PR #823, opened 2016-06-14
Document support for container cluster managers Mesos and Kubernetes.

mattray:
Mesos is thoroughly documented, Kubernetes is commented out and I'm still updating in a separate PR.


mattray:
@juliandunn fixes included but the switch from private->public orphaned this PR apparentely. Picking it back up in #827

PR #890, opened 2016-06-16
[Hold] Adds Homebrew Cask to Get Habitat page

ryankeairns:
Adding homebrew cask commands to the Get Habitat page on the website.

Signed-off-by: Ryan Keairns <rkeairns@chef.io>

### DIFF
--- a/www/source/docs/get-habitat.html.md
+++ b/www/source/docs/get-habitat.html.md
@@ -16,6 +16,14 @@ _Habitat for Mac requires a 64-bit processor running Mac OS X version 10.9+._
 <a class="button" href="https://api.bintray.com/content/habitat/stable/darwin/x86_64/hab-$latest-x86_64-darwin.zip?bt_package=hab-x86_64-darwin">Download Habitat for Mac OS X</a>
 <a class="button secondary" href="https://www.docker.com/products/docker-toolbox">Download Docker Toolbox</a>
 
+#### Homebrew Cask
+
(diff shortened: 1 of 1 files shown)


thesentinels:
By analyzing the blame information on this pull request, we identified @adamhjk to be a potential reviewer

smith:
If we're going to tell people to use homebrew, do we have a way for them to easily upgrade to the latest version without us having to be blocked by homebrew merging a PR? I don't know if we should list this as a supported thing unless it gives users a reliably good experience.

@juliandunn any thoughts on that?

juliandunn:
I have no idea as I'm actually not a Homebrew user - I still use MacPorts :)

ryankeairns:
@schisamo do you have any thoughts on Nathan's question?

schisamo:
@ryankeairns I don't think you can just have a generic Homebrew or Cask formula that always just always points to the latest version of something.

smith:
We're not too far from having a `hab selfupdate` type of feature, where right now you can do `hab install core/hab/X.Y.Z/YYYYMMDDTTTT` to get the latest, we could just ship a version that can upgrade itself when needed then never have to worry about it again.

Until then I'm not 100% sure about telling our users to use homebrew if they end up having a lousy experience.

ryankeairns:
I second that and vote that we wait until such a solution for upgrading is in place before promoting this.

I'll close this PR for now and bring it back once we get the update capabilities in place. Until then it'll be available on Cask just not promoted by us here on this page.

Thanks for your input!

PR #902, opened 2016-06-17
Updates to Travis for website deploy

smith:
* We've switched from pro to non-pro travis, so all of the secure keys
  have been regenerated
* Update the bucket policy to add the thing that was causing the deploy
  to the new bucket on the habitat org to fail
* Remove the slack notification, because it's disabled anyway (Homu
  posts notifications to Slack instead)
* Set `path_style = false` in the s3_sync; `true` is only needed if you
  have special characters in your bucket name; we don't
* Remove the Travis deploy for the bucket on the CIA account (currently
  the active website.)

Deploying to the new bucket will not work until the terraform change to
the bucket policy is applied.

This makes it so the website is not deployed to the bucket it's being
served from now, so the Fastly service will need to be pointed at the
new bucket once the deploy succeeds.

gif-keyboard-10398425249175422067

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,9 +135,6 @@ matrix:
     script:
     - "(cd components/builder-web && npm run travis)"
     - sudo -E bash -c "./support/ci/build_builder_web.sh"
-    #after_success:
-    #- "hart=$(source ./results/last_build.env && echo $pkg_artifact)"
-    #- "if [ -n \"$hart\" ]; then hab pkg upload ./results/$hart; fi"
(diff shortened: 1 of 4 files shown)


thesentinels:
By analyzing the blame information on this pull request, we identified @cwebberOps, @adamhjk, @fnichol, @ryankeairns and @magwalk to be potential reviewers

metadave:
gif-keyboard-15441619571874564483

smith:
Manually merging.

PR #937, opened 2016-06-21
Only build the "auto" branch

smith:
The Homu readme (https://github.com/barosl/homu,
https://github.com/servo/homu), says to do `only: -auto` in .travis.yml,
but we have master and auto.

This probably will make it so we don't do that extra build on master.

gif-keyboard-4269918751022729145

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 branches:
   only:
-  - master
   - auto
 env:
   global:


thesentinels:
By analyzing the blame information on this pull request, we identified @cwebberOps, @fnichol and @adamhjk to be potential reviewers

smith:
hmmm this also makes it so pull requests don't build, which is not what we want.

PR #979, opened 2016-06-23
[plans/gcc] Add fortran to enabled languages

stevendanna:

### DIFF
--- a/plans/gcc/plan.sh
+++ b/plans/gcc/plan.sh
@@ -127,7 +127,7 @@ do_build() {
       --with-mpfr=$(pkg_path_for mpfr) \
       --with-mpc=$(pkg_path_for libmpc) \
       --with-native-system-header-dir=$headers \
-      --enable-languages=c,c++ \
+      --enable-languages=c,c++,fortran \
       --enable-lto \
       --enable-plugin \
       --enable-shared \


thesentinels:
By analyzing the blame information on this pull request, we identified @fnichol and @smith to be potential reviewers

stevendanna:
This adds about 8MB to the core/gcc download.  Most distributions split this out into a gfortran or gcc-fortran package and I'd be happy to do that here if we want.  

I'm also happy to just continue using my own origin for this: https://app.habitat.sh/#/pkgs/ssd/gcc/5.2.0/20160621134439 since I doubt it is widely needed.  I currently need it as it is required to package `R`.

adamhjk:
@fnichol what do you think we should do here?

fnichol:
For the moment, this seems fine by me. I suspect that after one or two more additions and then we may want to re-evaluate. The most important consideration is that this Plan is in our critical toolchain build path, so it must work and be well-behaved.

@stevendanna, from what you can tell this doesn't have any additional runtime or buildtime deps? I'll run this in a stage1 Studio to make sure that our initial builds won't have any surprises.

fnichol:
Well I got a build of gcc, although with a bunch of test suite failures. I'd suspect that's a missed setup/patch for fortran-ness but not a deal breaker. @stevendanna, did this patch lead to a working R package down the line?

stevendanna:
@fnichol I got a "working" R package as in it compiles and I can run trivial code.  I'm still working on other deps I need to do more robust testing

thesentinels:
:umbrella: The latest upstream changes (presumably c78a767) made this pull request unmergeable. Please resolve the merge conflicts.

reset:
@stevendanna we just moved the core plans over to their own [repository](https://github.com/habitat-sh/core-plans). Would you mind opening your PR over there and linking to this discussion? I'll close this one for now but that doesn't mean it's rejected ;)

PR #1043, opened 2016-07-06
Remove mapping of the REST API in dev shell

adamhjk:
When using the developer shell, you don't want the REST API mapped to
the outside. Doing so causes multiple development shells to fail to
start, which makes the dev shell way less useful.

Instead, if you need to hit the REST API inside a dev shell, you should
just fire up two shells and go nuts. :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,10 @@ ifneq ($(IN_DOCKER),)
 	docker_cmd := env http_proxy= https_proxy= docker
 	compose_cmd := env http_proxy= https_proxy= docker-compose
 	common_run := $(compose_cmd) run --rm $(run_args)
-	run := $(common_run) -p 9636:9636 shell
+	run := $(common_run) shell
 	docs_host := ${DOCKER_HOST}
 	docs_run := $(common_run) -p 9633:9633 shell
-	web_dev_run := $(common_run) -p 9636:9636 shell
+	web_dev_run := $(common_run) shell
 else
 	run :=
 	docs_host := 127.0.0.1


Makefile
```diff
@@ -20,10 +20,10 @@ ifneq ($(IN_DOCKER),)
 	docker_cmd := env http_proxy= https_proxy= docker
 	compose_cmd := env http_proxy= https_proxy= docker-compose
 	common_run := $(compose_cmd) run --rm $(run_args)
-	run := $(common_run) -p 9636:9636 shell
+	run := $(common_run) shell
 	docs_host := ${DOCKER_HOST}
 	docs_run := $(common_run) -p 9633:9633 shell
-	web_dev_run := $(common_run) -p 9636:9636 shell
```
fnichol: I believe this external mapping could be preserved if the intention was to run a web browser against this port.

adamhjk: It wasn't the intent on `9636` - that's the supervisor REST API port. The doc run mapping is preserved.

adamhjk: Or do you literally mean "I want to use the web browser to hit the REST API when I'm running the web dev environment?"

smith: That's why it was added. If you do `make start-web-dev`, you can hit the API and web app on 9636.

Before this I was running with the ugly hack Timberman showed me:

`env HAB_DEPOT_URL="https://willem.habitat.sh/v1/depot -p 9636:9636" make shell`

(notice how the quotes are set up on that var, to, with trickery, pass `-p` into `docker-compose`.)

What I've been doing since running multiple `make shell`s was broken is doing `docker run -it ID bash` to open new shells in the same container.

I don't have much of an opinion on how to do things here, just showing how I managed to get by before and after.

adamhjk: I don't care as much about the web dev portion having the port forwarded, since one assumes you don't run more than one of them at a time. :)

fnichol:
@thesentinels r+

smith:
@thesentinels r+

thesentinels:
:pushpin: Commit 30d98d2 has been approved by `smith`

thesentinels:
:hourglass: Testing commit 30d98d2 with merge bf60b82...

thesentinels:
:sunny: Test successful - [travis](https://travis-ci.org/habitat-sh/habitat/builds/143144120)

PR #1049, opened 2016-07-11
[plan-build] Add PKG_CONFIG_PATH metadata to packages

stevendanna:
Plans can use the `pkg_pconfig_dirs` variable to indicate that they
export pkg-config configuration. This variable will then be used to
populate PKG_CONFIG_PATH during the build.

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -157,6 +157,14 @@
 # pkg_bin_dirs=(bin)
 # ```
 #
+# ### pkg_pconfig_dirs
+# An array of paths, relative to the final install of the software,
+# where pkg-config metadata (.pc files) can be found.  Used to populate
+# PKG_CONFIG_PATH for software that depends on your package.
+# ```
+# pkg_pconfig_dirs=(lib/pkgconfig)
+# ```
+#
 # ### pkg_svc_run
 # The command to start the service, if needed. Should not fork!
 # ```
@@ -1667,6 +1675,16 @@ _build_environment() {
         export LDFLAGS="$trimmed"
       fi
     fi
+
+    if [[ -f "$dep_path/PKG_CONFIG_PATH" ]]; then
+      local data=$(cat ${dep_path}/PKG_CONFIG_PATH)
+      local trimmed=$(trim $data)
+      if [[ -n "$PKG_CONFIG_PATH" ]]; then
+        export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:${trimmed}"
+      else
+        export PKG_CONFIG_PATH="$trimmed"
+      fi
+    fi
   done
 
   # Create a working directory if it doesn't already exist from `do_unpack()`
@@ -1680,6 +1698,7 @@ _build_environment() {
   build_line "Setting CXXFLAGS=$CXXFLAGS"
   build_line "Setting CPPFLAGS=$CPPFLAGS"
   build_line "Setting LDFLAGS=$LDFLAGS"
+  build_line "Setting PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
   return 0
 }
 
@@ -1793,6 +1812,7 @@ do_default_install() {
 #
 # * `$pkg_prefix/BUILD_DEPS` - Any dependencies we need build the package
 # * `$pkg_prefix/CFLAGS` - Any CFLAGS for things that link against us
+# * `$pkg_prefix/PKG_CONFIG_PATH` - Any PKG_CONFIG_PATH entries for things that depend on us
 # * `$pkg_prefix/DEPS` - Any dependencies we need to use the package at runtime
 # * `$pkg_prefix/EXPOSES` - Any ports we expose
 # * `$pkg_prefix/FILES` - blake2b checksums of all files in the package
@@ -1856,6 +1876,18 @@ _build_metadata() {
   done
   if [[ -n "${cxxflags_part}" ]]; then
     echo $cxxflags_part > $pkg_prefix/CXXFLAGS
+  fi
+
+  local pconfig_path_part=""
+  for inc in "${pkg_pconfig_dirs[@]}"; do
+    if [[ -z "$pconfig_path_part" ]]; then
+      pconfig_path_part="${pkg_prefix}/${inc}"
+    else
+      pconfig_path_part="${pconfig_path_part}:${pkg_prefix}/${inc}"
+    fi
+  done
+  if [[ -n "${pconfig_path_part}" ]]; then
+    echo $pconfig_path_part > $pkg_prefix/PKG_CONFIG_PATH
   fi
 
   local path_part=""
--- a/www/source/docs/reference/plan-syntax.html.md
+++ b/www/source/docs/reference/plan-syntax.html.md
@@ -119,6 +119,15 @@ that depends on your package.
   pkg_bin_dirs=(bin)
   ~~~
 
+pkg_pconfig_dirs
+: Optional. An array of paths, relative to the final install of the
+software, where pkg-config metadata (.pc files) can be found. Used to
+populate `PKG_CONFIG_PATH` for software that depends on your package.
+
+  ~~~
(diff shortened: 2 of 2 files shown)


stevendanna:
Like I said in my original report, it isn't clear to me that this is worth introducing a version incompatibility into the plan syntax (i.e. some plans might not build unless they use a version of plan-build that includes this); however, I didn't want to leave the idea hanging in my head so here is a PR :D.  I've noticed that at least a few plans in core aren't being built correctly because of missing PKG_CONFIG_PATH setup

miketheman:
Related: https://github.com/habitat-sh/core-plans/pull/37#discussion_r70341979

reset:
This is something that I've wanted for a while. This might require a version bump but that's alright!

@thesentinels r+

thesentinels:
:pushpin: Commit df94013 has been approved by `reset`

thesentinels:
:hourglass: Testing commit df94013 with merge 8f961a7...

reset:
gif-keyboard-15573287912978953646

miketheman:
@reset I think you may have overlooked the related conversation - it's likely not needed if the other packages export bin for pkg-config to pick up.

metadave:
props for updating the docs!

thesentinels:
:sunny: Test successful - [travis](https://travis-ci.org/habitat-sh/habitat/builds/144162826)

reset:
@miketheman I saw the related conversation but it seemed specific to libpng since libpng contains the libpng-config binary. That's not the case with many other pieces of software, though. They have pkg-config definitions but not always an associated config binary. That sounds correct, yeah?

reset:
@miketheman here's my inspiration on that: https://github.com/libarchive/libarchive/tree/master/build/pkgconfig

libarchive is a piece of software that I built in the early days of Habitat and really wanted to export pkg_config but never had the time to go and do it!

miketheman:
@reset I guess I was expecting a little more discussion before accepting a public api extension.
If anything, I think I'd expect this to automatically do thr Right Thing, similar to C*FLAGS and automatically create the PKG_CONFIG_PATH for me at the well-established patterns of locations, and thereby remove the need for the human to do anything out of the ordinary.

reset:
@miketheman if you'd like to take a crack at a better implementation I'm all for it. This PR contained the implementation that I had in my mind so it was easy to accept as is.

The plan definition is declarative all around. For example, we ask you to specify lib and bin paths and don't provide any defaults. The CFLAGS are only auto-populated correctly if you specify a lib path. This PR mimics that behaviour by populating the PKG_CONFIG_PATH with the contents of the metafile in the same manner as CFLAGS.

What do you think we're missing here?

reset:
@miketheman the implementation in this PR is inline with the declarative nature of the plan file. For example: the CFLAGS metadata is only populated if you specify a value for `pkg_lib_dirs`. This feels the same way to me: declare some paths that contain pkg_config definitions and the metadata will be populated and present for this and any other plans consuming the target as a dependency.

This was a pretty easy PR to accept because it's exactly how I would have done it. If you have a better implementation or a different idea I'm open to continue the conversation of course.

(note: apologies if you get double pinged on this message, it seems my original response got deleted or the hotel wifi is playing TRICKS!)

miketheman:
@reset I understand the approach, indeed.

I think that now that we have at least 4 instances in which we're exporting environment variables to be accessible for other packages, and this one is the sole one that contains an API entrypoint (the others are magical), if feels like there's a better way.

Otherwise we end up with more api hooks, knobs and tunables, etc ad nauseum.

Maybe we should attempt to auto-populate PKG_CONFIG_PATH in a similar manner?

reset:
But the others aren't magic, they still need you to specify the lib/bin paths before they get populated.

Other than recurring through the source tree to find .pc files I don't see how we could automagically populate it. Is that what you're suggesting or do you have another idea?

miketheman:
I guess that's the "magic" part.

How about instead of creating an API extension, the PKG_CONFIG_PATH variable is automatically set from any packages exporting a `lib`, since convention places the .pc files in `pkg_path/lib/pkgconfig/` ?
Similar to how the CFLAGS variables work, it's exported by the source package as a lib dir, and then hab auto-builds any related paths.

I'm sorry if I'm harping on this, but the "it has to be easy" principle of habitat feels like the more knobs created, the worse off it gets, because you have to "know more" to solve a pkg-config issue of an upstream plan that you're likely not responsible for, you only want to build your own thing.

reset:
@miketheman absolutely no reason to apologize, it probably isn't coming over properly in text but I'm happy with the conversation we're having and I want us to come to a conclusion that feels right.

It won't work to just search the lib dir for `.pc` files. Some software may drop off the `.pc` files in there, but according to the [offical guide](https://people.freedesktop.org/~dbn/pkg-config-guide.html) there's no standard place in the source tree of a package for where the `.pc` files should live, just where they should be placed on package install.

For example: libarchive places it's `.pc` files in a directory `$src_root/build`. On install the `.pc` files are installed to one of standard locations: `/usr/lib/pkgconfig` (the other being `/usr/share/pkgconfig`).

I still think our only option would be to search for `.pc` files in the tree and add the paths they reside in to a metadata file which we read the contents from and append to `PKG_CONFIG_PATH`. My only reservation here is that everything else is declarative in nature and this would be the only METAFILE we populate automagically for you.

miketheman:
@reset In trying to figure out the source of my hesitation, I think it's similar to the problem surfaced in the [libpng package](https://github.com/habitat-sh/core-plans/pull/40) where, as a library, it was built, and exported the `lib/` directory faithfully, but since there was a binary involved to assist a downstream builder that wasn't explicitly exported, the yak needed shaving one step further before the downstream builder was able to accomplish their build.

In your use case about libarchive, we can see that the .pc files actually get to the "right place" during `make install` (sample from a quick build today):

     /hab/pkgs/core/coreutils/8.24/20160612075329/bin/install -c -m 644 build/pkgconfig/libarchive.pc '/hab/pkgs/core/libarchive/3.2.0/20160717125959/lib/pkgconfig'

This follows the convention of placing .pc in `lib/pkgconfig/*` for subsequent packages to pick up, which is probably why NixOS has an auto-setup hook like so: https://github.com/NixOS/nixpkgs/blob/7475728593a49f09c4b7b959b15513aee38ab4b4/pkgs/development/tools/misc/pkgconfig/setup-hook.sh

They add both potential locations of .pc files for each known any time a package declares a build dep on pkg-config:

>  pkg-config
> Adds the `lib/pkgconfig` and `share/pkgconfig` subdirectories of each build input to the `PKG_CONFIG_PATH` environment variable.

ref https://nixos.org/nixpkgs/manual/#ssec-setup-hooks

So in the libarchive scenario, the .pc files end up in the place we expect them to be, and setting the PKG_CONFIG_PATH dynamically (a lá C*FLAGS) with a set of "known locations" should allow them to be found, and reduces the amount of things a package maintainer needs to know about exporting for Someone Else's Software, and the overall public API.

stevendanna:
If we were to add magic, I think that my weak preference would be that we leave `pkg_pconfig_dir` as a configurable but give it a default value of `${pkg_lib_dirs[@]/%/\/pkgconfig}` which would cover the majority of the cases but allow package maintainers the ability to export a different pkgconfig dir.  I don't love the "search the tree for `*.pc` files unless there is a clear way for maintainers to turn that behavior off.

> reduces the amount of things a package maintainer needs to know about exporting for Someone Else's Software

I'm skeptical that this should be a goal. As a user of `core` I want `core/FOO` to be an expertly built version of `FOO` which means that the maintainer understands the package well.

miketheman:
@stevendanna anecdotally, if the "automatically add lib/pkgconfig to PKG_CONFIG_PATH" was in place for libpng, the "export the binary" wouldn't have been needed either, since it's also in the conventional location.

reset:
@stevendanna @miketheman ah yeah of course, it's post-installed location *should* become a standard location so we can auto-populate the variable which will make things easier.

@stevendanna we can either allow them to override or ask them to move the files to the proper location as part of the build process. I still like exporting the PKG_CONFIG_PATH metafile, though. I think we should keep that part to mirror CFLAGS and the others.

Issue #1039, opened 2016-07-06
Docs needed for do_check callback


miketheman:
The `do_check` callback is a powerful mechanism to validate the build process has completed successfully, and does not appear in the docs: https://www.habitat.sh/docs/reference/plan-syntax/#callbacks

adamhjk:
When we document this, lets make it clear that you need to set `env DO_CHECK=true`, not just `DO_CHECK`.

bscott:
This was resolved in PR#1062

miketheman:
Thanks! Using #1062 (without a `PR`) to autolink.

PR #1062, opened 2016-07-13
Resolves issue #1039 - do_check() documentation

bscott:
Adds `do_check()` callback documentation.

### DIFF
--- a/www/source/docs/reference/plan-syntax.html.md
+++ b/www/source/docs/reference/plan-syntax.html.md
@@ -255,6 +255,10 @@ do_download()
 do_verify()
 : The default implementation tries to verify the checksum specified in the plan against the computed checksum after downloading the source tarball to disk. If the specified checksum doesn't match the computed checksum, then an error and a message specifying the mismatch will be printed to stderr. You should not need to override this behavior unless your package does not download any files.
 
+do_check()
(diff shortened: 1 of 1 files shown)


bscott:
r+

metadave:
you'll need an @thesentinals before the r+

smith:
And it's not working right now. See https://habitat-sh.slack.com/archives/general/p1468417818008298

www/source/docs/reference/plan-syntax.html.md
```diff
@@ -255,6 +255,10 @@ do_download()
 do_verify()
 : The default implementation tries to verify the checksum specified in the plan against the computed checksum after downloading the source tarball to disk. If the specified checksum doesn't match the computed checksum, then an error and a message specifying the mismatch will be printed to stderr. You should not need to override this behavior unless your package does not download any files.
 
+do_check()
+: The default implementation runs post-compile and checks provided two conditions
```
smith: I don't think this is clear. The default implementation does nothing, and you're expected to put something like `make test`, or whatever runs your plan's tests after it's built, in it.

These listed in the order they are run, so `do_check` should go between `do_build` and `do_install`.

bscott:
@smith thanks for the feedback

smith:
Looks great!

giphy

PR #1073, opened 2016-07-18
fix_interpreter: warn+skip on non-regular files

### LINKED ISSUE #1045: Wildcard fix_interpreter throws error if the path contains directories
``` bash
fix_interpreter "$pkg_prefix/neo4j/bin/*" core/bash bin/bash
```
causes:
```
sed: couldn't edit /hab/pkgs/keyvan/neo4j-enterprise/3.0.3/20160707073749/neo4j/bin/tools: not a regular file
```

Right solution might be to ignore directories or at least accepting an ignore-directories flag.
Also a more appropriate error message would be appreciated.

srenatus:
Fixes #1045.

Signed-off-by: Stephan Renatus <srenatus@chef.io>

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1316,6 +1316,11 @@ fix_interpreter() {
     fi
 
     for t in ${targets}; do
+      if [[ ! -f $t ]]; then
+        debug "Ignoring non-file target: ${t}"
+        continue
+      fi
+
       build_line "Replacing '${interpreter_old}' with '${interpreter_new}' in '${t}'"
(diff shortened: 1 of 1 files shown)


components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1316,6 +1316,11 @@ fix_interpreter() {
     fi
 
     for t in ${targets}; do
+      if [[ ! -f $t ]]; then
+        warn "Ignoring non-file target: ${t}"
```
reset: This would be a common thing and not something we're actually try to call attention to. This might be better off being `debug` level instead of `warn` level

srenatus: Ok, updated.

smith:
Thanks!

gif-keyboard-11766646103904955700

PR #1079, opened 2016-07-18
Publish web site to acceptance

jtimberman:
This commit modifies the deploy_website.sh script to publish the web
site to acceptance when we're on a pull request.

### DIFF
--- a/support/ci/deploy_website.sh
+++ b/support/ci/deploy_website.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-
 git log HEAD~1..HEAD | grep -q '!!! Temporary Commit !!!'
 is_tmp_commit=$?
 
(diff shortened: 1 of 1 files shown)


thesentinels:
@jtimberman, thanks for your PR! By analyzing the annotation information on this pull request, we identified @smith to be a potential reviewer

smith:
Is this how our workflow is going to work? I thought that in capital-W Workflow you had already merged the change to master once the verify stage and approval was completed. Am I wrong about that?

I'm not opposed to this being the workflow, but I'd like to understand what we're going for.

jtimberman:
I'm not of a strong opinion about this being the workflow. However, unlike Chef's Workflow, here we don't have an additional gate between "acceptance" and "delivered" (live/production for non Chef-using observers). That is, here once we merge to master, that's it, right? Unless there's an additional thing that Homu can do, but I'm not super familiar with how it works.

adamhjk:
If we want to customize it any further, we'll have to do some custom deployment hooks. Right now, you're right - there is no way to deploy in to acceptance other than to just go straight to production.

jtimberman:
I've updated the script to publish to both acceptance and live. Is that what you mean, @adamhjk?

support/ci/deploy_website.sh
```diff
@@ -9,12 +8,17 @@ set -eu
...
 if [ "${TRAVIS_PULL_REQUEST}" = "false" ] &&
    [ "${TRAVIS_BRANCH}" = "auto" ] &&
    [[ $is_tmp_commit = 1 ]]; then
+  AWS_BUCKET=habitat-www-acceptance ./bin/middleman s3_sync
   ./bin/middleman s3_sync
-  curl -H "Fastly-Key: ${FASTLY_API_KEY}" -X POST "https://api.fastly.com/service/${FASTLY_SERVICE_KEY}/purge_all"
-else echo "Not on master; skipping website deploy"; fi
+  # If we have published, then we need to purge the CDN cache.
+  curl -H "Fastly-Key: ${FASTLY_API_KEY}" \
+       -X POST "https://api.fastly.com/service/${FASTLY_SERVICE_KEY}/purge_all"
```
smith: I think this will only purge the live site, since the `FASTLY_SERVICE_KEY` is the part of the URL that changes on the Fastly UI when you switch between the different services.

smith:
I don't think it makes much sense to have an acceptance environment for the website that's exactly the same as production. Wouldn't it just be easier to make a redirect or alias from www.acceptance to www?

Having an acceptance environment to deploy the depot, build service, and web app is very important. Having one for the website is something we should absolutely do at some point, but right now I think it makes more sense just to close this PR and leave it work the way it does.

jtimberman:
👍

PR #1081, opened 2016-07-19
[plan-build] Use `make -j` in default steps

smith:
The `make` man page says:

> If the -j option is given without an argument, make will not limit the
> number of jobs that can run simultaneously.

Using `make -j` by default makes it use all available cores and
noticeably speeds up build times.

Not sure if this is the correct thing to do, but if it is, let's merge
it!

gif-keyboard-14521149750607829157

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1760,7 +1760,7 @@ do_build() {
 # Default implementation for the `do_build()` phase.
 do_default_build() {
   ./configure --prefix=$pkg_prefix
-  make
+  make -j
 }
 
 # Will run post-compile tests and checks, provided 2 conditions are true:
@@ -1810,7 +1810,7 @@ do_install() {
 
 # Default implementation for the `do_install()` phase.
 do_default_install() {
-  make install
+  make -j install
 }
 
 # **Internal** Write out the package data to files:


metadave:
:+1:, gif-keyboard-5399516555533657948

might be worth one more set of eyes before merging

miketheman:
My current understanding is that with no number, there's no limit to the amount of concurrent jobs - so this may saturate a system to the point at which nothing can progress reasonably.

I've seem use of calculating `-j(cpu count + 1)` as well as using the `-l 2.5` flag to limit tasks when the load average rises above 2.5.

miketheman:
There also exists the "problem" that any plan that overrides the `do_build` phase will not benefit from the defaults - not a crazy problem, but something that we could explore using the `MAKEFLAGS` env var if that makes more sense.

stevendanna:
I've been using `make -j $(nproc)`.  However, if we do this by default,  I think we'll definitely want to provide some sort of flag or variable since a number of pieces of software have problems with parallel builds.

smith:
And with `nproc` you need a build dep on coreutils.

My vote is use `-j`, and if it causes anybody problems, we can try to generalize making the concurrency level configurable throughout all the callback defaults. I don't have strong feelings either way though.

metadave:
I like both `-j` and  `make -j $(nproc)`, it looks like `nproc` works in the studio as well.

for reference:
```
[core] hab_dev:root:/src$ command -v nproc
/usr/bin/nproc
```

fnichol:
I'd like to take a quick re-survey of several other packaging systems, namely ArchLinux and NixOS before committing to this default. If memory serves, there were good reasons to leave this up to the Plan author…

scassiba:
I've been experimenting with `-j` with and without `nproc` with a plan I'm writing.

I find that if I don't give an argument to `-j` (such as `$(nproc)`) then `make` will endlessly spawn new processes until they start getting killed due to memory exhaustion, resulting in failed builds. I found this was the case even when increasing the memory allocated to Docker to as much as my host computer would allow while still being (barely) usable. My friends, that is not good eats.

If there has to be a change to the defaults, I highly suggest not using `-j` without arguments. Using make jobs with `nproc` is friendlier to the system and operator, whether or not they know about it.

miketheman:
Looking at the [Homebrew package builder](https://github.com/Homebrew/brew/blob/2e747aa910ccfd645388a52517f285047a3276fe/Library/Homebrew/extend/ENV/super.rb), the default is to:

- determine how many hardware cpu cores there are, set to at least 1
- set the `MAKEFLAGS` environment variable to `MAKEFLAGS="-j{{cores}}"`
- allow overriding both the `MAKEFLAGS` env var, as well as calling `ENV.deparallelize` in user 'plans' which is a helper for the same.

This way, the defaults are "use as many slots as there are cores" - similar to `make -j$(nproc)` while allowing for non-unlimited task slots, and a user interface to change it.

I think this might end up complicating the code more than the proposed PR, but will ultimately be safer.

miketheman:
NixOS does a similar fashion of "use envvar, detect with `nproc`, default to 1, export envvar" approach, as can be [seen here](https://github.com/NixOS/nixpkgs/blob/3afa2460386b1472128e302d1db13dba54e68e12/pkgs/stdenv/generic/setup.sh#L360-L374).
The variable is then used in the `build`, `check`, and `install` phases by passing the value to `-j` to `make`.
All the while checking for the value of `enableParallelBuilding`, which allows the user to set this to false to override the behavior, similar to brew's `ENV.deparallelize`.

In NixOS, all of this assumes that the default hooks are used and are not overridden by the end user - something which we see plenty of customization in the core packages so far.

Nix also provides a `configure` phase, one that we have bundled together in habitat with the `do_build()` hook - making the decision less clear if this is the right path to follow.

metadave:
I like @miketheman's proposal ala Nix.

smith:
Closing this. I agree that the approach from Nix that Mike suggested would be the best thing to do, but not going to implement it in this PR.

PR #1091, opened 2016-07-20
Set the SSL_CERT_FILE for the director

jtimberman:
We need to set the SSL_CERT_FILE in the environment in order for the
processes started by the director to know where to find it, so that
HTTPS requests can have the SSL certificates verified. This change was
done manually on "live." When we set up the "acceptance" environment,
the Sign In process was throwing a 500 that looks like this:

ERROR:habitat_builder_api::http::handlers: github authentication, err=HyperError(Ssl(OpenSslErrors([UnknownError { library: "SSLroutines", function: "ssl3_get_server_certificate", reason: "certificate verify failed" }])))

This isn't because we are using a self-signed certifcate on acceptance
though, it's because the application couldn't verify GitHub's
certificate because it didn't know about the CA certs.

This change will hardcode the release-specific cert.pem, so if that
changes, it will probably fail on startup in the future.

### DIFF
--- a/terraform/files/hab-director.service
+++ b/terraform/files/hab-director.service
@@ -2,6 +2,7 @@
 Description=Habitat Director
 
 [Service]
+ExecStartPre=/bin/bash -c "/bin/systemctl set-environment SSL_CERT_FILE=$(hab pkg path core/cacerts)/ssl/cert.pem"
 ExecStart=/bin/hab-director start -c /hab/etc/director/config.toml
 
 [Install]


thesentinels:
@jtimberman, thanks for your PR! By analyzing the annotation information on this pull request, we identified @reset to be a potential reviewer

reset:
As you noted this will break as soon as a new release of `core/cacerts` is released. This can dynamically be populated in a few different ways, but here's [a gist](https://gist.github.com/nickjacob/9909574) that has a particularly good solution with some commentary below it.

It's probably safe enough to dynamically set this to whatever the latest version of cacerts is. Technically we should be finding the exact version of cacerts that the package we're starting was built with, but at this level we don't know what the director is starting from this file.

This is a good temp fix but I need to figure out a solution in the plan/software for the longer term

reset:
Actually, this might be as simple to fix as setting `core/cacerts` as a `pkg_dep` and not just a `pkg_build_dep` in the builder-api plan

smith:
There's discussion on #1075 that is directly related to this.

fnichol:
I'm guessing we should be constructing our own SSL context when standing up the Hyper/Iron HTTP server much like what we had to do with HTTP client-side [setups](https://github.com/habitat-sh/habitat/blob/9e25233fafe0433864ab47d0d8ede6344d30be8a/components/http-client/src/lib.rs#L33-L76).

On a philosophical level, I'd argue that the package should handle its own SSL needs itself, otherwise it has leaked its dependencies back outside into the external OS. Having said that, in the short term, adding an explicit `core/cacerts` to the `$pkg_deps` array would at least ensure that we have such a package installed.

jtimberman:
This is working in acceptance.

reset:
I think this is good enough for now but we should follow through and make sure the app "just works" without us setting this environment variable outside of it

@thesentinels r+

PR #1095, opened 2016-07-22
api-shell make target

smith:
`make api-shell` starts a devshell with redis and the builder-api,
builder-router, builder-sessionsrv, and builder-vault services running.

An `api` script that can start, stop, and restart these services is
available in the shell.

With this running, along with using the new defaults in
components/builder-web/habitat.conf.sample.js, you can run a local
Builder API service and have the locally running (using Node on your
local machine on port 3000) web app connect to it.

This replaces `make start-web-dev` and adds Redis to the Dockerfile for
the devshell image.

## How is this different than `make shell`?

`make shell` can be run multiple times and start multiple dev shells,
which is great if you want to be working on individual components inside the
shell.

I does not expose any ports, because there would be conflicts if you
tried to run more than one instance.

`make api-shell` exposes port 9636, and is meant to be run once, but you
can still use `cargo` and everything else that you can in `make shell`,
so somebody working on the API can make changes, recompile, and run them
while being able to connect to the API service.

## How is this different than `make start-web-dev`?

`make start-web-dev` ran the Habitat packages for these services, so you
could have a locally running API, but did not give you the ability to
easily make changes to those components. It ran the web app inside the
director, while `make api-shell` expects you to run the web app locally
on your machine (we should add the ability to run the web app inside a
container at some point, but this functionality is not included here.)

gif-keyboard-7212439022391031913


smith:
@thesentinels try

smith:
@thesentinels try

reset:
@smith looks great! Merge when ready :)

PR #1103, opened 2016-07-25
Move Core Plans Maintainers back to main MAINTAINERS.md

smith:
Feels like it would be better to have just one of these files.

gif-keyboard-10564341984700588675


smith:
Closing. See habitat-sh/core-plans#95.

PR #1104, opened 2016-07-25
[core] add quickcheck and ServiceGroup parsing property

metadave:
Hi!

Now that the [Rust implementation of QuickCheck](https://github.com/BurntSushi/quickcheck) (aka "QC") doesn't require a [nightly Rust build](https://users.rust-lang.org/t/quickcheck-has-a-macro-on-rust-stable/6636), I'm including it in Habitat core. 

The first property that I've included generates service, group, and an optional organization to pass into `ServiceGroup::from_str()`. While testing, I found several bugs with the current regex we're using.

If you'd like to see some of the crazy Strings that QC generates,
run the test:

	cargo test -- prop_service_group_from_str --nocapture
	# some of the output may mess up tmux etc

### DIFF
--- a/components/core/src/lib.rs
+++ b/components/core/src/lib.rs
@@ -26,6 +26,9 @@ extern crate rustc_serialize;
 extern crate sodiumoxide;
 extern crate libsodium_sys;
 #[cfg(test)]
+#[macro_use]
+extern crate quickcheck;
+#[cfg(test)]
 extern crate tempdir;
 extern crate time;
 extern crate toml;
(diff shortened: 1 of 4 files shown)


thesentinels:
@metadave, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol and @reset to be potential reviewers

components/core/src/lib.rs
```diff
@@ -26,6 +26,9 @@ extern crate rustc_serialize;
 extern crate sodiumoxide;
 extern crate libsodium_sys;
 #[cfg(test)]
+#[macro_use]
```
adamhjk: Can these be on the same line? I want to say you can do:

```rust
#[cfg(test), macro_use]
```

metadave: My vote would be to change those across all components, as we don't use the single line attribute elsewhere. I'd do this in a separate PR.

thesentinels:
:umbrella: The latest upstream changes (presumably #1136) made this pull request unmergeable. Please resolve the merge conflicts.

PR #1123, opened 2016-08-04
Copy no_proxy env variable if set

stephenbm:
This just adds the no_proxy env variable to the hab studio if it's set (same as http and https proxy settings)


thesentinels:
@stephenbm, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol, @mattray and @adamhjk to be potential reviewers

PR #1133, opened 2016-08-08
add Habitat testing framework

metadave:
This PR adds a crude Inspec/Rspec based testing framework for Habitat.

- `core/net-rools` and `core-inspec` from [this PR](https://github.com/habitat-sh/core-plans/pull/128) **MUST** be merged and uploaded to the Depot in order for the tests to work.
    - if you'd like to test before merging, you can set `INSPEC_PACKAGE=metadave/inspec` in `test.sh`. I've pushed `metadave/net-tools` and `metadave/inspec` to the Depot for testing only.
- The testing framework uses a combination of Inspec and RSpec. 
   - initially, Inspec is used to see if the testing environment is "clean"
   - most testing is performed via RSpec
   - In the future, we can run Inspec tests _remotely_ against Docker containers etc via a command like `inspec exec ./controls/ -t docker://f028618d58aa`, _however_, Rspec tests **WILL NOT** be executed remotely.
- Mixlib::Shellout - the `ctx.cmd_expect` function does not use Mixlib::Shellout. `cmd_expect` is designed to match output (possibly) long running processes, such as `hab start core/redis`. 
   - `cmd_expect` currently matches against an exact String, however, it could easily be made to match against regexes. 
- Travis stuff will come in another PR.

As this is a first pass on a testing framework, it's not perfect. Please see the readme for more info!


thesentinels:
@metadave, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol, @adamhjk, @reset, @smith and @jtimberman to be potential reviewers

test/banner
```diff
@@ -0,0 +1,33 @@
+                                    _    _       _     _ _        _
```
reset: What happen to Willem's ASCII face? :trollface:

test/test.sh
```diff
@@ -0,0 +1,111 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+# The list of all specs to run after basic and env tests are run.
+# WITHOUT .rb suffix.
+all_specs=(crypto)
+
+# sadly, this is NOT a banner of a cat. But, with a pull request,
```
reset: gif-keyboard-10866906255235050798

test/README.md
```diff
@@ -0,0 +1,149 @@
+# Habitat testing
+
+## Running all tests
+
+First, ensure that all Habitat related environment variables are unset:
```
reset: Is there instead a way for us to run these tests in a "clean room" environment? I think this could become tedius since the developers are also habitat users

reset:
@metadave I tried to run the suite by following the instructions in the README and this is what I got:

screen shot 2016-08-08 at 12 29 01 pm

By reading the `test.sh` file I can see that we generate a log and got my answer there. Maybe we want to message the user on failure to check the logs or pipe the last X lines or the failed lines to the console?

test/test.sh
```diff
@@ -0,0 +1,111 @@
...
+# YOU could make it so.
+cat banner
+
+# load in common test env vars
+HAB=/bin/hab
```
reset: Should we try to look for this by using their `PATH` instead? Our OSX installer puts the binary to `/usr/local/bin/hab` instead of `/bin/hab`.

(actually, should I be running this test suite on my local machine or inside a docker container?)

test/README.md
metadave: I originally started implementing the test stuff so it could run completely safely in any environment, but there were issues with:

- tl;dr - making the test suite use it's own safe env was extremely complicated
- managing a separate cache of keys inside and outside of a studio
  - setting the `HAB_CACHE_KEY_PATH` env var isn't passed on through to a studio, and the paths wouldn't be the same inside and outside of the studio anyways.
- trying to set alt `HAB_STUDIOS_HOME`, `HAB_STUDIO_ROOT`, `HAB_ROOT` didn't provide the desired results.

we talked a bit on Slack about making a separate Makefile target such as `make test-shell`, but in the end, it seemed like `make shell` got us what we needed.

metadave:
@reset until the `core/net-tools` + `core/inspec` core-plans PR is merged, you need to
 set `INSPEC_PACKAGE=metadave/inspec` in `test.sh`. I've pushed `metadave/net-tools` and `metadave/inspec` to the Depot for testing only.

I simply piped the install cmd output to a log file, as it was displaying a lot of not-very-helpful log information. Regular test output is detailed, and if you want a ton of detail, you can set `HAB_TEST_DEBUG=1`

test/spec/basic.rb
```diff
@@ -0,0 +1,129 @@
...
+
+require 'pathname'
+
+# TODO: .rspec file doesn't seem to be honored, so we need
+# to manually include the spec_helper here
+require_relative 'spec_helper'
+
+describe "Habitat CLI" do
+    # TODO: maybe extract this into a module in the future
+    before(:all) do
```
reset: Ruby's indentation standard is 2 spaces. Forgive me for including stylistic comments in a PR
gif-keyboard-1300383516384764049

test/test.sh
metadave: I've never tried it on my local machine, as the supervisor doesn't work there. However, I specifically call out `/hab/bin` in case we want to change it to a compiled version that sits in `/src/components/hab/target/debug/hab`. Theres definitely more shell stuff we can do here to make it more pleasant, but I'd like to iterate via separate PR's.

test/spec/basic.rb
metadave: lol, wtf, I formatted all .rb files with Vim's Ruby formatter.

reset:
Changing the location of my hab binary to `/usr/local/bin/hab` got me this far:

screen shot 2016-08-08 at 12 39 01 pm

It doesn't look like any tests actually ran but there's a success error code. I'm running this on my OSX machine and not inside a Docker container. I might be doing it wrong!

metadave:
@reset I definitely didn't target OSX as the first platform that this could run on

reset:
@metadave I got it all running on my machine! Looks good, I'd make a few modifications to the README and then merge on your own when ready:

* Specify that you should be in `make shell`
* Specify that you should be in the `test` directory before running anything

metadave:
thanks @reset, README.md updated.

PR #1136, opened 2016-08-09
Use Cargo "workspaces" to manage a project-wide Cargo.lock file

reset:
There is now a Cargo.toml and Cargo.lock file at the root of the Git project. This is what the
Cargo RFC refers to as a "Virutal Manifest". This references all other crates and they reference
back to the virtual manifest. All crate targets are now built to $root/targets instead of a
target directory in their respective crate.

Workspaces were officially added to Cargo in version 0.12 which will be included in Rust 1.11. You
can, and should, download Cargo 0.12 or greater to compile our project. For more information about
how to use workspaces see the docs or RFC:

* [Documentation](http://doc.crates.io/manifest.html#the-workspace-section)
* [RFC](https://github.com/rust-lang/rfcs/blob/master/text/1525-cargo-workspace.md)


thesentinels:
@reset, thanks for your PR! By analyzing the annotation information on this pull request, we identified @metadave, @fnichol, @adamhjk and @smith to be potential reviewers

fnichol:
Also of use is the [manifest doc page](https://github.com/rust-lang/cargo/blob/master/src/doc/manifest.md#the-workspace-section) out of master.

PR #1114, opened 2016-07-28
Rename 'views' to 'channels'

reset:
* Depot API endpoints renamed from `/views` to `/channels`
* `/views` endpoints remain as an alias
* Update docs to reflect naming changes

Signed-off-by: Jamie Winsor <jamie@vialstudios.com>

PR #1141, opened 2016-08-10
Adding routes for downloading packages and keys from views

nsdavidson:
I was trying to experiment with views/channels, but found that packages and keys failed to download when starting the supervisor with `--url http://mydepot/v1/depot/views/myview`.  Not sure if this is the implementation that was planned, but it's similar to what's in #1114 for adding the channel routes.  

Running this locally allows views to work as I would expect.  Starting a supervisor with the `--url http://mydepot/v1/depot/views/myview` and `--strategy at-once` options allows only the promoted version of a package available in `myview` to be deployed.  Also, promoting a new version to the view prompts the supervisor to update.

Signed-off-by: Nolan Davidson <ndavidson@chef.io>

### DIFF
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -1304,6 +1304,15 @@ pub fn router(depot: Arc<Depot>) -> Result<Chain> {
     let depot33 = depot.clone();
     let depot34 = depot.clone();
     let depot35 = depot.clone();
+    let depot36 = depot.clone();
+    let depot37 = depot.clone();
+    let depot38 = depot.clone();
+    let depot39 = depot.clone();
+    let depot40 = depot.clone();
+    let depot41 = depot.clone();
+    let depot42 = depot.clone();
+    let depot43 = depot.clone();
+
 
     let router = router!(
         get "/channels" => move |r: &mut Request| list_channels(&depot28, r),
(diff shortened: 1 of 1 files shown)


reset:
@nsdavidson The changes that you made here conveniently work because of how the URLs layout, but the `--url` flag's initial intent was to be the exact URL to the depot's root and not scoped by the view/channel. Your implementation here might be the right way to go, though. Sticking with the name "channel", my original intent was to expose a `--channel` flag to scope things properly. When you don't specify a channel you're using the "default channel" which is the entire pool of packages. I'm going to chat with the team and see if we want to still expose that flag or go this route instead.

nsdavidson:
@reset Sounds good.  

I should have mentioned initially that I based the approach of using the `--url` to point to a specific channel from the current docs (https://www.habitat.sh/docs/share-packages-overview <- bottom of the page).  If we go the route of exposing a separate flag to specify that, we need to make sure the docs get updated too.

reset:
@nsdavidson thanks for pointing that out! I'll make sure the docs are up to date as well, I think what is there right now snuck through during our launch craziness even though it wasn't just yet ready ;)

thesentinels:
:umbrella: The latest upstream changes (presumably #1150) made this pull request unmergeable. Please resolve the merge conflicts.

metadave:
@reset @nsdavidson this PR at least needs a rebase, is it still valid with the latest changes on master?

PR #1150, opened 2016-08-12
Rename depot/depot-client to builder-depot/builder-depot-client

reset:
The depot and depot-client are actually components of Builder, not Habitat. This is a simple directory rename for now, but the crate and Habitat packages will also be renamed when we push the next major version of Builder


metadave:
@thesentinels r+

sticking with the magic theme for the rest of the day:
gif-keyboard-2861608032444231872

thesentinels:
:pushpin: Commit e4b410b has been approved by `metadave`

PR #1163, opened 2016-08-16
Habitat publishing guide and program fixes

reset:
* Repair the `bintray-publish` executing environment by exporting the necessary
  package's bin paths in both of the bintray-publish binaries
* `publish-studio` no longer accepts a package identifier *or* a path to a `.hart`
  file. This was necessary when releasing for Mac because we would attempt to install and then
  later execute the Mac version of `hab` from within our Linux container
* Documentation updates
* Moved bintray-publish from the support directory to the components directory as it's a component of hab even though it's a part of our release process


thesentinels:
@reset, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol to be a potential reviewer

Issue #1120, opened 2016-08-03
Allow ignoring of certain file types (e.g. *.swp) in config/ dir during build


mivok:
When editing files with vim, it opens up a filename.swp file while the editor is open. If you build a package while the editor is open on a config file (as you do often during development), the config/config.json.swp file ends up included in the final package, resulting in the following error when you go to run the package with hab start:

```
DEBUG:habitat_sup::service_config: Config template .config.json.swp at "/hab/pkgs/chefops/foo/0.1.0/20160803162046/config/.config.json.swp"
hab-sup(ER)[src/error.rs:322:8]: IOError(Error { repr: Custom(Custom { kind: InvalidData, error: StringError("stream did not contain valid UTF-8") }) })
```

It would be nice to be able to specify a list of files to exclude from the package, possibly via environment variable that is passed to the studio (so you could just put something like `export HAB_CONFIG_EXCLUDE="*.sw* *~ *.bak"` in your .bashrc and have it work), so that these files are not included in the final package.

It might also be worth having defaults for common editor swap and backup files in the excludes too, so that people don't run into this issue by default.

efyang:
So to be clear you would have a list of defaults that would be overridden if the environment variable was set?

mivok:
Yes, although I haven't thought through the implementation in detail, so it's quite possible there's a better way.

efyang:
Alright, I'll try making a basic implementation of it this week.

chefsalim:
Hey @efyang that is awesome, let us know if you need any help!  This work is on our radar/list to work on, let me know if you don't get to it in the next day or so...

efyang:
@chefsalim I should be able to get started on it tomorrow or the day after, I just need a little time to get acquainted with the internals and the internal structure. I'll definitely ask if I have questions though, thanks!

efyang:
[...]

PR #1171, opened 2016-08-17
files in a packages config directory are logged upon parse failure

metadave:
If an editor leaves a backup/swap file in the `./config` dir of a plan, the supervisor will fail to start with:
```
hab-sup(ER)[src/error.rs:322:8]: IOError(Error { repr: Custom(Custom { kind: InvalidData, error: StringError("stream did not contain valid UTF-8") }) })
```
This PR displays the config file that failed to parse. 

See also: https://github.com/habitat-sh/habitat/issues/1120

1__tmux


reset:
So much better 😎

Issue #1184, opened 2016-08-23
Overriding nested (table) config vars clobbers defaults


chetan:
Configuration is only shallow-merged[1] currently, which creates issues when overriding vars inside a table. For example, given this default config: 

```toml
[server]
port = "8080"
shutdown-port = "8005"
redirect-port = "8443"
```
(from https://github.com/habitat-sh/core-plans/blob/master/tomcat8/default.toml)

If you override the port using `HAB_TOMCAT="server = { port=9090 }"` then the other `[server]` configs like `shutdown-port` get clobbered in the final  config (entirely missing), resulting in empty strings when using those vars. 

[1] https://github.com/habitat-sh/habitat/blob/master/components/sup/src/service_config.rs#L313

metadave:
potential fix here: https://github.com/habitat-sh/habitat/pull/1185

juliandunn:
This got fixed (for realzies this time) in #1307 and #1308, released with Habitat 0.10.2.

PR #1185, opened 2016-08-23
service_config::toml_merge handles toml::Tables

metadave:
fixes https://github.com/habitat-sh/habitat/issues/1184

### DIFF
--- a/components/sup/src/service_config.rs
+++ b/components/sup/src/service_config.rs
@@ -314,6 +314,28 @@ struct Cfg {
 fn toml_merge(left: &toml::Table, right: &toml::Table) -> toml::Table {
     let mut final_map = toml::Table::new();
     for (left_key, left_value) in left.iter() {
+        // attempt to merge nested tables
+        let a = right.get(left_key).and_then(|v| v.as_table());
+        let b = left.get(left_key).and_then(|v| v.as_table());
+        match (a, b) {
+            (Some(left_t), Some(right_t)) => {
+                let new_table = toml_merge(left_t, right_t);
+                let table_as_value = toml::Value::Table(new_table);
+                final_map.insert(left_key.clone(), table_as_value);
+                continue;
+            }
+            (Some(_), _) => {
+                // this occurs when the value type in left != type in right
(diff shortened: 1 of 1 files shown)


thesentinels:
@metadave, thanks for your PR! By analyzing the annotation information on this pull request, we identified @adamhjk, @fnichol, @reset and @smith to be potential reviewers

adamhjk:
Did this merge correctly? I see a "Closed with unmerged commits" here.

robbkidd:
T'isn't merged.

robbkidd:
I've tinkered with this branch in my dev environment. I've confirmed the tests added in this change fail-as-expected when the old form of the `toml_merge` function is used. And the test failures match what I'm seeing when running a package with nested values and I provide a custom config.

I don't quite know how to wire up running my build of `hab-sup` as the supervisor for that package though because they're in separate repos/working directories/containers.

metadave:
@robbkidd inside the dev-shell:

        /src/target/debug/hab-sup start core/somepackage

robbkidd:
OK. Managed to run a service with `hab-sup` from this branch. I believe the merge is currently overriding in the wrong direction: values from `default.toml` are kept, not the overrides from the environment.

robbkidd:
```

robbkidd:
Now that I'm looking at this again, the two uses of `toml_merge` in the code there have the direction opposite each other.

robbkidd:
Building off of @metadave's work here, I rebased this branch and flipped the order of the params in calls to `toml_merge`. Running my thing with `hab-sup` built from [the new branch](https://github.com/habitat-sh/habitat/compare/master...robbkidd:dp_toml_merge_table_fix) runs with environment values taking precedence over defaults. There's some thoughts about refactoring the implementation of the config merge, but it appears that at least with these two commits the "default values in nested tables disappear when some are overridden" and "config precedence flows in the expected order" issues may be fixed.

robbkidd:
Rebased branch + precedence fix offered for review in #1294

metadave:
thanks @robbkidd, you rock!

Issue #1175, opened 2016-08-18
setuid bit is lost upon install


miketheman:
- [x] The OS (including version) where you are running any of the Habitat commands.
```
# uname -a
Linux f9f3298c3c78 4.4.17-moby #1 SMP Fri Aug 12 22:26:18 UTC 2016 x86_64 GNU/Linux
# hab --version
hab 0.9.0/20160815225003
```

When creating a package that sets a file with a `setuid` on it, the locally-installed version works correctly.

However, when installing from a package, the setuid bit is lost.

```shellsession
[1][default:/src:0]# cd sudo
[2][default:/src/sudo:0]# build
...
[3][default:/src/sudo:0]# ls -la $(hab pkg path miketheman/sudo)/bin/sudo
-rwsr-xr-x 1 root root 158816 Aug 18 15:05 /hab/pkgs/miketheman/sudo/1.8.17p1/20160818150414/bin/sudo
[4][default:/src/sudo:0]# rm -fr /hab/pkgs/miketheman/sudo
[5][default:/src/sudo:0]# hab pkg install results/miketheman-sudo-1.8.17p1-20160818150414-x86_64-linux.hart
↓ Downloading miketheman-20160616143833 public origin key
    81 B / 81 B | [===================================================================================================================] 100.00 % 1.90 MB/s
☑ Cached miketheman-20160616143833 public origin key
→ Using core/acl/2.2.52/20160612075215
→ Using core/attr/2.4.47/20160612075207
→ Using core/coreutils/8.25/20160729192409
→ Using core/gcc-libs/5.2.0/20160612075020
→ Using core/glibc/2.22/20160612063629
→ Using core/gmp/6.1.0/20160612064724
→ Using core/libcap/2.24/20160612075226
→ Using core/linux-headers/4.3/20160612063537
→ Using core/ncurses/6.0/20160612075116
→ Using core/vim/7.4.1089/20160729215447
✓ Installed miketheman/sudo/1.8.17p1/20160818150414
★ Install of miketheman/sudo/1.8.17p1/20160818150414 complete with 1 new packages installed.
[6][default:/src/sudo:0]# ls -la $(hab pkg path miketheman/sudo)/bin/sudo
-rwxr-xr-x 1 root root 158816 Aug 18 15:05 /hab/pkgs/miketheman/sudo/1.8.17p1/20160818150414/bin/sudo
```

To be perfectly clear:
`-rwsr-xr-x` turns into `-rwxr-xr-x`

[...]

PR #1187, opened 2016-08-24
[core] Use 'Owner' and 'Permissions' during unpack

fujin:
* Per the [libarchive-rust documentation](https://github.com/chef/libarchive-rust/blob/master/src/archive.rs#L201-L212), the `Permissions` option may be
  used to fully restore the original permissions. When used in
  combination with the `Owner` option, the original owner as well as the
  setuid/setgid bits should be properly set during installation.

I was unfortunately having some trouble getting my dev `make shell` to build a
test-setuid plan.sh with this, so would welcome an additional test.

References #1175, perhaps!

Signed-off-by: AJ Christensen <aj@junglistheavy.industries>

### DIFF
--- a/components/core/src/package/archive.rs
+++ b/components/core/src/package/archive.rs
@@ -188,6 +188,8 @@ impl PackageArchive {
         let writer = writer::Disk::new();
         let mut extract_options = ExtractOptions::new();
         extract_options.add(ExtractOption::Time);
+        extract_options.add(ExtractOption::Owner);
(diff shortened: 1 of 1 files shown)


thesentinels:
@fujin, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol, @metadave, @reset and @stevendanna to be potential reviewers

reset:
gif-keyboard-13661974733786825745

PR #1189, opened 2016-08-25
Adds access token scope to CLI

freethejazz:
This simply adds a clarifying detail, that the access token only needs the `user:email` scope for setting up depot with GitHub authentication.

Signed-off-by: Jonathan Freeman <freethejazz@gmail.com>

### DIFF
--- a/components/hab/src/command/cli.rs
+++ b/components/hab/src/command/cli.rs
@@ -81,7 +81,7 @@ pub mod setup {
         para("While you can build and run Habitat packages without sharing them on the public \
               depot, doing so allows you to collaborate with the Habitat community. In addition, \
               it is how you can perform continuous deployment with Habitat.");
-        para("The depot uses GitHub authentication with an access token \
+        para("The depot uses GitHub authentication with an access token with the user:email scope \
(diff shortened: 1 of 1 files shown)


thesentinels:
@freethejazz, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol, @juliandunn, @metadave and @davidwrede to be potential reviewers

reset:
@freethejazz Thanks for adding some clarity to the CLI for users, this should be good enough for now!

In the future I think we need something in our documentation instead and a link to it in our CLI. That way we can explain a bit more about *why* we need certain scopes and the list can expand if we add additional required or optional scopes

freethejazz:
There already is a bit of documentation in the getting started guide, but that might need to be augmented/moved to more a more general doc page: https://www.habitat.sh/tutorials/getting-started-setup-environment/ (see the _Note:_ at the bottom of the page)

reset:
@freethejazz I think you're right, we should call it out in a more general sense. @davidwrede what do you think?

davidwrede:
Agreed. I'll open an issue and assign to myself.

PR #1190, opened 2016-08-25
Set the pkg user/group for builder-api-proxy

jtimberman:
With the change in #1050, we need to start nginx in builder-api-proxy as root, and then drop privileges to hab/hab.

We also need to put the cache in the right place.


thesentinels:
@jtimberman, thanks for your PR! By analyzing the annotation information on this pull request, we identified @adamhjk and @reset to be potential reviewers

reset:
gif-keyboard-2270681443418391568

Issue #1192, opened 2016-08-26
Port Hab command-line tool to Windows


smurawski:
Goal - a "native" hab.exe that supports interacting with the depot and supervisors, as well as entering a Docker-based studio for building 64 bit Linux applications.

Native Dependencies:

- [X]  OpenSSL [Build script](https://github.com/smurawski/hab-build-script/blob/master/openssl_static.ps1)
- [X]  Libarchive  [Build script](https://github.com/smurawski/hab-build-script/blob/master/openssl_static.ps1)
- [X]  Libsodium [Build script](https://github.com/smurawski/hab-build-script/blob/master/openssl_static.ps1)
- [X]  bzip2 [Build script](https://github.com/smurawski/hab-build-script/blob/master/openssl_static.ps1)
- [X]  xz [Build script](https://github.com/smurawski/hab-build-script/blob/master/openssl_static.ps1)
- [X]  zlib [Build script](https://github.com/smurawski/hab-build-script/blob/master/openssl_static.ps1)

Habitat Project Crates Required:

- [X] common
- [X] core (platform specific-y things are stubbed out currently)
- [X] http_client 
- [X] builder-depot-client
- [X] builder-protocol

smurawski:
We will have a Windows `hab.exe` binary with our next release (probably 0.11.0).

PR #1193, opened 2016-08-26
Extract linux/mac specific behavior in the habitat core crate

smurawski:
(from the `users` crate and `libc`) to modules.

Stubbed out windows versions of those functions.

re: #1192


thesentinels:
@smurawski, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol, @metadave, @reset and @adamhjk to be potential reviewers

reset:
@smurawski great first Rust PR! The only change I would suggest is to name the `platforms` module `os` instead. This would mirror the naming convention in the standard library (`std::os`) which defines "OS-Sepcific functionality"

PR #1204, opened 2016-09-01
[ci/rustfmt] Uninstall then install rustfmt when upgrading.

fnichol:
This change is necessary if using an older version of Cargo which does
not yet support the `--force` flag on `cargo install`.


thesentinels:
@fnichol, thanks for your PR! By analyzing the annotation information on this pull request, we identified @adamhjk to be a potential reviewer

PR #1207, opened 2016-09-01
[dockerize] take advantage of docker layers

chetan:
This modifies the docker exporter to take advantage of layering. The current implementation creates an image consisting of a single layer with all the dependencies and packages installed. Instead, we create two images:
- A base image which includes only runtime dependencies
- Named as `origin/pkg_base:<hash>` where hash is computed from the names (idents) of the runtime deps
- Also tagged as `origin/habitat_export_base:<hash>` to allow re-use across packages which share the same set of dependencies. For instance, we have many apps which will only depend on `core/tomcat8`.
- A "runtime" image which installs the final package(s) and uses previous image as the base (`FROM origin/pkg_base:<hash>`)

This gives much better reuse of layers and totally avoids re-building the base image between app builds. Final images:

```
REPOSITORY                                  TAG                    IMAGE ID            CREATED             SIZE
chetan/foobar                               0.1.0-20160826204445   c3b2ed108437        5 seconds ago        177.8 MB
chetan/foobar                               latest                 c3b2ed108437        5 seconds ago        177.8 MB
chetan/habitat_export_base                  dadf80f42a2a2028       7921e8c8aac2        16 minutes ago       159.8 MB
chetan/foobar_base                          dadf80f42a2a2028       7921e8c8aac2        16 minutes ago       159.8 MB
```

### DIFF
--- a/components/pkg-dockerize/bin/hab-pkg-dockerize.sh
+++ b/components/pkg-dockerize/bin/hab-pkg-dockerize.sh
@@ -83,6 +83,15 @@ find_system_commands() {
   fi
 }
 
+
+add_trailing_slash() {
+  STR="$1"
+  length=${#STR}
+  last_char=${STR:length-1:1}
+  [[ $last_char != "/" ]] && STR="$STR/"; :
+  echo $STR
+}
+
 # Wraps `dockerfile` to ensure that a Docker image build is being executed in a
 # clean directory with native filesystem permissions which is outside the
 # source code tree.
@@ -92,14 +101,18 @@ build_docker_image() {
     hab pkg install $1 # try to install it
   fi
 
+  if [[ -n "$DOCKER_REGISTRY_URL" ]]; then
+    DOCKER_REGISTRY_URL=$(add_trailing_slash "$DOCKER_REGISTRY_URL")
+  fi
+
   pkg_name=$(package_name_for $1)
   pkg_origin=$(package_origin_for $ident_file)
   pkg_ident=$(package_ident_for $ident_file)
   pkg_version=$(version_num_for $ident_file)
 
   BASE_PKGS=$(base_pkgs $@)
-  DOCKER_BASE_TAG="${pkg_ident}_base:$(base_pkg_hash $BASE_PKGS)"
-  DOCKER_BASE_TAG_ALT="${pkg_origin}/habitat_export_base:$(base_pkg_hash $BASE_PKGS)"
+  DOCKER_BASE_TAG="${DOCKER_REGISTRY_URL}${pkg_ident}_base:$(base_pkg_hash $BASE_PKGS)"
+  DOCKER_BASE_TAG_ALT="${DOCKER_REGISTRY_URL}${pkg_origin}/habitat_export_base:$(base_pkg_hash $BASE_PKGS)"
 
   # create base layer image
   DOCKER_CONTEXT="$($_mktemp_cmd -t -d "${program}-XXXX")"
@@ -107,6 +120,9 @@ build_docker_image() {
   docker_base_image $@
   popd > /dev/null
   rm -rf "$DOCKER_CONTEXT"
+
+
+  DOCKER_RUN_TAG="${DOCKER_REGISTRY_URL}${pkg_ident}"
 
   # build runtime image
   DOCKER_CONTEXT="$($_mktemp_cmd -t -d "${program}-XXXX")"
@@ -191,9 +207,6 @@ EOT
 }
 
 docker_image() {
-  local version_tag="${pkg_ident}:${pkg_version}"
-  local latest_tag="${pkg_ident}:latest"
-
   local pkg_file=$(ls /hab/cache/artifacts/$(cat $(hab pkg path $pkg_ident)/IDENT | tr '/' '-')-*)
   cp -a $pkg_file $DOCKER_CONTEXT/
   pkg_file=$(basename $pkg_file)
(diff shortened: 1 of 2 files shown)


thesentinels:
@chetan, thanks for your PR! By analyzing the annotation information on this pull request, we identified @adamhjk, @fnichol, @juliandunn, @mattray and @julian7 to be potential reviewers

reset:
@chetan after reading up on Docker layers I think this is a great addition!

@fnichol do you want to take a peek at this PR before we merge it in since you did most/all of the work in the dockerize component?

chetan:
Just pushed another commit which adds support for private docker registries. If you export DOCKER_REGISTRY_URL="docker.example.com:443" then the images will be tagged accordingly and pushed at the end.

fnichol:
Hey @chetan, I still have this on my to-review pile sorry it's taking longer than I'd like.

fnichol:
Hi @chetan, sorry to keep this as an open loop. @juliandunn and I have started talking about this PR on the product side and the goals about what we're after (faster downloads,  smaller images, using Docker-specific features where appropriate).

We should be able to follow up by next week one way or the other on this PR. Thank you so much for your patience here!

eeyun:
Ok. So we've totally dropped the ball on this one, my apologies @chetan. I've thought about it enough and I think that a large subset of folks who are using habitat to package up docker containers would be pretty happy with this change. So I'm :+1: on it.

The only change I'd like to see is adding a feature-flag for uploading to the docker hub _as well as_ the environment variable. As far as precedence is concerned the feature flag should supersede the environment variable imo.

@reset @fnichol and @smurawski can we get some eyes on this?

chetan:
Sure, I can add the feature-flag. Something like `--push`?

chetan:
Ok, think I finally got this!



```
USAGE:
  hab-pkg-dockerize [--repo <repo>] [--push] <PKG_IDENT>

FLAGS:
    --help           Prints help information

OPTIONS:
    --repo=URL       If given, prefix the tag with the URL
    --push           Push the built images to the configured repository

ARGS:
    <PKG_IDENT>      Habitat package identifier (ex: acme/redis)

EXAMPLE:
    hab-pkg-dockerize --repo docker.private.com:443 --push core/nginx

    Would create and push the following images/tags:

    docker.private.com:443/core/nginx                     1.10.1-20161207041036   380c892f61ab        2 minutes ago       183.7 MB
    docker.private.com:443/core/nginx                     latest                  380c892f61ab        2 minutes ago       183.7 MB
    docker.private.com:443/core/habitat_export_base       43691665acf57304        e965f9e803c4        13 hours ago        181.9 MB
    docker.private.com:443/core/nginx_base                43691665acf57304        e965f9e803c4        13 hours ago        181.9 MB
```

adamhjk:
@fnichol your eyes would be good here - this looks good to me.

chetan:
Bump

eeyun:
@chetan we should totally bother you with another video call about this. I think the only remaining hangup we had with this PR was a concern around whether or not we'd have trouble getting the base layer updated . @fnichol might remember the discussion there. Otherwise I can take some time to re-review the PR and and we can finally get it sorted out.

chetan:
Not sure I see where you're going with that, but I'm open to feedback. We're going to be using this more heavily starting real soon so should be able to shake out issues on our end as well.

eeyun:
Yeah I totally need to dig back into the PR because it's been too long. I'll see if perhaps @fnichol and I can pair on this today.

chetan:
Going to close this PR for now. My branch is changing a bunch so perhaps I'll reopen once it settles a bit.

mumoshu:
Hi @chetan @eeyun @fnichol @reset, this is SO NICE!

I had been doing [complex bash-scripting](https://github.com/mumoshu/kubeherd/blob/master/builds/alpine-rails/imagebuild) to orchestrate [s2i](https://github.com/openshift/source-to-image) and docker builds so that I can leverage all the docker layer caching goodies to build minimal docker images in a fast and reproducible manner.

Habitat with this PR seems to achieve the same goal in a far more elegant way.
And leveraging docker-layer caching for faster builds and faster deploys are common-patterns(imho) seen in the docker world.
Baking this into Habitat would attract even more heavy docker users to use it.

Anything I can do to revive this?

christophermaier:
@mumoshu The Docker exporter has since been rewritten in Rust, so any revival of this feature would also need to be rewritten. At the time, it seemed that the feeling on the basic layering use case was positive; if this is something you are interested in, could you open a new issue detailing the aspects of a layer-aware Docker exporter that you'd most like to see?

If speed and reproducibility are your main concerns, though, the current exporter is rather fast and reproducible by definition, since it's using Habitat packages. It doesn't really take advantage of layering, though, since all the dependencies (indeed, most all the files involved) are added in a single layer.

mumoshu:
@christophermaier Thanks for the clarification!
I'll submit a dedicated issue once I get a more concrete picture of what I want to achieve from it.

> If speed and reproducibility are your main concerns, though, the current exporter is rather fast and reproducible by definition, since it's using Habitat packages

Awesome! I was certainly not aware of that.

Then,

> It doesn't really take advantage of layering, though, since all the dependencies (indeed, most all the files involved) are added in a single layer

This seems like what I'd like to solve today.
Basically, I'd like only the changed image layers(=habitat packages, hopefully) are transmitted among CI jobs, docker registries and nodes running the resulting docker images for faster deploys of habitat-built docker images.

PR #1210, opened 2016-09-02
depot/builder-api refactors

reset:
This is a first of many CUTS (clean up the shit) pull requests. This leverages the recent work that landed in master to centralize on a single authentication mechanism across all of our http gateways and uses the synchronous route function added to the routing module to simplify the code in the depot. I also found a bug in origin key upload which this will address.


components/builder-depot/src/server.rs
```diff
@@ -196,26 +74,9 @@ pub fn origin_create(depot: &Depot, req: &mut Request) -> IronResult<Response> {
...
-                _ => unreachable!("unexpected msg: {:?}", rep),
-            }
-        }
-        Err(e) => {
-            error!("{:?}", e);
-            Ok(Response::with(status::ServiceUnavailable))
-        }
+    match conn.route::<OriginCreate, Origin>(&request) {
+        Ok(origin) => Ok(render_json(status::Created, &origin)),
+        Err(err) => Ok(render_net_error(&err)),
```
fnichol: Much more compact! The sync version also reads well.

components/builder-depot/src/server.rs
```diff
@@ -1251,187 +955,90 @@ fn dont_cache_response(response: &mut Response) {
...
+        get "/pkgs/:origin/:pkg/latest" => show_package,
+        get "/pkgs/:origin/:pkg/:version" => list_packages,
+        get "/pkgs/:origin/:pkg/:version/latest" => show_package,
+        get "/pkgs/:origin/:pkg/:version/:release" => show_package,
+
+        get "/pkgs/:origin/:pkg/:version/:release/download" => download_package,
         post "/pkgs/:origin/:pkg/:version/:release" => {
-            move |r: &mut Request| upload_package(&depot17, r)
+            if depot.config.insecure {
+                XHandler::new(upload_package)
```
fnichol: Nice! I believe this would avoid the need for *any* token for insecure mode?

reset: Yep! That was the previous behavior I believe

fnichol: Just tried it out (uploading to a local `hab-depot start --insecure`), works like cake!

fnichol:
👍 

[insert gif here due to 3g-over-phone-in-car]

PR #1219, opened 2016-09-08
add/improve make targets for running & building builder services

reset:
* Replace `make api-shell` target with `make bldr-run-shell`
* Add cmake to Dockefile package list - required for builder-worker
* Add Procfile for Linux which additionally starts Redis
* Add builder-worker to be started by Linux & Mac Procfile


support/Procfile.linux
```diff
@@ -0,0 +1,8 @@
+redis: redis-server
```
raskchanky: Not a blocker, but I'm curious why the linux Procfile starts redis and the mac one doesn't.

raskchanky:
One small question, but this looks great.

gif-keyboard-11381497733256267935

support/Procfile.linux
reset: @raskchanky just because if you've installed Redis by Homebrew then launchd will be automatically starting it for you. In the long run I hope that we can use pure Habitat packages in both Linux and Mac to link against and run against, but we need some more work still ;)

PR #1221, opened 2016-09-09
Quick fix for sorting package identifiers returned by the Depot API

reset:
This is a first pass implementation to get a fix out the door. We're
pulling a bulk of strings out of the database and allocating identifiers
for each then performing a sort in memory. This is going to be alright
for now, but in the future we should use weights in Redis' sorted sets
and adjust the rank of all identifiers on insertion of a new package
so we only perform the sorting overhead on write instead of each read

### DIFF
--- a/components/builder-depot/src/data_store.rs
+++ b/components/builder-depot/src/data_store.rs
@@ -141,12 +141,13 @@ impl PackagesIndex {
         let conn = self.pool().get().unwrap();
         match conn.zrange::<String, Vec<String>>(Self::key(&id.to_string()), offset, count) {
             Ok(ids) => {
-                let ids = ids.iter()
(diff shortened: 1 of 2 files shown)


thesentinels:
@reset, thanks for your PR! By analyzing the annotation information on this pull request, we identified @metadave and @fnichol to be potential reviewers

raskchanky:
I think I understand what's going on here, and from what I can tell, it looks good.

gif-keyboard-11019382013941327395

adamhjk:
This looks good. Yay for traits making the `sort` logic just work on the iterator.

Merge it!

adamhjk:
gif-keyboard-3394374292705165834

PR #1225, opened 2016-09-13
Move buffer trimming to the final buffer

adamhjk:
We used to have a bug, and that bug was that we weren't reading an
entire file from the tarball; isntead, we were only getting the very
first chunk.

We fixed that bug. Yay us!

However, we wanted all the data we read from our metadata files to have
no trailing whitespace or newlines (because hey - you really just want
the data you want, not the extra garbage on the end that humans like.
Silly humans.) We did this trimming on the content of the block we read.

When you then generate a file that has the buffer boundary pefectly
aligned with a newline, you will get your newline chomped off, and an
error from habitat that your metadata is malformed.

This commit fixes that issue by moving the trim to the final buffer,
rather than on every call.

Special shoutout to @fnichol, who helped debug this.

gif-keyboard-5320514194343445738

### DIFF
--- a/components/core/src/package/archive.rs
+++ b/components/core/src/package/archive.rs
@@ -199,12 +199,13 @@ impl PackageArchive {
 
     fn read_deps(&mut self, file: MetaFile) -> Result<Vec<PackageIdent>> {
         let mut deps: Vec<PackageIdent> = vec![];
+        let filename = format!("{}", &file);
         match self.read_metadata(file) {
             Ok(Some(body)) => {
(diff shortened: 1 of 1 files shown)


thesentinels:
@adamhjk, thanks for your PR! By analyzing the annotation information on this pull request, we identified @metadave, @fnichol, @reset, @stevendanna and @fujin to be potential reviewers

reset:
gif-keyboard-1563116012786367253

fnichol:
gif-keyboard-1730024699812803851

PR #1228, opened 2016-09-14
[hab,sup] Add common UI subsystem to manage output display.

fnichol:
This change is a first move to add a consistent "design language" to the
user-facing tools--primarily the `hab` program. It introduces a `UI`
struct in the common crate which knows how to write headings, status
messages, etc in the context of consoles that could be interactive or
noninteractive and terminals that support color or ones that do not
support color.

Most of the work to date is focused around proper terminal capability
detection (via the `term` crate) and interactive status (via the
`libc::isatty` and `kernel32::GetConsoleMode` calls). More testing and
validation against Windows platforms is required, but the minimum work
is there.

The `UI` struct has several methods on it that help authors of `hab`
subcommands present a consistent response to an end user. The `.begin()`
and `.end()` methods properly decorate and color the respective
messages, and the `.status()` method helps steer the subcommand author
into a verb-first message with consistent symbol decoration.

A pseudo-code example might be:

```
try!(ui.begin(format!("Installing {}", &pkg_ident)));
// install package and determine fully qualified identifier
// ...
let fqpi = ...;
try!(ui.status(Status::Downloading, &fqpi));
// ...
try!(ui.begin(format!("Install of {} successful", &fqpi)));
```

which might yield output similar to:

```
» Installing core/jq-static
↓ Downloading core/jq-static/1.5/20160730001211
...
★ Install of core/jq-static/1.5/20160730001211 sucessful
```

The `UI` struct is responsible for determining coloring capability,
context, and strategy, which symbol decorations to use by default, etc.

Still do do in future PRs:

* Move progress bar logic into `UI` to better determine if progress bars should be displayed or not
* Add warning, error, and fatal methods to `UI` and remove several one-off colored `println!()` calls
* Move input/query logic into `UI` (from `hab setup` logic)


thesentinels:
@fnichol, thanks for your PR! By analyzing the annotation information on this pull request, we identified @metadave, @reset, @seth, @nsdavidson, @juliandunn and @adamhjk to be potential reviewers

components/common/src/ui.rs
```diff
@@ -0,0 +1,368 @@
...
+    Uploading,
+    Using,
+    Verified,
+    Custom(char, String),
+}
+
+impl Status {
+    pub fn parts(&self) -> (char, String, Colour) {
+        match *self {
+            Status::Applying => ('↑', "Applying".into(), Colour::Green),
```
metadave: 👏  I love this

metadave:
👍 

gif-keyboard-14534286600092507396

components/common/src/command/package/install.rs
```diff
@@ -39,7 +39,6 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
-use ansi_term::Colour::{Blue, Green, Yellow};
```
reset: This is the only place that we were leveraging the `ansi_term` crate so we can now drop it as a dependency from this crate's Cargo.toml

fnichol: We'll still need it in the common crate's deps as that's our current coloring implementation in the [UI](https://github.com/habitat-sh/habitat/blob/db9caa8c5f704d38a227e2130dae1afca6f52ace/components/common/src/ui.rs#L18).

I think there's still a couple of spots in the hab crate that still use this crate to paint some errors, but I have those next in my sights. Hopefully when this is all over we only have 1 reference to `ansi_term` (in the common crate) which will be an internal implementation detail only (i.e. free to swap, change, update in the future). Closer!

fnichol:
cc @smurawski as a heads up that we're making our platform differences slightly wider with this PR--I think we looked at this last week, but it may not yet be 100% on the Windows side

reset:
gif-keyboard-4927756588733505730

Excellent first pass on wrangling our UI patterns into an abstraction. In a future pass I think we should simplify the UI by:

* Making the UI a globally available singleton by placing it on the heap as a mutable static ref and exposing macros/functions to send messages to the UI. This might be accomplished by directly addressing the static ref or by sending messages through a channel
* Altering the begin/end portion of messaging the UI to leverage a closure. This will ensure the end message is always associated with the begin regardless if we had a success or failure or if the programmer forgot to provide one

components/common/src/command/package/install.rs
reset: Ah, my mistake!

PR #1232, opened 2016-09-14
Clean the middleman build before building

reset:
This addresses the weirdness we noticed late last week after deploying the site. A few pages were not generated and the s3 sync deleted them resulting in a few unwanted 404s. There's likely a bug in Middleman's logic for determining what is new/changed which we can side step by just deleting the output directory before building

### DIFF
--- a/www/Makefile
+++ b/www/Makefile
@@ -1,6 +1,9 @@
-build:
+build: clean
 	bundle exec middleman build
 .PHONY: build
+
+clean:
+	rm -Rdf build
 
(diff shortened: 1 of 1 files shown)


thesentinels:
@reset, thanks for your PR! By analyzing the annotation information on this pull request, we identified @adamhjk to be a potential reviewer

raskchanky:
gif-keyboard-17861947114191544080

fnichol:
Agreed, better to start from clean that a potentially dirty build.

PR #1234, opened 2016-09-14
On windows, order matters - case does not

smurawski:
We may want to document the behavior, but it is core to Windows that case doesn't matter for environment variables and rust seems to play nice that way.

I had to switch the order of some of the environment variable sets since on Windows HTTP_PROXY, HTTP_proxy, http_PROXY, and http_proxy (and any other capitalization variant) are the same variable, so those test assignments just overwrite each other.

Another one off the list from #1192 

### DIFF
--- a/components/http-client/src/proxy.rs
+++ b/components/http-client/src/proxy.rs
@@ -221,8 +221,8 @@ impl ProxyBasicAuthorization {
 /// use std;
 /// use habitat_http_client::proxy;
 ///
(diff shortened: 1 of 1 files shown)


thesentinels:
@smurawski, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol to be a potential reviewer

reset:
gif-keyboard-18270585192302949495

Issue #1116, opened 2016-08-02
Supervisor attempts to perform config substitutions on directories in the config directory


tevert:
I found this by accidentally 'cp -R <mysrc>/config <myPkg>/config', which creates '<mypkg>/config/config'. 
The supervisor attempts to perform handlebars substitution on _everything_ in the config directory, which naturally blows up on the nested directory. 

Suggested possible fixes: 
- Nicer error message than a generic IO error
- Skip directories
- Recurse inside directories

bdangit:
oh probably explains why can't have nice nested config directories inside the config folder.

fnichol:
Looks like #1236 is a candidate fix here, we'll follow up in the PR to close this loop, thanks!

christophermaier:
Closing for fix in #1236

PR #1236, opened 2016-09-15
Recurse into subdirectories inside config.

nsdavidson:
Potential fix for #1116.

Currently the supervisor blows up when there is a directory inside a packages config directory.  This change adds the ability for the supervisor to recreate any subdirectories and render the config files in them.

Signed-off-by: Nolan Davidson ndavidson@chef.io

### DIFF
--- a/components/sup/src/package/mod.rs
+++ b/components/sup/src/package/mod.rs
@@ -37,6 +37,7 @@ use service_config::ServiceConfig;
 use supervisor::Supervisor;
 use util::path::busybox_paths;
 use util::users as hab_users;
+use walkdir::WalkDir;
 
 static LOGKEY: &'static str = "PK";
 const INIT_FILENAME: &'static str = "init";
@@ -243,23 +244,27 @@ impl Package {
     /// helpers above.
     pub fn config_files(&self) -> Result<Vec<String>> {
         let mut files: Vec<String> = Vec::new();
-        let config_dir = self.config_from().join("config");
-        debug!("Loading configuration from {:?}", config_dir);
-        match std::fs::read_dir(config_dir) {
-            Ok(config_dir) => {
-                for config in config_dir {
-                    let config = try!(config);
-                    match config.path().file_name() {
-                        Some(filename) => {
-                            debug!("Looking in {:?}", filename);
-                            files.push(filename.to_string_lossy().into_owned().to_string());
-                        }
-                        None => unreachable!(),
-                    }
-                }
+        let cfg_dir = self.path().join("config");
+
+        match std::fs::metadata(&cfg_dir).map(|d| d.is_dir()) {
+            Err(_) | Ok(false) => {
+                return Ok(files)
             }
-            Err(e) => {
-                debug!("No config directory in package: {}", e);
+            Ok(true) => ()
+        };
+
+        for entry in WalkDir::new(&cfg_dir) {
+            if let Err(e) = entry {
+                debug!("Can't process config file: {}", e);
+                continue;
+            }
+            let entry = entry.unwrap();
+            if entry.path().is_file() {
+                if let Ok(rel_path) = entry.path().strip_prefix(&cfg_dir) {
+                    let f = rel_path.to_string_lossy().into_owned();
+                    debug!("Detected config file: {}", &f);
+                    files.push(f);
+                }
             }
         }
         Ok(files)
--- a/components/sup/src/service_config.rs
+++ b/components/sup/src/service_config.rs
@@ -17,7 +17,8 @@
 use std::ascii::AsciiExt;
 use std::collections::HashMap;
 use std::env;
-use std::fs::File;
+use std::fs::{File, create_dir_all};
+use std::path::Path;
 use std::io::prelude::*;
 
 use ansi_term::Colour::Purple;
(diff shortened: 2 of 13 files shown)


thesentinels:
@nsdavidson, thanks for your PR! By analyzing the annotation information on this pull request, we identified @adamhjk, @metadave, @fnichol, @reset and @smith to be potential reviewers

components/sup/src/package/mod.rs
```diff
@@ -233,21 +234,16 @@ impl Package {
...
-                        Some(filename) => {
-                            files.push(filename.to_string_lossy().into_owned().to_string());
-                        }
-                        None => unreachable!(),
-                    }
+        if std::fs::metadata(self.path().join("config")).unwrap().is_dir() {
+            for entry in WalkDir::new(self.path().join("config")) {
+                let entry = entry.unwrap();
+                if entry.path().is_file() {
+                    let rel_path = entry.path().strip_prefix(&self.path().join("config").to_str().unwrap()).unwrap().to_string_lossy().into_owned().to_string();
```
metadave: we try to handle as many `Result` values with pattern matching. In general, I'm suspicious whenever I see an unhandled `unwrap()` in the wild.

I took a shot at refactoring this method a bit:

```
 pub fn config_files(&self) -> Result<Vec<String>> {
        let mut files: Vec<String> = Vec::new();
        let cfg_dir = self.path().join("config");

        match std::fs::metadata(&cfg_dir).map(|d| d.is_dir()) {
            Err(_) | Ok(false) => {
                // error reading config dir or dir doesn't exist
                return Ok(files)
            }
            Ok(true) => ()
        };

        for entry in WalkDir::new(&cfg_dir) {
            if let Err(e) = entry {
                debug!("Can't process config file {}", e);
                continue;
            }
            let entry = entry.unwrap();
            if entry.path().is_file() {
                if let Ok(rel_path) = entry.path().strip_prefix(&cfg_dir) {
                    let f = rel_path.to_string_lossy().into_owned();
                    println!("Detected config file = {}", &f);
                    files.push(f);
                }
            }
        }
        Ok(files)
    }
```
In the case of `entry` above, to reduce the amount of blocks + indentation, I check to see if `entry` is an `Err` and move on, or unwrap if it's not. This is circumventing the compilers exhaustive pattern matching, but I've used this technique carefully in other spots to reduce too many nested `{ }`'s.

components/sup/src/service_config.rs
```diff
@@ -196,6 +197,11 @@ impl ServiceConfig {
                 debug!("Configuration {} does not exist; restarting", filename);
                 outputln!("Updated {}", Purple.bold().paint(config));
                 self.config_hash.insert(filename.clone(), file_hash);
+                let parent_path = Path::new(&filename).parent();
+                match parent_path {
```
metadave: for a match where you only care about one pattern (the `Some` in this case), you can use an `if let`:

```
if let Some(pp) = Path::new(&filename).parent {
  try!(create_dir_all(pp));
}
```

nsdavidson:
@metadave Haven't had a chance to look too deeply into testing yet, but I refactored it for better unwrap() handling.

Will dig into testing options next week, but I'm thinking it will require putting a directory structure down to test properly?

metadave:
there may be some existing Rust tests that you could use as a start, and we also have a [test framework](https://github.com/habitat-sh/habitat/tree/master/test) that I'm cleaning up right now to run in Travis etc.

mwrock:
Sorry it has been so long since this has been looked up. We would definitely consider pulling this in but at this point the supervisor code has churned quite a bit and we'd need a rebase and also conflict cleanup.

mwrock:
I'm going to close this since it has grown stale and by now the code has diverged so far. We really appreciate the PR and we are also debating whether we want to introduce this in the future. If anyone thinks there is a pressing need, we can absolutely discuss further.

reset:
@nsdavidson btw, I completely forgot about this PR but I recently refactored the way we render templates out to disk. I left [this comment](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/config.rs#L231-L233) which describes exactly where to put this feature in the new code base if this is still desired.

I didn't do it myself because I just wasn't super sure that we really wanted it. I couldn't personally think of a use case for putting templates into directories

goldie80:
This unfortunately blocks me from trying out Habitat currently. Our application config uses a directory per environment today.

I'd like to initially add a "habitat" environment to proof of concept using habitat but still coexist with our current configuration today.

I'm not a rust programmer but I wouldn't mind trying if I knew my contribution would be valued..

bdangit:
@goldie80, never fear the power of submitting a PR. :). You'll find this community a very helpful and accepting one. So your PR(s) are definitely valued. 

Now to the point of environments as directories, is it by any chance that the environmental configs you are providing are "static"?  Essentially if you packaged a Dev, CI, Prod and etc, there are essentially hard coded settings.  If that's the case then you could create a plan where your `hook/init` will symlink or install the set of configs given an environmental flag you define. 

Example

```
# default.toml
environment = prod


# hook/init
environment={{ cfg.environment }}
if [[ $environment == "prod" ]]; then
  # symlink pkgpathfor yourpkg/files/prod to svc/config
elif [[ $environment == "dev" ]]; then 
  ...
elif other env
  ...
fi
```

In this scenario your configs would not live in `configs` but just in an aribitary folder inside `$pkg_prefix`

PR #1240, opened 2016-09-15
Always render run script with chpst

jtimberman:
This commit makes two primary changes to the build script:

- The run script generated from pkg_svc_run in the plan will always
  render with chpst set to run with pkg_svc_user and pkg_svc_group
- The generated run script no longer prefixes the pkg_svc_run command
  with "$pkg_prefix"

These changes are necessary to ensure that the sane default for built
packages does the right thing for the most cases. We assume that the
binary used in the pkg_svc_run will be in a bin dir defined in the plan
so it ends up in the PATH.

Additionally, we need some other changes:

- Update plans to reflect this change in the various components' plans
- The proxy package can now use pkg_svc_run instead of having its own
  run hook

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1993,11 +1993,14 @@ do_default_build_config() {
   return 0
 }
 
-# Write out the `$pkg_prefix/run` file. If a file named `hooks/run` exists, we
-# skip this step. Otherwise, we look for `$pkg_svc_run`, and use that.
+# Write out the `$pkg_prefix/run` file. If a file named `hooks/run`
+# exists, we skip this step. Otherwise, we look for `$pkg_svc_run`,
+# and use that. We assume that the binary used in the `$pkg_svc_run`
+# command is set in the $PATH.
 #
-# If the `$pkg_svc_user` is set to a value that is not `root`, we change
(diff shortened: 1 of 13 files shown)


components/plan-build/bin/hab-plan-build.sh
```diff
@@ -2009,32 +2012,27 @@ do_build_service() {
...
-  if [[ -f "$PLAN_CONTEXT/hooks/run" ]]; then
+  if [[ -f "${PLAN_CONTEXT}/hooks/run" ]]; then
+    build_line "Using run hook ${PLAN_CONTEXT}/hooks/run"
     return 0
   else
     if [[ -n "${pkg_svc_run}" ]]; then
-      if [[ -n "$pkg_svc_user" && "${pkg_svc_user}" != "root" ]]; then
-        cat <<EOT >> $pkg_prefix/run
+      # We use chpst to ensure that the script works outside `hab-sup`
+      # for debugging purposes, or under a `hab-director`.
```
metadave: 🍰

metadave:
:+1:

components/plan-build/bin/hab-plan-build.sh
jtimberman: 🚀 🍰

PR #1241, opened 2016-09-15
Static directory permissions

jtimberman:
This commit adds the static directory to the default directory
structure of services in /hab/svc and it ensures the ownership is set
consistently to avoid permissions errors.

We also need the static directory copied in the api application
instead of linked because the parent (/hab/svc/hab-builder-api) will
be owned by root, which creates a permissions issue for the init hook
to continue running and set up the habitat.conf.js config file.

This pull request requires the changes in #1240

### DIFF
--- a/components/builder-api/habitat/hooks/init
+++ b/components/builder-api/habitat/hooks/init
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-ln -sf {{pkg.path}}/static {{pkg.svc_path}}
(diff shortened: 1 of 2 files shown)


thesentinels:
@jtimberman, thanks for your PR! By analyzing the annotation information on this pull request, we identified @reset, @adamhjk, @fnichol and @metadave to be potential reviewers

metadave:
lgtm
gif-keyboard-8388711334349770750

PR #1246, opened 2016-09-16
Convert supervisor config to global singleton

adamhjk:
This commit converts to using a global singleton for configuration,
rather than threading a reference to everything that needs configuration
data.

Since we need to populate our `Config` struct with data from `clap`, we
need to be able to modify it at runtime with an argument, rather than
simply use `lazy_static!`. This implementation stores the configuration
as a static pointer to a location on the heap via `Box`, and updates
the pointer via `mem::transmute` in a `gcache()` method. This method
is called early in `main` - right after we populate the configuration
struct, in fact.

Once `gcache()` has been called (one time only; subsequent calls will be
ignored), you can access the `Config` struct through the
`config::gconfig()` method.

This removes all the references to threading an `&Config` through the
code base, and replaces them with calls to `gconfig()`.

In particular, it cleans up the need for lifetimes in the topology and
worker code, which is a nice side effect.

gif-keyboard-11431932797035670168

### DIFF
--- a/components/sup/src/config.rs
+++ b/components/sup/src/config.rs
@@ -21,6 +21,8 @@
 //! See the [Config](struct.Config.html) struct for the specific options available.
 
 use std::str::FromStr;
+use std::sync::{Once, ONCE_INIT};
+use std::mem;
 
 use hcore::package::PackageIdent;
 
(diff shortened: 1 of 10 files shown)


thesentinels:
@adamhjk, thanks for your PR! By analyzing the annotation information on this pull request, we identified @fnichol, @metadave, @reset, @smith, @juliandunn and @adamhjk to be potential reviewers

components/sup/src/topology/mod.rs
```diff
@@ -95,13 +94,11 @@ pub enum State {
...
 
 /// The topology `Worker` is where everything our state machine needs between states lives.
-pub struct Worker<'a> {
+pub struct Worker {
     /// The package we are supervising
     pub package: Arc<RwLock<Package>>,
     /// Name of the package being supervised
     pub package_name: String,
-    /// A pointer to our current Config
-    pub config: &'a Config,
```
fnichol: Bingo. That's a rather nice simplification.

components/sup/src/config.rs
```diff
@@ -30,6 +32,32 @@ use topology::Topology;
...
+/// The Static Global Configuration.
+///
+/// This sets up a raw pointer, which we are going to transmute to a Box<Config>
+/// with the first call to gcache().
+static mut CONFIG: *const Config = 0 as *const Config;
+
+/// Store a configuration, for later use through `gconfig()`.
+///
+/// MUST BE CALLED BEFORE ANY CALLS TO `gconfig()`.
+pub fn gcache(config: Config) {
```
fnichol: I presume that if we ever had to update the `Config`, we'd call this again with a new copy (I noticed that `.gconfig()` doesn't return a `mut` borrow)?

adamhjk: Right now, the Config struct is immutable after it is created. It's passed around as a reference, without mutability. 

If we decide we want it to be mutable, we will need to refactor into having the static hold a `RwLock<Config>` instead. Given that right now, for the supervisor at least, these are only options populated from the CLI, we're fine to let it remain immutable.

adamhjk:
Fixed the test suite - this should pass now. @fnichol, you have to approve my changes for merge. :)

fnichol:
Haha, I hold all the keys!

PR #1255, opened 2016-09-18
Update Cargo dependencies

adamhjk:
This commit updates all the Cargo dependencies. The only major change is
an update to Iron's Router, which adds a `url_for()` function to routes,
which requires us to add a third option to the router calls (in order to
produce the named routes.)


fnichol:
Nice, I was going to do this tomorrow. Looking good

gif-keyboard-8810445208805844864

PR #1256, opened 2016-09-18
Iterate on plans with --config-from

adamhjk:
gif-keyboard-9626797244353975288

This commit enables the `--config-from` option to the supervisor. It
allows you to pass the directory where your habitat plan lives, and when
you run the supervisor, it will use the configuration it finds there
rather than the configuration in the package.

While this violates a bunch of useful properties in production, it's
very useful for plan creation. Here is an example:

```bash
$ build redis
[.. hab builds stuff ..]
$ hab start core/redis
[ .. redis starts ..]
```

At this point, you might realize a hook is wrong, or a configuration
value isn't set. Currently, testing this would either require you to
edit the installed path in `/hab/pkgs`, or recompile all of redis. With
this option, you can instead say:

```bash
$ hab start core/redis --config-from /src
```

And edit the configuration file in place, test to your hearts content,
and then issue one final build when you're sure eveything works.

In future work, it would be cool to include some file watching logic, to
automatically reload the service when your configuration changes. For
now, though, this works great, and it doesn't take any new dependencies.

### DIFF
--- a/components/sup/src/config.rs
+++ b/components/sup/src/config.rs
@@ -139,12 +139,24 @@ pub struct Config {
     version_number: u64,
     organization: Option<String>,
     ring: Option<String>,
+    config_from: Option<String>,
 }
 
 impl Config {
     /// Create a default `Config`
     pub fn new() -> Config {
         Config::default()
+    }
+
+    /// Set the config file from directory
+    pub fn set_config_from(&mut self, config_from: Option<String>) -> &mut Config {
+        self.config_from = config_from;
+        self
+    }
+
+    /// Return the config file from directory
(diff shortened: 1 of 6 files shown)


thesentinels:
@adamhjk, thanks for your PR! By analyzing the annotation information on this pull request, we identified @metadave, @fnichol, @reset, @smith, @jtimberman and @adamhjk to be potential reviewers

adamhjk:
Let's not merge this till I write some docs and a test.

adamhjk:
Ok, this now includes docs. Looks like the test suite is almost ready, but not quite - so I skipped adding a test.

PR #1257, opened 2016-09-18
[devshell] Use rustup in Dockerfile.

fnichol:
This change "aggressively upgrades" us to use the new Rustup tooling
rather than the currently stable rustup.sh installer. Upon the release
of the next stable Rust, we should be able to drop the requirement of a
newer nightly Cargo installation further reducing the line count.

gif-keyboard-4040447348298661739


thesentinels:
@fnichol, thanks for your PR! By analyzing the annotation information on this pull request, we identified @metadave, @chetan, @reset, @jtimberman and @smith to be potential reviewers

reset:
For anyone following along, the rustup project will replace the default installation instructions on rust-lang.org once it's finished. It also happens to be a successor to multirust which is a huge plus!

PR #1258, opened 2016-09-19
[hab/mac] Prompt for core origin key when performing a Mac hab release.

fnichol:
This change alters the ordering of the `mac-build.sh` script to prompt
the user for a core origin signing key as soon as possible in the
process. As a consequence, a `wget` command is no longer present for the
mac `install.sh` program and therefore `curl` is used instead (Homebrew
installs the `wget` command whereas `curl` is available on stock Mac
installations).

gif-keyboard-4148750678897312232

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -191,13 +191,16 @@ target component.
 
 # How-To: Build Mac Components
 
-1. Turn on and enter virtual machine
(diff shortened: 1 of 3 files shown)


thesentinels:
@fnichol, thanks for your PR! By analyzing the annotation information on this pull request, we identified @reset and @juliandunn to be potential reviewers

PR #1259, opened 2016-09-19
[Makefile] Repair `serve-docs` target & support Docker for Mac.

fnichol:
This change updates the doc build paths to use the workspaces target
path and also fixes the default/fallback URL for the web server to use
`127.0.0.1` which is what the Docker for Mac app will use.


thesentinels:
@fnichol, thanks for your PR! By analyzing the annotation information on this pull request, we identified @reset, @adamhjk, @jtimberman and @metadave to be potential reviewers

PR #1260, opened 2016-09-20
Update Admin, API, and Depot route definitions with names

reset:
In https://github.com/habitat-sh/habitat/commit/fb2402e00c3f57822851ade4167210c0e7fe1bd4 we bumped our dependency on `iron-router` which requires a name field when declaring routes for use with `url_for()`.

We're not using the new named route functionality just yet (though, I'm happy it's arrived!) so these names may change int he future but this will get us compiling.


metadave:
+1

PR #1262, opened 2016-09-20
Cleaning up the windows build docs

smurawski:
After some feedback from habislack and moving more quickly across a number of crates, I restructured the build so it was a bit more consistent.  Also, similar to the mac build, the build script can handle verifying dependencies, so the build pre-req install instructions are much shorter.


metadave:
+1

PR #1264, opened 2016-09-20
move platform dependent calls from hab to core

smurawski:
First steps to getting a functional hab command on Windows.

* wrap std::os::unix::ffi::OsStrExt and add a windows equal
* wrap std::os::unix::symlink and add a windows mock.  There are windows symlink calls but they differ for file and directory.

With this changeset, hab compiles but panics due to calling unimplemented user resolution from core.

ref: #1192 

### DIFF
--- a/components/core/src/os/filesystem/windows.rs
+++ b/components/core/src/os/filesystem/windows.rs
@@ -15,6 +15,7 @@
 use libc::{c_int, c_char};
 use std::ffi::CStr;
 use std::path::Path;
+use std::io;
 
 pub fn chown(r_path: *const c_char, uid: u32, gid: u32) -> c_int {
     unimplemented!();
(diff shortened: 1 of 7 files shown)


components/core/src/os/filesystem/windows.rs
```diff
@@ -25,7 +26,11 @@ pub fn chmod(r_path: *const c_char, mode: u32) -> c_int {
...
             false => 1,
-            true => 0
-        } 
+            true => 0,
+        }
     }
 }
+
+pub fn symlink<P: AsRef<Path>, Q: AsRef<Path>>(src: P, dst: Q) -> io::Result<()> {
+    unimplemented!();
```
reset: Are Junctions still a thing in Windows? How do they achieve symlinks these days (if at all)?

smurawski: Junctions are still a thing.  the rust standard library has functions for the symlinks in windows, but they require you to determine if the source is a directory or file and call the specific function.  Shouldn't be hard to implement here.

smurawski: Junctions can cross drive boundaries and are the only link type on windows that can do directory links.  Hard links have to be on the same physical and logical volume and can only reference individual files.  

 https://msdn.microsoft.com/en-us/library/windows/desktop/aa365006(v=vs.85).aspx

reset:
gif-keyboard-13850742219115196093

Great work man, so excited for hab on Windows!

PR #1270, opened 2016-09-21
Stub out uname and effective_uid to make hab run

smurawski:
This gets us a basic `hab` command.  `hab setup` and `hab install` minimally work.

We are just stubbing out the basic uname info that we'd expect from a windows box and pretending to be root.


smurawski:
progress on #1192

Issue #1105, opened 2016-07-25
Unclear message when token has incorrect scope


jtimberman:
In order to upload a package, the local user must have a GitHub token with the `user:email` scope. When they do, everything is super cool.

```
% hab pkg upload results/core-openssh-7.2p2-20160725172511-x86_64-linux.hart -z "$(cat ~/.github/habitat_token)"
☛ Artifact signed with core-20160612031944.pub
» Uploading origin key core-20160612031944.pub
→ Public key revision already exists in the depot core-20160612031944.pub
» Uploading results/core-openssh-7.2p2-20160725172511-x86_64-linux.hart
→ Exists core/cacerts/2016.04.20/20160612081125
→ Exists core/glibc/2.22/20160612063629
→ Exists core/linux-headers/4.3/20160612063537
→ Exists core/openssl/1.0.2h/20160708160802
→ Exists core/zlib/1.2.8/20160612064520
↑ Uploading results/core-openssh-7.2p2-20160725172511-x86_64-linux.hart
    613.97 KB / 613.97 KB | [========================================================================================================================] 100.00 % 326.79 KB/s  ✓ Uploaded core/openssh/7.2p2/20160725172511
★ Upload of core/openssh/7.2p2/20160725172511 complete.
```

However, if the user has created the token with more limited scope, such as forgetting to allow read access to `user:email`, the user has a different experience:

```
% hab pkg upload results/core-openssh-7.2p2-20160725172511-x86_64-linux.hart -z "$(cat ~/.github/habitat_token)"
☛ Artifact signed with core-20160612031944.pub
» Uploading origin key core-20160612031944.pub
→ Public key revision already exists in the depot core-20160612031944.pub
» Uploading results/core-openssh-7.2p2-20160725172511-x86_64-linux.hart
→ Exists core/cacerts/2016.04.20/20160612081125
→ Exists core/glibc/2.22/20160612063629
→ Exists core/linux-headers/4.3/20160612063537
→ Exists core/openssl/1.0.2h/20160708160802
→ Exists core/zlib/1.2.8/20160612064520
↑ Uploading results/core-openssh-7.2p2-20160725172511-x86_64-linux.hart
    613.97 KB / 613.97 KB | [==========================================] 100.00 % 648.30 KB/s  Unexpected response from remote
✗✗✗
✗✗✗ 503 Service Unavailable
✗✗✗
```

Since I can log into the API server and view the server-side log, I thought perhaps I would get more interesting information, however that was not the case:

[...]

PR #1277, opened 2016-09-23
Improve client & server error reporting

reset:
This change will attempt to convert non-200 HTTP responses from the
Depot and API servers into a NetError which now implements
std::error::Error and is bubbled up back to the `hab` crate. These
errors contain a code and a message to help identify what exactly
went wrong with the users remote request.

Additional changes around erorr handling have been made as well. We
will now continue to validate already cached sessions even if the
GitHub API is down and we also report errors from their API more
gracefully.

There was quite a bit of work here so I would recommend a reviewer
to read each commit if they'd like a full understanding of what all went in
to this PR!


thesentinels:

Thank you for contributing!

metadave:
+1

reset:
@habitat-sh/habitat-core-maintainers this one could use one more look over, I made some additional changes and commits over my 3 hour flight back home. This should wrap things up and get us to a good place!

eeyun:
Am I wrong that this PR also solves this open issue https://github.com/habitat-sh/habitat/issues/1105 ?

I realized last week this was opened before the discussion we had around cleaning up error handling related to the `401 Unauthorized` messages from users having the incorrect GH User Token scoping.

raskchanky:
gif-keyboard-12704431809050293618

raskchanky:
@eeyun It _looks_ like it would, but I'll defer to @reset for the final word on that.

PR #1290, opened 2016-09-27
Support windows command resolution

smurawski:
Adds search by PATHEXT extensions to the command search of PATH.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -201,11 +201,37 @@ pub fn find_command(command: &str) -> Option<PathBuf> {
                 if candidate.is_file() {
                     return Some(candidate);
                 }
+                else {
+                    match find_command_with_pathext(&candidate) {
+                        Some(result) => return Some(result),
+                        None => {}
+                    } 
+                }
             }
             None
         }
         None => None,
     }
+}
+
+// Windows relies on path extensions to resolve commands like `docker` to `docker.exe`
+// Path extensions are found in the PATHEXT environment variable.
+fn find_command_with_pathext(candidate: &PathBuf) -> Option<PathBuf> {
+    match henv::var_os("PATHEXT") {
+        Some(pathexts) => {
+            for pathext in env::split_paths(&pathexts) {
+                let mut source_candidate = candidate.to_path_buf();
+                let extension = pathext.to_str().unwrap().trim_matches('.');
+                source_candidate.set_extension(extension);
+                let current_candidate = source_candidate.to_path_buf(); 
+                if current_candidate.is_file() {
+                    return Some(current_candidate);
+                }
+            }
+        },
+        None => return None,
+    };
+    None
 }
 
 /// Returns whether or not the current process is running with a root effective user id or not.
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -185,6 +185,48 @@ pub fn svc_var_path(service_name: &str) -> PathBuf {
 ///
 /// If the command represents an absolute path, then the `PATH` seaching will not be performed. If
 /// no absolute path can be found for the command, then `None` is returned.
+///
+/// On Windows, the PATHEXT environment variable contains common extensions for commands, 
+/// for example allowing "docker.exe" to be found when searching for "docker".
+///
+/// # Examples
+///
+/// Behavior when the command exists on PATH:
+///
+/// ```
+///
+/// use std::env;
+/// use std::fs;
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -201,13 +201,39 @@ pub fn find_command(command: &str) -> Option<PathBuf> {
...
+        Some(pathexts) => {
+            for pathext in env::split_paths(&pathexts) {
+                let mut source_candidate = candidate.to_path_buf();
+                let extension = pathext.to_str().unwrap().trim_matches('.');
+                source_candidate.set_extension(extension);
+                let current_candidate = source_candidate.to_path_buf(); 
+                if current_candidate.is_file() {
+                    return Some(current_candidate);
+                }
+            }
```
reset: Like in Ruby, the last line of a function is the return value in Rust. Short-circuiting the function with `return Some(T)` on 228 is great, but we should fall through to a None below this line. That would allow you to remove the `;` on 233 & the explicit call to `return None` on 232.

These are small suggestions but they'll make the function control flow read a bit easier and be more "rust-like"

components/core/src/fs.rs
```diff
@@ -201,13 +201,39 @@ pub fn find_command(command: &str) -> Option<PathBuf> {
                 if candidate.is_file() {
                     return Some(candidate);
                 }
+                else {
+                    match find_command_with_pathext(&candidate) {
```
reset: The match here would be unnecessary if the `;` on line 202 was removed along with the `None` on line 211. `find_command_with_pathext` already returns an `Option` so if you just call that without the `;` appended to the end you'll get the return value you're looking for!

components/core/src/fs.rs
smurawski: Definitely appreciate the feedback on making it more rust-like!

components/core/src/fs.rs
smurawski: Thanks for the feedback!

I think that would only work for the first path in the path variable checked.  The second part of the match `None => {}` is what lets things proceed on to the next path to check.

I'm going to write some tests to validate the behavior.

reset: Nope, you're right. I looked quick and didn't notice we were in an iterator and trying to short-circuit out of it in this function as well

smurawski:
@reset I've made some changes to the structure and behavior and added a bunch of tests.  Mind giving it another 👀 ?

components/core/src/fs.rs
```diff
@@ -199,7 +241,13 @@ pub fn find_command(command: &str) -> Option<PathBuf> {
             for path in env::split_paths(&paths) {
                 let candidate = PathBuf::from(&path).join(command);
                 if candidate.is_file() {
+                    println!("Found {:?}", candidate);
```
reset: This looks like it's either debug output or something left over from iterating. If it's debug output let's use `debug!()` instead of `println!()` and if it's just left over from iterating remove it

components/core/src/fs.rs
```diff
@@ -208,7 +256,208 @@ pub fn find_command(command: &str) -> Option<PathBuf> {
...
+                setup_environment();
+                let target_path = canonicalize("./tests/fixtures/plan.sh").unwrap();
+                let result = find_command("plan.sh");
+                let found_path = result.unwrap();
+                assert_eq!(found_path, target_path);
+            }
+        }
+    }
+
+    #[cfg(target_os="windows")]
```
reset: What a neat way to group tests in a platform dependent way. I hadn't seen us use submodules in a test module yet but this is a nice pattern

smurawski: I started using the modules as a way to group the tests to keep the function names shorter, but it became handy when I needed to make one windows only :)

components/core/src/fs.rs
smurawski: Yeah, thought I killed all those.  I'll remove it.

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/163539457).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/163539457).

smurawski:
progress on #1192

PR #1293, opened 2016-09-27
Require less broad OAuth scopes

reset:
Previously we required the `user:email` scope on the API and
`user:email`/`read:org` on the front-end application. When feature flags
were added we needed the ability to read which GitHub teams a user was a
part of in order to determine which features flags to grant the user.

To accomplish this we began requiring the full `user` scope on the
back-end which is one way to grant permissions to read a user's teams,
but it also requires the user to allow us full read/write to their user
profile.

This change ensures the front-end and the back-end both require the same
OAuth scopes and removes the requirement for full access to the `user`
scope. We once again only require `user:email` and `read:org`

gif-keyboard-13473140267849866289


thesentinels:

Thank you for contributing!

raskchanky:
Looks good to me.

gif-keyboard-2490127458359523764

fnichol:
Nice, tighter is better here!

PR #1294, opened 2016-09-28
service_config::toml_merge handles toml::Tables (rebased w/ changed order)

robbkidd:
Rebased @metadave's `dp_toml_merge_table_fix` branch from #1185 and added on a fix for config value precedence in the merge. 

I'm not sure how to re-add the "types mismatch in config _values_" test, though.

### DIFF
--- a/components/sup/src/service_config.rs
+++ b/components/sup/src/service_config.rs
@@ -312,6 +312,28 @@ struct Cfg {
 fn toml_merge(left: &toml::Table, right: &toml::Table) -> toml::Table {
     let mut final_map = toml::Table::new();
     for (left_key, left_value) in left.iter() {
+        // attempt to merge nested tables
+        let a = right.get(left_key).and_then(|v| v.as_table());
+        let b = left.get(left_key).and_then(|v| v.as_table());
+        match (a, b) {
+            (Some(left_t), Some(right_t)) => {
+                let new_table = toml_merge(left_t, right_t);
+                let table_as_value = toml::Value::Table(new_table);
+                final_map.insert(left_key.clone(), table_as_value);
+                continue;
+            }
+            (Some(_), _) => {
+                // this occurs when the value type in left != type in right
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/service_config.rs
```diff
@@ -345,20 +368,20 @@ impl Cfg {
     }
 
     fn to_toml(&self) -> toml::Value {
-        let mut left = toml::Table::new();
-        if let Some(toml::Value::Table(ref right)) = self.default {
-            left = toml_merge(&left, right);
+        let mut output_toml = toml::Table::new();
```
metadave: I think `output_toml` is more clear, nice!

metadave: actually, all the vars in this block have better names, thank you!

components/sup/src/service_config.rs
```diff
@@ -698,6 +723,30 @@ mod test {
...
+        let default_config = "[server]
+            port = \"8080\"
+            shutdown-port = \"8005\"
+            redirect-port = \"8443\"";
+
+        let override_config = "server = { port=\"9090\" }";
+        let default_toml = toml::Parser::new(default_config).parse().unwrap();
+        let override_toml = toml::Parser::new(override_config).parse().unwrap();
+        {
+            let result = toml_merge(&override_toml, &default_toml);
```
metadave: I still think the order of the args to `toml_merge` is backwars, BUT, this PR fixes the issue and affects less code than switching the order of args. So, +1

metadave:
Excellent!

I'll merge when Travis is green.

gif-keyboard-13605179485467846138

components/sup/src/service_config.rs
robbkidd: Ya. I concur. This is a short term fix and there's probably goodness to be found in reworking the merge logic.

PR #1310, opened 2016-09-30
SWIM as an independent library

adamhjk:
This adds an independent habitat_swim library. It uses small UDP packets
(under 512 bytes) and a much simpler implementation.

* Simpler architecture - one server, two threads (Inbound / Outbound)
* All timing happens in the application, not the network
* Trace file support with TRACE_SWIM=1
* Test suite based on building a "network", with software based
  blacklisting and pausing for partitions.
* Uses Protobufs for the network serialization

TODO:

  * Implement sharing of membership status as piggyback
  * Finish test suite for SWIM basic, and SWIM+piggyback
  * Documentation

### DIFF
--- a/components/swim/src/error.rs
+++ b/components/swim/src/error.rs
@@ -1,0 +1,69 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use protobuf;
+
+use std::error;
+use std::fmt;
+use std::io;
+use std::result;
+
+pub type Result<T> = result::Result<T, Error>;
+
+#[derive(Debug)]
+pub enum Error {
+    BadMessage(String),
+    CannotBind(io::Error),
+    ProtobufError(protobuf::ProtobufError),
+    SocketSetReadTimeout(io::Error),
+    SocketSetWriteTimeout(io::Error),
+    ServerCloneError,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let msg = match *self {
+            Error::BadMessage(ref err) => format!("Bad Message: {:?}", err),
+            Error::CannotBind(ref err) => format!("Cannot bind to port: {:?}", err),
+            Error::ProtobufError(ref err) => format!("ProtoBuf Error: {}", err),
+            Error::SocketSetReadTimeout(ref err) => {
+                format!("Cannot set UDP socket read timeout: {}", err)
+            }
+            Error::SocketSetWriteTimeout(ref err) => {
+                format!("Cannot set UDP socket write timeout: {}", err)
+            },
+            Error::ServerCloneError => format!("Cannot clone the underlying UDP socket"),
+        };
+        write!(f, "{}", msg)
+    }
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        match *self {
+            Error::BadMessage(ref _err) => "Bad Protobuf Message; should be Ping/Ack/PingReq",
+            Error::CannotBind(ref _err) => "Cannot bind to port",
+            Error::ProtobufError(ref err) => err.description(),
+            Error::SocketSetReadTimeout(ref _err) => "Cannot set UDP socket read timeout",
+            Error::SocketSetWriteTimeout(ref _err) => "Cannot set UDP socket write timeout",
+            Error::ServerCloneError => "Cannot clone the underlying UDP socket",
+        }
+    }
+}
+
+impl From<protobuf::ProtobufError> for Error {
+    fn from(err: protobuf::ProtobufError) -> Error {
+        Error::ProtobufError(err)
+    }
+}
--- a/components/swim/src/member.rs
+++ b/components/swim/src/member.rs
@@ -1,0 +1,722 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Tracks membership. Contains both the `Member` struct and the `MemberList`.
+
+use std::collections::{hash_map, HashMap};
+use std::fmt;
+use std::iter::IntoIterator;
+use std::net::SocketAddr;
+use std::ops::{Deref, DerefMut};
+use std::sync::{Arc, RwLock};
+
+use uuid::Uuid;
+use rand::{thread_rng, Rng};
+use time::SteadyTime;
+
+use rumor::RumorKey;
+use message::swim::{Member as ProtoMember, Membership as ProtoMembership,
+                    Membership_Health as ProtoMembership_Health};
+
+/// How many nodes do we target when we need to run PingReq.
+const PINGREQ_TARGETS: usize = 5;
+
+/// The health of a node.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Health {
+    Alive,
+    Suspect,
+    Confirmed,
+}
+
+/// Maps our internal health to the wire protocols health.
+impl From<ProtoMembership_Health> for Health {
+    fn from(pm_health: ProtoMembership_Health) -> Health {
+        match pm_health {
+            ProtoMembership_Health::ALIVE => Health::Alive,
+            ProtoMembership_Health::SUSPECT => Health::Suspect,
+            ProtoMembership_Health::CONFIRMED => Health::Confirmed,
+        }
+    }
+}
+
+impl From<Health> for ProtoMembership_Health {
+    fn from(pm_health: Health) -> ProtoMembership_Health {
+        match pm_health {
+            Health::Alive => ProtoMembership_Health::ALIVE,
+            Health::Suspect => ProtoMembership_Health::SUSPECT,
+            Health::Confirmed => ProtoMembership_Health::CONFIRMED,
+        }
+    }
+}
+
+impl<'a> From<&'a Health> for ProtoMembership_Health {
+    fn from(pm_health: &'a Health) -> ProtoMembership_Health {
+        match pm_health {
+            &Health::Alive => ProtoMembership_Health::ALIVE,
+            &Health::Suspect => ProtoMembership_Health::SUSPECT,
+            &Health::Confirmed => ProtoMembership_Health::CONFIRMED,
+        }
+    }
+}
+
+impl fmt::Display for Health {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &Health::Alive => write!(f, "Alive"),
+            &Health::Suspect => write!(f, "Suspect"),
+            &Health::Confirmed => write!(f, "Confirmed"),
+        }
+    }
+}
+
+/// A member in the swim group. Passes most of its functionality along to the internal protobuf
+/// representation.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Member {
+    pub proto: ProtoMember,
+}
+
+impl Member {
+    /// Creates a new member with a unique UUID and an incarnation of zero.
+    pub fn new() -> Member {
+        let mut proto_member = ProtoMember::new();
+        proto_member.set_id(Uuid::new_v4().simple().to_string());
+        proto_member.set_incarnation(0);
+        Member { proto: proto_member }
+    }
+
+    /// Returns the socket address of this member.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if the address is un-parseable. In practice, it shouldn't be
+    /// un-parseable, since its set from the inbound socket directly.
+    pub fn socket_address(&self) -> SocketAddr {
+        match self.get_address().parse() {
+            Ok(addr) => addr,
+            Err(e) => {
+                panic!("Cannot parse member {:?} address: {}", self, e);
+            }
+        }
+    }
+}
+
+impl Deref for Member {
+    type Target = ProtoMember;
+
+    fn deref(&self) -> &ProtoMember {
+        &self.proto
+    }
+}
+
+impl DerefMut for Member {
+    fn deref_mut(&mut self) -> &mut ProtoMember {
+        &mut self.proto
+    }
+}
+
+impl From<ProtoMember> for Member {
+    fn from(member: ProtoMember) -> Member {
+        Member { proto: member }
+    }
+}
+
+impl<'a> From<&'a ProtoMember> for Member {
+    fn from(member: &'a ProtoMember) -> Member {
+        Member { proto: member.clone() }
+    }
+}
+
+impl From<SocketAddr> for Member {
+    fn from(socket: SocketAddr) -> Member {
+        let mut member = Member::new();
+        member.set_address(format!("{}", socket));
+        member
+    }
+}
+
+impl From<Member> for RumorKey {
+    fn from(member: Member) -> RumorKey {
+        RumorKey::new("member", member.get_id())
+    }
+}
+
+impl<'a> From<&'a Member> for RumorKey {
+    fn from(member: &'a Member) -> RumorKey {
+        RumorKey::new("member", member.get_id())
+    }
+}
+
+impl<'a> From<&'a &'a Member> for RumorKey {
+    fn from(member: &'a &'a Member) -> RumorKey {
+        RumorKey::new("member", member.get_id())
+    }
+}
+
+// This is a Uuid type turned to a string
+pub type UuidSimple = String;
+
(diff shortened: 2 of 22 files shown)


thesentinels:

Thank you for contributing!

components/swim/Cargo.lock
```diff
@@ -0,0 +1,313 @@
+[root]
```
reset: This lockfile shouldn't be in version control since we're using a common workspace. I see that you've got the right line in your `Cargo.toml` to enable the workspace so this was probably generated before you added that and can be safely deleted

components/swim/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+            Error::BadMessage(ref _err) => "Bad Protobuf Message; should be Ping/Ack/PingReq",
+            Error::IO(ref err) => err.description(),
+            Error::ProtobufError(ref err) => err.description(),
+        }
+    }
+}
+
+impl From<io::Error> for Error {
+    fn from(err: io::Error) -> Error {
+        Error::IO(err)
```
reset: We've got this pattern throughout our codebase but it's the source of our most indestinguishable error messages. An IO Error from the IO module doesn't include any contextual information about where/what went wrong, so I think it's not appropriate to call `Error::from(T)` on these error types.

I'd recommend nipping this in the bud before people keep using it by starting here by not implementing `From<io::Error>` for your error type and generating the error messages with more context instead

components/swim/src/member.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn new() -> Member {
+        let mut proto_member = ProtoMember::new();
+        proto_member.set_id(Uuid::new_v4().simple().to_string());
+        proto_member.set_incarnation(0);
+        Member{
+            proto: proto_member,
+        }
+    }
+
+    pub fn new_from_proto(proto: ProtoMember) -> Member {
```
reset: This function is just fine, but the `Into<T>` trait may be the more "rust appropriate" way to do it. That would let you call `proto.into()` instead of `Member::new_from_proto(proto)`

components/swim/src/member.rs
```diff
@@ -0,0 +1,259 @@
...
+        Member {
+            proto: proto,
+        }
+    }
+
+    pub fn socket_address(&self) -> SocketAddr {
+        match self.get_address().parse() {
```
reset: I would perform this work in the constructor(s) to fail fast. That way a message doesn't propogate through the system any further than necessary before we hit the `panic!()` on line 70

components/swim/src/member.rs
```diff
@@ -0,0 +1,259 @@
...
+impl MemberList {
+    pub fn new() -> MemberList {
+        MemberList{
+            members: HashMap::new(),
+            health: HashMap::new(),
+        }
+    }
+
+    pub fn insert(&mut self, member: Member, health: Health) -> Option<Member> {
+        self.health.insert(String::from(member.get_id()), health);
```
reset: `member.take_id()` will transfer ownership to the caller so you don't need to allocate a new string with `String::from()`.

The Protobuf generated code will create two functions for types which don't implement copy - `get_X()` and `take_X()` where get returns a reference and take returns the type

components/swim/src/member.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.members.insert(String::from(member.get_id()), member)
+    }
+
+    pub fn health_of(&self, member: &Member) -> Option<&Health> {
+        self.health.get(member.get_id())
+    }
+
+    pub fn insert_health(&mut self, member: &Member, health: Health) -> Option<Health> {
+        self.health.insert(String::from(member.get_id()), health)
```
reset: Same comment here as for line 109

components/swim/src/member.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    pub fn members(&self) -> Vec<&Member> {
+        self.members.values().collect()
+    }
+
+    pub fn check_list(&self) -> Vec<Member> {
```
reset: I don't think you need to allocate a new list of Members here, just a ref should be alright by looking at the callers of `check_list()`. You can get a mutable member list with `values_mut()` to pass to `thread_rng::shuffle()` if that's what the motivation was.

components/swim/src/message/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod swim;
```
reset: Missing the license header at the top of this file

components/swim/src/server/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub fn rounds(&self) -> isize {
+        self.rounds.load(Ordering::SeqCst)
+    }
+
+    pub fn update_round(&self) {
+        let current_round = self.rounds.load(Ordering::SeqCst);
+        match current_round.checked_add(1) {
+            Some(_number) => {
+                self.rounds.fetch_add(1, Ordering::SeqCst);
+                return
```
reset: Explicit call to `return` here is unncessary, you'll drop out of the match and return `()`

components/swim/src/server/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            } ,
+            None => {
+                debug!("Exceeded an isize integer in rounds. Congratulations, this is a very long running supervisor!");
+                self.rounds.store(0, Ordering::SeqCst);
+                return
```
reset: Same comment here as line 93

components/swim/src/server/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            panic!("You should never, ever get here, judy");
+        });
+        let _ = thread::Builder::new().name("outbound".to_string()).spawn(move || {
+            outbound::Outbound::new(&server_b, rx_inbound, timing).run();
+            panic!("You should never, ever get here, bob");
+        });
+    }
+
+    pub fn add_to_blacklist(&self, addr: SocketAddr) {
+        let mut blacklist = self.blacklist.write().expect("Write lock for blacklist is poisoned");
```
reset: Good usage of `expect()`, this is exactly how you write assertions ;)

components/swim/src/server/inbound.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub fn run(&self) {
+        let mut recv_buffer: Vec<u8> = vec![0; 1024];
+        loop {
+            if self.server.pause.load(Ordering::Relaxed) {
+                thread::sleep(Duration::from_millis(100));
+                continue;
+            }
+            match self.server.socket.recv_from(&mut recv_buffer[..]) {
+                Ok((length, addr)) => {
+                    if self.server.check_blacklist(&addr) {
```
reset: Love how this code has testing hooks and escape hatches built in. This is excellent!

components/swim/src/server/inbound.rs
```diff
@@ -0,0 +1,171 @@
...
+                        println!("Not processing message from {} - it is blacklisted", addr);
+                        continue;
+                    }
+                    let msg: Swim = match protobuf::parse_from_bytes(&recv_buffer[0..length]) {
+                        Ok(msg) => msg,
+                        Err(e) => {
+                            error!("Error parsing protobuf: {:?}", e);
```
reset: We may want to forcefully disconnect the caller if they are sending us garbage. I know this is a back-end service but I'm a little paranoid

components/swim/src/server/inbound.rs
```diff
@@ -0,0 +1,171 @@
...
+                Some(member) => {
+                    member.socket_address()
+                },
+                None => {
+                    error!("PingReq request {:?} for invalid target", msg);
+                    return;
+                }
+            }
+        };
+        outbound::ping(self.server, target_addr, Some(Member::new_from_proto(msg.get_pingreq().get_from().clone())));
```
reset: You can take ownership instead of cloning with `msg.take_pingreq().take_from()` here

components/swim/src/server/inbound.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn process_ack(&self, addr: SocketAddr, msg: Swim) {
+        trace_swim!(&self.server, "recv-ack", &format!("{}", addr), Some(&msg));
+        info!("Ack from {}@{}", msg.get_ack().get_from().get_id(), addr);
+        if msg.get_ack().has_forward_to() {
+            {
```
reset: What's the purpose of the additional block (`{}`) here?

components/swim/src/server/inbound.rs
```diff
@@ -0,0 +1,171 @@
...
+        match self.tx_outbound.send((addr, msg)) {
+            Ok(()) => {},
+            Err(e) => panic!("Outbound thread has died - this shouldn't happen: #{:?}", e)
+        }
+    }
+
+    fn process_ping(&self, addr: SocketAddr, mut msg: Swim) {
+        trace_swim!(&self.server, "recv-ping", &format!("{}", addr), Some(&msg));
+        if msg.get_ping().has_forward_to() {
+            outbound::ack(self.server, addr, Some(Member::new_from_proto(msg.get_ping().get_forward_to().clone())));
```
reset: Ownership here as well, no need to clone

components/swim/tests/common/net.rs
```diff
@@ -0,0 +1,197 @@
...
+            if SteadyTime::now() <= next_period {
+                thread::sleep(Duration::from_millis(100));
+            } else {
+                return
+            }
+        }
+    }
+}
+
+macro_rules! assert_health_of {
```
reset: These macros are SUPER clean and useful. This makes test suites wonderful

components/swim/Cargo.lock
adamhjk: Removed.

components/swim/src/error.rs
adamhjk: This is a great suggestion. 

Does anyone have an idea for something more ergonomic than the full match that still allows for the early return? I couldn't find one.

components/swim/src/member.rs
adamhjk: This is gloriously better. Done.

components/swim/src/member.rs
adamhjk: The socket address can change in the life of the program (in particular for a given member) based on whether or not we receive pings directly from the member. We check it more than once, actually - once here, and then earlier. It's really just a guard against manual addresses being bullshit on read. In the "normal" life of the program, all the addresses are direct translations of SocketAddr, so we're going to pass.

components/swim/src/member.rs
adamhjk: I do need to allocate a new string, because its the key for the hash, and it has to be owned to be the key. I can't take it, because I need it to still be in the member when I insert it as a value.

components/swim/src/member.rs
adamhjk: Same reply! 🎱

components/swim/src/member.rs
adamhjk: You do need it to be owned. The member list changes async behind the scenes - not just in order, but the contents of a given member may change while we're walking the list in a given round. By making the copy, we're free to forget about any state transition that happens *during* the walk from another source, which is what we want.

components/swim/src/message/mod.rs
adamhjk: Good catch!

components/swim/src/server/inbound.rs
adamhjk: That could be a thing we use the blacklist for eventually. I'm inclined to take this as a future note, and see if we want to add it later.

components/swim/src/server/inbound.rs
adamhjk: Yep, good point.

components/swim/src/server/inbound.rs
adamhjk: Nervous tick around lifetimes of locks.

adamhjk: (I removed it)

components/swim/src/server/inbound.rs
adamhjk: Roger

components/swim/src/error.rs
reset: What I had in mind to solve this problem in the other crates would be to do something along these lines:

```rust
match func_might_err() {
  Ok(val) => val,
  Err(io::Error) => {
    msg = "contextual reason for failure";
    Error::ContextualReason(msg)
  },
  Err(e) => Error::from(e),
}
```

components/swim/src/member.rs
reset: Ok that sounds reasonable!

components/swim/src/member.rs
reset: Ah yeah, I should have looked at the line below ;)

adamhjk:
More commits - this makes the test suite work for all the SWIM Basic tests, and adds a little bit of Ruby to snarf the swim trace files and print a UML diagram.

adamhjk:
This is ready for final review. Next step is to move on to the gossip layer.

components/swim/src/server/inbound.rs
```diff
@@ -0,0 +1,198 @@
...
+                      addr,
+                      msg.get_ack().get_forward_to().get_id(),
+                      msg.get_ack().get_forward_to().get_address(),
+                      );
+                msg.mut_ack().mut_from().set_address(format!("{}", addr));
+                outbound::forward_ack(self.server, forward_to_addr, msg);
+                return;
+            }
+        }
+        let membership = {
```
reset: I don't think this scoping is necessary

components/swim/src/server/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub blacklist: Arc<RwLock<HashSet<SocketAddr>>>,
+}
+
+impl Clone for Server {
+    fn clone(&self) -> Server {
+        let socket = match self.socket.try_clone() {
+            Ok(socket) => socket,
+            Err(e) => {
+                error!("Failed to clone socket; trying again: {:?}", e);
+                return self.clone();
```
reset: This will blow the stack if you aren't ever able to clone the socket. I know this is a super fatal case, but any application that consumes this crate's Server and uses it could die from a stack overflow which might not be how they want to clean up ;). Maybe a finite number of retries before we call it good or don't implement `Clone` for server but instead expose a `try_clone()` function of your own

components/swim/src/server/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    ///
+    /// * Returns `Error::CannotBind` if the socket cannot be bound
+    /// * Returns `Error::SocketSetReadTimeout` if the socket read timeout cannot be set
+    /// * Returns `Error::SocketSetWriteTimeout` if the socket write timeout cannot be set
+    pub fn new<A: ToSocketAddrs>(addr: A, member: Member, trace: Trace) -> Result<Server> {
+        let socket = match UdpSocket::bind(addr) {
+            Ok(socket) => socket,
+            Err(e) => return Err(Error::CannotBind(e)),
+        };
+        match socket.set_read_timeout(Some(Duration::from_millis(1000))) {
```
reset: This would be better written with [map_err](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err)

```
try!(socket.set_read_timeout(Some(Duration::from_millis(1000))).map_err(|e| Error::SocketSetReadTimeout(e)));
```

components/swim/src/server/outbound.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    /// Run the outbound thread. Gets a list of members to pinmg, then walks the list, probing each
+    /// member.
+    ///
+    /// If the probe completes before the next protocol period is scheduled, waits for the protocol
+    /// period to finish before starting the next probe.
+    pub fn run(&mut self) {
+        loop {
+            if self.server.pause.load(Ordering::Relaxed) {
```
reset: Love how this stuff is all testable by default

components/swim/src/member.rs
```diff
@@ -0,0 +1,706 @@
...
+    }
+
+    /// Inserts a member into the member list with the given health.
+    pub fn insert(&self, member: Member, health: Health) -> bool {
+        let share_rumor: bool;
+        let mut start_suspicion: bool = false;
+        let mut stop_suspicion: bool = false;
+
+        // If we have an existing member record..
+        if let Some(current_member) = self.members
```
adamhjk: Yeah, it's gotten pretty hairy. It's fairly similar to the logic that's in the existing implementation, which also look similar. It's made worse by needing to not contend for the locks.

components/swim/src/member.rs
```diff
@@ -0,0 +1,706 @@
...
+
+    /// Returns a protobuf membership record for the given member id.
+    pub fn membership_for(&self, member_id: &str) -> ProtoMembership {
+        let mut pm = ProtoMembership::new();
+        let health = self.health
+            .read()
+            .expect("Health lock is poisoned")
+            .get(member_id)
+            .expect("Should have membership before calling membership_for")
+            .clone(); // This is basically free, since its just an emum
```
adamhjk: Nice!

components/swim/src/member.rs
```diff
@@ -0,0 +1,706 @@
...
+        let mhealth = match health {
```
adamhjk: You are the master of seeing where a Trait could simplify code. It's so nice.

components/swim/src/server/inbound.rs
adamhjk: Yep.

components/swim/src/server/mod.rs
adamhjk: Yep. One of the things I'm worried about here is that very little of this code should be able to fail. We really should only be panicing if we have locks we can't fix.

components/swim/src/server/mod.rs
adamhjk: Much cleaner.

components/swim/src/server/outbound.rs
adamhjk: Me too. It really helped my brain to think about it as integration testing rather than relying purely on functional tests.

adamhjk:
Pushed changes addressing @reset comments. Going to merge when the CI clears, and move on to the gossip layer.

adamhjk:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/167117565).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/167117565).

PR #1320, opened 2016-10-03
[hab] Retire `core/hab-dynamic` Plan.

fnichol:
This change removes the "hab-dynamic" Plan (and therefore package) from
the source tree.

Some background:

> In the early days of Habitat...

Originally all Rust-implemented programs in Habitat were dynamically
compiled for Linux, that is they loaded shared libraries such as
libsodium, libarchive, libc, etc. As a team, our desired starting point
for a new user was a small, portable download from which we could
"hydrate" (i.e. install) Habitat on a system. To this end we started
producing a statically compiled ELF binary for the `hab` program. In
other words a portable binary with no linked libraries required at run
time. As step towards this goal for a short period we produced a
`core/hab-static` package to test this support. Once we were satisfied
with its performance, we cut this statically compiled behavior over to
the canonical `core/hab` package which left us with the old dynamically
compiled Plan. At the time, we simply renamed this to be the
`core/hab-dynamic` package. Clearly though, there is no longer any
active work on this particular artifact and therefore we can bid it a
fond farewell.

---

Also (burying the lead here), this will save us 5-10 minutes on the
release process as we have one less package to build.

gif-keyboard-14046575732250768551


thesentinels:

Thank you for contributing!

reset:
gif-keyboard-9252270238555812706

PR #1329, opened 2016-10-05
[hab] Attempt to call `sudo` on `hab studio *` for non-root users.

fnichol:
This change set comprises 2 parts. The contents of each commit message is aggregated below:

# [hab] Attempt to call `sudo` on `hab studio *` for non-root users.

This change only currently affects Linux non-root users who attempt to run the following subcommands:

* `hab studio *`
* `hab pkg build *`

The behavior introduced here will attempt to re-exec the command with prepended with a `sudo -p "[sudo hab-studio] password for %u:" -E` in front of the original `hab` subcommand. If the `sudo` command cannot be found on the `PATH` then an error message is displayed to the user and the program terminates with a non-zero exit code. The end result is that, for non-root users on systems using sudoers:

    hab pkg build ...

Will be the same as running:

    sudo hab pkg build ...

Examples
--------

Running as a non-root user with appropriate permissions in sudoers:

```
> hab studio enter
[sudo hab-studio] password for jdoe:
   hab-studio: Creating Studio at /hab/studios/home--jdoe (default)
   hab-studio: Importing jdoe secret origin key
★ Imported secret origin key jdoe-20161005175939.
   hab-studio: Entering Studio at /hab/studios/home--jdoe (default)
   hab-studio: Exported: HAB_ORIGIN=jdoe

[1][default:/src:0]#
```

Running as a non-root user when the `sudo` command cannot be found on `PATH`:

```
> sudo mv /usr/bin/sudo /usr/bin/ssudo

> hab studio enter
∅ Could not find the `sudo' command on the filesystem or in PATH.
∅ Running Habitat Studio requires root or administrator privileges. Please retry this command as a super user or use a privilege-granting facility such as sudo.

✗✗✗
✗✗✗ Root or administrator permissions required to complete operation
✗✗✗
```

# [core] Add core `become_command()` to abstract Unix/Windows exec diffs.

This change extracts the formerly named `exec_command()` function from
hab's `exec` module into a common interface `become_command()` in core's
newly created `os::process` module. The intent of this function
philisophically is to turn the current process into some other process
that takes over for the duration of the new process. In other words, the
existing process ceases to do any work.

There are 2 implementations for this function, which depend on the
underlying target platform:

Unix (i.e. Linux, Mac, BSD, etc)
--------------------------------

This implementation uses [`execvp(3)`][] which is now provided by the
[`exec()`][] function on Rust's stdlib [`CommandExt`][] trait. This
insures that there isn't a deeply nested process chain and that signals
such as `Ctrl+C` correctly kill all related processes. If a developer
needs to create a child process and supervise it or act on the child's
results then this is **not** the appropriate API to invoke, hence the
name `become_command()`.

[`execvp(3)`]: https://linux.die.net/man/3/execvp
[`exec()`]:
https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.exec
[`CommandExt`]:
https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html

Windows
-------

This implementation uses Rust's stdlib [`Command`] to execute the
command as a child process and then exits with the child's exit code.
Due to the lack of `exec*()` behavior. Note that there is an older
`_execv()` in a POSIX compatibility layer in MSVCRT, but the native
approach is to use a child process here, and so that's what we do.
Also note that while the `hab` process is not replaced by the new
process, we explicitly terminate the `hab` process after the child
process completes. If a developer needs to create a child process and
supervise it or act on the child's results then this is **not** the
appropriate API to invoke, hence the name `become_command()`.

[`Command`]:
https://doc.rust-lang.org/std/process/struct.Command.html#method.exec

### DIFF
--- a/components/core/src/os/process/linux.rs
+++ b/components/core/src/os/process/linux.rs
@@ -1,0 +1,39 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ffi::OsString;
+use std::path::PathBuf;
+use std::os::unix::process::CommandExt;
+use std::process::Command;
+
+use error::Result;
+
+pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    become_exec_command(command, args)
+}
+
+/// Makes an `execvp(3)` system call to become a new program.
+///
+/// Note that if sucessful, this function will not return.
+///
+/// # Failures
+///
+/// * If the system call fails the error will be returned, otherwise this function does not return
+fn become_exec_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    debug!("Calling execvp(): ({:?}) {:?}", command.display(), &args);
+    let error_if_failed = Command::new(command).args(&args).exec();
+    // The only possible return for the above function is an `Error` so return it, meaning that we
+    // failed to exec to our target program
+    return Err(error_if_failed.into());
+}
--- a/components/core/src/os/process/windows.rs
+++ b/components/core/src/os/process/windows.rs
@@ -1,0 +1,39 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ffi::OsString;
+use std::path::PathBuf;
+use std::process::{self, Command};
+
+use error::Result;
+
+pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    become_child_command(command, args)
+}
+
+/// Executes a command as a child process and exits with the child's exit code.
+///
+/// Note that if sucessful, this function will not return.
+///
+/// # Failures
+///
+/// * If the child process cannot be created
+fn become_child_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    debug!("Calling child process: ({:?}) {:?}",
+           command.display(),
+           &args);
+    let status = try!(Command::new(command).args(&args).status());
+    // Let's honor the exit codes from the child process we finished running
+    process::exit(status.code().unwrap())
+}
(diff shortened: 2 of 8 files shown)


thesentinels:

Thank you for contributing!

rrxtns:
@fnichol while I'm sad to see my recent contribution to exec.rs be summarily removed, this seems like 200% the right thing.  I can poke at this in Windows land adhoc if you need.

components/core/src/os/process/linux.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use std::ffi::OsString;
+use std::path::PathBuf;
+use std::os::unix::process::CommandExt;
+use std::process::Command;
+
+use error::Result;
+
+pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    become_exec_command(command, args)
```
reset: Just curious why we're delegating to a private func here?

components/core/src/os/process/windows.rs
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+use std::ffi::OsString;
+use std::path::PathBuf;
+use std::process::{self, Command};
+
+use error::Result;
+
+pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    become_child_command(command, args)
```
reset: Same question here

components/core/src/os/process/windows.rs
```diff
@@ -0,0 +1,41 @@
...
+///
+/// Note that if sucessful, this function will not return.
+///
+/// # Failures
+fn become_child_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
+    debug!("Calling child process: ({:?}) {:?}",
+           command.display(),
+           &args);
+    let mut command = Command::new(command);
+    for arg in &args {
```
reset: No need to iterate over and call `arg()` for each, there's [args()](https://doc.rust-lang.org/std/process/struct.Command.html#method.args)!

components/hab/src/command/studio.rs
```diff
@@ -93,24 +98,52 @@ mod inner {
...
+        // Otherwise we will try to re-run this program using `sudo`
+        match find_command(SUDO_CMD) {
+            Some(sudo_prog) => {
+                let mut args: Vec<OsString> =
+                    vec!["-p".into(), "[sudo hab-studio] password for %u: ".into(), "-E".into()];
+                args.append(&mut env::args_os().collect());
+                Ok(try!(process::become_command(sudo_prog, args)))
+            }
+            None => {
+                try!(ui.warn(format!("Could not find the `{}' command on the filesystem or in \
```
reset: I think the error here should read something like "Could not find the `{}` command. Is it in your PATH?". We should  always try and provide a way to fix the error along with what went wrong

fnichol: Sounds good with me--I'm pretty sure I copy/pasta'd a similar error from elsewhere.

fnichol:
I'll update, rebase, repush and bust the Travis cache if I get the same failure.

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/165424666).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/165424666).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/165424666) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/165944968).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/165944968).

PR #1331, opened 2016-10-06
[studio] Properly propagate the exit code when `build` fails.

### LINKED ISSUE #1285: hab pkg build returns code 0 on error
OS = Linux
hab = 0.9.3/20160916191534

hab pkg build . 
Output
```
....
cp: omitting directory '/src/habitat/../source/app.tar.gz'
   app: Build time: 0m14s
   app: Exiting on error
Script done, file is /src/results/logs/..2016-09-27-135632.log
# echo $?
0
```

fnichol:
Can I write more words to explain this fix than the fix itself? It seems
I already did.

The `hab-studio build` subcommand (which is also invoked via `hab studio
build` and `hab pkg build`) has a shell function called [`record()`][]
to log the entire build activity in a log file via the [`script(1)`][]
program. It turns out that the exit code of the child process is not
returned by default as I initially assumed and therefore the `script`
call always return an exit code of `0`.

This fix adds a `--return` flag (short form of which is `-e`) to return
the exit code of the child proces--in our case this is the `build`
program itself. Adding this exit code return properly bubbles back the
exit code to its parent process (the bash program inside the
`chroot(1)`) and we get a proper exit code when `hab-studio`
terminates...as we expected in the first place.

Note that I went with the short flag form of `-e` in an attempt to make
this portable across to the BusyBox version if it happens to be
activated. This is one situation where I deviated from my defaulting to
always use the long, explicit version in calling programs and
scripts--we can't all remember every `docker` and `netstat` flag, right?

Closes #1285

[`record()`]:
https://github.com/habitat-sh/habitat/blob/f470c41146bd3c796adb0a610d3994497557dde0/components/studio/bin/hab-studio.sh#L604-L622
[`script(1)`]:
http://man7.org/linux/man-pages/man1/script.1.html

/cc @reset I think this is going to help you out this week...

gif-keyboard-143735727654487103

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -617,7 +617,7 @@ record() {
     $bb mkdir -p $($bb dirname $log)
     unset BUSYBOX LOGDIR
 
-    $bb script -c "$bb env -i $env $cmd $*" $log
+    $bb script -c "$bb env -i $env $cmd $*" -e $log
   ); return $?
 }
 


thesentinels:

Thank you for contributing!

reset:
gif-keyboard-16729791309457023454

Of course it would be a two character change, good work @fnichol

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/165399791).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/165399791).

fnichol:
Noice!

PR #1337, opened 2016-10-07
Make Admin/Elevated Admin Token == root for hab commands

smurawski:
This plumbs in a process token check to see if the user account has admin rights and, if UAC is enabled, if the token is elevated or filtered.

Admins (without UAC) and admins with elevated tokens will be treated as root by the `hab` command.

There are some additions to the core crate - an admincheck.c which is a wrapper around the win32 api calls required to validate above in an easier to consume fashion, a build.rs file to compile and link that C code, as well as updates to the build script to handle the native code linking to system apis.

### DIFF
--- a/components/core/src/os/users/admincheck.c
+++ b/components/core/src/os/users/admincheck.c
@@ -1,0 +1,90 @@
+#include <windows.h>
+#include <winnt.h>
+#include <ShlObj.h>
+
+#pragma comment (lib,"shell32.lib")
+
+TOKEN_ELEVATION_TYPE s_elevationType = TokenElevationTypeDefault;
+BOOL                 s_bIsAdmin = FALSE;
+
+// returns 0 if an admin
+// returns 1 if not
+int GetElevationState()
+{
+    if (GetProcessElevation(&s_elevationType, &s_bIsAdmin)) {
+        switch(s_elevationType) {
+            // Default user or UAC is disabled
+            case TokenElevationTypeDefault:  
+                if (IsUserAnAdmin()) {
+                    return 0;
+                } else {
+                    return 1;
+                }
+            break;
+            // Process has been successfully elevated
+            case TokenElevationTypeFull:
+                if (IsUserAnAdmin()) {
+                    return 0;
+                } else {
+                    return 1;
+                }
+            break; 
+            // Process is running with limited privileges
+            case TokenElevationTypeLimited:
+                if (IsUserAnAdmin()) {
+                    // Not sure what capabilities a filtered admin has
+                    // So, for now, I'll treat that as non-admin
+                    return 1;
+                } else {
+                    return 1;
+                }
+            break;
+        }
+        return 1;
+    }
+    return 1;
+} 
+
+BOOL GetProcessElevation(TOKEN_ELEVATION_TYPE* pElevationType, BOOL* pIsAdmin) {
+   HANDLE hToken = NULL;
+   DWORD dwSize;
+
+   // Get current process token
+   if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken))
+      return(FALSE);
+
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

components/core/src/os/users/admincheck.c
```diff
@@ -0,0 +1,90 @@
...
+#include <ShlObj.h>
+
+#pragma comment (lib,"shell32.lib")
+
+TOKEN_ELEVATION_TYPE s_elevationType = TokenElevationTypeDefault;
+BOOL                 s_bIsAdmin = FALSE;
+
+// returns 0 if an admin
+// returns 1 if not
+int GetElevationState()
```
chefsalim: If this is an int masquerading as a bool that we need for unsafe interface, consider changing name to be something like "IsElevatedAdmin" or some such.

components/core/src/os/users/admincheck.c
```diff
@@ -0,0 +1,90 @@
...
+{
+    if (GetProcessElevation(&s_elevationType, &s_bIsAdmin)) {
+        switch(s_elevationType) {
+            // Default user or UAC is disabled
+            case TokenElevationTypeDefault:  
+                if (IsUserAnAdmin()) {
+                    return 0;
```
chefsalim: Consider single return rather than multiple points of return.

components/core/src/os/users/admincheck.c
```diff
@@ -0,0 +1,90 @@
...
+                    // So, for now, I'll treat that as non-admin
+                    return 1;
+                } else {
+                    return 1;
+                }
+            break;
+        }
+        return 1;
+    }
+    return 1;
```
chefsalim: Consider simplifying the function to a single expression.

components/core/src/os/users/admincheck.c
smurawski: Thanks @chefsalim - first of all, hopefully this is temporary and we can move to straight ffi against, the requisite apis, but I left this an int rather than a bool in case I need to do more discreet separation of user elevation types later (could drop in an enum or something).

components/core/src/os/users/admincheck.c
raskchanky: Honestly not trying to bike shed here, but I think returning early is a great thing.  Keeping track of the state of your function all the way through, just for the purpose of having a single return at the end feels like additional complexity for no real benefit.

raskchanky:
I don't know Windows very well, but I think I followed most of what was going on here, and it looks good to me (for whatever that's worth 😁)

gif-keyboard-4307876432385209987

components/core/src/os/users/admincheck.c
smurawski: I updated the function to return different values based on the state of the user token, which will be useful in upcoming "process permission awareness" scenarios.   It still has the early returns, but has more types.  Again, this is one of those tech debt items that should be replaced with a "rusty" solution, but the win32 API doesn't make that easy.

chefsalim: Yeah not a big deal... @raskchanky Returning early is fine but we have 7 return statements so we're returning early, middle and late ;) Having multiple points of returns leaves us more exposed to leaks over time as code is maintained is the school of thought.

chefsalim:
LGTM 👍

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/165944069).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/165944069).

smurawski:
progress on #1192

Issue #1089, opened 2016-07-20
Package names with a dot generate invalid default group


pharaujo:
I'm trying to package in habitat an upstream service that has a dot in its name (e.g. `service.something`).

Building the `.hart` works with no warnings. However, when running the service without specifying a `--group`, the generated group becomes `service.something.default` which can't be used with `hab config apply` because it's invalid.

My suggestions are either making these names an error, emitting a warning in the build process, or removing the dots from the generated default group.

I'm running `hab 0.8.0/20160708161531`.

metadave:
thanks @pharaujo, we'll take a look. 

related Slack discussion: https://habitat-sh.slack.com/archives/general/p1469025611000427

metadave:
@habitat-sh/habitat-core-maintainers what are your thoughts on limiting package names during build to `^[A-Za-z0-9_-]+$`

beddari:
As a user I'm perfectly happy about your suggestion @metadave

reset:
@metadave packages are most certainly not supposed to be allowed to have a dot (`.`) in their name. Oops!

davidwrede:
Docs will be corrected shortly. Apologies for any confusion!

metadave:
@davidwrede we'll also need to update `hab-plan-build.sh` with the [regex above](https://github.com/habitat-sh/habitat/issues/1089#issuecomment-234004136), but having doc updates would be excellent!

PR #1345, opened 2016-10-10
[build] Validate characters in plan name

chefsalim:

This change address https://github.com/habitat-sh/habitat/issues/1089 .  The package name will be validated to conform to the supported format, and plan build will fail if it does not.

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -2300,6 +2300,9 @@ if [[ -n "$HAB_ORIGIN" ]]; then
   pkg_origin="$HAB_ORIGIN"
 fi
 
+# Validate metadata
+build_line "Validating plan metadata"
+
 # Test for all required metadata keys
 required_variables=(
   pkg_name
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

reset:
@chefsalim looks good! This will cover the client validation but we should also make sure we're validating with the same rules on the server side - specifically when uploading a package. Not necessary for this PR, merge when ready!

reset:
gif-keyboard-9268426778400619820

chefsalim:
Cool, ok will open a new card for server validation.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -2314,6 +2317,12 @@ do
   fi
 done
 
+# Test to ensure package name contains only valid characters
+if [[ ! "${pkg_name}" =~ ^[A-Za-z0-9_-]+$ ]];
```
raskchanky: Does the hyphen in the regex need to be backslash escaped?

chefsalim: No doesn't seem to need it - it works correctly with a hypen in the package name.

PR #1363, opened 2016-10-14
[core] Validate cache path when running hab studio

chefsalim:

This change adds a check to make sure we catch a case that was missed when automatically elevating a user running hab studio. Specifically, we want to make sure that the user's .hab directory is present (ie, hab setup has been done) before we automatically elevate. If we auto-elevate without the .hab directory present, it can get created with root owner which is not what we want.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -44,6 +44,17 @@ const SVC_PATH: &'static str = "hab/svc";
 lazy_static! {
     static ref EUID: u32 = users::get_effective_uid();
 
+    static ref MY_CACHE_PATH: PathBuf = {
+        if *EUID == 0u32 {
+            PathBuf::from(ROOT_PATH)
+        } else {
+            match env::home_dir() {
+                Some(home) => home.join(format!(".{}", ROOT_PATH)),
+                None => PathBuf::from(ROOT_PATH),
+            }
+        }
+    };
+
     static ref MY_CACHE_ANALYTICS_PATH: PathBuf = {
         if *EUID == 0u32 {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -44,6 +44,17 @@ const SVC_PATH: &'static str = "hab/svc";
 lazy_static! {
     static ref EUID: u32 = users::get_effective_uid();
 
+    static ref MY_CACHE_PATH: PathBuf = {
```
smurawski: Is there a reason we are adding this rather than using the existing `MY_CACHE_KEY_PATH`?

components/hab/src/command/studio.rs
```diff
@@ -111,6 +111,15 @@ mod inner {
             return Ok(());
         }
 
+        // Make sure we have run 'hab setup' already (i.e, .hab folder exists)
+        let cache_dir = cache_path(None);
```
smurawski: Can we do this check somewhere that both Linux and Mac/Windows studio invocations could benefit?  Since we attempt to map the key cache location from the host to the docker-based studio, checking for `hab setup` would be good feedback for new users.

components/core/src/fs.rs
chefsalim: Yeah - if the user elect's to not create an origin key we won't have a key file/path.  So we needed a different check, and I elected to check the root path existence (could also have used the analytics path since the analytics opt in/out files are always created, but this seemed cleaner long term).

Issue #1082, opened 2016-07-19
Need a better installation/bootstrap method for hab binary itself and mixlib-install integration


lamont-granquist:
AFAIK right now there's two ways to bootstrap getting hab onto a box:
- The downloads page (which is awful UX for automation)
- Doing a curl | bash of `https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh` which only works on linux and that URL looks hella brittle.

I don't much care about the backend of where the bits come from, but it should probably be made installable from mixlib-install.

smith:
It's hosted on Bintray (https://bintray.com/habitat/stable/hab-x86_64-linux/0.8.0-20160708161531#files/linux/x86_64, for example), so whatever we use with Bintray for Chef should work for Habitat too.

lamont-granquist:
We currently use omnitruck.  If that's the way forward then we need channels setup in omnitruck so that we can curl bash install.sh there and install it from bintray.  Then the work needs to get done so that mixlib-install could pull it from omnitruck.

Paging @schisamo

juliandunn:
I'm not sure this is an issue that is in Habitat's domain to fix. As Nathan mentions, the packages are available from Bintray-format links, which could be automated however you want.

lamont-granquist:
Until it gets identified who should fix it and it gets passed off, its still a UX issue encountered while trying to use habitat.

jtimberman:
I'm actively working on this and plan to have a PR up soon.

reset:
@lamont-granquist the changes in #1364 make this process better, but it doesn't add mixlib-install integration. It's unlikely that somebody from the Habitat team will be adding that in the near future, but we can give you anything you need to get it done!

igalic:
is there a way to reproduce whatever is on bintray myself?
I understand that a lot of this centered around docker, but… uhm… `make install` would still be nice.

adamhjk:
@igalic that's a slightly different issue. Adding a `make install` development target seems like a fine addition to the `Makefile`. 

I do think we need to provide a less brittle install script, but I wouldn't put it high on the list of things to do this instant.

reset:
For the short term, we should document this on the installation page and make these instructions prominent at the top of the README

christophermaier:
Closing due to age and the fact that it was likely fixed a long time ago.

PR #1364, opened 2016-10-14
[JEX-346] Unify the Linux and MacOS install scripts

jtimberman:
We borrow some idioms and helpers from the omnitruck install.sh script
to make a better cross-platform install.sh script for Linux and MacOS.
This allows us to have a single installation script that handles both
platforms, and allows for easier extension to other platforms in the
future.

As the "mac" install.sh will no longer be necessary, it is removed
from the repository.

Signed-off-by: Joshua Timberman <joshua@chef.io>


thesentinels:

Thank you for contributing!

components/hab/install.sh
```diff
@@ -1,58 +1,318 @@
...
+}
+
+checksum_mismatch() {
+  echo "Package checksum mismatch!"
+  report_bug
+  exit 1
+}
+
+unable_to_retrieve_package() {
+  echo "Unable to retrieve a valid package!"
```
reset: We should include the reason we failed here and not just the outcome of the install. It looks like we've failed because we were unable to locate a suitable means to retrieve a remote asset over HTTP. We should also include some instructions like, "Hey dummy! Install wget or something!", but... maybe more friendly ;)

components/hab/install.sh
```diff
@@ -1,58 +1,318 @@
...
+
+  url=`echo $1`
+  if test "x$platform" = "xsolaris2"; then
+    if test "x$platform_version" = "x5.9" -o "x$platform_version" = "x5.10"; then
+      # solaris 9 lacks openssl, solaris 10 lacks recent enough credentials - your base O/S is completely insecure, please upgrade
+      url=`echo $url | sed -e 's/https/http/'`
+    fi
+  fi
+
+  # we try all of these until we get success.
```
reset: I love how we try a bunch of different ways to get the result we want

reset:
Closes #1082

jtimberman:
@reset Updated!

lamont-granquist:
This is an improvement, but I'm not certain it really closes #1082 

You're basically copypasta'ing install.sh into here which is kinda what mixlib-install was designed to avoid.  There's going to be additional issues which crop up that you'll rediscover and eventually we'll wind up having requirements to add it to mixlib-install.

jtimberman:
I think it closes the first half of #1082, but not the second. That is, it's a better installation/bootstrap method for the `hab` binary on Linux and MacOS. It certainly doesn't integrate with mixlib/install. That is work that will come later, by Engineering Services. 

What I've copy/pasted from the omnitruck install.sh is the helpers, primarily to detect which program to use for downloading, comparing the sha256 checksums, and the related functions for error handling. Habitat's install.sh will diverge from Chef's anyway because Habitat is not concerned with multiple different platform and platform-version specific package managers.

PR #1368, opened 2016-10-17
hab pkg export deb

smith:
This is in an experimental state, but will allow you to do

`hab pkg export deb myorigin/mypackage`

and spit out a .deb that you can install on a debian system.

Discuss.

Known issues
------------

* If you have a package with the same a name, but a newer version, of an
  existing package, it will happily overwrite that, even though one is a
  hab package and the other is a system-level thing, so if you `dpkg -i`
  the package you built with `hab pkg export deb core/findutils`,
  bye-bye `find` command. It will be removed and replaced with the one
  hidden away in /hab
* If you have a description, license, maintainer, or upstream URL with a
  ":" it might cut it off because of the poor method I'm using to extract
  this data from the manifest
* Invalid package names don't get safely converted correctly because of
  my wrong regexes

Features not implemented
------------------------

* Doesn't do `preinst`, `postinst`, `prerm`, or `postrm`
* No support for `conffiles`
* Multi-line descriptions
* Doesn't do depends/conflicts/replaces
* Priority and Section are not customizable

Open Questions
--------------

* Do we install everything into /hab? This is what we do now, but if
  there's packages already there you need to do `--force-overwrite`, which
  is probably fine, since things are immutable-ish, but not sure if that's
  what we should be doing .
* Do we put the origin in the package filename or the package name
  somewere?
* There's a "Vendor" field in debs, but hab doesn't have anything that
  really corresponds (there's already "Maintainer".) I'm just sticking
  the origin in there for now. Should we do something else?

gif-keyboard-10948128314208408466

### DIFF
--- a/components/hab/src/command/pkg.rs
+++ b/components/hab/src/command/pkg.rs
@@ -184,17 +184,24 @@ pub mod export {
 
         pub fn format_for(_ui: &mut UI, value: &str) -> Result<ExportFormat> {
             match value {
-                "docker" => {
-                    let format = ExportFormat {
-                        pkg_ident: try!(PackageIdent::from_str("core/hab-pkg-dockerize")),
-                        cmd: "hab-pkg-dockerize".to_string(),
-                    };
-                    Ok(format)
-                }
                 "aci" => {
                     let format = ExportFormat {
                         pkg_ident: try!(PackageIdent::from_str("core/hab-pkg-aci")),
                         cmd: "hab-pkg-aci".to_string(),
+                    };
+                    Ok(format)
+                }
+                "deb" => {
+                    let format = ExportFormat {
+                        pkg_ident: try!(PackageIdent::from_str("core/hab-pkg-deb")),
+                        cmd: "hab-pkg-deb".to_string(),
+                    };
+                    Ok(format)
+                }
+                "docker" => {
+                    let format = ExportFormat {
+                        pkg_ident: try!(PackageIdent::from_str("core/hab-pkg-dockerize")),
+                        cmd: "hab-pkg-dockerize".to_string(),
                     };
                     Ok(format)
                 }
--- a/components/pkg-deb/bin/hab-pkg-deb.sh
+++ b/components/pkg-deb/bin/hab-pkg-deb.sh
@@ -1,0 +1,305 @@
+#!/bin/bash
+#
+# # Usage
+#
+# ```
+# $ hab-pkg-deb [PKG ...]
+# ```
+#
+# # Synopsis
+#
+# Debian package exporter for Habitat artifacts
+#
+# # License and Copyright
+#
+# ```
+# Copyright: Copyright (c) 2016 Chef Software, Inc.
+# License: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ```
+
+# Everything in this script was shamelessy and gratefully copied from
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

smith:
> I know this is experimental as you pointed out, but I think that you need to implement depends before this could begin to be useful. That is, unless you're the consumer of this and you don't need it!

The use case is basically replacing Omnibus. Since an exported hab deb would include all of its dependencies, it's unlikely that it would need depends at all in normal usage. None of the Omnibus packages that Chef ships have any dependencies that I know of.

components/pkg-deb/plan.sh
```diff
@@ -0,0 +1,43 @@
+pkg_name=hab-pkg-deb
+pkg_origin=core
+pkg_version=$(cat "$PLAN_CONTEXT/../../VERSION")
+pkg_description="Export a Debian package from Habitat."
```
smith: Changed this and force-pushed.

components/hab/src/command/pkg.rs
```diff
@@ -184,17 +184,24 @@ pub mod export {
...
                     let format = ExportFormat {
-                        pkg_ident: try!(PackageIdent::from_str("core/hab-pkg-dockerize")),
-                        cmd: "hab-pkg-dockerize".to_string(),
+                        pkg_ident: try!(PackageIdent::from_str("core/hab-pkg-aci")),
+                        cmd: "hab-pkg-aci".to_string(),
                     };
                     Ok(format)
                 }
-                "aci" => {
+                "deb" => {
```
ryankeairns: I'm not very familiar with `deb` packages, other than the limited understanding I've gained recently while researching package managers. That said, I agree that we should have a discussion about _when/why_ to build upon the package export feature since, as you stated above, we anticipate the need for even more image formats.

smurawski:
To respond to the question:

> Do we install everything into /hab? This is what we do now, but if there's packages already there you need to do --force-overwrite, which is probably fine, since things are immutable-ish, but not sure if that's what we should be doing 

I think requiring `--force-overwrite` is ok given builds would install to a version and timestamp qualified path, installing over the top would seem to be an edge case or exceptional circumstances.

components/hab/src/command/pkg.rs
mwrock: Agreed and also to provide an omnibus equivilent to other platforms as well.

components/pkg-deb/bin/hab-pkg-deb.sh
```diff
@@ -0,0 +1,305 @@
...
+# Build the debian package
+build_deb() {
+  # Install packages using the studio
+  env FS_ROOT="$staging_dir" hab pkg install "$PKG"
+  # Delete files we won't be needing
+  rm -rf "${staging_dir:?}/hab/cache"
+  # Make a DEBIAN directory
+  mkdir -p "$staging_dir/DEBIAN"
+
+	# Set these variables in advance, since they may or may not be in the manifest,
```
fnichol: Any reason this block is further indented?

smith: No

mwrock:
I wonder if we are thinking about this from the wrong direction. At least I think I might be. When I think of Omnibus feature parity and packages, I was thinking of pumping out debs, msis, etc that look basically the same as those same artifacts do today in omnibus with no habitat specifics inside.

If we think of producing artifacts for a true habitat application, do those packaging formats make sense to pursue? Why woukld you prefer to install a habitat app via a deb or msi into your habitat environment vs a hart?

adamhjk:
I'm not sure this is at all the right way to go.

1) If you don't create a 1:1 relationship between a hab package and a deb, you'll wind up with a sea of potential conflicts (not the least of which will be on hab itself.) You could literally never install two of these packages at the same time.

2) If you do create a 1:1 relationship, how is this better than simply building a deb for `hab` itself, then making a stub-package that simply depends on that and calls `hab pkg install` for you? If you wanted to be offline, you could even packge the hart files in the deb, if you really, really wanted to. (But.. ew)

3) You'll always wind up in trouble around pre/post install.

reset:
I spoke with @smith over slack last week about this feature and suggested that we don't include package exporters in Habitat. If the Omnibus team wants to move forward with translating `.hart`s into other formats they can create a separate repo with this software and package it up as a `.hart` just like in this PR and get the same results.

I'm going to close this PR out. If there's something we can do to unblock any of the Omnibus team's work to make this happen outside of the Habitat codebase feel free to reach out!

PR #1370, opened 2016-10-17
Appveyor building `hab` and testing ported crates

smurawski:
I moved the execution of the build and test cycles into a powershell script (`./support/ci/appveyor.ps1`) which is called by (`./support/ci/appveyor.bat`).  

We use a batch file to start things off because AppVeyor's PowerShell host treats stderr from native executables (like `cargo`) as errors on the error stream and adds all sorts of ugly output.

Calling PowerShell.exe from a batch file, while an odd hack, works to keep the output looking good but have the better scripting experience of PowerShell.

Now, the build will check if any files have changed in the ported crates and, if so, it'll run cargo test against all ported crates and then build `hab.exe` and run `hab --version` to validate `hab.exe` can run successfully.

Builds can be forced by setting a value to $env:HAB_FORCE_BUILD and tests can be forced with $env:HAB_FORCE_TEST

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,7 @@ image: Visual Studio 2015
 
 cache:
     - c:\Program Files (x86)\Rust
+    - c:\Users\appveyor\.cargo
 
 branches:
   only:
@@ -15,6 +16,11 @@ skip_tags: true
 
 clone_folder: c:\projects\habitat
 
+environment:
+  matrix:
+    - hab_build_action: "test;build"
+      hab_components: "core;http-client;builder-protocol;builder-depot-client;common;hab"
+
 install:
   - ps: ./components/hab/win/win-build.ps1
   - rustc -V
@@ -23,5 +29,4 @@ install:
 build: false
 
 test_script:
-  - cd c:\projects\habitat\components\hab
-  - cargo build --verbose
+  - c:\projects\habitat\support\ci\appveyor.bat
--- a/support/ci/appveyor.bat
+++ b/support/ci/appveyor.bat
@@ -1,0 +1,1 @@
+powershell -nologo -noprofile -executionpolicy bypass -file c:\projects\habitat\support\ci\appveyor.ps1 
--- a/support/ci/appveyor.ps1
+++ b/support/ci/appveyor.ps1
@@ -1,0 +1,30 @@
+pushd c:\projects\habitat
+
+$RunTests = (git diff master --name-only | 
+                where-object {$_ -like 'components/*'} | 
+                foreach-object { $_ -replace 'components/(\w+-*\w*)/.*$', '$1'} | 
+                sort-object -unique | 
+                where-object {($env:HAB_COMPONENTS -split ';') -contains $_}
+            ).count -ge 1
+
+foreach ($BuildAction in ($env:hab_build_action -split ';')) {
+    if (($RunTests -or (test-path env:HAB_FORCE_BUILD)) -and ($BuildAction -like 'build')) {
+        pushd "c:/projects/habitat/components/hab"
+        cargo build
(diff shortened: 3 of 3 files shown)


thesentinels:

Thank you for contributing!

appveyor.yml
```diff
@@ -23,5 +30,4 @@ install:
 build: false
 
 test_script:
-  - cd c:\projects\habitat\components\hab
-  - cargo build --verbose
+  - appveyor.bat
```
mwrock: Would be great to see this in powershell because `%%A` makes me reactively stab my eyes :)

smurawski: yeah, the problem is the output from cargo via powershell makes me want to stab my eyes.

mwrock: haha. ok fair enough!

appveyor.yml
```diff
@@ -15,6 +16,13 @@ skip_tags: true
...
 
+environment:
+  matrix:
+    - hab_build_action: build
+      hab_components: hab
+    - hab_build_action: test
+      hab_components: "core;http-client;builder-protocol;builder-depot-client;common"
+
 install:
   - ps: ./components/hab/win/win-build.ps1
```
mwrock: Not a bad idea to add a `& /path/to/hab.exe --version`

appveyor.yml
```diff
@@ -15,6 +16,13 @@ skip_tags: true
 
 clone_folder: c:\projects\habitat
 
+environment:
+  matrix:
+    - hab_build_action: build
+      hab_components: hab
```
smurawski: Correct (though I should add hab to the collection of for the test run).

appveyor.yml
smurawski: After the build?  It'll only exist after the cargo build, not in the test cycle.

appveyor.yml
reset: Yeah I'd add `hab` to the list of components to be tested. Do you think there's much value in a matrix just ensuring the component builds? The cargo test task will build the software in addition to running the test suite. It's true that we could be using `cfg(test)` for conditional compilation outside of just marking modules as test modules, but we don't really do that.

Whatever you feel is most appropriate just make sure we run the test suite on all components ;)

smurawski:
@mwrock @reset I changed the direction a bit. Matt, your eyes will bleed less as I've moved to a PowerShell script for the conditional logic.  Jamie, we are building hab.exe and running tests against all the crates.  

There is no benefit at this time (and, in fact, an impediment thanks to cargo workspaces) in splitting the builds onto multiple machines.

appveyor.yml
```diff
@@ -15,6 +16,11 @@ skip_tags: true
 
 clone_folder: c:\projects\habitat
 
+environment:
+  matrix:
+    - hab_build_action: "test;build"
```
mwrock: Any reason why not to break out the matrix like:

```
- hab_build_action: "build"
- hab_components: "hab"

- hab_build_action: "test"
- hab_components: "hab"

- hab_build_action: "test"
- hab_components: "http-client"

etc, etc
```

Something like this would allow the appveyor.ps1 to be much simpler (no looping or string parsing) and the test actions to be more clear.

smurawski: yeah, it's WAAAY slower.  
* sharing a workspace for each build step allows single pass compilation of dependencies
* if we don't want the build matrix to progress serially, there's a dollar cost ($100/mo just to get two concurrent jobs).

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/168670790).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/168670790).

PR #1388, opened 2016-10-22
add a `hab pkg header` command to read the hart file header

smurawski:
I had a few minutes this morning and thought this might be a handy subcommand.

/cc @juliandunn 

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -209,6 +209,13 @@ pub fn get() -> App<'static, 'static> {
                     "A path to a Habitat Artifact \
                     (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
             )
+            (@subcommand header =>
+                (about: "Returns the Habitat Artifact header")
+                (aliases: &["hea", "head", "heade", "header"])
+                (@arg SOURCE: +required {file_exists}
+                    "A path to a Habitat Artifact \
+                    (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
+            )
         )
         (@subcommand ring =>
             (about: "Commands relating to Habitat rings")
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -212,6 +212,7 @@ pub fn get() -> App<'static, 'static> {
             (@subcommand header =>
                 (about: "Returns the Habitat Artifact header")
                 (aliases: &["hea", "head", "heade", "header"])
+                (@setting Hidden)
                 (@arg SOURCE: +required {file_exists}
                     "A path to a Habitat Artifact \
                     (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
--- a/components/hab/src/command/pkg.rs
+++ b/components/hab/src/command/pkg.rs
@@ -571,3 +571,33 @@ pub mod verify {
         Ok(())
     }
 }
+
+pub mod header {
+    use std::path::Path;
+
+    use common::ui::UI;
+    use hcore::crypto::artifact;
+    use std::io::{self, Write};
+
+    use error::Result;
+
+    pub fn start(ui: &mut UI, src: &Path) -> Result<()> {
+        try!(ui.begin(format!("Reading Artifact Header for {}", &src.display())));
+        try!(ui.para(""));
+        let artifact_header = artifact::get_artifact_header(src);
+        if artifact_header.is_ok() {
+            let header = artifact_header.unwrap();
+            try!(io::stdout()
+                .write(format!("Format Version : {}\n", header.format_version).as_bytes()));
+            try!(io::stdout().write(format!("Key Name       : {}\n", header.key_name).as_bytes()));
+            try!(io::stdout().write(format!("Hash Type      : {}\n", header.hash_type).as_bytes()));
+            try!(io::stdout()
+                .write(format!("Raw Signature  : {}\n", header.signature_raw).as_bytes()));
+        } else {
+            try!(ui.warn("Failed to Read Artifact Header."));
+        }
+        try!(ui.para(""));
+        try!(ui.end(format!("Finished Reading Artifact Header for {}.", &src.display())));
+        Ok(())
+    }
+}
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -209,6 +209,13 @@ pub fn get() -> App<'static, 'static> {
                     "A path to a Habitat Artifact \
                     (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
             )
+            (@subcommand header =>
+                (about: "Returns the Habitat Artifact header")
+                (aliases: &["hea", "head", "heade", "header"])
```
smurawski: This command has fewer alias because of some lettering overlap with help.

juliandunn:
Thanks! This could use a test I think - I imagine we _might_ need to change the header when we get to Windows, and I'm sure we'll break this if so.

smurawski:
@juliandunn since I'm using our artifact api and not manually parsing the header, I'm using tested code paths which will need to be updated if we change the package header.  The subcommand code is mostly output using our ui library functions, so there isn't much to independently test.

smurawski:
hab_pkg_header

adamhjk:
I would make the output here more parse-able - it's pretty this way, but it'll not be very useful if you turn the colors off. Should it be colon-separated? @fnichol?

smurawski:
@adamhjk how about a switch to output JSON?

adamhjk:
@smurawski your powershell is showing. :) a good example of how you might use this, to say make a choice based on the type of hart you have:

```bash
$ hab pkg header foo.hart | grep "Format Version" | cut -d ":" -f 1
```

Would print the version if the output had "Format Version: HART-1"

fnichol:
We've been favoring the simplest structured output as we can get away with, primarily so that in minimal environments a user wouldn't need extra tools or languages (in other words, can we pass the "do you have cat-and-cut?" test). In that spirit the colon-delimited version looks to work decently. However, I can't think of any subcommands that (as of yet) have structured output meant to be consumed by other scripts, aside from `hab pkg path`.

With a product hat on, I'd be interested in what value we're providing with this command? Up to this point the structure of the package metadata/frontmatter has been internal implementation details that we could conceivably change in a forwards compatible way. Don't get me wrong, there is most likely value to this--I'm just coming up with few ideas this morning.

smurawski:
The idea for the subcommand @fnichol was primarily for habitat developers as they build and validate packages (rather than having to remember `head -n5` or some such to return the package format (especially since on Windows that command would be different and, depending on your version of powershell, could be pretty ugly).

ryankeairns:
@smurawski , et al, I'll post a quick mockup of slightly revised layout for the output.

reset:
@smurawski @fnichol @ryankeairns I am inclined to agree with @fnichol here, the header is an implementation detail which should be invisible to the end-user. What if we made this a hidden CLI command?

ryankeairns:
@reset that makes sense to me... you're suggesting it basically doesn't show up in the `help` list of commands, right?

ryankeairns:
As for the output, I'd propose a format like this:

hab pkg header

reset:
@ryankeairns that's right, it's still there but it's not showing up in the help

smurawski:
@ryankeairns @reset @fnichol 
Closer?
hab_pkg_header_hidden

reset:
@smurawski I like it!

smurawski:
@reset  I don't have the nice green from @ryankeairns, but I think I can sort that out.

ryankeairns:
@smurwaski looks great, thanks for making the changes.

components/hab/src/command/pkg.rs
```diff
@@ -571,3 +571,33 @@ pub mod verify {
...
+    use std::path::Path;
+
+    use common::ui::UI;
+    use hcore::crypto::artifact;
+    use std::io::{self, Write};
+
+    use error::Result;
+
+    pub fn start(ui: &mut UI, src: &Path) -> Result<()> {
+        try!(ui.begin(format!("Reading Artifact Header for {}", &src.display())));
```
juliandunn: I would recommend not capitalizing all this, and also, we have in the past decided not to refer to `.hart`s as artifacts, just as packages. So altogether:

`Reading package header for {}`

components/hab/src/command/pkg.rs
```diff
@@ -571,3 +571,33 @@ pub mod verify {
...
+        if artifact_header.is_ok() {
+            let header = artifact_header.unwrap();
+            try!(io::stdout()
+                .write(format!("Format Version : {}\n", header.format_version).as_bytes()));
+            try!(io::stdout().write(format!("Key Name       : {}\n", header.key_name).as_bytes()));
+            try!(io::stdout().write(format!("Hash Type      : {}\n", header.hash_type).as_bytes()));
+            try!(io::stdout()
+                .write(format!("Raw Signature  : {}\n", header.signature_raw).as_bytes()));
+        } else {
+            try!(ui.warn("Failed to Read Artifact Header."));
```
juliandunn: Ditto here:

`Failed to read package header`

components/hab/src/command/pkg.rs
```diff
@@ -571,3 +571,33 @@ pub mod verify {
...
+        }
+        try!(ui.para(""));
+        try!(ui.end(format!("Finished Reading Artifact Header for {}.", &src.display())));
```
juliandunn: I'd suggest just eliminating this message.

components/hab/src/command/pkg.rs
smurawski: `Artifact` is the term we use elsewhere in the cli help, so we should probably clean that up as well if we just want to use package. (I've got no problem with that, I was just trying to stay consistent with the rest of the cli.

smurawski:
@juliandunn - updated messaging per your comments
image

juliandunn:
LGTM, let's get a 👍 from one of the maintainers and perhaps we can :shipit: in tomorrow's release!

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/170925828).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/170925828).

PR #1443, opened 2016-11-02
pull the supervisor waitpid into core to be handled separately by os

mwrock:
This is one step forward in allowing the supervisor to build on windows. The windows implementation for waiting on a process remains unimplemented and that will be added in a separate PR likely leveraging win api calls to `OpenProcess` and `WaitForSingleObject`.

### DIFF
--- a/components/core/src/os/process/linux.rs
+++ b/components/core/src/os/process/linux.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use libc::{pid_t, c_int};
 use std::ffi::OsString;
 use std::path::PathBuf;
 use std::os::unix::process::CommandExt;
@@ -19,8 +20,16 @@ use std::process::Command;
 
 use error::Result;
 
+extern "C" {
+    fn waitpid(pid: pid_t, status: *mut c_int, options: c_int) -> pid_t;
+}
+
 pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
     become_exec_command(command, args)
+}
+
+pub fn wait_for_exit(pid: u32, status: *mut c_int) -> u32 {
+    unsafe { waitpid(pid as i32, status, 1 as c_int) as u32 }
 }
 
 /// Makes an `execvp(3)` system call to become a new program.
--- a/components/sup/src/supervisor.rs
+++ b/components/sup/src/supervisor.rs
@@ -27,8 +27,9 @@ use std::process::{Command, Stdio, Child};
 use std::thread;
 
 use hcore;
+use hcore::os::process;
 use hcore::package::PackageIdent;
-use libc::{pid_t, c_int};
+use libc::c_int;
 use time::{Duration, SteadyTime};
 
 use error::{Result, Error};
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

raskchanky:
This looks like it's mostly just shuffling bits around?  Seems fine to me.  I wonder if there's value in re-evaluating how some of the process shenanigans are accomplished and leveraging something like Rust's `std::process::Command`.  I'm guessing that's likely out of scope for this work.

gif-keyboard-1414087175926888008

mwrock:
@raskchanky I'd definitely prefer to use std lib code. Its highly probably there could be something out there to do what we need but from the looks of `std::process::command` it can either launch a child process and continue or launch and wait but not launch and continue and allow you to wait later.

mwrock:
oh nevermind @raskchanky after further investigation it looks like you can take the `child` returned from `spawn` and then call `wait` on that.

I'm gonna look at that more closely and may just close that if it looks like it pans out.

smurawski:
Yeah, I'd play with the command functionality.. that pretty much wraps createprocess and makes it a safe call.

mwrock:
Yeah looking more closely at the supervisor code I think using `command` and `child` is the way to go when we move to implement windows, but that means the supervisoe will hold on to a `child` and not a `pid` which means lots of changes here and I'd prefer to do that as part of a separate PR.

raskchanky:
@mwrock Yeah, makes sense.  This one looks good as is once CI passes.

components/core/src/os/process/linux.rs
```diff
@@ -12,17 +12,26 @@
...
+extern "C" {
+    fn waitpid(pid: pid_t, status: *mut c_int, options: c_int) -> pid_t;
+}
+
 pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
     become_exec_command(command, args)
 }
 
+pub fn wait_for_exit(pid: u32, status: *mut c_int) -> u32 {
+    return waitpid(pid, &mut status, 1 as c_int).unwrap as u32;
```
smurawski: If you leave off the `;`, you won't need the `return` as rust will return the result of the last expression.

Also, this should be wrapped in `unsafe {}` (rather than the later code in the supervisor.

mwrock:
heh. and it turns out that when you JUST build it on windows locally, any issues in the linux class is blissfully ignored.

components/sup/src/supervisor.rs
```diff
@@ -313,8 +309,8 @@ impl Supervisor {
 
         unsafe {
```
smurawski: You probably should be able to remove this `unsafe` block in favor of the more scoped block in core.

mwrock:
OK I made the suggested changes and got things building/passing on linux. Just pushed a merged and signed off commit.

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/172947378).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/172947378).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/172947378).

PR #1447, opened 2016-11-03
Move user functions from sup to core

mwrock:
This moves the usage of the users crate out of the supervisoe and into core. Most functions were already implemented but I wrapped a couple that were not already there. We will add windows implementations in a separate PR.

### DIFF
--- a/components/core/src/os/users/mod.rs
+++ b/components/core/src/os/users/mod.rs
@@ -16,10 +16,10 @@
 mod windows;
 
 #[cfg(windows)]
-pub use self::windows::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user};
+pub use self::windows::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user, get_current_username, get_current_groupname};
 
 #[cfg(not(windows))]
 pub mod linux;
 
 #[cfg(not(windows))]
-pub use self::linux::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user};
+pub use self::linux::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user, get_current_username, get_current_groupname};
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

components/core/src/os/users/mod.rs
```diff
@@ -16,10 +16,10 @@
 mod windows;
 
 #[cfg(windows)]
-pub use self::windows::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user};
+pub use self::windows::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user, get_current_username, get_current_groupname};
```
mwrock: Is there a better way to do this rather than specifying every function? Especially if you are exposing all public functions?

smurawski: We can use a `*` to export all the functions, structs, etc in a module.  In this case, we'd want to bury `GetUserTokenStatus` because a wildcard would export that unsafe function which is really more of an internal implementation detail.  (I think we can just make that function not public and we should be alright.) 

Rust style/patterns tend to lean toward explicit imports and exports so you know what you are getting from where.

There's a good thread on the Rust forums on this topic as well.. 

https://users.rust-lang.org/t/confused-on-mod-use-pub-simple-example/1973/6?u=smurawski

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/172958623).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/172958623).

components/core/src/os/users/mod.rs
smurawski: @mwrock we can move to an `*` on the import and move the `GetUserTokenStatus` to a non-public function.  It's not a big deal, it does compile and doesn't seem to break anything.  I'm not opposed to doing the `pub use` that way since we control that crate.

PR #1450, opened 2016-11-03
Move process signalling to core

mwrock:
Habitat was failing to link because of the C call to `Kill`. This lifts the signalling out of sup and into core where is can be handled per os. Will need to backfill windows in the near future.

### DIFF
--- a/components/core/src/os/process/linux.rs
+++ b/components/core/src/os/process/linux.rs
@@ -21,6 +21,7 @@ use std::process::Command;
 use error::Result;
 
 extern "C" {
+    fn kill(pid: i32, sig: u32) -> u32;
     fn waitpid(pid: pid_t, status: *mut c_int, options: c_int) -> pid_t;
 }
 
@@ -30,6 +31,13 @@ pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
 
 pub fn wait_for_exit(pid: u32, status: *mut c_int) -> u32 {
     unsafe { waitpid(pid as i32, status, 1 as c_int) as u32 }
+}
+
+/// send a Unix signal to a pid
+pub fn send_signal(pid: u32, sig: u32) -> u32 {
+    unsafe {
+        kill(pid as i32, sig)
+    }
 }
 
 /// Makes an `execvp(3)` system call to become a new program.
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

adamhjk:
Are we sure this belongs in Core? Does any other part of Habitat need to call it?

smurawski:
@adamhjk the idea behind moving it to core was as a central place to resolve our platform dependent behaviors.  It's a fair point for discussion, but that seemed logical as everything depends on core.

adamhjk:
We shouldn't be compiling and linking to code that we don't need, if we can help it. So unless at least two components are going to call it, we shouldn't put it in core, would be my feelings.

smurawski:
Rust, on linux, links to libc, by default (as far as I'm aware) so there aren't additional system libraries being linked for this function.

smurawski:
That call is also in director, but director is not long for this world.

mwrock:
I don't think this adds any additional linking. It fixes this error:

```
error LNK2019: unresolved external symbol kill referenced in function _ZN11habitat_sup4util7signals18send_signal_to_pid17h98cd7e679a126a7cE
```

So on linux you would link to glibc either way but the msvc library is missing the kill signal.

The supervisor is the only place where this function is being called. However I think putting low level OS specific calls in core is a solid pattern. These should definitely be as limited and as low level as possible. Ideally there should be very little inside of these os specific modules.

mwrock:
that all said I think I'm gonna further refactor this so that the function stays in signals.rs and that calls to core rather than supervisor.rs calling to core.

mwrock:
ok I have rearranged this a bit. Mow the only thing in core (for linux) is the one line call to libc's `Kill` so supervisor.rs remains untouched calling into signals.rs which really does "own" the signal logic and signal.rs calls into core.

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/173289565).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/173289565).

PR #1458, opened 2016-11-07
add target awareness to PackageArchive and PackageInstall

smurawski:
Adds validation of the TARGET (platform and architecture) from the package to the host running/installing the package.

This impacts both the `hab` and `hab-sup` executables and prevents the installation or execution of packages whose target does not match the host OS.

### DIFF
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -55,9 +55,9 @@ pub enum Error {
     /// Occurs when a package target string cannot be successfully parsed.
     InvalidPackageTarget(String),
     /// Occurs when validating a package target for an unsupported architecture.
-    InvalidPackageTargetArchitecture(String),
+    InvalidArchitecture(String),
     /// Occurs when validating a package target for an unsupported platform.
-    InvalidPackageTargetPlatform(String),
+    InvalidPlatform(String),
     /// Occurs when a service group string cannot be successfully parsed.
     InvalidServiceGroup(String),
     /// Occurs when making lower level IO calls.
@@ -126,16 +126,12 @@ impl fmt::Display for Error {
                         e)
             }
             Error::InvalidPackageTarget(ref e) => {
-                format!("Invalid package target: {:?}. A valid target is in the form \
+                format!("Invalid package target: {}. A valid target is in the form \
                          architecture-platform (example: x86_64-linux)",
                         e)
             }
-            Error::InvalidPackageTargetArchitecture(ref e) => {
-                format!("Invalid package target architecture: {:?}.", e)
-            }
-            Error::InvalidPackageTargetPlatform(ref e) => {
-                format!("Invalid package target platform: {:?}.", e)
-            }
+            Error::InvalidArchitecture(ref e) => format!("Invalid architecture: {}.", e),
+            Error::InvalidPlatform(ref e) => format!("Invalid platform: {}.", e),
             Error::InvalidServiceGroup(ref e) => {
                 format!("Invalid service group: {:?}. A valid service group string is in the form \
                          service.group (example: redis.production)",
@@ -195,10 +191,8 @@ impl error::Error for Error {
             Error::InvalidPackageTarget(_) => {
                 "Package targets must be in architecture-platform format (example: x86_64-linux)"
             }
-            Error::InvalidPackageTargetArchitecture(_) => {
-                "Unsupported target architecture supplied."
-            }
-            Error::InvalidPackageTargetPlatform(_) => "Unsupported target platform supplied.",
+            Error::InvalidArchitecture(_) => "Unsupported target architecture supplied.",
+            Error::InvalidPlatform(_) => "Unsupported target platform supplied.",
             Error::InvalidServiceGroup(_) => {
                 "Service group strings must be in service.group format (example: redis.production)"
             }
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -52,6 +52,12 @@ pub enum Error {
     FileNotFound(String),
     /// Occurs when a package identifier string cannot be successfully parsed.
     InvalidPackageIdent(String),
+    /// Occurs when a package target string cannot be successfully parsed.
+    InvalidPackageTarget(String),
+    /// Occurs when validating a package target for an unsupported architecture.
+    InvalidPackageTargetArchitecture(String),
+    /// Occurs when validating a package target for an unsupported platform.
+    InvalidPackageTargetPlatform(String),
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

smurawski:
hab-only-installs-packages-for-target

smurawski:
hab-sup-only-run-packages-for-target

components/core/src/error.rs
```diff
@@ -117,6 +125,17 @@ impl fmt::Display for Error {
                          origin/name (example: acme/redis)",
                         e)
             }
+            Error::InvalidPackageTarget(ref e) => {
+                format!("Invalid package target: {:?}. A valid target is in the form \
```
reset: Just `{}` is fine here, `{:?}` is the "debug" output and not the "display" output of your type

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn platform(&self) -> &str;
+    fn architecture(&self) -> &str;
+    fn valid_for(&self, uname: Uname) -> Result<()>;
+}
+
+/// Describes the platform (operating system/kernel)
+/// and architecture (x86_64, i386, etc..) that a package is built for
+#[derive(RustcEncodable, RustcDecodable, Debug, Clone, Hash)]
+pub struct PackageTarget {
+    pub platform: String,
```
reset: I think it's a bit more appropriate for both the platform and architecture should be represented as enums instead of Strings. It's a bit more performant but more importantly, anytime we can use the type system we should ;)

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub fn new<T: Into<String>>(platform: T, architecture: T) -> Self {
+        let valid_platform = PackageTarget::supported_platform(platform).unwrap();
+        let valid_arch = PackageTarget::supported_architecture(architecture).unwrap();
+        PackageTarget {
+            platform: valid_platform,
+            architecture: valid_arch,
+        }
+    }
+
+    fn supported_platform<T: Into<String>>(value: T) -> Result<String> {
```
reset: If you go the enum route, this function would become an implementation of the `FromStr` trait just like you have below for `PackageTarget` itself

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+        let platform = value.into();
+        let valid_platforms = vec!["linux", "windows",];
+        if valid_platforms.iter().any(|&x| x == platform) {
+            Ok(platform)
+        } else {
+            Err(Error::InvalidPackageTargetPlatform(platform))
+        }
+    }
+
+    fn supported_architecture<T: Into<String>>(value: T) -> Result<String> {
```
reset: Same here, this would become a `FromStr` implementation

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+use std::result;
+use std::str::FromStr;
+use os::system::{uname, Uname};
+
+use error::{Error, Result};
+
+pub trait Target: fmt::Display + Into<PackageTarget> {
+    fn platform(&self) -> &str;
+    fn architecture(&self) -> &str;
+    fn valid_for(&self, uname: Uname) -> Result<()>;
```
reset: I might recommend calling this `validate` instead of `valid_for` if you're returning a Result. The latter sounds like it would return a boolean where the former, to me at least, sounds like an API which returns an `Ok` or an `Err`

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+                                                self.platform,
+                                                architecture,
+                                                platform,)))
+        }
+    }
+}
+
+impl Default for PackageTarget {
+    fn default() -> PackageTarget {
+        let current_system = uname().unwrap();
```
reset: This will cause a panic without message in a default implementation which is a bit of a no-no in the API. Is there a "safer" way to represent a default if we can't get the uname or perhaps we do want to panic but we should provide a bit of a message here.

`expect()` is just like `unwrap()` but it allows you to specify a panic message

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+impl fmt::Display for PackageTarget {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}-{}", self.architecture, self.platform)
+    }
+}
+
+impl AsRef<PackageTarget> for PackageTarget {
```
reset: This is the default implementation of `AsRef` which should be automatically available on any type. I don't think you need to provide it here but I could be wrong!

components/core/src/package/target.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+impl FromStr for PackageTarget {
+    type Err = Error;
+
+    fn from_str(value: &str) -> result::Result<Self, Self::Err> {
+        let items: Vec<&str> = value.split("-").collect();
+        let (architecture, platform) = match items.len() {
+            2 => (items[0], items[1]),
```
reset: In here if you go the route of using enums to represent the Architecture and Platform and have implemented `FromStr` on them, you can use that to chain an error upwards on error in this FromStr implementation

components/core/src/package/target.rs
smurawski: @reset I thought the unwrap would pass along the error that bubbled out of the uname call (which is what I wanted to preserve.

mwrock:
Would also be great to include doc changes here. I assume it would go into the "writing plans" docs.

juliandunn:
bill

components/core/src/package/target.rs
```diff
@@ -0,0 +1,186 @@
...
+                unimplemented!()
+            }
+        }
+    }
+
+    pub fn current_architecture() -> Architecture {
+        if cfg!(target_arch = "x86_64") {
+            Architecture::X86_64
+        } else {
+            unimplemented!()
```
reset: Perfect!

reset:
@thesentinels approve

gif-keyboard-16559914678116768512

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/174311323).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/174311323).

PR #1468, opened 2016-11-09
Adding windows user implementations

mwrock:
This adds a new crate to wrap windows api calls around user calls and is called from core. `cargo test` now passes for the superviso on windows and the supervisor no longer needs a `SVC_GROUP` to run. Also the machine account is given the same "root" privileges we give to `root` on linux.

### DIFF
--- a/components/core/src/os/filesystem/windows.rs
+++ b/components/core/src/os/filesystem/windows.rs
@@ -17,7 +17,7 @@ use std::ffi::CStr;
 use std::path::Path;
 use std::io;
 
-pub fn chown(r_path: *const c_char, uid: u32, gid: u32) -> c_int {
+pub fn chown(r_path: *const c_char, uid: String, gid: String) -> c_int {
     unimplemented!();
 }
 
--- a/components/core/src/os/users/linux.rs
+++ b/components/core/src/os/users/linux.rs
@@ -40,3 +40,7 @@ pub fn get_effective_uid() -> u32 {
 pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
     linux_users::get_user_by_name(username).map(|u| PathBuf::from(u.home_dir()))
 }
+
+pub fn root_level_account() -> String {
+    "root".to_string()
+}
--- a/components/core/src/os/users/windows.rs
+++ b/components/core/src/os/users/windows.rs
@@ -12,26 +12,40 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
 use std::path::PathBuf;
+
+use habitat_win_users::account::Account;
 
 extern "C" {
     pub fn GetUserTokenStatus() -> u32;
 }
 
-pub fn get_uid_by_name(owner: &str) -> Option<u32> {
-    unimplemented!();
+fn get_sid_by_name(name: &str) -> Option<String> {
+  match Account::from_name(name) {
+    Some(acct) => Some(acct.sid.to_string()),
+    None => None
+  }
 }
 
-pub fn get_gid_by_name(group: &str) -> Option<u32> {
-    unimplemented!();
+pub fn get_uid_by_name(owner: &str) -> Option<String> {
+  get_sid_by_name(owner)
+}
+
+pub fn get_gid_by_name(group: &str) -> Option<String> {
+  get_sid_by_name(group)
 }
 
 pub fn get_current_username() -> Option<String> {
-    unimplemented!();
+  match env::var("USERNAME").ok() {
+    Some(username) => Some(username.to_lowercase()),
+    None => None
+  }
 }
 
+// this is a no-op on windows
 pub fn get_current_groupname() -> Option<String> {
-    unimplemented!();
+    Some(String::new())
 }
 
 pub fn get_effective_uid() -> u32 {
@@ -40,4 +54,27 @@ pub fn get_effective_uid() -> u32 {
 
 pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
     unimplemented!();
+}
+
+pub fn root_level_account() -> String {
+    env::var("COMPUTERNAME").unwrap().to_uppercase() + "$"
+}
+
+#[cfg(test)]
+mod tests {
+    use std::env;
+
+    use super::*;
+
+    #[test]
+    fn downcase_current_username() {
+      env::set_var("USERNAME", "uSer");
+      assert_eq!(get_current_username().unwrap(), "user")
+    }
+
+    #[test]
+    fn return_none_when_no_user() {
+      env::remove_var("USERNAME");
+      assert_eq!(get_current_username(), None)
+    }
 }
--- a/components/sup/src/util/users.rs
+++ b/components/sup/src/util/users.rs
@@ -64,10 +64,10 @@ fn check_pkg_user_and_group(pkg_install: &PackageInstall) -> Result<Option<(Stri
             let current_user = current_user.unwrap();
             let current_group = current_group.unwrap();
 
-            if current_user == "root" {
+            if current_user == users::root_level_account() {
                 Ok(Some((user, group)))
             } else {
-                if current_user == user && current_group == group {
+                if current_user == user && (cfg!(target_os = "windows") || current_group == group) {
                     // ok, sup is running as svc_user/svc_group already
                     Ok(Some((user, group)))
                 } else {
@@ -89,7 +89,7 @@ fn get_default_user_and_group() -> Result<(String, String)> {
     let uid = users::get_uid_by_name(DEFAULT_USER);
     let gid = users::get_gid_by_name(DEFAULT_GROUP);
     match (uid, gid) {
-        (Some(uid), Some(gid)) => return Ok((DEFAULT_USER.to_string(), DEFAULT_GROUP.to_string())),
+        (Some(_), Some(_)) => return Ok((DEFAULT_USER.to_string(), DEFAULT_GROUP.to_string())),
         _ => {
             debug!("hab:hab does NOT exist");
             let user = users::get_current_username();
@@ -119,12 +119,4 @@ pub fn get_user_and_group(pkg_install: &PackageInstall) -> Result<(String, Strin
         let defaults = try!(get_default_user_and_group());
         Ok(defaults)
     }
-}
-
-pub fn user_name_to_uid(user: &str) -> Option<u32> {
-    users::get_uid_by_name(user)
-}
-
-pub fn group_name_to_gid(group: &str) -> Option<u32> {
-    users::get_gid_by_name(group)
 }
--- a/components/win-users/Cargo.toml
+++ b/components/win-users/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "habitat_win_users"
+version = "0.0.0"
+authors = ["Matt Wrock <matt@mattwrock.com>"]
+description = "Habitat library for win32 account api calls"
+workspace = "../../"
+
+[target.'cfg(windows)'.dependencies]
+kernel32-sys = "*"
+widestring = "*"
+winapi = "*"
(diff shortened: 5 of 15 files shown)


thesentinels:

Thank you for contributing!

mwrock:
This is building and tests have been added. The following remains:

- [ ] tests are randomly failing on a corrupt pointer
- [ ] refactor sup users to test if a user belongs to the svc_group instead of checking if the "current" group matches since windows has no notion of "current group"

mwrock:
Another thought here: we defer groups  on windows and MVP only allows users. That seems reasonable.

components/core/src/os/filesystem/windows.rs
```diff
@@ -17,7 +17,7 @@ use std::ffi::CStr;
 use std::path::Path;
 use std::io;
 
-pub fn chown(r_path: *const c_char, uid: u32, gid: u32) -> c_int {
+pub fn chown(r_path: *const c_char, uid: String, gid: String) -> c_int {
```
smurawski: Changing this signature is going to be a problematic unless you add platform awareness around https://github.com/habitat-sh/habitat/blob/master/components/core/src/util/perm.rs#L65 

Perhaps, since users are represented so differently, we expose a User struct whose internals vary by platform (one implemented in users/windows and one in users/linux.  Then the chown and chmod signatures could take the users::User struct and each look for what they care about.

components/core/src/os/users/windows.rs
```diff
@@ -12,22 +12,32 @@
...
+  get_sid_by_name(owner)
+}
+
+pub fn get_gid_by_name(group: &str) -> Option<String> {
+  get_sid_by_name(group)
 }
 
 pub fn get_current_username() -> Option<String> {
-    unimplemented!();
+    Some(env::var("USERNAME").unwrap())
```
smurawski: case will likely matter for any equality checks, so maybe force a downcase?

And what behavior would you like if `$env:USERNAME` is empty or not present? (maybe this is a place for a `match` or `unwrap_or_default`.

components/win-users/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "habitat_win_users"
+version = "0.0.0"
+authors = ["Matt Wrock <matt@mattwrock.com>"]
+description = "Habitat library for win32 account api calls"
+build = "../build.rs"
```
smurawski: You only need to specify a build script if you need to compile additional code side by side or control your linking to an external library.  That said, we could pull the C that we have in core over to this crate (as well as the build script), since the checking of UAC/admin status seems to fit feature set for this crate.

components/win-users/src/account.rs
```diff
@@ -0,0 +1,76 @@
...
+
+  let ret = unsafe {
+    LookupAccountNameW(null_mut(), wide.as_ptr(), sid, &mut sid_size as LPDWORD, domain, &mut domain_size as LPDWORD, sid_type)
+  };
+  if ret == 0 {
+    panic!("Failed to retrieve SID for {}: {}", name, Error::last_os_error());
+  }
+
+  let domain_str = unsafe { WideCString::from_ptr_str(domain).to_string_lossy() };
+  Some(Account {name: name, system_name: system_name, domain: domain_str, account_type: unsafe { *sid_type }, sid: Sid {raw: sid}})
```
smurawski: Is the account name you come up with here always going to be consistent with the account name you resolve with users::get_current_username() which pulls from the environment (when looking up the current user)?

components/win-users/tests/account.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+#[test]
+fn bogus_account_returns_none() {
+  assert_eq!(Account::from_name("bogus".to_string()).is_none(), true)
+}
+
+#[test]
+fn user_account_returns_user_type() {
+  let acct_type = Account::from_name("Administrator".to_string()).unwrap().account_type;
```
smurawski: what about a test with all lowercase or all caps?  How does that function respond then?

components/win-users/Cargo.toml
mwrock: yeah I just copied this from another crate that did not have its own build script. I'm tempted to bring over the elevation check but was thinking of converting it to ffi.

components/win-users/src/account.rs
mwrock: like you mentioned above its probably wise to downcase also, its possible for the name to be a SPN or a `domain\user` format. I'm considering having `get_current_username` prepend domain, but that still may not guarantee consistensy. This is definitely something that needs following up on.

components/win-users/Cargo.toml
smurawski: Yeah, I started with FFI, but was fighting it too long and just needed to ship the PR.

components/win-users/tests/account.rs
smurawski: given the cross platform nature of the code base and how much case matter elsewhere, might be worth including in a test or doc test just for clarification/edification of newcomers.

components/win-users/Cargo.toml
mwrock: I totally hear you there. It took a looong time to eek out this bit of code here. I see alot of it in our future so I'm hoping it goes faster.

components/core/src/os/filesystem/windows.rs
mwrock: Just to circle back here, there will definitely be changes to the API here but I want to defer those changes to when we get to the consuming code.

components/win-users/tests/account.rs
mwrock: see new test for the sid struct to demonstrate case agnosticism

components/core/src/os/filesystem/windows.rs
reset: @smurawski is right, the interface for a function which is not marked unsafe should not include raw pointers. I like your approach just fine, though, of just getting things in and moving and coming around and fixing them. Please leave a `MW TODO: <msg>` comment wherever you intend to do that so we can easily find them 😄

components/core/src/os/users/linux.rs
```diff
@@ -40,3 +40,7 @@ pub fn get_effective_uid() -> u32 {
 pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
     linux_users::get_user_by_name(username).map(|u| PathBuf::from(u.home_dir()))
 }
+
+pub fn root_level_account() -> String {
+    "root".to_string()
```
reset: This should just be a static str unless the implementation is going to change in the future to dynamically allocate a string

components/core/src/os/users/windows.rs
```diff
@@ -12,26 +12,40 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
 use std::path::PathBuf;
 
+use habitat_win_users::account::*;
```
reset: It's frowned upon to include an entire module in this way unless it's a module which just re-exports things for you or you're in a test module. This is totally fine for now, but just giving you a heads up!

components/core/src/os/users/windows.rs
```diff
@@ -12,26 +12,40 @@
...
+
 extern "C" {
     pub fn GetUserTokenStatus() -> u32;
 }
 
-pub fn get_uid_by_name(owner: &str) -> Option<u32> {
-    unimplemented!();
+fn get_sid_by_name(name: &str) -> Option<String> {
```
reset: This would be a more appropriate function head:

```
fn get_sid_by_name<T: AsString>(name: T) -> Option<String>
```

Reasoning here is that you're just going to call `to_string()` on whatever you get anyway

components/win-users/src/account.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct Account {
+  pub name: String,
+  pub system_name: Option<String>,
+  pub domain: String,
+  pub account_type: SID_NAME_USE,
+  pub sid: Sid
+}
+
+impl Account {
+  pub fn from_name(name: String) -> Option<Account> {
```
reset: You can get away with just accepting `&str` here instead of `String` because you're just going to call `clone()` on the param in `lookup_account()` down below

components/core/src/os/users/windows.rs
```diff
@@ -41,3 +55,19 @@ pub fn get_effective_uid() -> u32 {
 pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
     unimplemented!();
 }
+
+pub fn root_level_account() -> String {
+    env::var("COMPUTERNAME").unwrap().to_uppercase() + "$"
+}
+
+#[test]
```
reset: All tests should be placed in a module named `tests` by convention, example:

```
#[cfg(test)]
mod tests {
    use super::*;
}
```

This will get you access to the functions/types you're trying to test within that scope

components/sup/src/util/users.rs
```diff
@@ -64,10 +64,10 @@ fn check_pkg_user_and_group(pkg_install: &PackageInstall) -> Result<Option<(Stri
             let current_user = current_user.unwrap();
             let current_group = current_group.unwrap();
 
-            if current_user == "root" {
+            if current_user == users::root_level_account() {
```
reset: If you were to convert `users::root_level_account()` to a `&'static str` then you'd call `current_user.as_slice()` here to do the comparison

components/win-users/tests/account.rs
```diff
@@ -0,0 +1,50 @@
...
+extern crate habitat_win_users;
+extern crate winapi;
+
+use std::env;
+
+use winapi::winnt::{SidTypeUser, SidTypeWellKnownGroup};
+
+use habitat_win_users::account::Account;
+
+#[test]
```
reset: These tests look like unit tests, I'd place them in a sub module named `tests` inside the Account module

components/win-users/tests/sid.rs
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern crate habitat_win_users;
+
+use habitat_win_users::account::Account;
+
+#[test]
```
reset: Same comment with these tests as the above

components/core/src/os/users/linux.rs
mwrock: Since the windows implementation returns `env::var("COMPUTERNAME").unwrap().to_uppercase() + "$"` which is actually a `String` I'd have to convert one or the other.

components/core/src/os/users/windows.rs
mwrock: Hmm I'm getting compile errors with get_sid_by_name<T: AsString> and not finding refs to that on google. Is this a typo?

components/win-users/src/account.rs
mwrock: how about `system_name` in the `from_name_and_system` function below? Seems like name should be the same type for both and in that case then it seems like also making `system_name` a `&str` would be best too. Does that sound right?

components/win-users/tests/account.rs
mwrock: Is the convention to put only functional/integration tests in separate files?

components/core/src/os/users/windows.rs
mwrock: ahh `Into<String>` very cool but turns out unnecessary since I'm switching to `&str` upstream which gets rid of the need to `to_string()`

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/176225624).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/176225624).

PR #1484, opened 2016-11-17
Initial infra for builder metrics

chefsalim:

This change adds some initial infrastructure for funnel events - specifically a new event module in core that can be used to record events to disk, that will eventually be harvested by a collector and sent out to an external service. The events are written in JSON format to the service var folder. Currently this module is only being used by the builder-api to plumb in a single event as a proof of concept - more to come shortly.

Note that there is some overlap and a bit of copy-pasta from the analytics module. This will get resolved over time as we consolidate among the two modules.

1fnpdkhfzdgou

### DIFF
--- a/components/builder-api/src/http/handlers.rs
+++ b/components/builder-api/src/http/handlers.rs
@@ -17,6 +17,7 @@
 use bodyparser;
 use depot::server::check_origin_access;
 use hab_core::package::Plan;
+use hab_core::event::*;
 use hab_net;
 use hab_net::http::controller::*;
 use hab_net::routing::Broker;
@@ -30,6 +31,12 @@ use router::Router;
 use rustc_serialize::base64::FromBase64;
 
 include!(concat!(env!("OUT_DIR"), "/serde_types.rs"));
+
+lazy_static! {
+    static ref EVENTLOG: EventLogger = {
+        EventLogger::new("hab-builder-api")
+       };
+}
 
 pub fn github_authenticate(req: &mut Request) -> IronResult<Response> {
     let code = {
@@ -223,6 +230,12 @@ pub fn project_create(req: &mut Request) -> IronResult<Response> {
         }
         Err(_) => return Ok(Response::with((status::UnprocessableEntity, "rg:pc:2"))),
     }
+
+    EVENTLOG.record_event(Event::ProjectCreate {
+        origin: origin.get_name(),
+        package: project.get_id(),
+    });
+
     project.set_owner_id(session.get_id());
     request.set_project(project);
     match conn.route::<ProjectCreate, Project>(&request) {
--- a/components/core/src/event.rs
+++ b/components/core/src/event.rs
@@ -1,0 +1,166 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/core/src/event.rs
```diff
@@ -0,0 +1,139 @@
...
+// }
+
+const CLIENT_ID_METAFILE: &'static str = "CLIENT_ID";
+
+#[derive(Debug, Clone)]
+pub struct Event {
+    name: String,
+    clientid: String,
+    timestamp: String,
+    properties: BTreeMap<String, String>,
```
reset: Are we structuring our Event struct like this because that is the format that Segment expects? We might want to format this however we see fit and then massage it to the event type that Segment expects with a `ToSegment` trait. That way we can fully leverage the type system in Rust.

If this is actually just a grab bag of extra data then ignore this comment ;)

components/core/src/event.rs
```diff
@@ -0,0 +1,139 @@
...
+    fs::create_dir_all(parent_dir).expect("Unable to create directory");
+    let mut file = File::create(&file_path).expect("Unable to create file");
+    file.write_all(content.as_bytes()).expect("Unable to write file");
+}
+
+fn timestamp() -> String {
+    let (secs, subsec_nanos) = match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(duration) => (duration.as_secs(), duration.subsec_nanos()),
+        Err(e) => {
+            debug!("Cannot generate system time: {}", e);
```
reset: This should be `error!()` and not `debug!()`. We want the service to keep going even if it's generating less than ideal data, but we want our monitoring tools to pick it up and alert us!

components/core/src/event.rs
```diff
@@ -0,0 +1,139 @@
...
+    if file_path.exists() {
+        read_file(&file_path)
+    } else {
+        let uuid = Uuid::new_v4().hyphenated().to_string();
+        write_file(&cache_dir, &file_path, &uuid);
+        uuid
+    }
+}
+
+pub fn record_event(name: &str) {
```
reset: Here we could use the enum I suggested above

components/core/src/event.rs
```diff
@@ -0,0 +1,139 @@
...
+    let timestamp: &str = &timestamp();
+    let clientid: &str = &client_id();
+    let event = Event::new(name, timestamp, clientid);
+
+    let cache_dir = cache_analytics_path(None);
```
reset: The `cache_analytics_path()` is for the client analytics but these are server side analytics. We want to store these instead in the `var` or `data` directory of the service. I think what we're looking for here is `core::fs::svc_var_path()`

components/builder-api/src/http/handlers.rs
```diff
@@ -17,6 +17,7 @@
 use bodyparser;
 use depot::server::check_origin_access;
 use hab_core::package::Plan;
+use hab_core::event::*;
```
smurawski: (minor style nit) 

Rust imports tend to be very explicit about the bits you are pulling in and avoiding wildcard imports.  It's not just a style thing, but can help prevent future breakage due to overlapping names (You can only have one of a non-module qualified name defined at any point.)

Source - https://aturon.github.io/style/imports.html#avoid-use-*,-except-in-tests.

reset:
@chefsalim this is an even better looking starting place, let's move forward with it!

gif-keyboard-6648375357332554636

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/177150698).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/177150698).

PR #1495, opened 2016-11-20
WIP: Enable relative paths in pkg_source

Hoverbear:
This is a WIP PR to enable `pkg_source` to point to some arbitrary folder relative to `plan.sh` such as `../`.

This is primarily useful in the "in repo" variant workflow described in https://github.com/habitat-sh/habitat/pull/1416#issuecomment-257402252.

To do:

* [ ] Support local/relative tarballs or non-folders.
* [ ] Figure out how to best structure things.
* [ ] ???

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1246,13 +1246,16 @@ download_file() {
 # Will return 0 if the shasums match, and 1 if they do not match. A message
 # will be printed to stderr with the expected and computed shasum values.
 verify_file() {
-  build_line "Verifying $1"
-  local checksum=($($_shasum_cmd $HAB_CACHE_SRC_PATH/$1))
-  if [[ $2 = $checksum ]]; then
-    build_line "Checksum verified for $1"
+  local file="$1"
+  local expected="$2"
+
+  build_line "Verifying $file"
+  local checksum=($($_shasum_cmd $HAB_CACHE_SRC_PATH/$file))
+  if [[ $expected = $checksum ]]; then
+    build_line "Checksum verified for $file"
   else
-    warn "Checksum invalid for $1:"
-    warn "   Expected: $2"
+    warn "Checksum invalid for $file:"
+    warn "   Expected: ${expected}"
     warn "   Computed: ${checksum}"
     return 1
   fi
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

Hoverbear:
@eeyun Would love some feedback from you here.

eeyun:
@Hoverbear I think the concept is delightful. Would be nice to be able to clean up plans for applications that are following the `/app/habitat && /app/source` pattern from what we do now:

```
pkg_source=nosuchfile.tar.gz
pkg_upstream_url=
pkg_shasum=TODO
```

Hoverbear:
@eeyun Can you think of any desirable/undesirable semantics to be aware of?

eeyun:
Hmm not off the top of my head. But, it might be nice to be able to use a `src` or `source` directory instead of a tarball as well. I know that we already "support" this, but having something explicitly declared might not be bad from a composition point of view? I dunno this might be something we should get some more thoughts and feelings on though. 

As it stands right now its pretty straightforward to just need to know what your directory structure should be but /shrug

Hoverbear:
Did some refactoring on this on the weekend. Need to figure out a good way of testing this. :)

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1569,7 +1567,15 @@ do_download() {
 
 # Default implementation for the `do_download()` phase.
 do_default_download() {
-  download_file $pkg_source $pkg_filename $pkg_shasum
+  if [[ $pkg_source == "."* ]]; then
+    # File is a local path, so is already downloaded.
+    cp -r $pkg_source $HAB_CACHE_SRC_PATH/$pkg_filename
```
rhass: @Hoverbear What do you think about putting the copy into a different function such as `do_source_copy` or `do_local_copy` and calling the function within this `if` statement?

My hope is rather than telling people to write a `do_download` callback to customize local file copies, this would allow us to provide a more logically named callback for users should they want to change the copy command or behavior. Do you think that makes sense or would end up being a more confusing user experience?

rhass:
With regards to supporting relative tarballs and non-folders. I have [started playing with an idea](https://github.com/rhass/habitat/compare/rhass/blueprints/RFC3986_compliant_paths) which is to leverage the static-musl curl, which will take care of this for us. I haven't rested it yet, but feel free to poke at it if you feel like running "alpha" quality code. :wink: 

As an added bonus, if the curl thing works it will help pave the way to use URI/URN for pkg_source since it would make files/paths accessed via `file://`. This could open up some very interesting avenues in the long run, eg: nfs://, magnet://, sftp://,  and s3:// are just a few interesting prospects I can think up.

rhass:
@Hoverbear I opened PR #1977 which includes some of the great work you did. Feel free to poke at it and see if you have any issues.

reset:
I'm going to close this in favor of the work done in #2074. @Hoverbear please feel free to reopen if the work there isn't a match with what you envisioned!

Hoverbear:
Awesome. Thanks.

PR #1505, opened 2016-11-22
Updates docoumentation with working links.

dmccown:
Fixes an instance of bad links and updates the documentation with a more
recent example.

Signed-off-by: David McCown <dmccown@chef.io>


reset:
@dmccown thanks for the doc updates!

PR #1510, opened 2016-11-23
More Travis Fun

smurawski:
First off, I should RTFM better - TRAVIS_PULL_REQUEST is the PR # or false, not true or false, so my conditional was wrong.  

Second, somewhere, sometimes, Travis seems to lose the executable status of ./support/ci/deploy_unstable.sh.  What's odd is that it works on some builds but not others.  I added a conditional into the travis steps to check if it is executable and if not, make it so.  It feels so ugly, but I'm not sure what else to do there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -208,6 +208,7 @@ matrix:
           - /root/.cargo
       before_install:
         - ./support/ci/fast_pass.sh || exit 0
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
@thesentinels approve

PR #1513, opened 2016-11-24
Re-implement sidecar module as http_gateway

reset:
* Remove sidecar module and replace with http_gateway
* Refactorings to allow state of Manager to be shared with http gateway
* Add initial raml documentation for supervisor http gateway

gif-keyboard-7839051923796985666


thesentinels:

Thank you for contributing!

components/butterfly/src/message/mod.rs
```diff
@@ -41,3 +44,117 @@ pub fn unwrap_wire(payload: &[u8], ring_key: &Option<SymKey>) -> Result<Vec<u8>>
         Ok(wire.take_payload())
     }
 }
+
+impl Encodable for swim::Election {
```
adamhjk: I love how this comes together.

components/sup/doc/api.raml
```diff
@@ -0,0 +1,75 @@
+#%RAML 0.8
+---
+title: Habitat Supervisor API
+
+baseUri: https://{rootUri}
```
adamhjk: This should be http, not https, unless we're auto-generating certificates. That's probably a neat idea, actually, but it wouldn't prove a ton security wise.

components/sup/src/http_gateway.rs
```diff
@@ -0,0 +1,194 @@
...
+use router::Router;
+use rustc_serialize::json;
+
+use config::gconfig;
+use error::{Result, Error, SupError};
+use health_check;
+use manager::{self, census};
+use util::signals;
+
+static LOGKEY: &'static str = "HG";
```
adamhjk: gif-keyboard-12254865723829879485

components/sup/src/http_gateway.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(PartialEq, Eq, Debug)]
+pub struct ListenAddr(pub SocketAddrV4);
+
+impl Default for ListenAddr {
+    fn default() -> ListenAddr {
+        ListenAddr(SocketAddrV4::new(Ipv4Addr::new(0, 0, 0, 0), 9631))
```
adamhjk: Should this be configurable, or a constant?

components/sup/src/http_gateway.rs
```diff
@@ -0,0 +1,194 @@
...
+impl DerefMut for ListenAddr {
+    fn deref_mut(&mut self) -> &mut SocketAddrV4 {
+        &mut self.0
+    }
+}
+
+impl FromStr for ListenAddr {
+    type Err = SupError;
+
+    fn from_str(val: &str) -> Result<Self> {
```
adamhjk: Why not just use the normal SocketAddr traits, and let this work for both v4 and v6?

Specifically: https://doc.rust-lang.org/std/net/enum.SocketAddr.html should let you take an arbitrary v4 or v6 socket.

components/sup/src/http_gateway.rs
```diff
@@ -0,0 +1,194 @@
...
+                self.0
+                    .http(*gconfig().http_listen_addr())
+                    .expect("unable to start http-gateway thread");
+            }));
+        Ok(handle)
+    }
+}
+
+fn butterfly(req: &mut Request) -> IronResult<Response> {
+    let state = req.get::<persistent::Read<ManagerState>>().unwrap();
```
adamhjk: turbofish to the rescue.

components/sup/src/http_gateway.rs
```diff
@@ -0,0 +1,194 @@
...
+fn census(req: &mut Request) -> IronResult<Response> {
+    let state = req.get::<persistent::Read<ManagerState>>().unwrap();
+    let data = state.census_list.read().unwrap();
+    Ok(Response::with((status::Ok, json::encode(&*data).unwrap())))
+}
+
+fn config(req: &mut Request) -> IronResult<Response> {
+    let state = req.get::<persistent::Read<ManagerState>>().unwrap();
+    let service_group =
+        ServiceGroup::new(req.extensions.get::<Router>().unwrap().find("svc").unwrap(),
```
adamhjk: What happens when this panics? Can it essentially not panic because of the shape of the route?

components/sup/src/main.rs
```diff
@@ -153,31 +152,16 @@ fn config_from_args(subcommand: &str, sub_args: &ArgMatches) -> Result<()> {
...
-                                            DEFAULT_HTTP_LISTEN_IP,
-                                            DEFAULT_HTTP_LISTEN_PORT));
-
-    debug!("HTTP IP = {}", &sidecar_ip);
-    debug!("HTTP port = {}", &sidecar_port);
-
-    config.set_http_listen_ip(sidecar_ip);
-    config.set_http_listen_port(sidecar_port);
-
+    if let Some(addr_str) = sub_args.value_of("listen-http") {
```
adamhjk: yeah - since this is stringy, I think my comment above about just taking `SocketAddr` rather than a specific v4 is the right thing, as we will then work transparently with v6.

components/sup/src/manager/mod.rs
```diff
@@ -15,35 +15,54 @@
...
 use manager::signals::SignalEvent;
+use package::Package;
+use http_gateway;
+use util;
 
 static LOGKEY: &'static str = "MR";
 
-#[derive(Debug)]
+#[derive(Clone)]
+pub struct State {
```
adamhjk: This turned out very nicely.

components/sup/src/manager/mod.rs
```diff
@@ -183,8 +193,8 @@ impl Manager {
 
     /// Walk each service and check if it has an updated package installed via the Update Strategy.
     pub fn check_for_updated_packages(&mut self) {
-        for service in self.services.iter_mut() {
-            service.check_for_updated_package();
+        for service in self.state.services.write().unwrap().iter_mut() {
```
adamhjk: Man, I hate taking this lock so frequently for an operation that almost never will actually write. Given how the code is written, I think it won't matter much, since we don't have a ton of parallel read locks outstanding, so the cost is probably minimal. But I pour one out for the lock-free services list. :)

components/sup/src/manager/mod.rs
```diff
@@ -104,74 +115,73 @@ impl Manager {
...
         }
-
         Ok((false, update))
     }
 
     fn check_for_incoming_signals(&mut self) -> bool {
         match signals::check_for_signal() {
             Some(SignalEvent::Shutdown) => {
-                for service in self.services.iter_mut() {
+                for service in self.state.services.write().unwrap().iter_mut() {
```
adamhjk: These unwrap's all need to become `expect("Services lock is poisoned")` or something similar, rather than just the raw explosion. Makes a big difference in debug hunting - the raw unwrap tells you almost nothing about the lock you're digging for.

components/sup/src/manager/mod.rs
```diff
@@ -104,74 +115,73 @@ impl Manager {
...
                     outputln!("Shutting down {}", service);
                     service.down()
                         .unwrap_or_else(|err| outputln!("Failed to shutdown {}: {}", service, err));
                 }
                 true
             }
             Some(SignalEvent::Passthrough(signal_code)) => {
-                for service in self.services.iter() {
+                for service in self.state.services.read().unwrap().iter() {
```
adamhjk: Should be `expect`

components/sup/src/manager/mod.rs
adamhjk: Should be expect

components/sup/src/manager/mod.rs
```diff
@@ -240,22 +257,24 @@ impl Manager {
                 last_census_update = ncu;
             }
 
-            for service in self.services.iter_mut() {
+            for service in self.state.services.write().unwrap().iter_mut() {
```
adamhjk: Should be `expect`

components/sup/src/manager/mod.rs
```diff
@@ -240,22 +257,24 @@ impl Manager {
...
+                        .service_config_for(&service.service_group_str(),
+                                            service.service_config_incarnation) {
                     service_config_updated = service.write_butterfly_service_config(config);
                     service.service_config_incarnation = Some(incarnation);
                 }
 
                 // Reconfigure if neccessary
                 if census_updated || service_config_updated {
-                    service.reconfigure(&self.census_list);
+                    service.reconfigure(&self.state.census_list.read().unwrap());
```
adamhjk: Should be `expect`.

components/sup/src/manager/mod.rs
```diff
@@ -266,7 +285,7 @@ impl Manager {
 
                 // Start or restart the service
                 if service.needs_restart || service.is_down() {
-                    match service.restart(&self.census_list) {
+                    match service.restart(&self.state.census_list.read().unwrap()) {
```
adamhjk: Should be `expect`

components/sup/src/manager/service_updater.rs
```diff
@@ -0,0 +1,143 @@
...
+        if service.update_strategy == UpdateStrategy::None {
+            return false;
+        }
+        let rx = Worker::new(service).start();
+        self.workers.insert(service.package.ident().clone(), rx);
+        true
+    }
+}
+
+struct Worker {
```
adamhjk: This whole refactor is great.

adamhjk:
I'm going to carry this patch for you, @reset. I need your refactor of the manager and friends to land to avoid further merge conflicts as I add back in encryption to the service configuration and service files.

components/sup/src/http_gateway.rs
reset: Exactly, you can't panic here - the handler would never be hit

components/sup/doc/api.raml
adamhjk: Fixed

components/sup/src/http_gateway.rs
reset: This will inline so we don't need to move it to a constant. This is literally the "default ip addr/port" for the http_gateway so I don't think this should be configurable here, instead we configure by the CLI & assign it to the Config struct. This is just what will be populated into that config struct by default

components/sup/src/http_gateway.rs
reset: Great suggestion - I totally didn't know about that enum!

reset:
@adamhjk great code review, I didn't even know about the SocketAddr enum, thank you!

I addressed all of your comments. I totally didn't see your last message about handling this for me, sorry about that. I needed some programming time on this vacation anyway ;)

adamhjk:
@thesentinels merge

adamhjk:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/179099564).

adamhjk:
gif-keyboard-2766689288864813007

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/179099564).

PR #1517, opened 2016-11-28
Add `hab file` commands and encryption

adamhjk:
* Adds in the `hab file` commands
* Turns security and encryption on for both `config apply` and `file
  upload`
* Breaks the `config apply` and `file upload` commands into a separate
  `hab-butterfly` component and command.
* Fixes the order of arguments to be the same between file upload and
  config apply.
* Allows for encryption of configuration as well as files.
* Improves the `butterfly` http gateway output to render as strings if
  the contents are UTF-8.
* Allows for the hashing of a pure byte buffer.
* Failed initialization will now block starting/restarting.

gif-keyboard-15632042006024311536

Signed-off-by: Adam Jacob <adam@chef.io>

PR #1518, opened 2016-11-28
The great clean up in ing

adamhjk:
Depends on #1517.

This commit cleans up all the warnings in the supervisor, core, common, and director. It removes un-needed crates, deletes quite a bit of dead code, and removes any use of Wonder, including refactoring the directors signal handling.

It should compile a full `make test` pass without warnings, and all tests should pass.

gif-keyboard-17861947114191544080


thesentinels:

Thank you for contributing!

adamhjk:
The issue here is we need to define a `Signal` struct in the Windows signal section, and export it.

smurawski:
I think the actual windows process signals are - https://msdn.microsoft.com/en-us/library/xdkz3x12.aspx

PR #1551, opened 2016-12-10
[sup] Reimplement recursive toml config merging.

fnichol:
This change reimplements the logic which is responsible for merging down
runtime, environment, user, and default config for a service. The
initial implementation was a shallow merge only, meaning that sub tables
where not merged with each other--only top level keys. The next
implementation built on this code by supporting deep merging of toml
tables if present. There were a few issues with this implementation when
different value types were present for the same key. Additionally, the
Rust match-based logic was extended to support this which resulted in a
harder to understand function. Lastly, using recursion as part of the
implementation introduced a potential attack vector where a user could
craft a data structure deep enough to overflow the stack and cause the
supervisor to unwind and terminate.

This current implementation reduces the merging logic to the following:

* There are 2 toml tables: `me` (the current, config which will be
  mutated in place) and `other` (a reference to a config which override
  the current config).
* For each key in `other`, perform one of two things:
* If the values at `key` in `me` and `other` both contains toml tables,
  then call `toml_merge_recurse` on those value tables to recursively
  merge them down.
* Otherwise copy the value of `other` at `key` into the value of `key`
  in `me`. It makes no difference if the value types differ or are the
  same--the `other` table will always override anything in `me`.

The merge logic also contains a maximum recursive depth check in an
attempt to prevent burning through all stack space.

### DIFF
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -39,6 +39,13 @@ use VERSION;
 static LOGKEY: &'static str = "SC";
 static ENV_VAR_PREFIX: &'static str = "HAB";
 
+/// The maximum TOML table merge depth allowed before failing the operation. The value here is
+/// somewhat arbitrary (stack size cannot be easily computed befhorehand and different libc
+/// implementations will impose different size constraints), however a parallel data structure that
+/// is deeper than this value crosses into overly complex territory when describing configuration
+/// for a single service.
+static TOML_MAX_MERGE_DEPTH: u16 = 30;
+
 /// The top level struct for all our configuration - this corresponds to the top level namespaces
 /// available in `config.toml`.
 #[derive(Debug, RustcEncodable)]
@@ -92,7 +99,7 @@ impl ServiceConfig {
         let sys = try!(self.sys.to_toml());
         top.insert(String::from("sys"), sys);
 
-        let cfg = self.cfg.to_toml();
+        let cfg = try!(self.cfg.to_toml());
         top.insert(String::from("cfg"), cfg);
 
         let svc = self.svc.to_toml();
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/config.rs
```diff
@@ -316,41 +316,39 @@ struct Cfg {
...
-    for (ovr_key, ovr_value) in ovr.iter() {
-
-        // attempt to merge nested tables
-        let existing_v = existing.get(ovr_key).and_then(|v| v.as_table());
-        let ovr_v = ovr.get(ovr_key).and_then(|v| v.as_table());
-        if let None = existing.get(ovr_key) {
-            final_map.insert(ovr_key.clone(), ovr_value.clone());
+// Recursively merges two TOML tables
+fn toml_merge(me: &toml::Table, other: &toml::Table) -> toml::Table {
+    let mut result = me.clone();
```
reset: Could we take a mutable reference to `me` instead of cloning the data? This is going to be quite costly

components/sup/src/manager/service/config.rs
```diff
@@ -316,41 +316,39 @@ struct Cfg {
...
+    for (key, other_value) in other.iter() {
+        if is_toml_value_a_table(key, me) && is_toml_value_a_table(key, other) {
+            result.insert(key.clone(),
```
reset: We may want to consider consuming `other` if we aren't going to be using it after we call this function. That will allow us to *take* the key/values and also reduce our allocations

components/sup/src/manager/service/config.rs
reset: Since this is a generic function for performing TOML merges (albeit private to our Config class) I would be careful about weighing on how small our current data set is. Somebody is going to call this when one of us isn't paying attention on a struct that could dynamically grow to massive sizes.

At the very least I would rename this to something like `toml_merge_and_new_from()` or something that indicates it's evil allocation tendencies :)

reset:
@fnichol the logic looks great! Let's sort out the over-allocations and merge

reset:
gif-keyboard-13260391840291368851

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/185565857).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/185565857).

PR #1552, opened 2016-12-10
Terraform Admin Srv

reset:
This PR adds Terraform resources for the builder-admin and builder-admin-proxy services. It also makes some small changes to the builder-api-proxy plan which fixes the permissions issue when a worker process attempts to write to the service directory


thesentinels:

Thank you for contributing!

PR #1577, opened 2016-12-14
Supervisor on windows can "supervise" process for state change

mwrock:
This adds support for windows processes under a supervisor's supervision. Previously when the supervisor checked the status of a running process executed from a supervisor's run hook, we hard coded a `waitpid` status of `0` which means there is no transition state for the process and it is running normally. This allowed us to at least run the supervisor and launch run hooks on windows but prevented us from taking any sort of intervention steps.

This PR adds a `HabChild` struct that has different `status` trait implementatins on linux and windows and both return a `HabExitStatus`. Linux runs the familiar `waitpid` and now windows will obtain a process `HANDLE` which it can probe for status. In the event that the process dies immediately and there is no handle to obtain, we call `wait()` and cache the exit code and return that the next tine the supervisor requests status.

This PR also refactors the process status "interpretation" logic formerly used by linux:

1. `supervisor.rs` previously had several internal functions to retrieve exit code or signal. That is now moved to the `HabExitStatus` struct and it mostly leverages `libc` for these functions. The logic in `HabExitStatus` is modeled after the `std::process::ExitStatus` logic which will return an exit code if the process terminated otherwise it assumes the termination was signaled and it returns the signal.
2. The previous logic in the supervisor had a "catch all" arm that assumed that it was possible that a termination could succesfully be reported that was neither from a signal or normal exit. Based on how we are calling `waitpid`(with the `WNOHANG` option only), that is not possible.
3. The former supervisor had logic that if `waitpid` returned a pid that was different from the pid being supervised, it would check their signal/exit code. Again, based on the way we are calling `waitpid` (with a specific pid) we will ONLY receive termination status for THAT pid. The intent of this logic appears to have been for reaping zombies and perhaps did just that at one time but I am removing it here since it does not fit well within the new structure and is effectively dead code. That said we may want to pursue a separate fix to address zombies propperly since the current supervisor has no awareness of any process other than the one launched from the run hook.

### DIFF
--- a/components/core/src/os/process/linux.rs
+++ b/components/core/src/os/process/linux.rs
@@ -12,32 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use libc::{pid_t, c_int};
+use libc;
 use std::ffi::OsString;
 use std::path::PathBuf;
 use std::os::unix::process::CommandExt;
-use std::process::Command;
+use std::process::{self, Command};
 
-use error::Result;
+use error::{Error, Result};
+
+use super::{HabExitStatus, ExitStatusExt};
 
 extern "C" {
     fn kill(pid: i32, sig: u32) -> u32;
-    fn waitpid(pid: pid_t, status: *mut c_int, options: c_int) -> pid_t;
+    fn waitpid(pid: libc::pid_t, status: *mut libc::c_int, options: libc::c_int) -> libc::pid_t;
 }
 
 pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
     become_exec_command(command, args)
 }
 
-pub fn wait_for_exit(pid: u32, status: *mut c_int) -> u32 {
-    unsafe { waitpid(pid as i32, status, 1 as c_int) as u32 }
-}
-
 /// send a Unix signal to a pid
 pub fn send_signal(pid: u32, sig: u32) -> u32 {
-    unsafe {
-        kill(pid as i32, sig)
-    }
+    unsafe { kill(pid as i32, sig) }
 }
 
 /// Makes an `execvp(3)` system call to become a new program.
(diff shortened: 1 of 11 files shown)


thesentinels:

Thank you for contributing!

components/core/src/os/process/linux.rs
```diff
@@ -54,3 +52,116 @@ fn become_exec_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
...
+            0 => Ok(HabExitStatus { status: None }),
+            -1 => Err(Error::WaitpidFailed(format!("Error calling waitpid on pid: {}", self.pid))),
+            _ => Ok(HabExitStatus { status: Some(exit_status as u32) })
+        }
+    }
+}
+
+impl ExitStatusExt for HabExitStatus {
+    fn code(&self) -> Option<u32> {
+        unsafe {
```
reset: It doesn't appear that you aren't performing any unsafe operations in this code block. You should remove the unsafe block here if that's the case!

components/core/src/os/process/windows.rs
```diff
@@ -47,3 +50,147 @@ fn become_child_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
...
+        }
+        else {
+            return Some(proc_handle);
+        }
+    }
+}
+
+pub struct Child {
+    handle: Option<winapi::HANDLE>,
+    last_status: Option<u32>,
```
mwrock: Its technically a `u32` on Windows - a `DWORD`. While unix is `i32`, they are always positive so casting from `i32` to `u32` seemed like the safest thing to do.

components/core/src/os/process/mod.rs
```diff
@@ -12,15 +12,61 @@
...
 #[cfg(not(windows))]
-pub use self::linux::{become_command, wait_for_exit, send_signal};
+pub use self::imp::{become_command, send_signal};
+
+pub struct HabChild {
+    inner: imp::Child,
+}
+
+impl HabChild {
+    pub fn from_child(child: &mut Child) -> HabChild {
```
mwrock: I don't think I can use that trait with a mutable paramerter. correct?

mwrock: `impl From<&mut Child> for HabChild {` yields:
```
   |
45 | impl From<&mut Child> for HabChild {
   |           ^ expected lifetime parameter
```

reset: Yeah you totally can, I just didn't tell you the right way to use it ;)

```
imp<'a> From<&'a mut Child> for HabChild {
    fn from(other: &mut Child) -> HabChild {
        HabChild { inner: impl::Child::new(child) }
    }
}
```

components/core/src/os/process/windows.rs
mwrock: It is but here is the [signature](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683189(v=vs.85).aspx) for GetExitCodeProcess. Its a pointer to a DWORD.

components/core/src/os/process/windows.rs
```diff
@@ -47,3 +50,147 @@ fn become_child_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
...
+impl Child {
+    // On windows we need the process handle to capture status
+    // Here we will attempt to get the handle from the pid but if the
+    // process dies before we can get it, we will just wait() on the
+    // std::process::Child and cache the exit_status which we will return 
+    // when ststus is called.
+    pub fn new(child: &mut process::Child) -> Child {
+        let (win_handle, status) = match handle_from_pid(child.id()) {
+            Some(handle) => (Some(handle), None),
+            _ => (None, Some(child.wait().unwrap().code().unwrap() as u32))
```
mwrock: We actually want to panic if it errors here because in that case we have we have lost contact with the process, but we should totally panic with a better error.

reset: We can't panic the entire process because the Supervisor will be supervising multiple services in the very near future. We can bubble an error up to the top and not handle it for now and that's just fine!

components/core/src/os/process/windows.rs
reset: Ok, no worries - perfectly reasonable!

mwrock:
All the comments should now be addressed @reset

reset:
gif-keyboard-5738271046109575339

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/184626443).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/184626443).

PR #1598, opened 2016-12-22
[devshell] Re-add `redis-server` pkg for `make bldr-run-shell` target.

fnichol:
I made an incorrect assumption that the `redis-server` Ubuntu package
was not being used for anything, when it was for running the Builder
service stack via the `make bldr-run-shell` script. This change readds
that package to the default "devshell".

gif-keyboard-6325974613419619595


thesentinels:

Thank you for contributing!

reset:
Good catch!

gif-keyboard-5166927647261727844

PR #1603, opened 2016-12-23
implement graceful shutdown on windows

mwrock:

### DIFF
--- a/components/core/src/os/process/windows.rs
+++ b/components/core/src/os/process/windows.rs
@@ -17,13 +17,14 @@ use std::path::PathBuf;
 use std::process::{self, Command};
 use std::ptr;
 use std::io;
+use time::{Duration, SteadyTime};
 
 use kernel32;
 use winapi;
 
 use error::{Error, Result};
 
-use super::{HabExitStatus, ExitStatusExt};
+use super::{HabExitStatus, ExitStatusExt, ShutdownMethod};
 
 const STILL_ACTIVE: u32 = 259;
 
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

components/core/src/os/process/windows.rs
```diff
@@ -27,6 +28,10 @@ use super::{HabExitStatus, ExitStatusExt};
 
 const STILL_ACTIVE: u32 = 259;
 
+const PROCESS_EXITED: i32 = 0;
```
reset: These are private to the module but returned from a public function - they should be public so they can be referenced by name

reset: They also might be better represented by an enum if that fits with your API

mwrock: Yeah I thought about the enum but that did not mesh with the Linux side which uses the libc signal constants.

Thinking again about this maybe a better approach would be to define an enum in mod.rs that just signifies graceful or abrupt shutdown.

mwrock: Then it's clear enough to infer based on platform what actually happened

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/186462582).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/186462582).

PR #1469, opened 2016-11-10
Add metrics endpoint to supervisor

jamesc:
This add a new endpoint `/metrics` to the supervisor that exports prometheus-style metrics.

Right now it exports HTTP metrics for all the endpoints in the side car and a `habitat_package_info` metrics which shows package metadata.  Sample output is:

```
# HELP habitat_package_info package version information
# TYPE habitat_package_info gauge
habitat_package_info{name="redis",origin="core",release="20161104175435",version="3.2.4"} 1
# HELP http_request_duration_seconds HTTP request latencies in seconds.
# TYPE http_request_duration_seconds histogram
http_request_duration_seconds_bucket{handler="config",le="0.005"} 1
http_request_duration_seconds_bucket{handler="config",le="0.01"} 1
http_request_duration_seconds_bucket{handler="config",le="0.025"} 1
http_request_duration_seconds_bucket{handler="config",le="0.05"} 1
http_request_duration_seconds_bucket{handler="config",le="0.1"} 1
http_request_duration_seconds_bucket{handler="config",le="0.25"} 1
http_request_duration_seconds_bucket{handler="config",le="0.5"} 1
http_request_duration_seconds_bucket{handler="config",le="1"} 1
http_request_duration_seconds_bucket{handler="config",le="2.5"} 1
http_request_duration_seconds_bucket{handler="config",le="5"} 1
http_request_duration_seconds_bucket{handler="config",le="10"} 1
http_request_duration_seconds_bucket{handler="config",le="+Inf"} 1
http_request_duration_seconds_sum{handler="config"} 0.000456555
http_request_duration_seconds_count{handler="config"} 1
http_request_duration_seconds_bucket{handler="metrics",le="0.005"} 4
http_request_duration_seconds_bucket{handler="metrics",le="0.01"} 4
http_request_duration_seconds_bucket{handler="metrics",le="0.025"} 4
http_request_duration_seconds_bucket{handler="metrics",le="0.05"} 4
http_request_duration_seconds_bucket{handler="metrics",le="0.1"} 4
http_request_duration_seconds_bucket{handler="metrics",le="0.25"} 4
http_request_duration_seconds_bucket{handler="metrics",le="0.5"} 4
http_request_duration_seconds_bucket{handler="metrics",le="1"} 4
http_request_duration_seconds_bucket{handler="metrics",le="2.5"} 4
http_request_duration_seconds_bucket{handler="metrics",le="5"} 4
http_request_duration_seconds_bucket{handler="metrics",le="10"} 4
[...]

PR #1613, opened 2017-01-03
Add infra for operational metrics

chefsalim:

This change adds infrastructure for adding operational metrics to the Hab builder services. The underlying infrastructure is based on a statsd aggregator, howver the metrics API abstracts out any specific statsd dependency.  In order for the metrics to be sent to statsd, the following env variable needs to be configured: HAB_STATS_ADDR, with a value of [IP Addr]:[Port], eg, 192.168.1.1:8125

giphy

### DIFF
--- a/components/builder-core/Cargo.toml
+++ b/components/builder-core/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "builder_core"
+version = "0.0.0"
+authors = ["Adam Jacob <adam@chef.io>", "Jamie Winsor <reset@chef.io>", "Fletcher Nichol <fnichol@chef.io>", "Joshua Timberman <joshua@chef.io>", "Dave Parfitt <dparfitt@chef.io>", "Steven Murawski <smurawski@chef.io>"]
+workspace = "../../"
+build = "../bldr-build.rs"
+
+[dependencies]
+log = "*"
+statsd = "*"
+
+[dependencies.habitat_core]
+path = "../core"
+
+[features]
+functional = []
--- a/components/builder-core/src/metrics.rs
+++ b/components/builder-core/src/metrics.rs
@@ -1,0 +1,232 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::{Once, ONCE_INIT};
+use std::sync::mpsc::{channel, sync_channel, Sender, Receiver, SyncSender};
+use std::thread;
+use statsd::Client;
+use hab_core::env;
+
+// Statsd Application name
+pub const APP_NAME: &'static str = "bldr";
+
+// Statsd Listener Address
+pub const STATS_ENV: &'static str = "HAB_STATS_ADDR";
+
+// Supported metrics
+#[derive(Debug, Clone)]
+pub enum Counter {
+    SearchPackages,
+}
+
+// Supported metrics
+#[derive(Debug, Clone)]
+pub enum Gauge {
+    PackageCount,
+}
+
+// Helper types
+#[derive(Debug, Clone, Copy)]
+enum MetricType {
+    Counter,
+    Gauge,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum MetricOperation {
+    Increment,
+    Decrement,
+    SetValue,
+}
+
+type MetricId = &'static str;
+type MetricValue = f64;
+type MetricTuple = (MetricType, MetricOperation, MetricId, Option<MetricValue>);
+
+trait Metric {
+    fn id(&self) -> &'static str;
+}
+
+// One-time initialization
+static mut SENDER: *const Sender<MetricTuple> = 0 as *const Sender<MetricTuple>;
+
+static INIT: Once = ONCE_INIT;
+
+fn sender() -> Sender<MetricTuple> {
+    unsafe {
+        INIT.call_once(|| {
+            SENDER = Box::into_raw(Box::new(init()));
+        });
+        (*SENDER).clone()
+    }
+}
+
+// init creates a worker thread ready to receive and process metric events,
+// and returns a channel for use by metric senders
+fn init() -> Sender<MetricTuple> {
+    let (tx, rx) = channel::<MetricTuple>();
+    let (rztx, rzrx) = sync_channel(0); // rendezvous channel
+
+    thread::Builder::new()
+        .name("metrics".to_string())
+        .spawn(move || receive(rztx, rx))
+        .expect("couldn't start metrics thread");
+
+    match rzrx.recv() {
+        Ok(()) => tx,
+        Err(e) => panic!("metrics thread startup error, err={}", e),
+    }
+}
+
+// receive runs in a separate thread and processes all metrics events
+fn receive(rz: SyncSender<()>, rx: Receiver<MetricTuple>) {
+    let mut client = statsd_client();
+    rz.send(()).unwrap(); // Blocks until the matching receive is called
+
+    loop {
+        let (mtyp, mop, mid, mval): MetricTuple = rx.recv().unwrap();
+        debug!("Received metrics tuple: {:?}", (mtyp, mop, mid, mval));
+
+        match client {
+            Some(ref mut cli) => {
+                match mtyp {
+                    MetricType::Counter => {
+                        match mop {
+                            MetricOperation::Increment => cli.incr(mid),
+                            MetricOperation::Decrement => cli.decr(mid),
+                            _ => error!("Unexpected metric operation: {:?}", mop),
+                        }
+                    }
+                    MetricType::Gauge => {
+                        match mop {
+                            MetricOperation::SetValue => cli.gauge(mid, mval.unwrap()),
+                            _ => error!("Unexpected metric operation: {:?}", mop),
+                        }
+                    }
+                }
+            }
+            None => (),
+        }
+    }
+}
+
+fn statsd_client() -> Option<Client> {
+    match env::var(STATS_ENV) {
+        Ok(addr) => {
+            match Client::new(&addr, APP_NAME) {
+                Ok(c) => Some(c),
+                Err(e) => {
+                    debug!("Error creating statsd client: {:?}", e);
+                    None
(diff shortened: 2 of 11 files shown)


thesentinels:

Thank you for contributing!

components/core/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::thread;
+use std::sync::mpsc::{channel, sync_channel, Sender, Receiver, SyncSender};
+use std::sync::{Once, ONCE_INIT};
+use statsd::Client;
+use env;
```
fnichol: Could you lexically sort and arrange the imports here? I'm struggling to find the style docs we've been unofficially woking off, but generally speaking we have 3 stanzas/group of imports: 1) `std::*` imports, 2) extern crate imports, 3) current crate imports ([sup example](https://github.com/habitat-sh/habitat/blob/23afffd3ef53a80def00731702faa53faf635d9a/components/sup/src/manager/service_updater.rs#L15-L35), [rust/fmt-rfcs example](https://github.com/rust-lang-nursery/fmt-rfcs/blob/b4c703ce19441089e065775a26a37dbf4150aad5/example/lists.rs#L11-L18))

components/core/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+
+// Statsd Application name
+pub const APP_NAME: &'static str = "bldr";
```
fnichol: Is this infra meant for Builder specifically or Habitat in general? If it's the former then a common `builder-*` crate might be a better home (`core` generally is a dependency of all Habitat higher level components) or if it's the latter then we may eventually want this as a parameter to using this module.

adamhjk: Why is this bldr and not `hab`? I assume its because we're going to be embedding this in various bldr components, but what about the `hap-sup` itself, which also could use a generic metrics endpoint, both for sending and recieving?

components/core/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+        Ok(addr) => Some(Client::new(&addr, APP_NAME).unwrap()),
+        Err(_) => None,
+    }
+}
+
+impl Counter {
+    pub fn increment(&self) {
+        sender()
+            .send((MetricType::Counter, MetricOperation::Increment, &self.id(), None))
+            .unwrap();
```
fnichol: Is there a reason that `unwrap()` is used in the client side API? I believe that if this call returns `Err` (such as if the metrics thread crashes and therefore hangs up the receiver side), this would cause the client/calling side to also panic and unwind its thread. If we're okay ignoring failures, it might be better to explicitly squelch any `Err` rather than crashing out the thread.

components/core/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+impl Gauge {
+    pub fn set(&self, val: f64) {
+        sender()
+            .send((MetricType::Gauge, MetricOperation::SetValue, &self.id(), Some(val)))
+            .unwrap();
+    }
+}
+
+impl Metric for Counter {
+    fn id(&self) -> &'static str {
```
fnichol: Nice use of lifetimes to save a String allocation!

components/core/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+                    }
+                }
+            }
+            None => (),
+        }
+    }
+}
+
+fn statsd_client() -> Option<Client> {
+    match env::var(STATS_ENV) {
```
adamhjk: Should this become a CLI flag as well?

adamhjk:
Generically, I'm also worried about that statsd implementations reliance on a deprecated library (`clock_ticks`) - which should probably get ported to SteadyTime. I don't think we should block the work on it, though.

components/core/src/metrics.rs
fnichol: Possibly also a config value for the `builder-*` components. In general we don't support many environment variables directly in lower level code with the exception perhaps of `HAB_CACHE_KEY_PATH`.

components/core/src/metrics.rs
adamhjk: better for this to return an error and let the client do the crashing if it needs to.

adamhjk: at the very least, it should be `expect()`

components/core/src/metrics.rs
chefsalim: Sure.. it looks roughly in that order already.. just swap the first two lines?

components/core/src/metrics.rs
chefsalim: This is meant for Builder operational metrics, as per the card: "As a developer of the Builder service, I need a way to define application counters".  I'll look at moving it to a 'builder-core' crate.

components/core/src/metrics.rs
chefsalim: That seems reasonable, will update.

components/core/src/metrics.rs
chefsalim: thanks to @reset for the trait/static str trick :)

components/core/src/metrics.rs
chefsalim: statsd is probably not the right strategy for hab-sup.. there's a different PR to add metrics to the sidecar that seems to cover hab-sup?  https://github.com/habitat-sh/habitat/pull/1469

components/core/src/metrics.rs
chefsalim: Because this API needs to be dead simple to use from any builder component, the low level env var approach is simpler/more workable than a config.  I'd prefer to see how the env approach works out and if needed we can look at config value(s).

components/core/src/metrics.rs
fnichol: Looks like they are in the correct order, just need a blank line between the stanza sections. It helps a lot when visually scanning the top of an `.rs` file to see what extern and internal dependencies are being used in a module.

components/core/src/metrics.rs
fnichol: Cool, I assumed there was a good story behind this one, just double checking.

components/core/src/metrics.rs
fnichol: Probably better to return an error unless the caller can't easily cope with failures--I'm not sure.

@adamhjk I've been wanting to have a chat with the team about `.expect()`. While much, much nicer and more expressive that `.unwrap()`, it still causes a panic on `Err`/`None`. Looking back through some of our code in the last few months I'm concerned that we've introduced more places which can terminate a thread of execution.

components/core/src/metrics.rs
fnichol: Hehe, I was about to correct this as fishy until I thought about the lifetime of the `"my string literal"`. I'd bet I have places to drop in this trick too.

components/core/src/metrics.rs
adamhjk: Yes - if you really can't recover, you should use `expect()`. If it's possible to recover, you should return a `Result` and let the caller do the needful. 

In particular, butterfly suffers from some of this - it needs a refactor to either use the new capabilities to capture an unwind at the thread boundary and restart, or refactor to deal with errors at the thread boundary.

components/core/src/metrics.rs
adamhjk: All those components also take configuration options from the CLI, and that's the existing pattern. I would prefer to stick with it, unless we have a deeper reason to drop it.

components/core/src/metrics.rs
adamhjk: If statsd is the wrong strategy for hap-sup, why is it the right strategy for the build service? #1469 uses the prometheus style for distribution, and includes an example of how some internal supervisor metrics could be exposed, but doesn't provide any sort of contract for a service to expose metrics which are then collected. As far as protocols to do that go, statsd is pretty great, and there is an existing prometheus bridge. Whatever we decide for the sup is eventually what we should use for builder, as these are all open components into a broader ecosystem. The other direction is a fine way (we add instrumentation to the builder services, which grows into the supervisor) - but that's how it should work, because relaying application instrumentation is part of what the supervisor needs to do long term.

fnichol:
Could you add an entry in the `LIB` line in the `Makefile` as well and add the directory path to the `AFFECTED_DIRS` environment variables in `.travis.yml`? This will ensure that CI will fire the appropriate tasks for the new crate.

components/builder-core/src/metrics.rs
```diff
@@ -0,0 +1,224 @@
...
+                        }
+                    }
+                }
+            }
+            None => (),
+        }
+    }
+}
+
+fn statsd_client() -> Option<Client> {
```
fnichol: I'd still prefer to see a `fn statsd_client(url: &Url)` or secondarily `fn statsd_client(url: &str)` as the mechanism by which to inject the Stasd service dependency. At the end of the day the Builder components will be using and consuming their own config (as well as setting suitable defaults for their `Config` structs) so we'll need that plumbing eventually. Additionally, if the string provided by the user/environment doesn't parse out to an appropriate URL I suspect that the `Client::new(...).unwrap()` call will `panic!` the metrics thread and will panic the main thread as a result (when the client doesn't properly start). Finally, having an environment variable drive this makes the service spawning more complicated as we need to add `Environment` declarations in our systemd units (or an extra environment variable if launched via a container) that is out-of-band from the remaining service config (all provided via Habitat).

components/builder-core/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "builder_core"
```
fnichol: Nice!!

components/builder-core/src/metrics.rs
chefsalim: That's actually where we had started initially, however backed away from an explicit config. One, we'll need to plumb in this config for around 9 different builder components that have configs. It seems like it might be harder to manage and make sure we get everything consistent. Secondly, and more challengingly, the way the user facing API is envisioned where it can be called from any component (not just the web services), without explicitly initing by the caller, plumbing in the service config into the one-time init and/or the API call makes the implementation more complicated and messy. The point about a potential panic is a good one, I can add some more checks against a parsing panic. We should chat more offline on the systemd implications.

reset:
@adamhjk @fnichol @chefsalim I need the `builder-core` crate that @chefsalim has created in this PR for the PR that I'm working on for GitHub build notifications. If nobody objects, I'd like to merge this in but we can continue the code review in this thread and then issue a new PR with the changes to address.

Everyone alright with that?

reset:
@thesentinels approve

gif-keyboard-7713248155969918795

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/191734267).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/191734267).

PR #1621, opened 2017-01-05
[bldr] Push common args into their subcommands to prevent mis-invoking.

fnichol:
This change clears up a usability issue that I stumbled upon myself,
initially in `core/bldr-api`. I wanted to use a custom config and ran
the following:

    bldr-api --config /path/to/config.toml start

To my confusion, no matter what I did, my config was not getting picked
up. Upon re-reading the code and double checking against the clap API, I
learned that the `global` attribute on an argument will leave the arg
match at the level it was provided by the user (and not also copied down
to the subcommand). This means that both of these invocations result in
a started service:

    bldr-api --config /path/to/config.toml start
    bldr-api start --config /path/to/config.toml

The catch is that only the second one actually applies the config. Oi.

This fix provided here drops the `global` attribute on the `Arg` and
pushes them down into the `start` subcommand. Currently all the Builder
services only provide a `start` subcommand but in the future additional
subcommands may be added. At this point there is some `start` hardcoding
that needs to be generalized anyway and any copious arg duplication can
be dealt with then.

The new behavior is as follows:

```sh
> bldr-api start --config /path/to/config.toml
```

```sh
> bldr-api --config /path/to/config.toml start
error: Found argument '--config' which wasn't expected, or isn't valid in this context

USAGE:
    bldr-api [SUBCOMMAND]

For more information try --help
```

### DIFF
--- a/components/builder-admin/src/main.rs
+++ b/components/builder-admin/src/main.rs
@@ -49,11 +49,11 @@ fn app<'a, 'b>() -> clap::App<'a, 'b> {
         (about: "Habitat builder-admin")
         (@setting VersionlessSubcommands)
         (@setting SubcommandRequiredElseHelp)
-        (@arg config: -c --config +takes_value +global
-            "Filepath to configuration file. [default: /hab/svc/hab-builder-admin/config.toml]")
-        (@arg port: --port +takes_value +global "Listen port. [default: 8080]")
         (@subcommand start =>
             (about: "Run the builder-admin server")
+            (@arg config: -c --config +takes_value
+                "Filepath to configuration file. [default: /hab/svc/hab-builder-admin/config.toml]")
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

adamhjk:
This looks good - as always, your PR explanations are legendary.

Lame that it doesn't work the way one would expect with the `global` setting.

gif-keyboard-5672741902532227213

@thesentinels approve

thesentinels:

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/189362981).

fnichol:
gif-keyboard-14819133391177303892

PR #1629, opened 2017-01-09
Enhancements to `hab plan init`

smith:
* Add all of the variables listed in the docs
* Make the variable be listed in the same order as they are in the docs
  (this is probably the wrong order for what we want in the end, but
  it's now consistent between the two and can be argued about later.)
* Allow all "pkg_*" variables (except name and origin) to be set by environment
  variables. So now you can type:

```
    env pkg_svc_user=me pkg_deps="(core/foo core/bar)" \
      pkg_license="('MIT' 'Apache-2.0')" pkg_bin_dirs="(bin sbin)" \
      pkg_version=1.0.0 pkg_description="foo" pkg_maintainer="you" \
      hab plan init foo
```

and it does what you would want.

Signed-off-by: Nathan L Smith <smith@chef.io>

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -226,7 +226,9 @@ pub fn get() -> App<'static, 'static> {
             (@subcommand init =>
                 (about: "Generates common package specific configuration files. Executing without \
                     argument will create a `habitat` directory in your current folder for the plan. \
-                    If `PKG_NAME` is specified it will create a folder with that name.")
+                    If `PKG_NAME` is specified it will create a folder with that name. \
+                    Environment variables (those starting with 'pkg_') that are set will be used \
+                    in the generated plan")
                 (aliases: &["i", "in", "ini"])
                 (@arg PKG_NAME: +takes_value "Name for the new app.")
                 (@arg ORIGIN: --origin -o +takes_value "Origin for the new app")
--- a/components/hab/src/command/plan.rs
+++ b/components/hab/src/command/plan.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 pub mod init {
+    use std::env;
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

mwrock:
This looks good, I'm just curious about the value of the environment variable use case. I'm not saying its not valuable, but it just seems to me just as simple to edit the template than to set the environment variables. So I'd just like to hear the motivation and put this out there to protect against "feature bloat."

smith:
> I'm not saying its not valuable, but it just seems to me just as simple to edit the template than to set the environment variables. So I'd just like to hear the motivation and put this out there to protect against "feature bloat."

It seemed to me that this is how I would expect it to work.

I saw @fnichol using https://github.com/jpmens/jo and liked the way that works. With jo, you could also say it's just as easy to edit a blank file and start writing JSON, which might be true, but having an obvious way to pre-populate the values is even nicer.

fnichol:
That is rather cute :smile: I could see some interesting potential for one-off package builds and possibly some CI creation tasks.

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/190341121).

mwrock:
Yeah I was thinking more on this and could definitely see the value for auto generating plans.

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/190341121).

reset:
@smith great additions here, thank you! I can especially see the value in setting the generated values of a plan by env var when using an automated process to generate plans

PR #1630, opened 2017-01-09
Added sig key checking to core crypto

georgemarshall:
This is in response to a public key that I uploaded to the depot that was a bad copy/paste with invalid chars. The goal is to do a basic check to make sure the key will not break things.

### DIFF
--- a/components/core/src/crypto/keys/sig_key_pair.rs
+++ b/components/core/src/crypto/keys/sig_key_pair.rs
@@ -15,7 +15,7 @@
 use std::fs;
 use std::path::{Path, PathBuf};
 
-use rustc_serialize::base64::{STANDARD, ToBase64};
+use rustc_serialize::base64::{STANDARD, FromBase64, ToBase64};
 use rustc_serialize::hex::ToHex;
 use sodiumoxide::crypto::sign;
 use sodiumoxide::crypto::sign::ed25519::SecretKey as SigSecretKey;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

mwrock:
This looks really great! My only comment is that it looks like there is a pattern here where the number in the error signifies a "type" of error. I'm deferring to @reset  or @fnichol to confirm this is indeed the pattern, but if so, perhaps the error string should be `write_sig_key_from_str:4 Malformed sig key string`.

georgemarshall:
From my gathering the number seems more like a "section" reference than an error "type".

mwrock:
Right but with the aim to make it clear what logic branch caused the error so having a separate number here I think it would be faster to deduce the nature of the error.

reset:
@mwrock @georgemarshall this isn't a "NetError" so the formatting of the message body doesn't need to be used here - you can use whatever message is helpful to send back to the user.

The formatting that @mwrock is talking about is in the form `{service}:{operation}:{case}`. Service is the service or module which the error occured in. Operation is what was happening at the time. Case is the branch of code that caused the error. This is a static message (no string concatentation or injection) and is meant to be used at a later date to build up a knowledge base for self help and to assist in localizing clients.

reset:
@thesentinels approve

gif-keyboard-12597638913184257927

Thanks for catching this, very sorry for the problem it caused you!

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/190379457).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/190379457).

PR #1634, opened 2017-01-13
Resolve peer address before assigning it's IP to member

elliott-davis:
The previous behavour of a peer binding would not resolve a dns entry before trying to unwrap a string. This caused a thread panic that would kill the supervisor.
The new behavour will try and resolve the ip's for dns entries and die if it is unable to do so.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -17,7 +17,7 @@ pub mod service;
 pub mod signals;
 pub mod service_updater;
 
-use std::net::SocketAddr;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::thread;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


mwrock:
Awesome like a possom!

Rerunning travis and appveyor whose breaks look unrelated.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/191710918).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/191710918).

components/sup/src/manager/mod.rs
```diff
@@ -87,7 +87,10 @@ impl Manager {
                                                  None));
         outputln!("Butterfly Member ID {}", server.member_id());
         for peer_addr in gconfig().gossip_peer() {
-            let addr: SocketAddr = peer_addr.parse().unwrap();
+            let addrs: Vec<SocketAddr> = peer_addr.to_socket_addrs()
+                                                  .expect("Unable to resolve domain")
```
adamhjk: This function returns a `Result` - we should add an error for failure to resolve, and return an error, so that the tooling can handle things like pretty printing the error.

mwrock: Doesn't the `expect` mostly take care of this? It conveys how the input failed and if it does fail, I don't see harm in the panic as long as its made clear to the user why.

adamhjk: The application knows how to exit cleanly with an error message, with pretty colors and details about precisely where the error was generated in the source. Panic makes sure you can't do that.

Also, this gives optionality to the caller - if you wanted to panic, you easily can.

PR #1667, opened 2017-01-24
Add Package Exports to build and runtime systems

reset:
This commit introduces a new concept for Habitat packages to make
a subset of service configuration data available to consumers through
the `--bind` runtime flag of a service.

A package which declares one or more exports is known as a "producer"
and the service using a `--bind` runtime flag to connect to a service
group running the producing package is known as the "consumer".

A package declares an export with the newly created `pkg_exports` plan
variable. This is an associative array which maps the services
configuration into a public key space. This public key space acts as
a promise, or contract, between the producer and consumer to mitigate
changes in configuration structure between updates.

The public configuration of a bind can be accessed within a newly
created `.cfg` accessor in a member of a bind from within any of our
rendered templates.

* Add `pkg_exports` to plan syntax
* Add EXPORTS metadata file to `.hart` files
* Update service rumor to include *exported* configuration data
* Update census entry to contain configuration data from a service
  rumor
* Update pkg plan generator template to include `pkg_exports` stub
* Additional refactors for readability, memory efficiency, and log
  usability

gif-keyboard-9971416907674337106


thesentinels:

Thank you for contributing!

components/butterfly/src/message/mod.rs
```diff
@@ -108,7 +109,7 @@ impl Serialize for swim::Rumor {
 
 impl Serialize for swim::Service {
     fn serialize<S>(&self, serializer: &mut S) -> result::Result<(), S::Error> where S: Serializer {
-        let mut state = try!(serializer.serialize_struct("service", 9));
+        let mut state = try!(serializer.serialize_struct("service", 10));
```
jmink: Totally ignorable nitpick: Does it make sense to make 10 a constant?

components/butterfly/src/rumor/service_config.rs
```diff
@@ -102,16 +104,23 @@ impl ServiceConfig {
...
+            let encoded = try!(str::from_utf8(self.get_config())
+                .map_err(|e| Error::ServiceConfigNotUtf8(self.get_service_group().to_string(), e)));
+            try!(self.parse_config(&encoded))
+        };
+        Ok(config)
+    }
+
+    fn parse_config(&self, encoded: &str) -> Result<toml::Value> {
+        toml::Value::from_str(encoded)
+            .map_err(|e| Error::ServiceConfigDecode(self.get_service_group().to_string(), e))
```
jmink: Feel free to ignore manager meddling: This just hit the rule of three.  Would a helper function be useful here?

components/butterfly/src/server/mod.rs
```diff
@@ -725,21 +726,18 @@ impl Server {
...
-                            Err(e) => {
-                                warn!("Cannot decrypt service config for {}: {}", service_group, e)
-                            }
-                        }
+        self.service_config_store.with_rumor(service_group, "service_config", |maybe_sc| {
+            if let Some(sc) = maybe_sc {
+                if incarnation.is_none() || sc.get_incarnation() > incarnation.unwrap() {
+                    match sc.config() {
+                        Ok(config) => result = Some((sc.get_incarnation(), config)),
+                        Err(err) => warn!("{}", err),
```
jmink: Feel free to ignore manager meddling: What's the logic behind making the error message less verbose?

components/butterfly/src/rumor/service_config.rs
reset: This is the helper function :). We call this twice above

components/butterfly/src/message/mod.rs
reset: I wouldn't make any of these constants. They're only used to determine the amount of elements we wish to serialize and the lib only requires them to know how much to allocate

components/butterfly/src/server/mod.rs
reset: The error returned has a much better message associated with it now. In Rust, this syntax will call the fmt trait implementation of the thing it's replacing. So that's where the improved error is. We don't need any additional context anymore here

components/butterfly/src/rumor/service_config.rs
jmink: Awesome.  Carry on then.

components/butterfly/src/server/mod.rs
jmink: Oh cool.  That's awesome!

components/butterfly/src/error.rs
```diff
@@ -43,29 +45,31 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let msg = match *self {
-            Error::BadMessage(ref err) => format!("Bad Message: {:?}", err),
-            Error::CannotBind(ref err) => format!("Cannot bind to port: {:?}", err),
+            Error::BadMessage(ref err) => format!("Bad Message, {:?}", err),
```
adamhjk: I use colon for that *everywhere* in this code base, and I think `:` is more appropriate than a comma, since you're not completing a sentence. I'm a 2/10 about how much it matters, but we should decide, and my preference is colon's not comma's.

components/butterfly/src/rumor/mod.rs
```diff
@@ -28,6 +28,11 @@ pub mod service;
 pub mod service_config;
 pub mod service_file;
 
+pub use self::election::Election;
```
adamhjk: I find this kind of re-exporting confusing, because my imports no longer map to where I can explore the source on disk. I'm fine with it, and it's a common enough pattern, but I think the verbose mechanism is better.

components/core/src/package/install.rs
```diff
@@ -22,9 +22,9 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 pub use types::package_install::*;
+use super::{Identifiable, MetaFile, PackageIdent, Target, PackageTarget};
```
adamhjk: Using super here feels less clear than the former path, which used `package`. Why the change?

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -8,7 +8,7 @@ param (
 
 # # License and Copyright
 # ```
-# Copyright: Copyright (c) 2015 Chef Software, Inc.
+# Copyright: Copyright (c) 2016-2017 Chef Software, Inc.
```
adamhjk: The copyright notice should be the year the work first appeared. So this should just become

`Copyright (c) 2017`

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -177,6 +177,18 @@
 # pkg_expose=(80 443)
 # ```
 #
+# ### pkg_exports
+# An associative array representing configuration data which should be gossiped to peers. The keys
+# in this array represent the name the value will be assigned and the values represent the toml path
+# to read the value.
+# ```
+# pkg_exports=(
+#   [key]=token
```
adamhjk: What happens if we want to export something that's in the `sys` namespace, or the `pkg` namespace? The examples have unqualified names.

For example, the IP doesn't show up in `cfg`, it shows up in `sys`.

components/sup/src/manager/census.rs
```diff
@@ -14,37 +14,37 @@
...
 use std::collections::HashMap;
-use std::str::FromStr;
+use std::str::{self, FromStr};
 
 use hcore::package::PackageIdent;
 use hcore::service::ServiceGroup;
-use butterfly::rumor::service::Service as ServiceRumor;
-use butterfly::rumor::election::{Election as ElectionRumor, Election_Status};
+use butterfly;
+use butterfly::rumor::{Service as ServiceRumor, Election as ElectionRumor};
```
adamhjk: Yeah, I see those imports coming back. I think this is worse, not better, even though the above is more terse. It maps to the actual locations on disk - with this indirection, you have to go code spelunking, discover the re-export, and *then* find it on disk. I'm fine if its the pattern we want to follow, but I'm :-1: personally.

components/sup/src/manager/census.rs
```diff
@@ -301,12 +301,25 @@ impl CensusEntry {
...
+        self.set_ip(String::from(rumor.get_ip()));
+        self.set_hostname(String::from(rumor.get_hostname()));
+        self.set_port(format!("{}", rumor.get_port()));
+        self.set_exposes(rumor.get_exposes().iter().map(|p| format!("{}", p)).collect());
+        // JW TODO: we can't crash the supervisor because we've received bad data. This could have
+        // gotten to us from a supervisor speaking a slightly different protocol or from a
+        // malicious service speaking our gossip protocol. We should consider ignoring this update
+        // altogether perhaps? We can't just unload the service because that would also give an
+        // attacker what they want.
+        self.set_package_ident(PackageIdent::from_str(rumor.get_package_ident())
```
adamhjk: Yep. I think we should ignore the update.

components/sup/src/util/toml.rs
```diff
@@ -0,0 +1,228 @@
+// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
```
adamhjk: This should just be `Copyright (c) 2016` - only ever need the date work first appeared.

adamhjk: Or 2017, since this is a new file

components/butterfly/src/error.rs
reset: Ah yeah, I should have made a comment about this - totally forgot about the change.

All of our servers log with the format of "msg, {errKey}={errMsg}". I do that on the servers so we can later easily setup patterns to match on in a log aggregator for things we care about being alerted on or just for ease of debugging.

We don't necessarily need to standardize on it here since these messages aren't going into a log aggregator. I just saw these and thought "oh that's a server and that's how the logs are supposed to be formatted".

Additionally, we can standardize on commas or colons or any other separator in the servers if you have a preference!

I'll change these back

components/butterfly/src/rumor/mod.rs
reset: The exports still exist on disk, this is common in larger Rust code bases because it helps guide people through the public API of your modules. It allows you to place things in different files for logical organization as a developer but also exposing the module as a single logical unit for the public API consumer.

These are re-exported here because when you're dealing with Butterfly rumors as an API consumer you don't necessarily care that they're in their own sub modules, you just care about the 4 types of rumors and want to access them. It makes for less lines on import.

You'll also see re-exporting commonly used when your API asks for people to use a number of public types to interact with it in the form of a `prelude` module. I was actually one step away from including that until I noticed we were aliasing the rumors on import to avoid a namespace collision ;)

components/core/src/package/install.rs
reset: I think it's more appropriate to use `super` and `self` because it makes the module more self contained. It can be picked up and moved anywhere else in the codebase without worrying about where on disk it resides and which modules surround it

components/plan-build-ps1/bin/hab-plan-build.ps1
reset: Got it! After all these years I didn't actually know how that worked ;)

components/plan-build/bin/hab-plan-build.sh
reset: Good question! I totally forgot to follow up on this with you after our discussion before I got in and implemented it.

My thinking was that we could restructure the keyspace of the bind to look more similar to that of the `ServiceConfig`, but with a smaller surface. In this PR I began to flatten the namespace by putting all of the exported configuration in `bind.{member}.cfg`. I was thinking that we'd also probably want to gossip the contents of our `ServiceConfig.sys` as well and that'd become `bind.{member}.sys`. I don't think we need to actually perform the mapping here because the contents of `ServiceConfig.sys` is both small and already structured for us. Every peer will have the same sys.

I can see us also wanting to export a small portion of the `Pkg` struct as `bind.{member}.pkg` as well. I don't think we need to share all of it - for example, the consumer doesn't care about the producer's exports, path, or really any of the in the svc_* fields. They're more of a private member and only usable by the producer's side.

After writing this out to you I think I'm going to go ahead and make these changes in this PR and we should continue the discussion about it before merging!

components/sup/src/manager/census.rs
reset: If we were (and we should soon!) generating docs for Butterfly the value of this might be more apparent. The generated docs would show the same content regardless of the module path that you searched for the rumor struct on.

Here's a great example in serde. When reading the docs you can check out the [re-exports at the top of the crate](https://docs.serde.rs/serde/#reexports). It makes the API surface clear to the reader without asking them to read each submodule. Your docs are also easier to write because you don't need to write public documentation for each submodule, instead you can write it at the container module level and explain to people how to use your API

components/sup/src/manager/census.rs
reset: So we should just not update our census entry, yeah? That's what I was thinking but I wanted to make sure

reset:
@adamhjk here's a sample of the rendered bind now

screen shot 2017-01-27 at 8 35 22 am

I've added a `sys` and `pkg` section to go alongside the `cfg` section. `sys` only has 4 fields which all make sense to make public and half of their data is duplicated for now on the bind as `ip` and `hostname`.

I looked at `pkg` and I think really just these four parts are what you care about as a consumer. We actually already had this on the bind, it was just named as `package_ident` which isn't inline with our convention.

Another PR will come to remove `pkg_exposes` which will help clean some of this up. I wanted to make that a separate PR because it's going to be a sizable change in itself.

adamhjk:
Looks great

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/196057528).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/196057528).

PR #1668, opened 2017-01-24
Add builder-db component

adamhjk:
This commit adds a builder-db component. It serves as the base layer for
using PostgreSQL for persistence for builder services. Currently, it
provides 3 things:

* A simple `Pool`, which governs access to the database
* A `Migrator`, which handles database migrations
* A `test` namespace, which exports macros and functionality neccessary
  to build integration tests for services which use access the database

The top level module documentation includes some advice on design and
usage of the database within builder.

To run the test suite, you must first, in a separate terminal:

```sh
$ mkdir -p /hab/svc/postgresql
$ cp components/builder-db/tests/db/* /hab/svc/postgresql
$ hab start core/postgresql
```

You can then use `cargo test` as normal.

As we gain more experience writing the service components, I expect we
will automate that part of the process - hence, I'm leaving it out of
the formal documentation for now.

### DIFF
--- a/components/builder-db/src/data_store.rs
+++ b/components/builder-db/src/data_store.rs
@@ -1,0 +1,31 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::error;
+use std::convert::From;
+
+use pool::Pool;
+use migration::Migrator;
+use error::Error;
+
+pub trait DataStore<E: From<Error>> {
+    fn migrator(&self) -> &Migrator;
+
+    fn pool(&self) -> &Pool;
+
+    fn setup(&self) -> Result<(), E> {
+        let result = self.migrator().setup()?;
+        Ok(result)
+    }
+}
--- a/components/builder-db/src/error.rs
+++ b/components/builder-db/src/error.rs
@@ -1,0 +1,93 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::error;
+use std::fmt;
+use std::result;
+
+use r2d2;
+use postgres;
+
+#[derive(Debug)]
+pub enum Error {
+    ConnectionTimeout(r2d2::GetTimeout),
+    FunctionCreate(postgres::error::Error),
+    FunctionRun(postgres::error::Error),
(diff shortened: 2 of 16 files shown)


thesentinels:

Thank you for contributing!

components/builder-db/src/data_store.rs
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+use std::error;
+use std::convert::From;
+
+use pool::Pool;
+use migration::Migrator;
+use error::Error;
+
+pub trait DataStore<E: From<Error>> {
```
reset: This would be a great place to use an [associated type](https://doc.rust-lang.org/book/associated-types.html) instead of a standard generic for the `Error` type

components/builder-db/src/error.rs
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
```
reset: Excellent work with how the errors are being handled in this crate so far. We've come a long way :)

components/builder-db/src/pool.rs
```diff
@@ -0,0 +1,94 @@
...
+impl DerefMut for Pool {
+    fn deref_mut(&mut self) -> &mut r2d2::Pool<PostgresConnectionManager> {
+        &mut self.inner
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct TestConnectionCustomizer;
+
+impl r2d2::CustomizeConnection<postgres::Connection, r2d2_postgres::Error> for TestConnectionCustomizer {
```
reset: This is really slick

components/builder-db/src/pool.rs
```diff
@@ -0,0 +1,94 @@
...
+// We will use this to allocate test schmeas
+static GLOBAL_SCHEMA_COUNT: AtomicUsize = ATOMIC_USIZE_INIT;
+
+#[derive(Debug, Clone)]
+pub struct Pool {
+    inner: r2d2::Pool<PostgresConnectionManager>,
+}
+
+impl Pool {
+    pub fn new(connection_url: &str,
```
reset: A constructor that sleeps the calling thread is a little awkward. I might suggest moving this off to a "start" function or something along those lines

reset:
@adamhjk really great start so far!

components/builder-db/src/data_store.rs
adamhjk: Turns out I think we don't need this trait at all. I'm leaving it here for now because it's working, but I think in the end we won't implement it.

components/builder-db/src/error.rs
adamhjk: right? felt better.

components/builder-db/src/pool.rs
adamhjk: Thanks!

components/builder-db/src/pool.rs
adamhjk: Sure - easy enough to do.

adamhjk:
@reset can you look at these appveyor errors? they look like they are serde related.

reset:
@adamhjk I think there's something up with caching in appveyor there. It's not able to see the API changes we implemented for in the Regex crate, either. I'm not sure how to blow the cache in appveryor. You've rebased against master for this PR, yeah? I also recently updated us to Rust 1.14 in the appveyor config on master

adamhjk:
Yep, it's rebased on master

components/builder-db/src/pool.rs
adamhjk: This is actually not easy to do, I take it back. We would need to wrap the inner section in an Option to make this work, since we need the pool to be persistent, and the pool object itself will sleep. We could move the retry logic out if we wanted, but you're still going to sleep the thread. Ie: you wouldn't ever call `Pool.new()` anymore, because it literally can't return from an error - so you'll call `Pool.start()` all the time, which I'm not sure is better.

adamhjk:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/195708691).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/195708691).

Issue #1280, opened 2016-09-25
Spammy build output


chetan:
#1233 considerably improves the output of running `hab pkg build ...` by quieting the `hab pkg install` output during studio setup, but once the build actually starts, packages installed by `hab-plan-build.sh` output progress bars again. I've found the fix is to modify/override the following method:

``` bash
_install_dependency() {
  if [[ -z "${NO_INSTALL_DEPS:-}" ]]; then
    $HAB_BIN install -u $HAB_DEPOT_URL "$dep" || true
  fi
  return 0
}
```

from https://github.com/habitat-sh/habitat/blob/master/components/plan-build/bin/hab-plan-build.sh#L616-L621

the relevant line becomes:

```
    ($HAB_BIN install -u $HAB_DEPOT_URL "$dep" | cat -) || true
```

This line also needs `--verbose` removed:

```
 $_xz_cmd --compress -6 --threads=0 --verbose $tarf
```

https://github.com/habitat-sh/habitat/blob/master/components/plan-build/bin/hab-plan-build.sh#L2185

While this does work, there should probably be a better way of passing the correct behavior into this script. Perhaps something like wget's `--no-verbose`?

chetan:
Did some more digging into this and discovered the following - 

* `hab pkg build` initially quiets output by requiring `isatty` on stdin/stdout/stderr
* however once inside the `chroot`ed env, I observe the following

```
$ hab pkg build . 2>&1 | cat - # redir/pipe to remove tty

$ ps auxwwwf | grep ..
root     15565  0.1  0.0  25176  3792 /hab/studios/testing/dev/pts/10 Ss+ 17:03   0:00          |               |               \_ /hab/pkgs/core/bash/4.3.42/20161213234235/bin/bash /hab/pkgs/core/hab-plan-build/0.15.0/20161222203825/bin/hab-plan-build .

$ ls -al /proc/15565/fd/
total 0
dr-x------ 2 root root  0 Jan 18 17:04 .
dr-xr-xr-x 9 root root  0 Jan 18 17:03 ..
lrwx------ 1 root root 64 Jan 18 17:04 0 -> /hab/studios/testing/dev/pts/10
lrwx------ 1 root root 64 Jan 18 17:04 1 -> /hab/studios/testing/dev/pts/10
lrwx------ 1 root root 64 Jan 18 17:04 2 -> /hab/studios/testing/dev/pts/10
```

[...]

PR #1682, opened 2017-01-27
[hab,studio] Add noninteractive & no coloring modes via environment variables.

fnichol:
The primary use case driving this feature is running builds in a CI environment. Some logging streams do not cope well with progress bar style output which behave by using carriage returns to the terminal to set its cursor back to the start of the line and "redraw" the current line. These logging streams frequently start outputting each progress tick/redraw as a full new line and very quickly consume a large amount of output. In systems such as Travis there is a upper bound to the number of log lines provided back to the user. As we care about the build output not spinners of doom, we needed a way to turn these off.

This change adds a way for users of Habitat tooling (such as `hab`, `hab pkg build`, `hab studio`, etc.) to explicitly drop single-line progress bar indicators or to disable text coloring by setting 2 new environment variables. As this feature is targeted for CI environments, environment variables were used so that the command issued in CI is virtually identical to the command issued on a developer's workstation. Additionally, running a `hab pkg build` will invoke a `hab` process which invokes a `hab-studio` process which may contain further `hab` subprocess before completing. The amount of work required to only support explicit flags on all the related tools would have made this feature much more complex and error prone. The choice of prefixing these environment variables with `HAB_` ensures that there's almost no chance these variables would affect the build of any software and so can be safely passed over the Studio's barrier of isolation.

Currently, `hab`'s  UI system will check 2 things to determine its behavior:

* Coloring is determined by checking the `TERM` environment variable to check the local terminfo database. If the `TERM` supports color, then `hab` will add ansi coloring.
* Progress bars is determined by checking if the input and output streams (i.e. stdin, stdout, stderr) have a TTY assigned (using `libc::isatty()` or equivalent depending in platform).

This change checks an environment variable `HAB_NONINTERACTIVE` which if set to `"true"`, will override the default detection logic and tell the UI that it does *not* have a TTY assigned. As a consequence, downloading and uploading progress bars will be disabled.

Additionally, another environment variable `HAB_NOCOLORING` is checked, and if set to `"true"` will disable all coloring from `hab`, `hab-studio`, `hab-plan-build`, etc.

tl;dr
-----

To skip spinners in CI either:

    env HAB_NONINTERACTIVE=true hab pkg build ...

or

    export HAB_NONINTERACTIVE=true
    hab pkg build ...

References #1280 

gif-keyboard-4359273550099528466

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -83,15 +83,20 @@ SUBCOMMANDS:
     version   Prints version information
 
 ENVIRONMENT VARIABLES:
-    HAB_ORIGIN        Propagates this variable into any studios
-    HAB_ORIGIN_KEYS   Installs secret keys (\`-k' option overrides)
-    HAB_STUDIOS_HOME  Sets a home path for all Studios (default: /hab/studios)
-    HAB_STUDIO_ROOT   Sets a Studio root (\`-r' option overrides)
-    NO_SRC_PATH       If set, do not mount source path (\`-n' flag overrides)
-    QUIET             Prints less output (\`-q' flag overrides)
-    SRC_PATH          Sets the source path (\`-s' option overrides)
-    STUDIO_TYPE       Sets a Studio type when creating (\`-t' option overrides)
-    VERBOSE           Prints more verbose output (\`-v' flag overrides)
+    HAB_NOCOLORING      Disables text coloring mode despite TERM capabilities
+    HAB_NONINTERACTIVE  Disables interactive progress bars despite tty
+    HAB_ORIGIN          Propagates this variable into any studios
+    HAB_ORIGIN_KEYS     Installs secret keys (\`-k' option overrides)
+    HAB_STUDIOS_HOME    Sets a home path for all Studios (default: /hab/studios)
+    HAB_STUDIO_ROOT     Sets a Studio root (\`-r' option overrides)
+    NO_SRC_PATH         If set, do not mount source path (\`-n' flag overrides)
+    QUIET               Prints less output (\`-q' flag overrides)
+    SRC_PATH            Sets the source path (\`-s' option overrides)
+    STUDIO_TYPE         Sets a Studio type when creating (\`-t' option overrides)
+    VERBOSE             Prints more verbose output (\`-v' flag overrides)
+    http_proxy          Sets an http_proxy environment variable inside the Studio
+    https_proxy         Sets an https_proxy environment variable inside the Studio
+    no_proxy            Sets a no_proxy environment variable inside the Studio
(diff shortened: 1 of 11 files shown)


components/studio/bin/hab-studio.sh
```diff
@@ -83,15 +83,20 @@ SUBCOMMANDS:
...
+    HAB_STUDIOS_HOME    Sets a home path for all Studios (default: /hab/studios)
+    HAB_STUDIO_ROOT     Sets a Studio root (\`-r' option overrides)
+    NO_SRC_PATH         If set, do not mount source path (\`-n' flag overrides)
+    QUIET               Prints less output (\`-q' flag overrides)
+    SRC_PATH            Sets the source path (\`-s' option overrides)
+    STUDIO_TYPE         Sets a Studio type when creating (\`-t' option overrides)
+    VERBOSE             Prints more verbose output (\`-v' flag overrides)
+    http_proxy          Sets an http_proxy environment variable inside the Studio
+    https_proxy         Sets an https_proxy environment variable inside the Studio
+    no_proxy            Sets a no_proxy environment variable inside the Studio
```
mwrock: Great to see complete env var docs! We should update https://github.com/habitat-sh/habitat/blob/master/www/source/docs/reference/environment-vars.html.md as well.

mwrock:
This is super great. I'd merge but think its important not to forget about the web docs.

fnichol:
@mwrock Good call--I updated the docs and rebased. In all honesty it was a fluke that I found our `hab-studio` usage output and realized we were missing a couple entries. All better now!

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/197069080).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/197069080).

bixu:
@fnichol / @eeyun, I think this is still broken-feeling. When I set `HAB_NOCOLORING=true` in my shell, I still see colored text in my Studio prompts. For me specifically this is a style issue (I prefer black-on-white text in my terminal for eye strain reasons), but I can also imaging this being an actual accessibility issue as well as problematic for certain types of automation.

Also a personal preference: I feel fairly strongly that a low-level tool like Habitat should have color turned OFF by default and only be toggled on by a user config/env var.  <oldmanrant>Colored output has been a constant source of annoyance for me and my colleagues (for various reasons) since we adopted Habitat.</oldmanrant>
https://s3media.247sports.com/Uploads/Assets/762/405/3405762.gif

Issue #1194, opened 2016-08-26
Add dedicated content about GitHub scopes needed in CLI setup


davidwrede:
Per #1189, we should have a dedicated section or topic that describes what GitHub scopes or other authentication mechanisms we require as part of both setup and/or general workflow when using the `hab` CLI to upload and download keys, packages, etc.

smurawski:
@davidwrede just a quick ping to bring this back to your attention.

davidwrede:
Thanks @smurawski for the ping. I'm pivoting over to Habitat work this week and next. I'll add this to my queue.

PR #1710, opened 2017-02-03
Updated overview page + added a bit more info on OAuth scopes

davidwrede:
In response to #1189 and #1194, I've updated the notes in the tutorial to point people to an existing section in our docs that talks about OAuth scopes in the context of authenticating with the depot (which is the only place they seem to be used). If/when new scopes are required, we can break the scopes content out into a new topic.

Also updated the overview page to more accurately reflect the current nav structure of the docs.


thesentinels:

Thank you for contributing!

reset:
@davidwrede good additional clarification here, thank you!

PR #1712, opened 2017-02-04
 Rename pkg_expose to pkg_exposes and refactor to leverage pkg_exports

reset:
This changes the accepted values for pkg_expose from a numeric literal
to a string representing a key in pkg_exports containing a valid port
number. The contents of EXPOSES will still contain the numeric
literal so this is not a breaking change for the Habitat runtime,
just a breaking change for hab-plan-build users.

* Rename pkg_expose to pkg_exposes
* A numeric literal is no longer a valid pkg_expose
* The default value of the pkg_export must be a valid port number
  or package generation will halt
* Package generation will fail if the pkg_export is missing

gif-keyboard-601488567894941889

### DIFF
--- a/components/builder-api/habitat/default.toml
+++ b/components/builder-api/habitat/default.toml
@@ -1,3 +1,6 @@
+[srv]
+port = 9636
+
 [ui]
 app_url          = "https://willem.habitat.sh/v1"
 community_url    = "https://www.habitat.sh/community"
--- a/components/builder-api/habitat/plan.sh
+++ b/components/builder-api/habitat/plan.sh
@@ -9,7 +9,10 @@ pkg_deps=(core/glibc core/openssl core/coreutils core/gcc-libs core/zeromq core/
   core/libarchive)
 pkg_build_deps=(core/protobuf core/protobuf-rust core/coreutils core/cacerts
   core/rust core/gcc core/git core/pkg-config core/node core/phantomjs)
-pkg_expose=(9636)
+pkg_exports=(
+  [port]=srv.port
+)
+pkg_exposes=(port)
 bin="bldr-api"
 pkg_svc_run="$bin start -c ${pkg_svc_path}/config.toml"
 
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -171,12 +171,6 @@
 # pkg_svc_run="bin/haproxy -f $pkg_svc_config_path/haproxy.conf"
 # ```
 #
-# ### pkg_expose
-# An array of ports this service exposes to the world.
-# ```
-# pkg_expose=(80 443)
-# ```
-#
 # ### pkg_exports
 # An associative array representing configuration data which should be gossiped to peers. The keys
 # in this array represent the name the value will be assigned and the values represent the toml path
@@ -187,6 +181,14 @@
 #   [port]=server.port
 #   [host]=server.host
 # )
+# ```
+#
+# ### pkg_exposes
+# An array of `pkg_exports` keys containing default values for which ports that this package
+# exposes. These values are used as sensible defaults for other tools. For example, when exporting
+# a package to a container format.
+# ```
+# pkg_exposes=(port)
 # ```
 #
 # ### pkg_origin
@@ -255,7 +257,7 @@
 # pkg_bin_dirs=(bin)
 # pkg_deps=(glibc pcre openssl zlib)
 # pkg_svc_run="bin/haproxy -f $pkg_svc_config_path/haproxy.conf"
-# pkg_expose=(80 443)
+# pkg_exposes=(port)
 # pkg_exports=(
 #   [mode]=mode
 #   [port]=server.port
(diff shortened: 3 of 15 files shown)


thesentinels:

Thank you for contributing!

reset:
We'll need to follow up with a Windows PR. The `rq` tool doesn't have a pre-built Windows binary and we need to be able to parse the `default.toml` and validate the port in `plan-build-ps1`

adamhjk:
In your example where you use srv.port - that would map to cfg.srv.port, yes? Might choose a different example.

reset:
@adamhjk Yeah it would, I wrote that a few times in this update - which example? I can make the examples whatever makes most sense for the reader

components/builder-api/habitat/default.toml
```diff
@@ -1,3 +1,6 @@
+[srv]
+port = 9636
```
smurawski: Is this actually supposed to configure the port that builder-api is running on or is it only a convenience mechanism for `pkg_expose`? (configuring the port would be my impression.)

reset: It's something that should be used to configure the port, but right now we read the `config.toml` that the Supervisor generates directly. We'll need to change it up just a bit soon. I'm going to take a pass on the builder services as soon as the `pkg_bind_required`/`pkg_bind_optional` work is done and this will get picked up in that work

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1215,6 +1215,22 @@ abspath() {
   fi
 }
 
+# Convert a string into a numerical value.
+function to_int() {
+    local -i num="10#${1}"
+    echo "${num}"
+}
+
+# Return 0 if the given value is a valid port and 1 if not.
```
fnichol: Same thing with `port_is_valid()` -> `_port_is_valid()`?

components/plan-build/plan.sh
```diff
@@ -18,6 +18,7 @@ pkg_deps=(
   core/grep
   core/gzip
   core/hab
+  core/rq
```
fnichol: Could you add a check for this command in `hab-plan-build.sh`'s [`_find_system_commands()`](https://github.com/habitat-sh/habitat/blob/2aae1efc68da24389c1accb096b5f5df8395074d/components/plan-build/bin/hab-plan-build.sh#L495-L513) function? This function becomes important on net-new target systems in the future as they are special programs that won't necessarily appear on a vanilla system and are needed for the build program to be well behaved before a Studio environment is possible.

www/source/about/habitat-and-modern-app.html.slim
```diff
@@ -110,7 +110,11 @@ pre
     pkg_deps=(core/glibc chefops/jdk8)
     pkg_bin_dirs=(bin)
     pkg_lib_dirs=(lib)
-    pkg_expose=(9200 9300)
+    pkg_exports=(
+      [http-port]=network.port
```
fnichol: Well I like the way this is expressed, seems much more self-documented!

components/builder-api/habitat/plan.sh
```diff
@@ -9,7 +9,10 @@ pkg_deps=(core/glibc core/openssl core/coreutils core/gcc-libs core/zeromq core/
   core/libarchive)
 pkg_build_deps=(core/protobuf core/protobuf-rust core/coreutils core/cacerts
   core/rust core/gcc core/git core/pkg-config core/node core/phantomjs)
-pkg_expose=(9636)
```
fnichol: Any reason you chose to many this singular (i.e expose) vs plural to match pkg_deps, etc.?

reset: @fnichol it's always been called pkg_expose. Maybe I should rename it now since this is probably the right time to break things anyway ;)

components/builder-api/habitat/default.toml
adamhjk: This line was the exact one I referenced above. I would make it not `srv`, since we have a namespace for that already.

components/builder-api/habitat/plan.sh
adamhjk: Lets rename it.

reset: Will do!

components/builder-api/habitat/default.toml
reset: Oh I see, you just are suggesting to use a different namespace in the toml for our config than `srv`. Can do that for sure. It's not coming to my mind immediately though, where else do we use `srv` in config namespaces?

components/builder-api/habitat/plan.sh
fnichol: Nice! I actually forgot this was the original. Yay for consistency

reset:
Approving this one based on a verbal OK from @adamhjk and @fnichol 

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/199026472).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/199026472).

PR #1724, opened 2017-02-08
Package activity should be properly rooted in a windows studio

mwrock:
This ensures that package handling inside of a windows studio occurs in the studio root directory in `c:\hab\studios\<studio path>\pkgs`. The following are impacted:

1. building packages that depend on other packages will be loaded and, if necessary downloaded, into the studio root.
2. starting services from the supervisor will load them from the studio root
3. The supervisor will load dependent packages from the studio root
4. When starting services, the bin PATHS of the service will be appended to the studio root directory

NOTE: This only affects windows. Linux studios, do not need this and can live in /hab due to the chrooted environment. This code will ignore the FS_ROOT variable for linux except where it already explicitly uses it (ie bintray).

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -22,11 +22,6 @@ use error::Result;
 use super::package::PackageInstall;
 
 
-/// The default filesystem root path
-#[cfg(not(target_os="windows"))]
-pub const FS_ROOT_PATH: &'static str = "/";
-#[cfg(target_os="windows")]
-pub const FS_ROOT_PATH: &'static str = concat!(env!("SYSTEMDRIVE"), "/");
 /// The default root path of the Habitat filesystem
 pub const ROOT_PATH: &'static str = "hab";
 /// The default path for any analytics related files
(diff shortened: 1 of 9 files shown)


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -41,10 +36,24 @@ pub const CACHE_SRC_PATH: &'static str = "hab/cache/src";
...
+/// Using this variable could lead to broken supervisor services and it should
+/// be used with extreme caution.
+pub const FS_ROOT_ENVVAR: &'static str = "FS_ROOT";
 /// The root path containing all runtime service directories and files
 const SVC_PATH: &'static str = "hab/svc";
 
 lazy_static! {
+    /// The default filesystem root path
+    pub static ref FS_ROOT_PATH: String = {
+        if cfg!(target_os = "windows") && henv::var(FS_ROOT_ENVVAR).is_ok() {
```
fnichol: I'm wondering what this would look like as 2 `pub static ref` statements with `#[cfg(…)]` attributes. The reason here is that I'm worried about the platform-dependant code getting tucked too far in and hidden. I realize that the 2 statements would look similar but am hopeful that it'll be super clear to any future development.

components/sup/src/command/start.rs
```diff
@@ -84,7 +84,7 @@ pub fn package() -> Result<()> {
         return Err(sup_error!(Error::RootRequired));
     }
 
-    match Package::load(gconfig().package(), None) {
+    match Package::load(gconfig().package(), Some(&*FS_ROOT_PATH)) {
```
fnichol: I traced this through down to `core::package::PackageInstall::resolve_package_install()` and we look okay here. Long term we might want to consider changing the `Option<&Path>` to a `&Path` but this is more than fine right now.

components/core/src/fs.rs
mwrock: That was my initial approach @fnichol but the compiler was seeing them as duplicate declarations and I am guessing because `lazy_static!` is a macro and something funky was happening by nesting those macros. Here is the error:
```
error[E0428]: a value named `FS_ROOT_PATH` has already been defined in this module
   --> src\fs.rs:47:1
    |
47  | lazy_static! {
    | ^
    | |
    | previous definition of `FS_ROOT_PATH` here
    | already defined
    |
    = note: this error originates in a macro outside of the current crate

error[E0119]: conflicting implementations of trait `std::ops::Deref` for type `fs::FS_ROOT_PATH`:
   --> src\fs.rs:47:1
    |
47  | lazy_static! {
    | ^
    | |
    | first implementation here
    | conflicting implementation for `fs::FS_ROOT_PATH`
    |
    = note: this error originates in a macro outside of the current crate

error: Could not compile `habitat_core`.
```
So I went with the above implementation instead.

fnichol: All right, I would have done the same 😄

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/199763829).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/199763829).

PR #1734, opened 2017-02-09
Refactor Supervisor main loop to tick over each Service on each tick

reset:
This refactor moves much of the logic of the manager's main loop into
a `tick()` function on the Service struct. This function is designed
to return quickly and encapsulate tending to a service on each tick
of the manager. Creating this abstraction allows us to better
represent the execution flow of the supervisor. The Service struct
now owns everything it needs to tend to itself on instruction by
the manager.

Additional changes and improvements:

* Remove `config_from` from global supervisor config. This is now
  part of the `ServiceSpec`/`Service` module
* Better validation of parsing Bindings
* Removed `habitat_sup::package::Package` in favor of
  `habitat_core::package::PackageInstall`
* Improve ergonomics of a number of APIs dealing with file paths
  * Use `AsRef<Path>` over `&Path` where possible
  * Use `PathBuf` over `String` where possible
* Move Supervisor specifics from core crate into supervisor crate
  * Added `habitat_sup::fs` module
  * Move `pkg_svc_*` functions from `PackageInstall` struct to
    `habitat_sup::fs` module


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -41,8 +40,6 @@ pub const PKG_PATH: &'static str = "hab/pkgs";
 /// Using this variable could lead to broken supervisor services and it should
 /// be used with extreme caution.
 pub const FS_ROOT_ENVVAR: &'static str = "FS_ROOT";
-/// The root path containing all runtime service directories and files
-const SVC_PATH: &'static str = "hab/svc";
```
fnichol: This does make more sense in `sup` as it isn't a broadly shared idea

components/core/src/package/install.rs
```diff
@@ -307,41 +307,6 @@ impl PackageInstall {
         &self.installed_path
     }
 
-    /// Returns the root path for service configuration, files, and data.
```
fnichol: Also makes more sense in `Service`

components/sup/src/manager/service/config.rs
```diff
@@ -60,30 +63,71 @@ pub struct ServiceConfig {
     // Set to 'true' if we have data that needs to be sent to a configuration file
     #[serde(skip_deserializing)]
     pub needs_write: bool,
+    #[serde(skip_deserializing)]
+    supported_bindings: Vec<(String, String)>,
```
fnichol: Perhaps a `Vec<(String, ServiceConfig)>` instead? I was hoping to validate this input a lot earlier in the Supervisor's lifecyle and this might help.

components/sup/src/manager/service/config.rs
```diff
@@ -60,30 +63,71 @@ pub struct ServiceConfig {
...
-            bind: bind,
+            cfg: Cfg::new(package, config_root)?,
+            svc: Svc::default(),
+            bind: Bind::default(),
             needs_write: true,
+            supported_bindings: Self::split_bindings(bindings)?,
         })
     }
 
+    fn split_bindings(bindings: &[String]) -> Result<Vec<(String, String)>> {
```
fnichol: I was considering performing this earlier in CLI parsing (later when writing into a `ServiceSpec` potentially from disk) which would help us catch `--bind db:redis` syntax errors more quickly. Thoughts?

components/sup/src/manager/service/config.rs
```diff
@@ -317,17 +317,17 @@ pub struct Cfg {
 }
 
 impl Cfg {
-    fn new(pkg: &Package) -> Result<Cfg> {
+    fn new<T: AsRef<Path>>(package: &Package, config_root: T) -> Result<Cfg> {
```
fnichol: Yeah, that's probably the right use of the `--config-from` that doesn't infect the code in a strange way. It'll be up to the Supervisor to strongly discourage using an override when not in development. This may also help us in testing

components/sup/src/manager/service/config.rs
```diff
@@ -478,93 +478,32 @@ pub struct Pkg {
     pub svc_files_path: String,
     pub svc_static_path: String,
     pub svc_var_path: String,
-    pub svc_user: Option<String>,
-    pub svc_group: Option<String>,
-    pub svc_user_default: String,
-    pub svc_group_default: String,
+    pub svc_user: String,
```
fnichol: Much nicer, only the *used* values and not optional--there is always a set value.

components/sup/src/manager/service/config.rs
```diff
@@ -478,93 +478,32 @@ pub struct Pkg {
...
-            version: version,
-            release: release,
+    fn new(package: &Package, runtime: &RuntimeConfig) -> Result<Pkg> {
+        let ident = package.pkg_install.ident().clone();
+        Ok(Pkg {
             ident: ident.to_string(),
-            deps: deps,
-            exposes: exposes,
-            exports: exports,
-            path: pkg_install.installed_path().to_string_lossy().into_owned(),
```
fnichol: A bunch of `pkg_install` feature envy, no?

components/sup/src/manager/service/config.rs
```diff
@@ -478,93 +478,32 @@ pub struct Pkg {
...
-            svc_user: svc_user,
-            svc_group: svc_group,
-            svc_user_default: default_svc_user,
-            svc_group_default: default_svc_group,
-        }
+            origin: ident.origin,
+            name: ident.name,
+            version: ident.version.expect("Couldn't read package version"),
+            release: ident.release.expect("Couldn't read package release"),
+            deps: package.tdeps.clone(),
```
fnichol: Was `tdeps` on purpose here?

components/sup/src/manager/service/mod.rs
```diff
@@ -77,26 +81,31 @@ impl ServiceSpec {
             topology: Topology::default(),
             update_strategy: UpdateStrategy::default(),
             binds: vec![],
-            dev_config_from: None,
+            config_from: None,
```
fnichol: For a future discussion: I'm hoping to only use this (i.e. set `Some(…)`) through the `hab-sup start <PKG_IDENT_OR_ARTIFACT> …` invocation.

components/sup/src/manager/service/mod.rs
```diff
@@ -77,26 +81,31 @@ impl ServiceSpec {
...
-    pub needs_restart: bool,
-    pub package: Package,
     pub cfg_incarnation: u64,
-    pub service_group: ServiceGroup,
-    pub topology: Topology,
-    pub update_strategy: UpdateStrategy,
+    pub config: ServiceConfig,
+    pub config_root: PathBuf,
     pub current_service_files: HashMap<String, u64>,
+    pub depot_url: String,
```
fnichol: I keep wanting this to be a `Url` but keep falling back to a `String`. Partly because there is no Serialize implemented on `url:Url` and partly because our codebase pre-validates this value before inserting it anyway.

components/sup/src/manager/service/mod.rs
```diff
@@ -77,26 +81,31 @@ impl ServiceSpec {
...
+    pub health_check: HealthCheck,
     pub initialized: bool,
     pub last_restart_display: LastRestartDisplay,
-    pub supervisor: Supervisor,
-    pub depot_url: String,
+    pub needs_restart: bool,
+    pub package: Package,
+    pub service_group: ServiceGroup,
+    pub smoke_check: SmokeCheck,
     pub spec_ident: PackageIdent,
```
fnichol: Does `spec_ident` make sense to you? I needed a record of the "initial requirement" for the package to be booting into a service. In this way if you `hab start core/redis/1.2.3` we know how to check for updates more correctly. One interesting side effect of calling `hab start /path/to/core-redis-1.2.3-20170209010203-x86_64-linux.hart` is that `spec_ident` will be set to `core/redis/1.2.3/20170209010203` and would therefore *not* be updated. I think this is correct behavior albeit potentially surprising for a user on first blush.

components/sup/src/manager/service/mod.rs
```diff
@@ -106,62 +115,102 @@ impl Service {
                                               spec.organization.as_ref().map(|x| &**x))?;
         let (svc_user, svc_group) = try!(util::users::get_user_and_group(&package.pkg_install));
         let runtime_config = RuntimeConfig::new(svc_user, svc_group);
-        let supervisor = Supervisor::new(package.ident().clone(), &service_group, runtime_config);
+        let config_root = spec.config_from.unwrap_or(package.pkg_install.installed_path().clone());
```
fnichol: Yuuuuuup.

components/sup/src/manager/service/mod.rs
```diff
@@ -106,62 +115,102 @@ impl Service {
...
-            current_service_files: HashMap::new(),
-            last_restart_display: LastRestartDisplay::None,
-            initialized: false,
-            cfg_incarnation: 0,
+            hooks: HookTable::default(),
         })
     }
 
+    /// Create the service path for this package.
+    pub fn create_svc_path(&self) -> Result<()> {
```
fnichol: Better location by far!

components/sup/src/config.rs
```diff
@@ -121,6 +121,4 @@ impl fmt::Display for GossipListenAddr {
 pub struct Config {
     // Currently only used by `ServiceConfig`
     pub service_config_http_listen: http_gateway::ListenAddr,
```
fnichol: Now it's just down to you, lil buddy

components/builder-api/src/config.rs
```diff
@@ -64,6 +66,7 @@ impl Default for Config {
             github_client_secret: DEV_GITHUB_CLIENT_SECRET.to_string(),
             ui_root: None,
             events_enabled: false, // TODO: change to default to true later
+            log_dir: "/tmp".to_string(),
```
mwrock: we probably want to abstract temp directory since it would be the temp environment variable on windows.

fnichol: Potentially use something like [`std::env::temp_dir()`](https://doc.rust-lang.org/std/env/fn.temp_dir.html) which appears to do the Right Things™?

components/sup/src/manager/service/config.rs
reset: @fnichol this could become `Vec<(String, ServiceGroup)>`, I'll make that change

components/sup/src/manager/service/config.rs
reset: The earlier that we validate input the better. I would move this function to that area in a future refactor if I were you for sure!

components/hab/src/main.rs
```diff
@@ -161,13 +160,11 @@ fn start(ui: &mut UI) -> Result<()> {
 }
 
 fn sub_cli_setup(ui: &mut UI) -> Result<()> {
-    let fs_root = henv::var(FS_ROOT_ENVVAR).unwrap_or(FS_ROOT_PATH.to_string());
-    let fs_root_path = Some(Path::new(&fs_root));
```
fnichol: We'll still need these `FS_ROOT_ENVVAR` checks here, for non-Windows Studio and `hab pkg export` tooling. The check used in `core` only applies to Windows when potentially working in the Studio--I can fill you in more if that's helpful, but we'll want to preserve these checks in `main.rs`.

components/sup/src/command/start.rs
```diff
@@ -83,7 +81,7 @@ pub fn package(cfg: ManagerConfig, spec: ServiceSpec, local_artifact: Option<&st
         return Err(sup_error!(Error::RootRequired));
     }
 
-    match Package::load(&spec.ident, Some(&*FS_ROOT_PATH)) {
+    match PackageInstall::load(&spec.ident, Some(&Path::new(&*FS_ROOT_PATH))) {
```
fnichol: Finally, happy to see this!

components/sup/src/manager/mod.rs
```diff
@@ -110,25 +110,21 @@ impl Manager {
...
-        let service_rumor = ServiceRumor::new(self.state.butterfly.member_id().to_string(),
-                                              service.package.ident(),
-                                              &service.service_group,
-                                              &*svc_cfg.sys,
-                                              Some(&cfg))?;
-        self.state.butterfly.insert_service(service_rumor);
-
+        // JW TODO: We can't just set the run path for the entire process here, we need to set
+        // this on the supervisor which will be starting the process itself to support multi
+        // services in hab-sup.
```
fnichol: Agreed, I'll have to deal with this pretty soon in `Supervisor`.

components/hab/src/main.rs
reset: @fnichol I adressed this by creating a private static ref which encapsulates this behaviour and uses that throughout this module

components/core/src/event.rs
```diff
@@ -284,7 +282,7 @@ mod test {
 
     #[test]
     fn event_logger_path() {
-        let event_logger: EventLogger = EventLogger::new("foo", true);
+        let event_logger: EventLogger = EventLogger::new("/hab/svc/foo/var", true);
```
fnichol: Might want to consider using `svc_var_path("foo")` in the future.

components/core/src/fs.rs
```diff
@@ -41,16 +40,19 @@ pub const PKG_PATH: &'static str = "hab/pkgs";
...
 lazy_static! {
-    /// The default filesystem root path
-    pub static ref FS_ROOT_PATH: String = {
+    /// The default filesystem root path.
+    ///
+    /// WARNING: On Windows this variable mutates on first call if an environment variable with
+    ///          the key of `FS_ROOT_ENVVAR` is set.
+    pub static ref FS_ROOT_PATH: PathBuf = {
+        // JW TODO: When Windows container studios are available the platform reflection should
+        // be removed.
```
fnichol: 👍

components/hab-butterfly/src/main.rs
```diff
@@ -33,24 +35,32 @@ use clap::ArgMatches;
...
-
 const HABITAT_BUTTERFLY_PORT: u64 = 9638;
-
 const MAX_FILE_UPLOAD_SIZE_BYTES: u64 = 4096;
 
+lazy_static! {
+    /// The default filesystem root path to base all commands from. This is lazily generated on
+    /// first call and reflects on the presence and value of the environment variable keyed as
+    /// `FS_ROOT_ENVVAR`.
+    static ref FS_ROOT: PathBuf = {
```
fnichol: Probably could have been a simple function as the subcommands only ever get invoked once but this works for me.

components/sup/src/manager/service/config.rs
```diff
@@ -57,33 +60,78 @@ pub struct ServiceConfig {
...
     pub svc: Svc,
     pub bind: Bind,
+    #[serde(skip_serializing)]
+    pub config_root: PathBuf,
     // Set to 'true' if we have data that needs to be sent to a configuration file
-    #[serde(skip_deserializing)]
+    #[serde(skip_serializing, skip_deserializing)]
     pub needs_write: bool,
+    #[serde(skip_serializing, skip_deserializing)]
+    supported_bindings: Vec<(String, ServiceGroup)>,
```
fnichol: 👍

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/200518966).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/200518966).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/200518966).

components/builder-worker/src/config.rs
```diff
@@ -54,9 +53,7 @@ impl Default for Config {
         jobsrv.insert("heartbeat".to_string(), "5567".to_string());
         Config {
             auth_token: "".to_string(),
-            data_path: hab_core::fs::svc_data_path("hab-builder-worker")
-                .to_string_lossy()
-                .into_owned(),
+            data_path: "/tmp".to_string(),
```
adamhjk: Why would this switch to /tmp? We're breaking out of the `/hab/svc/FOO` in a way that feels wrong. What's the intent?

components/core/src/event.rs
```diff
@@ -262,9 +260,9 @@ pub struct EventLogger {
 }
 
 impl EventLogger {
-    pub fn new(service_name: &str, enabled: bool) -> Self {
+    pub fn new<T: Into<PathBuf>>(log_dir: T, enabled: bool) -> Self {
         EventLogger {
-            log_dir: svc_var_path(&service_name),
+            log_dir: log_dir.into(),
```
adamhjk: Is this logging to a file rather than stdout because we want to collect the data for some other reason? Services in general should log to stdout.

components/core/src/fs.rs
```diff
@@ -41,16 +40,19 @@ pub const PKG_PATH: &'static str = "hab/pkgs";
...
         if cfg!(target_os = "windows") && henv::var(FS_ROOT_ENVVAR).is_ok() {
-            henv::var(FS_ROOT_ENVVAR).unwrap()
+            PathBuf::from(henv::var(FS_ROOT_ENVVAR).unwrap())
```
adamhjk: Cheers.

components/core/src/event.rs
reset: Yes, these are event logs to be collected by another service

components/builder-worker/src/config.rs
reset: This is the default configuration for running the service by `cargo run`. It's set to `/tmp` (should probably be `std::env::temp_dir()` like the others) because you may not have the Habitat directory structure on your machine when running tests. This wasn't previously an issue because event logging is disabled by default. If anyone changed that, it would break the test suite.

I actually only made this change becuase `hab_core::fs::svc_data_path()` no longer exists. That's part of the `habitat_sup` module and not `hab_core`

PR #1755, opened 2017-02-12
Fix configuration and hook compilation errors

reset:
This PR addresses some bugs that I introduced from a large refactor for multi-tenant supervisor support in #1734. I also removed a few duplicate fields from the `CensusEntry` struct which have moved to the struct in it's `sys` field.

tldr; This allows the supervisor to run any service that defines it's own `run` hook once again and fixes a number of edge cases I introduced due by optimizing out unnecessary allocations we were previously doing on every server tick.


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -491,32 +338,20 @@ impl Service {
...
-            }
-            self.initialized = true;
+        if self.initialized {
+            return;
+        }
+        outputln!(preamble self.service_group, "Initializing");
+        self.hooks.load_hooks(&self.supervisor.runtime_config,
+                              &self.config,
+                              &self.service_group);
+        if let Some(err) = self.run_hook(HookType::Init).err() {
```
bodymindarts: This fails because after loading the hooks they need to be compiled.

components/sup/src/manager/service/mod.rs
```diff
@@ -529,11 +364,15 @@ impl Service {
         match self.config.write() {
             Ok(true) => {
                 self.needs_restart = true;
+                self.hooks.compile(&self.service_group, &self.config);
```
bodymindarts: The first time this code gets called hooks haven't been loaded yet.

bodymindarts:
Here is a PR that loads the hooks in `Service::new()` which addresses the above issue:
https://github.com/habitat-sh/habitat/pull/1756

bodymindarts:
I've rebased my changes on top of this branch including the above fix.
Here you can merge my changes into this branch: https://github.com/habitat-sh/habitat/pull/1757

bodymindarts:
There seems to be another problem other than the hooks not being loaded / compiled properly (addressed in https://github.com/habitat-sh/habitat/pull/1756)

The values of `{{svc.leader.ip}}` and `{{svc.leader.port}}` aren't being rendered properly.
Investigating...

bodymindarts:
Okay I see that `{{svc.leader.ip}}` is can be found under `{{svc.leader.address}}` or `{{svc.leader.sys.ip}}` but the `leader.port` seems to be gone completely. Don't know if that has to do with this PR or other changes between `v0.16.0` and head.

components/sup/src/manager/service/mod.rs
reset: Great catch, I've got some changes on my branch to push up to optimize calling of compilation which will sort out these bugs from sneaking in for the future. Thanks for reviewing this for me ;)

reset:
@bodymindarts thank you so much for being patient and joining in on the development of the supervisor with us! I'm sorry that you were greeted with such a large and difficult to navigate refactoring in exactly the same spot of code that you were modifying. I swear, it's not always like this :).

You're spot on that `{{svc.leader.ip}}` and `{{svc.leader.port}}` have been removed. This upcoming release adds a new feature called [pkg_exports](https://github.com/habitat-sh/habitat/pull/1667) which replace the hack we had in place for providing templates with `{{svc.leader.port}}`.

`{{svc.leader.address}}` remains for now - I need to have a conversation with @adamhjk and see if I'm correct that `{{svc.leader.address}}` is *always* the same as `{{sys.ip}}` or if it has a different intended use.

I've pushed some changes to this branch to address timing issues with hook loading and compilation which should also be performant enough for now. Your comments from the work I pushed late last night helped greatly in getting that fixed ❤️ . Let's get this merged through to get to a working base and then let's get your timing issues integrated into that. I think we can get them in for the next release for sure!

bodymindarts:
@reset I'm very happy to join the dev efforts and have noticed a bunch of aspects in this code section that have potential for improvement. Was just focusing on the fixes that are blocking me in adding HA support to the core/postgresql package.

Its unfortunate we keep colliding. Perhaps we can find a moment to talk on zoom before I rework it again. Also if I get a sense of what the plan is in terms of potential features that affect the supervisor I could also take that into account if I end up doing more refactoring / bug fixing.

components/sup/src/manager/service/hooks.rs
```diff
@@ -188,13 +192,24 @@ impl HookTable {
...
-            .expect("Can't determine user:group");
-        match fs::metadata(&template) {
-            Ok(_) => Some(Hook::new(hook_type, template, concrete, user, group)),
+    fn load_hook<T>(&self,
+                    hook_type: HookType,
+                    runtime_cfg: &RuntimeConfig,
+                    hooks: T,
+                    templates: T)
+                    -> Option<Hook>
+        where T: AsRef<Path>
```
fnichol: Just a note that with type `T`, both `hooks` and `templates` must be the same input type (i.e. both `&str`, `String`, etc.) otherwise you need a generic per-input ([like here](https://github.com/habitat-sh/habitat/blob/57a51506f486211c7db1668208ac2c5e7c5bd0fc/components/common/src/command/package/install.rs#L67-L68)).

reset: I'll just add a second generic since the goal of this is to be an ergonomics improvement and not to save on character count ;)

mwrock:
This LGTM!

reset:
@bodymindarts that's a great idea, what time zone are you in? It might be easiest to sync up on the Habitat slack and go from there. I'm also @reset on there

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/201244648).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/201244648).

Issue #1742, opened 2017-02-09
Inconsistency between {{svc.me.leader}} && {{svc.me.follower}}


bodymindarts:
In a 3 node setup there will be 1 node where 
```
{{svc.me.leader}} = 'true'
{{svc.me.follower}} = 'false'
```
 and 2 where 
```
{{svc.me.leader}} = 'false'
{{svc.me.follower}} = 'true'
```

After killing the leader the situation is as follows:
1 node has : 
```
{{svc.me.leader}} = 'false'
{{svc.me.follower}} = 'true'
```
and 1 has:
```
{{svc.me.leader}} = 'false'
{{svc.me.follower}} = 'false'
```

So there is obviously still a leader (`follower != 'true'`) but you cannot discover it via querying `{{svc.me.leader}}`

adamhjk:
@bodymindarts has the election finished on the followers?

adamhjk:
We dug a little more into this via zoom. We found a couple things:

1) We are currently re-configuring during interstitial election states; we really only want to reconfigure when an election is finished, not while its in progress.

2) We run reconfigure before we run the init hook - this should be reversed

3) Lets write some test coverage for when hooks get triggered :)

PR #1763, opened 2017-02-14
Run hooks after leader is elected and in correct order

bodymindarts:
When Topology::Leader is selected hooks are only run when the
election is completed. This garuantees {{svc.me.leader}} and
{{svc.me.follower}} will be in a consistent state when the hooks are
executed.
Also the execution of the reconfigure hook is now consistently after the
run hook (also on restarts) as is documented.

Fixes https://github.com/habitat-sh/habitat/issues/1742

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -43,7 +43,7 @@ use self::hooks::{HOOK_PERMISSIONS, HookTable, HookType};
 use error::{Error, Result, SupError};
 use fs;
 use manager::signals;
-use manager::census::{CensusList, CensusUpdate};
+use manager::census::{CensusList, CensusUpdate, ElectionStatus};
 use prometheus::Opts;
 use supervisor::{Supervisor, RuntimeConfig};
 use util;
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -370,24 +417,12 @@ impl Service {
...
-                outputln!(preamble self.service_group,
-                          "Failed to write service configuration: {}",
-                          e);
-            }
+        if let Some(err) = self.run_hook(HookType::Reconfigure).err() {
+            outputln!(preamble self.service_group,
+                      "Reconfiguration hook failed: {}",
+                      err);
+        } else {
+            self.needs_reconfiguration = false;
```
reset: Only clearing this bit on run hook success means that the service can get in an infinite restart loop if a bad reconfigure hook is provided. It might be best to just warn and clear this no matter what to try and just keep running if possible. What do you think?

adamhjk: it's a good question - if you fail and retry, you can write reconfigure scripts that converge, rather than a certain to work the first time. If you clear the bit and move on, you really can't - at least, you would need to put that logic inside the reconfigure hook. 

I'm not sure what I think :)

reset:
@bodymindarts this looks great! Just the one comment that I left and we discussed and I think we can merge this one in

components/sup/src/manager/service/mod.rs
reset: @adamhjk I didn't think of a "let it crash" model for the hooks. I think either way will work as long as it's documented

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/201657483).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/201657483).

PR #1776, opened 2017-02-16
Refresh and reinitialize service after package update

mwrock:
Updates are currently broken because updated packages are not refreshing internal package data, reloading hooks and reinitializing the service.

This PR ensures that when a package is updated:
1. The data in the service, config and supervisor is refreshed with new package metadata and identity
2. All hooks are reloaded with the new template location
3. All hooks are recompiled
4. the init hook is run
5. and the run hook is copied to the run location

This uses an Arc<RwLock<<PackageInstall>> to share and access package data accross the service and supervisor to keep them synchronized.

### DIFF
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -81,10 +81,9 @@ impl ServiceConfig {
     pub fn new(package: &PackageInstall,
                mgr_cfg: &ManagerConfig,
                runtime_cfg: &RuntimeConfig,
-               config_root: Option<PathBuf>,
+               config_root: PathBuf,
                bindings: Vec<(String, ServiceGroup)>)
                -> Result<ServiceConfig> {
-        let config_root = config_root.unwrap_or(package.installed_path.clone());
         Ok(ServiceConfig {
             pkg: Pkg::new(package, runtime_cfg)?,
             hab: Hab::new(),
@@ -226,6 +225,17 @@ impl ServiceConfig {
         }
         self.needs_write = false;
         Ok(should_restart)
+    }
+
+    pub fn reload_package(&mut self,
+                          package: &PackageInstall,
+                          config_root: PathBuf,
+                          runtime: &RuntimeConfig)
+                          -> Result<()> {
+        self.config_root = config_root;
+        self.pkg = Pkg::new(package, runtime)?;
+        self.cfg = Cfg::new(package, &self.config_root)?;
+        Ok(())
     }
 }
 
@@ -661,7 +671,7 @@ mod test {
         let sc = ServiceConfig::new(&pkg,
                                     &ManagerConfig::default(),
                                     &RuntimeConfig::default(),
-                                    None,
+                                    PathBuf::from("/hab/pkgs/neurosis/redis/2000/20160222201258"),
                                     Vec::new())
             .unwrap();
         let toml = sc.to_toml().unwrap();
@@ -675,7 +685,7 @@ mod test {
         let sc = ServiceConfig::new(&pkg,
                                     &ManagerConfig::default(),
                                     &RuntimeConfig::default(),
-                                    None,
+                                    PathBuf::from("/hab/pkgs/neurosis/redis/2000/20160222201258"),
                                     Vec::new())
             .unwrap();
         let toml = sc.to_toml().unwrap();
@@ -689,7 +699,7 @@ mod test {
         let sc = ServiceConfig::new(&pkg,
                                     &ManagerConfig::default(),
                                     &RuntimeConfig::default(),
-                                    None,
+                                    PathBuf::from("/hab/pkgs/neurosis/redis/2000/20160222201258"),
                                     Vec::new())
             .unwrap();
         let toml = sc.to_toml().unwrap();
--- a/components/sup/src/manager/service_updater.rs
+++ b/components/sup/src/manager/service_updater.rs
@@ -99,7 +99,7 @@ impl ServiceUpdater {
             Some(&mut UpdaterState::AtOnce(ref mut rx)) => {
                 match rx.try_recv() {
                     Ok(package) => {
-                        service.package = package;
+                        service.update_package(package);
                         service.needs_restart = true;
                         return true;
                     }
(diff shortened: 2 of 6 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/config.rs
```diff
@@ -81,12 +82,17 @@ impl ServiceConfig {
     pub fn new(package: &PackageInstall,
                mgr_cfg: &ManagerConfig,
                runtime_cfg: &RuntimeConfig,
-               config_root: Option<PathBuf>,
+               config_root_from_config: Option<PathBuf>,
                bindings: Vec<(String, ServiceGroup)>)
                -> Result<ServiceConfig> {
-        let config_root = config_root.unwrap_or(package.installed_path.clone());
+        let config_root = match config_root_from_config {
```
reset: These 4 lines do the same as the replaced line except they also cost an additional allocation. Why the change?

components/sup/src/manager/service/config.rs
```diff
@@ -81,12 +82,17 @@ impl ServiceConfig {
...
+            Some(ref root) => root.clone(),
+            None => package.installed_path.clone(),
+        };
         Ok(ServiceConfig {
-            pkg: Pkg::new(package, runtime_cfg)?,
+            pkg: Pkg::new(package,
```
reset: I think we do want to pass a ref to the runtime config and not just the user/group.

We should totally be storing a ref to the RuntimeConfig and not just pull it's parts out in the Pkg constructor like we previously had, but that refactor/optimization hasn't happened yet. This is in line with the spirit of this PR - don't duplicate memory into caches that get stale ;)

components/sup/src/manager/service/config.rs
```diff
@@ -227,6 +234,18 @@ impl ServiceConfig {
         self.needs_write = false;
         Ok(should_restart)
     }
+
+    pub fn reload_package(&mut self, package: &PackageInstall) -> Result<()> {
+        self.config_root = match self.config_root_from_config {
```
reset: Same comment here [as above](https://github.com/habitat-sh/habitat/pull/1776/files#r101658477). This form causes an additional allocation

components/sup/src/manager/service/config.rs
```diff
@@ -227,6 +234,18 @@ impl ServiceConfig {
...
+            Some(ref root) => root.clone(),
+            None => package.installed_path.clone(),
+        };
+        self.pkg = Pkg::new(package,
```
reset: I see here why you've opted to change the constructor now! What we really want to do is update our RuntimeConfig so when the supervisor restarts it uses it. Else, we'll update the Pkg struct in ServiceConfig but the supervisor itself won't use the new user/group.

If it happens to today, it's just a side effect of how the code was structured. This reload function needs to reload everything about the Service which is informed by the Package

components/sup/src/manager/service/config.rs
```diff
@@ -227,6 +234,18 @@ impl ServiceConfig {
...
+                            self.pkg.svc_user.clone(),
+                            self.pkg.svc_group.clone())?;
+        self.cfg = Cfg::new(package, &self.config_root)?;
+        Ok(())
```
reset: Before exiting this function I think you want to throw away the entire hook table and load it again. Hooks aren't just additive, they could also be removed between package updates. Doing this here might also simplify some logic below - you probably don't even need to ask for the hooks to be compiled since rebuilding the entire table will clear the incarnation counter and they'll be recompiled on the next call to `run_hook()`

components/sup/src/manager/service_updater.rs
```diff
@@ -99,7 +99,7 @@ impl ServiceUpdater {
             Some(&mut UpdaterState::AtOnce(ref mut rx)) => {
                 match rx.try_recv() {
                     Ok(package) => {
-                        service.package = package;
+                        service.update_package(package);
```
reset: This is all much better - great encapsulation

components/sup/src/manager/service/config.rs
mwrock: because we are owning `config_root_from_config` and assigning it to our own property below, The replaced line had to clone `config_root_from_config `. and it would need to clone regardless of whether it would use it (like if it was `None`) so I broke it out to only clone if it was used. Maybe its just a micro optimization since cloning `None` is little to none.

components/sup/src/manager/service/config.rs
mwrock: Yeah I do exactly that in the service's update package right after this call. It did not seem right to mess with hooks inside the service config.

mwrock:
Yeah the data structures did not feel super but I did not want to introduce fundamental refactorings while we are in the midst of the mult process work.

components/sup/src/manager/service/config.rs
reset: Oops! I meant to leave this comment on `update_package` on the `Service` struct and not this function on `ServiceConfig`. I see that you've defined some new functions for recompiling the templates but what I am suggesting is to completely throw away the old `HooksTable` and reload it

mwrock: funny i did that in pass one and yeah, in retrospect I think thats better. Maybe when I put it back I'll remember why I thought the inplace change was better. I'll make that change.

reset:
gif-keyboard-8115030937132548043

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/202479246).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/202479246).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/202479246).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/202479246).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/202479246).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/202479246).

PR #1791, opened 2017-02-17
[sup] Refactor Service to install itself & load from a ServiceSpec.

fnichol:
This internal refactoring tries to accomplish

* Push the initial package installation into a `Service` concern--it
  should be responsible for ensuring that the desired package is
  installed.
* Allow a `Service` to be created or "loaded" from a `ServiceSpec` (and
  possible a couple of other small bits of info). This makes future work
  to start additional services much easier as they load from a simple
  value struct. **Note:** the `Service::new` wasn't overriden and remains
  public so that future testing work could construct a `Service` from a
  pretend or faked out `PackageInstall`.
* Divest a `Service` from needing to know about a `ManagerConfig` as
  this value struct is consumed when the `Manager` is first built--future
  services will not be able to use this struct once the `Manager` is
  running.
* Prepare the codebase for a use of the `Manager` without any services.
* Begin to add the concept of a `Service` being added (i.e. initialized,
  booted up, brought into being) and being removed (i.e. decommissioned,
  torn down, deleted) in the `Manager`'s lifecycle.'
* Begin to further centralize the logic to install a package in the
  Supervisor that will eventually be used anywhere this task is required,
  leading to fewer bugs and one code path for all future package
  installation improvements and fixes.

As this is a refactoring change, and being incremental not everything
here is perfect or as nice as I'd like. Future refactorings will
endeavor to tighten up the code even more. This was enough to unlock the
feature work of multiple services.

### DIFF
--- a/components/sup/src/command/start.rs
+++ b/components/sup/src/command/start.rs
@@ -50,109 +50,41 @@
 use std::path::Path;
 
 use ansi_term::Colour::Yellow;
-use common::command::package::install;
+use common;
 use common::ui::UI;
-use depot_client::Client;
-use hcore::fs::{am_i_root, cache_artifact_path, FS_ROOT_PATH};
-use hcore::package::{PackageIdent, PackageInstall};
+use hcore::fs::{self, FS_ROOT_PATH};
 
 use {PRODUCT, VERSION};
 use error::{Error, Result};
 use manager::{Manager, ManagerConfig};
-use manager::{Service, ServiceSpec, UpdateStrategy};
+use manager::ServiceSpec;
 
 static LOGKEY: &'static str = "CS";
 
-/// Creates a [Package](../../pkg/struct.Package.html), then passes it to the run method of the
-/// selected [topology](../../topology).
-///
-/// # Failures
-///
-/// * Fails if it cannot find a package with the given name
-/// * Fails if the `run` method for the topology fails
-/// * Fails if an unknown topology was specified on the command line
 pub fn package(cfg: ManagerConfig, spec: ServiceSpec, local_artifact: Option<&str>) -> Result<()> {
     let mut ui = UI::default();
-    if !am_i_root() {
-        try!(ui.warn("Running the Habitat Supervisor requires root or administrator privileges. \
-                      Please retry this command as a super user or use a privilege-granting \
-                      facility such as sudo."));
-        try!(ui.br());
+    if !fs::am_i_root() {
+        ui.warn("Running the Habitat Supervisor requires root or administrator privileges. \
+                   Please retry this command as a super user or use a privilege-granting \
+                   facility such as sudo.")?;
+        ui.br()?;
         return Err(sup_error!(Error::RootRequired));
     }
 
-    match PackageInstall::load(&spec.ident, Some(&Path::new(&*FS_ROOT_PATH))) {
-        Ok(mut package) => {
-            match spec.update_strategy {
-                UpdateStrategy::None => {}
-                _ => {
-                    outputln!("Checking Depot for newer versions...");
-                    // It is important to pass `spec.ident` to `show_package()` instead
-                    // of the package identifier of the loaded package. This will ensure that
-                    // if the operator starts a package while specifying a version number, they
-                    // will only automatically receive release updates for the started package.
-                    //
-                    // If the operator does not specify a version number they will
-                    // automatically receive updates for any releases, regardless of version
-                    // number, for the started  package.
-                    let depot_client = try!(Client::new(&spec.depot_url, PRODUCT, VERSION, None));
-                    let latest_pkg_data = try!(depot_client.show_package(&spec.ident));
-                    let latest_ident: PackageIdent = latest_pkg_data.get_ident().clone().into();
-                    if &latest_ident > package.ident() {
-                        outputln!("Downloading latest version from Depot: {}", latest_ident);
-                        let new_pkg_data = try!(install::start(&mut ui,
-                                                               &spec.depot_url,
-                                                               &latest_ident.to_string(),
-                                                               PRODUCT,
-                                                               VERSION,
-                                                               Path::new(&*FS_ROOT_PATH),
-                                                               &cache_artifact_path(None),
-                                                               false));
-                        package = try!(PackageInstall::load(&new_pkg_data, Some(&*FS_ROOT_PATH)));
-                    } else {
-                        outputln!("Already running latest.");
-                    };
-                }
-            }
-            start_package(package, cfg, spec)
-        }
-        Err(_) => {
-            outputln!("{} is not installed",
-                      Yellow.bold().paint(spec.ident.to_string()));
-            let new_pkg_data = match local_artifact {
-                Some(artifact) => {
-                    try!(install::start(&mut ui,
-                                        &spec.depot_url,
-                                        &artifact,
-                                        PRODUCT,
-                                        VERSION,
-                                        Path::new(&*FS_ROOT_PATH),
-                                        &cache_artifact_path(None),
-                                        false))
-                }
-                None => {
-                    outputln!("Searching for {} in remote {}",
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/main.rs
```diff
@@ -101,30 +101,34 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
...
             (@arg PERMANENT_PEER: --("permanent-peer") -I "If this service is a permanent peer")
             (@arg RING: --ring -r +takes_value "Ring key name")
-            (@arg STRATEGY: --strategy -s +takes_value {valid_update_strategy}
-                "The update strategy; [default: none] [values: none, at-once, rolling]")
-            (@arg TOPOLOGY: --topology -t +takes_value {valid_topology}
-                "Service topology; [default: none]")
+            (@arg PKG_IDENT_OR_ARTIFACT: +required
+                "A Habitat package identifier (ex: acme/redis) or filepath to a Habitat Artifact \
+                (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
+            (@group SVC_ARGS =>
```
smurawski: This is sweet.

fnichol: This is yet another reason why I love clap so much. The library and @kbknapp have always delivered for me--I can design a user experience and clap lets me implement it without adding new features.

kbknapp: Aw, thanks! 😄

reset: So much cleaner!

components/sup/src/command/start.rs
```diff
@@ -50,109 +50,41 @@
...
-                                        VERSION,
-                                        Path::new(&*FS_ROOT_PATH),
-                                        &cache_artifact_path(None),
-                                        false))
-                }
-            };
-            let package = try!(PackageInstall::load(&new_pkg_data, Some(&*FS_ROOT_PATH)));
-            start_package(package, cfg, spec)
-        }
+    if let Some(artifact) = local_artifact {
```
reset: I am feeling that starting an artifact from a `.hart` might ultimately be a bad idea. We need to do some not-so-elegant things to solve edge cases that it introduces. In design theory if you create a rule, you know the rule is wrong if you need to create "patch rules" especially if you're patching patches.

I think we only ever provided this because we wanted to save somebody some keystrokes by avoiding them to first need to run the install command on their hart before issueing a start command. As we evolve the idea of a ServiceSpec I don't think there's a place for a local artifact in it anymore - essentially what the ServiceSpec has is a "goal" of what to start. That goal can change depending on the service group that the supervisor joins. For example, if you join a service group that's running a different version of the package than you are and the service group is configured in the rolling update strategy then you're going to be upgraded/downgraded to the version of the software everyone else is running.

What do you think? Are we just doing this to save some keystrokes? Should we just ask people to run install first?

components/sup/src/manager/mod.rs
```diff
@@ -145,6 +146,13 @@ impl Manager {
         loop {
             let next_check = SteadyTime::now() + TimeDuration::milliseconds(1000);
             if self.check_for_incoming_signals() {
+                let mut services = self.state
+                    .services
+                    .write()
+                    .expect("Services lock is poisend!");
+                for service in services.drain(..) {
```
reset: I love this. `drain` creates an iterator for you *and* clears the vec at the same time

components/sup/src/manager/service/mod.rs
```diff
@@ -183,6 +193,51 @@ impl Service {
...
+    pub fn load(spec: ServiceSpec,
+                gossip_listen: &GossipListenAddr,
+                http_listen: &http_gateway::ListenAddr)
+                -> Result<Service> {
+        let mut ui = UI::default();
+        let package = match PackageInstall::load(&spec.ident, Some(&Path::new(&*FS_ROOT_PATH))) {
+            Ok(package) => {
+                match spec.update_strategy {
+                    UpdateStrategy::AtOnce | UpdateStrategy::Rolling => {
+                        try!(util::pkg::maybe_install_newer(&mut ui, &spec, package))
```
reset: We actually shouldn't perform a `maybe_install_newer()` action here for `UpdateStrategy::Rolling`. We were before, but that was bad behaviour. When the service joins the group it's going to look at it's peers and see what version it should be on.

In the behaviour we had (and by proxy, this behaviour) a service with a broad goal of a package ident could be started, update, and then join a service group which is in the rolling strategy. As soon as it joins it could notice it's peers are on a different version - maybe the election leader had a more strict goal perhaps - and then the service would downgrade itself and *then* join the service group.

tldr; add `UpdateStrategy::Rolling` to the same match arm as `UpdateStrategy::None` :)

components/sup/src/manager/service/mod.rs
```diff
@@ -183,6 +193,51 @@ impl Service {
...
+                try!(util::pkg::install(&mut ui, &spec.depot_url, &spec.ident))
+            }
+        };
+        Self::new(package, spec, gossip_listen, http_listen)
+    }
+
+    pub fn add(&self) -> Result<()> {
+        outputln!("Adding {}",
+                  Yellow.bold().paint(self.package().ident().to_string()));
+        // JW (and now also FN) TODO: We can't just set the run path for the entire process here,
```
reset: LOL! I guess I never thought how you represent notational buy-in before ;)

reset:
gif-keyboard-15913864449047670170

@fnichol just a few conversational bits that we should address before merging. Great work!

components/sup/src/command/start.rs
fnichol: Yeah, I agree with that, especially the issue of adding on rule exceptions that explode with *bigly* over time. In trying to avoid that I did 2 things:

* Leave any notion of a local artifact out of the `ServiceSpec`--the artifact's fully qualified ident is used for the `ServiceSpec`'s ident and that's it
* Treat the local artifact case quickly by naiively installing it to disk and moving on, in effect "boiling off" the fact that a local artifact was asked for which is why it never leaves the command starting stage and enters the library internals (I've been thinking of it similar to an environment variable--get it gone as soon as possible before you call too deep)

With these 2 tricks I don't see a way we'd enable an end user to provide a config file (mapping to a `ServiceSpec`) that has an artifact file in it or allow any update strategies to work (you asked to run that specific ident and nothing else).

components/sup/src/manager/mod.rs
fnichol: Yeah, I was really happy to find and use this! Only catch is that the call takes a mutable borrow on self, meaning that if `self.remove(&self)` needed to be `self.remove(&mut self)` we have a problem. I can't remember why I split it off (for sure a compiler suggestion that I wasn't ready to deal with), but I hope we can deal with that when we need to.

components/sup/src/manager/service/mod.rs
fnichol: That's super! I tried to preserve *existing behavior* as much as possible, so I'm proud of myself that I succeeded, heh 😄 

I agree, we didn't have this quite right. Is it easier to follow up with a commit (or PR) to make this correct so there's a commit and words around it, or roll it into this commit and rebase/push?

components/sup/src/manager/service/mod.rs
fnichol: We really do collectively own this codebase!

components/sup/src/manager/service/mod.rs
reset: @fnichol totally up to you, whatever your preference ;)

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/202830146).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/202830146).

PR #1795, opened 2017-02-18
Suitability

### LINKED ISSUE #1767: Elect 'healthiest' follower when leader goes down
When running clustered statefull services where follower nodes count on streaming replication to sync with a leader node we want to elect the follower that is 'furthest along' as the new leader.

Are there any thoughts as to how this can be achieved? Either as a work around or supported in the core. I'd be happy to brainstorm and hack on a solution.

bodymindarts:
Add support for suitability hook.
By printing a parsable u64 as the last line of stdout this hook allows services to report a priority by which they should be elected the new leader when an election is initiated.

Closes #1767

### DIFF
--- a/components/core/src/os/system/mod.rs
+++ b/components/core/src/os/system/mod.rs
@@ -52,7 +52,7 @@ impl fmt::Display for Architecture {
 pub enum Platform {
     Linux,
     Windows,
-    MacOS,
+    Darwin,
 }
 
 impl fmt::Display for Platform {
@@ -82,7 +82,7 @@ impl FromStr for Platform {
         match platform.as_ref() {
             "linux" => Ok(Platform::Linux),
             "windows" => Ok(Platform::Windows),
-            "macos" => Ok(Platform::MacOS),
+            "darwin" => Ok(Platform::Darwin),
             _ => return Err(Error::InvalidPlatform(value.to_string())),
         }
     }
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

reset:
This looks *excellent*! Per our video conversation, let's stream the output of a hook into a log file for each hook instead of storing it in a String in memory. The log should be truncated before each run so we just have the last output of the hook stored in there. Everything else should fall into place nicely :)

gif-keyboard-13847292863988949832

components/sup/src/manager/service/mod.rs
```diff
@@ -685,6 +685,21 @@ impl Service {
         false
     }
 
+    pub fn suitability(&mut self) -> Option<u64> {
+        if !self.initialized {
+            return None;
+        }
+        self.run_hook(HookType::Suitability)
+            .ok()
```
adamhjk: I like how terse this is, but do we actually want to throw away the error?

components/sup/src/manager/service/mod.rs
```diff
@@ -685,6 +685,21 @@ impl Service {
...
+            .as_ref()
+            .and_then(|hook_out| hook_out.stdout().lines().last())
+            .and_then(|line| line.trim().parse::<u64>().ok())
```
adamhjk: Similar question here - if we can't parse the line into a `u64`, we get none, and so the suitability rules go back to default. This can be fine, but we loose any chance to warn the user that the hook isn't written right.

adamhjk:
gif-keyboard-13697484317413336881

reset:
@bodymindarts if you want any help refactoring this after the changes I made to `hooks.rs` I can totally jump in with you or correct the merge conflicts for you - let me know!

PR #1804, opened 2017-02-22
Speed up blake2b hash generation during build

### LINKED ISSUE #1800: Slow hash generation for lots of files
When the package being built contains lots of files (in my test case, a nodejs app with 50k+ files, blake2b generation can be very slow due to (I think) shell overhead:

Timestamps show elapsed times:
```
00:02:47.404    xxx: Generating blake2b hashes of all files in the package
00:04:08.652    xxx: Generating signed metadata FILES
```

(not entirely sure why the times differ so much above and below, but the `user` time appears to match)

```sh
$ time find . -type f     | sort     | while read file; do hab pkg hash "$file"; done > ~/hashes2

real	4m1.489s
user	1m55.800s
sys	0m49.140s
```

vs a simple ruby script for comparison:

```sh
$ time find . -type f | sort | ruby -rblake2 -ne 'puts Blake2.hex(File.read($_.strip)) + "  " + $_;'  > ~/hashes-ruby-sorted

real	0m4.644s
user	0m3.784s
sys	0m0.844s
```

Obviously I'm using a different blake2b impl here, and it could be that the impl used by `hab` is just slow. Needs some more investigation.. 

Ubuntu 14.04.5 LTS + hab 0.17.0/20170214234052

chetan:
Fixes #1800 (see discussion there)

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -315,24 +315,8 @@ fn sub_pkg_hash(m: &ArgMatches) -> Result<()> {
             // read files from stdin
             let stdin = io::stdin();
             for line in stdin.lock().lines() {
-                match line {
-                    Ok(line) => {
-                        let file = line.trim_right();
-                        if Path::new(&file).is_file() {
-                            match command::pkg::hash::start(&file) {
-                                Ok(_) => (),
-                                Err(error) => {
-                                    writeln!(io::stderr(), "{}: error hashing file: {}", &file, error).unwrap();
-                                }
-                            }
-                        } else {
-                            writeln!(io::stderr(), "{}: No such file or directory", &file).unwrap();
-                        }
-                    },
-                    Err(error) => {
-                        writeln!(io::stderr(), "error: {}", error).unwrap();
-                    },
-                }
+                let file = line.unwrap();
+                try!(command::pkg::hash::start(file.trim_right()));
             }
             Ok(())
         },
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -27,6 +27,7 @@ extern crate log;
 use std::env;
 use std::ffi::OsString;
 use std::io::{self, Read};
+use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::thread;
@@ -304,10 +305,33 @@ fn sub_pkg_export(ui: &mut UI, m: &ArgMatches) -> Result<()> {
 }
 
 fn sub_pkg_hash(m: &ArgMatches) -> Result<()> {
-    let source = m.value_of("SOURCE").unwrap(); // Required via clap
-
     init();
-    command::pkg::hash::start(&source)
+    match m.value_of("SOURCE") {
+        Some(source) => {
+            // hash single file
+            command::pkg::hash::start(&source)
+        },
+        None => {
+            // read files from stdin
+            let stdin = io::stdin();
+            for line in stdin.lock().lines() {
+                match line {
+                    Ok(line) => {
+                        let file = line.trim_right();
+                        if Path::new(&file).is_file() {
+                            command::pkg::hash::start_multi(&file);
+                        } else {
+                            println!("File: '{}' cannot be found", &file);
+                        }
+                    },
+                    Err(error) => {
+                        println!("error: {}", error);
+                    },
+                }
+            }
+            Ok(())
+        },
+    }
 }
 
 fn sub_plan_init(ui: &mut UI, m: &ArgMatches) -> Result<()> {
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -321,11 +321,11 @@ fn sub_pkg_hash(m: &ArgMatches) -> Result<()> {
                         if Path::new(&file).is_file() {
                             command::pkg::hash::start_multi(&file);
                         } else {
-                            println!("File: '{}' cannot be found", &file);
+                            writeln!(io::stderr(), "{}: No such file or directory", &file);
                         }
                     },
                     Err(error) => {
-                        println!("error: {}", error);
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/main.rs
```diff
@@ -304,10 +305,33 @@ fn sub_pkg_export(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
+        },
+        None => {
+            // read files from stdin
+            let stdin = io::stdin();
+            for line in stdin.lock().lines() {
+                match line {
+                    Ok(line) => {
+                        let file = line.trim_right();
+                        if Path::new(&file).is_file() {
+                            command::pkg::hash::start_multi(&file);
```
mwrock: This feels just a bit awkward to me to have 2 functions (`start` and `start_multi`) where one differs just in output, but I see why you are doing it. I wonder if it would be better to simply call `hash::hash_file(&file)` here and then spit out our output.

chetan: Yeah, I agree. As I said in the other thread, if we change the output of the original `hab pkg hash` command then this could be simplified.

chetan: Updated in latest commit

components/hab/src/main.rs
```diff
@@ -304,10 +305,38 @@ fn sub_pkg_export(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
+                                    writeln!(io::stderr(), "{}: error hashing file: {}", &file, error).unwrap();
+                                }
+                            }
+                        } else {
+                            writeln!(io::stderr(), "{}: No such file or directory", &file).unwrap();
+                        }
+                    },
+                    Err(error) => {
+                        writeln!(io::stderr(), "error: {}", error).unwrap();
+                    },
```
mwrock: you can let the errors from `hash::start` bubble up and `main::start` will render them propperly.

chetan: The `Err` match on line 332 is actually from the lines() iterator. I'm not sure if it's really needed or if we can just `line.unwrap()` instead.

The `Err` match on 324 is needed to log individual errors within a list of files, if there was an issue hashing one of them. Not really sure the set of possible errors here..

chetan: Here's one possibility -

```sh
$ ls -l foo bar
---------- 1 chetan chetan 5 Feb 22 19:56 bar
-rw-rw-r-- 1 chetan chetan 5 Feb 22 19:56 foo
$ ls foo bar | ../../target/debug/hab pkg hash
bar: error hashing file: Permission denied (os error 13)
8b1533ef62b27202859d78676417a162156a53d466899fba7bbd928913f9ce9f  foo
```

mwrock: ahh i see. I'd still return a `Result` with the `err` in it instead of writing to stderr from this function. This kiips the actual writing of the error messages centralized using the `ui` in `main::start`

chetan: In that case processing would just halt at the error. I guess that's ok in this case since it is, _in theory_, a rare edge case.

chetan: Also, should a `file not found` return the error too?

mwrock: Shouldn't halt, returning the `Err` in the `Result` will get caught by https://github.com/chetan/habitat/blob/99f053b331c024c295eaa4471db9a5cd99b6a081/components/hab/src/main.rs#L71 and then rendered.

mwrock: I don't think you need to handle file not exists. https://github.com/chetan/habitat/blob/99f053b331c024c295eaa4471db9a5cd99b6a081/components/core/src/crypto/hash.rs#L32 should catch that.

chetan:
@mwrock ok, it's much simpler now. how's this?

components/hab/src/main.rs
```diff
@@ -304,10 +305,22 @@ fn sub_pkg_export(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
-    command::pkg::hash::start(&source)
+    match m.value_of("SOURCE") {
+        Some(source) => {
+            // hash single file
+            command::pkg::hash::start(&source)
+        },
+        None => {
+            // read files from stdin
+            let stdin = io::stdin();
+            for line in stdin.lock().lines() {
```
mwrock: This all is looking better! We should avoid the possible panic by using `for line in try!(stdin.lock().lines()) {` and dropping the `unwrap`.

smurawski:
@chetan You'll want to check if your `rustfmt` tool is up to date (0.7.1).

smurawski:
And thanks for the contribution @chetan. :)  Looking forward to getting this merged.  🎉

chetan:
Yup, my rustfmt is probably really old. Will update and push again

components/hab/src/main.rs
chetan: I think you meant `let file = try!(line);`? The other way didn't work as the `for` loop requires an iterator.

mwrock: yes! perfect.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/205044810).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/205044810).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/205044810).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/205044810).

reset:
@chetan wow, that's so much faster - thank you!

PR #1810, opened 2017-02-23
Refactor Hooks and add SmokeTest hook

reset:
This change stubs out a new hook to be run on service start after the
first initial successful health check. The Smoke Test hook is our first
long running hook which needs to be executed in a background thread
so I haven't hooked it up just yet. More refactoring is necessary to
the way we schedule a service to do work before the background runner
can be implemented.

In addition to the new hook this commit also refactors hooks into a
concrete struct implementing the new `Hook` trait eliminating the
HookType enum. This change allows us to move hook execution behaviour
into the hooks module. Running a hook now has an associated type for
it's return which can be massaged into any type. For example, the
`HealthCheckHook` and `SmokeTestHook` return a health/smoke check
instead of an i32. This will allow us to build more complex return
values from hooks and allows us to define them on the hook itself.

Lastly, hooks are no longer compiled on demand. This prior change
meant that people running a hook needed a mutable reference of the
Service struct. We now just compile after configuration and during
initialization.

gif-keyboard-3488498315577892481

### DIFF
--- a/components/sup/src/manager/service/hooks.rs
+++ b/components/sup/src/manager/service/hooks.rs
@@ -17,109 +17,348 @@ use std::fmt;
 use std::io::BufReader;
 use std::io::prelude::*;
 use std::path::{Path, PathBuf};
-use std::process::Child;
+use std::process::{Child, ExitStatus};
+use std::result;
 
 use hcore;
 use hcore::service::ServiceGroup;
+use serde::{Serialize, Serializer};
 
-use error::{Error, Result};
+use super::health;
+use error::Result;
 use manager::service::ServiceConfig;
 use supervisor::RuntimeConfig;
 use templating::Template;
 use util;
 
 pub const HOOK_PERMISSIONS: u32 = 0o755;
-pub const INIT_FILENAME: &'static str = "init";
-pub const HEALTHCHECK_FILENAME: &'static str = "health_check";
-pub const FILEUPDATED_FILENAME: &'static str = "file_updated";
-pub const RECONFIGURE_FILENAME: &'static str = "reconfigure";
-pub const SMOKETEST_FILENAME: &'static str = "smoke_test";
-pub const RUN_FILENAME: &'static str = "run";
-
 static LOGKEY: &'static str = "HK";
 
-#[derive(Debug, Serialize)]
-pub struct Hook {
-    pub htype: HookType,
-    pub template: PathBuf,
-    pub path: PathBuf,
-    pub user: String,
-    pub group: String,
+#[derive(Debug, Copy, Clone)]
+pub struct ExitCode(i32);
+
+impl Default for ExitCode {
+    fn default() -> ExitCode {
+        ExitCode(-1)
+    }
 }
 
-impl Hook {
-    pub fn new(htype: HookType,
-               template: PathBuf,
-               path: PathBuf,
-               user: String,
-               group: String)
-               -> Self {
-        Hook {
-            htype: htype,
-            template: template,
-            path: path,
-            user: user,
-            group: group,
+pub trait Hook: fmt::Debug + Sized {
+    type ExitValue: Default;
+
+    fn file_name() -> &'static str;
+
+    fn load<C, T>(service_group: &ServiceGroup, concrete_path: C, template_path: T) -> Option<Self>
+        where C: AsRef<Path>,
+              T: AsRef<Path>
+    {
+        let concrete = concrete_path.as_ref().join(Self::file_name());
+        let template = template_path.as_ref().join(Self::file_name());
+        match std::fs::metadata(&template) {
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -732,11 +706,6 @@ impl Service {
         Ok(())
     }
 
-    fn run_hook(&mut self, hook: HookType) -> Result<()> {
-        self.hooks.compile(&self.service_group, &self.config);
-        self.hooks.try_run(hook, &self.service_group)
-    }
-
     /// Update our own service rumor with a new configuration from the packages exported
```
mwrock: funny @reset, my hook compile PR earlier initially removed the `run_hook` function from `service/mod` and replaced all `run_hook` calls with `self.hooks.try_run` and I reverted it because I was nervous that handlebar templating may have changed data that needs to be recompiled prior to running the hook. Like for example, a member is confirmed dead changing the live members, if you do not recompile the templates on every run, you risk having stale members in your hook no?

mwrock: a recently expired peer would not trigger a reconfigure or initialize

mwrock: or would it?

bodymindarts: If there are config files that depend on those values a reconfigure would be triggered on the next `tick`. That gives a 1 second upper bound to how stale the hooks can be.

components/sup/src/manager/service/hooks.rs
```diff
@@ -17,116 +17,358 @@ use std::fmt;
...
+pub struct ExitCode(i32);
+
+impl Default for ExitCode {
+    fn default() -> ExitCode {
+        ExitCode(-1)
     }
+}
 
-    /// Run a compiled hook.
-    pub fn run(&self, service_group: &ServiceGroup) -> Result<()> {
```
bodymindarts: Do we really need to pass `ServiceGroup` around in all the functions. It appears that its only being used for output purposes. Can't we just pass it once on creation?

bodymindarts:
General question: Is it favourable to treat the `run` script like the other hooks? The handling of that script is all taken care of in https://github.com/habitat-sh/habitat/blob/master/components/sup/src/supervisor.rs. Couldn't we delete the run hook from the `hooks.rs` file?

components/sup/src/manager/service/mod.rs
bodymindarts: Perhaps restructuring [this code](https://github.com/habitat-sh/habitat/blob/42c3ff5678cb2c8a18a896138c98708e36da19d8/components/sup/src/manager/service/mod.rs#L418-L421) to ensure `reconfigure` is called not only if `self.config.write()` return `Some(true)` but also if `self.hooks.compile(&self.service_group, &self.config)` indicate a change has been persisted. I guess this would require hashing the compiled hooks like we do the config files.

mwrock: Turns out I have learned that changes to census data results in a config change and triggers a reconfigure which clears it all up for me and makes me fine with removing recompile before every hook execution.

mwrock:
I'm gonna go ahead and approve but @bodymindarts has good points worthy of revisiting!

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/204689779).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/204689779).

components/sup/src/manager/service/mod.rs
reset: Yeah, I was just hitting it hard with a hammer when I was making sure things were compiled on call. It changed the signatures of the functions to need to be mutable and made the API worse in the end. I'm pretty sure having the hooks compile in the two spots we are now is all the places it needs to be called in. We can add additional places at no cost because the caching is still in place when calling compile

components/sup/src/manager/service/hooks.rs
reset: We would need to clone it to pass it on creation. In Rust you can't have a member that holds a reference to another of your members. This is the recommended way out - to pass by ref when calling

reset:
@bodymindarts I think it's a good question to ask if the HookTable should try and manage the run hook or not. I'm not sure right now, it feels like that hook is for sure special, but I haven't run into anything yet that made me want to move it off the HookTable. Do you have an idea of what the code would look like if we moved it off? I'm not married to either way ;)

Issue #1579, opened 2016-12-14
RFC: a mechanism to declare runtime environment variables at build time


robbkidd:
### Problem:

`$(hab pkg path origin/thingie)` in a hook does not guarantee that that the path returned is the version the hook's package was built against. The presence of this command in a hook should be considered a code smell.

```
habitat-sh/core-plans > ack --files-with-matches "hab pkg path" | ack "hooks" | wc -l
      10
```

First notion I can find on this problem is [a forum post](https://forums.habitat.sh/t/allow-env-vars-to-be-passed-from-plan-to-runtime/264) from @elliott-davis.

### Affects: 

Anyone running multiple habitat packages on a single host, the bare-metal scenario. Should not affect those running habitat packages exported to containers.

### Solutions Today:

* Paths can be computed at build time and ...
    * written out to wrapper scripts that act like binstubs for the real binaries, see this pattern in `wrap_with_cert_path` (used for [cargo-nightly](https://github.com/habitat-sh/core-plans/blob/d882abe701aaadc0bd60364df2af38ef0a08676b/cargo-nightly/plan.sh#L40-L54) and [rust](https://github.com/habitat-sh/core-plans/blob/d882abe701aaadc0bd60364df2af38ef0a08676b/rust/plan.sh#L91-L109)) and `write_shim` (used for [github_changelog_generator](https://github.com/habitat-sh/core-plans/blob/d882abe701aaadc0bd60364df2af38ef0a08676b/github_changelog_generator/plan.sh#L45-L59))
    * written out to a file that is sourced by hooks, see notional work on packaging Supermarket with Habitat ([write out in plan](https://github.com/chef/supermarket/blob/a7133c66182aa9410f6f5872170b136497e3a677/habitat/supermarket/plan.sh#L139-L150) and then [source it in a hook](https://github.com/chef/supermarket/blob/a7133c66182aa9410f6f5872170b136497e3a677/habitat/supermarket/hooks/init#L4))

### Solution Idea

Throwing a noodle at the wall:

* A `do_pkg_exports` function that would write out given exports from a plan to an `EXPORTS` file at `$pkg_prefix`.
    * The generator for the default run hook would source this `EXPORTS` file.
    * A `make_binstub` (similar to `fix_interpreter`) to be called during build/install to rename an original binary (like in `wrap_with_cert_path` & `write_shim`) and write out a binstub/wrapper/shim to the original binary's location that sources the `EXPORTS` file.

[...]

PR #1830, opened 2017-02-27
pkg_env: Add support for plan environment variable definitions

georgemarshall:
This is an attempt to solve environment variable definitions as described in #1579. It is also the next iteration of my PoC [habitat/pythonpath](https://github.com/georgemarshall/habitat/tree/pythonpath) branch and [python-plans](https://github.com/georgemarshall/python-plans).

- [x] [core] parsing `ENVIRONMENT` metadata
- [x] [core] cascade support for `ENVIRONMENT`
- [x] [core] compatibility support for some older metadata files
- [x] [hab] export command for runtime environment
- [x] [hab] support for setting runtime environment
- [x] [plan-build] `pkg_env`/`pkg_build_env` vars and helper functions
- [x] [plan-build] `ENVIRONMENT` and `BUILD_ENVIRONMENT` metadata files
- [x] [plan-build] parsing `ENVIRONMENT` and `BUILD_ENVIRONMENT` metadata
- [x] [plan-build] cascade support for `ENVIRONMENT` and `BUILD_ENVIRONMENT`
- [x] [plan-build] compatibility support for some older metadata files
- [x] [sup] support for setting runtime environment

**punted items**
- [plan-build] refactor _set_environment to use a bottom-up approach instead of top-down
- [plan-build] add overwrite support for environment variables that are already defined.
- [studio] add export for environment

## Plan usage

```bash
pkg_name=myapp
pkg_deps=(python/django)
pkg_bin_dirs=('bin') # This is an alias `add_path_env 'PATH' ${pkg_bin_dirs[@]}`
pkg_env_sep=(
  ['PYTHONPATH']=':' # separator used for cascading other artifacts
)
pkg_env=(
  ['DJANGO_SETTINGS_MODULE']='myapp.settings' # This will not cascade
)
# pkg_build_env=() # Build time envs

do_begin() {
  # Sets PYTHONPATH and adds $pkg_prefix
  add_path_env 'PYTHONPATH' 'lib/python3.6/site-packages'
}

# export environment variables in pkg_env/pkg_build_env and all deps ENVIRONMENT/BUILD_ENVIRONMENT
# _build_environment called
# ...
# do_prepare_wrapper

do_install() {
  # Will fail because DJANGO_SETTINGS_MODULE has already been set
  add_env 'DJANGO_SETTINGS_MODULE' 'myapp.new_settings'

  # This is set after _build_environment and will not be exported into current shell
  add_env 'HTTPS' 1
}

# Render the final values of pkg_env_sep/pkg_env/pkg_build_env
# _build_metadata
```

### ENVIRONMENT_SEP
This file contains the Internal Field Separator (IFS) used for joining/cascading environment variables. If a variable is not set, further attempts to set a value for its accompanying environment variable will result in an error. At the moment there is a limitation when using whitespace characters such as ` ` when setting an IFS.
```
PATH=:
PYTHONPATH=:
```

### ENVIRONMENT/BUILD_ENVIRONMENT
These files contain the names and values of the packages specific environment variables. At the moment there are limitations when using `\n` in a definition.
```
PATH=/hab/pkgs/georgemarshall/hab-plan-build/0.19.0/20170313042443/bin:/hab/pkgs/georgemarshall/hab-plan-build/0.19.0/20170313042443/sbin
PYTHONPATH=/hab/pkgs/georgemarshall/hab-plan-build/0.19.0/20170313042443/lib:/hab/pkgs/georgemarshall/hab-plan-build/0.19.0/20170313042443/lib64
```


thesentinels:

Thank you for contributing!

reset:
@georgemarshall glad to see somebody in the community picking this up! I just wanted to share a few concerns that I had thought about when thinking about how to write this feature - you might already even have these solved in your head ;)

1. What do we do with environment variables which are a list of things and not a simple key/value. Do we append to them? If so, which appends last and has priority over others?
1. How do we handle collisions with environment variables which aren't a list?

georgemarshall:
1. In terms of "lists" environmental variables do not support anything beyond strings as far as I know. So if you wanted support that it would require a "join" function.
```bash
join_path ':' ('my/path', 'myother/path')
```
2. In terms of type collisions I do not anticipate this since values will will all be rendered to a string such as above and also in the same manor `PATH` is assembled in each package.
3. The part that gets tricky is deciding what variables get cascaded or replaced. In the case when we cascade; do we append, prepend or allow the plan to even put it somewhere in the middle? Could we assume everything that ends with `PATH` such as `PYTHONPATH` can be joined with `:`?
4. In terms of @adamhjk proposal, it may be the most viable approach to handling the ugliness of defining cascade and overwrite behavior. It may lead to some ugliness with the `ENVIRONMENT` file, but the goal is to be rust/bash parsable.
```bash
# cacerts plan
build() {
  add_pkg_svc_export SSL_CERT_FILE $pkg_prefix/ssl/certs/cacert.pem
  add_pkg_build_export SSL_CERT_FILE $pkg_prefix/ssl/certs/cacert.pem
}
```

reset:
@georgemarshall 
1. Yep! That's exactly what I mean - it's just key/value where the value is stringy, but the value might have self describing structure in it. In the case of a path it's a list becuase it's got the `:` joining portions together.
2. The collision that I'm worried about is when two packages both want to set the same key `FOO` to a different value and `FOO` isn't a "list" type of environment variable.

@adamhjk's proposal would work and it could be a great starting place, but I'm worried that without setting up a very minimal type contract around an environment variable it could result in weird runtime errors which we want to avoid.

For example: if one package sets `FOO` as a single value while another attempts to append things to `FOO` we might encounter a runtime issue with the first package who didn't expect there to be a `:` separated list of values.

A minimal type contract here might just be two things: Do I expect the value of this env var to be a single value or a list of values separated by a `:`. There's other ways to represent a stringy-list but I don't think I've ever seen anyone use anything other than a `:` at the env var level.

georgemarshall:
I am thinking about adding a special marker for definitions that cascade. The marker would be something like `<|` and `|>` containing the join characters in the middle. While this can be done by hand, it would be best to use the helper methods `add_env`, `add_path_env`, `add_build_env` and `add_path_build_env`. In the example bellow, the markers are placed in the values so that it does not interfere with key lookups.

```bash
pkg_env=(
  PERL5LIB='<|:|>lib/perl5'
  PYTHONPATH='<|:|>lib/python3.6/site-packages'
)
```

adamhjk:
I think you can handle this with shell just fine. 

```
add_env PERL5LIB $PERL5LIB:$pkg_prefix/lib/perl5
```

If I wanted to stomp `$PERL5LIB`, I would just stomp it. No need to over-complicate it - this is what any shell programmer would expect.

adamhjk:
(I really think you don't need to put guard-rails in shell that aren't in shell - it's shell, and thats okay)

adamhjk:
Also - @georgemarshall - this is awesome that you're implementing this.

gif-keyboard-7251744493274087411

reset:
@adamhjk I think we do need to have some sort of protection in the build system for mutating environment variables which were set from a dependency. It's a protection we can provide to avoid runtime issues. At the very least a warning letting you know that you've changed an environment variable that something else has set?

georgemarshall:
After putting more thought into it, the "football" token does not seem to be the most practical approach. Maybe the cascade functionality should just work off a separate associative array. So under the hood it could be defined as:
```bash
_env_join=(
  LDFLAGS=' '
  PATH=':'
  PERL5LIB=':'
  PYTHONPATH=':'
)
```
There would be an associated helper functions or it could be apart of the ones I already am thinking of. In general it allows for joining values between packages. Granted it would be redundant across all packages, but it also could be used to ensure a consensus. So you could not define `_env_join[PATH]=')'('` while everyone else uses `:`. Maybe some keys have a predefined static values that cannot be changed.

I could see something such as:
```bash
add_path_env PATH ':' ('bin', 'sbin')
# or
add_path_env PATH ('bin', 'sbin')
```
The join/cascade would only occur if it was set in `_env_join`, otherwise it would be overwritten by parent packages. So it could be a topdown approach starting with the root plan, then in order of `pkg_deps` similar to the current way `PATH` is built.
```bash
pkg_name=hab
pkg_origin=core
pkg_deps=(
  core/bash
  core/binutils
  core/bzip2
)

do_build() {
  add_path_env PATH ':' ('bin', 'sbin')
  # or
  add_env PATH ':' "$pkg_prefix/core/hab/bin:$pkg_prefix/core/hab/sbin"
}
# Runtime PATH
# PATH='core/hab/bin:core/hab/sbin:core/bash/bin:core/binutils/bin:core/bzip2/bin'
```
`PYTHONPATH` works identically so I want to make this functionality generic.

* * *

Hopefully my `PATH` illustration shows that `add_env PERL5LIB $PERL5LIB:$pkg_prefix/lib/perl5` may not be viable due to the values of previous packages polluting the next.

georgemarshall:
@adamhjk and @reset can I get some 👀 on `plan-build`? I believe that part to be feature complete.

georgemarshall:
@fnichol can I get some 👀 on this?

fnichol:
I'll take a good pass on this shortly, thank you for all your hard work on this!

georgemarshall:
After testing on my end, everything appears to be working as intended.
Tested with https://github.com/georgemarshall/python-plans

eeyun:
It might not be possible but, would be _awesome_ if we could get eyes on this today before our release tomorrow. @reset or @fnichol or @adamhjk if any of you have time for review today :smile:

components/core/src/package/install.rs
```diff
@@ -318,35 +375,124 @@ impl PackageInstall {
...
+        let tdeps = self.load_tdeps()?;
         for dep in tdeps.iter() {
             if idents.contains(dep.ident()) {
                 continue;
             }
-            let mut p = try!(dep.paths());
-            run_paths.append(&mut p);
+            let env = dep.environment()?;
+            let env_sep = dep.environment_sep()?;
+            if !env.is_empty() {
```
bdangit: this code block looks similar to https://github.com/habitat-sh/habitat/pull/1830/files#diff-b9b27ac18067cb62ebad852e9abc0b10R413

Perhaps make this DRY?

components/core/src/package/install.rs
```diff
@@ -251,6 +252,62 @@ impl PackageInstall {
         self.read_deps(MetaFile::TDeps)
     }
 
+    /// Returns a Rust representation of the mappings defined by the `pkg_env` plan variable.
+    ///
+    /// # Failures
+    ///
+    /// * The package contains a Environment metafile but it could not be read or it was malformed.
+    pub fn environment(&self) -> Result<HashMap<String, String>> {
```
bdangit: The logic is very similiar to https://github.com/habitat-sh/habitat/pull/1830/files#diff-b9b27ac18067cb62ebad852e9abc0b10R288 except the object is for EnvironmentSep vs Environment

This could be DRY.

adamhjk:
This looks good to me - I'm +1 on merging. 

gif-keyboard-14998113990778886694

reset:
This is *super* good work @georgemarshall. Let's get this merged in for the release and try to get the punted items which involve the package protocol (metafiles) and the UX of the feature sorted out if we can by tomorrow along with some docs to get people going with this. If we can't by tomorrow, no worries, we can get it done for the release afterwards ;)

gif-keyboard-4989721567643263133

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/218993633).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/218993633).

PR #1835, opened 2017-02-28
Make depot upload and download target aware

mwrock:
This PR allows uploads and downloads to and from the depot to be able to filter on target. Uploads are initially saved to a temp file where the target is discovered and then the file is moved to the appropriate file name based on its target. Downloads will find the latest package for the target of the caller by adding target redis indexes. The supported target is now a `Vector` instead of a single target.

Note that this does not implement target awareness for channels. The channels persistence code is currently completely in memory and it makes no sense to go to the trouble to refactor this code to be filterable by channel when it will be replaced with the new data persistence and especially when channels are not currently being used.


thesentinels:

Thank you for contributing!

components/builder-depot/src/server.rs
```diff
@@ -517,40 +505,43 @@ fn upload_package(req: &mut Request) -> IronResult<Response> {
...
         debug!("Unsupported package platform or architecture {}.",
                target_from_artifact);
-        remove_file(&archive.path).expect("Unable to remove temp file");
         return Ok(Response::with(status::NotImplemented));
     };
 
+    match depot.datastore.packages.find(&ident) {
+        Ok(_) |
+        Err(dbcache::Error::EntityNotFound) => {
+            if let Some(_) = depot.archive(&ident, &target_from_artifact) {
```
raskchanky: This is a total nit and not important, but since you don't care about the value inside the `Some`, you could also write this as:

`if depot.archive(&ident, &target_from_artifact).is_some()`

components/builder-protocol/protocols/depotsrv.proto
```diff
@@ -11,10 +11,11 @@ message Package {
...
   required string manifest = 3;
-  repeated PackageIdent deps = 4;
-  repeated PackageIdent tdeps = 5;
-  repeated uint32 exposes = 6 [packed=true];
-  optional string config = 7;
+  required string target = 4;
+  repeated PackageIdent deps = 5;
+  repeated PackageIdent tdeps = 6;
+  repeated uint32 exposes = 7 [packed=true];
+  optional string config = 8;
```
raskchanky: I might be way off about this, but I want to say that when you modify protobufs, you should add new fields to the end of the definition, so that the order of the existing fields is preserved.  I might be mis-remembering a conversation I had with @reset so it's possible this is no big deal.

components/core/src/config.rs
```diff
@@ -1428,59 +1435,51 @@ mod test {
...
     }
 
     #[test]
     fn parse_into_package_target_invalid_target() {
-        let mut actual = PackageTarget::default();
+        let mut actual = vec![PackageTarget::new(Platform::Linux, Architecture::X86_64),
+                              PackageTarget::new(Platform::Windows, Architecture::X86_64)];
         let toml = toml_from_str(r#"
-            field = "nope"
+            field = ["x86_64-linux", "LOVE IS ALL AROUND YOU"]
```
raskchanky: So good 😂

raskchanky:
Looks great to me overall.  You're welcome to ignore my nit 😁 I just had one question about the ordering of the fields in the protobuf but otherwise awesome!

components/builder-protocol/protocols/depotsrv.proto
mwrock: I just did some research on this and it seems field ordering really does not matter and unrecognized fields will simply be ignored. I was operating under the assumption that maybe required fields should be first, but it seems that is not the case either. @reset please confirm.

components/builder-depot/src/server.rs
mwrock: i like it

components/builder-protocol/protocols/depotsrv.proto
mwrock: ok per our discusion in standup, I moved target to the end along with its field number, recompiled and retested.

smurawski:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/206763293).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/206763293).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/206763293).

PR #1888, opened 2017-03-02
corrected travis env var value

eeyun:
So this checks whether their is a value other than "false" in $TRAVIS_PULL_REQUEST which should give us a sane check to determine whether to deploy.


thesentinels:

Thank you for contributing!

PR #1927, opened 2017-03-08
Port the vault to PostgreSQL

adamhjk:
* This moves the `builder-vault` to using PostgreSQL as the data store
* Updates the Project protocol buffer definitions to reflect their being
  tied to origin
* Updates the builder API to use the new protocols
* Moves the authorization check on invitation accept into the API
* Updates the jobsrv to match the new project data
* Updates the scheduler to match the new project data
* Updates the worker to match the new project data

Sorry for the code-bomb. :)

gif-keyboard-6346447494100748124


thesentinels:

Thank you for contributing!

components/builder-vault/src/config.rs
```diff
@@ -62,8 +70,24 @@ impl ConfigFile for Config {
...
+        let mut connection_db = String::new();
+        try!(toml.parse_into("cfg.datastore_connection_db", &mut connection_db));
+
+        cfg.datastore_connection_url = format!("postgresql://{}@{}/{}",
+                                               connection_user,
+                                               connection_address,
+                                               connection_db);
+        try!(toml.parse_into("cfg.datastore_connection_retry_ms",
+                             &mut cfg.datastore_connection_retry_ms));
+        let mut timeout_seconds = 0;
```
chefsalim: This should be initialized to 3600 (default value), not 0, since we want a sane value even if we are parsing only a partial set of config values.

components/builder-protocol/protocols/jobsrv.proto
```diff
@@ -42,5 +42,5 @@ message JobGet {
 
 message JobSpec {
   required uint64 owner_id = 1;
-  required vault.Project project = 2;
+  required vault.OriginProject project = 2;
```
chefsalim: Switch these to optional as well while we're in this code?

components/builder-protocol/protocols/vault.proto
```diff
@@ -124,33 +133,33 @@ message OriginSecretKeyGet {
...
-  required Project project = 1;
+message OriginProjectUpdate {
+  optional uint64 requestor_id = 1;
+  optional OriginProject project = 2;
 }
 
-message ProjectDelete {
-  required string id = 1;
+message OriginProjectDelete {
+  required string name = 1;
```
chefsalim: Switch these to optional?

components/builder-vault/src/config.rs
adamhjk: Nice catch! that was a subtle bug, actually.

components/builder-protocol/protocols/jobsrv.proto
adamhjk: Yeah - I'll just clean the whole thing up.

components/builder-protocol/protocols/vault.proto
adamhjk: Yar

adamhjk:
gif-keyboard-3488230272478315025

Just pushed changes that address everything @chefsalim pointed out - @reset, you want to do a review pass?

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/209475540).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/209475540).

components/builder-api/src/http/handlers.rs
```diff
@@ -162,6 +163,25 @@ pub fn list_user_origins(req: &mut Request) -> IronResult<Response> {
     }
 }
 
+pub fn validate_origin_invitation(conn: &mut BrokerConn,
+                                  req: &OriginInvitationAcceptRequest)
+                                  -> IronResult<bool> {
+    let mut request = OriginInvitationValidateRequest::new();
+    request.set_invite_id(req.get_invite_id());
+    request.set_account_accepting_request(req.get_account_accepting_request());
+    match conn.route::<OriginInvitationValidateRequest, OriginInvitationValidateResponse>(&request) {
```
reset: Boolean responses can be represented as `NetOk`, no need to define a `XResponse`. The bad case should be represented as a `NetErr`

components/builder-api/src/http/handlers.rs
```diff
@@ -176,6 +196,11 @@ pub fn accept_invitation(req: &mut Request) -> IronResult<Response> {
     }
 
     let mut conn = Broker::connect().unwrap();
+
+    if !try!(validate_origin_invitation(&mut conn, &request)) {
```
reset: Yes! All authentication and permissions should be checked on the gateway nodes and not piped through to the services themselves. We can *trust all* messages once they get past a gateway

components/builder-db/src/migration.rs
```diff
@@ -18,11 +18,15 @@ use pool::Pool;
...
     pool: &'a Pool,
+    sequence_number: i64,
 }
 
 impl<'a> Migrator<'a> {
     pub fn new(pool: &'a Pool) -> Migrator {
-        Migrator { pool: pool }
+        Migrator {
+            pool: pool,
+            sequence_number: 1,
```
reset: Great improvement here

components/builder-protocol/protocols/vault.proto
```diff
@@ -56,101 +56,110 @@ message CheckOriginAccessRequest {
...
 }
 
-message Project {
-  required string id = 1;
-  required uint64 owner_id = 2;
-  required string plan_path = 3;
-  oneof vcs {
-    VCSGit git = 4;
-  }
+message OriginProject {
```
reset: The `id` was `{name}/{origin}` before, this change means the messages will be slightly larger which is OK, but we need to assert that the name and origin don't differ from the combined field.

We also lose the typing around the VCS so we'll need to validate in the application code as well

reset:
gif-keyboard-9413707485692314217

Can't wait for this to hit acceptance!

components/builder-api/src/http/handlers.rs
adamhjk: My question here was if we wanted `NetErr` to really be only for errors, and a false response isn't necessarily an error.

components/builder-protocol/protocols/vault.proto
adamhjk: They can't differ, as the database sets it for you from the component parts. You can set it all day to bullshit, and it'll get set right on re-read. :)

I did shoot the typing on VCS in the head - mostly because there wasn't any real benefit to doing it in the database layer anymore, and making it stringy-typed meant it was going to be easier to extend. We could go back the other way.

Issue #1896, opened 2017-03-03
Add habitat to homebrew


bodymindarts:
I tried adding habitat to the S&W brew tap but have difficulties finding a link that will deterministically download a specific version and is easily discoverable.

smith:
It would not qualify as "easily discoverable", but https://github.com/habitat-sh/habitat/blob/master/components/hab/install.sh takes these options:

```
FLAGS:
    -c    Specifies a channel [values: stable, unstable] [default: stable]
    -h    Prints help information
    -v    Specifies a version (ex: 0.15.0/20161222215311)
```

bodymindarts:
interesting. That script can perform install of a specific version but discovering the `0.15.0/20161222215311` part is the problem. How do I can I programatically get the timestamp of `0.17.0` for example.

smith:
I think that what would need to happen here are improvements to the installer script.

If you type `install.sh -v 0.17.0` you get the exact same result if you type `install.sh -v 🐣` or any version that does not match an exact version/release that exists on Bintray.

The two things that need to be fixed that may get this working for you are:

* Give a failure message and exit non-zero if the version you are asking for is not found
* Allow specifying only a version without a release and make the program get the latest release for that version

eeyun:
I agree here. I think this could be chocked up to a change in the "expected behavior" realm. 

I'd take this a step further and say we ought to get the install.sh available as a  first-class citizen with instructions on how to use it via the website.

fnichol:
@bodymindarts All Linux `core/hab` [releases](https://app.habitat.sh/#/pkgs/core/hab) after June 2016 (when we launched) will have a corresponding Bintray version. Unfortunately our Depot currently doesn't have support for multiple platform/target packages (i.e. linux/64bit vs mac/64bit). We are working towards this however, and once we've landed the requisite work, you'll be able to use a Depot API endpoint for this query for all the Mac and Windows packages too. Once the multi-target Depot support is working we should be able to work towards eliminating Bintray as the initial download point and fold that into the Depot itself.

The reason that the [`install.sh` script](https://github.com/habitat-sh/habitat/blob/master/components/hab/install.sh) only takes a `VERION/RELEASE` "version" string is that each Bintray version is more like a compound key of a Habitat version string and release string joined with a dash (this was the closest we could get at the time to fit the notion of Habitat versioning into the Bintray API).

[...]

PR #1929, opened 2017-03-09
Added support for installing habitat from a semi-qualified version number

eeyun:
This is probably a dirty hack. Obviously temporary. Should run on busybox.

This is to respond to some of the issues presented in https://github.com/habitat-sh/habitat/issues/1896

### DIFF
--- a/components/hab/install.sh
+++ b/components/hab/install.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+# shellcheck disable=SC2039
 #
 # Copyright (c) 2010-2016 Chef Software, Inc. and/or applicable contributors
 #
@@ -21,6 +22,7 @@ set -eu
 if [ -n "${DEBUG:-}" ]; then set -x; fi
 
 BT_ROOT="https://api.bintray.com/content/habitat"
+BT_SEARCH="https://api.bintray.com/packages/habitat"
 
 main() {
   # Use stable Bintray channel by default
@@ -52,6 +54,7 @@ main() {
   info "Installing Habitat 'hab' program"
   create_workdir
   get_platform
+  get_version
   download_archive
   verify_archive
   extract_archive
@@ -79,7 +82,7 @@ USAGE:
 FLAGS:
     -c    Specifies a channel [values: stable, unstable] [default: stable]
     -h    Prints help information
-    -v    Specifies a version (ex: 0.15.0/20161222215311)
+    -v    Specifies a version (ex: 0.15.0, 0.15.0/20161222215311)
 
 USAGE
 }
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

eeyun:
Wanted to update this with more context for reviewers.

More or less we take the version provided and check whether it's formed in an expected way. If it is then we pass that along to the standard process. If it's not, then we take the version number provided and make a call to a different BT API and compare the available versions to the version provided. If we find one that matches the semi-qualified version then we pull the fully qualified version and pass that along to the standard process.

What I don't like about it: It's a _lot_ of string manipulation in bash, which feels bad. A tool like json_pp or jq would be ideal in the case of manipulating the returned json but I hesitated to bundle a full json library or perl script into what should eventually become a curlbash.

fnichol:
We're getting pretty darn close on this--we think we have a mildly evil `sed` that'll operate on the minified JSON data to get us what we need and with the input cases (short version with one match, short version with multiple matches, version that isn't present, invalid string)

eeyun:
Ok I've plugged in the changes that fletcher and I were pairing on this afternoon.  We now have a sed command to strip out our version numbers rather than awk, and this switch cleared up 2x bugs.

First: awk on busybox had slightly different functionality which caused the previous data manipulation to return the wrong things.
Second: The awk manipulation had a bug where it was possible for an incorrect version to be returned

This should work for as long as the APIs on BinTray continue to return minified json. If they switch that up, we'll need to change the sed script but the rest of the logic should remain sound.

fnichol:
I added some final changes to the commit and I think we're really, really good here now:

* Updated the help message to show you can use a short version (ex: `"0.15.0"`)
* Fixed a bug in our trap cleanup code that didn't properly propagate the exit code (at least the second time I've made this exact mistake)
* Called `get_version` from `main()` to make it clearer when we determine our version
* Removed the `for` loop over 1 variable in `get_version()`--now it's a simple if/else
* Added a couple of `info` lines in `get_version()` to explain why we're making API calls to Bintray and what the value of our resulting Bintray version string
* Moved the location of `get_version()` to be ordered with the other top-level functions called by `main()`
* Wrapped the contents of `get_version()` to 80-chars (our shell code style)
* Sorted the `need_cmd` calls in `get_version()` lexically and added a missing one for `grep`
* Made the path to the version.json file absolute (i.e. it's set to `${workdir}/version.json`) to make future debugging via `DEBUG=1 sh install.sh` calls easier to trace
* Ran through `shellcheck` to ensure it continues to lint well (also added a couple of explicit lint ignores that are known-good in our situation)
* Squashed the commits down to 1 and rebased against master

Trust me, that's not as much as you'd think, mostly small changes for future maintenance ease.

I'm all 👍 on this now, we should be good to merge!!

reset:
@thesentinels approve

thesentinels:

reset:
gif-keyboard-17792092523274989603

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/209871582).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/209871582).

Issue #1951, opened 2017-03-14
Document "blueprints" (real name TBD) Feature Usage


rhass:

rhass:
WIP: https://gist.github.com/rhass/7e4c7ade90539b8a97356b90748fb183

rhass:
This is ready for review! :smile:

mhedgpeth:
Have you considered making the terminology a bit more domain specific?

I.e.:

builder = how things get built
for example:
builder = java-maven-tomcat

A builder could also be:
builder = bash

Which is what it is now

You could also have:

source_retriever = github
github_source_url = https://github.com/mhedgpeth/cafe
github_branch = master
github_revision_id = 438290fjdklsafdj;afdsa

I think the product should "feel" like blueprints are a first class, default citizen. Which means you call it what it is in a DDD sense, not add a concept/metaphor (which is what you do when you add on to an existing domain language)

adamhjk:
What we have is a single syntax for writing plans on a given platform, written in the native scripting language of that platform.

We then provide `scaffolding` for as many different kinds of applications as you want - these are published as normal habitat packages, making it easy to ship new scaffolding types. 

What's important here is how the experience degrades - you need to be able to go from a no touch, fully declarative experience, to one that is "normal go18, but we use..." by simply overriding a callback. This is in contrast to writing a full DSL - what we learned early in the development of habitat is that for people who are already used to Chef, it feels great. But the abstraction actually doesn't buy you much - when we got rid of the DSL, you started being able to cut-and-paste directions from the upstream into the packaging system itself, and given the artifact as immutable, you don't need much of the behavior you do in something like Chef.

reset:
We've got the initial docs done and additional improvements coming down the pipe, gonna close this one out for now in favor of what is on our tracker

PR #1959, opened 2017-03-15
Add documentation for blueprints.

rhass:
This is the initial documentation and specification for the blueprints feature.

Relates to issue #1951

### DIFF
--- a/www/source/docs/blueprints.md
+++ b/www/source/docs/blueprints.md
@@ -1,0 +1,66 @@
+# Blueprints
+Blueprints are default implementations of the build phases and runtime hooks for your application. They also provide some default health check hooks where appropriate to ensure your application is functioning reliably. Customized Blueprints can be created to facilitate re-usability of common patterns in your organization for developing, building, and running your applications.
+
+## Getting Started
+### New Projects
+Simply run `hab plan init` to leverage blueprints. Habitat will look in your application code for known project types and the appropriate `pkg_blueprint`.
+
+If you would like to specify the desired blueprint to use, you can run `hab plan init --blueprint core/java8-maven-tomcat`
+
+The resulting plan will contain all the available callbacks and variables relevant to the blueprint type.
+
+### Existing Plans
+
+To begin using Blueprints, you will need to add the appropriate `pkg_blueprints` 
+
+```bash
+pkg_name="my_java_app"
+....
+blueprint="core/blueprint-java8-maven-tomcat"
+```
+
+### Available Blueprints
+
+* `core/blueprint-golang`
+* `core/blueprint-ruby-mri2.2.6-rails-unicorn`
+* `core/blueprint-ruby-mri2.3.4-rails-unicorn`
+* `core/blueprint-ruby-mri2.4.0-rails-unicorn`
+* `core/blueprint-java8-maven-tomcat`
+* `core/blueprint-java8-maven-jboss`
+* `core/blueprint-java8-gradle-tomcat`
+* `core/blueprint-java8-gradle-jboss`
+* `core/blueprint-java7-ant`
+* `core/blueprint-python-unicorn`
+
+## Callbacks
+Each blueprint defines a set of callbacks which are unique to the blueprint type. This allows you to complex projects which may need to leverage different blueprints as needed within your project. Each of these callbacks define overrides of base level callbacks.
+
+### ruby _(This will likely live in the blueprint README.md)_
+#### do_blueprint_bundler
+#### do_blueprint_rake
+#### do_blueprint_rack
+#### do_blueprint_more_ruby_stuff 
+
+## Variables
+These are variables which each blueprint honors, allowing the plan author to consume the value to override a particular behavior. 
+### ruby
+:warning: _This section will likely live in the Ruby blueprint README.md and is here for feature clarification_ :warning:
+* pkg_ruby_build_cmd - Overrides default binary and path.
+* pkg_ruby_install_cmd
+* pkg_ruby_bundle_cmd
+* pkg_ruby_run_opts  - This functions similar to `pkg_svc_run` as the result of using this option will intelligently generate a hook file for you. However, you also can create the [hook file](#Run-Hooks) directly if you need to have more control over the resulting output.
+
+## Run Hooks
+#### Health Checks
+Each blueprint comes with a default health check for your application.
+
+*TODO: Write the directions*
+
+## Build Artifacts
+### Auto-Ingestion
+Each blueprint defines a default value for `pkg_source` in [URI syntax](https://tools.ietf.org/html/rfc3986). You can override this value within your plan should you application source live in a non-standard location.
+
+## Blueprints Internals
+Each Blueprint defines it's own `pkg_build_deps` and `pkg_deps` arrays with any required pckages and is merged into the TDEPS (transitive dependencies) as it would any other package.
+
+


thesentinels:

Thank you for contributing!

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
+# Blueprints
+Blueprints are default implementations of the build and runtime phases for your application. They also provide some default health check hooks where appropriate to ensure your application functioning reliably. Customized Blueprints can be created to facilitate reusability of common patterns in your organization for developing, building, and running your applications.
```
raskchanky: I think you need "is" between "application" and "functioning".

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+pkg_blueprints=("core/blueprints/java8-maven-tomcat")
+               ^     ^^^^^^^^^^^                    ^
+               |______|_______|_____________________|
+                 Arrays and additional namespace TBD -- just random thoughts at the moment.
+```
+* add `auto` package to `build_deps`
+* set toolchain variable
+
+#### Optional Variables
+* `pkg_src` - now accepts URI notation. Each blueprint contains a default URI which is relevant to the type of blueprint being used.
```
reset: I don't think this is inherently part of blueprints - this is just an overall quality of life improvement we need to make. @Hoverbear started the work [here](https://github.com/habitat-sh/habitat/pull/1495) already for us actually!

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+*  `pkg_blueprint_opts` - Options which override some default runtime parameters. Eg: `--env DEV`. This operates similar to `pkg_svc_run` as the result of using this option will intelligently generate a hook file for you. However, you also can create the [hook file](#Run-Hooks) directly if you need to have more control over the resulting output.
```
reset: I think `pkg_blueprint_opts` ends up actually being represented as overridable package variables and callbacks  similar to my [comment above](https://github.com/habitat-sh/habitat/pull/1959/files#r106311922). I know that this was something I quickly wrote in my editor as an example - I think I totally got it wrong after thinking about it.

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+pkg_name="my_java_app"
+....
+pkg_blueprints=("core/blueprints/java8-maven-tomcat")
+               ^     ^^^^^^^^^^^                    ^
+               |______|_______|_____________________|
+                 Arrays and additional namespace TBD -- just random thoughts at the moment.
+```
+* add `auto` package to `build_deps`
+* set toolchain variable
+
```
mwrock: I would move these 2 lines above the snippet and include example usage in the snippet.

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+#### Optional Variables
+* `pkg_src` - now accepts URI notation. Each blueprint contains a default URI which is relevant to the type of blueprint being used.
+*  `pkg_blueprint_opts` - Options which override some default runtime parameters. Eg: `--env DEV`. This operates similar to `pkg_svc_run` as the result of using this option will intelligently generate a hook file for you. However, you also can create the [hook file](#Run-Hooks) directly if you need to have more control over the resulting output.
+
+### Available Blueprints
```
reset: This looks like a great list for our initial offerings even though I know you said it's a mockup ;). We should get some feedback from @tashimi on what she think, too.

@fnichol and I spoke about tackling `go`  as one of the first, I think that could be an easy home run

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+Each blueprint defines a set of callbacks which are unique to the blueprint type. This allows you to complex projects which may need to leverage different blueprints as needed within your project. Each of these callbacks define overrides of base level callbacks.
+
+### ruby _(This will likely live in the blueprint README.md)_
+#### do_bundler
+#### do_rake
+#### do_rack
+#### do_lotsmorestuff 
+
+## Run Hooks
+#### Health Checks
```
reset: I love that we could potentially fill out somebody's healt check/smoke test/init/run hooks for them. This is going to be a killer feature

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+               |______|_______|_____________________|
+                 Arrays and additional namespace TBD -- just random thoughts at the moment.
+```
+* add `auto` package to `build_deps`
+* set toolchain variable
+
+#### Optional Variables
+* `pkg_src` - now accepts URI notation. Each blueprint contains a default URI which is relevant to the type of blueprint being used.
+*  `pkg_blueprint_opts` - Options which override some default runtime parameters. Eg: `--env DEV`. This operates similar to `pkg_svc_run` as the result of using this option will intelligently generate a hook file for you. However, you also can create the [hook file](#Run-Hooks) directly if you need to have more control over the resulting output.
+
```
mwrock: One suggestion would be to have `hab plan init` include this variable with all possible options for the pack.

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+## Callbacks
+Each blueprint defines a set of callbacks which are unique to the blueprint type. This allows you to complex projects which may need to leverage different blueprints as needed within your project. Each of these callbacks define overrides of base level callbacks.
+
+### ruby _(This will likely live in the blueprint README.md)_
+#### do_bundler
+#### do_rake
+#### do_rack
+#### do_lotsmorestuff 
+
```
mwrock: Same as above comment: would be good to include commented function stubs for each of these.

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+```bash
+pkg_name="my_java_app"
+....
+pkg_blueprints=("core/blueprints/java8-maven-tomcat")
+               ^     ^^^^^^^^^^^                    ^
+               |______|_______|_____________________|
+                 Arrays and additional namespace TBD -- just random thoughts at the moment.
+```
+* add `auto` package to `build_deps`
```
rhass: Oops! Forgot to yank that from the origin rev. Thanks!

www/source/docs/blueprints.md
bdangit: :+1:

www/source/docs/blueprints.md
fnichol: Maybe "build" to "build phases" and "runtime phases" to "runtime hooks"?

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
+# Blueprints
+Blueprints are default implementations of the build and runtime phases for your application. They also provide some default health check hooks where appropriate to ensure your application functioning reliably. Customized Blueprints can be created to facilitate reusability of common patterns in your organization for developing, building, and running your applications.
+
+## Getting Started
+### New Projects
+Simply run `hab plan init` to leverage blueprints. Habitat will look in your application code for known project types and the appropriate `pkg_blueprint`.
+
+If you would like to specify the desired blueprint to use, you can run `hab plan init java8-maven-tomcat`
```
fnichol: Given our current implementation of `hab plan init`, `java8-maven-tomcat` would map to `PKG_NAME`. My guess is this would be an option (i.e. `--blueprint java8-maven-tomcat` or similar) but I don't think it's critical for this doc.

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Existing Plans
+
+To begin using Blueprints, you will need to add the appropriate `pkg_blueprints` 
+
+```bash
+pkg_name="my_java_app"
+....
+pkg_blueprints=("core/blueprints/java8-maven-tomcat")
```
fnichol: Ha, that's a nice point--we do this with our `core/hab-backline` [plan](https://github.com/habitat-sh/habitat/blob/master/components/backline/plan.sh) which only exists to bundle up dependencies into 1 logical unit.

In this case, would we tentatively use single-case version of the variable name such as `pkg_blueprint`?

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+               ^     ^^^^^^^^^^^                    ^
+               |______|_______|_____________________|
+                 Arrays and additional namespace TBD -- just random thoughts at the moment.
+```
+* add `auto` package to `build_deps`
+* set toolchain variable
```
fnichol: In that case, it might make sense to either make a slightly longer prefix (like `pkg_blueprint_`) or go for an alternate `blueprint_` prefix. That would keep the build program free to still control the `pkg_` namespace for any future work we might need and provide Blueprint authors a safer namespace that won't be collided with in the future. It should also be clearer to a Plan author which are core metadata (i.e. "where do I look for the `pkg_name` docs?") and any Blueprint metadata (i.e. "where do I look for the `blueprint_node_build_cmd` docs?")

www/source/docs/blueprints.md
fnichol: Agreed with above!

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+#### Health Checks
+Each blueprint comes with a default health check for your application.
+#### blueprint_opts
+Follow these directions if you wish to create the hook which would otherwise be created by the optional  `pkg_blueprint_opts` variable.
+
+*TODO: Write the directions*
+
+## Build Artifacts
+### Auto-Ingestion
+Each blueprint defines a default value for `pkg_src` in [URI syntax](https://tools.ietf.org/html/rfc3986). You can override this value within your plan should you application source live in a non-standard location.
```
fnichol: I'd like to see `pkg_source` ultimately become an optional variable, only to be used when your Plan isn't included with your codebase. At the moment we have no convenience variable for `$HAB_CACHE_SRC_PATH/$pkg_dirname` (which would be something like `/hab/cache/src/zlib-1.2.3` for most third-party core plans) but having such a variable (whether it's internal to the program or a public, user facing tunable) would help. Currently we have [`$PLAN_CONTEXT`](https://github.com/habitat-sh/habitat/blob/849fa27188b3c7cc3361d4edfd37c9a71e1d3bef/components/plan-build/bin/hab-plan-build.sh#L2414-L2440) which is used when we need to refer back to the `plan.sh` itself or the `./hooks` directory, etc.

We could change the build program slightly to the following:
1. if `pkg_source` is set, do what we do today and set an internal var `_src_root="$HAB_CACHE_SRC_PATH/$pkg_dirname"`
2. if `pkg_source` is not set, then set `_src_root="$(dirname "$PLAN_CONTEXT")"` (if we have the `./habitat/plan.sh` situation and `_src_root="$PLAN_CONTEXT"` if plan.sh is in the root of the codebase)

then each build phase wrapper uses `pushd "$_src_root"` (updating lines like [this](https://github.com/habitat-sh/habitat/blob/849fa27188b3c7cc3361d4edfd37c9a71e1d3bef/components/plan-build/bin/hab-plan-build.sh#L1858)).

Anyway, that's a sidetrack, but I'd really like to *not* use `$pkg_source` for plan.sh files that live with a codebase

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,58 @@
...
+* `core/blueprint/java8-maven-jboss`
+* `core/blueprint/java8-gradle-tomcat`
+* `core/blueprint/java7-ant`
+* `core/blueprint/python-unicorn`
+
+## Callbacks
+Each blueprint defines a set of callbacks which are unique to the blueprint type. This allows you to complex projects which may need to leverage different blueprints as needed within your project. Each of these callbacks define overrides of base level callbacks.
+
+### ruby _(This will likely live in the blueprint README.md)_
+#### do_bundler
```
fnichol: With similar reasoning above, we might want to consider namespacing these Blueprint-specific sub-phases to make it clear that they come from a Blueprint and will be case-specific (i.e. not a general case for the build program).

fnichol:
This is so darn close to what I was hoping for, I am excited!

www/source/docs/blueprints.md
rhass: @fnichol Hrmm... I like the idea of identifying the source of the function, I also want to keep the name as simple as possible to avoid having to fully understand what a blueprint is in full when getting started. This is to say, if we use the `blueprint` term for everything it make it seem like more glue rather than core functionality.

As I think about this, there could be some different approaches we may want to try or consider. One super rough thought that comes to mind would be a top level wrapper of sorts:
```bash
blueprint=ruby23-mri-unicorn

do_blueprint() {
  do_bundler() {
     bundle exec foo
  }
}
``` 

I am not entirely sure how this would work under the hood since bash is not object oriented ... but `do_blueprint`. But if we can figure this out it would make it a nice clean way to declare all the bits which are blueprint related.

adamhjk: You need to namespace them. It's a good case for not using `blueprint`, because thats a lot to type.

`do_auto_bundler()` makes a lot more sense.

www/source/docs/blueprints.md
adamhjk: Absolutely. We could maybe make some magic happen here if you set `pkg_source` to `local` or something.

rhass:
@reset @fnichol Take another look when you can. Thanks!

mwrock:
The description in the markdown looks great! Did you mean to delete the files you did? I'm guessing that was unintentional.

rhass:
@mwrock Whoa! I didn't even notice that somehow! :man_facepalming: Thanks!

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,66 @@
...
+The resulting plan will contain all the available callbacks and variables relevant to the blueprint type.
+
+### Existing Plans
+
+To begin using Blueprints, you will need to add the appropriate `pkg_blueprints` 
+
+```bash
+pkg_name="my_java_app"
+....
+blueprint="core/blueprint-java8-maven-tomcat"
```
reset: `pkg_blueprint` here - it'll become obvious why we've namespaced all of the plan variables with `pkg_` as soon as blueprints come together, actually 😄. We'll probably suggest that people use a prefix by convention for the feature (`bp_` or something similar) to make sure we don't have conflicts in our namespace

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,66 @@
...
+### New Projects
+Simply run `hab plan init` to leverage blueprints. Habitat will look in your application code for known project types and the appropriate `pkg_blueprint`.
+
+If you would like to specify the desired blueprint to use, you can run `hab plan init --blueprint core/java8-maven-tomcat`
+
+The resulting plan will contain all the available callbacks and variables relevant to the blueprint type.
+
+### Existing Plans
+
+To begin using Blueprints, you will need to add the appropriate `pkg_blueprints`
```
reset: `pkg_blueprint` - I don't think there will ever be plurality here. The blueprint itself would be composed of others so you wouldn't need to pick more than one

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,66 @@
+# Blueprints
+Blueprints are default implementations of the build phases and runtime hooks for your application. They also provide some default health check hooks where appropriate to ensure your application is functioning reliably. Customized Blueprints can be created to facilitate re-usability of common patterns in your organization for developing, building, and running your applications.
```
reset: health checks are just *one* hook that we might provide implementations for. I think it's more appropriate to say "They also provide some default implementations of various life-cycle hooks where appropriate"

www/source/docs/blueprints.md
```diff
@@ -0,0 +1,66 @@
...
+#### do_blueprint_bundler
+#### do_blueprint_rake
+#### do_blueprint_rack
+#### do_blueprint_more_ruby_stuff 
+
+## Variables
+These are variables which each blueprint honors, allowing the plan author to consume the value to override a particular behavior. 
+### ruby
+:warning: _This section will likely live in the Ruby blueprint README.md and is here for feature clarification_ :warning:
+* pkg_ruby_build_cmd - Overrides default binary and path.
```
reset: We may consider using another prefix to notate overridable blueprint variables aside from `pkg_`. Let's go with this for now but keep that in mind

reset:
@rhass this is coming together really nicely, I think this is a great start and we should move on with initial implementation based on this spec 🎉

@fnichol you wanna take one last pass?

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/213114678).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/213114678).

PR #1964, opened 2017-03-17
Persist scheduler store to Postgres

chefsalim:
This change adds Postgres-based persistence to the builder-scheduler data store. 

giphy 13


thesentinels:

Thank you for contributing!

components/builder-scheduler/src/config.rs
```diff
@@ -71,7 +71,7 @@ impl Default for Config {
         Config {
             routers: vec![SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 5562))],
             job_servers: vec![jobsrv],
-            datastore_connection_url: String::from("postgresql:://hab@127.0.0.1/builder_db_test"),
```
fnichol: ha good catch--I'm going to guess the double colon was some Rustacean muscle memory

components/builder-scheduler/src/data_store.rs
```diff
@@ -12,215 +12,335 @@
...
+        let pool = Pool::new(&config.datastore_connection_url,
+                             config.pool_size,
+                             config.datastore_connection_retry_ms,
+                             config.datastore_connection_timeout,
+                             config.datastore_connection_test)?;
+        Ok(DataStore { pool: pool })
+    }
+
+    /// Create a new DataStore from a pre-existing pool; useful for testing the database.
+    pub fn from_pool(pool: Pool) -> Result<DataStore> {
```
fnichol: I love these constructors

components/builder-scheduler/src/data_store.rs
```diff
@@ -12,215 +12,335 @@
...
 
-        assert!(!self.group_map.contains_key(&id));
-        self.group_map.insert(id, self.groups.len() - 1);
+        let mut group = Group::new();
+        group.set_id(id);
+        group.set_state(GroupState::Pending);
+        group.set_projects(projects);
 
-        println!("Group created: {:?}", group_data);
+        debug!("Group created: {:?}", group);
```
fnichol: 👍

components/builder-scheduler/src/data_store.rs
```diff
@@ -12,215 +12,335 @@
...
+        let project_rows = &conn.query("SELECT * FROM get_projects_for_group_v1($1)",
+                   &[&(group_id as i64)])
+            .map_err(Error::GroupGet)?;
 
-        let mut jobs = RepeatedField::new();
-        for group_job in group_data.jobs.iter() {
-            jobs.push(group_job.clone());
-        }
-        group.set_jobs(jobs);
+        assert!(project_rows.len() > 0); // should at least have one
```
fnichol: I appreciate these and the associated comment

components/builder-scheduler/src/config.rs
adamhjk: It was copy-pasta error!

components/builder-scheduler/src/data_store.rs
```diff
@@ -12,215 +12,335 @@
...
 
-        Ok(DataStore::group_data_to_group(&group_data))
+        Ok(group)
     }
 
-    pub fn get_group(&self, group_id: u64) -> Option<Group> {
-        if self.group_map.contains_key(&group_id) {
-            let index = *self.group_map.get(&group_id).unwrap();
-            assert!(self.groups[index].id == group_id);
+    pub fn get_group(&self, group_id: u64) -> Result<Option<Group>> {
```
adamhjk: I would change this to take the type of the request protobuf. That brings it more inline with the way the final refactoring of the `originsrv` works, and has made the ergonomics much nicer.

components/builder-scheduler/src/data_store.rs
```diff
@@ -12,215 +12,335 @@
...
-    pub fn create_group(&mut self, project_names: Vec<String>) -> Result<Group> {
-        let mut projects = Vec::new();
-        for name in project_names {
-            let project = ProjectData {
-                name: name,
-                state: proto::ProjectState::NotStarted,
-            };
-            projects.push(project);
-        }
+    pub fn create_group(&self, project_names: Vec<String>) -> Result<Group> {
```
adamhjk: Rather than take a `Vec<String>` here, I would change this to take the incoming protocol protobuf. That's how the `originsrv` works now, and it's been much nicer.

adamhjk: (You're already returning the protobuf type, which is what you want.)

components/builder-scheduler/src/server/handlers.rs
```diff
@@ -50,11 +45,7 @@ pub fn schedule_get(req: &mut Envelope,
     let msg: proto::ScheduleGet = try!(req.parse_msg());
     println!("ScheduleGet: {:?}", msg);
 
-    let group_opt;
-    {
-        let ds = state.datastore().read().unwrap();
-        group_opt = ds.get_group(msg.get_group_id());
-    }
+    let group_opt = state.datastore().get_group(msg.get_group_id()).unwrap();
```
adamhjk: This unwrap needs a comment about why its safe, and at the very least should be an expect.

components/builder-scheduler/src/server/scheduler.rs
```diff
@@ -302,10 +275,7 @@ impl ScheduleMgr {
                 proto::GroupState::Pending
             };
 
-            {
-                let mut ds = self.datastore.write().unwrap();
-                ds.set_group_state(group_id, new_state).unwrap();
-            }
+            self.datastore.set_group_state(group_id, new_state).unwrap();
```
adamhjk: Needs a comment, and should be `expect` so we know what it is if it fails.

support/db/start.sh
```diff
@@ -31,8 +31,9 @@ echo "Creating builder_jobsrv database if neccessary"
 sudo -u hab -E TERM=vt100 hab pkg exec core/postgresql createdb -O hab -h 127.0.0.1 builder_jobsrv
 echo "Creating builder_originsrv database if neccessary"
 sudo -u hab -E TERM=vt100 hab pkg exec core/postgresql createdb -O hab -h 127.0.0.1 builder_originsrv
+echo "Creating builder_scheduler database if neccessary"
+sudo -u hab -E TERM=vt100 hab pkg exec core/postgresql createdb -O hab -h 127.0.0.1 builder_scheduler
```
adamhjk: Nice catch!

components/builder-scheduler/src/data_store.rs
chefsalim: The suggestion makes sense, however the reason this is a vec string is that we don't actually pass in the list of packages in the protobuf, so we don't have a protobuf to pass in that contains this info.  In the current design paradigm, the dependent packages will be constructed by the message handler, and passed to the datastore.

adamhjk: Where do the project names come from, though? Aren't they extracted from the request in the handler?

adamhjk: In particular, I'm talking about lines 34 and 48 in `handlers.rs`, where we extract data from the inbound protocol and pass that to the data store. Instead, we should pass a reference to the protocol message and do the extraction inside the datastore.

components/builder-scheduler/src/data_store.rs
chefsalim: This function gets called from both a ScheduleGet request and also by a Job notification. So we could have it take a ScheduleGet as the incoming param, but that is one extra (though minor) step for the Job notification path.

components/builder-scheduler/src/server/handlers.rs
chefsalim: Good catch - it's actually not safe, will add a check.

components/builder-scheduler/src/server/scheduler.rs
chefsalim: Will add an error check.

components/builder-scheduler/src/data_store.rs
adamhjk: I've been adding a wrapper in those cases, where you can call it by the non-protocol bit with something like `get_group_by_id`, and `get_group` just takes the protocol, extracts the id, and then calls `get_group_by_id` under the hood.

adamhjk:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/213128667).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/213128667).

PR #1969, opened 2017-03-19
Serialize/Deserialize `ServiceSpec` & implement `SpecWatcher` to load/unload service definitions from disk

fnichol:
This changeset was made in a way that is either additive to the codebase (and therefore not yet used) or as a refactoring of existing code (in the case of the serializing of several pre-existing types). The table should be about set now to wire the `SpecWatcher` into the `Manager` to pick up service spec files in a similar style to runit and daemontools do with a directory in `/sv/service`.

This is happening!

---

## [sup] Implement `Deserialize` for `ServiceSpec` via serde/toml.
This change allows a `ServiceSpec` to be created from a TOML string
representation with custom type error checking, validation of
required fields, etc. A future feature will use this implementation to
read a `.toml` file from disk.

The following related changes were made or added:

* Introduce `ServiceBind` which knows how to create itself from a str
(via `FromStr`)
* Implement `Deserialize` for `Topology`
* Implement `Deserialize` for `UpdateStrategy`
* Create a generic `Deserialize` implementation in `core` for types
implementing `FromStr` (used by `UpdateStrategy`, `Topology`,
`ServiceBind`, and indirectly by `PackageIdent`)

## [sup] Implement `Serialize` for `ServiceSpec` via serde/toml.  …
This change produces a TOML string representation of a `ServiceSpec` via
a `to_toml_string()` function which uses the toml crate's serde
serializer implementation. A future feature will use this implementation
to write a `.toml` file to disk.

The following related changes were made or added:

* Implement `Serialize` and `Display` for `Topology`
* Implement `Serialize` and `Display` for `UpdateStrategy`
* Implement `Serialize` and `Display` for `ServiceBind`
* Create a generic `Serialize` implementation in `core` for types
implementing `ToString` (via implementing `Display`) which is used
indirectly by `PackageIdent`

## [sup] Implement `SpecWatcher` to pick up spec file changes on disk.

This change adds the `SpecWatcher` but doesn't wire it in anywhere yet
(it is been well tested in isolation in the meantime). The `SpecWatcher`
uses file system events (appropriate to the respective platforms) to
trigger a scan/load/reconcile activity which is the mechanism by which
the Supervisor's `Manager` will be able to add, remove, and (soon)
update running services.

Setting up the `SpecWatcher` is fairly trivial--invoke the `run()`
function giving it a directory path to watch (note this directory must
exist and is checked for):

```
let tmpdir = TempDir::new("specs").unwrap();
let mut watcher = SpecWatcher::run(tmpdir.path()).unwrap();
```

In order to furnish the `Manager` with an initial list of services to
load, the `SpecWatcher` provides a `.initial_events()` method. This
returns a `Vec<SpecWatcherEvent>` where each event is an
`AddService(ServiceSpec)` entry.

```
let initial_events = watcher.initial_events().unwrap();
```

Finally, on each `Manager` tick a `.new_events()` method is provided
which returns the same `Vec<SpecWatcherEvent>` but this time each event
could be an `AddService(ServiceSpec)` or a `RemoveService(ServiceSpec)`
entry. Think of this as tracking the state of a directory with version
control, each patch being a diff to the current desired state.

A few notes:

* While `SpecWatcher` deals with a changed `ServiceSpec` (i.e. where the
on-disk version for a service is different from what is currently
running), the implementation is the simple version: emit a
`RemoveService(..)` event followed by an `AddService(..)` event with the
new `ServiceSpec` desired state. This will, in effect, like restarting a
service. Not the end of the world but until (and if) we can support
updating a service in-place, this keeps us is a good(ish) place.
* The spec file names must conform to `<NAME>.spec.toml` where `<NAME>`
is the name component of a package identifier. In other words, a spec
file whos' contents are `ident = "core/redis"` must be named
`redis.spec.toml`. Failure to do so results in a warning message (with
hints as to why and how to fix it) and the file is explicitly skipped.
Remember that all spec files will be re-scanned many times in the future
so each scan is a chance that the user has fixed their problem and the
system will recover. In this way we try to fail-safe, or "do no further
harm".
* The same strategy is taken for malformed spec files or files that
cannot be properly read due to IO/permission issues/etc.: a warning
message is display and the file is explicitly skipped.
* Currently a file system event needs to fire for the scanning to occur.
There should probably be a timer that ensures the directory is being
scanned some minimum number of times per period (like at least once
every 30 seconds or 2 minutes). Most likely room for improvement
here--using this module live in the Supervisor will give us the feedback
we need.


thesentinels:

Thank you for contributing!

fnichol:
gif-keyboard-6478525090938551104

gif-keyboard-6935520905077062508

components/core/src/util/mod.rs
```diff
@@ -14,3 +14,52 @@
...
+
+use std::error;
+use std::fmt;
+use std::marker::PhantomData;
+use std::result;
+use std::str::FromStr;
+
+use serde;
+
+pub fn deserialize_using_from_str<T, E, D>(d: D) -> result::Result<T, D::Error>
```
reset: This is a nice generic serde helper!

components/sup/src/error.rs
```diff
@@ -161,7 +171,9 @@ impl fmt::Display for SupError {
             Error::EnvJoinPathsError(ref err) => format!("{}", err),
             Error::FileNotFound(ref e) => format!("File not found at: {}", e),
             Error::InvalidBinding(ref binding) => {
-                format!("Invalid binding - must be ':' delimited: {}", binding)
+                format!("Invalid binding \"{}\", must be of the form <NAME>:<SERVICE_GROUP> where \
```
reset: Much better error message

components/sup/src/manager/spec_watcher.rs
```diff
@@ -0,0 +1,600 @@
...
+                            continue;
+                        }
+                        // All other errors are unexpected and should be dealt with up the calling
+                        // stack.
+                        _ => return Err(e),
+                    }
+                }
+            };
+            let file_stem = spec_file.file_name()
+                .expect("name should not be ..")
```
reset: It looks like these `expect()`s could be hit by the end user? If that's the case, we should swap these to warning messages. We should only use `expect`/`unwrap` as a synonym for a `debug_assertion` to communicate to the developer that they did something wrong

reset:
@fnichol this is wonderful! I've only got one message in line that you should take a peek at regarding some  calls to `expect()`. Super excited to see this wired into the supervisor soon.

We should document the format of the service spec, too, but we can hit that in a future PR

components/sup/src/manager/spec_watcher.rs
fnichol: Yeah, I like that idea. When I wrote this part my thought was: we've just proven that the file exists, path is good etc.--this really should never fail. However, when you use the word "never", it happens. I'll fix this up with the same error recovery logic (warn the user that we can't proceed with this entry and move to the next).

I agree with our use of `.unwrap()` and `expect()` which should be justifiable--perhaps even with a comment nearby to explain why it is okay in this case. For example, I have a couple of `.expect()` calls in `SpecWatcher` around fetching values from `HashMap`s. I know they are present as I'm using a key value which I previously extracted from the same `HashMap`. If those `Option<T>`s return `None` something has corrupted memory (that's one reason for tests in this module--a logic error might trigger an `.expect()` here to panic, but not in production).

components/sup/src/error.rs
fnichol: It's been helpful writing these error messages as their contents will be written out for humans to read and action against--I was one of those early humans in development :smile:

reset:
@fnichol merge when ready ;)

components/sup/src/manager/service/hooks.rs
```diff
@@ -722,6 +722,8 @@ impl HookOutput {
         }
     }
 
+    // TODO fn: this method appears to not be used in the codebase, is it necessary?
```
mwrock: I'd be inclined to just rip this out if no one can think where this might be used.

mwrock:
this looks great!

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/213110356).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/213110356).

PR #1976, opened 2017-03-21
[sup] Activate the MVP multi-service Supervisor! (behind feature flag)

fnichol:
Hey there, how are ya? Are you interested in running more than one
service with a single Habitat Supervisor? Great! Let's talk.

First off, this is the first integration of several moving parts, and as
such is known to be not fully capable yet. But that's cool, because you
won't get the new behavior unless you start the Supervisor with an
environment variable to enable the `Multi` feature. Setting:

    HAB_FEAT_MULTI=true

will start the Supervisor with an extra component (internally called the
`SpecWatcher`) which will monitor a directory of service specification
TOML files, called Service Specs (internally referred to as a
`ServiceSpec` or a "spec" for short) following a strategy similar to
runit or daemontools.

The exiting behavior of `hab-sup start` is preserved except in one
circumstance described in the `Breaking Behavior` section below.

Quick TL;DR
-----------

gif-keyboard-9576611079189079384

All of the file formats, directories, and start modes shown here are
explained below, so if this feels a touch magical...it's because you
haven't learned about it yet!

To run a single Supervisor that is running `core/redis`, `core/nginx`,
and `core/postgresql` (admittedly a bit contrived) you'll create 3
Service Specs TOML files in the default spec watch directory and then
start the Supervisor in its default mode. Assuming you've compiled the
Supervisor yourself locally:

```
sudo mkdir -p /hab/sup/default/specs
echo 'ident = "core/redis"' | sudo tee /hab/sup/default/specs/redis.spec.toml
echo 'ident = "core/nginx"' | sudo tee /hab/sup/default/specs/nginx.spec.toml
echo 'ident = "core/postgresql"' | sudo tee /hab/sup/default/specs/postgresql.spec.toml

sudo -E env HAB_FEAT_MULTI=true ./target/debug/hab-sup start
```

To kill the foregrounded Supervisor (as always), you can type `Ctrl+c`.
Then to start the Supervisor again with all previous services, the
same Supervisor identifier, etc.:

```
sudo -E env HAB_FEAT_MULTI=true ./target/debug/hab-sup start
```

For more details on the above example, read on!

Breaking Behavior
-----------------

First a note about potentially breaking behvaior. This change introduces
one breaking change wthat ill impact existing Habitat users who run more
than one Supervisor per host/VM/container/etc. More information is
provided in the `Supervisor State Directory` section below.

This only applies when running more than one Supervisor--you know this
because you will have had to explicitly set non-default values for the
`--listen-gossip` and `--listen-http` options. There are 2 strategies to
choose from:

1. Eliminate the multiple Supervisors launching via SystemD, init, etc.,
write a corresponding Service Spec TOML file per service (described
below in more detail), and launch a single Supervisor with `hab start`
(note the lack of package identifiers, topology strategies, etc.)
2. Alternatively, for each exisiting Supervisor, add an
`--override-name` option to the command with some unique value. This
will allow the Supervisors to write their own state information to
distinctly different directories.  Once stable, you are highly
encouraged to migrate to the single Supervisor strategy as explained
directly above.

If you have any questions or concerns, feel free to contact the Habitat
team in Slack (http://slack.habitat.sh/), on our forums
(https://forums.habitat.sh/), or by email (humans@habitat.sh).

Service Spec TOML Files
-----------------------

gif-keyboard-18288171251351446621

While a near future workflow with the Supervisor shouldn't require users
to immediately understand Service Spec files (hereafter referred to as
specs), it might be instructional to those who wish to seed the
Supervisor with multiple service definitions before it boots to
understand the file format.

The file a flat key/value structure with no nesting and where each key
corresponds to a service-related option or flag currently existing on
the `hab-sup start` subcommand. Here is the current help output for
`hab-sup start`:

```
hab-sup-start
Start a Habitat-supervised service from a package or artifact

USAGE:
    hab-sup start [FLAGS] [OPTIONS] <PKG_IDENT_OR_ARTIFACT>

FLAGS:
        --no-color          Turn ANSI color off
    -I, --permanent-peer    If this service is a permanent peer
    -v                      Verbose output; shows line numbers
    -h, --help              Prints help information

OPTIONS:
        --bind <BIND>...                   One or more service groups to bind to a configuration
        --config-from <CONFIG_DIR>         Use package config from this path, rather than the package itself
    -u, --url <DEPOT_URL>                  Use a specific Depot URL (ex: http://depot.example.com/v1/depot)
        --group <GROUP>                    The service group; shared config and topology [default: default].
        --listen-gossip <LISTEN_GOSSIP>    The listen address for the gossip system [default: 0.0.0.0:9638]
        --listen-http <LISTEN_HTTP>        The listen address for the HTTP gateway [default: 0.0.0.0:9631]
        --org <ORGANIZATION>               The organization that a service is part of
        --peer <PEER>...                   The listen address of an initial peer (IP[:PORT])
    -r, --ring <RING>                      Ring key name
    -s, --strategy <STRATEGY>              The update strategy; [default: none] [values: none, at-once, rolling]
    -t, --topology <TOPOLOGY>              Service topology; [default: none]

ARGS:
    <PKG_IDENT_OR_ARTIFACT>    A Habitat package identifier (ex: acme/redis) or filepath to a Habitat Artifact (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)
```

In particular, here is an invocation of the Supervisor with all service-related options spelled out explicitly:

```
hab-sup start \
  origin/name/1.2.3/20170223130020 \  # required
  --group jobs \                      # default is: "default"
  --org acmecorp \                    # default is: <unset> (optional value)
  --url http://example.com/depot \    # default is: "https://willem.habitat.sh/v1/depot"
  --topology leader \                 # default is: "none"
  --strategy at-once \                # default is: "none"
  --bind cache:redis.cache@acmecorp \ # default is <unset> (option values)
  --bind db:postgres.app@acmecorp
```

A corresponding spec file, named `name.spec.toml` would be:

```
ident = "origin/name/1.2.3/20170223130020"
group = "jobs"
organization = "acmecorp"
depot_url = "http://example.com/depot"
topology = "leader"
update_strategy = "rolling"
binds = ["cache:redis.cache@acmecorp", "db:postgres.app@acmecorp"]
```

There is only *1* required key/value pair in a spec file:

```
ident = "origin/name"
```

As with the `hab-sup start` command, all other key/value pairs have the
same defaults and are optional.

Supervisor Start Modes
----------------------

There are 2 slightly different behaviors when starting the Supervisor
via `hab-sup start` (or `hab start`). Previously, a Package Identifier
or local Package Artifact was a required argument, however now this is
optional. See the following sections for details:

 ### Specifying No Identifiers Or Artifacts

gif-keyboard-14610798927384970348

In this mode on a brand new system, the Supervisor starts with no
services to run. Currently the only way to add services to the
Supervisor is by adding a spec file in the specs watch directory
(described below). Future features will add user tooling to automate the
creation and removal of these files, but interacting directly with these
spec files is still useful when working with configuration management
tools such as Chef or in container preparation workflows like creating a
Docker container.

For init systems such as SystemD, runit, etc. the new recommended
command to start the Supervisor would be:

    hab start

that is, without specifying any Package Identifiers. In this invocation,
the Supervisor will check its specs watch directory for specs to load
and start all related services. Restarting the entire Supervisor (and
therefore also all supervised services) would mean killing the `hab-sup
start` process and re-running `hab start`. All previous Supervisor state
and desired service specifications will be used when it boots next.

 ### Specifying A Package Identifier

gif-keyboard-11126045281174846351

This is intended for one-off Supervisor launches such as in testing,
local development or simple deployments such as in containers, etc. A
spec file for the corresponding Package Identifier (and related options
and flags) will be created in the specs watch directory (described
below) and the Supervisor will start. Think of this as a convenience
"short cut" of 2 activities:

1. Add a service to be supervised
2. Start the Supervisor

If you intend to start more than one service, you are highly suggested
to start the supervisor with *no* Package Identifiers, as described
above. More user tooling will be added in the near future to make the
activity of adding (and removing) additional services easier.

 ### Specifying A Local Package Artifact

gif-keyboard-13041148434355848888

This is intended primarily for local development and testing and is not
recommended for use in production--you are highly encouraged to use the
*no* Package Identifiers scenario described above, or secondarily the
Package Identifier scenario described above.

This follows the same behavior as above, with one major distinction: the
Supervisor will install the local Package Artifact before continuing and
lock the Package Identifier to the fully qualified value extracted from
the artifact. For example a Package Artifact
`"origin-name-1.2.3-20170223130020-x86_64-linux.hart"` would produce a
spec file containing the line:

    ident = "origin/name/1.2.3/20170223130020"

That would mean, in effect, there is no update strategy (there is only
one possible package that satisfies a fully qualified Package
Identifier) and any `update_strategy` value would be null-and-void.

Supervisor State Directory
--------------------------

gif-keyboard-12629087991134924323

When the Supervisor starts, it determines a directory for itself to
persist some Supervisor-specific data and state. By default, this
directory is:

     /hab/sup/default

The Habitat team would like to note that once the multi-service support
in the Supervisor is stable, we highly recommend running *one*
Supervisor per host/VM/container/etc. The setup is much simpler, the
network traffic is far less, and most improvement and innovation will
assume this mode of operation. There may be users and use cases for
running more than one Supervisor and for these scenarios you can supply
an `--override-name` option to the Supervisor when starting. Currently,
the only use for this value is in determining the state directory of the
Supervisor. For example, starting the Supervisor with
`hab-sup start --override-name secondary` would use a Supervisor state
directory of `/hab/sup/secondary`.

Again, it is our assumption that most Habitat users will use one
Supervisor per host/VM/container/etc. Our hypothesis follows that this
shouldn't impact many users and the vast majority of users (who even
need to know about this path), would use `/hab/sup/default/*` without
thinking about it.

Service Specs Watch Directory
-----------------------------

gif-keyboard-6893989295495612039

Given the determination of the Supervisor state directory above, the
Supervisor will create a child directory called `specs/` (by default
`/hab/sup/default/specs`) which will be monitored for spec file changes
during the running of the Supervisor.

During the booting process, the Supervisor will use any spec files found
in the specs directory to populate itself and start the corresponding
services.

---

There are several know issues and while things aren't perfect, this is more than an initial start. Celebrate!

gif-keyboard-5617735478235085312

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -106,11 +106,14 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
                 "The listen address for the gossip system [default: 0.0.0.0:9638]")
             (@arg LISTEN_HTTP: --("listen-http") +takes_value
                 "The listen address for the HTTP gateway [default: 0.0.0.0:9631]")
+            (@arg NAME: --("override-name") +takes_value
+                "The name for the state directory if launching more than one Supervisor \
+                [default: default]")
             (@arg PEER: --peer +takes_value +multiple
                 "The listen address of an initial peer (IP[:PORT])")
             (@arg PERMANENT_PEER: --("permanent-peer") -I "If this service is a permanent peer")
             (@arg RING: --ring -r +takes_value "Ring key name")
-            (@arg PKG_IDENT_OR_ARTIFACT: +required
+            (@arg PKG_IDENT_OR_ARTIFACT:
                 "A Habitat package identifier (ex: acme/redis) or filepath to a Habitat Artifact \
                 (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
             (@group SVC_ARGS =>
@@ -179,6 +182,20 @@ fn sub_start(m: &ArgMatches) -> Result<()> {
     if let Some(addr_str) = m.value_of("LISTEN_HTTP") {
         cfg.http_listen = try!(http_gateway::ListenAddr::from_str(addr_str));
     }
+    if let Some(name_str) = m.value_of("NAME") {
+        cfg.name = Some(String::from(name_str));
+        outputln!("");
+        outputln!("{} Running more than one Habitat Supervisor is not recommended for most",
+                  Red.bold().paint("CAUTION:".to_string()));
+        outputln!("{} users in most use cases. Using one Supervisor per host for multiple",
+                  Red.bold().paint("CAUTION:".to_string()));
+        outputln!("{} services in one ring will yield much better performance.",
+                  Red.bold().paint("CAUTION:".to_string()));
+        outputln!("");
+        outputln!("{} If you know what you're doing, carry on!",
+                  Red.bold().paint("CAUTION:".to_string()));
+        outputln!("");
+    }
     if m.is_present("PERMANENT_PEER") {
         cfg.gossip_permanent = true;
     }
@@ -230,16 +247,31 @@ fn sub_start(m: &ArgMatches) -> Result<()> {
     }
 
     let mut maybe_local_artifact: Option<&str> = None;
-    let ident_or_artifact = m.value_of("PKG_IDENT_OR_ARTIFACT").unwrap();
-    let ident = if Path::new(ident_or_artifact).is_file() {
-        maybe_local_artifact = Some(ident_or_artifact);
-        try!(PackageArchive::new(Path::new(ident_or_artifact)).ident())
-    } else {
-        try!(PackageIdent::from_str(ident_or_artifact))
+    let maybe_spec = match m.value_of("PKG_IDENT_OR_ARTIFACT") {
+        Some(ident_or_artifact) => {
+            let ident = if Path::new(ident_or_artifact).is_file() {
+                maybe_local_artifact = Some(ident_or_artifact);
+                try!(PackageArchive::new(Path::new(ident_or_artifact)).ident())
+            } else {
+                try!(PackageIdent::from_str(ident_or_artifact))
+            };
+            Some(try!(spec_from_matches(&ident, m)))
+        }
+        None => None,
     };
-    let spec = try!(spec_from_matches(&ident, m));
-
-    try!(command::start::package(cfg, spec, maybe_local_artifact));
+    if !feat::is_enabled(feat::Multi) && !m.is_present("PKG_IDENT_OR_ARTIFACT") {
+        // For now, mimick the "required argument" error message to preserve original behavior of a
+        // require pkg ident or artifact
+        println!("{} The following required arguments were not provided:",
+                 clap::Format::Error("error:"));
+        println!("    {}", clap::Format::Error("<PKG_IDENT_OR_ARTIFACT>"));
+        println!("");
+        println!("{}", m.usage());
+        println!("");
+        println!("For more information try {}", clap::Format::Good("--help"));
+        std::process::exit(1);
+    }
+    try!(command::start::package(cfg, maybe_spec, maybe_local_artifact));
     Ok(())
 }
 
@@ -310,7 +342,7 @@ fn valid_url(val: String) -> result::Result<(), String> {
 }
 
 fn enable_features_from_env() {
-    let features = vec![(feat::List, "LIST")];
+    let features = vec![(feat::List, "LIST"), (feat::Multi, "MULTI")];
 
     for feature in &features {
         match henv::var(format!("HAB_FEAT_{}", feature.1)) {
--- a/components/sup/src/command/start.rs
+++ b/components/sup/src/command/start.rs
@@ -47,6 +47,7 @@
 //! See the [documentation on topologies](../topology) for a deeper discussion of how they function.
 //!
 
+use std;
 use std::path::Path;
 
 use ansi_term::Colour::Yellow;
@@ -56,12 +57,16 @@ use hcore::fs::{self, FS_ROOT_PATH};
 
 use {PRODUCT, VERSION};
 use error::{Error, Result};
+use feat;
 use manager::{Manager, ManagerConfig};
 use manager::ServiceSpec;
 
 static LOGKEY: &'static str = "CS";
 
-pub fn package(cfg: ManagerConfig, spec: ServiceSpec, local_artifact: Option<&str>) -> Result<()> {
+pub fn package(cfg: ManagerConfig,
+               service_spec: Option<ServiceSpec>,
+               local_artifact: Option<&str>)
+               -> Result<()> {
     let mut ui = UI::default();
     if !fs::am_i_root() {
         ui.warn("Running the Habitat Supervisor requires root or administrator privileges. \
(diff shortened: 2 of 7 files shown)


thesentinels:

Thank you for contributing!

adamhjk:
One high level comment - I suspect having a dynamic location for the spec files will lead to pretty nasty side effects, particularly in trying to automate the insertion of spec files. One common case we've already encountered is k8s, where we need to specify a bind address to ensure we get the correct routing to peers. 

I suggest we make this a command line option rather than dynamically discovered - so you can specify `--cfg-path poop` and get `/hab/svc/poop`. In general, once this is accepted and turned on by default, we won't want anyone running more than one supervisor per logical system anyway.

fnichol:
@adamhjk Yeah, @reset, @tashimi and I just finished a conversation around the location of that root directory. If we were to pull the notion of an organization into the Supervisor we could use that to specify the directory name (i.e. the user has to type something explicit). Honestly, I don't know why it didn't occur to me to try a straight up option like `--cfg-path` (or similar)--that additional opt-in behavior might be all right and pretty straight forward.

fnichol:
The more I think about this, the more I'm liking the explicit option idea on `hab-sup start`. Favoring explicitness over convention magic usually leads to a better time and this is an escape hatch for people that have their own uses anyway. @reset How does an additional option strike you?

reset:
@fnichol I think an explicit option passed to `hab-sup start` is a totally alright way to go. I think we should add the flag `--id {stringy-var}` which will determine the data directory to use for that Supervisor (ie. `/hab/sup/badchoices/data`) and if you don't pass a value for ID we just write to `/hab/sup/default/data`. We should warn on startup that this is a power user feature to run more than one supervisor on a machine and things might get bumpy!

smith:
`--peer` is not in the example `hab sup start` or TOML. Do all services running on a supervisor automatically select a 127.0.0.1 peer or is this something that has to be specified in the spec?

fnichol:
@smith The reason that `--peer` isn't in a spec file is because it's an option related to the Supervisor itself, not a particular service. Same goes for `--permanent`, `--listen-gossip`, `--listen-http`, etc. Make sense?

fnichol:
@reset I could do both (add an option and throw a nice human-useful warning like we do elsewhere). Think it's better to append a new commit on this PR or rebase into this one? I guess we could also merge this and I could follow up with a quick PR but it's nice having all the doc/code context bundled together. Hm, I think I've talked myself into rebasing

reset:
@fnichol add the `--id` flag and warning when you use it and rebase ;)

components/sup/src/main.rs
```diff
@@ -110,7 +110,7 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
                 "The listen address of an initial peer (IP[:PORT])")
             (@arg PERMANENT_PEER: --("permanent-peer") -I "If this service is a permanent peer")
             (@arg RING: --ring -r +takes_value "Ring key name")
-            (@arg PKG_IDENT_OR_ARTIFACT: +required
```
reset: 😍

fnichol: Did you see my (evil) hack to preserve the appearance of this being required without the flag? That'll be deleted soon enough, but no surprises for people in the meantime.

reset: Yes, I saw it below 😆

smith:
If I create a spec file while it's running I don't see a new service start. Is this one of the known issues? I can get on a call to show you whenever if you want to take a look.

When I restart the sup it does pick up the new thing. 

also thank you this is...
mind

fnichol:
Okay, we now have an `--id` option with updated commit and PR docs. I think we're all good here, yes?

fnichol:
@smith Ah yes, I was going to add a list of known-issues--this actually being one 😄 

From my first look into this, I'm guessing that the synchronous mode in `notify` is going to work better for us than the asynchronous mode I currently have implemented. As this wasn't the only issue, and the fact that we're behind a feature flag I was comfortable going ahead and fixing this up next. But at least I can put this in the "confirmed problem" bucket, so thank you!

adamhjk:
So one more johnny-come-lately thing. Supervisors already have an `id`, and it's the `guid` we generate for butterfly. Do we care? If we do, do we want to call this field something like `--sup-name`?

fnichol:
I'm okay with that--as you can tell by the PR it took a bit of work gymnastics to make clear what this was for while not trying to make it look too appealing.

fnichol:
`name` might sound less authoritative than `id`. At least an update to that flag should be a quicker fix

fnichol:
@adamhjk Okay, talked with @reset and we landed on a forceful yet not as inherently visible `--override-name` option name as we're…override the name! That hopefully shouldn't lead to confusion with existing internal Supervisor and/or Butterfly identifiers.

Once we're green here, I think we're good to merge. I'll update the PR text here too…

fnichol:
For any early adopters of this feature (congrats to you!) we still have several known shortcomings/issues/etc. which, as we address them, I will try to link back to this PR for clarity.

Currently on my hit-list:

* Set `PATH` for child processes properly for each service without prior starting states infecting later children (I didn't get to this yet)
* `SpecWatcher` is not triggering as well as I'd like (or at all sometimes). Going to swap the async calls to sync calls and add one more worker thread so we still get async-like polling in the main tick loop. This will help for adding spec files after start (feels like magic when you see it in action)
* Removing services by deleting specs are most likely not 100%. So far, I know that the prometheus code which registers a gauge needs a requisite "undo" action when we remove the service. There are probably other things we need to consider

Otherwise…

gif-keyboard-12575141538236895379

reset:
gif-keyboard-3672926384459081470

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/213984792).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/213984792).

fnichol:
To update the list above, here's where we are:

>> Currently on my hit-list:

* ~~Set `PATH` for child processes properly for each service without prior starting states infecting later children (I didn't get to this yet)~~
* ~~`SpecWatcher` is not triggering as well as I'd like (or at all sometimes). Going to swap the async calls to sync calls and add one more worker thread so we still get async-like polling in the main tick loop. This will help for adding spec files after start (feels like magic when you see it in action)~~
* Removing services by deleting specs are most likely not 100%. So far, I know that the prometheus code which registers a gauge needs a requisite "undo" action when we remove the service. There are probably other things we need to consider

PR #1892, opened 2017-03-03
Remove confusing install step from tutorial.

phiggins:
The first step of the prerequisites section of the tutorial is to install hab. https://www.habitat.sh/tutorials/getting-started/overview/

Later on in the tutorial, the user is walked through installing hab. https://www.habitat.sh/tutorials/getting-started/mac/setup-environment/

Signed-off-by: Pete Higgins <pete@peterhiggins.org>

PR #1978, opened 2017-03-22
Revert "Remove confusing install step from tutorial."

phiggins:
Reverts habitat-sh/habitat#1892

While the original copy is confusing, after removing it I heard from someone else new to habitat that it was confusing in a different way because now the tutorial doesn't link to the download page at all. Rather than attempt to fix things again I'll just admit I have no business writing docs and walk away. :)


thesentinels:

Thank you for contributing!

reset:
gif-keyboard-10492019305871040303

Issue #1973, opened 2017-03-21
Concurrent Services


reset:
The main thread of the supervisor currently "schedules" time to each service by allowing each service to "tick" on every server tick. This is a great first step to moving logic for servicing a service to the Service struct itself, but we now run the risk of blocking execution of other services if a service tick runs for a longer than usual period of time and, in a worst case, indefinitely blocking execution.

A service "tick" includes operations for running external code in the form of hooks and this is where the potential for blocking execution lies. We need to refactor the Manager and Service relationship so the Service executes from within it's own thread and receives information from the Manager when it's ready over a channel. Making this change may also require some refactorings to the way the http-gateway reads information about services.

This refactor is a requirement for backgrounding long running hooks (ie. Smoke Test hook)

reset:
This is a pre-requisite of #2680

PR #1987, opened 2017-03-23
Decouple http_gateway module from Manager's shared state

reset:
This commit removes the reliance of obtaining read locks on
the Manager's shared state in the http-gateway to communicate
supervisor state to a consumer. Essentially, the http-gateway
can no longer be used to interupt program flow.

* Removes `manager::State` struct and replaces it with
  `manager::FsCfg`. This is used to communicate file system config
  to the http-gateway and service threads. This struct is write once
  on boot and read only.
* Census, Butterfly, Service List, and last run health check results
  are now all cached to disk in an atomic way so the http-gateway
  can consume them.

Associated with #1973

gif-keyboard-17720029054697840210

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -20,7 +20,7 @@ pub mod spec_watcher;
 
 use std::collections::HashMap;
 use std::fs::{self, File};
-use std::io::{Read, Write};
+use std::io::{BufWriter, Read, Write};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use std::thread;
@@ -34,6 +34,7 @@ use butterfly::server::timing::Timing;
 use butterfly::server::Suitability;
 use hcore::crypto::{default_cache_key_path, SymKey};
 use hcore::service::ServiceGroup;
+use serde_json;
 use time::{SteadyTime, Duration as TimeDuration};
 use toml;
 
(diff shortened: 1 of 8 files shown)


thesentinels:

Thank you for contributing!

reset:
@adamhjk the stem of the filename is unique while the extension is generic

components/sup/src/manager/mod.rs
```diff
@@ -413,15 +461,107 @@ impl Manager {
...
+        self.persist_census_state();
+        debug!("Writing butterfly state to disk");
+        self.persist_butterfly_state();
+        debug!("Writing services state to disk");
+        self.persist_services_state();
+    }
+
+    fn persist_census_state(&self) {
+        let tmp_file = self.fs_cfg.census_data_path.with_extension("dat.tmp");
+        let file = match OpenOptions::new()
```
fnichol: Using `OpenOptions` is fine, but did you know that [`File::create`](https://doc.rust-lang.org/std/fs/struct.File.html#method.create) implements the same behavior (opens in write-only, creates file if not exists, otherwise truncates)?

reset: Ah yeah, I totally can just use that instead here - I was pulling this from another piece of code that I wrote when I also wanted to be able to read the file. Just swapped it to `File::create`!

reset:
@thesentinels approve

thesentinels:

components/sup/src/manager/mod.rs
fnichol: Noice!

fnichol:
gif-keyboard-6626283132427792382

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/214456593).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/214456593).

PR #1989, opened 2017-03-23
Add package migration and graph to builder-scheduler

chefsalim:
This change adds a couple of key features to the builder scheduler. It creates a new table in the Postgres DB to track the global package index, along with associated data store functions. In addition, it adds the capability to the builder scheduler to migrate packages into the Postgres store (via their on-disk backups). It also hooks in the reverse dependency graph, and is now able to schedule build jobs for the package reverse dependencies. Finally, it also updates the 'spider' dev tool to allow it to function against the DB, and also adds a new capability to resolve package names.

giphy 14


thesentinels:

Thank you for contributing!

fnichol:
Our project has officially reached the Taylor Swift Line--we got to PR 1989!

gif-keyboard-15680712121516960247

components/hab-spider/src/main.rs
```diff
@@ -22,38 +22,53 @@ extern crate walkdir;
...
+
+pub mod data_store;
+pub mod error;
 
 use std::io::{self, Write};
 use clap::{Arg, App};
 use time::PreciseTime;
 use bldr_core::package_graph::PackageGraph;
-use bldr_core::file_walker::FileWalker;
+//use bldr_core::file_walker::FileWalker;
```
mwrock: did you mean to delete this?

components/hab-spider/src/main.rs
```diff
@@ -22,38 +22,53 @@ extern crate walkdir;
...
+use data_store::DataStore;
 
 fn main() {
     env_logger::init().unwrap();
 
     let matches = App::new("hab-spider")
         .version("0.1.0")
         .about("Habitat package graph builder")
-        .arg(Arg::with_name("PATH").help("The path to the packages root").required(true).index(1))
+        .arg(Arg::with_name("URL").help("The DB connection URL").required(true).index(1))
```
mwrock: Instead of pointing to a DB connection and maintaining its own datastore, have you considered it connecting to the scheduler service and using its api for data access?

mwrock:
If I understand correctly, the scheduler service's package store is intended to be a global package index different from the "soon to be" originsrv package store which will be sharded.

Just curious if we have thought about how these stores will remain in sync. Or if we'll just cross that bridge when we get there (I'm hoping early next week).

adamhjk:
The primary service (origin) will have a column tracking wether the index service has written or not. It'll have a thread doing periodic, randomized back off - so it will keep sending messages to update the other service until it gets positive ack.

components/hab-spider/src/main.rs
chefsalim: I left it in as we may add the ability to 'migrate' on-disk packages from hab-spider as well. I'll remove it for now though.

components/hab-spider/src/main.rs
chefsalim: Yes, I did consider that, however it looked like it would be quite a bit more complicated to go that route and would need to add new APIs that the scheduler service doesn't actually need to be exposing.

mwrock: ok. that sounds fair!

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/215684639).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/215684639).

PR #1992, opened 2017-03-24
Add mutually exclusive process lock to Supervisor

reset:
Since the Supervisor is writing and reading from disk, we can no
longer allow multiple Supervisors of the same "name" run at the
same time. This change prevents that from happening by writing a
LOCK file to the Supervisor's service directory.

Running two supervisors is not recommended, but it is supported by
means of the `--override-name` flag on the CLI. If this is not
specified, the Supervisor will now exit with a helpful error message
explaining how to start multiple.

gif-keyboard-13954893931446060778

Related: #1973


thesentinels:

Thank you for contributing!

fnichol:
I'll take a look here shortly and get this in, thanks!

components/sup/src/manager/mod.rs
```diff
@@ -613,6 +629,19 @@ impl Suitability for SuitabilityLookup {
     }
 }
 
+fn obtain_process_lock(fs_cfg: &FsCfg) -> Result<()> {
+    match OpenOptions::new().write(true).create_new(true).open(&fs_cfg.proc_lock_file) {
+        Ok(_) => Ok(()),
```
adamhjk: We should be writing the PID of the supervisor to this lock file

components/sup/src/manager/mod.rs
```diff
@@ -613,6 +629,19 @@ impl Suitability for SuitabilityLookup {
...
+        Err(_) => Err(sup_error!(Error::NoProcessLock)),
```
adamhjk: Here we should be checking to see if the PID mentioned in the lock file is present, and if it is not, we should release the lock and try again.

fnichol: Yep, makes sense to me--it makes the file potentially actionable and has a way to determine if a stale file is legit or not.

adamhjk: Right - even if you don't put the logic in to release the lock, writing something in it thats actionable is smart.

reset:
I should have dropped a comment in here regarding why the PID is no longer in the lock file: there's no functionality in std Rust to get your own PID so you use libc. However, Windows doesn't support this mechanism.

I wanted to get this working in time for next week and I'm out until Tuesday. This should be sufficient enough but we should absolutelyake the changes that @adamhjk mentioned!

mwrock:
This functionality actually does exist on Windows. See https://github.com/habitat-sh/habitat/blob/master/components/core/src/os/process/windows.rs#L53 which calls:
```
let proc_handle = kernel32::OpenProcess(winapi::PROCESS_QUERY_LIMITED_INFORMATION |
                                                winapi::PROCESS_TERMINATE,
                                                winapi::FALSE,
                                                pid as winapi::DWORD);
```

reset:
@mwrock perfect! I knew we had to have that somewhere ;). Hold on merging this for now so I can put that in later today

reset:
@adamhjk @fnichol @mwrock this is now much more friendly and complete. Thanks to @mwrock for the Windows tips, I did some reading today on the Windows API and I feel a lot more comfortable ;)

components/net/src/server.rs
```diff
@@ -171,7 +171,7 @@ pub trait NetIdent {
 
     fn net_ident() -> String {
         let hostname = super::hostname().unwrap();
-        let pid = unsafe { libc::getpid() };
+        let pid = process::current_pid();
```
reset: I snuck this change in to this PR because it was quick and related to the work in the `core::os::process` module. This change will allow us to be  portable in the `builder-worker` barring any other Linux specific dependencies we may have ;)

adamhjk:
Awesome

components/sup/src/manager/mod.rs
```diff
@@ -613,6 +630,53 @@ impl Suitability for SuitabilityLookup {
...
+                Ok(file) => {
+                    let reader = BufReader::new(file);
+                    if let Some(Ok(line)) = reader.lines().next() {
+                        if let Ok(pid) = line.parse::<u32>() {
+                            if process::is_alive(pid) {
+                                return Err(sup_error!(Error::ProcessLocked(pid)));
+                            }
+                        }
+                    }
+                    release_process_lock(fs_cfg);
```
reset: @adamhjk technically by doing this there is a race condition where another supervisor of the same name is simultaneously attempting to obtain a process lock. It'll be exceedingly rare, but if one process hits this function while another just did this check and exited the function call below, we could still end up with two supervisors.

I think it's alright to do this for now but what do you think of using `flock` (and the Windows counterpart if existing)  instead in a future improvement?

adamhjk:
Yep - also a race where the process ids are being recycled quickly, and you
get a false positive. Flock has downsides too - most of them around the
conditions of a given file system deciding to make it a transparent no-op.
This was reasonable for sysv init for a long time - I think we will be
okay.

components/sup/src/manager/mod.rs
```diff
@@ -303,11 +313,7 @@ impl Manager {
         loop {
             let next_check = SteadyTime::now() + TimeDuration::milliseconds(1000);
             if self.check_for_incoming_signals() {
-                let mut services = self.services.write().expect("Services lock is poisend!");
-                for service in services.drain(..) {
-                    self.remove_service(&service)?;
-                }
-                outputln!("Habitat thanks you - shutting down!");
+                self.shutdown();
```
fnichol: Would it be worth considering call to this on `Drop` of `Manager`? It would have to be idempotent--and maybe that's overkill

components/core/src/os/process/windows.rs
```diff
@@ -32,6 +32,16 @@ pub fn become_command(command: PathBuf, args: Vec<OsString>) -> Result<()> {
...
 }
 
+/// Get process identifier of calling process.
+pub fn current_pid() -> u32 {
+    unsafe { kernel32::GetCurrentProcessId() as u32 }
+}
+
+/// Determines if a process is running with the given process identifier.
+pub fn is_alive(pid: u32) -> bool {
+    handle_from_pid(pid).is_some()
```
mwrock: if this `is_some` then we want to make sure to call ` kernel32::CloseHandle(handle)` before returning `true`

components/net/Cargo.toml
```diff
@@ -7,11 +7,12 @@ workspace = "../../"
 [dependencies]
 bitflags = "*"
 fnv = "*"
+habitat_builder_protocol = { path = "../builder-protocol" }
+habitat_core = { path = "../core" }
```
fnichol: Yeah! Was wondering if anyone saw that I did this with `components/sup/Cargo.toml` as a test

components/core/src/os/process/windows.rs
reset: @mwrock is that because `handle_from_pid` allocates memory outside of Rust and doesn't handle the cleanup?

components/net/Cargo.toml
reset: Totally saw it :)

components/core/src/os/process/windows.rs
mwrock: Generally good hygene. Here is a quote from the docs: 

> In general, CloseHandle invalidates the specified object handle, decrements the object's handle count, and performs object retention checks. After the last handle to an object is closed, the object is removed from the system.

The docs on `OpenProcess` state this should be done. The `child` struct does this in a `drop`.

reset: @mwrock yeah I just took a peek at the docs and saw this and updated the PR - great catch! I made the changes in a separate commit to show the cleanup more clearly

components/core/src/os/process/windows.rs
```diff
@@ -229,6 +211,32 @@ impl ExitStatusExt for HabExitStatus {
     }
 }
 
+struct ProcessHandle(winapi::HANDLE);
```
mwrock: Nice!

mwrock:
Awesome. As long as appveyor comes out the other side green, I'm good.

mwrock:
I don't mind carrying to green. I have a windows environment handy here.

reset:
@mwrock I really would appreciate that, I seriously need to get my Windows toolchain setup but don't have time this weekend. I pushed some changes which I believe will solve the compilation errors but I'm not 100% sure. Want to see them through for me if they don't make this green? Thanks so much man!

mwrock:
After cleaning up the few trivial compile issues, this ultimately hit a wall with `moving borrowed content` errors with `Child` trying to take ownership of the handle as it passed it to its win32 functions. The `ProcessHandle` wrapper could not implement `Copy` (which would solve this) because it implemented `drop`. I tried to create another wrapper that would implement `Copy` and own the handle and then have `ProcessHandle` wrap that and be in charge of the `Drop`. I kept hitting the moved content issues. I'm pretty sure there is a way to make the wrapper work but to just get this working, I'm abandoning the wrapper for a simple call to `CloseHandle` after a successful `OpenProcess`. Not as clean but works.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/215170696).

fnichol:
@mwrock Nicely done! Was fun watching this one come together.

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/215170696).

adamhjk:
@mwrock I think you would need to move it into a `RefCell` - but honestly, I think I prefer the simplicity of what you've done here, versus the addition of the implicit RwLock putting it in a `RefCell` adds.

https://doc.rust-lang.org/1.6.0/std/cell/index.html

reset:
@adamhjk I don't think that simplicity wins the day here - the new type pattern with a Drop trait is how the Rust book recommends that you provide a Safe wrapper around an FFI call like this. Since it's private we don't need to go the extra mile, but at the very least we need to mark this function as `unsafe`

Issue #1979, opened 2017-03-22
Create framework for loading and injecting "scaffolding"


rhass:
We need a way to `source` scaffolding into the build phases.

reset:
Completed!

PR #2000, opened 2017-03-27
Add support for Scaffolding to plan-build.

rhass:
Add the variable `pkg_scaffolding` and private callbacks `_load_scaffolding` and `_inject_scaffolding_dependency` to plan-build. This is the framework in plan-build to support the feature.

Implements GH-1979


thesentinels:

Thank you for contributing!

mwrock:
This looks good to me. I think before we go any further with this we should get serious about the name. I think we have agreed that it won't be blueprints.

adamhjk:
I really do think the name here should be `automatic`, with non-automatic plans being called `manual` plans. Partly because of the dual-nature we're making in plans here, and partly because it hits on one of our internal naming rules - that we call things what they are/what they do. What these do is automatically configure your plan. Any other name will require us to write *that* sentence after the name. /cc @reset @tashimi

rhass:
I just want to call out one thing really quick, unrelated to the naming stuff. This does not inject the correct dependency in to the dep chain yet. Nevertheless, I think this is good enough to at least get the ball rolling so we can add support for different languages/build tools/webapp frameworks, and I can circle back to the dependency injection later.

reset:
@adamhjk I'm cool with the name `automatic` and `manual`

mwrock:
yeah those names are great and we can use photos of stick shifts and automatic shifters as icons!

tashimi:
@adamhjk @reset ‘automatic‘ and ‘manual‘ work. Let's ship it!

rhass:
I dub thee sir automatic of habitat

Let me know if we should change any of the phrasing in the output to "automatic plan(s)" or anything like that.

mwrock:
Should we also change your doc page to reflect the naming?

adamhjk:
gif-keyboard-6714613561839233137

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1603,25 @@ _resolve_dependencies() {
   _validate_deps
 }
 
+function do_automatic() {
+  do_default_automatic "$1"
```
fnichol: This shouldn't need to take an argument--as is, `$1` is empty anyway. These hook functions tend to be declarative anyway.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1603,25 @@ _resolve_dependencies() {
...
+  return $?
+}
+
+# Loads specified automatic to extend plan DSL.
+function do_default_automatic() {
+  local automatic="${automatic:-$1}"
```
fnichol: Should this be `pkg_automatic`? I think that's what we had in the docs--makes things more consistent for the Plan author ("all plan variables are prefixed with `pkg_`"). Given that this variable comes from a `plan.sh`, we should be able to use this variable outright with a local or checking its value.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1603,25 @@ _resolve_dependencies() {
...
+
+  if [[ -n "$automatic" ]]; then
```
fnichol: Maybe use `${pkg_automatic:-}` which ends up being an empty string if this variable isn't set. At some point I'd like to update the rest of the code to this style so we can use a `set -u` on the whole program (when you reference an unused variable, the program dies).

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1603,25 @@ _resolve_dependencies() {
...
+    build_line "Loading Automatic $automatic"
+    if source "$(pkg_path_for $automatic)/bin/automatic.sh"; then
```
fnichol: I was thinking about the location for loading this file…it might make more sense to pull this out of the `PATH` of the blueprint package. Perhaps putting it under `lib/automatic.sh` (or even `libexec/automatic.sh`) that way if there were ever more than one, we wouldn't have (potentially) multiple `automatic.sh` files showing in `$PATH`. As it's a plugin of sorts it probably makes sense to think of it as a library implementation (hence why I was thinking putting it under `lib/`)

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1603,25 @@ _resolve_dependencies() {
...
+      build_line "Automatic loaded"
+    else
+      exit_with "Failed to load Automatic $automatic" $ret
```
fnichol: The `$ret` variable isn't set anywhere--maybe just pick a non-zero value here?

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -2536,6 +2555,8 @@ _resolve_dependencies
 
 _set_path
 
+do_automatic
```
fnichol: I'd argue that we want to make this an internal/private function only and not a user-facing part of our build API. Naming this `_do_automatic` or even possibly `_load_automatic` would be okay. (Sorry about the repeat here, I realized this at the end and circled back to the top)

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1603,25 @@ _resolve_dependencies() {
   _validate_deps
 }
 
+function do_automatic() {
```
fnichol: I'd argue that we want to make this an internal/private function only and not a user-facing part of our build API. Naming this `_do_automatic` or even possibly `_load_automatic` would be okay. In that case, we wouldn't need a default implementation of `do_default_automatic`--that's just how the build program works (just like we don't have a default implementation for setting the PATH because users shouldn't override this).

components/plan-build/bin/hab-plan-build.sh
rhass: Oops. That's leftover from a small refactor I did at one point and forgot to fix it. Thanks!

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1609,6 +1615,7 @@ function _load_automatic() {
 
   if [[ -n "$automatic" ]]; then
     build_line "Loading Automatic $automatic"
+    _install_dependency "$automatic"
```
rhass: oops! Ignore this line, I am about to remove it.

www/source/docs/scaffolding.md
```diff
@@ -0,0 +1,70 @@
...
+* `core/scaffolding-java8-gradle-tomcat`
+* `core/scaffolding-java8-gradle-jboss`
+* `core/scaffolding-java7-ant`
+* `core/scaffolding-python-unicorn`
+
+## Callbacks
+Each scaffolding defines a set of callbacks which are unique to the scaffolding type. This allows you to complex projects which may need to leverage different scaffoldings as needed within your project. Each of these callbacks define overrides of base level callbacks.
+
+### scaffolding_{flavor}_begin
+Each scaffolding contains a begin callback which is executed immediately while the `scaffolding.sh` file is loaded.
```
fnichol: Would this be added in a future PR? To make this work and to allow the function to be optional running something like this:

```sh
if [[ "$(type -t _scaffolding_begin)" = "function" ]]; then
  _scaffolding_begin
fi
```

in `_load_scaffolding()` should do the trick (assuming that the function name is `_scaffolding_begin` being a static name and signalling that it's internal).

rhass: @fnichol We get this behavior for free somewhat by sourcing the file, and it will execute as the file is loaded. For example:

https://github.com/habitat-sh/core-plans/pull/454/files#diff-ab80c934a69deeae9dc5fcfdc761f0b4R96
https://github.com/habitat-sh/core-plans/pull/454/files#diff-ab80c934a69deeae9dc5fcfdc761f0b4R41

The disadvantage to my approach might be with testability in the long term, so it may be worth changing it to check for a function.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1603,6 +1611,22 @@ _resolve_dependencies() {
...
+# application/project type.
+_load_scaffolding() {
+  local scaffolding="${pkg_scaffolding:-}"
+
+  if [[ -n "$scaffolding" ]]; then
+    build_line "Loading Scaffolding $scaffolding"
+    if source "$(pkg_path_for $scaffolding)/lib/scaffolding.sh"; then
+      build_line "Scaffolding loaded"
+    else
+      exit_with "Failed to load Scaffolding $scaffolding" 1
```
reset: It's true that we failed to load, but why? It would be nice to provide a helpful message that `lib/scaffolding.sh` was not found in the root of the package

www/source/docs/scaffolding.md
```diff
@@ -0,0 +1,70 @@
...
+The resulting plan will contain all the available callbacks and variables relevant to the scaffolding type.
+
+### Existing Plans
+
+To begin using Scaffolding, you will need to add the appropriate `pkg_scaffolding`.
+
+```bash
+pkg_name="my_java_app"
+....
+scaffolding="core/scaffolding-java8-maven-tomcat"
```
reset: `scaffolding` -> `pkg_scaffolding`

components/plan-build/bin/hab-plan-build.sh
rhass: Good point!

www/source/docs/scaffolding.md
fnichol: Yeah, I'd agree that a function would lead to easier testing. Something we've tried to do with core plans is to not have any adverse side effects when a `plan.sh` is sourced in other than setting the package/build-related variables (we use this when [computing](https://github.com/habitat-sh/core-plans/blob/b10d18da313a44679ea142389b1bbb13c6881807/bin/build-dependent-order.rb#L69-L81) dependent package rebuilds). I'd rather have an explicit loading function so we know what "bucket" any code execution should belong in--if it turns out there is no appropriate bucket then maybe we've missed a build phase or an internal setup event.

fnichol:
Aside from adding an internal `_scaffolding_begin` check and invocation we look all set up for scaffolding in the build system--possibly even completely set up (i.e. all modifications to the build program are handled here). If we can get this last optional load function I'm 100% thumbs up!!

gif-keyboard-6564272303338671313

gif-keyboard-3881794901460246957

gif-keyboard-9443312930772415038

rhass:
@fnichol Good to go! =)

reset:
@thesentinels approve

gif-keyboard-4823868336315547589

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/217298494).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/217298494).

PR #2038, opened 2017-04-06
Add load/unload/stop/run command to Supervisor

reset:
The multi-tenant supervisor work is nearly complete but still lacks the ability to start a transient service vs a permanent service. Currently, any services started by `hab start` are recorded as a persistent service but this should instead be transient - a service which will not be started if the supervisor itself experiences a restart.

* Add `hab sup load` command to start or mark a running service as persistent
* Add `hab sup unload` command to remove a persistent service and stop it if it is running
* Modify `hab sup start <PKG_IDENT>` command to require an argument for `PKG_IDENT` and start it's target as a transient service
* Add `hab sup run` command to start a Supervisor with no additional transient services
* Add `hab sup stop` command to stop a running permanent or transient service

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -295,6 +295,35 @@ pub fn get() -> App<'static, 'static> {
                     (@arg ORG: "The service organization")
                 )
             )
+            (@subcommand load =>
+                (about: "Load a service to be started and supervised by Habitat from a package or \
+                    artifact. Services started in this manner will persist through Supervisor \
+                    restarts.")
+                (@setting Hidden)
+            )
+            (@subcommand unload =>
+                (about: "Unload a persistent or transient service started by the Habitat \
+                    supervisor. If the Supervisor is running when the service is unloaded the \
+                    service will be stopped.")
+                (@setting Hidden)
+            )
+            (@subcommand start =>
+                (about: "Start a loaded, but stopped, Habitat service or a transient service from \
+                    a package or artifact. If the Habitat Supervisor is not already running this \
+                    will additionally start one for you.")
+                (@setting Hidden)
+            )
+            (@subcommand stop =>
+                (about: "Stop a running Habitat service.")
+                (@setting Hidden)
+            )
+            (after_help: "\nALIASES:\
+                \n    load       Alias for: 'sup load'\
+                \n    unload     Alias for: 'sup unload'\
+                \n    start      Alias for: 'sup start'\
+                \n    stop       Alias for: 'sup stop'\
+                \n"
+            )
         )
         (@subcommand studio =>
             (about: "Commands relating to Habitat Studios")
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[macro_use]
 extern crate clap;
 extern crate env_logger;
 extern crate hab;
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -497,6 +497,10 @@ fn exec_subcommand_if_called(ui: &mut UI) -> Result<()> {
         }
         ("sup", _) => command::sup::start(ui, env::args_os().skip(2).collect()),
         ("start", _) => command::sup::start(ui, env::args_os().skip(1).collect()),
+        ("service", "load") |
+        ("service", "unload") |
+        ("service", "start") |
+        ("service", "stop") => command::sup::start(ui, env::args_os().skip(2).collect()),
         _ => Ok(()),
     }
 }
--- a/components/sup/src/command/start.rs
+++ b/components/sup/src/command/start.rs
@@ -47,7 +47,6 @@
 //! See the [documentation on topologies](../topology) for a deeper discussion of how they function.
 //!
 
-use std;
 use std::path::Path;
 
 use ansi_term::Colour::Yellow;
@@ -62,10 +61,10 @@ use manager::ServiceSpec;
 
 static LOGKEY: &'static str = "CS";
 
-pub fn package(cfg: ManagerConfig,
-               service_spec: Option<ServiceSpec>,
-               local_artifact: Option<&str>)
-               -> Result<()> {
+pub fn run(cfg: ManagerConfig,
+           service_spec: Option<ServiceSpec>,
+           local_artifact: Option<&str>)
+           -> Result<()> {
     let mut ui = UI::default();
     if !fs::am_i_root() {
         ui.warn("Running the Habitat Supervisor with root or superuser privileges is recommended")?;
@@ -84,16 +83,12 @@ pub fn package(cfg: ManagerConfig,
                                                      &fs::cache_artifact_path(None),
                                                      false)?;
         }
-
-        // If we're starting with a package, then we assume that any prior specs are out of
-        // date.
-        // TODO fn: there are some assumptions to be tested here...
-        let specs_path = Manager::specs_path_for(&cfg);
-        if specs_path.exists() && specs_path.is_dir() {
-            std::fs::remove_dir_all(specs_path)?;
-        }
         Manager::save_spec_for(&cfg, spec)?;
     }
-    let mut manager = Manager::load(cfg)?;
-    manager.run()
+    if !Manager::is_running(&cfg)? {
+        let mut manager = Manager::load(cfg)?;
+        manager.run()
+    } else {
+        Ok(())
+    }
 }
--- a/components/sup/src/command/start.rs
+++ b/components/sup/src/command/start.rs
@@ -57,7 +57,6 @@ use hcore::fs::{self, FS_ROOT_PATH};
 
 use {PRODUCT, VERSION};
 use error::Result;
-use feat;
 use manager::{Manager, ManagerConfig};
 use manager::ServiceSpec;
 
@@ -72,12 +71,7 @@ pub fn package(cfg: ManagerConfig,
         ui.warn("Running the Habitat Supervisor with root or superuser privileges is recommended")?;
         ui.br()?;
     }
-
-    // TODO fn: once we remove the `Multi` feature flag a refactoring of this function will clean
-    // things up substantially. If it looks slightly awkward, this was done with an eye to the end
-    // state, not this intermediate state. Cheers!
-
-    if let Some(spec) = service_spec.as_ref() {
+    if let Some(spec) = service_spec {
         if let Some(artifact) = local_artifact {
             outputln!("Installing local artifact {}",
                       Yellow.bold().paint(artifact));
@@ -90,26 +84,16 @@ pub fn package(cfg: ManagerConfig,
                                                      &fs::cache_artifact_path(None),
                                                      false)?;
         }
-    }
 
-    if feat::is_enabled(feat::Multi) {
-        if let Some(spec) = service_spec {
-            // If we're starting with a package, then we assume that any prior specs are out of
-            // date.  TODO fn: there are some assumptions to be tested here...
-            let specs_path = Manager::specs_path_for(&cfg);
-            if specs_path.exists() && specs_path.is_dir() {
-                std::fs::remove_dir_all(specs_path)?;
-            }
-            Manager::save_spec_for(&cfg, spec)?;
+        // If we're starting with a package, then we assume that any prior specs are out of
+        // date.
+        // TODO fn: there are some assumptions to be tested here...
+        let specs_path = Manager::specs_path_for(&cfg);
+        if specs_path.exists() && specs_path.is_dir() {
+            std::fs::remove_dir_all(specs_path)?;
         }
-
-        let mut manager = Manager::load(cfg)?;
-        manager.run()
-    } else {
-        let mut manager = Manager::load(cfg)?;
-        if let Some(spec) = service_spec {
-            manager.add_service(spec)?;
-        }
-        manager.run()
+        Manager::save_spec_for(&cfg, spec)?;
     }
+    let mut manager = Manager::load(cfg)?;
+    manager.run()
 }
--- a/components/sup/src/lib.rs
+++ b/components/sup/src/lib.rs
@@ -300,8 +300,7 @@ lazy_static!{
 
 features! {
     pub mod feat {
-        const List = 0b00000001,
-        const Multi = 0b00000010
+        const List = 0b00000001
     }
 }
 
--- a/components/sup/src/lib.rs
+++ b/components/sup/src/lib.rs
@@ -304,5 +304,5 @@ features! {
     }
 }
 
-const PRODUCT: &'static str = "hab-sup";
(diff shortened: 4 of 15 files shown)


thesentinels:

Thank you for contributing!

fnichol:
@reset I'll take a look at this pretty quick. Sadly I had a slightly different approach I was working on, but I'm starting to like the verb action names here better as they are less ambiguous. If you have time today we should sync up on this.

components/sup/src/manager/service/spec.rs
```diff
@@ -225,6 +255,40 @@ impl serde::Serialize for ServiceBind {
...
 
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub enum StartStyle {
+    Persistent,
+    Transient,
+}
+
+impl Default for StartStyle {
+    fn default() -> StartStyle {
+        StartStyle::Transient
```
fnichol: I'd say we want this to default to `StartStyle::Persistent` to favor service state sticking around. If a particular invocation is intended to be transient (`hab start` perhaps), we can explicitly feed this value to override a permanent default.

components/sup/src/manager/spec_watcher.rs
```diff
@@ -29,8 +30,8 @@ use manager::service::ServiceSpec;
 
 static LOGKEY: &'static str = "SW";
 const WATCHER_DELAY_MS: u64 = 2_000;
-const SPEC_FILE_EXT: &'static str = "spec.toml";
```
fnichol: I'd have argued against dropping the `.toml` extension here. The user in mind here is someone using a configuration management tool who would be writing these files when setting up a host (the same reasoning for having `.sh` on the Plan file). So in effect, this *could* be in someone's text editor. I know that made the Rust file stem logic extra crazy (as I see you found below) but not a big deal either way. Anyway, that was my thinking.

components/hab/src/main.rs
```diff
@@ -498,6 +497,10 @@ fn exec_subcommand_if_called(ui: &mut UI) -> Result<()> {
         }
         ("sup", _) => command::sup::start(ui, env::args_os().skip(2).collect()),
         ("start", _) => command::sup::start(ui, env::args_os().skip(1).collect()),
+        ("service", "load") |
```
fnichol: It might make sense to group all Supervisor/Service-related verb-named subcommands under a `service` group, just like we do in `hab`. In other words, `hab-sup load` would be `hap-sup service load` (or similar). This would help to clarify exactly what `hap-sup service start` would do, as opposed to what `hab-sup start` (and by extension, `hab start`) might do.

components/sup/src/command/start.rs
```diff
@@ -57,27 +56,21 @@ use hcore::fs::{self, FS_ROOT_PATH};
...
 use manager::{Manager, ManagerConfig};
 use manager::ServiceSpec;
 
 static LOGKEY: &'static str = "CS";
 
-pub fn package(cfg: ManagerConfig,
-               service_spec: Option<ServiceSpec>,
-               local_artifact: Option<&str>)
-               -> Result<()> {
+pub fn run(cfg: ManagerConfig,
```
fnichol: I made the same tweak--it kept its original name for an awfully long time.

components/sup/src/main.rs
```diff
@@ -156,6 +220,69 @@ fn sub_config(m: &ArgMatches) -> Result<()> {
...
+fn sub_load(m: &ArgMatches) -> Result<()> {
+    if m.is_present("VERBOSE") {
+        sup::output::set_verbose(true);
+    }
+    if m.is_present("NO_COLOR") {
+        sup::output::set_no_color(true);
+    }
+    let cfg = mgrcfg_from_matches(m)?;
+    let ident_or_artifact = m.value_of("PKG_IDENT_OR_ARTIFACT").unwrap();
+    let ident = if Path::new(ident_or_artifact).is_file() {
```
fnichol: I don't think we need to preserve the artifact-handling here, probably only in the dev-friendly `hab-sup start` case.

components/sup/src/main.rs
```diff
@@ -156,6 +220,69 @@ fn sub_config(m: &ArgMatches) -> Result<()> {
...
+    let default_spec = ServiceSpec::default_for(ident);
+    let spec_file = Manager::spec_path_for(&cfg, &default_spec);
+    if let Ok(spec) = ServiceSpec::from_file(&spec_file) {
+        if !m.is_present("FORCE") {
+            return Err(sup_error!(Error::ServiceLoaded(spec.ident)));
+        }
+    }
+    let mut spec = spec_from_matches(default_spec.ident, m)?;
+    spec.start_style = StartStyle::Persistent;
+    Manager::save_spec_for(&cfg, spec)
```
fnichol: The non-arg parsing logic probably should go into a proper subcommand module/function. There isn't currently any user feedback from this subcommand and that messaging can be duplicated in `hab-sup start` by invoking the same logic.

components/sup/src/manager/mod.rs
```diff
@@ -291,18 +309,15 @@ impl Manager {
         Ok(())
     }
 
-    pub fn remove_service(&self, service: &Service) -> Result<()> {
-        // TODO FN: there is more to removing a service, more to follow
-        service.remove()?;
+    pub fn remove_service(&self, service: &mut Service) -> Result<()> {
```
fnichol: If we're going with load/unload we should think about updating the internal functions and references to load/unload as well. It forces the distinction between loading and starting (vs. unloading and stopping).

components/sup/src/manager/service/spec.rs
```diff
@@ -49,6 +83,16 @@ pub struct ServiceSpec {
     pub binds: Vec<ServiceBind>,
     #[serde(skip_deserializing, skip_serializing)]
     pub config_from: Option<PathBuf>,
+    #[serde(
+        deserialize_with = "deserialize_using_from_str",
+        serialize_with = "serialize_using_to_string"
```
fnichol: Way to re-use generics!

components/sup/src/manager/service/spec.rs
```diff
@@ -62,66 +106,50 @@ impl ServiceSpec {
...
     }
 
     pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
-        let file = File::open(&path).map_err(|e| {
-                         sup_error!(Error::ServiceSpecFileRead(path.as_ref().display().to_string(),
-                                                               e.to_string()))
-                     })?;
+        let file =
+            File::open(&path).map_err(|err| {
+                             sup_error!(Error::ServiceSpecFileIO(path.as_ref().to_path_buf(), err))
```
fnichol: Nice, I didn't know about `Path.to_path_buf()`--much nicer.

components/sup/src/manager/service/spec.rs
```diff
@@ -62,66 +106,50 @@ impl ServiceSpec {
...
         debug!("Writing service spec to '{}': {:?}",
                path.as_ref().display(),
                &self);
-        let dst_path = path.as_ref()
-            .parent()
-            .ok_or(sup_error!(Error::ServiceSpecFileWrite(path.as_ref().display().to_string(),
-                                                          "cannot determine parent directory"
-                                                              .to_string())))?;
+        let dst_path =
+            path.as_ref().parent().expect("Cannot determine parent directory for service spec");
```
fnichol: A reason for calling `.expect()` here? I think you argued me over to returning an error here. Given I had originally used an `.expect()` assertion I'm forced to understand but was convinced to err on the side on more extreme error handling in spec file IO.

components/hab/src/cli.rs
```diff
@@ -295,6 +295,35 @@ pub fn get() -> App<'static, 'static> {
...
+                (@setting Hidden)
+            )
+            (@subcommand unload =>
+                (about: "Unload a persistent or transient service started by the Habitat \
+                    supervisor. If the Supervisor is running when the service is unloaded the \
+                    service will be stopped.")
+                (@setting Hidden)
+            )
+            (@subcommand start =>
+                (about: "Start a loaded, but stopped, Habitat service or a transient service from
```
fnichol: looks like we're missing a `\` at the end here which wraps the line in the terminal.

components/sup/src/lib.rs
```diff
@@ -300,10 +300,9 @@ lazy_static!{
...
-        const List = 0b00000001,
-        const Multi = 0b00000010
+        const List = 0b00000001
     }
 }
 
-const PRODUCT: &'static str = "hab-sup";
-const VERSION: &'static str = env!("CARGO_PKG_VERSION");
+pub const PRODUCT: &'static str = "hab-sup";
+pub const VERSION: &'static str = env!("CARGO_PKG_VERSION");
```
fnichol: Was changing the implementation here on purpose (this won't put the `build.rs`-driven version string into the binary)?

fnichol: Ah, I see we had the `VERSION` const in bin and lib which is why we were seeing `0.0.0` in some cases. This was a pre-exiting bug no doubt--pretty sure we should be using `include_str!(concat!(env!("OUT_DIR"), "/VERSION"))` in its place.

components/sup/src/main.rs
```diff
@@ -95,13 +100,67 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
...
+            (@arg RING: --ring -r +takes_value "Ring key name")
+        )
         (@subcommand sh =>
             (about: "Start an interactive Bourne-like shell")
             (aliases: &[])
         )
         (@subcommand start =>
-            (about: "Start a Habitat-supervised service from a package or artifact")
-            (aliases: &["st", "sta", "star"])
+            (about: "Start a loaded, but stopped, Habitat service or a transient service from
```
fnichol: Missing a `\` here at the end of line.

components/sup/src/main.rs
```diff
@@ -114,26 +173,31 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
                 [default: default]")
             (@arg PEER: --peer +takes_value +multiple
                 "The listen address of an initial peer (IP[:PORT])")
-            (@arg PERMANENT_PEER: --("permanent-peer") -I "If this service is a permanent peer")
+            (@arg PERMANENT_PEER: --("permanent-peer") -I "If this Supervisor is a permanent peer")
```
fnichol: Much better and more correct. The previous description wasn't even technically true.

components/sup/src/manager/mod.rs
```diff
@@ -115,6 +115,24 @@ pub struct Manager {
 }
 
 impl Manager {
+    pub fn is_running(cfg: &ManagerConfig) -> Result<bool> {
+        let state_path = Self::state_path_from(&cfg);
+        let fs_cfg = FsCfg::new(state_path);
+
+        match read_process_lock(&fs_cfg.proc_lock_file) {
+            Ok(pid) => Ok(process::is_alive(pid)),
+            Err(SupError { err: Error::ProcessLockCorrupt, .. }) => Ok(false),
```
fnichol: I've been trying to find this destructuring idiom with our errors, nice!

components/sup/src/manager/service/mod.rs
```diff
@@ -292,36 +288,6 @@ impl Service {
         self.supervisor.check_process()
     }
 
-    pub fn register_metrics(&self) {
```
fnichol: On purpose?

fnichol:
I'm working from my source tree and notes to find any gaps in our approaches and I think I've hit most of them.

components/sup/src/main.rs
```diff
@@ -95,13 +100,67 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
...
+                "The service group; shared config and topology [default: default].")
+            (@arg DEPOT_URL: --url -u +takes_value {valid_url}
+                "Use a specific Depot URL (ex: http://depot.example.com/v1/depot)")
+            (@arg TOPOLOGY: --topology -t +takes_value {valid_topology}
+                "Service topology; [default: none]")
+            (@arg STRATEGY: --strategy -s +takes_value {valid_update_strategy}
+                "The update strategy; [default: none] [values: none, at-once, rolling]")
+            (@arg BIND: --bind +takes_value +multiple
+                "One or more service groups to bind to a configuration")
+            (@arg FORCE: --force -f "Load or reload an already loaded service. If the service was \
```
fnichol: Probably a good call, I could see why we could need something like this.

components/sup/src/manager/service/spec.rs
reset: Input validation is done at a higher level than this. We do actually mean to panic here because we're communicating back to the developer that they fucked something up if we ever hit this

components/sup/src/main.rs
reset: Oh totally, I also dropped `--config-from` because of this very reason. I'll remove this code path - it'll make things much simpler

components/sup/src/lib.rs
reset: I'll switch this one's implementation with the other - I did do this on purpose but I forgot which was the right one ;)

components/sup/src/manager/service/mod.rs
reset: Yeah, the "metrics" which were registered here were completely useless. It was a fake Gauge for "version number" of the software. We already have this in output from the http-gateway without this metric

reset:
@fnichol thanks for the awesome code review on this! Per our video chat, there's two remaining edge cases in here that I'm going to address and push up, then we can go ahead and merge

fnichol:
Checking this out now…

components/sup/src/manager/mod.rs
reset: Ya we can do that, let's hit it in another refactor if we feel the words should be the same. It's cool either way with me

components/sup/src/manager/spec_watcher.rs
reset: For anyone reading, @fnichol and I had an offline conversation about this. The `/hab` data directory should be completely managed by the `hab` CLI and nothing else. Another configuration management tool, like Chef, should never *ever* write to anything in this directory. Some of the directories here have human readable files including the `/hab/sup` directory, but the files are for the machine. Because of this, we don't need to include extensions on them for our editors. It's best to not leak implementation details in file name in this case. We may be serializing these things out as TOML today, but we could swap to something else in the future if we fancy it!

components/sup/src/manager/service/spec.rs
reset: These generic deserializers were so nice to find

fnichol:
gif-keyboard-2824787143830811919

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/219506636).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/219506636).

PR #2048, opened 2017-04-06
Add Last Value Caching (LVC) to EventSrv

### LINKED ISSUE #2023: EventSrv subscriber OnConnect receives seed of initial events
Currently, subscribers to the EventSrv will not receive any events about the system until a change is made. This would make the behaviour of something like a visualization dashboard pretty weak - it would need to receive an event from every member before populating any data.

The EventSrv will need to be changed in these three ways to achieve the desired behaviour:

* EventEnvelope will need to have a field added to identify the sending-service 
* EventSrv should store the last event for each sending-service and member-id it last received
* EventSrv should send all cached events to subscribers OnConnect

christophermaier:
When subscribers connect to the EventSrv, they will have missed
whatever events were published before they connected. Depending on how
frequently events are produced, it may take some time before a
subscriber receives any messages. This can be disorienting and
potentially misleading, particularly for dashboard-type clients.

By implementing the LVC pattern, we can cache recent messages within
the EventSrv process, and send those out to any new subscribers. The
EventSrv process itself acts as a proxy between event publishers and
event subscribers. Currently, we cache the latest messages from each
service and from each ring member, providing a "snapshot" of the
current state of affairs. Messages are received from publishers, the
cache is updated as appropriate, and the message is finally relayed on
to subscribers.

By using a ZeroMQ `XPUB` socket inside EventSrv, we can access the
special subscription messages sent by `SUB` sockets. Examining these
allows us to know when to send our cached messages out. Since there
can be duplication between the "latest messages from each service" and
the "latest messages from each ring member", we deduplicate when
sending messages to new subscribers.

Fixes #2023


thesentinels:

Thank you for contributing!

components/eventsrv/src/main.rs
```diff
@@ -12,39 +12,122 @@
...
+use zmq::{Context, SUB, XPUB};
 
 fn main() {
     let ctx = Context::new();
-    let pull_sock = ctx.socket(PULL).unwrap();
-    let push_sock = ctx.socket(PUSH).unwrap();
-    assert!(pull_sock.connect("tcp://127.0.0.1:34567").is_ok());
-    assert!(push_sock.connect("tcp://127.0.0.1:45678").is_ok());
+
+    // TODO: need to better parameterize these ports;
```
reset: There will be one EventSrv running per compute resource. Another way of saying this is - one EventSrv per supervisor on `n` instances/nodes/compute resources

reset: That said, we definitely shouldn't hardcode them either ;)

components/eventsrv/src/main.rs
```diff
@@ -12,39 +12,122 @@
...
+    // they can't be hard-coded if we're going to have multiple event
+    // servers running on a given host.
+    //
+    // Maybe we gossip this information around?
+
+    // Publishers need to connect their PUB socket to this.
+    let sub_sock = ctx.socket(SUB).unwrap();
+    assert!(sub_sock.bind("tcp://*:34570").is_ok());
```
reset: This is completely fine for now, but an `assert!()` should be considered a message to a developer. In the future, we will want to output a reason & remediation for why the socket couldn't bind

components/eventsrv/src/client.rs
```diff
@@ -32,16 +33,16 @@ use std::io::prelude::*;
...
-use zmq::{Context, PUSH};
+use rand::{thread_rng, Rng};
+use zmq::{Context, PUB};
 use protobuf::Message;
 
 use message::event::{EventEnvelope, EventEnvelope_Type};
 
 fn main() {
     let ctx = Context::new();
-    let socket = ctx.socket(PUSH).unwrap();
```
reset: This is better for sure because the client needs to connect to the server, not the other way around. However, I think you maybe want a `DEALER/REP` socket here. A `PUB/SUB` socket will send our message to *every* `SUB` socket that we're connected to and what we want to do is fair-queue among them

components/eventsrv/src/main.rs
```diff
@@ -12,39 +12,122 @@
...
-    assert!(push_sock.connect("tcp://127.0.0.1:45678").is_ok());
+
+    // TODO: need to better parameterize these ports;
+    // they can't be hard-coded if we're going to have multiple event
+    // servers running on a given host.
+    //
+    // Maybe we gossip this information around?
+
+    // Publishers need to connect their PUB socket to this.
+    let sub_sock = ctx.socket(SUB).unwrap();
```
reset: If we were to change the client to a `DEALER` this would become a `REP` socket

components/eventsrv/src/client.rs
reset: http://zguide.zeromq.org/php:chapter3#The-DEALER-to-REP-Combination

reset:
@christophermaier great implementation of LVC here and exactly what we needed. Much improved having the producing client connect to the server. See my comments on `DEALER`/`REP` and let me know what you think

christophermaier:
@reset Agreed that `PUB`/`SUB` on the event generation side isn't the way we want to go. Still reading through docs, but do you have thoughts on `DEALER`/`REP` vs. `PUSH`/`PULL` as a replacement pair?

raskchanky:
@christophermaier @reset If I'm reading the docs correctly, it sounds like DEALER sockets will fair queue, just like PUSH/PULL will.  http://zguide.zeromq.org/page:all#Recap-of-Request-Reply-Sockets

christophermaier:
@raskchanky @reset I think we only need `PUSH`/`PULL` for getting events into the EventSrv from the various producers. It's just a one-way interaction; `PUSH` and `PULL` are both unidirectional sockets (no reply is needed). Plus, for `DEALER`/`REP`, we need to mess around with envelope frames (not that that's a huge problem, but still). Thoughts?

reset:
@raskchanky @christophermaier `PUSH`/`PULL` is just fine - I incorrectly read that you couldn't bind with a `PULL` socket but you totally can. We just need to make sure the "stable" parts of the system bind and the transient things connect (i.e. EventSrv has two binding ports and both clients connect)

raskchanky:
@christophermaier @reset I agree that PUSH/PULL feels a bit more natural than DEALER/REP in the sense that the producer doesn't require a reply - it's a one way interaction.  I'm not 100% sure how that fits with @reset's comment of
>the client needs to connect to the server, not the other way around

I honestly don't have enough experience with zeromq to make a definitive call one way or the other on this.

raskchanky:
@reset Thanks for the clarification on the difference between `bind` and `connect`.  That wasn't obvious to me from the docs.

reset:
@raskchanky yep, you got it! Let's keep em `PUSH`/`PULL` - that absolutely makes the most sense

components/eventsrv/src/client.rs
```diff
@@ -32,23 +33,30 @@ use std::io::prelude::*;
...
 
-    let arg = match env::args().last() {
-        Some(a) => a,
-        None => panic!("Pass the path for the file to parse"),
-    };
+    // Everything else is treated as a port number
+    for p in args {
+        let push_connect = format!("tcp://localhost:{}", p);
+        println!("connecting to {}", push_connect);
+        assert!(socket.connect(&push_connect).is_ok());
```
raskchanky: Just for my own edification - we changed this from `bind` to `connect` because of @reset's comment about the "stable" parts of the system should `bind` and the transient parts `connect`.  Is that right?

components/eventsrv/src/lib.rs
```diff
@@ -12,13 +12,125 @@
...
+pub fn proxy(frontend_port: i32, backend_port: i32) {
+    let ctx = Context::new();
+
+    let pull_sock = ctx.socket(PULL).unwrap();
+    let pull_bind = format!("tcp://*:{}", frontend_port);
+    assert!(pull_sock.bind(&pull_bind).is_ok());
+
+    let xpub_sock = ctx.socket(XPUB).unwrap();
+    let xpub_bind = format!("tcp://*:{}", backend_port);
+    assert!(xpub_sock.bind(&xpub_bind).is_ok());
```
raskchanky: These changed from `connect` to `bind` because this is the server, which is the "stable" part of the system.  Is that right?

components/eventsrv/src/lib.rs
```diff
@@ -12,13 +12,125 @@
...
+            let timestamp = event.get_timestamp();
+            let service = event.get_service().to_string();
+            if service.is_empty() {
+                warn!("missing service: {:?}", event);
+                continue;
+            }
+
+            println!("EVENTSRV: Timestamp {}", timestamp);
+            println!("EVENTSRV: Member ID {}", member_id);
+            println!("EVENTSRV: Service {}", service);
```
raskchanky: Is it worth changing these to `debug!`?

components/eventsrv/src/lib.rs
```diff
@@ -12,13 +12,125 @@
...
+
+            // Store the bytes of the message in the cache. For the
+            // service cache, we also record the member ID, and vice
+            // versa for the member cache; these data will be used for
+            // deduplication of messages being sent to new subscribers.
+            service_cache.insert(service.clone(), (member_id, bytes.clone()));
+            member_cache.insert(member_id, (service, bytes.clone()));
+
+            println!("EVENTSRV: Service Cache {:?}", service_cache.keys());
+            println!("EVENTSRV: Member Cache {:?}\n", member_cache.keys());
```
raskchanky: Same here re: `debug!`

components/eventsrv/src/main.rs
```diff
@@ -13,38 +13,27 @@
...
+    let port2 = args.remove(1);
+    let backend_port: i32 = port2.parse().unwrap();
+
+    assert!(frontend_port != backend_port);
+
+    println!("Frontend port is {}", frontend_port);
+    println!("Backend port is {}", backend_port);
+    println!("Starting proxy service...");
+
+    habitat_eventsrv::proxy(frontend_port, backend_port);
```
raskchanky: I love the way you cleaned up this file.  The distinction now between `main.rs` and `lib.rs` is fantastic.

components/eventsrv/src/subscriber.rs
```diff
@@ -41,10 +49,11 @@ fn main() {
...
                 let timestamp = event.get_timestamp();
-
-                println!("Timestamp {}", timestamp);
-                println!("Member ID {}\n", member_id);
-                debug!("{}\n", received_payload);
+                let service = event.get_service();
+                println!("SUBSCRIBER: Timestamp {}", timestamp);
+                println!("SUBSCRIBER: Member ID {}", member_id);
+                println!("SUBSCRIBER: Service {}", service);
+                println!("SUBSCRIBER: Payload {}\n", received_payload);
```
raskchanky: Should these `println!`s be `debug!`s?

components/eventsrv/src/client.rs
christophermaier: Yup, exactly.

components/eventsrv/src/subscriber.rs
christophermaier: @raskchanky I was thinking about removing them; what's the project's general approach to logging?

raskchanky: @christophermaier From what I've seen so far, it's common to use `debug!` for noisy output that doesn't need to be seen by the end user, but would be helpful when debugging.  If it doesn't need to be seen by the end user _and_ it's also not helpful when debugging, then just get rid of it.

raskchanky:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/219890981).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/219890981).

reset:
gif-keyboard-1580674926938910351

PR #2064, opened 2017-04-08
Remove `pkg_source` from required plan variables.

rhass:
The pkg_source variable is primarily only useful when building applications which are built from tarballs.

In cases were users build from local source, I think URI notation for this variable is the most desired option. However, there are many situations such as interpreted languages in which this variable doesn't serve a good purpose. Moreover, since the plan validation is run before we can inject scaffolding, we cannot optionally override the behavior in the validation for this variable for specific languages.

See original discussion in Slack as well:
https://habitat-sh.slack.com/archives/C0VBY57UP/p1491604342661130

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -2657,7 +2657,6 @@ build_line "Validating plan metadata"
 required_variables=(
   pkg_name
   pkg_origin
-  pkg_source
   pkg_version
 )
 for var in "${required_variables[@]}"


thesentinels:

Thank you for contributing!

mwrock:
Currently `pkg_source` is only used in the default download function and tangentally in the default verify and unpack functions via `pkg_filename` which can be derrived from `pkg_sourc` if not explicitly provided. If a plan's source is local, these values are not used so it makes sense to me for this variable tro be optional. We might consider having the default download function complain if it is not defined which should only occur if that function is called.

robbkidd:
> the default download function complain if it is not defined which should only occur if that function is called.

As an author of plans that mostly work with local source, I'm a fan of optional but bark if default download doesn't have a `pkg_source` URL to work with.

rhass:
This was implemented in 33c79bb17b0bd6b3002ff9d04c164f12407c9b9c

Issue #2073, opened 2017-04-10
Adding dependencies in scaffolding fails


smith:
In a scaffold, if you do something like:

```
_scaffolding_begin() {
  pkg_deps=(${pkg_deps[@]} core/coreutils core/node)
}
```

and then try to run a program in one of those packages' `pkg_bin_dirs`, it will fail, not able to find that program.

This is because hab-plan-build calls `_set_environment` before it calls `_load_scaffolding`, and `_set_environment` uses the values of `pkg_deps` and `pkg_build_deps` to populate the environment variables such as PATH.

If you add packages to those arrays after `_set_environment` has been called, it won't set environment variables for them.

rhass:
@smith Would it be possible for me to work with you tomorrow to try out my fix against your code as well? You might be able to do it on your own, but I have been having to set `HAB_ORIGIN=core` to use plan-build with the changes since there are a few internal bits which seem expect `core` for `plan-build` rather than any random origin.

PR #2075, opened 2017-04-10
Fix dependency issues in Scaffolding.

rhass:
This moves the call to `_load_scaffolding` to before the ` _set_environment` call and adds a call to rerun the `_resolve_dependencies` to the `_load_scaffolding` function after the Scaffolding
has been sourced into plan-build.

By re-running `_resolve_dependencies` we ensure any build-time or run-time dependencies which are defined within the Scaffolding are downloaded and properly added to the path. This is common for interpreted languages which have tools to download and install packages which are written in the targeted language, thereby requiring the interpreter to be installed during both build and runtime.

Example `lib/scaffolding.sh`:
```bash
scaffolding_python_pkg="core/python"

_scaffolding_begin() {
  pkg_deps=(${pkg_deps[@]} $scaffolding_python_pkg)
  pkg_build_deps=(${pkg_build_deps[@]} $scaffolding_python_pkg)
}
```
Resolves Issue #2073.

Signed-off-by: Ryan Hass <rhass@users.noreply.github.com>

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1763,6 +1763,9 @@ _load_scaffolding() {
   if [[ "$(type -t _scaffolding_begin)" == "function" ]]; then
     _scaffolding_begin
   fi
+
+  build_line "Resolving Scaffolding dependencies"
+  _resolve_dependencies
 }
 
 # **Internal**  Build `$PATH` containing each path in our own
@@ -2768,11 +2771,11 @@ _inject_scaffolding_dependency
 # Download and resolve the dependencies
 _resolve_dependencies
 
-# Set up runtime environment
-_set_environment
-
 # Load scaffolding plans if they are being used.
 _load_scaffolding
+
+# Set up runtime environment
+_set_environment
 
 # Download the source
 mkdir -pv "$HAB_CACHE_SRC_PATH"


thesentinels:

Thank you for contributing!

smith:
This works, but if you put a package in `pkg_deps` in the scaffolding, then put the same package in the `pkg_deps` in the plan itself, you'll get a confusing error:

```
The following runtime dependencies have more than one version
release in the full dependency chain:

  * core/coreutils ( core/coreutils/8.25/20161208223423 core/coreutils/8.25/20161208223423 )
  * core/node ( core/node/6.9.5/20170314063939 core/node/6.9.5/20170314063939 )
```

We should do some kind of `uniq` on the pkg_deps and pkg_build_deps arrays to make sure this doesn't happen, because you should not have to know what dependencies are declared in a scaffolding you're using in normal cases.

fnichol:
@smith We'll need to be very careful about what goes in the Plan's `pkg_deps` as this set is scanned for multiple instances with differing versions. It's a little strange, but you need to remember that a scaffolding's run deps become a Plan's build deps because from the point of view of a scaffolding package, that is what it needs to do its work (a scaffolding's "runtime" is during the "buildtime" of another package).

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1763,6 +1763,9 @@ _load_scaffolding() {
   if [[ "$(type -t _scaffolding_begin)" == "function" ]]; then
     _scaffolding_begin
   fi
+
+  build_line "Resolving Scaffolding dependencies"
+  _resolve_dependencies
```
rhass: I think that is probably correct. I may have an issue in my testing methodology in which I am attempting to simulate a clean studio by doing `rm -rf /hab/pkgs/core/python*` to ensure the deps get downloaded. This could be introducing a red herring on my part if we assume the TDEPS are installed when the direct dependencies are already installed.

Issue #1983, opened 2017-03-23
RFC: Post-Run hook for service initialization


bodymindarts:
For many data services creation of specific users / roles or datastores is required. This needs to happen once the service has already started.
Proposal: add a post-run hook that will get executed after initial startup.

Issue #1393, opened 2016-10-25
Proposal: Habitat post run hook


dmajere:
Sometimes we need to run not only some tasks before application will run, but also after.
It's would be nice to have postinit hook for habitat.
So the flow would be
init -> run -> post

bdangit:
what could those tasks be?

dmajere:
Well, I have it mostly with erlang applications, when you need to start erlang node first and then do something with it. A good example, probably, is rabbitmq, you need to have it started before you'll be able to do rabbitctl create_vhost and so on.

bdangit:
fair point! I've seen some plans where people will do this during `init` hook where they would start the service under dameonize mode, configure it, shut it down and then let `run` run the service under `hab-sup`.  Its clunky but it does qualify as "init".  

The `post` might be tricky unless `hab-sup` is able to spawn another task to do what you want.

eeyun:
I'm very curious about this issue. I'd like to get some more community weigh in on it. It would be good to discuss whether or not its a realistic thing. It _might_ be tricky. Unfortunately the supervisor is the piece of the codebase I have the LEAST exposure to and is simultaneously the most in flux currently :).

reset:
@dmajere we've been adding lifecycle hooks as we needed them but haven't found a true use for `after-run` just yet. The problem that I see with adding any lifecycle hooks which happen after `run` is that you lose the ability to coordinate in case of a failure. Basically anything that would happen in `after-run` can't be critical to the service.

Initializing a database's schema is actually not the job of the database service itself, but the consumer of the database service. We actually added a topology for just this called the "initializer topology". Also as @bdangit points out, the init hook can be leveraged in clever ways to ensure all initialization steps are taken before the service runs.

I'm all for adding additional lifecycle hooks - let's just make sure we've got some concrete examples of where to use them. If we need to modify the `init` hook into multiple steps to add clarity we could investigate that as well.

reset:
Closed by #2088

PR #2088, opened 2017-04-12
Add PostRun hook to hook table

bodymindarts:
Closes https://github.com/habitat-sh/habitat/issues/1983

Signed-off-by: Justin Carter <justin@starkandwayne.com>

### DIFF
--- a/components/sup/src/manager/service/hooks.rs
+++ b/components/sup/src/manager/service/hooks.rs
@@ -379,6 +379,60 @@ impl Hook for RunHook {
 }
 
 #[derive(Debug, Serialize)]
+pub struct PostRunHook {
+    render_pair: RenderPair,
+    stdout_log_path: PathBuf,
+    stderr_log_path: PathBuf,
+}
+
+impl Hook for PostRunHook {
+    type ExitValue = ExitCode;
+
+    fn file_name() -> &'static str {
+        "post-run"
+    }
+
+    fn new(service_group: &ServiceGroup, pair: RenderPair) -> Self {
+        PostRunHook {
+            render_pair: pair,
+            stdout_log_path: stdout_log_path::<Self>(service_group),
+            stderr_log_path: stderr_log_path::<Self>(service_group),
+        }
+    }
+
+    fn handle_exit<'a>(&self,
+                       service_group: &ServiceGroup,
+                       _: &'a HookOutput,
+                       status: &ExitStatus)
+                       -> Self::ExitValue {
+        match status.code() {
+            Some(code) => ExitCode(code),
+            None => {
+                outputln!(preamble service_group,
+                    "{} exited without a status code", Self::file_name());
+                ExitCode::default()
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
Should we change the state of the service or delay its completed initialization until the post run hook has completed?

My concern is that once the run hook has completed, other services can bind to this service and the service may not be "prepared" for binding until the post run hook completes.

bodymindarts:
@mwrock thats a good point but I would like to mention that currently the `initialized`status of a service has no baring whatsoever on the rest of the system. Even though the gossiped messages have a field for that it is never being populated or read out.
That doesn't mean we shouldn't think about this, just that any decision wouldn't come to bare until that information actually gets gossiped.

That being said what do you think is correct:
`hooks/init` gets repeated until it succeeds
-> `hooks/run`
-> `hooks/post-run` until it succeeds.
Or should the entire 
`hooks/init`
`hooks/run`
`hooks/post-run`

sequence get repeated until all have succeeded?

mwrock:
Yeah looking at the flow again I see what you are saying. `initialized` does not even state that or guarantee that the service run hook has completed. I'm fine with this as is but am curious what others think before merging. @reset had some comments on this topic in https://github.com/habitat-sh/habitat/issues/1393 and I am curious what his thoughts are here.

bodymindarts:
Actually I understand the concerns that @reset mentions in that ticket.

I definitely see a need for this hook but it makes sense to discuss what it
means for a service to be initialized and what the expected behavior is.

In the context of 1 service this doesn't have much baring but in systems
with multiple collaborators it is very important to have clear contacts.
Otherwise there is no way to reliably orchestrate longer life cycles
including updates of dependant (potentially clustered services) services.

In BOSH there is the equivalent 'post-start' Skript and it is considered
part of the initialization.
Eg. When updating a cluster and a post-start script falls then the whole
update will terminate (leaving the remaining nodes in the old state). When
an operator continues the update ('bosh deploy) will be re start the
proceces on the ndes that failed (including the post-start script) before
attempting to update the rest of the nodes.

reset:
@mwrock @bodymindarts I think this is a perfectly acceptable implementation and I think it actually prevents what I was worried about in #1393. What I didn't want to happen was for the service to start and then a rumor to be sent which indicated to it's peers that all was good. This will actually block after service start for the hook to complete before that rumor is sent.

gif-keyboard-14807548089388691765

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/223261398).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/223261398).

Issue #2033, opened 2017-04-03
The builder-scheduler should parallelize dependency rebuilds


chefsalim:
This is a work item to be able to schedule builds in a more parallel manner.  Packages that are ready to be built, and have their dependencies already re-built should be scheduled in paralle.

PR #2093, opened 2017-04-13
Dispatch jobs in parallel

chefsalim:
This change allows the builder-scheduler to dispatch multiple projects at a time to the job scheduler. This allows for group builds to proceed in parallel (as long as the dependencies are ready).

NOTE: This change has not been fully tested yet.


thesentinels:

Thank you for contributing!

components/builder-scheduler/src/data_store.rs
```diff
@@ -259,15 +291,19 @@ impl DataStore {
         let mut rng = thread_rng();
         let id = rng.gen::<u64>();
 
-        conn.execute("SELECT insert_group_v1($1, $2)",
-                     &[&(id as i64), &project_names])
+        let (project_names, project_idents): (Vec<String>, Vec<String>) =
+            project_tuples.iter().cloned().unzip();
+
+        conn.execute("SELECT insert_group_v1($1, $2, $3)",
```
mwrock: this should probably be performed in a transaction

chefsalim: @mwrock Per Postgres docs, the function is atomic - "Functions and trigger procedures are always executed within a transaction established by an outer query — they cannot start or commit that transaction, since there would be no context for them to execute in"

adamhjk: I think the question here is if you are manipulating more than one table within insert_group_v1, you might need to run it from a transaction in order to roll back on failure.

chefsalim: Yeah I may be mis-understanding here, but are Postgres functions not atomic (and therefore do not support sub-transactions within functions)? Ie, if there is an error within the function and a rollback happens, it will roll back everything everything (ie, both table changes).

reset:
Closes #2033

components/builder-scheduler/src/data_store.rs
mwrock: My concerns was that if there was a failure, you could end up with only SOME of the rows added but on second thought, that may not necessarily be a bad thing in this case.

chefsalim: That does not seem possible given the stated postgres function rules.  I'll test it out and report back the findings.

mwrock: and my apologies for not reading the above more carefully! if the functions are atomic than we are totally good.

chefsalim: OK I tossed in a RAISE exception after the first loop statement and run the function, it cleanly rolls everything back.

mwrock:
This LGTM. But looks like travis is angry. probably a merge issue?

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/222940221).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/222940221).

PR #2074, opened 2017-04-10
[plan-build] Make `pkg_source` optional so local builds are first class.

fnichol:
*Note* This is a non-breaking change, meaning that all existing Plans
should continue to operate as before.

This change drops the requirement of including the `pkg_source` variable
in a Plan and alters the default behavior of the corresponding build
phases (namely `do_download()`, `do_verify()`, and `do_unpack()`) to do
the right thing.

Rationale
---------

The hypothesis here is that the vast majority of Plans will exist
alongside application source code, much like adding a project `Gemfile`
for Ruby dependencies or a `.travis.yml` for CI instructions. In that
sense, from the point of view of the Plan file, I am building **my**
application, not someone else's.

Previous to this change the build system (and by extension the defaults
in Plan files) optimized for building **other** software (not **my**
software) such as third party services like Redis, MySQL, or
intermediate software libraries like OpenSSL, zlib, etc. In these
cases, the `plan.sh` was not co-located with the source code meaning
that some source artifact needed to be downloaded, verified, and
extracted. In the case where the `plan.sh` is hosted with the source
code, there is nothing to download when building--the source tree is
already mounted into the Studio under `/src` and ready to be used.

In a CI or build environment such as Travis, Jenkins, or Builder (the
Habitat distributed build service), an SCM clone will have already
occurred before a package build will be invoked via `hab pkg build`. If a
clean checkout is provided, then there should be no additional work that
needs to happen inside the Studio instance with the help of the build
program. This means we can operate directly on the mounted-in directory
when performing a build which was part of the original Studio design.

Net-New Plan Author Variables
-----------------------------

As a part of implementing this feature, 2 new user-facing variables are
introduced which a Plan author *might* be inclined to use when writing a
Plan. The internals of the build program have been updated to use these
directly in an attempt to make more clear what is happening and where.
The variables are:

* `$SRC_PATH`: This is the directory that contains the source code to
  be built. During each relevant build phase (namely `do_prepare()`,
  `do_build()`, and `do_install()`), the current working directory will
  be `$SRC_PATH`. For local applications this will be the same directory
  which contains either `./plan.sh` or `./habitat/plan.sh`. For external
  source-downloaded software, this will be in the cache path, as
  described next.
* `$CACHE_PATH`: This is a directory which will be freshly created each
  time the build program is executed and lives under the
  `/hab/cache/src` directory. Its value is derived from the
  `$pkg_dirname` variable and is exactly equivalent to
[...]

PR #2101, opened 2017-04-14
Add `disable_callbacks` helper.

rhass:
I found myself writing things like this a lot:

```
do_download() {
  return 0
}

do_verify() {
  return 0
}

do_unpack() {
  return 0
}
```

In the spirit of scaffolding I decided to add this helper to simplify this common task. This function makes disabling callbacks easier to read and perform, while making the call more DRY with a simple one-line call: `disable_callbacks "do_download" "do_verify" "do_unpack"`

Signed-off-by: Ryan Hass <rhass@users.noreply.github.com>

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1636,6 +1636,26 @@ pkg_interpreter_for() {
     return 1
 }
 
+# Disables callbacks if they exist. Accepts multiple arguments.
+# Callbacks must be prefixed with `do_` to provide some level of
+# safety for disabling functions.
+# ```
+# disable_callbacks "do_download" "do_verify" "do_unpack"
+# ```
+disable_callbacks() {
+  local callback_names
+  callback_names=("$@")
+
+  for callback in "${callback_names[@]}"; do
+    if [[ "$callback" =~ ^do_.* ]] && [[ "$(type -t "$callback")" == "function" ]]; then
+      build_line "Disabling callback $callback"
+      eval "$callback() { return 0; }"
+    else
+      build_line "Unable to disable callback $callback"
+    fi
+  done
+}
+
 # ## Build Phases
 #
 # Stub build phases, in the order they are executed. These can be overridden by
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

rhass:
@mwrock I am not sure if we can do something like this in Windows/PowerShell or not, but it would be nice if we can.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1580,9 +1583,17 @@ fix_interpreter() {
     local targets=$1
     local pkg=$2
     local int=$3
-    local interpreter_old=".*${int}"
-    local interpreter_new="$(pkg_interpreter_for ${pkg} ${int})"
-    if [[ -z $interpreter_new || $? -ne 0 ]]; then
+    local interpreter_old
+    local interpreter_new
```
rhass: Oops! This does not belong in here.

mwrock:
sure can:
```
C:\dev\habitat\components\sup [master]> function some-func() {write-host "yo"}
C:\dev\habitat\components\sup [master]> Set-Content -Path Function:\some-func -Value "write-host 'yo2'"
C:\dev\habitat\components\sup [master]> some-func
yo2
```

bdangit:
can this be merged in? It would be one less thing I have to build to play with scaffolding next time Hab is released.

fnichol:
Hey all, we discussed this in our triage yesterday but I forgot to close this loop. In short, we're going to close this for now and keep an eye on the use cases that may warrant this in this future.

To recap the conversation, we don't want Plan authors to be no-op'ing those 3 build phases anymore as this was resolved as of #2074 which is shipped in the 0.21.0 release. If there continue to be situations where Plan authors need to constantly skip default build phases, we need to know about it because chances are there is a much better fix we can land. Additionally (and speaking for myself), I'd rather limit additional helper functions which perform a reasonably straight forward task which we should be avoiding in the first place.

@bdangit per your comment above, it's our hope that using a Scaffolding would mean that you shouldn't have to do much extra in your Plan. If using a Scaffolding means that you need need to skip build phases, then we have more work to do with Scaffolding. I'll be submitting a PR in the next day or so with a first pass at a Ruby (Rails/Rack) Scaffolding which might help to explain where I'm coming from.

bdangit:
thanks @fnichol.  this makes sense.  My python plans that depended on the scaffolding-python depended on this.  Sounds like https://github.com/habitat-sh/core-plans/issues/468 will need to get re-worked.

PR #2116, opened 2017-04-18
[plan-build] Add Scaffolding direct deps as direct build deps for Plans.

### LINKED ISSUE #2097: Investigate adding Scaffolding direct run deps as direct Plan build deps

fnichol:
This change aims to fulfill a design goal of Scaffolding: within reason,
a Plan author should be able to do what a Scaffolding package provides
directly in their `plan.sh`. A secondary design goal is that Scaffolding
package authors should not need to call any internal API functions from
the build program. The latter goal surfaced a design issue in the
previous Scaffolding implementation around direct vs. transitive
dependencies.

There are several helper functions that a Plan author (and therefore
Scaffolding package author) may use that are marked as part of the
public API. For example, the `pkg_path_for()` will return the absolute
path to the pkg_prefix for a given dependency. The implementation for
`pkg_path_for()` (correctly) only looks at **direct** build and run
dependencies--not indirect/transitive dependencies. If a Plan author is
using a Ruby Scaffolding package that pulls in Bundler for build-time
dependency resolution, then there is a good change they would want to
call something like: `pkg_path_for core/bundler`. The problem here, is
that the previous implementation would fail the `pkg_path_for()` call and
report that the `core/bundler` dependency is not a direct run or build
dependency.

This change injects the Scaffolding pacakge into the direct build
dependency array `${pkg_build_deps[@]}` as before, but now also adds
each direct run dependency for the Scaffolding package into the direct
**build** dependency array of the Plan. For example, a Ruby Scaffolding
package that pulls Bundler in as its run dependency, would yield a Plan
`${pkg_build_deps[@]}` array which contained `core/ruby` as well as
`core/bundler/1.13.7/20170104000124`. Note that these Scaffolding
dependencies are full qualified in order for the same exact versions of
Scaffolding dependencies to be captured as Plan build dependencies.

Finally, as the number of dependency resolving functions has grown, the
activity of resolving dependencies is captured in the
`_resolve_dependencies()` internal function to be used in the main
program flow.

---

There are additional refactoring commits which have attached comments for more detail.

Closes #2097

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -736,6 +736,33 @@ _get_tdeps_for() {
   return 0
 }
 
+# **Internal** Returns (on stdout) the `DEPS` file contents of another locally
+# installed package which contain the set of all direct run dependencies. An
+# empty set could be returned as whitespace and/or newlines.  The lack of a
+# `DEPS` file in the desired package will be considered an unset, or empty set.
+#
+# ```
+# _get_deps_for /hab/pkgs/acme/a/4.2.2/20160113044458
+# # /hab/pkgs/acme/dep-b/1.2.3/20160113033619
+# # /hab/pkgs/acme/dep-c/5.0.1/20160113033507
+# # /hab/pkgs/acme/dep-d/2.0.0/20160113033539
+# # /hab/pkgs/acme/dep-e/10.0.1/20160113033453
+# # /hab/pkgs/acme/dep-f/4.2.2/20160113033338
+# # /hab/pkgs/acme/dep-g/4.2.2/20160113033319
+# ```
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/223306203).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/223306203).

reset:
gif-keyboard-3733563235945033432

Issue #2125, opened 2017-04-18
Support running services under different accounts on windows


mwrock:
This is something we currently support on linux via `hab_svc_group` and `hab_svc_group`. The windows supervisor simply launches services under the same identity that the supervisor itself runs as. We do this because starting a windows process under a different identity is "hard" especially in rust and may require some kind of management of the identities password.

IMO, providing the ability to start services under a different identity may be the #1 blocker preventing users from running habitat in production on windows.

We need to investigate:

* do we need to store the user password
* if so, where (secure string in a condig?, Winsdows "Private Data"?)
* rewrite windows process creation to use a Win32 API that supports alternate credentials (not trivial)

mwrock:
I've done some research as to how to best approach this and here is how I think we should proceed:

* `hab sup start` and `hab sup load` will take an additional optional `password` argument on Windows.
* The cli will use the Windows `DPAPI` functions to encrypt the password and add the encrypted string to the service spec file
* The supervisor will use DPAPI to decrypt this string when starting/restarting services and use it to generate a logon token for creating a process.

**Security Implications and Limitations**:
* Running services under local `SYSTEM` or `NETWORK` account will not require a password
* The encrypted password can only be decrypted on the local machine. It will be important that we make this point clear. While `DPAPI` does allow one to restrict the decryption operations to the same user that encreypted the data, it can be assumed that the user running the cli and the user of the running supervisor may likely be different.
* Its possible to add a byte array to the encryption/decryption operations to add "complexity" to the encryption. We could conceivably add data here to more tightly lock down the encryption. However, the challenge is that it would need to be something specific to the local hab cli and supervisor processes and not discoverable via source code.
* Note that I looked into the possibility of using X509 certs but this is not supported for local windows logons. The closest mechanism would be smart card certs which are not a good fit for services.

smurawski:
@mwrock I like the direction. however, rather than DPAPI, couldn't we require service group encryption and use that to keep the password in sync but confidential between member of the service group?

mwrock:
[...]

Issue #2115, opened 2017-04-18
Stopping/Unloading Services on windows should kill child processes


mwrock:
Some services like mysql to dot completely clean up processes when the siupervisor stops or the service is unloaded. `myslqd.exe` is still running. Curiously this does not happen for all services. A dotnet app service when unloaded will succesfully stop `dotnet.exe`. I currently work around this with mysql by running: `Taskkill /IM hab-sup.exe /T /F` which explicitly kills any child process of `hab-sup`.

mwrock:
Here is what is happening: When a windows supervisor service stops, the supervisor:
1. tells itself to ignore any ctrl+C signal
1. Broadcasts a ctrl+C  signal
1. All processes that share the supervisor's console will respond to the signal
1. The supervisor tells itself it is now ok to respond to future ctrl+c signals (in case someone actively uses tctrl+c on the supervisor)

MySql is known for not responding to ctrl+c (this is fairly rare) and needs to be forcefully terminated. So what happens is the ctrl+c signal is only honored by the powershell.exe process that the run hook started and that in turn started mysqld.exe.

mwrock:
So this all unfortunately uncovers a rather major flaw in the windows supervisor that was fine when running one service per supervisor but may be dangferous in a multi-service supervisor world. That is that the ctrl+c event gets sent to ALL services. So stopping/unloading a single service will effectively stop all the other services as well. Fortunately its a "graceful" stop and the supervisor will immediately restart the services it did not mean to stop but this is still very bad behavior.

So we really have 2 problems here: dont stop everyone and if a child of someone we WANTED to stop does not stop, forcibly kill it.

mwrock:
For console style applications on windows, sending a "console control event" is the idiomatic windows way to gracefully terminate console applications. Applications with real windows act differently and can receive a `WM_CLOSE` event for graceful shutdown. For the habitat scenario, we will almost always (if not absolutely always) be dealing with console applications. There is no way to prevent individual processes in a single console from receiving signals UNLESS you are THAT process and you can tell yourself to ignore them. Since we do not control the source code, we forfeit this control over our service processes.

What we do have control over is:
* whether a service is spawned in the supervisr console, a new console or a detached console (no console)
* We (the supervisor) can detach itself from its current console and attach to another console

So here is a proposed solution:
[...]

PR #2129, opened 2017-04-19
Take full control over windows process creation

mwrock:
This sets the stage for #2115 and #2125 and gives us total control over windows process creation. This PR does not change any behavior or add any functionality but lays tyhe groundwork for some future work.

It turns out that rust 1.16.0 exposed more of the windows create process API. However, in order to create processes that are owned by different users, the `CreateProcessW` API call will not work. We will need to use another call instead. There are a couple options there and we can decide on the best one when tackling #2125. Most of the parameters will match `CreateProcessW`.


thesentinels:

Thank you for contributing!

Issue #2131, opened 2017-04-19
Run supervisor in the background inside a studio


mwrock:
This issue comes from inspiration out of #2104 and is aimed to make service development a smoother experience inside of a studio.

When a user enter's a studio via `hab studio enter` we should call `hab sup run` to start a supervisor in the background and pipe its output to a file. As a developer iterates on their service(s), they can `load`/`unload` services and the running supervisor should pickup/drop those services.

Obviously the supervisor output will not be streaming out to the studio. Its a reasonable assumption that a developer will actually want to use the studio and interact with it without streaming text getting in the way. It is also reasonable to assume that the developer will want to see the output from time to time. To do so, we will provide a `tail_sup` script that will stream the tail of the supervisor log file.

mwrock:
closed via #2133

PR #2133, opened 2017-04-20
Run the supervisor in a studio

mwrock:
This starts an empty supervisor inside of an interactive studio. The supervisor is started in the background and is terminated when the studio exits. A convenience function is added to the studio profile `tail_sup` (very open to better names) which runs `tail -f` on the log file that the supervisor output is piped to.

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -679,9 +679,23 @@ enter_studio() {
     set -x
   fi
 
+  echo "$studio_supervisor_start_command" | $bb chroot "$HAB_STUDIO_ROOT" \
+    $studio_env_command -i $env $studio_run_command
+
+  trap stop_supervisor EXIT
+
   # Become the `chroot` process
-  exec $bb chroot "$HAB_STUDIO_ROOT" \
+  $bb chroot "$HAB_STUDIO_ROOT" \
     $studio_env_command -i $env $studio_enter_command $*
+}
+
+# **Internal** Run when an entered studio exits.
+stop_supervisor() {
+  lock_file="$HAB_STUDIO_ROOT/hab/sup/default/LOCK"
+
+  if [ -f $lock_file ]; then
+    $bb kill $($bb cat $lock_file)
+  fi
 }
 
 # **Internal** Run a build command using a Studio.
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -6,6 +6,7 @@ studio_build_environment=
 studio_build_command="record \${1:-} $HAB_ROOT_PATH/bin/build"
 studio_run_environment=
 studio_run_command="$HAB_ROOT_PATH/bin/hab pkg exec core/hab-backline bash -l"
+studio_supervisor_start_command="start_supervisor"
 
 pkgs="${HAB_BACKLINE_PKG:-core/hab-backline}"
 
@@ -65,6 +66,7 @@ finish_setup() {
   local coreutils_path=$(_pkgpath_for core/coreutils)
 
   $bb mkdir -p $v $HAB_STUDIO_ROOT$HAB_ROOT_PATH/bin
+  $bb mkdir -p $v $HAB_STUDIO_ROOT$HAB_ROOT_PATH/sup/default
 
   # Put `hab` on the default `$PATH`
   _hab pkg binlink --dest $HAB_ROOT_PATH/bin core/hab hab
@@ -83,6 +85,12 @@ finish_setup() {
 exec $HAB_ROOT_PATH/bin/hab pkg exec core/hab-plan-build hab-plan-build \$*
 EOF
   $bb chmod $v 755 $HAB_STUDIO_ROOT$HAB_ROOT_PATH/bin/build
+
+  $bb cat <<TAIL_SUP > $HAB_STUDIO_ROOT$HAB_ROOT_PATH/bin/slog
+#!$bash_path/bin/sh
+exec tail -f /hab/sup/default/log.out
+TAIL_SUP
+  $bb chmod $v 755 $HAB_STUDIO_ROOT$HAB_ROOT_PATH/bin/slog
 
   # Set the login shell for any relevant user to be `/bin/bash`
   $bb sed -e "s,/bin/sh,$bash_path/bin/bash,g" -i $HAB_STUDIO_ROOT/etc/passwd
@@ -104,6 +112,20 @@ export TERMINFO=$(_pkgpath_for core/ncurses)/share/terminfo
 if [ -f /src/.studiorc ];then
   source /src/.studiorc
 fi
+
+start_supervisor() {
+  if [ -z \$NO_BG_SUP ]; then
+    $HAB_ROOT_PATH/bin/hab sup run > /hab/sup/default/log.out &
+    echo "** The Habitat Supervisor has been started in the background."
+    echo "** Use 'hab sup start' and 'hab sup stop' to start and stop services."
+    echo "** Use the 'slog' command to stream the supervisor log."
+    echo "** Adding '\\\NO_BG_SUP=1' to your .studiorc file will disable the background supervisor."
+    echo ""
+  else
+    echo "** \\\$NO_BG_SUP was set. The Habitat Supervisor is not running."
+    echo ""
+  fi
+}
 
 # Add command line completion
 source <(hab cli completers --shell bash)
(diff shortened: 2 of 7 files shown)


thesentinels:

Thank you for contributing!

mwrock:
This is all working in a linux studio. I want to double check docker and add documentation before merging.

mwrock:
This fixes #2131

components/studio/bin/hab-studio.sh
```diff
@@ -679,11 +679,25 @@ enter_studio() {
...
+  $bb chroot "$HAB_STUDIO_ROOT" \
     $studio_env_command -i $env $studio_enter_command $*
 }
 
+# **Internal** Run when an entered studio exits.
+stop_supervisor() {
+  lock_file="$HAB_STUDIO_ROOT/hab/sup/default/LOCK"
+
+  if [ -f $lock_file ]; then
+    $bb kill -9 $($bb cat $lock_file)
```
reset: Should we try to let the supervisor gracefully exit instead of forcing it with `-9` immediately?

components/studio/libexec/hab-studio-type-default.sh
```diff
@@ -84,6 +86,12 @@ exec $HAB_ROOT_PATH/bin/hab pkg exec core/hab-plan-build hab-plan-build \$*
 EOF
   $bb chmod $v 755 $HAB_STUDIO_ROOT$HAB_ROOT_PATH/bin/build
 
+  $bb cat <<TAIL_SUP > $HAB_STUDIO_ROOT$HAB_ROOT_PATH/bin/tail_sup
```
reset: What about just simply calling it `log` or `logs` like how we have the `build` command built in?

reset: By the way, holy shit was this a good idea

fnichol:
I still need to try this out, but I want to avoid us running software by default inside a Studio. This would incur more software being downloaded and installed on each new Studio instance and isn't relevant for post-build failures (when a `hab pkg build` fails and you want to `hab studio enter` to trace an error).

Could we add an explicit command inside the Studio to do a similar thing so the user knows a Supervisor is being executed? I don't like the idea of other active processes running in the build environment when a package build is being executed. At least not without asking for it. Otherwise maybe use an opt-in flag to do the same thing on a `hab studio enter`? I'd still much rather see a `run-sup` or equivalent command inside the Studio though.

components/studio/libexec/hab-studio-type-default.sh
```diff
@@ -6,6 +6,7 @@ studio_build_environment=
 studio_build_command="record \${1:-} $HAB_ROOT_PATH/bin/build"
 studio_run_environment=
 studio_run_command="$HAB_ROOT_PATH/bin/hab pkg exec core/hab-backline bash -l"
+studio_supervisor_start_command="$HAB_ROOT_PATH/bin/hab sup run > /hab/sup/log.out &"
```
reset: You should place this in `/hab/sup/default/out.log` - the supervisor that we're starting is "named" default. If somebody starts a separate supervisor they might want to consider outputting that to `/hab/sup/<name>/out.log`

reset:
@mwrock super quick and clean work man!

gif-keyboard-6463566812690874366

components/studio/bin/hab-studio.sh
mwrock: fur sure. reflecting on this, I have no idea why I used `-9`

components/studio/libexec/hab-studio-type-default.sh
mwrock: To me `log<s>` feels vague. Its unclear if it is build log or supervisor logs. In a studio, my assumption would be build logs.

mwrock:
I can see your point @fnichol . I see this as being helpful to the mac development workflow where a studio is the only place you can run a supervisor. Honestly I think that if we ask the user t ostart the supervisor with a command, we might as well just guide them to run `hab sup run`.

Thinking about this more, I wonder if it makes sense to totally rethink the api around starting services. Having a `hab start` command that does 2 things based on the state of the supervisor is kind of weird. Its also weird having `hab start` and `hab load` doing nearly the same thing if the supervisor is running. I know this might sound extreme but what if we just eliminate `hab start` or deprecate it. I'm just gonna stop myself here but worth thinking about.

components/studio/libexec/hab-studio-type-default.sh
smith: I think this is fine for now but somebody should open an issue to replace this with `hab sup logs -f`.

Since this is for studio use, I'm not concerned about the growth of this log file, but if habitat starts producing logs anywhere but stdout/stderr or the system journal, it has a responsibility to rotate them.

mwrock: Yeah I think that ideally we would have a `hab sup start -d` and the `-d` would indicate daemonized and it would emit logs in a sane way to a known location with rotation. Then `hab sup log` would consume those logs. I'm tempted just to do that here, but should just leave it for a separate pr.

reset: This can't become a command to `hab-sup` because we don't actually log to file by convention. We're just redirecting output here in the studio to provide this particular functionality - so it would make sense that this stay a command of the studio

reset:
@fnichol we don't need to run the supervisor if the studio is initiated by `hab pkg build`, but I think it definitely makes sense when using an interactive studio with `hab studio enter`.

The most common concern that we've heard after releasing the multi-tenant supervisor is the confusion around `hab start` having what is viewed as a "dual purpose". I think we actually got this UX completely correct, but we need to make small changes to our docs and studio to have it feel natural to people. In most cases, you will be using `hab start` and `hab stop` to do one of two things:

# Start/Stop a transient service in an interactive studio so you can test your package or use `--config-from` to test iterations on your hooks, etc
# Modify a permanent service's desired state in a production environment to start or stop

The studio would be the place that we would push people to would accomplish #1. We can additionally ask somebody to run `hab sup run > /path/to/log &` and then tail it, but the goal of this PR was to automate that and abstract away tailing logs to a single, easy to remember command, just like `build` from within the studio

reset:
Per the offline discussion with @mwrock, @fnichol, and myself, we're gonna make a few changes before merging this in

* Add an environment variable (perhaps `BG_SUP=false` or something along these lines?) which can be set in a `.studiorc` to disable automatic starting of a backgrounded supervisor in an interactive studio
* Ensure that we don't automatically start a backgrounded supervisor in a non-interactive studio (I think this might already be done?)
* Add some output when entering the interactive studio stating if the backgrounded service is started or stopped. If it's stopped, mention the env var that disabled it. If it's started, we should introduce the `slog` command and tell people how to `start`/`stop` services, if there's room mention `--config-from`, and maybe link to some docs if that's helpful
* Rename the `tail_sup` command to `slog` (short for supervisor log)

jonsmorrow:
This is awesome and I dig the pre merge changes.

mwrock:
```
vagrant@vagrant:/vagrant/components/studio$ hab studio enter
   hab-studio: Creating Studio at /hab/studios/vagrant--components--studio (default)
   hab-studio: Importing core secret origin key
★ Imported secret origin key core-20160810182414.
   hab-studio: Entering Studio at /hab/studios/vagrant--components--studio (default)
   hab-studio: Exported: HAB_ORIGIN=core

** The Habitat Supervisor has been started in the background.
** Use 'hab sup load' and 'hab sup unload' to load and unload services.
** Use the 'slog' command to stream the supervisor log.
** Adding '$NO_BG_SUP=1' to your .studiorc file will disable the background supervisor.

[1][default:/src:0]# slog
hab-sup(MR): Butterfly Member ID 1c9c5d7601274140b318fc8d1d8946d9
hab-sup(MR): Starting butterfly on 0.0.0.0:9638
connecting to tcp://localhost:10001
connecting to tcp://localhost:10011
connecting to tcp://localhost:10021
♥
[2][default:/src:130]# exit
logout
vagrant@vagrant:/vagrant/components/studio$
```

Is the heart an accident?...I don't think so.

smith:
> `** Use 'hab sup load' and 'hab sup unload' to load and unload services.`

Should we prefer `start` and `stop` in a studio, since they allow `--config-from` and you're more likely going to want transient services here?

> `** Adding '$NO_BG_SUP=1' to your .studiorc file will disable the background supervisor.`

No `$` here.

mwrock:
thanks @smith fixed both comments

reset:
gif-keyboard-7821617314214773675

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/224489188).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/224489188).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/224489188).

PR #2164, opened 2017-04-25
updates the ubuntu setup docs with workaround for libreadline 7 error

nellshamrell:
Signed-off-by: Nell Shamrell nellshamrell@gmail.com

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -35,6 +35,8 @@ the latest version of rustfmt. An easy way to install it (assuming you have
 Rust installed as above), is to run `cargo install rustfmt` and adding
 `$HOME/.cargo/bin` to your `PATH`.
 
+**Note2:** While this Docker container will work well for getting started with Habitat development, you may want to consider using a VM as you start compiling Habitat components.  To do this, create a VM with your preferred flavor of Linux and follow the appropriate instructions for that flavor below.
+
 
 ## Mac OS X for Native Development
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

BUILDING.md
```diff
@@ -94,20 +96,45 @@ Then, run the system preparation scripts and try to compile the project:
 cp components/hab/install.sh /tmp/
 sh support/linux/install_dev_0_ubuntu_latest.sh
 sh support/linux/install_dev_9_linux.sh
-. ~/.profile
+```
+
+When you run the install_dev_9_linux.sh script, you may see this error.
```
rhass: I think we should fix the `install_dev_0_ubuntu_latest.sh` script ather than telling people to manually build and install libreadline7.

The cause of the problem is self-inflicted, since we are installing packages from Ubuntu 17.04 into 16.10. 
https://github.com/habitat-sh/habitat/blob/master/support/linux/install_dev_0_ubuntu_latest.sh#L32-L38

The script should check if the version of libreadline we want is available via apt, and if not then we canmanually grab and install the packages and deps we want.

nellshamrell: Would it make sense to merge this in for now, then open an issue saying that the script needs to be fixed?  While I would be happy to take a look at fixing the script, with ChefConf coming up so soon I think my time would be better spent on demo work at least until after ChefConf.  This would also be a great first contribution for a new contributor.

PR #2170, opened 2017-04-26
implement supervisor status command

mwrock:
This adds a `status command to the supervisor. It takes an optional package ident argument. If this argument is not given then the command reports status for all services running in the supervisor.

Further yhis fixes a bug in the service data persistence where service state was not being persisted when a service's state changes. In production, the `/services` monitoring endpoint only reports the initial service state when the supervisor first runs and this would typically report that the services are `Down` with a `Null` PID. Now service states are checked for a change on each supervisor tick and persisted to disk on change.

Note that the `state_entered` supervisor field is now a `Timespec` rather than a `SteadyTime`. This is because `SteadyTime` does not lend itself well  to serialization/deserialization. Its internal state will differ based on platform whereas `Timespec` internal data should be uniform.

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -44,9 +44,10 @@ use sup::error::{Error, Result};
 use sup::feat;
 use sup::command;
 use sup::http_gateway;
-use sup::manager::{Manager, ManagerConfig};
+use sup::manager::{Manager, ManagerConfig, ServiceStatus};
 use sup::manager::service::{DesiredState, ServiceBind, Topology, UpdateStrategy};
 use sup::manager::service::{ServiceSpec, StartStyle};
+use sup::supervisor::ProcessState;
 
 /// Our output key
 static LOGKEY: &'static str = "MN";
@@ -73,6 +74,7 @@ fn start() -> Result<()> {
         ("run", Some(m)) => sub_run(m),
         ("sh", Some(m)) => sub_sh(m),
         ("start", Some(m)) => sub_start(m),
+        ("status", Some(m)) => sub_status(m),
         ("stop", Some(m)) => sub_stop(m),
         ("unload", Some(m)) => sub_unload(m),
         _ => unreachable!(),
@@ -186,6 +188,14 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
                 "One or more service groups to bind to a configuration")
             (@arg CONFIG_DIR: --("config-from") +takes_value {dir_exists}
                 "Use package config from this path, rather than the package itself")
+        )
+        (@subcommand status =>
+            (about: "Query the status of Habitat services.")
+            (aliases: &["stat", "statu", "status"])
+            (@arg PKG_IDENT: +takes_value "A Habitat package identifier (ex: core/redis)")
+            (@arg NAME: --("override-name") +takes_value
+                "The name for the state directory if there is more than one Supervisor running \
+                [default: default]")
         )
         (@subcommand stop =>
             (about: "Stop a running Habitat service.")
@@ -312,6 +322,63 @@ fn sub_start(m: &ArgMatches) -> Result<()> {
     };
     try!(command::start::run(cfg, maybe_spec, maybe_local_artifact));
     Ok(())
+}
+
+fn sub_status(m: &ArgMatches) -> Result<()> {
+    if m.is_present("VERBOSE") {
+        sup::output::set_verbose(true);
+    }
+    if m.is_present("NO_COLOR") {
+        sup::output::set_no_color(true);
+    }
+    let cfg = mgrcfg_from_matches(m)?;
+    let mut services: Vec<ServiceStatus> = Vec::new();
+
+    if !Manager::is_running(&cfg)? {
+        outputln!("The supervisor is not running.");
+        std::process::exit(3);
+    }
+
+    match m.value_of("PKG_IDENT") {
+        Some(pkg) => {
+            match Manager::service_status(cfg, PackageIdent::from_str(pkg)?) {
+                Ok(status) => {
+                    services.push(status);
+                }
+                Err(_) => {
+                    outputln!("The {} service is not currently loaded.", pkg);
+                    std::process::exit(2);
+                }
+            }
+        }
+        None => services.append(&mut Manager::status(cfg)?),
+    }
+
+    for status in services {
+        match status.process.state {
+            ProcessState::Up => {
+                outputln!("The {} service entered the running state with PID: {} {} seconds ago.",
+                          status.package,
+                          status.process.pid.unwrap(),
+                          status.process.elapsed.num_seconds())
+            }
+            ProcessState::Down => {
+                outputln!("The {} service is not currently running.", status.package);
+            }
+            ProcessState::Start => {
+                outputln!("The {} service entered the starting state {} seconds ago.",
+                          status.package,
+                          status.process.elapsed.num_seconds());
+            }
+            ProcessState::Restart => {
+                outputln!("The {} service entered the restarting state {} seconds ago.",
+                          status.package,
+                          status.process.elapsed.num_seconds());
+            }
+        }
+    }
+
+    return Ok(());
 }
 
 fn sub_stop(m: &ArgMatches) -> Result<()> {
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -18,10 +18,12 @@ mod service_updater;
 mod spec_watcher;
 
 use std::collections::HashMap;
+use std::fmt;
 use std::fs::{self, File, OpenOptions};
-use std::io::{BufRead, BufReader, BufWriter, Read, Write};
+use std::io::{self, BufRead, BufReader, BufWriter, Read, Write};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
+use std::result;
 use std::thread;
 use std::sync::{Arc, RwLock};
 use std::sync::mpsc::channel;
@@ -38,10 +40,13 @@ use eventsrv_client::EventSrvClient;
 use hcore::crypto::{default_cache_key_path, SymKey};
 use hcore::fs::FS_ROOT_PATH;
 use hcore::service::ServiceGroup;
+use hcore::util::deserialize_using_from_str;
 use hcore::os::process;
+use hcore::package::{Identifiable, PackageIdent};
 use protobuf::Message;
+use serde;
 use serde_json;
-use time::{SteadyTime, Duration as TimeDuration};
+use time::{self, Timespec, Duration as TimeDuration};
 
 pub use manager::service::{Service, ServiceConfig, ServiceSpec, UpdateStrategy, Topology};
 use self::service::{DesiredState, StartStyle};
@@ -52,6 +57,7 @@ use config::GossipListenAddr;
 use census::CensusRing;
 use manager::signals::SignalEvent;
 use http_gateway;
+use supervisor::ProcessState;
 
 const MEMBER_ID_FILE: &'static str = "MEMBER_ID";
 const PROC_LOCK_FILE: &'static str = "LOCK";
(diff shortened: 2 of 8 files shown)


thesentinels:

Thank you for contributing!

mwrock:
Status for an individual service:
```
C:\dev\habitat\components\sup [status +1 ~2 -0 !]> ..\..\target\debug\hab-sup.exe status core/habitat-aspnet-sample
hab-sup(MN): The core/habitat-aspnet-sample/0.2.0/20170311090829 service entered the running state with PID: 15492 11 seconds ago.
```

Status for all services:
```
C:\dev\habitat\components\sup [status +1 ~2 -0 !]> ..\..\target\debug\hab-sup.exe status

hab-sup(MN): The core/habitat-aspnet-sample/0.2.0/20170311090829 service entered the running state with PID: 15492 83 seconds ago.
hab-sup(MN): The core/mysql/5.7.17/20170310233054 service entered the running state with PID: 4660 5 seconds ago.
```

mwrock:
Thanks a ton for this feedback @smith!
 Here is what I propose for exit codes:

`0` - single service running or `status` with no ident is passed and the supervisor is running.
`1` - single service `down`
`2` - single service `starting`
`3` - single service restarting
`4` - single service not loaded
`5` - supervisor not running

Thoughts?

mwrock:
Having a hard time getting a repro on your second point @smith:
```
C:\dev\habitat\components\sup [status +1 ~0 -0 !]> ..\..\target\debug\hab-sup.exe status core/mysql
hab-sup(MN): The core/mysql service is not currently loaded.
C:\dev\habitat\components\sup [status +1 ~0 -0 !]> ..\..\target\debug\hab-sup.exe load core/mysql
C:\dev\habitat\components\sup [status +1 ~0 -0 !]> ..\..\target\debug\hab-sup.exe status core/mysql
hab-sup(MN): The core/mysql/5.7.17/20170310233054 service is not currently running.
C:\dev\habitat\components\sup [status +1 ~0 -0 !]> ..\..\target\debug\hab-sup.exe status core/mysql
hab-sup(MN): The core/mysql/5.7.17/20170310233054 service entered the running state with PID: 5024 2 seconds ago.
```

One important not for testing is that you need to start the supervisor (via run or start) with the new binary. If you the supervisor is running from the prod binary, it will not refresh the statuses and the `status` command of the build binary may be stale.

mwrock:
ok I think I have a repro now. if i load it, stop it, kill the supervisor and then start the supervisor with a different service. The mysql service is stil loaded but stopped yet dead to the status command.

mwrock:
@smith I *think* I hit all of your feedback but would really appretiate it if you could run the latest in this PR and make sure I addressed the second point where it reported that the loaded service was not loaded.

mwrock:
Note that we used to only add services that had an initial `desired_state` of `Up` to the manager's service collection. This meant that out `/services` api did not report services that were loaded and stopped. We now load all services regardless of desired state. If the `desired_state` is `Down` we simply break out of that service's `tick` function early so it does nothing.

components/sup/src/manager/mod.rs
```diff
@@ -156,6 +206,26 @@ impl Manager {
...
+
+        Err(sup_error!(Error::ServiceNotLoaded(ident)))
+    }
+
+    pub fn status(cfg: ManagerConfig) -> Result<Vec<ServiceStatus>> {
+        let state_path = Self::state_path_from(&cfg);
+        let fs_cfg = FsCfg::new(state_path);
+
+        let dat = File::open(&fs_cfg.services_data_path)?;
+        Ok(serde_json::from_reader(&dat)?)
```
reset: Directly converting one error to our own type is a pattern from the earlier days of our codebase but it's something that we've been trying to get away from. It's more appropriate to create a specific error and pass the error you would be converting here instead directly to it as a member. For example, you'd maybe create the error `SvcStatFileParse(serde_json::Error)` which explicitly says that there was an error in parsing the svc stat file. Since you're holding onto the `serde_json::Error` in the error struct you can display that in anyway you'd like, including just writing it into your error string at the end

components/sup/src/manager/service/mod.rs
```diff
@@ -287,6 +289,9 @@ impl Service {
     }
 
     pub fn tick(&mut self, census_ring: &CensusRing) -> bool {
+        if self.desired_state == DesiredState::Down {
```
reset: This shouldn't be a necessary. Services which are in the `Down` state shouldn't be loaded so their tick function shouldn't be called

components/sup/src/main.rs
```diff
@@ -314,6 +324,73 @@ fn sub_start(m: &ArgMatches) -> Result<()> {
...
+                    outputln!("The {} service entered the restarting state {} seconds ago.",
+                              status.package,
+                              status.supervisor.elapsed.num_seconds());
+                    if query_one {
+                        std::process::exit(4);
+                    }
+                }
+            }
+        }
+    } else {
```
reset: Instead of an `if`/`else` block I'd recommend using a _function short-circuit_ where you return out of the function early.

```rust
if !Manager::is_running(&cfg)? {
    std::process::exit(6);
}

// ... rest of code
```

This would be a design pattern that we have in our coding standards had anyone ever gotten around to writing them yet ;)

components/sup/src/manager/service/mod.rs
```diff
@@ -92,7 +93,7 @@ pub struct Service {
     last_health_check: Instant,
     #[serde(skip_serializing)]
     manager_fs_cfg: Arc<manager::FsCfg>,
-    supervisor: Supervisor,
+    pub supervisor: Supervisor,
```
reset: We should keep this private and expose some public member functions on the `Service` struct to expose the API that you need. The `Supervisor` module should be considered internal

components/sup/src/supervisor.rs
```diff
@@ -104,7 +104,7 @@ pub struct Supervisor {
     pub package: Arc<RwLock<PackageInstall>>,
     pub preamble: String,
     pub state: ProcessState,
-    pub state_entered: SteadyTime,
+    pub state_entered: Timespec,
```
reset: From my understanding you had to make this swap because of serialization issues. It's been a while since we've had to revisit this module so this hasn't come up before, but the crate that we're using for time is actually old and deprecated which is why you don't have the support you need in it for serde.

[chrono](https://github.com/chronotope/chrono) replaced the [time](https://github.com/rust-lang-deprecated/time) crate and it does have support for serialization using serde. You might want to consider using that here instead or we could save that for a future refactor!

components/sup/src/manager/mod.rs
```diff
@@ -424,6 +495,10 @@ impl Manager {
                                     &self.butterfly.service_file_store);
             service_rumor_offset = 0;
 
+            if self.check_for_changed_services() {
```
reset: Good catch!

components/sup/src/manager/mod.rs
```diff
@@ -650,12 +741,7 @@ impl Manager {
     fn start_initial_services_from_watcher(&mut self) -> Result<()> {
         for service_event in self.watcher.initial_events()? {
             match service_event {
-                SpecWatcherEvent::AddService(spec) => {
```
reset: I don't think we should load services which are supposed to be in the `Down` state, I'd keep this block as is. The code makes assumptions that a "loaded" service has an active rumor which is shared along with participating in the tick loop. I think this complicates that logic by needing to short circuit things with a state check.

It's worth noting that right now Adam is working on some logic which will allow us to tombstone service rumors so when you stop a service it actually let's peers know that it went away

components/sup/src/main.rs
```diff
@@ -187,6 +189,14 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
             (@arg CONFIG_DIR: --("config-from") +takes_value {dir_exists}
                 "Use package config from this path, rather than the package itself")
         )
+        (@subcommand status =>
+            (about: "Query the status of a running Habitat service.")
```
reset: Technically you're not just asking about _running_ services ;)

components/sup/src/manager/mod.rs
reset: You may want to read the service specs directory to get a view of what is stopped vs what is running. As @smith pointed out, things can be "loaded" but not be running

mwrock: This was mainly a choice of where to put the complexity. If the `services` vec only has services with a `DesiredState` of `Up` then we need to change the persistence logic to do more than just iterate the `services` vec. If the "downed"services are not in the services vec, they won't get hydrated to the serialized services. We could aggregate and merge the down and up services into a new data structure that we serialize to `/services` but that seems even more complex.

components/sup/src/manager/mod.rs
mwrock: Is this what
```
impl From<serde_json::Error> for SupError {
 +    fn from(err: serde_json::Error) -> SupError {
 +        sup_error!(Error::SerdeJsonError(err))
 +    }
 +}
```
is doing?

mwrock: Oh I think i see...

You are saying bail on that `From<serde_json::Error>` implementation and actually match here and explicitly `Err` my `sup_error!(Error::SerdeJsonError(err))`

That does sound better

components/sup/src/supervisor.rs
mwrock: Yeah I stumbled upon chrono in the serde issues and decided to skip the dependency since I would have to do some custom deserialization anyway to represent the time of the state change as elapsed seconds.

smith:
This is much closer, but a couple things:

If I stop a transient service, it seems like it still stays in the list. Can you reproduce this? If there's no spec file it shouldn't show up:

```bash
[27][default:/src:130]# hab pkg exec smith/hab-sup hab-sup status
hab-sup(MN): The core/postgresql/9.6.1/20170424032128 service is not currently running.
hab-sup(MN): The core/redis/3.2.4/20170215222111 service is not currently running.
[28][default:/src:0]# ls /hab/sup/default/specs/
postgresql.spec
```

Also, I'm wondering about the correct exit code behavior. The first thing I want to do with this is:

* Is service x loaded? (I don't care whether or not it's running)
  * If not, `hab sup load x`
  * If so, `hab sup start x`

basically, I want to do `hab sup load --force` but not force a restart.

In this case I would want `hab sup status x`, when x is persistent but down, to exit 0, but in the case where I just wanted to know if it was up or down, I would probably expect a non-zero exit code.

This confusion makes me think either there's another missing feature (`hab sup load --yes-I-want-it-loaded-but-if-it's-already-loaded-then-thats-cool-never-mind`?) or some other simplification that could be done. Thoughts?

mwrock:
@smith I'm just pushed a new iteration with some refactorings based on comments from @reset. They had the fortunate byproduct of also fixing the issue with unloaded services showing up as not running in the status.

Thinking of your exit code comments, I do think that persistent services not running should emit a `0` exit code since that is not an "error" case. If it mattered to you that the service was running you could grep the output. I think simply changing the error of loading a loaded service to a warning would be best. We can open a separate issue for that.

mwrock:
Hmmm. Now I'm feeling torn over this exit code business. If a stopped transient service returns 0 because that is an acceptable state, then we should also return 0 got services in the "starting" and "reloading" states. You could of coarse grep on any of these states reliably, but status codes are nice. Here is what I think we should do: remove the exit codes for "acceptable" states and if that proves inconvenient, we later add a `-states` arg that will provide explicit exit codes for each possible state.

reset:
@mwrock when you stop a transient service it unloads from the supervisor completely - the service is stopped and the service spec should be deleted. I don't think you can ever see a transiet service as "stopped"

mwrock:
@reset yeah @smith was saying that he was seeing that in the previous state of this PR but I have not been able to repro and it would certainly be a bug

reset:
tenor-67443071

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/226929712).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/226929712).

Issue #1195, opened 2016-08-26
private hab-depot - issues uploading / downloading packages


throwanexception:
Using hab 0.9.0/20160816000307 and BldrDepot 0.9.0/20160815233945 on centos-7.2 (specifically the bento/centos-7.2 vagrant image).

I have no HAB environment variables defined.

I've tried this workflow from two VMs, one being the depot and another acting as the client, and together on the same VM to isolate any connectivity issues.

Built a package for tomcat-native:

```
pkg_origin=company
pkg_name=tomcat-native
pkg_version=1.2.8
pkg_maintainer="redacted"
pkg_license=('Apache 2.0')
pkg_source=http://www-us.apache.org/dist/tomcat/tomcat-connectors/native/${pkg_version}/source/${pkg_name}-${pkg_version}-src.tar.gz
pkg_shasum=408ece0b027c8967b3aa85533c5fca642827e235b1857d28df918a4eab861d30
pkg_build_deps=(core/gcc core/make core/diffutils core/coreutils core/file company/jdk7 core/ant core/apr core/openssl)
pkg_deps=(core/glibc core/apr)
pkg_lib_dirs=(lib)
pkg_include_dirs=(include)

do_build() {
    pushd ${HAB_CACHE_SRC_PATH}/${pkg_name}-${pkg_version}-src/native
    ./configure --prefix=${pkg_prefix} --with-apr=$(pkg_path_for core/apr) --with-java-home=$(pkg_path_for company/jdk7) --with-ssl=$(pkg_path_for core/openssl)
    make
}

do_install() {
    pushd ${HAB_CACHE_SRC_PATH}/${pkg_name}-${pkg_version}-src/native
    make install
}
```

It'll upload to the private hab-depot without any issues. When I attempt to install as part of a dependency chain ->

RUST_LOG=debug RUST_BACKTRACE=1 hab pkg install company/tomcat-example --url http://192.168.247.134:9632/v1

```
↓ Downloading company/tomcat-native/1.2.8/20160826151709
[...]

PR #2178, opened 2017-04-27
Remove implication that hab-depot currently works.

dmp1ce:
Signed-off-by: David Parrish <daveparrish@tutanota.com>

### DIFF
--- a/www/source/docs/share-packages-overview.html.md
+++ b/www/source/docs/share-packages-overview.html.md
@@ -14,7 +14,7 @@ You interact with a depot by performing the following operations:
 * Uploading the keys for that origin.
 * Uploading one or more packages into that origin.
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

dmp1ce:
Maybe the wording needs a little work, but I think it is important to let users know that they cannot setup a private depot currently.  #1195 has a little more information.

mwrock:
I definitely think its a great idea to more explicitly tell people that running a private depot is not supported. It is "possible" but is not a matter of running `core/ha-depot` and what works today may not work tomorrow until it stabilizes.

mwrock:
Thanks a bunch for inspiring discussion. We are closing this in favor of #2285

PR #2187, opened 2017-04-27
Remove gdb install step from ubuntu_latest script

cnunciato:
Also updates the doc. Verified on Ubuntu 16.04 and 16.10 VMs.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -96,45 +96,11 @@ Then, run the system preparation scripts and try to compile the project:
 cp components/hab/install.sh /tmp/
 sh support/linux/install_dev_0_ubuntu_latest.sh
 sh support/linux/install_dev_9_linux.sh
-```
-
-When you run the install_dev_9_linux.sh script, you may see this error.
-
-```
-The following packages have unmet dependencies:
- gdb : Depends: libreadline7 (>= 6.0) but it is not installable
-E: Unmet dependencies. Try 'apt-get -f install' with no packages (or specify a solution).
-```
-
-Unfortunately, running apt-get -f install does not currently fix the issue. To get around this,
-
-cd ~
-wget https://launchpad.net/ubuntu/+archive/primary/+files/readline_7.0.orig.tar.gz
-tar xzf readline_7.0.orig.tar.gz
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

raskchanky:
Why remove this?

cnunciato:
There was some discussion in #2164 about this (I was running into the same issue that prompted that change), and it looks like at least on 16.04 and 16.10, it doesn't seem to be necessary, so I figured it might be good to consider removing it. (I haven't tested on 17.04, though.) I could totally be missing the point of it, though. :)

raskchanky:
Ah, I wasn't aware of the context there.  Maybe this is the right move?  I don't think I've hit that error before myself.

cnunciato:
I think I probably missed the point, here -- I was assuming the items in this script were *necessary* libraries, so when I noticed this one wasn't necessary, I figured I might as well remove it. I failed to consider that gdb might actually provide something useful to people other than myself. ;) 

So maybe the thing to do is figure out how to install the right version. @rhass @nellshamrell does that sound right?

fnichol:
Nice! Well, that version of gdb wasn't strictly required but is, in fact an awesome way to debug Rust code (you need version 7.12 or higher). At the time I updated those docs, the latest Ubuntu didn't have that exact version so I hacked it in with the intention to drop it as soon as a newer Ubuntu release shipped with that gdb version.

As a project, our suggestion to developers is to use the latest version of Ubuntu anyway if possible (assuming that's your Linux distribution) so this works for me!

raskchanky:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/227656209).

cnunciato:
Nice! Thanks for the closer look, @raskchanky and @fnichol.

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/227656209).

Issue #1816, opened 2017-02-23
pkg_exports values are not gossiped when updated on  hab config apply


smacfarlane:
I have an nginx service that uses `pkg_exports` for the port it listens on, and an haproxy service that binds to the nginx service group and consumes the port value in its configuration.

When adding/removing nginx services, the haproxy configuration updates as expected.  However, when running `hab config apply nginx.default ...`,  the nginx services in the group restart on their new port, but haproxy doesn't update its configuration until nginx is stopped/started. 

Looking at the `/butterfly` endpoint after the `config apply`, it still contains the original value.

I have the following repo with a test case to reproduce:  https://github.com/smacfarlane/hab-bind-example

mwrock:
Should be fixed by #2190

PR #2190, opened 2017-04-28
remove service rumor offset to avoid dropping rumors

mwrock:
This is a potential fix for #1816. When a service package is updated (in this case from a `hab config apply`), we add a service rumor to the service rumor store. However this rumor does not update the census group members and as a result the binds of one service that would have been altered by another's change are never processed.

Its really a mystery what this offset is intended to accomplish. Its clearly there for a reason but perhaps it is legacy?

There is a good chance that I am not seeing its true purpose which may be important so I would really like @adamhjk and/or @reset to weigh in. As this PR stands, I now see config updates refresh the binds of other services.

### DIFF
--- a/components/sup/src/census.rs
+++ b/components/sup/src/census.rs
@@ -65,7 +65,6 @@ impl CensusRing {
     }
 
     pub fn update_from_rumors(&mut self,
-                              service_rumor_offset: usize,
                               service_rumors: &RumorStore<ServiceRumor>,
                               election_rumors: &RumorStore<ElectionRumor>,
                               election_update_rumors: &RumorStore<ElectionUpdateRumor>,
@@ -73,7 +72,7 @@ impl CensusRing {
                               service_config_rumors: &RumorStore<ServiceConfigRumor>,
                               service_file_rumors: &RumorStore<ServiceFileRumor>) {
         self.changed = false;
-        self.update_from_service_store(service_rumor_offset, service_rumors);
+        self.update_from_service_store(service_rumors);
         self.update_from_election_store(election_rumors);
         self.update_from_election_update_store(election_update_rumors);
         self.update_from_member_list(member_list);
@@ -90,9 +89,7 @@ impl CensusRing {
     }
 
     fn update_from_service_store(&mut self,
-                                 service_rumor_offset: usize,
                                  service_rumors: &RumorStore<ServiceRumor>) {
-        self.last_service_counter += service_rumor_offset;
         if service_rumors.get_update_counter() <= self.last_service_counter {
             return;
         }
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

bodymindarts:
@mwrock I think the purpose is to prevent a `Service` to get passed in a `CensusRing` (on the next tick) where `census_ring.changed == true` but the only thing that has changed is that it now is updated to contain the `ServiceRumor` itself triggered sending. The rumor that it sent is obviously already incorporated into its config, so I think this was either an optimization or a mechanism to prevent `reconfigure` to get triggered when nothing really changed.

I'm not sure wether or not this mechanism is still required. I think `Service` uses other tests (like hashing the config files) to discover if something has changed or not so its probably not needed. @reset will have to comment on how important the optimization aspect is.

mwrock:
Thanks @bodymindarts! I figured it was some kind of optimization or safeguard for double rumor handling. Interestingly in this case it shows that when an update causes a tick to return true we do not want to suppress that rumor because that will prevent other services from "hearing" it that need to know.

mwrock:
Took another look at this today and from what I can tell there are no adverse affects of updating service rumors twice or more. Each update appears to be idempotent. I also do not see how a rumor could ever update the member population if the offset is used.

reset:
@mwrock I think there is a chance for a race condition in here. In `update_from_service_store()` the call to `service_rumors.get_update_counter()` reads from an atomic and it's read to check if we need to update, then we update, and then it's read again at the end of the function to set the value of `last_service_counter`. If we just obtain a lock on `service_rumors` before reading the counter to a local variable we can avoid the potential race condition.

It's worth noting that this is *very* unlikely to occur, but this is exactly why the offset was added

mwrock:
hey @reset, what strikes me here is that all of the rumor update functions has that same race potential. In order for the race to be truly problematic the manager loop would have to be called from multiple threads which is the only entry point to `update_from_rumors` that I can tell and thats managed by a single thread.

The only thing from what I can tell that makes the service rumors more "special" than the others is that the manager cals `gossip_latest_service_rumor` and inserts a rumor to the service store but I don't see how that would cause multiple threads to enter `update_from_service_store` concurrently.

mwrock:
that all said, a lock is by no means a bad idea but we should apply a lock to each of the rumor update functions.

reset:
@mwrock you are completely correct :)

reset:
@thesentinels approve

tenor-42943700

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/228217571).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/228217571).

bodymindarts:
@mwrock thanks for this one :-)

PR #2226, opened 2017-05-04
Refactor ServiceConfig into RenderContext in supervisor

reset:
Refactor ServiceConfig into RenderContext in Habitat Supervisor

The spirit of this commit is to improve reliability when compiling templated
plan artifacts such as configuration files and hooks. We accomplish this by
adding a `RenderContext` module which is used in the templating module to
bind data to templates. This should improve overall readability of our code
and also ensure that we have a typed structure when deserializing data from
handlebars in the `templating::helpers` module. We leverage a zero-copy
deserialization implementation in `RenderContext` to ensure that we don't
duplicate data for each helper in a template.

* Add `templating::context` module and `RenderContext` struct. This is what
  is sent to our templating engine to be bound to all templates.
* Add `manager::sys` module and `Sys` struct. This represents the system as
  a whole and is eventually serialized in the RenderContext as `sys`
  * Move http_listen to `Sys` struct from `Manager` struct
  * Move gossip_listen to `Sys` struct `Manager` struct
* Add `Arc<Sys>` to `Service` struct. This holds a read only reference to
  the `Sys` struct of the `Manager` which loaded it. This ensures that we
  only allocate one `Sys` struct which doesn't change unless the Supervisor
  is restarted
* Rewrote `manager::service::config` module and added `Cfg` struct to it.
  This contains all configuration logic for the service and is passed
  to the `RenderContext` in the same way as the `Sys` struct
* Add `manager::service::package` module and `Pkg` struct to it. This is
  similar to both the `Sys` and `Cfg` modules in it's relation with
  `RenderContext`. It contains all information about the package the service
  is running
* Moved `supervisor` module to `manager::service::supervisor`
* Fix serialization/deserialization issues with SysInfo struct
* Fix API for `crypto::hash` - only return a `Result` if there is actually
  an error case
* Move command execution functions from the `util` module to
  `manager::service::exec`. The `create_command` function from that module has
  been renamed to `run_cmd` and now accepts a reference to a `Pkg` struct
  which will ensure that all commands are executed within the context of the
  service's package
* Remove `gossip.toml` and `config.toml` cache files from service directory.
  We no longer need to write these out since we now persist and reload our
  entire rumor store on supervisor boot.
* Improve error when loading a bad package from a package install
* Upgrade completely to serde 1.0+
  * Temporarily switch to git source for libsodium dependency
* Replace "Butterfly" in supervisor output with "Supervisor" or another more
  appropriate description which hides our implementation details
* Add function and module documentation


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -68,22 +71,23 @@ lazy_static! {
...
-    pub config: ServiceConfig,
     pub depot_url: String,
     pub spec_file: PathBuf,
     pub spec_ident: PackageIdent,
     pub start_style: StartStyle,
     pub topology: Topology,
     pub update_strategy: UpdateStrategy,
+    pub cfg: Cfg,
+    pub pkg: Pkg,
+    pub sys: Arc<Sys>,
```
bodymindarts: Why is Sys behind an Arc? Shouldn't it be immutable after start up? Or is this just to save memory instead of cloning?

components/sup/src/manager/service/mod.rs
```diff
@@ -98,45 +102,32 @@ pub struct Service {
...
-                                         &runtime_cfg,
-                                         config_root,
-                                         spec.binds.clone(),
-                                         &gossip_listen,
-                                         &http_listen)?;
-        let hook_template_path = svc_cfg.config_root.join("hooks");
-        let hooks_path = fs::svc_hooks_path(service_group.service());
-        let locked_package = Arc::new(RwLock::new(package));
+        let service_group = ServiceGroup::new(&pkg.name, spec.group, organization)?;
+        let config_root = Self::config_root(&pkg, spec.config_from.as_ref().map(|p| &**p));
```
bodymindarts: I really dislike the `&**p` type of code. I have seen it a lot in this code base where the `Deref` trait is invoked via `*` but this makes it impossible to know as a reader what type is actually falling out and it is very implicit. Could we find a more explicit way to do these type of conversions. Explicit in the sense that it is more readable. Perhaps the `From` trait would help but I'm not sure.
Just a method like `to_X` would already be helpful.

components/sup/src/manager/service/mod.rs
```diff
@@ -279,64 +264,67 @@ impl Service {
...
+
+        let svc_updated = self.update_templates(census_ring);
+        if self.update_service_files(census_ring) {
+            self.file_updated();
+        }
 
         match self.topology {
             Topology::Standalone => {
-                self.execute_hooks();
+                self.execute_hooks(census_ring);
```
bodymindarts: Not happy that the census_ring needs to be passed in here.
Previously the tick method had 2 explicit phases, rendering files / updating config and executing the hooks. Now this is complected again.

components/sup/src/manager/service/mod.rs
```diff
@@ -279,64 +264,67 @@ impl Service {
...
             Topology::Leader => {
                 let census_group = census_ring
                     .census_group_for(&self.service_group)
                     .expect("Service Group's census entry missing from list!");
-                let current_election_status = &census_group.election_status;
-                match census_group.election_status {
+                match census_group.topology_election_status {
                     ElectionStatus::None => {
-                        if self.last_election_status != *current_election_status {
+                        if self.last_election_status != census_group.topology_election_status {
```
bodymindarts: nice improvement

components/sup/src/manager/service/mod.rs
```diff
@@ -447,82 +407,44 @@ impl Service {
...
+    fn initialize(&mut self, census_ring: &CensusRing) {
         if self.initialized {
             return;
         }
         outputln!(preamble self.service_group, "Initializing");
-        self.hooks.compile(&self.service_group, &self.config);
-        if let Some(err) = self.copy_run().err() {
-            outputln!(preamble self.service_group, "Failed to copy run hook: {}", err);
+        {
+            let ctx = self.render_context(&census_ring);
```
bodymindarts: If this can be avoided and run in the `update_configuration` method I would find that much better.
Perhaps there could be code like:
```
if !self.initialized {
  ... always render everything
   return
}
```
in the `update_configuration` method.

components/sup/src/manager/service/mod.rs
```diff
@@ -641,9 +548,9 @@ impl Service {
         Ok(())
     }
 
-    fn execute_hooks(&mut self) {
+    fn execute_hooks(&mut self, census_ring: &CensusRing) {
         if !self.initialized {
-            self.initialize();
+            self.initialize(census_ring);
```
bodymindarts: Already commented on this.

components/sup/src/templating/context.rs
```diff
@@ -0,0 +1,149 @@
...
+            if let Some(group) = census.census_group_for(&bind.service_group) {
+                map.insert(bind.name.to_string(), BindGroup::new(group));
+            }
+        }
+        Binds(map)
+    }
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct BindGroup;
```
bodymindarts: Empty? I guess this is WIP.

components/sup/src/templating/mod.rs
```diff
@@ -13,20 +13,24 @@
...
+use std::fmt;
 use std::ops::{Deref, DerefMut};
 use handlebars::Handlebars;
+pub use self::context::RenderContext;
 
-pub struct Template(Handlebars);
+pub struct Registry(Handlebars);
 
-impl Template {
+impl Registry {
```
bodymindarts: Already commented on this from the POV of the client code in `hooks.rs`

components/sup/src/manager/service/mod.rs
reset: Ya this is just to save on memory instead of cloning. An `Arc` is basically a read-only thread safe pointer. This is totally immutable ;)

components/sup/src/manager/service/mod.rs
reset: This is the best way I've found to convert from an `Option<PathBuf>` to an `Option<&Path>`. The reason is that you need to unwrap the `Option` but there is no way to say unwrap me as `&Path`, only as `&PathBuf`.

With that said, I could just change the function signature that we're calling to accept `Option<&PathBuf`>.  Which would make the calling code a bit cleaner. It's not like this is an API we call very often! I'll make that change

What I *really want to have* is `Option<T> where T: AsRef<Path>` on the function we're calling so we can just pass our `Option<PathBuf>`. I'm sure theres a way to say it in Rust but I just haven't found it yet.

bodymindarts: Aha. For me this was not evident. If there is a cleaner way to express this I'm for it.

components/sup/src/manager/service/hooks.rs
```diff
@@ -144,7 +141,7 @@ pub trait Hook: fmt::Debug + Sized {
 
     fn path(&self) -> &Path;
 
-    fn template(&self) -> &Template;
+    fn registry(&self) -> &Registry;
```
reset: Yeah I was a 2/10 on this name but it was a hell of a lot better than Template which is what is actually being registered to this thing. I was also thinking `Renderer` so maybe `Template` was a bad name but `TemplateRenderer` is spot on. I'll make that change

components/sup/src/manager/mod.rs
```diff
@@ -48,16 +49,16 @@ use serde;
 use serde_json;
 use time::{self, Timespec, Duration as TimeDuration};
 
-pub use manager::service::{Service, ServiceConfig, ServiceSpec, UpdateStrategy, Topology};
-use self::service::{DesiredState, StartStyle};
+pub use self::service::{Service, ServiceSpec, UpdateStrategy, Topology};
```
reset: The way I've organized these are

* Imports from std
* Public imports from extern
* Imports from extern
* Public imports from our own crate
* Imports from our own crate

I chose this since the rest of the code is formatted (or should be... we really need to commit some rules to the wiki!)

* Imports
* Public Const
* Private Const
* Public Static
* Private Static
* Public Trait
* Private Trait
* Public Struct
* Private Struct
* Public Function
* Private Function

We've got a few other small things in there like recently organizing members of the struct into public/private sections and organization of functions within the struct itself.

If you think we can come up with something more readable then let's go ahead and do it

components/sup/src/manager/service/mod.rs
reset: This is actually how you're recommended to do things in Rust since you can't represent two members of a struct which hold a reference on each other without using runtime borrow checking. It's not something you often see in OO langauges but you do see this pattern in functional

components/sup/src/manager/mod.rs
bodymindarts: Okay this isn't documented nor adhered to consistently. I'm fine leaving it as is but we need to start a style guide to communicate this.

reset: Yeah we really do, I'll put that on my list of things to do for Friday

components/sup/src/manager/service/mod.rs
bodymindarts: I follow your reasoning but I'm talking about having 2 explicit phases.
* render config
* execute hooks

I can't see why this should have anything to do with the language.

components/sup/src/manager/service/mod.rs
reset: Yeah you're right, this actually can completely be avoided. It's already being called in the `update_templates` function that I've created from refactored bits of `update_configuration`. Gonna make those changes and push them up!

Cargo.lock
```diff
@@ -1617,11 +1612,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
 name = "sodiumoxide"
 version = "0.0.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/dnaq/sodiumoxide#77bc149ed67681f6504439098297021ca1e6869f"
 dependencies = [
  "libc 0.2.22 (registry+https://github.com/rust-lang/crates.io-index)",
- "libsodium-sys 0.0.14 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 0.9.15 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libsodium-sys 0.0.14 (git+https://github.com/dnaq/sodiumoxide)",
+ "serde 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
```
fnichol: Wow, gotta keep up, right?

components/butterfly/src/rumor/service.rs
```diff
@@ -151,18 +152,27 @@ impl Rumor for Service {
     }
 }
 
-#[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct SysInfo {
-    pub ip: String,
+    pub ip: IpAddr,
```
fnichol: :1st_place_medal:

components/core/src/crypto/hash.rs
```diff
@@ -29,12 +29,12 @@ const BUF_SIZE: usize = 1024;
...
 /// NOTE: the hashing is keyless
 pub fn hash_file<P: AsRef<Path>>(filename: &P) -> Result<String> {
-    let file = try!(File::open(filename.as_ref()));
+    let file = File::open(filename.as_ref())?;
     let mut reader = BufReader::new(file);
     hash_reader(&mut reader)
 }
 
-pub fn hash_string(data: &str) -> Result<String> {
+pub fn hash_string(data: &str) -> String {
```
fnichol: That's funny, I wonder if there used to be a possible error here?

components/sup/src/census.rs
```diff
@@ -388,31 +455,39 @@ impl CensusGroup {
...
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
 pub struct CensusMember {
     pub member_id: MemberId,
     pub pkg: Option<PackageIdent>,
 
     service: String,
     group: String,
     org: Option<String>,
-    cfg: toml::value::Table,
+    #[serde(
```
fnichol: That's kickass!

components/sup/src/main.rs
```diff
@@ -51,8 +51,7 @@ use sup::command;
 use sup::http_gateway;
 use sup::manager::{Manager, ManagerConfig, ServiceStatus};
 use sup::manager::service::{DesiredState, ServiceBind, Topology, UpdateStrategy};
-use sup::manager::service::{ServiceSpec, StartStyle};
-use sup::supervisor::ProcessState;
+use sup::manager::service::{ProcessState, ServiceSpec, StartStyle};
```
fnichol: That's probably a good sign!

components/sup/src/manager/mod.rs
fnichol: I guess this was part of our unspoken conventions that we're taking from Aaron Turon's [Rust Guidelines](https://aturon.github.io/style/imports.html) book.

components/sup/src/manager/mod.rs
```diff
@@ -240,15 +246,15 @@ impl Manager {
...
+        let server = butterfly::Server::new(sys.gossip_listen(),
+                                            sys.gossip_listen(),
                                             member,
                                             Trace::default(),
                                             ring_key,
                                             None,
                                             Some(&fs_cfg.data_path),
                                             Box::new(SuitabilityLookup(services.clone())))?;
-        outputln!("Butterfly Member ID {}", server.member_id());
+        outputln!("Supervisor Member-ID {}", sys.member_id);
```
fnichol: That's a nice tweak and more clearly communicate with the extra internal jargon.

components/sup/src/manager/mod.rs
```diff
@@ -258,40 +264,57 @@ impl Manager {
...
                     }
-                    Err(err) => return Err(sup_error!(Error::BadDataFile(file_path.clone(), err))),
                 }
             }
         }
+        sys.member_id = member.get_id().to_string();
+        member.set_persistent(sys.permanent);
+        // JW TODO: Should we really be setting these here? Seems like they should be auto set
+        // by the butterfly server constructor
+        member.set_swim_port(sys.gossip_port as i32);
```
fnichol: That is a bit surprising, but works for the moment!

components/core/src/error.rs
```diff
@@ -138,7 +139,9 @@ impl fmt::Display for Error {
                 format!("Invalid keypath: {}. Specify an absolute path to a file on disk.",
                         e)
             }
-            Error::ConfigFileIO(ref e) => format!("Error reading configuration file: {}", e),
+            Error::ConfigFileIO(ref f, ref e) => {
+                format!("Error reading configuration file, {}, {}", f.display(), e)
```
mwrock: nice. Always a bummer when you have no idea what file the error is about

components/sup/src/manager/service/supervisor.rs
```diff
@@ -64,64 +61,25 @@ impl fmt::Display for ProcessState {
...
 #[derive(Debug)]
 pub struct Supervisor {
     pub child: Option<HabChild>,
-    pub package: Arc<RwLock<PackageInstall>>,
     pub preamble: String,
     pub state: ProcessState,
     pub state_entered: Timespec,
     pub has_started: bool,
-    pub runtime_config: RuntimeConfig,
+    pid: Option<PathBuf>,
```
mwrock: I don't see anywhere that sets this except where we ser it to `None` when removing the pid file.

reset: We use this in `create_pidfile` to record what the path is before we delete it in `cleanup_pidfile`. After we do that, we set it to `None`

mwrock: that function uses `child.id()` (as it should) I do not see any use of `self.pid` there. I might just be being my usual dense self here.

components/sup/src/templating/helpers.rs
```diff
@@ -1,162 +0,0 @@
-// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
```
mwrock: great move to separate out the helpers.

components/sup/src/manager/service/supervisor.rs
reset: It's on line 214, 224, and 231

mwrock: oh totally missed 214. `pid_file_path` might be a better name but not a huge deal.

components/sup/src/manager/mod.rs
```diff
@@ -592,30 +616,24 @@ impl Manager {
                 .write()
                 .expect("Services lock is poisoned!")
                 .iter_mut() {
-            if self.updater
-                   .check_for_updated_package(service, &self.census_ring) {
-                service.populate(&self.census_ring);
-                self.gossip_latest_service_rumor(&service);
```
mwrock: hmm. did you find this gossiping was not necessary? I'd think it would be something you would want to rumor out but maybe something else takes care of the rumor. Just curious about the rationale.

mwrock:
You had mentioned in a previous standup that this may break templates out in the wild. Is that still the case? If so, what kind of template syntax would break?

reset:
@mwrock It should be pretty minimal, but the biggest changes are a bug fix to `eachAlive` which used to behave like `withAlive` (it would inherit the inner context without calling `this` or using a block param) and removing census members from things which aren't your own service group from `svc`.

Other than that, I think we're pretty 1:1

components/sup/src/manager/mod.rs
reset: Yep! This should be completely unnecessary. We are going to call `tick()` pretty much right after this on the service which is going to tell us that it updated. Once it does, we will call `gossip_latest_service_rumor()` as a very next step

mwrock:
would this break:
```
{{#if bind.has_backend }}
{{~#each bind.backend.members}}
{{~#if alive }}
    server {{sys.ip}} {{sys.ip}}:{{cfg.port}}
{{~/if}}
{{~/each}}
{{~/if}}
```

or would it need to be:
```
{{#if bind.has_backend }}
{{~#each bind.backend.members as |member|}}
{{~#if alive }}
    server {{member.sys.ip}} {{member.sys.ip}}:{{member.cfg.port}}
{{~/if}}
{{~/each}}
{{~/if}}
```

components/sup/src/manager/mod.rs
mwrock: ahh that makes sense

mwrock:
Oh just found a bug here that kills status deserialization. I can add a fix to this branch.

reset:
@mwrock the `has_{}` field is completely gone. We have enough power with the handlebars templating language that we don't need it anymore (prior we were using just mustache). You can do things like this now:

```
{{~#eachAlive bind.backend.members as |member|}}
    server {{member.sys.ip}} {{member.sys.ip}}:{{member.cfg.port}}
{{~else}}
    server {{sys.ip}} {{sys.ip}}:8080
{{~/eachAlive}}
```

You *must* use the block syntax (`as |x|`) when using any version of `each`. It was a bug before that we inherited the outer context there. If we want that functionality then we need to write a `withAlive` helper

mwrock:
added a fix for the deserialization bug. Will unleash sentinals when green

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/229349611).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/229349611).

PR #2235, opened 2017-05-05
[plan-build] Add optional helper to dynamically compute pkg_version.

fnichol:
(Note: this builds on #2234, which should be merged first)

This change adds an alternative way to determine the version of a Plan
when building a package. In seeing more Plans in the wild, it has become
clear that some Plan authors wish to determine the package version by
cat'ing a common version file, using a Git command, or by some other
means which is dynamic.

There are 2 challenges to computing a package version:

* how the version is computed
* when the version should be computed

To implement the "how", a Plan author needs to write a `pkg_version()`
function that prints the version string to standard output. The Plan
author then explicitly calls the new `update_pkg_version()` helper in a
build phase to determine the "when".

The `update_pkg_version()` can be called in the `do_before()` build
phase all the way up to but not including the `do_prepare()` build
phase, after which it is too late to safely recompute the other build
variables that depend on `$pkg_version`. If `update_pkg_version()` is
called too late, the build program will fail the build with an error
message.

Due to this new optional strategy, a Plan author can omit defining a
`pkg_version` variable, as long as they write a `pkg_version()`
function. This is checked for by the build program so that if the Plan
author omits the `pkg_version` variable and the `pkg_version()`
function, the build will fail.

Examples
--------

The following are partial Plan implementations with extra build phases
omitted for brevity.

This is a Plan which cats a static file in the source root of the
project to determine the version which can be safely determined in the
`do_before()` phase:

```
pkg_origin=acme
pkg_name=myapp

pkg_version() {
  cat "$SRC_PATH/version.txt"
}

do_before() {
  do_default_before
  update_pkg_version
}
```

Here is the same Plan which uses a Git command to compute the number of
commits in the repository as the version. Note that since we are using
the `git` command, we are adding the `core/git` package as a build
dependency:

```
pkg_origin=acme
pkg_name=myapp
pkg_build_deps=(core/git)

pkg_version() {
  git rev-list master --count
}

do_before() {
  do_default_before
  update_pkg_version
}
```

Finally, here is a partial implementation of a `cacerts` Plan which
greps a date (YYYY.mm.dd) from the CA certs file itself. The
`pkg_version()` function does some grep'ing and sed'ing before using the
date program to format the final version string to standard output. As
the downloaded file is required before running the version logic, the
`update_pkg_version()` helper is called in the `do_download()` build
phase:

```
pkg_origin=me
pkg_name=cacerts
pkg_source=http://curl.haxx.se/ca/cacert.pem

pkg_version() {
  local build_date

[...]

PR #2234, opened 2017-05-05
[plan-build] Introduce `do_before()` & `do_after()` build phases.

fnichol:
This change introduces 2 new build phases to the public build API:

* `do_before()` (with an empty `do_default_before()` implementation)
* `do_after()` (with an empty `do_default_after()` implementation)

After inspecting more Plans in the wild, it has become apparent that
there is a missing build phase which Plan authors would have used rather
than hooking into the `do_download()` phase to perform tasks like
dynamically calculating a new `$pkg_version` value, etc.

The `do_begin()` phase is often far too early in the build process to
perform many tasks as package dependencies are not yet loaded, the
`$PATH` is not yet set, and other build environment variables are not
yet present. The first publicly supported phase in which to perform
such tasks is `do_download()` which occurs directly after resolving
dependencies and setting up the build environment.

Rather than allowing this "hacked" pattern to continue unfettered, an
additional build phase is introduced **before** the downloading,
building, installing is started.

In an attempt to properly "bookend" the `do_before()` build phase, a
`do_after()` build phase is added after the `do_strip()` build phase. At
this point in the build, the package has been built, installed, and
stripped, but before the package metadata is written and before the
artifact is created and signed. This is a mirror to `do_begin()`
(extremely early) and `do_end()` (extremely late).

Future work may optionally exploit the `do_before()` phase by allowing a
documented location to recompute the `$pkg_version` value.

gif-keyboard-9918045239862007893

PR #2237, opened 2017-05-05
Setup environment before determining version in plan build

reset:
The changes in #2234 and #2235 are awesome but it looks like we've got a small issue. The `_set_environment` should probably run prior to determining the value for `pkg_version`. It uses that variable in itself and it populates metadata, including the PATH metafile, which includes that in it's path. Here's a sample of output:

screen shot 2017-05-05 at 1 26 54 am

I'm going to merge this one in for now to get us unblocked. @fnichol please feel free to make any additional changes if you think we should load some of the parts before and some after, etc


thesentinels:

Thank you for contributing!

PR #2241, opened 2017-05-05
Logs!

### LINKED ISSUE #2024: Streaming log output for Builder and long term storage on job complete
When a build is initiated we need to stream the output as close to realtime to a subscribing client on the `builder-api` gateway. The initial subscriber we should develop for is the `builder-web` client.

Once a job is complete we will need to store those logs in some sort of long term storage solution.

christophermaier:
Implements enough backend logic to create a "streaming build log"
experience, as well as laying the foundation for long-term log
storage.

Overview
========

Builder workers run `hab studio build` to create Habitat packages,
just as you would on your local workstation. Previously, the log
output of these jobs was written to disk on the worker machines, but
never sent anywhere else. We would like to a) stream this data to
users of our web UI so that they may monitor builds in real-time and
b) provide the ability to look at build output once a job has
completed.

To do this, several pieces of ZeroMQ messaging infrastructure are
added, allowing us to transmit log data from running jobs line-by-line
to the job server for consolidation and archiving, and for access via
the API. Both the worker server and the job server start log-related
processes in separate threads. The process on the worker server --
`LogForwarder` -- accepts lines of output from a build job (`LogPipe`)
and forwards them to the process on the job server -- `LogIngester` --
which accumulates the lines in files on disk, one per job. (It will
eventually also trigger archiving the logs in long-term storage
off-disk when they are complete, but that is not implemented here.)

We have separate threads and socket pairs handling the log processing
in order to decouple that processing from the actual execution of the
jobs, and to prevent log messages from filling up other more important
message queues (like actual job dispatch).

With log data centrally located on the job server, we can also create
an API for accessing a job's log data. Though we eventually would like
a Server-Sent Events (SSE) endpoint to stream log data to clients, the
current implementation does *not* use SSE (see below for why), but
uses polling instead (this is not long polling; it’s just making a
normal request periodically). The client makes an initial request and
receives back all the job's log data the server currently has, along
with offset data, as well as an indication of whether the log is
complete or not. If it is not complete, the client can wait and then
make another request, using the offset it received earlier. In this
way, the client can build up the entire log over a series of
requests. When the log is complete, the client can simply stop
polling.

Room for Improvement
====================

The polling API endpoint is the biggest target for
improvements. Because Iron and Hyper don’t currently support
asynchronous and non-blocking I/O, we can’t easily implement a SSE
endpoint for streaming logs back to clients. There are not really any
full-featured Rust SSE libraries yet. Websockets are more complexity
than this feature needs, and in any case, existing libraries appear to
all use a thread-per-connection approach. We can, however, write our
own small async server, based on something like
[Tokio](https://tokio.rs), and run it alongside our existing
Iron-based API server behind our Nginx proxy. While this won't
necessarily be difficult, we opted for the current polling approach to
more quickly implement the desired functionality and get feedback on
it sooner.

Log data “chunks” are currently sent from the workers one line at a
time. Once we have an SSE endpoint, this will give the most immediate
feedback to users, but we can increase the chunk size (e.g., multiple
lines at a time) we send out from workers if we encounter performance
issues.

When an API request for a portion of a job’s log comes in, we simply
read the file from disk and return the requested lines. Once we have
an SSE endpoint, the log chunk messages can simply be forwarded on to
the process serving those streams, thus eliminating this repeated file
reading.

Long-term, we’d like to ship completed logs off-disk to a remote
location like S3 (while allowing for S3-compatible solutions for
air-gapped or non-AWS installations) When a job log is finished, we
will send the logs off disk and record the URL at which they can be
found in the database entry for the job in a `log_url` field. The
presence of a value in this field indicates that a log is
“complete”. We don’t yet have the S3 upload functionality, so for now
we’ll simply add a `file://` URL (pointing to the on-disk location on
the job server) to the database upon completion.

Of Interest
===========

* The job server now listens on port 5569 for incoming log data from
  workers.

* The job server will not start if it cannot write job log data.

  Log data is accumulated on the job server in files-on-disk. The
  location is specifiable in config for the job server, defaulting to
  `/tmp`. The directory given must exist and be writable by the
  server, or the server will refuse to start.

* Workers now run with `HAB_NONINTERACTIVE` set.

  This disables the download progress bars in `hab studio build`
  output. Without this, we'd need to resort to hacks to deal with all
  the carriage returns and every "frame" of the "animation".

* Workers will refuse to start if multiple job servers exist

  Currently the Job Server is not sharded and we only run one. Thus,
  log chunks can be sent directly to the Job Server, instead of
  needing to be routed to the server that manages the particular job's
  shard (thus ensuring we consistently send logs to the same place for
  accumulation). If we were to shard the job server, the current
  implementation would fail horribly, with a single job's log output
  spread out across as many job servers as we have running. Thus, as a
  precaution, the LogMgr process will panic on startup if it is given
  multiple job servers to connect to, essentially preventing any
  workers from starting. This is more to guard against potential
  future breakage than anything.

* Logging uses a multi-frame ZMQ message protocol

  In order to notify clients that they can stop polling, and to know
  when to archive logs in long-term storage, we need to know when a
  log is complete. We use a two-frame message to convey this
  information. The first frame is a one-character code indicating the
  type of message that follows in the second frame. An `"L"` means a
  log chunk follows, while a `"C"` indicates a completion protocol
  message follows. The `LogIngester` takes action and parses messages
  based on this information.

Fixes #2024


thesentinels:

Thank you for contributing!

components/builder-jobsrv/src/config.rs
```diff
@@ -34,6 +34,11 @@ pub struct Config {
     /// List of net addresses for routing servers to connect to
     pub routers: Vec<RouterAddr>,
     pub datastore: DataStoreCfg,
+
+    /// Directory to which log output of running build processes will
+    /// be written. Defaults to the system temp directory. Must exist
+    /// and be writable by the server process.
+    pub log_dir: String
```
chefsalim: Might be better to default it to the service directory (eg /hab/svc/builder-jobsrv) instead of temp?

components/builder-jobsrv/src/data_store.rs
```diff
@@ -207,6 +207,15 @@ impl DataStore {
         migrator.migrate("jobsrv",
                          r#"CREATE INDEX pending_jobs_index_v1 on jobs(created_at) WHERE job_state = 'Pending'"#)?;
 
+        migrator.migrate("jobsrv", r#"ALTER TABLE jobs ADD COLUMN IF NOT EXISTS log_url TEXT DEFAULT NULL "#)?;
```
chefsalim: Might not need to do an alter here since we will be starting with fresh DBs in our environments, and other recent schema changes were done directly in the table.

components/builder-jobsrv/src/server/log.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Ensures that the specified log directory can be used.
+    ///
+    /// # Panics
+    /// * when the path does not exist
+    /// * when the path is not a directory
+    /// * when the path is not writable
+    pub fn validate(&self) -> Result<()> {
+        let meta = fs::metadata(&self.path)
+            .expect(format!("Build log directory {:?} doesn't exist!",
```
chefsalim: Is there a specific reason to not just create the dir if it doesn't exist?  That is the pattern we seem to use more in the codebase.

components/builder-jobsrv/src/server/log_ingester.rs
```diff
@@ -0,0 +1,142 @@
...
+                            debug!("Log complete for job {:?}", id);
+
+                            // TODO: Save logs in long-term storage,
+                            // like S3
+                            let log_file = self.log_dir.log_file_path(id);
+
+                            // Until we can ship things off to an
+                            // S3-alike, we'll just use a file URL
+                            let url = format!("file://{}", log_file.to_str().unwrap());
+                            self.data_store.set_log_url(id, &url)?;
```
chefsalim: Is it intentional to have the set_log_url happen only on a Complete chunk?

components/builder-jobsrv/habitat/default.toml
```diff
@@ -7,6 +7,8 @@ worker_heartbeat_listen = "0.0.0.0"
 worker_heartbeat_port = 5567
 publisher_listen = "0.0.0.0"
 publisher_port = 5568
+log_ingestion_listen = "0.0.0.0"
+log_ingestion_port = 5569
```
chefsalim: This is just a FYI - the publisher_port is deprecated (and should have been removed earlier) - feel free to take 5568 if you like :)

components/builder-jobsrv/src/server/log.rs
```diff
@@ -0,0 +1,48 @@
...
+/// Encapsulates the local filesystem directory in which in-process
+/// build job logs will be collected prior to being sent to long-term
+/// storage.
+#[derive(Clone, Debug)]
+pub struct LogDirectory {
+    path: PathBuf
+}
+
+impl LogDirectory {
+
```
chefsalim: V. minor nit - the de-factor formatting pattern in the codebase is to not have a blank line after the impl line. Feel free to ignore :)

chefsalim:
Looks really good, with a few comments added.

components/builder-api/doc/api.raml
```diff
@@ -214,6 +242,54 @@ securitySchemes:
...
+                    200:
+                        body:
+                            application/json:
+                                schema: jobLog
+                                example: |
+                                    {
+                                        "start": 0,
+                                        "stop": 2,
+                                        "content": [
+                                            "this is the first line",
```
cnunciato: Right now, the API is actually returning `"1: this is the first line"` (i.e., the string includes the line number). Can we remove that, and just return the content?

components/builder-jobsrv/src/config.rs
christophermaier: Hrm... that may be better, yeah.

components/builder-jobsrv/src/server/log_ingester.rs
christophermaier: Once we have off-disk storage, we'll look to see if there's a log URL whenever someone asks for the logs; if it's there, we'll just retrieve it and return it to the user as a complete log. On the other hand, if the log data is still coming in, we don't want to be constantly updating and retrieving from an S3 bucket, so we'll just keep using the local file on disk until it's complete.

In this scenario, there's never going to be a time where we have a log stored off-disk (and thus need a URL) but are also still receiving additional lines to append to it, so I think having an additional flag wouldn't add anything.

components/builder-jobsrv/habitat/default.toml
christophermaier: Ah, I didn't realize that. Probably best to remove that and reset port numbers in a separate PR, then 😄

components/builder-jobsrv/src/server/log.rs
christophermaier: I still need to run `rustfmt` over everything, as Travis notified me earlier 😄

components/builder-api/doc/api.raml
christophermaier: That's one of the little things I still need to fix up... I added that to help with debugging the earlier incarnations of this, just to convince myself I was actually getting all the lines. I'll be sure to remove them before this merges!

components/builder-jobsrv/src/server/log_ingester.rs
chefsalim: Right, the S3/local fs logic is fine.. the confusing thing (to me) was the job_log_get does not use the log url, though it is getting set to a file://.. url.  If we say, the location is always in the log_url (whether the location points to a local fs or a remote location), then that seems more consistent for at least the first time reader of the code. The way you have it is an extra special case to parse mentally. Not a big deal though, I wouldn't hold off the merge or anything.

components/builder-jobsrv/habitat/default.toml
chefsalim: Yeah that's cool.. :)

components/builder-jobsrv/src/config.rs
```diff
@@ -34,6 +34,11 @@ pub struct Config {
     /// List of net addresses for routing servers to connect to
     pub routers: Vec<RouterAddr>,
     pub datastore: DataStoreCfg,
+
+    /// Directory to which log output of running build processes will
+    /// be written. Defaults to the system temp directory. Must exist
+    /// and be writable by the server process.
+    pub log_dir: String,
```
reset: I'd recommend using `PathBuf` in place of `String` when dealing with a filepath. It's worth noting that `PathBuf` == `String` and `Path` == `str`, this was something that didn't click for me for a few months in

components/builder-jobsrv/src/server/log.rs
```diff
@@ -0,0 +1,44 @@
+use error::Result;
```
reset: We need a header on this file

components/builder-jobsrv/src/server/log.rs
```diff
@@ -0,0 +1,44 @@
...
+        LogDirectory { path: PathBuf::from(path) }
+    }
+
+    /// Ensures that the specified log directory can be used.
+    ///
+    /// # Panics
+    /// * when the path does not exist
+    /// * when the path is not a directory
+    /// * when the path is not writable
+    pub fn validate(&self) -> Result<()> {
```
reset: This function returns a Result but all of the error cases are a panic. You can have it just simply return `()` if you did intend to use panics, but we typically recommend only using those if the developer absolutely has made a mistake

components/builder-jobsrv/src/config.rs
christophermaier: Oh weird... I thought I'd set this to a `PathBuf`; will change. Good catch!

components/builder-jobsrv/src/server/log.rs
```diff
@@ -0,0 +1,44 @@
...
+/// build job logs will be collected prior to being sent to long-term
+/// storage.
+#[derive(Clone, Debug)]
+pub struct LogDirectory {
+    path: PathBuf,
+}
+
+impl LogDirectory {
+    /// Create a new `LogDirectory` wrapping `path`.
+    pub fn new(path: &str) -> Self {
```
reset: A pattern you'll commonly see in any functions which deal with paths is something like this

```
pub fn new<T>(path: T> -> Self where T: AsRef<Path> {
    LogDirectory { path: path.as_ref().to_pathbuf() }
}
```

This will allow you to use various types to interact with the public functions of your API which all can be referenced as a Path

components/builder-jobsrv/src/server/log_ingester.rs
```diff
@@ -0,0 +1,141 @@
+use config::Config;
```
reset: We need a header here, too ;)

components/builder-jobsrv/src/server/log_ingester.rs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            let cfg = self.config.read().unwrap();
+            let addr = cfg.net.log_ingestion_addr();
+            println!("Listening for log data on {}", addr);
+            self.intake_sock.bind(&addr)?;
+        }
+
+        rz.send(()).unwrap();
+
+        loop {
```
reset: Looks solid to me

components/builder-jobsrv/src/server/log.rs
```diff
@@ -0,0 +1,44 @@
+use error::Result;
+use std::fs;
+use std::path::PathBuf;
+
+/// Encapsulates the local filesystem directory in which in-process
+/// build job logs will be collected prior to being sent to long-term
+/// storage.
+#[derive(Clone, Debug)]
+pub struct LogDirectory {
```
reset: This is called the [newtype pattern](https://aturon.github.io/features/types/newtype.html) in Rust. It's recommended to represent this pattern as a [tuple struct](https://doc.rust-lang.org/book/structs.html#tuple-structs) with one element. You can later acces that element in your functions as `self.0`

components/builder-jobsrv/src/server/log.rs
christophermaier: Nice... still wrapping my head around the type system 😄

components/builder-worker/src/log_forwarder.rs
```diff
@@ -0,0 +1,81 @@
+use config::Config;
```
reset: Header here too 😄

components/builder-worker/src/log_forwarder.rs
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+        self.intake_sock.bind(INPROC_ADDR)?;
+
+        // Signal back to the spawning process that we're good
+        rz.send(()).unwrap();
+
+        // This hacky sleep is recommended and required by zmq for
+        // connections to establish
+        thread::sleep(Duration::from_millis(100));
```
reset: This is literally the worst thing about zmq

components/builder-worker/src/runner/mod.rs
```diff
@@ -250,11 +249,14 @@ impl Runner {
         let mut child = Command::new(command)
             .args(&args)
             .env_clear()
+            // This disables download progress bars; otherwise we have
+            // to filter out loads of carriage returns!
+            .env("HAB_NONINTERACTIVE", "true")
```
reset: tenor-23872439

reset:
tenor-49819909

@christophermaier This was some highly valuable and tough to land work, excellent job :)

components/builder-jobsrv/src/server/log.rs
christophermaier: @reset Is there a preferred way you'd like to handle this? I was intending for things to not start if the logs can't be written to the directory.

components/builder-jobsrv/src/server/log.rs
reset: If you want to hear about `AsRef` I'd love to tell you about it, just message me after our demo ;)

components/builder-jobsrv/src/server/log.rs
reset: You would want to return an `Err<T>` where T is `error::Error`. I'd add a descriptive error to `error::Error` which describes that the log directory couldn't be created. I normally wrap the underlying `io::Error` in the case, so I'll define something like `Error::LogDirCreate(PathBuf, io::Error)` and then give it a formatted message like this: `format!("Unable to create log directory at {}, {}", path, err)`

christophermaier:
Will merge now so as not to block other work... I'll submit a follow-up PR for the suggestions brought up in review.

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/230928226).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/230928226).

PR #2279, opened 2017-05-09
move user and group validation outside of service struct constructor so failed services can have persisted state

### LINKED ISSUE #2273: Inconsistency between sup load and status
Ubuntu 16.04.2 docker container, Habitat 0.22.0

The setup:

```
docker pull ubuntu:xenial
docker run -it ubuntu:xenial

apt-get update
apt-get install -y curl
curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | bash
hab sup run &
```

We now have Ubuntu 16.04, and hab and hab-sup 0.22.0. Try to load `core/nginx`:

```
root# hab sup load core/nginx
<installs nginx>
Waiting for the core/nginx service to enter the up state......hab-sup(MR): Starting core/nginx
hab-sup(MR): Unable to start core/nginx, hab-sup(UR)[src/util/users.rs:42:27]: Package requires group hab to exist, but it doesn't
.........
hab-sup(MN): The core/nginx service was not succesfully loaded: hab-sup(MR)[src/manager/mod.rs:228:12]: Service core/nginx not loaded
```

As expected, this fails because we haven't yet created the `hab` user required by the service. Let's add the user and then load the service:
```
root# useradd hab
root# hab sup status core/nginx
hab-sup(MN): The core/nginx service is not currently loaded.
root# hab sup load core/nginx
hab-sup(MN)[src/main.rs:247:23]: Service already loaded, unload 'core/nginx' and try again
```

The `status` command is checking whether the status persisted in the services.dat file is present. The `load` command is checking whether there is a spec file present.

```
root# curl http://localhost:9631/services
[]
root# cat /hab/sup/default/data/services.dat
[]
root# cat /hab/sup/default/specs/nginx.spec
ident = "core/nginx"
group = "default"
depot_url = "https://willem.habitat.sh/v1/depot"
topology = "standalone"
update_strategy = "none"
binds = []
desired_state = "up"
start_style = "persistent"
```

Removing the spec file, or running `unload` on the service will allow the `load` command to work properly.

mwrock:
This fixes #2273. Currently creating a `Service` struct checks the `svc_user` and `svc_group` to ensure they are valid. If this fails, The service fails upon loading. Strictly speaking the spec is saved so any subsequent attempt to load the service fails stating that it is already loaded. Further, because the `Service` cannot be constructed, it also cannot be persisted and therefore the `status` command will not include it and will state that it is not loaded when the service is explicitly queried.

This PR moves the user/group validation out of the Service creation and into a separate `assert_pkg_user_and_group` that the `Service` calls berfore creating the service directories. `Service::load` no longer creates the svc directories. It simply loads the service. The supervisor manager now calls the service to create those directories when adding active services.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -178,7 +178,7 @@ pub struct Manager {
     updater: ServiceUpdater,
     watcher: SpecWatcher,
     organization: Option<String>,
-    service_states: Vec<Timespec>,
+    service_states: HashMap<PackageIdent, Timespec>,
     sys: Arc<Sys>,
 }
 
@@ -273,7 +273,7 @@ impl Manager {
                watcher: SpecWatcher::run(&fs_cfg.specs_path)?,
                fs_cfg: Arc::new(fs_cfg),
                organization: cfg.organization,
-               service_states: Vec::new(),
+               service_states: HashMap::new(),
                sys: Arc::new(sys),
            })
     }
@@ -413,6 +413,18 @@ impl Manager {
                 return;
             }
         };
+
+        if let Err(e) = service.create_svc_path() {
+            outputln!("Can't create directory {}: {}",
+                      service.pkg.svc_path.display(),
+                      e);
+            outputln!("If this service is running as non-root, you'll need to create \
+                       {} and give the current user write access to it",
+                      service.pkg.svc_path.display());
+            outputln!("{} failed to start", &spec.ident);
+            return;
+        }
+
         self.gossip_latest_service_rumor(&service);
         if service.topology == Topology::Leader {
             self.butterfly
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

reset:
This is good enough to get us over the finish line! In a future refactor I might suggest avoiding cloning the current state of services into a HashMap owned by the manager struct itself. Aside from the additional allocations (which really aren't much) this creates a potential caching problem which could result in some bugs

@thesentinels approve

tenor-93850701

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/230503764).

PR #2311, opened 2017-05-11
fix updated package template rendering

mwrock:
This fixes a regression in package updating where an updated package's templates are not recompiled.

This came down to two bugs:

1. We stopped gossiping on package update and as a result, a recompile was not triggered. @reset and I had discussed this in the refactor and I believe there was the thought that the service tick would update the templates and that would trigger the update rumor to be gossiped. However, the templates are only updated if the census ring has changed. So this adds back the gossip on update.

2. Even with the above fixed and recompile actually running, we are recompiling against the old hook path. So this adds a `Hooktable::load` to the `update_package` function.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -626,8 +626,10 @@ impl Manager {
                 .write()
                 .expect("Services lock is poisoned!")
                 .iter_mut() {
-            self.updater
-                .check_for_updated_package(service, &self.census_ring);
+            if self.updater
+                   .check_for_updated_package(service, &self.census_ring) {
+                self.gossip_latest_service_rumor(&service);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/230993914).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/230993914).

reset:
@mwrock excellent catch, I think we also need to create a new `CfgRenderer` when we generate our new `HookTable`

PR #2312, opened 2017-05-11
builder service plan bug fixes

reset:
* Set a more sensible default value for enabling builds in api config
* Fix issue in plans where routers would not be properly configured
* Ensure SSL_CERT env var is set before running supervisor. This is a
  workaround for now, we really need to address this

tenor-253658283


thesentinels:

Thank you for contributing!

PR #2356, opened 2017-05-13
Basic file log for builder-scheduler

chefsalim:
This change adds some basic event logging to the scheduler service. The events are logged to a file, and in a format that allows easy import and creation of job schedule gantt charts. Currently, the file is re-created every time the scheduler starts, and there is no provision for anything like log rotation, etc.

giphy 32


thesentinels:

Thank you for contributing!

components/builder-jobsrv/src/data_store.rs
```diff
@@ -165,7 +165,7 @@ impl DataStore {
                                         FOR UPDATE SKIP LOCKED
                                         LIMIT $1
                                     LOOP
-                                        UPDATE jobs SET job_state='Dispatched', updated_at=now() WHERE id=r.id RETURNING * INTO r;
+                                        UPDATE jobs SET job_state='Dispatched', scheduler_sync=false, updated_at=now() WHERE id=r.id RETURNING * INTO r;
```
reset: Now that we've deployed to acceptance, we can't edit functions anymore, now we need to create a new version and have the server call that. We could destroy the database and recreate it just before we go live so we clear all the old funcs

components/builder-scheduler/src/config.rs
```diff
@@ -30,6 +30,8 @@ pub struct Config {
     pub worker_threads: usize,
     /// Path to packages on-disk (for migration)
     pub migration_path: String,
+    /// Path to scheduler event logs
+    pub log_path: String,
```
reset: I would use `PathBuf` instead of `String` to represent file paths. I've been cleaning this up as I go throughout the codebase

components/builder-scheduler/src/server/logger.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use protocol::jobsrv::Job;
+use protocol::scheduler::{Group, Project};
+
+pub struct Logger {
+    file: File,
+}
+
+impl Logger {
+    pub fn init(log_path: PathBuf) -> Self {
```
reset: Whenever we have a function that takes a `Path` or a `PathBuf` we should always use a signature like this:

```
pub fn init<T>(log_path: T> -> Self where T: AsRef<Path> {
    log_path.as_ref().to_path_buf()
}

File: components/builder-scheduler/src/server/mod.rs [hunk 4]
```diff
@@ -105,18 +106,16 @@ impl Dispatcher for Worker {
                 sock: &mut zmq::Socket,
                 state: &mut Self::State)
                 -> Result<()> {
-        println!("Message received: {}", message.message_id());
+        debug!("Message received: {}", message.message_id());
```
reset: 👍

components/builder-jobsrv/src/data_store.rs
chefsalim: This particular change is not critical, it's a nice to have, re-creating the jobsrv db before we go live to pick it up is fine.

components/builder-scheduler/src/config.rs
chefsalim: Done

components/builder-scheduler/src/server/logger.rs
chefsalim: Done

components/builder-protocol/src/scheduler.rs
```diff
@@ -145,6 +145,7 @@ impl Serialize for Group {
         try!(strukt.serialize_field("id", &self.get_id()));
         try!(strukt.serialize_field("state", &self.get_state()));
         try!(strukt.serialize_field("projects", &self.get_projects()));
+        try!(strukt.serialize_field("created_at", &self.get_created_at()));
```
christophermaier: Do you need to call `to_rfc3339()` on this?

components/builder-scheduler/src/server/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+            .to_path_buf()
+            .join("builder-scheduler.log");
+        Logger {
+            file: File::create(filepath).expect("Failed to initialize builder-scheduler log file"),
+        }
+    }
+
+    pub fn log(&mut self, msg: &str) {
+        let dt: DateTime<UTC> = UTC::now();
+        let fmt_msg = format!("{},{}\n", dt.format("%Y-%m-%d %H:%M:%S"), msg);
```
christophermaier: Can you use `to_rfc3339()` here, too?

components/builder-scheduler/src/server/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+                .as_secs();
+            let duration = stop.signed_duration_since(start)
+                .to_std()
+                .unwrap()
+                .as_secs();
+
+            format!("{},{},{},{}",
+                    offset,
+                    duration,
+                    start.format("%Y-%m-%d %H:%M:%S").to_string(),
```
christophermaier: formatting here, too.

components/builder-scheduler/src/server/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+    // Log format (fields are comma-separated)
+    //   Log entry datetime (UTC)
+    //   Entry type - G (group), J (job), P (project)
+    //   Group id
+    //   State
+    //   Project name (for job or project)
+    //   Start datetime (UTC) (only for jobs)
+    //   End datetime (UTC) (only for jobs)
+    //   Start offset (offset from group creation, in seconds, only for jobs)
+    //   Duration (job duration, in seconds, only for jobs)
```
christophermaier: Thinking a bit further out, I think it would be a good idea to think about using structured logging (e.g. JSON), sending everything to standard output, 12-factor app style, and using an ELK stack (or similar) to aggregate and analyze this kind of output.

Defining a type for each log event, with appropriate `From` trait implementations, would be an interesting way to implement things here, and could make a future transition to structured JSON logging pretty unintrusive. It would also very clearly and unambiguously document the contract of each log event.

components/builder-scheduler/src/server/logger.rs
chefsalim: Unfortunately I explicitly had to re-format differently from RFC3339 format in order to get Google Sheets to parse it automatically.

components/builder-scheduler/src/server/logger.rs
chefsalim: Yep, that's exactly the infra we have started down the path on. I initially started down that path, then moved to this specific format for a different need at the moment. We'll enable the structured JSON logging for all services down the road and can revisit the needs then.

components/builder-protocol/src/scheduler.rs
chefsalim: Not needed, as we we fill this field as RFC3339 initially, so the output is in that format.

components/builder-scheduler/src/server/logger.rs
chefsalim: Yep same as above - need to have a Google readable format.

components/builder-scheduler/src/server/logger.rs
christophermaier: Cool, good to hear that's the plan 😄 

Could you add some comments to this file, indicating that it's more of an "immediate need" kind of thing, rather than a pattern that we'd like to use long term? If we don't have a ticket for doing "real" logging, could you also add one and reference this PR? I'd like to make sure we don't lose track of this and that we come back to make a more long-term solution.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/231964258).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/231964258).

components/builder-jobsrv/src/data_store.rs
reset: Sounds good to me!

PR #2394, opened 2017-05-17
Fix log-streaming performance

cnunciato:
Currently the build log is being rendered with Angular’s default change-detection behavior, which is causing it to be rendered much more often than it needs to be. This change adds an RxJS BehaviorSubject to allow subscribers to be notified when log segments are received, and updates the log component to append only those segments, rather than replace its content completely. Also drops back to two seconds just to go a little easier on the network calls.

image

### DIFF
--- a/components/builder-web/app/actions/builds.ts
+++ b/components/builder-web/app/actions/builds.ts
@@ -59,11 +59,13 @@ export function fetchBuildLog(id: string, token: string, start = 0) {
       .then((data) => {
           dispatch(populateBuildLog(data));
 
-          if (data["is_complete"] && start !== 0) {
+          let complete = data["is_complete"];
+
+          if (complete && data["start"] !== 0) {
             setTimeout(() => { dispatch(fetchBuild(id, token)); }, 5000);
           }
-          else if (getState().builds.selected.stream) {
-            setTimeout(() => { dispatch(fetchBuildLog(id, token, data["stop"])); }, 1000);
+          else if (!complete && getState().builds.selected.stream) {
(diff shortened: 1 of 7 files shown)


components/builder-web/app/actions/builds.ts
```diff
@@ -59,11 +59,13 @@ export function fetchBuildLog(id: string, token: string, start = 0) {
       .then((data) => {
           dispatch(populateBuildLog(data));
 
-          if (data["is_complete"] && start !== 0) {
+          let complete = data["is_complete"];
+
+          if (complete && data["start"] !== 0) {
```
christophermaier: I'm not quite following what this is doing... if the last chunk you got was the last one of the log, fetch the entire log in 5 seconds?

cnunciato: Almost -- `fetchBuild` fetches the _job_ in 5 seconds (not the log, just the job metadata) in order to reflect its final state in the UI.

christophermaier: Aaaaah, makes sense 👍

christophermaier:
@thesentinels approve
giphy

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/233333729).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/233333729).

PR #2404, opened 2017-05-18
Various doc clarifications and minor fixes.

lamont:
Signed-off-by: Lamont Lucas <lamont@fastrobot.com>

### DIFF
--- a/www/source/docs/concepts-depot.html.md
+++ b/www/source/docs/concepts-depot.html.md
@@ -6,6 +6,10 @@ title: Depot
 
 A service for storing Habitat packages for download and upload.
 
+Private depots are in the works but currently only the public depot is supported. You may explore the public depot via its web interface at [https://app.habitat.sh](https://app.habitat.sh)
+
+In addition to viewing packages, you may use the origins pane to manage your origins, optionally uploading your public org key via the webui rather than the `hab origin key upload` command. 
+
 <hr>
 <ul class="main-content--link-nav">
   <li>Continue to the next topic</li>
--- a/www/source/docs/run-packages-export.html.md
+++ b/www/source/docs/run-packages-export.html.md
@@ -21,6 +21,9 @@ You can create a Docker container image for any package by performing the follow
 
 5. You can now exit the studio. The new Docker container image exists on your computer and can be examined with `docker images` or run with `docker run`.
 
+
+Note that the `hab pkg install <local artifact>` command will only work if the studio has a copy of the origin's public key. If the install command fails with a 404 on the origin in question, make sure the public depot has a copy of the origin's public key.
+ 
 For an example of running a Habitat service in a Docker container, see the [Run your service](/tutorials/getting-started-process-build) step in the Getting Started tutorial.
 
 ## Exporting to an Application Container Image (ACI)
--- a/www/source/docs/run-packages-monitoring.html.md
+++ b/www/source/docs/run-packages-monitoring.html.md
@@ -35,7 +35,7 @@ Connect to the supervisor of the running service using the following syntax. Thi
 
       curl http://172.17.0.2:9631/services
 
-> Note: The default listening port on the supervisor is 9631; however, that can be changed by using the `--listen-http` option when starting a service.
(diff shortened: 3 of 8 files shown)


thesentinels:

Thank you for contributing!

www/source/docs/concepts-depot.html.md
```diff
@@ -6,6 +6,10 @@ title: Depot
 
 A service for storing Habitat packages for download and upload.
 
+Private depots are in the works but currently only the public depot is supported. You may explore the public depot via its web interface at [https://app.habitat.sh](https://app.habitat.sh)
+
+In addition to viewing packages, you may use the origins pane to manage your origins, optionally uploading your public org key via the webui rather than the `hab origin key upload` command.
```
eeyun: I'll defer to others on this, but considering extensive changes coming to depot in a matter of a couple days we might want to hold off on this. Theres likely some better documentation coming here. (I could be wrong though :) )

www/source/docs/run-packages-export.html.md
```diff
@@ -21,6 +21,9 @@ You can create a Docker container image for any package by performing the follow
 
 5. You can now exit the studio. The new Docker container image exists on your computer and can be examined with `docker images` or run with `docker run`.
 
+
+Note that the `hab pkg install <local artifact>` command will only work if the studio has a copy of the origin's public key. If the install command fails with a 404 on the origin in question, make sure the public depot has a copy of the origin's public key.
```
eeyun: A good note. The default behavior for the studio is to leverage several possible configuration locations to programmatically locate the public key and  clone that key into itself. Might be better to change this to point folks towards the docs that walk people through setting up their workstation?

www/source/docs/run-packages-monitoring.html.md
```diff
@@ -35,7 +35,7 @@ Connect to the supervisor of the running service using the following syntax. Thi
 
       curl http://172.17.0.2:9631/services
 
-> Note: The default listening port on the supervisor is 9631; however, that can be changed by using the `--listen-http` option when starting a service.
+> Note: The default listening port on the supervisor is 9631; however, that can be changed by using the `--listen-http` option when starting a service.  Additionally the HTTP API port will only listen if the sup was started with the `--peer` or `--topology` flags
```
eeyun: Refer to above comment about the REST endpoints.

www/source/docs/run-packages-export.html.md
bdangit: This sounds like more a FAQ "I keep seeing 404 when i try to install a package, what is going on?"

Another thing to note is that the workaround is not to have to depend on the public depot.  As long as you have a way to distribute the pub key that was used during building of the hart, then is the pub key that needs to be ensured that is in the studio or whenever the install needs to happen.  More or less, if you distribute the hart, you need to distribute the pub key that is associated with it otherwise Hab attempts to locate that info from the depot.

www/source/docs/run-packages-overview.html.md.erb
```diff
@@ -17,6 +17,8 @@ When entering an interactive studio, a supervisor is started for you in the back
 
 Because the supervisor is running in the background, you will not see the supervisor output as you start your service. However you can use the `slog` (supervisor log) command that will stream the tail of the supervisor output.
 
+You can use the `HAB_DOCKER_OPTS` ENV variable to pass flags to the docker run that starts the studio. e.g. `HAB_DOCKER_OPTS="-p 8080:8080"` to expose port 8080 for the service you are running inside the studio out to your controlling host.
```
bdangit: Might want to note that this only works under Mac dev environment and this is only for hab studio work.  I'm not sure about Windows dev environment, but I'm guessing that studio for Windows is also docker so which means this may apply to them too.

But to reiterate this is ONLY needed for hab studio work.  The container that is exported out of `hab pkg export docker <pkg_ident>` is just a normal docker container and you would expose the ports normally just not with HAB_DOCKER_OPTS.

reset:
@lamont are you still interested in seeing this PR through? You've got some good changes here! If you want to address the feedback and force push we can get it in

lamont-sauce:
Yeah, let me rebase and address the feedback, I'd be happy to pick this back up.

reset:
I'm going to close this out for now since there hasn't been any movement in a few weeks. Please open if you'd like to see these things get pushed through!

PR #2408, opened 2017-05-18
do not randomize shard 0

mwrock:
This fixes a bug that came up in the data migration from redis to postgres. The bug is that if a record (an origin in our case) was created on shard 0, subsequent shard assignments based on its route key would be randomized and its a foreign key violation party!

We have not encountered this bug in our test environments due to a very narrow distribution of origins but in migrating the origins from production data, there are several origins legally residing in shard 0.

This simply removes the shard randomizer for keys that route to shard 0. I would really like to get @adamhjk 's feedback to make sure he did not have a case here that I am overlooking.

### DIFF
--- a/components/builder-router/src/server.rs
+++ b/components/builder-router/src/server.rs
@@ -277,11 +277,10 @@ impl Server {
     }
 
     fn select_shard(&mut self) -> u32 {
-        let route_hash = self.envelope.route_info().get_hash();
-        if route_hash == 0 {
-            (self.rng.gen::<u64>() % SHARD_COUNT as u64) as u32
+        if self.envelope.route_info().has_hash() {
+            (self.envelope.route_info().get_hash() % SHARD_COUNT as u64) as u32
         } else {
-            (route_hash % SHARD_COUNT as u64) as u32
+            (self.rng.gen::<u64>() % SHARD_COUNT as u64) as u32
         }
     }
 }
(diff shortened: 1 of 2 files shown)


bobotclown:
This same logic exists in `components/builder-router/src/server.rs` in the `select_shard` function.

adamhjk:
The question is - is it wrong in both places?

components/builder-router/src/server.rs
```diff
@@ -278,11 +278,7 @@ impl Server {
 
     fn select_shard(&mut self) -> u32 {
```
reset: I did some digging. The thing is that the protobuf library does an `unwrap_or(0)` when you call `get_hash()` on the route_info. What you want to do here is see if hash is `Some` or `None` and not pull the value out

```rust
    fn select_shard(&mut self) -> u32 {
        if self.envelope.route_info().has_hash() {
            (self.envelope.route_info().get_hash() % SHARD_COUNT as u64) as u32
        } else {
            (self.rng.gen::<u64>() % SHARD_COUNT as u64) as u32
        }
    }
```

reset:
tenor-16347367

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/233840483).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/233840483).

PR #2427, opened 2017-05-19
Ship logs to an S3-compatible object store

### LINKED ISSUE #2305: Implement off-site long-term storage for job logs
Currently, job logs are stored on disk on the job server. We want to provide off-site long-term storage for logs, using an S3-compatible stored (S3, minio, ceph, etc.) 

See #2024

christophermaier:
*DON'T MERGE YET*: Still working out some kinks around local archiving for dev. The key functionality is present here now, though.

When a job log is complete, we now archive it in an S3 API-compatible
object store (tested against S3 and [minio][1]), deleting it from the
job server. Correspondingly, when an API request comes in for a given
log, we retrieve the contents from the archive and forward it to the
user.

One side effect of this implementation is that we no longer record a
`log_url` for a job in the database, but rather simply mark it as
"archived". The archiver itself knows how to retrieve the logs given
a job ID.

The implementation use the `aws-sdk-rust` crate, which, despite the
general name, only interfaces with S3. This was chosen instead of the
more widely-used `rusoto` library specifically because `aws-sdk-rust`
has the ability to interact with arbitrary compatible object stores,
and not just AWS S3; this was a constraint of the feature, because we
did not want to mandate the use of AWS to run builder.

Configuration
=============

The job server's configuration now has an `[archive]` section. Here is
an example for using a bucket in S3's `us-east-1` region:

```toml
[archive]
key = "THIS_IS_A_KEY"
secret = "THIS_IS_A_SECRET"
bucket = "habitat-log-bukkit"
```

Alternatively, you may specify a region if you do not use the
`us-east-1` default.

```toml
[archive]
key = "THIS_IS_A_KEY"
secret = "THIS_IS_A_SECRET"
bucket = "habitat-log-bukkit"
region = "us-west-2"
```

To use Minio, you must supply an endpoint:

```toml
[archive]
key = "THIS_IS_A_KEY"
secret = "THIS_IS_A_SECRET"
bucket = "habitat-log-bukkit"
endpoint = "http://minio.mycompany.com:9000"
```

By default, Minio uses `us-east-1` as its region (for API
compatibility); if you override this in your Minio installation, you
should also include this region in your configuration.

The archiver will access objects in the bucket using the "virtual
bucket" model (e.g. `${BUCKET_NAME}.s3.amazonaws.com` for AWS. If
for some reason you want / need to use the path model
(e.g. `s3.amazonaws.com/${BUCKET_NAME}`), you will need to specify
both the appropriate region and the corresponding endpoint, such as:

```toml
   # ...
   region = "us-west-2"
   endpoint = "http://s3-us-west-2.amazonaws.com"
   # ...
```
(In the future this may become a configuration option.)

For the appropriate endpoint name, see
http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region For
Minio, the archiver currently always uses the path access model.

Failure Modes
=============

As soon as the log is recognized as being complete, we archive it. If
the archiving call fails for whatever reason, we do not delete the
local log file on disk. API calls to retrieve the log will still work,
being satisfied from the on-disk file, but it won't be marked as
"complete" in the API response, which would cause the UI to repeatedly
query for more lines. This is probably going to be rare, but we should
come up with a way to try and re-archive things that couldn't get
archived the first time around; we'll have the data on hand to
determine which file should be retried.

Fixes #2305

[1]: https://minio.io


thesentinels:

Thank you for contributing!

christophermaier:
Closing in favor of #2463

PR #2463, opened 2017-05-21
Archive job logs

### LINKED ISSUE #2305: Implement off-site long-term storage for job logs

christophermaier:
When a job log is complete, we now archive it for long-term
storage. This commit introduces a `LogArchiver` trait with two
implementations: a `LocalArchiver` which stores logs in the local
filesystem, and `S3Archiver`, which stores logs in an S3
API-compatible object store (tested against S3 and
[minio][1]). Following successful archiving, the temporary log file is
deleted from the job server. Thereafter, the archiver is used to
retrieve completed job logs (e.g., when servicing API requests)

One side effect of this implementation is that we no longer record a
`log_url` for a job in the database, but rather simply mark it as
"archived". The archiver itself knows how to retrieve the logs given a
job ID. A corrollary to this is that if you change how the archiver is
configured (e.g., switch from local to S3, change S3 buckets, switch
from AWS S3 to Minio, etc.) you are responsible for moving any
existing files such that your newly-configured archiver can find them.

The local archiver uses the local filesystem for storage. It avoids
the problems of large numbers of files in a single directory by
creating a nested directory structure based on the SHA256 checksum of
a job ID (e.g. `/archive/97/6e/48/3c/722477594578067456.log`). This is
the same approach that Chef's `bookshelf` cookbook storage engine
uses. This archiver is the recommended one to use for local
development purposes (though running Minio in a Docker container is
quite simple, too).

The S3 implementation use the `aws-sdk-rust` crate, which, despite the
general name, only interfaces with S3. This was chosen instead of the
more widely-used `rusoto` library specifically because `aws-sdk-rust`
has the ability to interact with arbitrary compatible object stores,
and not just AWS S3; this was a constraint of the feature, because we
did not want to mandate the use of AWS to run builder.

Configuration
=============

The job server's configuration now has an `[archive]` section. Here is
how to configure it for the local archiver:

```toml
[archive]
local = true
local_dir = "/path/to/my/properly/backed/up/archive"
```

The directory must exist and be writable.

Here is an S3 example for using a bucket in AWS's `us-east-1` region:

```toml
[archive]
key = "THIS_IS_A_KEY"
secret = "THIS_IS_A_SECRET"
bucket = "habitat-log-bukkit"
```

Alternatively, you may specify a region if you do not use the
`us-east-1` default.

```toml
[archive]
key = "THIS_IS_A_KEY"
secret = "THIS_IS_A_SECRET"
bucket = "habitat-log-bukkit"
region = "us-west-2"
```

To use Minio, you *must* supply an endpoint:

```toml
[archive]
key = "THIS_IS_A_KEY"
secret = "THIS_IS_A_SECRET"
bucket = "habitat-log-bukkit"
endpoint = "http://minio.mycompany.com:9000"
```

By default, Minio uses `us-east-1` as its region (for API
compatibility); if you override this in your Minio installation, you
should also include this region in your jobsrv configuration.

The S3 archiver will access objects in the bucket using the "virtual
bucket" model (e.g. `${BUCKET_NAME}.s3.amazonaws.com` for AWS. If for
some reason you want / need to use the path model
(e.g. `s3.amazonaws.com/${BUCKET_NAME}`), you will need to specify
both the appropriate region and the corresponding endpoint, such as:

```toml
   # ...
   region = "us-west-2"
   endpoint = "http://s3-us-west-2.amazonaws.com"
   # ...
```
(In the future this may become a configuration option.)

For the appropriate endpoint name, see
http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region For
Minio, the archiver currently always uses the path access model.

**Note** For either archiver, these configuration options _must_ be in place for the job server to start. We'll need to deal with this before deploying.

Failure Modes
=============

As soon as the log is recognized as being complete, we archive it. If
the archiving call fails for whatever reason, we do not delete the
local log file on disk. API calls to retrieve the log will still work,
being satisfied from the on-disk file, but it won't be marked as
"complete" in the API response, which would cause the UI to repeatedly
query for more lines. Additionally, if the temporary logs are not
backed up (the default directory for them is `/tmp`), they stand to be
lost if they are not archived. This is probably going to be rare, but
we should come up with a way to try and re-archive things that
couldn't get archived the first time around; we'll have the data on
hand to determine which file should be retried.

Fixes #2305

[1]: https://minio.io


components/builder-jobsrv/src/server/mod.rs
```diff
@@ -80,6 +84,14 @@ impl ServerState {
     fn log_dir(&self) -> &LogDirectory {
         self.log_dir.as_ref().unwrap()
     }
+
+    // fn get_foo<'a>(&'a self) -> &'a Foo {
```
mwrock: did you mean to leave this commented bit in?

christophermaier: Sigh... no 🤕

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/238042765).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/238042765).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/238042765).

PR #2542, opened 2017-06-06
make clear that the cmd means the public key

ShalokShalom:


thesentinels:

Thank you for contributing!

mwrock:
You can actually upload the private key as well both via the web UI or the command line. Is that not working for you?

ShalokShalom:
I got confused by the text:

" If you do not upload at least the public key for the origin, the depot will reject the upload of your packages under that origin."

and together with that:

"You can upload keys for the origin through the web interface for the depot, or by using the hab origin key upload command."

Since that sounds like this uploads both, the private and the public keys. 
I agree that my change makes less sense since it does not apply to the web UI, while the command to upload private keys look a bit different. 

I just like to make clear, that the common behavior is uploading a public key and how this is working. 
So long as the upload of a private key is something special, it should be marked so. 

I hope this is clear :/

ShalokShalom:
And where can i upload packages on the web interface?

https://bldr.habitat.sh/#/

mwrock:
You must use the hab cli to upload packages

ShalokShalom:
So the current text speaks about that wrong anyway, let me create another issue.

PR #2565, opened 2017-06-08
Create a .gitignore file with `hab plan init`

### LINKED ISSUE #2544: hab plan init should create .gitignore
at minimum, all the build results should be ignored

```
# .gitignore
results/*
*.DS_Store
```

cnunciato:
For git-managed projects, this adds a .gitignore file when one doesn’t already exist, appending only those lines that wouldn't result in duplicates.

Closes #2544.

### DIFF
--- a/components/hab/src/command/plan/init.rs
+++ b/components/hab/src/command/plan/init.rs
@@ -14,8 +14,8 @@
 
 use std::env;
 use std::fs::create_dir_all;
-use std::fs::{File, canonicalize};
-use std::io::Write;
+use std::fs::{File, canonicalize, OpenOptions};
+use std::io::{Write, BufRead, BufReader};
 use std::path::Path;
 use std::collections::HashMap;
 
@@ -28,6 +28,8 @@ const PLAN_TEMPLATE: &'static str = include_str!(concat!(env!("CARGO_MANIFEST_DI
                                                          "/static/template_plan.sh"));
 const DEFAULT_TOML_TEMPLATE: &'static str = include_str!(concat!(env!("CARGO_MANIFEST_DIR"),
                                                                  "/static/template_default.toml"));
+const GITIGNORE_TEMPLATE: &'static str = include_str!(concat!(env!("CARGO_MANIFEST_DIR"),
+                                                              "/static/template_gitignore"));
 
 pub fn start(ui: &mut UI,
              origin: String,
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/plan/init.rs
```diff
@@ -115,10 +117,40 @@ pub fn start(ui: &mut UI,
...
+fn render_gitignore(ui: &mut UI, root: &str) -> Result<()> {
+    let target_path = format!("{}/../.gitignore", root);
+    let path = Path::new(&target_path);
+
+    match path.exists() {
+        false => {
+            create_with_template(ui, &target_path, GITIGNORE_TEMPLATE)?;
+        }
+        true => {
+            let file = OpenOptions::new().read(true).write(true).append(true).open(path)?;
```
christophermaier: You can leave off `write(true)` here, as `append(true)` implies it.

components/hab/static/template_gitignore
```diff
@@ -0,0 +1,2 @@
+.DS_Store
```
christophermaier: Do we need this, given that it's not Habitat-specific?

smith: .DS_Store is not specific to Habitat and should be handled by the user in a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

If you agree, that would make the contents of this file just one line and you probably could just inline that string in init.rs instead.

components/hab/src/command/plan/init.rs
```diff
@@ -115,10 +117,40 @@ pub fn start(ui: &mut UI,
                your habitat. There are several hooks to create and tweak! See the full list \
                with info here: https://www.habitat.sh/docs/reference/plan-syntax/#hooks"));
 
+    try!(render_gitignore(ui, &root));
+
     try!(ui.end("A happy abode for your code has been initialized! Now it's time to explore!"));
     Ok(())
 }
 
+fn render_gitignore(ui: &mut UI, root: &str) -> Result<()> {
```
christophermaier: We should probably only do this if we detect a `.git` directory somewhere in the file path above us. Otherwise poor bazaar users will get a file they don't want.

components/hab/src/command/plan/init.rs
```diff
@@ -115,10 +117,40 @@ pub fn start(ui: &mut UI,
...
+    let target_path = format!("{}/../.gitignore", root);
+    let path = Path::new(&target_path);
+
+    match path.exists() {
```
chefsalim: might be cleaner to just do an if/else instead of match here?

components/hab/static/template_gitignore
```diff
@@ -0,0 +1,2 @@
...
+results/
```
christophermaier: This also might get weird with multi-package repositories (like `habitat` itself), where you would have multiple directories in a repository that you might run `hab pkg init` in. Should we add `/path/to/subproject/results` to a top-level `.gitignore`? Would we instead recommend that if you use that pattern you should always enter your studio from the top level directory and then run `build path/to/subproject` once you get in (so your artifacts always go into the top-level `results` directory?

christophermaier:
@cnunciato the code itself looks pretty good (nice first Rust! :clap:); most of my comments get into different edge cases we might want to think more about.

components/hab/static/template_gitignore
christophermaier: Tangentially related... the name `results` is rather general. I wonder if we should consider renaming to `hab-results`, `hab-artifacts`, etc.; basically make it obvious that this is a Habitat-related directory.

components/hab/src/command/plan/init.rs
cnunciato: Ah, excellent thanks.

components/hab/src/command/plan/init.rs
cnunciato: Yeah, that makes sense.

components/hab/src/command/plan/init.rs
cnunciato: I think so too. I was following other examples in this file, but I like that better also.

components/hab/static/template_gitignore
cnunciato: We should leave it out, yeah. I just copied it out of the Issue, but on second thought, it doesn't belong.

components/hab/static/template_gitignore
cnunciato: This is a good question and one I don't have a great answer for myself. I think if I were working that way, it might bug me a bit to have my tree noised up with a bunch of .gitignores -- especially if I've already put one (as we have) at the top level -- but I'd probably just tidy things up manually once I noticed what was happening. But that's not the awesomest user experience, I agree. 

We could maybe locate the first .git ancestor, look for a .gitignore there, and create/append at that location instead -- but would that always be the right place? I'm not sure. 🤔

mwrock: yeah I think its a really good question and think @cnunciato 's idea to let the user manage through a top level .gitignore  is likely the safest approach. Searching for ancestor ignore files feels a bit too "clever" for now.

components/hab/src/command/plan/init.rs
```diff
@@ -115,10 +117,65 @@ pub fn start(ui: &mut UI,
...
     try!(ui.end("A happy abode for your code has been initialized! Now it's time to explore!"));
     Ok(())
 }
 
+fn render_ignorefile(ui: &mut UI, root: &str) -> Result<()> {
+    let parent = format!("{}/..", root);
+    let expanded = canonicalize(&parent)?;
+    let current_path = Path::new(&expanded);
+
+    fn is_git_managed(path: &Path) -> bool {
```
reset: While it's _valid_ to define a function within another, I would recommend doing this outside of the `render_ignorefile()` function

ShalokShalom:
:+1:

raskchanky:
tenor-229256711

raskchanky:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/242952400).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/242952400).

PR #2632, opened 2017-06-21
Add more support for channels to hab cli

chefsalim:
This change adds more support to the hab cli for channels. It adds a HAB_DEPOT_CHANNEL environment variable that can be used to specify a default channel.  If the env variable is not specified, the default channel is 'stable'. This change also lets the 'hab pkg upload' command specify a channel to promote the package to after upload. It also adds a 'hab pkg promote' command to promote a package to another channel.

NOTE: A release of hab cli with this change should only happen subsequent to migrating Depot packages to stable.

### DIFF
--- a/components/builder-depot-client/src/lib.rs
+++ b/components/builder-depot-client/src/lib.rs
@@ -179,6 +179,19 @@ pub struct PackageResults<T> {
     pub package_list: Vec<T>,
 }
 
+#[derive(Clone, Deserialize)]
+pub struct OriginChannelIdent {
+    pub name: String,
+}
+
+impl Into<originsrv::OriginChannelIdent> for OriginChannelIdent {
+    fn into(self) -> originsrv::OriginChannelIdent {
+        let mut out = originsrv::OriginChannelIdent::new();
+        out.set_name(self.name);
+        out
+    }
+}
+
 pub trait DisplayProgress: Write {
     fn size(&mut self, size: u64);
     fn finish(&mut self);
@@ -413,6 +426,8 @@ impl Client {
     where
         I: Identifiable,
     {
+        // TODO: When channels are fully rolled out, we may want to make
+        //       the channel specifier mandatory instead of being an Option
         let mut url = if let Some(channel) = channel {
             channel_package_path(channel, package)
         } else {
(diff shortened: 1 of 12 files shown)


thesentinels:

Thank you for contributing!

reset:
@chefsalim this is excellent and exactly what we need. I think we should wait to merge until the Builder Web UI correctly represents packages in channels. I think @cnunciato and @ryankeairns were cooking up some details on what that looks like?

One more minor change I would suggest: if you attempt to install a package which isn't in a channel or you request latest of a package in a channel which is empty, we may want to recommend to the user to specify a different channel which we know isn't empty or does contain the package they're looking for

ryankeairns:
We did have a brief conversation on Monday, but I'd like to get @tashimi 's take on this too.

One thing we discussed adding to the UI - on the builds (list) view - was a filter to sort packages by channel, but I'm also thinking it could be even simpler to know which version of the package is in stable currently. I suspect the former incorrectly presumed that many packages could be in a given channel (that was my missed assumption at the time).

So a specific question I have, in order to clarify that, is can only one package/version/release be in stable at any given time? So if I had 3.2.5 in unstable and stable, and now 3.2.6 has come along - does 3.2.5 get pushed out? (I'm presuming it does).

The actual time to mock this initial version up feels small, just need to clarify that point. Also, depending upon urgency/priority (this is also where I'd like Tasha's input), we could fit this in to the existing UI (maybe as a simple table column or tag) before we revise the package detail view (where we're adding the README and consolidating/reducing clicking).

ryankeairns:
One other qq - are we sending channel info in the API currently, or is that work that still needs to be done as well?

reset:
@ryankeairns a channel may have any number of packages in it at once. When we speak about a package in this context I mean a fully qualified identifier - this means any number of versions + releases (the opposite of your assumption)

We're sending everything to the API regarding channels as far as I know. We may need to make some optimizations but it can most certainly be put in as is ;)

ryankeairns:
@reset roger that. Making some tweaks based on feedback from team, then meeting with Chris in the morning to get it in motion.

reset:
tenor-145068555

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/248122722).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/248122722).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/248122722).

PR #2648, opened 2017-06-24
Start services using specified hab_svc_user on windows

mwrock:
Currently we do not honor `hab_svc_user` on windows. We start windows based services under the same account that the supervisor is running as. This PR launches services using the logon token of `hab_svc_user` if specified.

Because we need to generate a logon token for the user, we need the logon credentials of that user and also need to persist those credentials so that the supervisor can use them if it needs to restart the service or start a newly loaded service. We use DPAPI to encrypt the user password and store it in the service spec on disk. Both `hab sup start` and `hab sup load` will take a `--password` parameter and store it in the spec. The password can only be decrypted by the hab binary as long as it is compiled with an additional byte array that provides added entropy to the encryption. If no byte array is provided at compile time, the passwords could feasibly be decrypted as long as the decryption is attempted on the same machine where the password was encrypted. We are storing a production byte array in base64 format in our password vault and passing to the appveyor build via a secure environment variable.

In order to start services using another user's logon token, the account  running the supervisor must have the `SE_INCREASE_QUOTA_NAME` and `SE_ASSIGNPRIMARYTOKEN_NAME` privileges assigned. The supervisor will validate these privileges are assigned and will return an error stating which is missing when attempting to start services with lacking privileges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,6 +240,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crypt32-sys"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "ctrlc"
 version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -861,6 +870,7 @@ name = "habitat_core"
 version = "0.0.0"
 dependencies = [
  "base64 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "crypt32-sys 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "errno 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "gcc 0.3.45 (registry+https://github.com/rust-lang/crates.io-index)",
  "habitat_win_users 0.0.0",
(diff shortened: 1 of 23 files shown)


mwrock:
Note that this is "mostly" working now. The following needs to occur before merge:

- [ ] Format and rebase with current master
- [ ] Cleanup all the new FFI (its a total mess) and move it out of `windows_child.
- [ ] Create a propper environment block for the environment variables. The current logic just inherits all vars from the supervisor user
- [ ] Implement the password encryption and decryption DPAPI calls
- [ ] Add a "secret" byte string to 1password and the appveyor secret vars. We'll use this to "add complexity" to the encryption so that ONLY hab can ever decrypt the passwords

mwrock:
fixes #2125

mwrock:
Just made a small but important change based on some more testing I did today: We now specify `LOGON32_LOGON_SERVICE` when generating the logon token for `hab_svc_user` this allows the service to run with full admin privileges (if the user is an administrator). Otherwise, the service could possibly encounter errors when performing admin operations.

This means that `hab_svc_user` must have this right assigned to it. If creating the process results in an error due to this right missing, we will return an error clearly stating what logon right needs to be granted to the user.

reset:
@mwrock this looks great! Resolve the lockfile conflicts and merge it in!

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/250588499).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/250588499).

PR #2699, opened 2017-07-07
fix the mac and windows build

mwrock:
This fixes the following:

* Fix the travis build number variable passed to the mac builder causing us to target an old dev build
* Update rust - it was using 0.16.0 which was raising compile errors on the latest code
* Add /usr/local/bin to path which without was breaking all sorts of things
* use `hab studio build` instead of the broken `hab pkg build` on windows. Will be fixing `hab pkg build` later.


thesentinels:

Thank you for contributing!

reset:
tenor-16280718

PR #2706, opened 2017-07-07
Default to unstable channel on package upload

reset:
Packages should be uploaded to the unstable channel by default and not
the stable channel. The stable channel can be specified as the destination
*as well as unstable* by passing `--channel stable` to the command

tenor-214863295

There's a little bit of cleanup in here as well

### DIFF
--- a/components/hab/src/command/pkg/upload.rs
+++ b/components/hab/src/command/pkg/upload.rs
@@ -52,26 +52,27 @@ use retry::retry;
 /// * Fails if it cannot find a package
 /// * Fails if the package doesn't have a `.hart` file in the cache
 /// * Fails if it cannot upload the file
-pub fn start<P: AsRef<Path>>(
+pub fn start<T, U>(
     ui: &mut UI,
     url: &str,
     channel: Option<&str>,
     token: &str,
-    archive_path: &P,
-    key_path: &P,
-) -> Result<()> {
+    archive_path: T,
+    key_path: U,
+) -> Result<()>
+where
+    T: AsRef<Path>,
+    U: AsRef<Path>,
+{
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

nellshamrell:
👍

components/hab/src/command/pkg/upload.rs
```diff
@@ -52,26 +52,27 @@ use retry::retry;
...
     token: &str,
-    archive_path: &P,
-    key_path: &P,
-) -> Result<()> {
+    archive_path: T,
+    key_path: U,
+) -> Result<()>
+where
+    T: AsRef<Path>,
+    U: AsRef<Path>,
```
christophermaier: What's the benefit of having two type aliases here instead of one, given that they're the the same?

reset: So the way this works is that the first time you match a generic it gets "bound" to the type which was passed to it. This means that if you pass an `&str` to the first instance of `P` and then a `Path` to the second, the function head won't get matched.

This function looks to be written before we knew how generics actually worked. You can see we re-used `P` and also referenced it with `&P` in the function signature even though we really just meant `P`

christophermaier: Ah, I suspected it might be that. Good to know.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/251298013).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/251298013).

PR #2707, opened 2017-07-07
Heartbeat refactor

nellshamrell:
This is the first step in moving the heartbeat code from builder-worker into the net crate.  Because this change is so invasive, I'm trying it in small steps, and this is the first very doable one.


thesentinels:

Thank you for contributing!

nellshamrell:
The HeartbeatManager portion of the code is still in builder-worker - that is so deeply intertwined with the job server I'm not sure it would be the best use of time to pull it out.  This at least moves the HeartbeatCli code to a central place - so we may be able to implement a heartbeat manager on each component that needs a heartbeat, and then use the central HeartbeatCli functionality as well.

nellshamrell:
upon closer inspection, this is not quite ready for review yet, will re-open after I make some more changes

nellshamrell:
Ok!  Ready to review again!

components/builder-protocol/protocols/jobsrv.proto
```diff
@@ -2,17 +2,6 @@ package jobsrv;
 import "protocols/net.proto";
 import "protocols/originsrv.proto";
 
-enum Os {
```
reset: `Os` and `WorkerState` do belong in this protocol file

components/net/src/heartbeat.rs
```diff
@@ -0,0 +1,68 @@
+use error::Result;
```
reset: Need a license at the top of this file :)

reset:
@nellshamrell I now see that this is the first of many planned changes so feel free to ignore my comments for now!

One thing I'd weigh here is should we make a generic heartbeat sender/receiver and how do we add additional things to it. So for example: The heartbeat for a sharded service may contain information about shards while the heartbeat for a worker contains the worker state

I'm 👍 for merging this in with that said if we want to do that in the next one!

components/net/src/heartbeat.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// In-memory zmq address for HeartbeatMgr
+const INPROC_ADDR: &'static str = "inproc://heartbeat";
+/// Protocol message to notify the HeartbeatMgr to begin pulsing
+const CMD_PULSE: &'static str = "R";
+/// Protocol message to notify the HeartbeatMgr to pause pulsing
+const CMD_PAUSE: &'static str = "P";
+
+#[derive(PartialEq)]
+enum PulseState {
```
chefsalim: Should PulseState also get removed from builder-worker/src/heartbeat.hs?  Looks like there are duplicates.

components/builder-jobsrv/src/server/worker_manager.rs
```diff
@@ -252,10 +253,10 @@ impl WorkerMgr {
 
     fn process_heartbeat(&mut self) -> Result<bool> {
```
chefsalim: I assume this will come later - if we abstract out the HB functionality, it would be nice to not have any knowledge of the HB implementation internals needed in the consuming modules.

nellshamrell:
Closing this after discussion of a better solution.

PR #2710, opened 2017-07-07
deploy scripts should upload to stable

mwrock:


thesentinels:

Thank you for contributing!

nellshamrell:
👍

reset:
@mwrock I think we want to deploy to stable only if this is an actual stable build. With this change I think we'll consider things from our `-dev` builds as stable

mwrock:
@reset currently `-dev` merges go to acceptance and prod releases go to prod. So both need to go to stable in their respective depots. I think that now that we have channels, moving from acceptance to prod/unstable might be better but I think that is a separate change.

reset:
@mwrock Ah I see, ya we should make a separate change for it but that's the right move IMO

reset:
tenor-174201456

PR #2729, opened 2017-07-13
Restart services when hook files change

### LINKED ISSUE #2448: Hook-only config updates do not restart services
- Mac OS 10.12

Given a plan that _only_ has a hook, but that hook has dynamic configuration, Habitat will rewrite the hook, but **will not** restart the service.

Example hook:

```sh
#!/bin/sh

exec http-echo \
  -listen="{{sys.ip}}:{{cfg.port}}" \
  -text="{{cfg.text}}" \
  2>&1
```

In this case, the service listens on `cfg.port` and renders `cfg.text`. When a configuration change is pushed:

```
$ hab config apply http-echo.default 2 <<< "port = 1234"
```

The configuration is changed (the hook actually changes on disk - verified by attaching to the container), but the service is not restarted. As a result, the service is now listening on the old port, but _new_ services would listen on the new port. Additionally, binding would fail, since it would bind to the old port.

As a workaround, creating a config file that depends on any configuration option triggers the service restart.

/cc @adamhjk

christophermaier:
Fixes #2448 and also introduces some new integration testing for the supervisor.


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/hooks.rs
```diff
@@ -97,19 +98,56 @@ pub trait Hook: fmt::Debug + Sized {
...
+            crypto::hash::hash_file(&self.path().to_path_buf()).map_err(|e| SupError::from(e))
+        } else {
+            Ok(String::new())
+        }
+    }
+
+    /// Write `maybe_new_content` into the hook file only if it's
+    /// different from what's already there. Return `true` if the file
+    /// content was changed.
+    fn maybe_change_content(&self, maybe_new_content: &String) -> Result<bool> {
```
reset: We should pefer `&str` over `&String` for API ergonomic reasons

components/sup/src/manager/service/hooks.rs
```diff
@@ -97,19 +98,56 @@ pub trait Hook: fmt::Debug + Sized {
 
     fn new(service_group: &ServiceGroup, render_pair: RenderPair) -> Self;
 
-    /// Compile a hook into it's destination service directory.
-    fn compile(&self, ctx: &RenderContext) -> Result<()> {
+    /// Cryptographically hash the contents of the compiled hook
```
reset: Excellent job documenting these functions

components/sup/tests/utils/test_sup.rs
```diff
@@ -0,0 +1,200 @@
...
+    bin
+}
+
+impl TestSup {
+    /// Create a new `TestSup` that will listen on randomly-selected
+    /// ports for both gossip and HTTP requests so tests run in
+    /// parallel don't step on each other.
+    ///
+    /// See also `new`.
+    pub fn new_with_random_ports<R, O, P, S>(
```
reset: We may want to consider using an `AtomicI16` instead of random ports. We may hit a port that is in use by another test which will result in random failures in the build service and could increase over time as the test suite grows.

I might also recommend that we use a particular port range and document it so you don't start a service on a port that is in use by another service on your system?

reset:
@christophermaier this is an excellent test suite!

tenor-98857088

components/sup/tests/utils/test_sup.rs
christophermaier: I think `AtomicI16` is a nightly-only API; is that a dependency we want to take on?

I can cut down the range of ports to look at to the ones above 49151, which is where registered ports stop and dynamic ones start. That still gives us a 1 in ~16,000 chance of randomly selecting the same port. I could also try testing if the port is currently free by trying to create a `TcpListener`on that port, though that's got race conditions itself, and adds a bit of extra complexity.

Once atomic integers are standard, that sounds like a good way to go, though.

Any thoughts on that approach?

components/sup/src/manager/service/hooks.rs
christophermaier: Derp, yeah.

components/sup/tests/utils/test_sup.rs
reset: Ah yep, that's part of the nightly API and I didn't even see the header! `AtomicIsize` is available to us and should do the job just fine.

I'm cool with however you want to avoid the port conflict - a retry is just fine!

reset:
@thesentinels approve

tenor-22538047

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/256192080).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/256192080).

PR #2737, opened 2017-07-14
[hab,studio] Support artifact caching across Studios.

fnichol:
## [hab,studio] Support artifact caching across Studios.

tenor-101473861

This change adds a new capability to the Studio software which allows the downloaded Habitat artifacts (i.e. `*.hart` files) to be shared between different Studio instances or the between setup and tear downs of the same Studio instance. The sharing is accomplished by mounting in a common artifact cache directory into each Studio so that all artifact downloads get combined into one location for the benefit of all instances.

For non-root users using Linux and for Windows and Mac users, a directory of `$HOME/.hab/cache/artifacts` will now be created and mounted into the Studio's artifact cache directory which is always at `/hab/cache/artifacts`. For a root user using Linux, the host's system `/hab/cache/artifacts` directory will be mounted into the Studio--the assumption we make in Habitat is that the root user would have their own home directory configuration that is different from the system's.

The net effect of this change is that users of the Studio should see a substantial reduction in bandwidth usage when building software. A couple of things to note however:

* The install logic will still (as before) check the Depot to ensure that we still have the latest artifact cached. If a newer artifact exists in the Depot, it will still be downloaded. Therefore, as of this feature, you will still need internet access when running Studio builds. Any potential future "offline" feature would require this caching as a prerequisite which is why it has been implemented first.
* While an artifact may be pre-cached, the Studio code still needs to install it into the Studio instance, meaning that on-disk extraction and artifact verification is still being performed every time (as before).

Studio Usage Additions
----------------------

The following new flags and options are added to the `hab-studio` program which gets called directly on Linux when using `hab studio` or `hab pkg build` calls:

* `-N` Do not mount the source artifact cache path into the Studio (default: mount the path).
* `-a <ARTIFACT_PATH>` Sets the source artifact cache path (default: `/hab/cache/artifacts`). For non-root users the default will be `$HOME/.hab/cache/artifacts`.

As well, and to match the patterns already established in the Studio codebase, the following Studio-only environment variables are added:

* `ARTIFACT_PATH` - Sets the source artifact cache path (\`-a' option overrides). This is named to line up with the `SRC_PATH` environment variable already in use.
* `NO_ARTIFACT_PATH` - If set, do not mount the source artifact cache path (\`-N' flag overrides).

Notes
-----

Some additional call outs:

* An additional core users function named `get_primary_gid_for_user` was added to help set correct ownership details on newly created user cache directories. There appears to be no good Windows equivalent call and therefore `None` is returned which should be correct behavior.
* Now that `/hab/cache/artifacts` in a Studio is assumed to be an externally bind mounted directory we should no longer clear out this directory when preparing Docker images, the default Docker Studio image, etc. Otherwise the built up user cache will be cleared for no good reason. This is why there are some updates which remove `rm` statements or add them in other cases.
* The Rust code in Hab's "silently use sudo and check on user settings and env vars" `studio` module is stating to become less clear than I'd like. While I don't believe I have this code set up 100% ideally, it works for the moment and isn't worse than before. Just an area I'd like to keep an eye on.
* While the benefits of saving on re-downloads is great, we introduce a new resource issue for users to keep an eye on: the size of their `~/.hab/cache/artifacts` directory.

## [hab] Refactor Docker Studio setup code.

tenor-27372330

This refactoring attempts to clarify *what* the code is doing at a high level and less about *how*. Prior to this change the entire operation of checking if the Docker engine is alive, pulling images, verifying volume mounts, etc. was done in one function resulting in potentially many `docker` spawned processes. While the number of process spawns hasn't been reduced, the logical operations have been extracted into their own functions named after the goal they represent.

### DIFF
--- a/components/core/src/os/users/linux.rs
+++ b/components/core/src/os/users/linux.rs
@@ -41,6 +41,10 @@ pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
     linux_users::get_user_by_name(username).map(|u| PathBuf::from(u.home_dir()))
 }
 
+pub fn get_primary_gid_for_user(username: &str) -> Option<u32> {
+    linux_users::get_user_by_name(username).map(|u| u.primary_group_id())
+}
+
 pub fn root_level_account() -> String {
     "root".to_string()
 }
--- a/components/core/src/os/users/mod.rs
+++ b/components/core/src/os/users/mod.rs
@@ -18,11 +18,13 @@ mod windows;
 
 #[cfg(windows)]
 pub use self::windows::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user,
-                        get_current_username, get_current_groupname, root_level_account};
+                        get_current_username, get_current_groupname, get_primary_gid_for_user,
+                        root_level_account};
 
 #[cfg(not(windows))]
 pub mod linux;
 
 #[cfg(not(windows))]
 pub use self::linux::{get_uid_by_name, get_gid_by_name, get_effective_uid, get_home_for_user,
-                      get_current_username, get_current_groupname, root_level_account};
+                      get_current_username, get_current_groupname, get_primary_gid_for_user,
+                      root_level_account};
--- a/components/core/src/os/users/windows.rs
+++ b/components/core/src/os/users/windows.rs
@@ -61,7 +61,7 @@ pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
     unimplemented!();
 }
 
-pub fn get_primary_gid_for_user(username: &str) -> Option<u32> {
+pub fn get_primary_gid_for_user(username: &str) -> Option<String> {
     None
 }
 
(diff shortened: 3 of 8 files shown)


thesentinels:

Thank you for contributing!

reset:
@fnichol I can't wait for this

tenor-261839036

fnichol:
Rebased, now let's see if we're green…

fnichol:
Okay, one more change to make our `core` `filesystem::chown` only fire on non-Windows systems then we should be good.

reset:
@fnichol perfect job and thank you to @mwrock for the fixes for Windows to push this over the line!

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/256243355).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/256243355).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/256243355) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

PR #2773, opened 2017-07-18
Make job archive configuration more robust

christophermaier:
This fixes a few issues we ran into on a recent deploy of builder to production.

### DIFF
--- a/components/builder-jobsrv/habitat/default.toml
+++ b/components/builder-jobsrv/habitat/default.toml
@@ -16,8 +16,4 @@ connection_timeout_sec = 3600
 
 [archive]
 backend = "local"
-key = ""
-secret = ""
-endpoint = ""
-bucket = ""
-region = "us-east-1"
+local_dir = "/tmp"
--- a/components/builder-jobsrv/src/server/log_archiver/s3.rs
+++ b/components/builder-jobsrv/src/server/log_archiver/s3.rs
@@ -70,7 +70,11 @@ impl S3Archiver {
         // Parameterize this if anyone ends up needing V2 signatures
         let signature_type = Signature::V4;
         let final_endpoint = match config.endpoint {
-            Some(url) => Some(extern_url::Url::parse(url.as_str())?),
+            Some(url) => {
+                let url = extern_url::Url::parse(url.as_str())
+                    .expect("Invalid endpoint URL given");
+                Some(url)
+            },
             None => None,
         };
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
@thesentinels approve

tenor-36439629

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/254983420).

components/builder-jobsrv/habitat/default.toml
```diff
@@ -16,8 +16,4 @@ connection_timeout_sec = 3600
 
 [archive]
 backend = "local"
-key = ""
```
reset: The `default.toml` should be considered an "interface" and expose all configurable pieces. I would include these bits with a sensible default if possible and just a blank string when it's something required. In the future I'd really love for us to investigate writing something that allows us to describe our interface much better than TOML itself so we can validate config on the Habitat level instead of bogusly configuring a service

components/builder-jobsrv/habitat/default.toml
```diff
@@ -16,8 +16,4 @@ connection_timeout_sec = 3600
...
-secret = ""
-endpoint = ""
-bucket = ""
-region = "us-east-1"
+local_dir = "/tmp"
```
reset: This is the default that we were really getting at from the change at our last customer off-site. We should only write to a service's directory. The other part of this commit was making `local_dir` automatically populate which is why this tunable was removed. You can see those changes [here](https://github.com/habitat-sh/habitat/blob/master/components/builder-jobsrv/habitat/config/config.toml#L18)

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/254983420).

robbkidd:
@christophermaier I'm seeing issues with this "essentially a revert of d0345238".

```
builder-jobsrv.default(SV): Starting service as user=hab, group=hab
builder-jobsrv.default(O): Syntax errors while parsing TOML configuration file:
builder-jobsrv.default(O):
builder-jobsrv.default(O): duplicate field `local_dir` for key `archive`
```

I suspect because the line in `config/config.toml` setting local_dir to the svc_data_path was left in.

My current `builder-jobsrv/config/config.toml` has this rendering:

PR #2803, opened 2017-07-23
Set spec watcher back to 100ms poll

reset:
It looks like we had accidentally set the spec watcher back to a
1_000ms poll (1 second) from 100ms. This will revert that change
which appears to have been accidental

/cc @christophermaier I think this came through in an unrelated PR from you. Did you mean to change this?

### DIFF
--- a/components/sup/src/manager/spec_watcher.rs
+++ b/components/sup/src/manager/spec_watcher.rs
@@ -661,7 +661,7 @@ mod test {
         active_specs: HashMap<String, ServiceSpec>,
     ) -> Vec<SpecWatcherEvent> {
         let start = Instant::now();
-        let timeout = Duration::from_millis(1000);
+        let timeout = Duration::from_millis(100);
         while start.elapsed() < timeout {
             let events = watcher.new_events(active_specs.clone()).unwrap();
             if !events.is_empty() {


thesentinels:

Thank you for contributing!

christophermaier:
@reset See https://github.com/habitat-sh/habitat/pull/2729/commits/4734858f74003aefb0988627f05aeadb3f8a8b2e

reset:
@christophermaier what were some of the test failures we were seeing? We might want to bump this just for testing - the reason it was lower was so that the CLI didn't feel sluggish when executing commands that waited on these events

christophermaier:
@reset They were literally failures because the timeout was too short; the expected change wasn't observed in the allotted time, and so the tests were failing. This didn't happen all the time, though I was seeing a decent amount of time in my VM and in Travis. Since we check every millisecond anyway, raising the upper bound shouldn't really make the tests run longer; it just gives us more buffer on slower machines.

Also, this is just in a test helper, and not anything in real code, so the behavior of the CLI will be unchanged.

reset:
@christophermaier oh my mistake, I thought this was a different poll interval 😂

christophermaier:
@reset No worries 😄

PR #2809, opened 2017-07-24
[studio] Kill Launcher process when shutting down Studio.

fnichol:
This change follows the introduction of the Launcher in the 0.26.0
release. Prior to this commit the Studio cleanup logic would attempt to
kill the Supevisor but with the introduction of the Launcher, this would
leave a `hab-launch` process behind and running in Linux Studios. With
this change, the process ID of the Launcher is recorded in a file and is
checked on Studio shutdown for which process to kill. As the Launcher is
now the parent process of the Supervisor, the full chain of processes
should terminate as expected.

When making this change, I made a couple of refactoring around the
`cleanup_studio()` function including extracting the 3 cleanup tasks
into their own functions and fixed on bug in `studio rm` where a local
variable `$v` was missing.

### DIFF
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -119,6 +119,7 @@ sup-run() {
   echo "--> Launching the Habitat Supervisor in the background..."
   echo "    Running: hab sup run \$*"
   hab sup run \$* > /hab/sup/default/sup.log &
+  echo \$! > /hab/sup/default/launch.pid
   echo "    * Use 'hab svc start' & 'hab svc stop' to start and stop services"
   echo "    * Use 'sup-log' to tail the Supervisor's output (Ctrl+c to stop)"
   echo "    * Use 'sup-term' to terminate the Supervisor"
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/studio/libexec/hab-studio-type-default.sh
```diff
@@ -119,6 +119,7 @@ sup-run() {
   echo "--> Launching the Habitat Supervisor in the background..."
   echo "    Running: hab sup run \$*"
   hab sup run \$* > /hab/sup/default/sup.log &
+  echo \$! > /hab/sup/default/launch.pid
```
reset: Super clean, I like that we didn't modify the launcher/supervisor code at all and instead wrote the pid in the studio for it's purposes

fnichol: Haha, I was worried until I remembered what Bash gave us. This ends up being the critical bit so the behavior didn't need to change much.

nellshamrell:
:shipit:

fnichol:
@nellshamrell I found one small thing in the cleanup that might have explained why needed to remove the device files afterwards--we were calling `trap cleanup_studio EXIT` in `rm_studio()` which would have been called *after* the `rm -rf /hab/studios/<dir>`. I tried to double check and verify this which is why I took those 2 lines out. Can't wait for this next release

reset:
@thesentinels approve

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/257100537).

mwrock:
not unless you create with `start-process`  and `-passthru` the process object. Then you can inspect that object for the id. And DAMN, I just assumud the LOCK file changed and did not catch that you were creating that. I need to ammend that pr now. Glad you brought this up!

PR #2811, opened 2017-07-25
Auto update channel default

reset:
This will resolve an issue where the new `--auto-update` flag was
incorrectly automatically retrieving updates from the unstable
channel by default. This change should additionally prevent mistakes
like this from escaping in the future by explicitly requiring a channel
and more clearly stating how our default channel is calculated.

tenor-215320992

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -37,10 +37,11 @@ use std::str::FromStr;
 use ansi_term::Colour::{Red, Yellow};
 use clap::{App, ArgMatches};
 use common::ui::UI;
-use hcore::env as henv;
+use hcore::channel;
 use hcore::crypto::{self, default_cache_key_path, SymKey};
 #[cfg(windows)]
 use hcore::crypto::dpapi::encrypt;
+use hcore::env as henv;
 use hcore::package::{PackageArchive, PackageIdent};
 use hcore::service::{ApplicationEnvironment, ServiceGroup};
 use hcore::url::{DEFAULT_DEPOT_URL, DEPOT_URL_ENVVAR};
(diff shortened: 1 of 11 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/main.rs
```diff
@@ -163,7 +164,7 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
             (@arg ENVIRONMENT: --environment -e +takes_value requires[APPLICATION]
                 "Environment name; [default: not set].")
             (@arg CHANNEL: --channel +takes_value
-                "Receive package updates from the specified release channel")
+                "Receive package updates from the specified release channel [default: stable]")
```
fnichol: 👍

components/sup/src/manager/mod.rs
```diff
@@ -116,7 +116,7 @@ pub struct ManagerConfig {
     pub auto_update: bool,
     pub eventsrv_group: Option<ServiceGroup>,
     pub update_url: String,
-    pub update_channel: Option<String>,
+    pub update_channel: String,
```
fnichol: Yeah, I like this better--forces us to determine the defaults upstream and keeps the internals simpler.

components/sup/src/main.rs
reset: I'm pretty sure this was an intermediary step in the first place before we had channels pipe through in the UI of Builder just so we could avoid confusion

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/257121625).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/257121625).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/257121625) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

PR #2826, opened 2017-07-26
store launcher pid in supervisor lock file

### LINKED ISSUE #2822: Launcher should obtain lock instead of Supervisor
The Supervisor on startup obtains a process lock by reading/writing to `/hab/sup/{name}/LOCK`. Since the root of the Superivsor's process tree is now the Launcher, the code to read/write/check for the process lock should instead be in the Launcher's crate.

This should immediately solve the issue with the `hab sup term` and the term function in the studio where it terminates the Supervisor but it is then immediately restarted.

mwrock:
This fixes #2822 and stores the launcher pid instead of the supervisor pid in the lock file. It also effectively reverts the recent studio PRs which this change renders unnecessary. Further, this removes the `term` command from windows since the current implementation is a "no-op" on windows.

A note on the implementation. #2822 specifies having the launcher "own" the lock. I started to go down the path of having the launcher crate maintain the file itself but it proved "awkward". The awkwardness was primarily around tracking the `state_path`. The supervisor needs this for its own artifact and the launcher would need that same path to write the lock file. While itr could look for a `--name` arg in the arguments to build the path, parsing the arguments here is ugly without clap and the supervisor still needs that path for its `term` command. In the end, sending the launcher pid to the supervisor ended up being MUCH simpler.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -37,6 +37,7 @@ use butterfly::trace::Trace;
 use butterfly::server::timing::Timing;
 use butterfly::server::Suitability;
 use hcore::crypto::{default_cache_key_path, SymKey};
+use hcore::env;
 use hcore::fs::FS_ROOT_PATH;
 use hcore::service::ServiceGroup;
 use hcore::os::process::{self, Signal};
@@ -60,6 +61,7 @@ use http_gateway;
 
 const MEMBER_ID_FILE: &'static str = "MEMBER_ID";
 const PROC_LOCK_FILE: &'static str = "LOCK";
+const LAUNCHER_PID_ENV: &'static str = "HAB_LAUNCHER_PID";
 
 static LOGKEY: &'static str = "MR";
 
@@ -1015,7 +1017,11 @@ where
             .as_ref(),
     ) {
         Ok(mut file) => {
-            match write!(&mut file, "{}", process::current_pid()) {
+            let pid = match env::var(LAUNCHER_PID_ENV) {
+                Ok(pid) => pid.parse::<u32>().expect("Unable to parse launcher pid"),
+                Err(_) => process::current_pid(),
+            };
+            match write!(&mut file, "{}", pid) {
                 Ok(()) => Ok(()),
                 Err(err) => {
                     Err(sup_error!(
--- a/components/launcher-protocol/src/lib.rs
+++ b/components/launcher-protocol/src/lib.rs
@@ -26,6 +26,7 @@ pub use message::supervisor::*;
 use message::net::*;
 
 pub const LAUNCHER_PIPE_ENV: &'static str = "HAB_LAUNCHER_PIPE";
+pub const LAUNCHER_PID_ENV: &'static str = "HAB_LAUNCHER_PID";
 /// Process exit code from Supervisor which indicates to Launcher that the Supervisor
 /// ran to completion with a successful result. The Launcher should not attempt to restart
 /// the Supervisor and should exit immediately with a successful exit code.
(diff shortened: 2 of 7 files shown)


thesentinels:

Thank you for contributing!

fnichol:
That's a clever way to solve the problem--I was thinking along the same lines ("the `state_path` is going to complicate things") but sending the WHAT to the sup so it continues to determine the WHERE was an interesting way to slice it.

fnichol:
@mwrock is it possibly you forgot to pass a file through `rustfmt`? I think that's why the Travis job failed

mwrock:
yeah my rustfmt environment looks to be hosed. Working to fix that now.

components/sup/src/manager/mod.rs
```diff
@@ -1015,7 +1017,11 @@ where
             .as_ref(),
     ) {
         Ok(mut file) => {
-            match write!(&mut file, "{}", process::current_pid()) {
+            let pid = match env::var(launcher_protocol::LAUNCHER_PID_ENV) {
```
reset: This is a really clever way to get this done. Very well thought out IMO to the point where the Supervisor can still be started by an older launcher

components/sup/Cargo.toml
```diff
@@ -28,6 +28,7 @@ habitat_core = { path = "../core" }
 habitat_depot_client = { path = "../builder-depot-client" }
 habitat-eventsrv-client = { path = "../eventsrv-client" }
 habitat-launcher-client = { path = "../launcher-client" }
+habitat-launcher-protocol = { path = "../launcher-protocol" }
```
reset: Instead of taking a direct dep on `habitat-launcher-protocol` I would instead re-pub the const that you want inside the client. This crate is supposed to be "hidden" to the supervisor (I should probably document that 😉)

You can see an example [here](https://github.com/habitat-sh/habitat/blob/master/components/launcher-client/src/lib.rs#L23)

mwrock: ok. I had debated that option too and went with this in an effort to keep the const in one place. Will update.

components/sup/src/manager/mod.rs
fnichol: That was exactly what I was thinking through--is there a release order to this? (I don't think it matters a lot, as long as sup/studio go together)

mwrock: yeah even if they were separate we'd be in the same place we are right now.

components/sup/Cargo.toml
reset: @mwrock I hear you. My goal here is to appropriate a pattern through our crates where the consumer of a client doesn't need to understand the protocol at all

It's worth noting that it's still in memory only in one place for any reader who is curious

Cargo.lock
```diff
@@ -1182,6 +1182,7 @@ dependencies = [
  "glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "habitat-eventsrv-client 0.0.0",
  "habitat-launcher-client 0.0.0",
+ "habitat-launcher-protocol 0.0.0",
```
reset: I think you need to re-run `cargo check` or something similar to regenerate this lock

mwrock: argh...right

reset:
@thesentinels approve

tenor-190264982

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/257911022).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/257911022).

PR #2829, opened 2017-07-26
Install initial habitat packages from bootstrap bundle

### LINKED ISSUE #2732: Stage 0 Bootstrap Improvement
Currently the Terraform bootstrap of the hab depot depends on hab packages getting installed from the depot itself which does not work.  We should instead bootstrap the required packages from an alternate location (eg S3).  There should also be a process to update the S3 location when the impacted packages are promoted to stable.  The new bootstrap should work for both prod and acceptance.

In addition, we should have a similar process for the bootstrap config TOML files that are currently stored on the datastore node.

christophermaier:
Rebuilding a Habitat Builder server instance in production is
currently harder than it needs to be, mainly because we've been using
Builder to bootstrap! If you need to replace a server in production,
chances are that the Builder isn't gonna be there for you. What to do?

Well, we can have a bundle of all the artifacts we need to get an
entire Builder infrastructure that's *not* coming from Builder, and
install from *that*. The solution has two parts. First, the
`create_bootstrap_bundle.sh` script is used to download all the
packages from a (currently functional!) Builder instance, `tar` them
up, and put them up into S3. This script is something that we can run
periodically (possibly with each release of the supervisor). All
bootstrap bundles created this way stay in S3, so it's theoretically
possible to stand up a Builder infrastructure from code at different
points in time, should we ever need / want to do such a
thing. Crucially, the script generates a manifest file for each
bundle, and also uploads this to S3, with a copy saved to the file
`LATEST`. All we need to do is grab the `LATEST` file, and it'll tell
us which tarball we should grab to bootstrap from.

The second part of the solution is to tweak our Terraform setup to
install packages from this tarball; that's largely the job of the
`install_base_packages.sh` script. It grabs the appropriate tarball on
every server instance and installs the desired packages into the local
`/hab/cache/artifacts` directory. With those packages in place, we can
run `hab svc load` for all our services without having to talk to
another Builder instance at all.

The previous `bootstrap.sh` script was removed, and the user creation
tasks it did were moved into the `filesystem.sh` script, which was
renamed to `foundation.sh` to reflect its more general purpose. This
was done to consolidate scripts.

Fixes #2732


thesentinels:

Thank you for contributing!

terraform/instances.tf
```diff
@@ -403,11 +449,14 @@ resource "aws_instance" "scheduler" {
...
   provisioner "remote-exec" {
     inline = [
+      "chmod +x /tmp/install_base_packages.sh",
+      "sudo /tmp/install_base_packages.sh core/builder-scheduler",
+
       "sudo mv /home/ubuntu/hab-sup.service /etc/systemd/system/hab-sup.service",
       "sudo systemctl daemon-reload",
       "sudo systemctl start hab-sup",
       "sudo systemctl enable hab-sup",
-      "sudo hab svc load core/builder-scheduler --group ${var.env} --bind router:builder-router.${var.env} --bind datastore:builder-datastore.${var.env} --bind depot:builder-api.${var.env} --strategy at-once --url ${var.depot_url} --channel ${var.release_channel}"
```
reset: This does require a bind to the api server. I think that was added relatively recently so a build of scheduler might help :)

terraform/scripts/create_bootstrap_bundle.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
+
+# Create a tarball of all the Habitat artifacts needed to run the
```
reset: There is so much polish in here that I would have never applied. So good :)

terraform/instances.tf
christophermaier: Whoa, thanks for catching that. That change shouldn't have been there in the first place.

terraform/scripts/create_bootstrap_bundle.sh
```diff
@@ -0,0 +1,181 @@
...
+########################################################################
+# Preliminaries, Helpers, Constants
+
+find_if_exists() {
+    command -v ${1} || { log "Required utility '${1}' cannot be found!  Aborting."; exit 1; }
+}
+
+# These are the key utilities this script uses. If any are not present
+# on the machine, the script will exit.
+aws=$(find_if_exists aws)
```
fnichol: That's a nice pattern, glad to see this!

terraform/scripts/create_bootstrap_bundle.sh
```diff
@@ -0,0 +1,181 @@
...
+                  core/builder-api-proxy
+                  core/builder-admin
+                  core/builder-admin-proxy
+                  core/builder-datastore
+                  core/builder-jobsrv
+                  core/builder-originsrv
+                  core/builder-router
+                  core/builder-scheduler
+                  core/builder-sessionsrv
+                  core/builder-worker)
```
fnichol: Bonus points for lexically sorting--this'll pay off

terraform/scripts/create_bootstrap_bundle.sh
```diff
@@ -0,0 +1,181 @@
...
+
+# Because Habitat may have already run on this system, we'll want to
+# make sure we start in a pristine environment. That way, we can just
+# blindly copy everything in ${sandbox_dir}/hab/cache/artifacts, confident
+# that those artifacts are everything we need, and no more.
+sandbox_dir=${this_bootstrap_bundle}
+mkdir ${sandbox_dir}
+log "Using ${sandbox_dir} as the Habitat root directory"
+
+for package in "${sup_packages[@]}" "${builder_packages[@]}"
```
fnichol: If we end up running this "raw" on a system we'll want to make sure we have Bash 4.x+ (IIRC) to get the array behavior. I don't think this'll be an issue, but that's come up with our build program in the past when running out of Studio.

reset:
tenor-262180878

terraform/scripts/install_base_packages.sh
```diff
@@ -0,0 +1,125 @@
...
+# script with the argument "core/builder-worker" (You could even just
+# pass "builder-worker"; both are recognized.)
+#
+# Multiple packages can be specified.
+#
+# Do *not* provide version or release information.
+declare -a services_to_install='()'
+if [ "$#" > 0 ]
+then
+    # Bash 4 adds the command to this array; let's slice it off
```
fnichol: Well, that's hilarious--but I guess it makes sense

terraform/scripts/create_bootstrap_bundle.sh
christophermaier: Indeed!

PR #2842, opened 2017-07-28
use 0.26.1 to build windows harts and then use current studio to test that it works

mwrock:
This does 2 things:

* Uses 0.26.1 to build the windows harts in appveyor because 0.27.0 will panic. We should use `latest` after 0.27.1 releases.
* Once the new studio is built. use that new studio to build the rest of the components. This should have been working before but had a bug because the built executable was copies one level too shallow. This should prevent a panicking studio from releasing again.

Signed-off-by: Matt Wrock <matt@mattwrock.com>

PR #2843, opened 2017-07-28
Honor channels and an alternate depot URL when exporting packages

reset:
* Exporters will now by default be downloaded from the stable
  channel before running
* Add `--channel` flag to `pkg export`
* Add `--url` flag to `pkg export`

tenor-201683009

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 #![cfg_attr(feature="clippy", feature(plugin))]
 #![cfg_attr(feature="clippy", plugin(clippy))]
 
@@ -324,10 +325,27 @@ fn sub_pkg_exec(m: &ArgMatches, cmd_args: Vec<OsString>) -> Result<()> {
 }
 
 fn sub_pkg_export(ui: &mut UI, m: &ArgMatches) -> Result<()> {
-    let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?; // Required via clap
-    let format = &m.value_of("FORMAT").unwrap(); // Required via clap
+    let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?;
+    let format = &m.value_of("FORMAT").unwrap();
+    let env_or_default = henv::var(DEPOT_URL_ENVVAR).unwrap_or(DEFAULT_DEPOT_URL.to_string());
+    let url = m.value_of("DEPOT_URL").unwrap_or(&env_or_default);
+    let channel = m.value_of("CHANNEL")
+        .and_then(|c| Some(c.to_string()))
+        .unwrap_or(channel::default());
+    let hab_url = m.value_of("HAB_DEPOT_URL").unwrap_or(&env_or_default);
+    let hab_channel = m.value_of("HAB_CHANNEL")
+        .and_then(|c| Some(c.to_string()))
+        .unwrap_or(channel::default());
     let export_fmt = command::pkg::export::format_for(ui, &format)?;
-    command::pkg::export::start(ui, &ident, &export_fmt)
+    command::pkg::export::start(
+        ui,
+        &url,
+        &channel,
+        &hab_url,
+        &hab_channel,
+        &ident,
+        &export_fmt,
+    )
 }
 
 fn sub_pkg_hash(m: &ArgMatches) -> Result<()> {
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

adamhjk:
So the actual changes to the studio are already done? This was literally just not piped through?

tenor-196572730

reset:
@adamhjk that's what it looks like 😢

mwrock:
we can ignore the appveyor breaks until https://github.com/habitat-sh/habitat/pull/2842 gets through

sdelano:
@reset does this mean that if I do `hab pkg export ...` that I'll also have to specify `--url` and `--channel`, or will that get picked up by default from what I used to install `hab`?

adamhjk:
tl;dr - this commit makes it so you get the stable channel for the supervisor, but what you actually want is two flags that pipe down through the shell code:

`--channel FOO`, which selects the channel for the package you want exported

`--supervisor-channel FOO`, whihc selects the channel for the supervisor version in the export.

adamhjk:
@sdelano that's a good suggestiont o default to whatever the hab you're running is, tho

sdelano:
@adamhjk if I'm reading the code correctly though the URL and CHANNEL params get passed to the install portion of `hab pkg export`, which actually installs the `hab-pkg-$type` package.

The origination of this PR is that:

> I install `hab` from stable. I run `hab package export`, but don't have the `hab-pkg-dockerizer` package installed, so when hab goes to install it I actually get an unstable version that doesn't correspond to the version of `hab` I have.

reset:
Another commit is going to be pushed up which exposes two channel and two URL flags. This will allow you to specify the toolchain's channel/url and your package's channel/url.

You should very rarely use the one to pull in an unstable toolchain but you _may_ want to use the one for your package itself

reset:
@adamhjk @sdelano we're good to go here

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/258380542).

reset:
@fnichol and I discussed overall improvements to some of our studio/exporter logic which will make this easier to maintain. We'll tackle those in another PR

components/hab/src/main.rs
```diff
@@ -324,10 +325,27 @@ fn sub_pkg_exec(m: &ArgMatches, cmd_args: Vec<OsString>) -> Result<()> {
...
+    let format = &m.value_of("FORMAT").unwrap();
+    let env_or_default = henv::var(DEPOT_URL_ENVVAR).unwrap_or(DEFAULT_DEPOT_URL.to_string());
+    let url = m.value_of("DEPOT_URL").unwrap_or(&env_or_default);
+    let channel = m.value_of("CHANNEL")
+        .and_then(|c| Some(c.to_string()))
+        .unwrap_or(channel::default());
+    let hab_url = m.value_of("HAB_DEPOT_URL").unwrap_or(&env_or_default);
+    let hab_channel = m.value_of("HAB_CHANNEL")
+        .and_then(|c| Some(c.to_string()))
+        .unwrap_or(channel::default());
```
sdelano: `channel::default()` actually looks like it'll look to `DEPOT_CHANNEL_ENVVAR` first and then eventually `"stable"` [link](https://github.com/habitat-sh/habitat/blob/master/components/core/src/channel.rs#L29). I think that makes sense for `channel` but for `hab_channel` do we want to default to stable here or go through that extra layer of the other envvar?

reset: I think it's appropriate for here, yeah. The env vars are just to establish an escape hatch to override the default values. They're largely supposed to be used from within a `.studiorc` for dev purposes.

We could of course create another escape hatch for just this command but I'd rather just suggest people to drive the CLI with the flags

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/258380542).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/258380542) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/258380542).

components/hab/src/main.rs
sdelano: Sounds good to me!

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/258380542).

PR #2847, opened 2017-07-28
Hab plan init rewrite with autodetection

eeyun:
Signed-off-by: eeyun <ihenry@chef.io>

This pr is the first iteration of rewriting the `hab plan init` subcommand to start using auto detection and push new users towards leveraging the scaffolding.

A reminder that this pr consists of the first two commits that move towards utilizing this autodiscovery behavior out of the box for new users.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -291,10 +291,11 @@ pub fn get() -> App<'static, 'static> {
                     Environment variables (those starting with 'pkg_') that are set will be used \
                     in the generated plan")
                 (aliases: &["i", "in", "ini"])
-                (@arg PKG_NAME: +takes_value "Name for the new app.")
+                (@arg PKG_NAME: +takes_value "Name for the new app")
                 (@arg ORIGIN: --origin -o +takes_value "Origin for the new app")
-                (@arg NO_CALLBACKS: --nocallbacks -f
-                    "Do not include callback functions in template")
+                (@arg WITH_DOCS: --withdocs -d "Include plan options documentation")
+                (@arg WITH_CALLBACKS: --withcallbacks -f
+                    "Include callback functions in template")
             )
         )
         (@subcommand ring =>
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -293,9 +293,13 @@ pub fn get() -> App<'static, 'static> {
                 (aliases: &["i", "in", "ini"])
                 (@arg PKG_NAME: +takes_value "Name for the new app")
                 (@arg ORIGIN: --origin -o +takes_value "Origin for the new app")
-                (@arg WITH_DOCS: --withdocs -d "Include plan options documentation")
-                (@arg WITH_CALLBACKS: --withcallbacks -f
+                (@arg WITH_DOCS: --withdocs "Include plan options documentation")
+                (@arg WITH_CALLBACKS: --withcallbacks
                     "Include callback functions in template")
+                (@arg WITH_ALL: --withall
+                    "Generate omnibus plan with all available plan options")
+                (@arg SCAFFOLDING: --scaffolding -s +takes_value
+                    "Specify explicit scaffolding type for your app")
             )
         )
         (@subcommand ring =>
--- a/components/hab/src/command/plan/init.rs
+++ b/components/hab/src/command/plan/init.rs
@@ -103,34 +103,26 @@ pub fn start(
 
     // We want to render the configured variables.
     if with_all || scaffold.is_none() {
-        let rendered_plan = try!(handlebars.template_render(FULL_PLAN_TEMPLATE, &data));
-        try!(create_with_template(
-            ui,
-            &format!("{}/plan.sh", root),
-            &rendered_plan,
-        ));
+        let rendered_plan = handlebars.template_render(FULL_PLAN_TEMPLATE, &data)?;
+        create_with_template(ui, &format!("{}/plan.sh", root), &rendered_plan)?;
     } else {
-        let rendered_plan = try!(handlebars.template_render(DEFAULT_PLAN_TEMPLATE, &data));
-        try!(create_with_template(
-            ui,
-            &format!("{}/plan.sh", root),
-            &rendered_plan,
-        ));
+        let rendered_plan = handlebars.template_render(DEFAULT_PLAN_TEMPLATE, &data)?;
+        create_with_template(ui, &format!("{}/plan.sh", root), &rendered_plan)?;
     }
-    try!(ui.para(
+    ui.para(
         "The `plan.sh` is the foundation of your new habitat. You can \
         define core metadata, dependencies, and tasks.",
-    ));
-    let rendered_default_toml = try!(handlebars.template_render(DEFAULT_TOML_TEMPLATE, &data));
-    try!(create_with_template(
+    )?;
+    let rendered_default_toml = handlebars.template_render(DEFAULT_TOML_TEMPLATE, &data)?;
+    create_with_template(
         ui,
         &format!("{}/default.toml", root),
         &rendered_default_toml,
-    ));
-    try!(ui.para(
+    )?;
+    ui.para(
         "The `default.toml` allows you to declare default values for `cfg` prefixed
         variables.",
-    ));
+    )?;
 
     let config_path = format!("{}/config/", root);
     match Path::new(&config_path).exists() {
@@ -148,10 +140,10 @@ pub fn start(
             create_dir_all(&config_path)?;
         }
     };
-    try!(ui.para(
+    ui.para(
         "The `config` directory is where you can set up configuration files for your app. \
         They are influenced by `default.toml`.",
-    ));
+    )?;
 
     let hooks_path = format!("{}/hooks/", root);
     match Path::new(&hooks_path).exists() {
@@ -169,22 +161,22 @@ pub fn start(
             create_dir_all(&hooks_path)?;
         }
     };
-    try!(ui.para(
+    ui.para(
         "The `hooks` directory is where you can create a number of automation hooks into \
         your habitat.",
-    ));
+    )?;
 
-    try!(ui.para(
+    ui.para(
         "For more information on any of the generated files: \
         https://www.habitat.sh/docs/reference/plan-syntax/#basic-settings \
         https://www.habitat.sh/docs/reference/plan-syntax/#runtime-configuration-settings \
         https://www.habitat.sh/docs/reference/plan-syntax/#hooks \
         https://www.habitat.sh/docs/reference/plan-syntax/#callbacks",
-    ));
+    )?;
 
     render_ignorefile(ui, &root)?;
 
-    try!(ui.end("An abode for your code has been initialized!"));
+    ui.end("An abode for your code has been initialized!")?;
     Ok(())
 }
 
--- a/components/hab/src/command/plan/init.rs
+++ b/components/hab/src/command/plan/init.rs
@@ -20,6 +20,7 @@ use std::path::Path;
 use std::collections::HashMap;
 
 use handlebars::Handlebars;
+use hcore::package::PackageIdent;
 
 use common::ui::{UI, Status};
 use error::Result;
@@ -46,6 +47,8 @@ pub fn start(
     origin: String,
     with_docs: bool,
     with_callbacks: bool,
+    with_all: bool,
+    scaffolding_ident: Option<PackageIdent>,
     maybe_name: Option<String>,
 ) -> Result<()> {
     ui.begin("Constructing a cozy habitat for your app...")?;
@@ -75,6 +78,15 @@ pub fn start(
     let mut data = HashMap::new();
     data.insert("pkg_name".to_string(), name);
     data.insert("pkg_origin".to_string(), origin);
+
+    let scaffold = match scaffolding_ident {
+        Some(ident) => Some(data.insert(
+            "scaffolding_ident".to_string(),
+            ident.to_string(),
+        )),
+        None => None,
+    };
+
     if with_callbacks {
         data.insert("with_callbacks".to_string(), "true".to_string());
     }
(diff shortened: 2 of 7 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/plan/init.rs
```diff
@@ -83,25 +102,35 @@ pub fn start(
...
-        define core metadata, dependencies, and tasks. More documentation here: \
-        https://www.habitat.sh/docs/reference/plan-syntax/",
-    )?;
-
-    let rendered_default_toml = handlebars.template_render(DEFAULT_TOML_TEMPLATE, &data)?;
-    create_with_template(
+        define core metadata, dependencies, and tasks.",
+    ));
+    let rendered_default_toml = try!(handlebars.template_render(DEFAULT_TOML_TEMPLATE, &data));
+    try!(create_with_template(
```
reset: We actually switched large portions of our codebase from `try!` to  the `?` syntax. They do the same thing but one is easier on our line length ;)

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+        try!(ui.para(
+            "For now, we'll generate a plan with all of the available plan options and callbacks. \
+            For more documentation on plan options try passing --withdocs or visit \
+            https://www.habitat.sh/docs/reference/plan-syntax/",
+        ));
+        Ok(None)
+    }
+}
+
+fn is_project_ruby(path: &Path) -> bool {
```
reset: I love the simplicity of these functions! I can't believe how easy it is to identify what a project is ;)

reset: When you're dealing with filepaths as a function argument it's best to use a generic representation instead of a ref to `Path` or `PathBuf`. Here is an example:

```
fn is_project_ruby<T>(path: T> -> bool where T: AsRef<Path> {
    if path.as_ref().join("Gemfile).is_file() {
        return true;
    }
    // ...
}
```

I would recommend changing the functions in this file to leverage this approach. It makes the function easier to deal with for other developers. With this representation you can pass any thing which implements `AsRef<Path>` which includes, `str`, `&str`, `String`, `&String`, `Path`, `&Path`, `PathBuf`, `&PathBuf`, `OsStr`, `&OsStr`, and more

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+    if path.join("Gemfile").is_file() {
+        return true;
+    }
+
+    if let Some(parent) = path.parent() {
```
reset: Why are we checking if the parent is a project of `X`? I would think we would just want to check the target dir, no?

components/hab/src/main.rs
```diff
@@ -46,7 +46,9 @@ use hcore::package::PackageIdent;
 use hcore::service::ServiceGroup;
 use hcore::url::{DEFAULT_DEPOT_URL, DEPOT_URL_ENVVAR};
 
-use hab::{analytics, cli, command, config, AUTH_TOKEN_ENVVAR, ORIGIN_ENVVAR, PRODUCT, VERSION};
+use hab::{analytics, cli, command, config, scaffolding, AUTH_TOKEN_ENVVAR, ORIGIN_ENVVAR, PRODUCT,
+          VERSION};
+// the above will include scaffolding
```
reset: This is kind of a given, the alive above it says so! :)

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+use hcore::package::PackageIdent;
+use common::ui::{UI, Status};
+
+// Check to see if the --scaffolding passed matches available core scaffolding
+// If not check if we've been given a pkg ident for a custom scaffolding
+pub fn scaffold_check(ui: &mut UI, maybe_scaffold: Option<&str>) -> Result<Option<PackageIdent>> {
+    match maybe_scaffold {
+        Some(scaffold) => {
+            init();
+            match scaffold {
```
reset: I would downcase this before attempting to match so you don't have a case sensitive match

```
scaffold.to_lowercase()
```

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+                "core/scaffolding-ruby" |
+                "ruby" |
+                "rails" |
+                "RUBY" |
+                "ruby-scaffolding" => {
+                    let ident = PackageIdent::from_str("core/scaffolding-ruby")?;
```
reset: It's safe to `unwrap()` here. We actually *want* to assert here if something is wrong

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+                }
+                _ => {
+                    let ident = PackageIdent::from_str(scaffold)?;
+                    let _ = ui.status(Status::Using, &format!("custom scaffolding: '{}'", ident));
+                    Ok(Some(ident))
+                }
+            }
+        }
+        // If nothing was passed try to autodiscover the codebase
+        // I'm not saying it's magic. But, MAGIC.
```
reset: tenor-66995151

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    if let Some(parent) = path.parent() {
+        return is_project_node(&parent);
+    }
+
+    return false;;
+}
+
+fn project_uses_gb(dir: &Path) -> io::Result<(bool)> {
```
reset: You just want `io::Result<bool>` here, it doesn't need to be in a tuple

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+            } else if path.is_file() {
+                if path.extension().unwrap() == "go" {
+                    return Ok(true);
+                }
+            } else {
+                return Ok(false);
+            }
+        }
+    }
+    return Ok(false);;
```
reset: Extra `;`, I didn't even know this would compile!

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+
+fn project_uses_gb(dir: &Path) -> io::Result<(bool)> {
+    if dir.is_dir() {
+        for entry in fs::read_dir(dir)? {
+            let entry = entry?;
+            let path = entry.path();
+            if path.is_dir() {
+                project_uses_gb(&path)?;
+            } else if path.is_file() {
+                if path.extension().unwrap() == "go" {
```
reset: If a file doesn't have an extension this will panic. I would write this as

```
if let Some(ext) == path.extension() {
    if ext == "go" {
        return Ok(true)
    }
}
```

components/hab/src/scaffolding.rs
eeyun: Oh damn, I had no idea, that's super useful!

components/hab/src/main.rs
eeyun: Yepppp thats leftover from my pseudo code. Good catch!

components/hab/src/scaffolding.rs
eeyun: Ah, I had these unrwapped before and took them out, will revert!

components/hab/src/scaffolding.rs
eeyun: In regards to the simplicity. Yeah it's super weird, the bulk of the work here is obviously NOT the actual validation of the project type but all the other stuff needed for the validation to occur. These functions might get fleshed out later to be a little smarter but for now these are the validations that @fnichol had agreed on.

components/hab/src/scaffolding.rs
eeyun: Yep, we only want to check the current, I think I left these in there from an alternative implementation I was thinking through. Happy to get rid of unnecessary nonsense!

components/hab/src/scaffolding.rs
eeyun: IT ME

components/hab/src/cli.rs
```diff
@@ -291,10 +291,15 @@ pub fn get() -> App<'static, 'static> {
                     Environment variables (those starting with 'pkg_') that are set will be used \
                     in the generated plan")
                 (aliases: &["i", "in", "ini"])
-                (@arg PKG_NAME: +takes_value "Name for the new app.")
+                (@arg PKG_NAME: +takes_value "Name for the new app")
                 (@arg ORIGIN: --origin -o +takes_value "Origin for the new app")
-                (@arg NO_CALLBACKS: --nocallbacks -f
-                    "Do not include callback functions in template")
+                (@arg WITH_DOCS: --withdocs "Include plan options documentation")
```
fnichol: Think you could make this `--with-docs` rather than `--withdocs`? That will match some of the other CLI arguments in Habitat like in the [Supervisor](https://github.com/habitat-sh/habitat/blob/182ef7b4035aea078ad98018257e5190f759f18b/components/sup/src/main.rs#L197). You should be able to do this with `--("with-docs")` using the clap macro DSL.

components/hab/src/cli.rs
```diff
@@ -291,10 +291,15 @@ pub fn get() -> App<'static, 'static> {
...
+                (@arg WITH_CALLBACKS: --withcallbacks
```
fnichol: `--with-callbacks`?

components/hab/src/command/plan/init.rs
```diff
@@ -143,15 +163,20 @@ pub fn start(
...
-               with info here: https://www.habitat.sh/docs/reference/plan-syntax/#hooks",
+        your habitat.",
+    )?;
+
+    ui.para(
+        "For more information on any of the generated files: \
+        https://www.habitat.sh/docs/reference/plan-syntax/#basic-settings \
+        https://www.habitat.sh/docs/reference/plan-syntax/#runtime-configuration-settings \
+        https://www.habitat.sh/docs/reference/plan-syntax/#hooks \
+        https://www.habitat.sh/docs/reference/plan-syntax/#callbacks",
```
fnichol: I much prefer the output containing URL links to being embedding in the `plan.sh`. Nicely done!

components/hab/src/main.rs
```diff
@@ -369,9 +370,21 @@ fn sub_pkg_hash(m: &ArgMatches) -> Result<()> {
 
 fn sub_plan_init(ui: &mut UI, m: &ArgMatches) -> Result<()> {
     let name = m.value_of("PKG_NAME").map(|v| v.into());
-    let origin = origin_param_or_env(&m)?;
-    let include_callbacks = !m.is_present("NO_CALLBACKS");
-    command::plan::init::start(ui, origin, include_callbacks, name)
+    let origin = try!(origin_param_or_env(&m));
```
fnichol: Perhaps use the `origin_param_or_env(&m)?` form?

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
```
fnichol: Congrats--your first new Rust module in Habitat (that I know of)!

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+    match maybe_scaffold {
+        Some(scaffold) => {
+            init();
+            match scaffold.to_lowercase().as_ref() {
+                "core/scaffolding-ruby" |
+                "ruby" |
+                "rails" |
+                "ruby-scaffolding" => {
+                    let ident = PackageIdent::from_str("core/scaffolding-ruby").unwrap();
+                    let _ = ui.status(Status::Using, &format!("ruby scaffolding '{}'", ident));
```
fnichol: You should be able to call `try!()` on this expression which will early-return any IO failures with:

`ui.status(Status::Using, &format!("ruby scaffolding '{}'", ident))?;`

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+                    ui.para("")?;
+                    Ok(Some(ident))
+                }
+                "core/scaffolding-go" |
+                "go" |
+                "golang" |
+                "go-scaffolding" => {
+                    let ident = PackageIdent::from_str("core/scaffolding-go").unwrap();
+                    let _ = ui.status(Status::Using, &format!("go scaffolding '{}'", ident));
```
fnichol: Same as above

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+                    ui.para("")?;
+                    Ok(Some(ident))
+                }
+                "core/scaffolding-node" |
+                "node" |
+                "node.js" |
+                "javascript" |
+                "js" => {
+                    let ident = PackageIdent::from_str("core/scaffolding-node").unwrap();
+                    let _ = ui.status(Status::Using, &format!("node scaffolding '{}'", ident));
```
fnichol: Same as above

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+{
+    if path.as_ref().join("main.go").is_file() ||
+        path.as_ref().join("Godeps/Godeps.json").is_file() ||
+        path.as_ref().join("vendor/vendor.json").is_file() ||
+        path.as_ref().join("glide.yaml").is_file() ||
+        project_uses_gb(path.as_ref()).expect("Result<bool> not returned from .go file check")
+    {
+        return true;
+    }
+    return false;
```
reset: Like Ruby, the last line of a block in Rust returns that value. So this,

```
fn what() -> bool {
    true
}
```

Will return true. However,

```
fn what() -> bool {
    true;
}
```

This will result in a compilation error. Placing a `;` at the end of a statement will force it to return `()` ("unit").

The `return` isn't necessary to return a value on the last line, it's just used for early returns.

You don't need to make any changes here, I just wanted to give you this knowledge ;)

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+                "javascript" |
+                "js" => {
+                    let ident = PackageIdent::from_str("core/scaffolding-node").unwrap();
+                    let _ = ui.status(Status::Using, &format!("node scaffolding '{}'", ident));
+                    ui.para("")?;
+                    Ok(Some(ident))
+                }
+                _ => {
+                    let ident = PackageIdent::from_str(scaffold).unwrap();
+                    let _ = ui.status(Status::Using, &format!("custom scaffolding: '{}'", ident));
```
fnichol: Same as above

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+    ui.begin("Attempting autodiscovery ")?;
+    ui.para(
+        "No scaffolding type was provided. Let's see if we can figure out \
+        what kind of application you're planning to package.",
+    )?;
+    let current_path = Path::new(".");
+    if is_project_ruby(&current_path) {
+        let ident = PackageIdent::from_str("core/scaffolding-ruby").unwrap();
+        ui.begin("We've detected your app as Ruby")?;
+        let _ = ui.status(Status::Using, &format!("scaffolding package: '{}'", ident));
```
fnichol: `ui.status(…);` should do the trick

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+        ui.para("")?;
+        Ok(Some(ident))
+    } else if is_project_golang(&current_path) {
+        let ident = PackageIdent::from_str("core/scaffolding-go").unwrap();
+        ui.begin("We've detected your app as Golang")?;
+        let _ = ui.status(Status::Using, &format!("scaffolding package: '{}'", ident));
```
fnichol: Same as above

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+        ui.para("")?;
+        Ok(Some(ident))
+    } else if is_project_node(&current_path) {
+        let ident = PackageIdent::from_str("core/scaffolding-node").unwrap();
+        ui.begin("We've detected your app as Node.js")?;
+        let _ = ui.status(Status::Using, &format!("scaffolding package: '{}'", ident));
```
fnichol: Same as above

reset:
@thesentinels approve

tenor-125604835

thesentinels:

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+        ui.warn(
+            "Unable to determine the type of app in your current directory",
+        )?;
+        ui.para(
+            "For now, we'll generate a plan with all of the available plan options and callbacks. \
+            For more documentation on plan options try passing --withdocs or visit \
+            https://www.habitat.sh/docs/reference/plan-syntax/",
+        )?;
+        Ok(None)
```
fnichol: Isn't that the best??

components/hab/src/scaffolding.rs
```diff
@@ -0,0 +1,167 @@
...
+
+fn is_project_golang<T>(path: T) -> bool
+where
+    T: AsRef<Path>,
+{
+    if path.as_ref().join("main.go").is_file() ||
+        path.as_ref().join("Godeps/Godeps.json").is_file() ||
+        path.as_ref().join("vendor/vendor.json").is_file() ||
+        path.as_ref().join("glide.yaml").is_file() ||
+        project_uses_gb(path.as_ref()).expect("Result<bool> not returned from .go file check")
```
fnichol: Very nice!

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/258751714).

components/hab/src/scaffolding.rs
eeyun: YES

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/258751714).

PR #2851, opened 2017-07-29
Better async job status handling

chefsalim:
This change fixes a race condition in the builder-scheduler where a job status can get lost. Now, the incoming job status is persisted by the handler thread, and later picked up from the datastore to be processed by the scheduler. Switching over to this mode also yields a simplification where we can eliminate the need to keep a separate socket around for status notifications.

tenor-248153294

### DIFF
--- a/components/builder-jobsrv/src/data_store.rs
+++ b/components/builder-jobsrv/src/data_store.rs
@@ -26,7 +26,6 @@ use postgres;
 use protobuf;
 use protocol::net::{NetOk, NetError, ErrCode};
 use protocol::{originsrv, jobsrv, scheduler};
-use std::str::FromStr;
 use protobuf::ProtobufEnum;
 
 /// DataStore inherints being Send + Sync by virtue of having only one member, the pool itself.
@@ -474,34 +473,32 @@ impl DataStore {
     pub fn update_job(&self, job: &jobsrv::Job) -> Result<()> {
         let conn = self.pool.get_shard(0)?;
         let job_id = job.get_id() as i64;
-        let job_state = match job.get_state() {
-            jobsrv::JobState::Dispatched => "Dispatched",
-            jobsrv::JobState::Pending => "Pending",
-            jobsrv::JobState::Processing => "Processing",
-            jobsrv::JobState::Complete => "Complete",
-            jobsrv::JobState::Rejected => "Rejected",
-            jobsrv::JobState::Failed => "Failed",
-        };
+        let job_state = job.get_state().to_string();
 
         // Note: the following fields may all be NULL. As currently
         // coded, if they are NULL, then the corresponding fields in
         // the database will also be updated to be NULL. This should
         // be OK, though, because they shouldn't be changing anyway.
-        let build_started_at = match job.has_build_started_at() {
-            true => Some(
-                DateTime::<UTC>::from_str(job.get_build_started_at()).unwrap(),
-            ),
-            false => None,
+        let build_started_at = if job.has_build_started_at() {
+            Some(job.get_build_started_at().parse::<DateTime<UTC>>().unwrap())
+        } else {
+            None
         };
-        let build_finished_at = match job.has_build_finished_at() {
-            true => Some(
-                DateTime::<UTC>::from_str(job.get_build_finished_at()).unwrap(),
-            ),
-            false => None,
+
+        let build_finished_at = if job.has_build_finished_at() {
+            Some(
+                job.get_build_finished_at()
+                    .parse::<DateTime<UTC>>()
+                    .unwrap(),
+            )
+        } else {
+            None
         };
-        let ident = match job.has_package_ident() {
-            true => Some(job.get_package_ident().to_string()),
-            false => None,
+
+        let ident = if job.has_package_ident() {
+            Some(job.get_package_ident().to_string())
+        } else {
+            None
         };
 
         let (err_code, err_msg) = if job.has_error() {
@@ -554,16 +551,9 @@ fn row_to_job(row: &postgres::rows::Row) -> Result<jobsrv::Job> {
     job.set_id(id as u64);
     let owner_id: i64 = row.get("owner_id");
     job.set_owner_id(owner_id as u64);
+
     let js: String = row.get("job_state");
-    let job_state = match &js[..] {
-        "Dispatched" => jobsrv::JobState::Dispatched,
-        "Pending" => jobsrv::JobState::Pending,
-        "Processing" => jobsrv::JobState::Processing,
-        "Complete" => jobsrv::JobState::Complete,
-        "Rejected" => jobsrv::JobState::Rejected,
-        "Failed" => jobsrv::JobState::Failed,
-        _ => return Err(Error::UnknownJobState),
-    };
+    let job_state: jobsrv::JobState = js.parse().map_err(Error::UnknownJobState)?;
     job.set_state(job_state);
 
     let created_at = row.get::<&str, DateTime<UTC>>("created_at");
--- a/components/builder-scheduler/src/data_store.rs
+++ b/components/builder-scheduler/src/data_store.rs
@@ -20,9 +20,11 @@ use config::Config;
 use error::{Result, Error};
 use chrono::{DateTime, UTC};
 
+use protocol;
 use protocol::jobsrv::{Job, JobState};
 use protocol::scheduler::*;
 use protobuf::RepeatedField;
+use protobuf::{parse_from_bytes, Message};
 
 // DataStore inherits Send + Sync by virtue of having only one member, the pool itself.
 #[derive(Debug, Clone)]
@@ -293,6 +295,50 @@ impl DataStore {
                                       )
                                   SELECT * FROM my_group;
                                 $$"#)?;
+
+        // A queue to hold messages to be processed. This is currently only for JobStatus
+        // messages to allow the scheduler to deal with them asynchronously and reliably.
+        // The message queue is meant to provide 'at-least-once' delivery reliability.
+        // Eventually this functionality could be useful more broadly and be moved out to
+        // a separate module.
+        migrator.migrate(
+            "scheduler",
+            r#"CREATE TABLE IF NOT EXISTS message_queue (
+                                   id bigserial PRIMARY KEY,
+                                   message bytea
+                            )"#,
+        )?;
+
+        // Insert a message to the message queue.
+        migrator.migrate(
+            "scheduler",
+            r#"CREATE OR REPLACE FUNCTION insert_message_v1 (
+                               message bytea
+                           ) RETURNS SETOF message_queue AS $$
+                                    INSERT INTO message_queue (message)
+                                    VALUES (message)
+                                    RETURNING *;
+                               $$ LANGUAGE SQL VOLATILE"#,
+        )?;
+
+        // Retrieve oldest message(s) from the front of the message queue.
+        // The bigserial id is used to determine the message order.
+        migrator.migrate("scheduler",
+                        r#"CREATE OR REPLACE FUNCTION get_front_message_v1(max_rows int) RETURNS SETOF message_queue AS $$
+                               SELECT * FROM message_queue
+                               ORDER BY id ASC
+                               LIMIT max_rows;
+                        $$ LANGUAGE SQL VOLATILE"#)?;
+
+        // Delete a message from the message queue
+        migrator.migrate(
+            "scheduler",
+            r#"CREATE OR REPLACE FUNCTION delete_message_v1 (msg_id bigint) RETURNS void AS $$
+                   DELETE FROM message_queue
+                   WHERE id = msg_id;
+            $$ LANGUAGE SQL VOLATILE"#,
+        )?;
+
         migrator.finish()?;
 
         Ok(())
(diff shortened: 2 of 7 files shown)


thesentinels:

Thank you for contributing!

components/builder-scheduler/src/data_store.rs
```diff
@@ -262,6 +265,70 @@ impl DataStore {
...
+                              BEGIN
+                                  DELETE FROM job_status
+                                  WHERE id IN (
+                                      SELECT id
+                                      FROM job_status
+                                      ORDER BY created_at ASC
+                                      LIMIT $1
+                                  );
+                              END
+                              $$ LANGUAGE plpgsql VOLATILE"#,
```
christophermaier: All these functions can be implemented in SQL rather than PL/pgSQL. Functions that are implemented in SQL are "transparent" to the planner, allowing it to optimize execution in ways that it can't when given the "black box" of a PL/pgSQL function.

Granted, for these simple functions, that's likely not going to make a huge impact, but always giving the system more information is probably a good habit to get into. As a nice bonus, you'll also get to drop the `BEGIN` and `END` keywords, so you get shorter functions 👅

components/builder-scheduler/src/data_store.rs
```diff
@@ -262,6 +265,70 @@ impl DataStore {
...
+                            r#"CREATE OR REPLACE FUNCTION insert_job_status_v1 (
+                               job_id bigint,
+                               owner_id bigint,
+                               job_state text,
+                               project_name text,
+                               net_error_code int,
+                               net_error_msg text,
+                               build_started_at timestamptz,
+                               build_finished_at timestamptz
+                           ) RETURNS SETOF job_status AS $$
```
christophermaier: Can our Rust database code deal with functions that return a single database type, or not? If not, that's fine; I was just thinking that declaring this as `RETURNS job_status` would be ideal, since it's only ever going to return one; there's no real possibility of it returning multiple rows.

components/builder-scheduler/src/data_store.rs
```diff
@@ -586,4 +653,134 @@ impl DataStore {
...
+        let job_id = job.get_id() as i64;
+        let owner_id = job.get_owner_id() as i64;
+
+        let job_state = match job.get_state() {
+            JobState::Dispatched => "Dispatched",
+            JobState::Pending => "Pending",
+            JobState::Processing => "Processing",
+            JobState::Complete => "Complete",
+            JobState::Rejected => "Rejected",
+            JobState::Failed => "Failed",
```
christophermaier: Should we implement `FromStr` (so we get can use `parse`) and `Display` for `JobState` instead, to handle the conversion to and from a String?

components/builder-scheduler/src/data_store.rs
```diff
@@ -586,4 +653,134 @@ impl DataStore {
...
+                DateTime::<UTC>::from_str(job.get_build_started_at()).unwrap(),
+            ),
+            false => None,
+        };
+        let build_finished_at = match job.has_build_finished_at() {
+            true => Some(
+                DateTime::<UTC>::from_str(job.get_build_finished_at()).unwrap(),
+            ),
+            false => None,
+        };
```
christophermaier: These might be more concise as `if` statements, rather than `match` statements.

components/builder-scheduler/src/data_store.rs
```diff
@@ -262,6 +265,70 @@ impl DataStore {
...
+
+        migrator.migrate("scheduler",
+                        r#"CREATE OR REPLACE FUNCTION get_job_status_v1 (integer) RETURNS SETOF job_status AS $$
+                               BEGIN
+                                   RETURN QUERY SELECT * FROM job_status
+                                        ORDER BY created_at ASC
+                                        LIMIT $1;
+                                   RETURN;
+                               END
+                               $$ LANGUAGE plpgsql VOLATILE"#)?;
```
christophermaier: The name and argument type of this function suggests that you're getting the status for the job whose ID you're supplying; naming it something like `get_oldest_active_job`, `pop_job_queue` or similar would be more intention-revealing.

Also, is there any time we're calling this without an argument of `1`?

components/builder-scheduler/src/data_store.rs
```diff
@@ -262,6 +265,70 @@ impl DataStore {
...
+            "scheduler",
+            r#"CREATE OR REPLACE FUNCTION delete_job_status_v1 (integer) RETURNS void AS $$
+                              BEGIN
+                                  DELETE FROM job_status
+                                  WHERE id IN (
+                                      SELECT id
+                                      FROM job_status
+                                      ORDER BY created_at ASC
+                                      LIMIT $1
+                                  );
```
christophermaier: I think this could be implemented in terms of a `JOIN` with the output of `get_job_status_v1`, rather than duplicating that function's logic here.

components/builder-scheduler/src/server/scheduler.rs
```diff
@@ -420,39 +381,53 @@ impl ScheduleMgr {
...
+                        }
 
-                    _ => (),
+                        jobsrv::JobState::Failed => self.update_group_state(job.get_owner_id())?,
+
+                        _ => (),
+                    }
+
+                    // Clear the top job status from the queue
+                    self.datastore.clear_job_status(1)?;
```
christophermaier: This seems like it could have the potential for subtle bugs; can we just remove, by ID, the entry for the job we just processed, instead of assuming that it's at the top of our queue?

components/builder-scheduler/src/data_store.rs
```diff
@@ -586,4 +653,134 @@ impl DataStore {
...
+                &err_msg,
+                &build_started_at,
+                &build_finished_at,
+            ],
+        ).map_err(Error::JobStatusInsert)?;
+
+        Ok(())
+    }
+
+    pub fn get_job_status(&self, count: i32) -> Result<Vec<Job>> {
```
christophermaier: Same comment re: naming applies here

components/builder-scheduler/src/data_store.rs
chefsalim: Yeah, I think we can go either way, however the way we are doing things above is part of a pattern that is established to help keep things DRY.  There's some explanatory comments at https://github.com/habitat-sh/habitat/blob/master/components/builder-jobsrv/src/data_store.rs#L124

components/builder-scheduler/src/data_store.rs
chefsalim: Will tweak the function name to make it clearer.  The current usage is always 1, however like a number of other functions like the pending jobs ones, this is designed to eventually be able to handle more than just one status at a time.

components/builder-scheduler/src/data_store.rs
chefsalim: Will simplify this by deleting based on id, as per your other comment.

components/builder-scheduler/src/data_store.rs
chefsalim: Yeah these are super simple functions that can go either way - we should look at a high level at our PL/pgSQL vs. SQL usage. PL/pgSQL seems to offer much more flexibility, including the ability to trap errors, better performance with plan caching when things don't get inlined, etc.  It would be nice to just pick one language and stick with it so we can copy/paste more easily and use the patterns for that language.  I've leaned toward pgSQL since that's what the vast bulk of our code is atm.

mwrock: I really like the to see as little PL/pgSQL as possible. Not such a big deal for these functions but removing as much complexity from the data layer even if it sacrifices some sugar is a +. The only exception is where there are blatant perf improvements to the PL/pgSQL route.

components/builder-scheduler/src/data_store.rs
chefsalim: Good catch - I shoulda done that.

components/builder-scheduler/src/data_store.rs
chefsalim: Yeah - some copy pasta here.

components/builder-scheduler/src/data_store.rs
chefsalim: Will tweak.

components/builder-scheduler/src/data_store.rs
adamhjk: You're wrong about the PL/pgSQL planner conversation. Each function gets each statement automatically made into a prepared one, and then the actual execution plan gets cached and re-used. It's the most efficient thing possible for the planner - to get the same, you would need to have every statement prepared in advance manually. There is no benefit to using raw SQL functions most of the time.

adamhjk: We also made a concerted decision to use functions as part of the contract between the application and the database. The alternative is putting this kind of logic in the application, which in large part is not better.

components/builder-scheduler/src/server/scheduler.rs
chefsalim: Yeah agreed, I think that will be simpler as well - will update.

components/builder-scheduler/src/data_store.rs
mwrock: Fair enough. My primary concern is to avoid the temptation of including extra logic in postgres procs and keeping them in rust but that may be a different topic entirely.

christophermaier: Postgres is actually able to inline code and fold function bodies into outer queries in some cases when the function is written in SQL, but can't do that when it's in PL/pgSQL.

Another good thing about writing in SQL is that it forces you to express your problem in terms of set operations, which is almost without exception going to be better than expressing it iteratively. For instance, I helped @raskchanky refactor a PL/pgSQL function that was making n+1 insert statements into an SQL statement that does the same, but in a single operation. True, nothing is stopping you from executing that same single query in a PL/pgSQL function, but that seems like just adding noise at that point.

Certainly, if there are situations where we really do need the capabilities of PL/pgSQL (like exception handling, for example), then by all means, we should use PL/pgSQL. You can do a lot more things in SQL in Postgres these days, though, so that's becoming increasingly unnecessary, particularly for the kinds of things we're doing on this project. That, coupled with the clarity that being forced to express your problem in set-oriented operations, makes me reach for SQL first.

adamhjk:
I care more about not abandoning the function boundaries we have now. If
it's a SQL function rather than a pl/pgsql one, great. If we move more
logic into the application, we have to start rethinking the design for the
forward compatibility of the back end. Not that we can't do it, but that's
why it is the way it is.

components/builder-scheduler/src/data_store.rs
adamhjk: Ya - the important thing is to not change the contract between the database and the application. We're consciously making the database interface be functions with versions in order to make forward migration and new behavior easy to add. It's not the only way to do that, but it's the way we chose, and we should make sure we don't change it accidentally. :) All things being equal, I'm 2/10 on wether a given function is written in SQL or PL/pgSQL. I'm 8/10 that we should keep the current function/app boundaries as they are, unless we want to really reconsider the entire approach.

components/builder-jobsrv/src/data_store.rs
```diff
@@ -554,16 +552,11 @@ fn row_to_job(row: &postgres::rows::Row) -> Result<jobsrv::Job> {
...
-    let job_state = match &js[..] {
-        "Dispatched" => jobsrv::JobState::Dispatched,
-        "Pending" => jobsrv::JobState::Pending,
-        "Processing" => jobsrv::JobState::Processing,
-        "Complete" => jobsrv::JobState::Complete,
-        "Rejected" => jobsrv::JobState::Rejected,
-        "Failed" => jobsrv::JobState::Failed,
-        _ => return Err(Error::UnknownJobState),
-    };
+    let job_state = jobsrv::JobState::from_str(&js).map_err(
```
christophermaier: Minor nit pick, but it seems like it's more idiomatic to use `parse` rather than calling `from_str` directly.

components/builder-scheduler/src/server/scheduler.rs
```diff
@@ -420,39 +381,53 @@ impl ScheduleMgr {
...
+                            self.update_group_state(job.get_owner_id())?
+                        }
 
-                    _ => (),
+                        jobsrv::JobState::Failed => self.update_group_state(job.get_owner_id())?,
+
+                        _ => (),
+                    }
+
+                    // Clear the top job status from the queue
```
christophermaier: This comment isn't quite in line with the implementation anymore.

adamhjk:
Looks good to me. In particular, this is a step toward the right thing regardless - it's either going to be fine for the long term, or need another refactor (one that deals with job volume/persistence in the database) - but assuming @christophermaier feels good, I say merge this.

Nice work all around, btw - @chefsalim for the code, and the team for the review.

christophermaier:
I agree with @adamhjk... I really like how this ended up; talking about the queue as a queue really makes the code flow very nicely. Nice job @chefsalim.

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/260440372).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/260440372).

PR #2855, opened 2017-08-01
Add --peer-watch-file flag

### LINKED ISSUE #2735: Implement `--peer-watch-file` flag for joining the ring
This flag could be used instead of the `--peer` flag; it would take a path to a file as a value. It would then watch it for changes (with inotify or something more portable). The contents of the file should be one or more IP addresses, just like in the `--peer` flag.

This would solve the issue of a different invocation for the initial supervisor in K8s, as all the invocations would include the `--peer-watch-file` flag. The file would be provided by the k8s controller/operator.

Also, you don't need to set up the k8s cluster to test this solution.

krnowak:
This is to fix #2735.

### DIFF
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -346,6 +346,13 @@ impl Server {
         Ok(())
     }
 
+    pub fn need_peer_seeding(&self) -> bool {
+        let m = self.member_list.members.read().expect(
+            "Members lock is poisoned",
+        );
+        m.is_empty()
+    }
+
     /// Blacklist a given address, causing no traffic to be seen.
     pub fn add_to_blacklist(&self, member_id: String) {
         let mut blacklist = self.blacklist.write().expect(
--- a/components/sup/src/manager/file_watcher.rs
+++ b/components/sup/src/manager/file_watcher.rs
@@ -57,7 +57,7 @@ pub trait Callbacks {
 }
 
 // Essentially a pair of dirname and basename.
-#[derive(Clone, Default)]
+#[derive(Clone, Debug, Default)]
 struct DirFileName {
     directory: PathBuf,
     file_name: OsString,
@@ -93,6 +93,7 @@ impl DirFileName {
 // status in mountinfo, when some change there happens.
 
 // Similar to DirFileName, but the file_name part is optional.
+#[derive(Debug)]
 struct SplitPath {
     directory: PathBuf,
     file_name: Option<OsString>,
@@ -114,6 +115,7 @@ impl SplitPath {
 }
 
 // Contents of this type are for `process_path`.
+#[derive(Debug)]
 struct ProcessPathArgs {
     // The beginning of the patch to process (very often it is root,
     // but not always).
@@ -138,12 +140,14 @@ struct ProcessPathArgs {
 
 // This struct tells that for `path` the previous item in chain is
 // `prev`.
+#[derive(Debug)]
 struct ChainLinkInfo {
     path: PathBuf,
     prev: Option<PathBuf>,
 }
 
 // This struct is passed together with some event actions.
+#[derive(Debug)]
 struct PathsActionData {
     dir_file_name: DirFileName,
     args: ProcessPathArgs,
@@ -152,6 +156,7 @@ struct PathsActionData {
 // This is stores information about the watched item
 //
 // TODO(krnowak): Rename it.
+#[derive(Debug)]
 struct Common {
     // TODO: maybe drop this? we could also use dir_file_name.as_path()
     path: PathBuf,
@@ -316,6 +321,7 @@ impl CommonGenerator {
 }
 
 // An item we are interested in.
+#[derive(Debug)]
 enum WatchedFile {
     Regular(Common),
     MissingRegular(Common),
@@ -389,6 +395,7 @@ fn simplify_abs_path(abs_path: &PathBuf) -> PathBuf {
 //
 // We translate `DebouncedEvent` to `EventAction`, and `EventAction`
 // to a list of `PathsAction`s.
+#[derive(Debug)]
 enum EventAction {
     Ignore,
     PlainChange(PathBuf),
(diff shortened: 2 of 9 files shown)


thesentinels:

Thank you for contributing!

krnowak:
I see that there is a bunch of rustfmt failures, will address that. Not sure about the `test manager::service::hooks::tests::compile_hook_table ... FAILED` thing, though.

reset:
@krnowak we've got to disable or fix that test, don't worry about it. I'm going to review this at some point today or tomorrow, the high level looks good as an alternative to the `--peer` flag

krnowak:
I have addressed the rustfmt issues, but there are some other (likely unrelated) failures.

christophermaier:
Not sure what was up with those test failures; I re-ran and they cleared up.

Looks good to me; I'll defer to @reset for the final 👍 , though.

adamhjk:
I'm a little confused about this implementation. Currently, you set peers on the CLI with:

`--peer WHATEVER`, and then they get added to the member list through `server.member_list.add_initial_member`.  Then in the outbound thread, if we have no members at all, we ping the `initial_members`. We will keep doing this until we reach `num_initial _members / 2 + 1)` (strictly more than half), at which point we assume we're fully connected.

This patch conflicts with `peer` as a CLI flag, which makes sense - either you seed through a file, or you seed through the CLI params. But then we move to add a new top-level addition to the server, the `peer_seed`, rather than adding them to the `member_list.initial_members`. The side effect is we have two paths to adding members - the `initial_peer` path and the `peer_seed` path. Why? It seems to me that we should follow the existing path, which is to add the initial members to initial members, and keep the logic we already have. Is there another behavior we are looking for?

I'm 👎  on merging this until we sort it.

components/sup/src/manager/mod.rs
```diff
@@ -838,6 +849,22 @@ impl Manager {
...
+    fn update_peers_from_watch_file(&mut self) -> Result<()> {
+        if !self.butterfly.need_peer_seeding() {
+            return Ok(());
+        }
+        match self.peer_watcher {
+            None => Ok(()),
+            Some(ref watcher) => {
+                if watcher.has_fs_events() {
+                    let members = watcher.get_members()?;
+                    self.butterfly.set_peer_seed(members)?;
```
adamhjk: This should be calling `self.butterfly.member_list.add_initial_members`

components/butterfly/src/server/mod.rs
```diff
@@ -119,6 +120,7 @@ impl Clone for Server {
             gossip_rounds: self.gossip_rounds.clone(),
             blacklist: self.blacklist.clone(),
             socket: None,
+            peer_seed: Arc::clone(&self.peer_seed),
```
adamhjk: I don't understand why this is here, rather than using the `member_list.initial_members` array.

krnowak:
An update here: for this feature to work inside kubernetes I need to implement a more complicated way of watching the path. Currently if I pass `--peer-watch-file=/habitat-operator/peer`, I'm trying to watch the `/habitat-operator` directory and filter the events that are related to the `peer` file. But when putting this file through kubernetes, it does not simply drop the file at the location we want. Instead it creates something similar to the following directory structure:

```
/habitat-operator
+ ..45.234.765634.123354 (this seems to be a mount point)
| + peer
+ data -> ..45.234.765634.123354
+ peer -> data/peer
```

And, when the operator updates the file, kubernetes does not update the `....45.234.765634.123354/peer` file, but instead it creates another directory with the new version of the file and rewires the symlinks.

So the current version of file watching is not enough.

krnowak:
Here's the updated branch. It still has a lot of rough edges, so the work is ongoing, but at least stuff seems to work.

I'm doing a bit refactoring (https://github.com/kinvolk/habitat/tree/krnowak/file-watcher-refactor) and adding to have a structured way for testing (https://github.com/kinvolk/habitat/tree/krnowak/new-file-watcher-tests).

We are going to test this code with k8s, I have already built habitat packages for hab (https://bldr.habitat.sh/#/origins/krnowak).

krnowak:
The CI will likely fail because of warnings, which I addressed in my refactoring branch.

krnowak:
Updated the PR again, this time there should be no warnings. No structured tests yet though, these are ongoing in my branch (https://github.com/kinvolk/habitat/tree/krnowak/new-file-watcher-tests). For now there is a test that checks the k8s scenario.

@adamhjk: Could you have a look? Hopefully the FileWatcher stuff is not too dense.

components/sup/src/manager/file_watcher.rs
```diff
@@ -0,0 +1,1463 @@
...
+            file_name: file_name.to_owned(),
+        })
+    }
+
+    fn as_path(&self) -> PathBuf {
+        self.directory.join(&self.file_name)
+    }
+}
+
+// TODO: handle mount events, we could use libc crate to get the
```
asymmetric: Think a `TODO(krnowak)` would be useful here, so people know who to ping in case they have questions.

components/sup/src/manager/file_watcher.rs
```diff
@@ -0,0 +1,1463 @@
...
+            WatchedFile::Symlink(c) |
+            WatchedFile::Directory(c) |
+            WatchedFile::MissingDirectory(c) => c,
+        }
+    }
+}
+
+// Similar to std::fs::canonicalize, but without resolving symlinks.
+//
+// I'm not sure if this is entirely correct, consider:
```
asymmetric: I would "sign" this as well, especially since you use the first person :)

christophermaier:
@krnowak What is the current status of this work; ready for review? Thanks!

krnowak:
@christophermaier: Yeah, it's ready for review. There is another PR which is based on this one that contains more test. I think it can be reviewed later though, after this one.

christophermaier:
@krnowak Sounds good; we'll get some eyes on this very soon.

PR #2860, opened 2017-08-02
Validate origin names with the same rules as the Depot

### LINKED ISSUE #2833: hab CLI allows creating origins that are invalid in depot
Hello!

I was walking through the Habitat tutorial and I created an origin with my usual IRC and GitHub username: SpamapS. I created it with the `hab setup` command, which happily accepted it. I used this in the mytutorial app, and it all worked fine for local tests. However, when I wanted to create an origin in the depot, I had to conform to the depot rules, which require the name to start with lowercase.

hab 0.26.1/20170715033447

```
LSB Version:	core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.2 LTS
Release:	16.04
Codename:	xenial
```

smacfarlane:
Resolves #2833 

First pass at validating origin names with the rules used on the depot. 

Length is validated separately from the structure of the origin because I was encountering `'Unable to compile regex: CompiledTooBig(10485760)'` when using the pattern `^[a-z0-9][\w\-_]{0,254}$`

One area I'd like to make better is that name is only validate when a key is generated, allowing you to have an invalid name if you skip key generation (first screenshot).  Guidance on where in the code a more general validation method should be placed would be appreciated.

4 docker-compose 2017-08-02 12-42-18
4 docker-compose 2017-08-02 12-58-16

Signed-off-by: Scott Macfarlane <macfarlane.scott@gmail.com>


thesentinels:

Thank you for contributing!

reset:
@smacfarlane we have a crate called `core` which is where shared modules between the supervisor and builder live. We would want to move the validation function to there and call it at any point user input is accepted. The key generation you've got here is one of them and I think the others are all server side in a handler in `builder-api` or `builder-depot` (which is mounted in `builder-api` for now).

What you have in this PR is still a clear improvement. Do you want to move forward with these changes or refactor and also tackle some server side stuff?

smacfarlane:
@reset Thanks!  I'd like to do some refactoring before merging.

smacfarlane:
I've done the initial refactor, I'm looking into the server-side of things now.  

One thing that I'm not super happy with is the experience of being dropped out of `setup`  on invalid input.  I'm thinking that at minimum the description of a valid origin should be presented to the user before the prompt.

reset:
@smacfarlane a validation function should just return a `bool` IMO. It doesn't belong in the crypto module, it probably belongs as a function in the `package/ident.rs` module since an origin is just a portion of a package identifier

smacfarlane:
Updated as per the guidance above.  One last thing that I'm not happy with is it feels like I'm abusing the left-over InvalidOrigin error to not have duplicate strings in the code. Should that be changed into a static string (in `package/ident.rs`)? 

I've also added valid origin description and move the origin generation into a loop to give users a chance to correct invalid input, rather than exit and forcing them to restart.  It felt like a better experience that way, but I thought it would be good to call that change out if there's UX reasons for a different flow.  

4 docker-compose 2017-08-04 10-25-03

reset:
@smacfarlane I hear what you're saying about using the same error message. I think the proper Rust way is to make the error as generic as possible and then add context around it from other callers. You can see this in some of our other work in this project where you'll run into something like `Error::KeyGen(String)` and the display function for that will result in something like:

"Unable to generate key: Origins must contain spaces ..."

Does this make sense?

reset:
Check this example out: https://github.com/habitat-sh/habitat/blob/master/components/sup/src/error.rs#L149

This is a way to give context to a serialization error for example

smacfarlane:
Conceptually, I understand (I think).  I've made a small change that allows the invalid origin name to be passed into the error for more context. 

However, and I may be getting hung up on nothing here, the context of why the origin is invalid (too long, invalid chars, etc) is lost when the is_valid_origin_name function returns. I've taken the approach of describing what a valid origin looks like rather than why the validation failed. This gives the user some context, rather than an `Invalid Origin: Foo` message.  However, it also means I have that description as a string in more than one location.

reset:
@smacfarlane you're totally right, if you can identify _why_ it's not correct then we should tell them *and* also tell them what a valid one looks like. Do you want to make that modification or merge as is? This is already a big improvement :)

smacfarlane:
I think I'm happy with where it's at as a first step.   If there's no revisions that need to be made let's merge it.

reset:
@thesentinels approve

tenor-255936180

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/262718827).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/262718827).

PR #2869, opened 2017-08-04
Receive initial peer list from a Terraform variable

reset:


thesentinels:

Thank you for contributing!

terraform/instances.tf
```diff
@@ -615,35 +615,12 @@ resource "aws_instance" "worker" {
...
-data "template_file" "hab_sup_seed" {
-  template = "${file("${path.module}/templates/hab-sup.service")}"
-
-  vars {
-    flags               = "--auto-update --channel ${var.release_channel} --events hab-eventsrv.default --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port} --permanent-peer"
-    gossip_listen_port  = "${var.gossip_listen_port}"
-    peer_ip             = "127.0.0.1"
-    log_level           = "${var.log_level}"
+    flags     = "--auto-update --channel ${var.release_channel} --events hab-eventsrv.default --listen-gossip 0.0.0.0:${var.gossip_listen_port} --listen-http 0.0.0.0:${var.http_listen_port}"
+    peers     = "${var.peers}"
```
christophermaier: Also, what's the reasoning behind having `peers` be a separate template variable, when it effectively ends up being prepended to the value of `flags` anyway? Why not just set it all up here?

Also also, does this blow up in the template if the variable is empty?

terraform/variables.tf
```diff
@@ -79,6 +79,10 @@ variable "private_subnet_id" {
   description = "Identifier for private AWS subnet"
 }
 
+variable "peers" {
+  description = "List of initial Supervisor peers"
+}
```
christophermaier: Based on this description I was expecting `peers` to be an actual list, but looking over at the [companion PR](https://github.com/habitat-sh/cloud-environments/pull/47), it seems that this is actually a string. I suppose this answers my earlier question about what happens if this "list" is empty, because without a default defined here, it has to be non-empty (there's the pathological case of an empty string, though, but we'd probably have much bigger issues if we got down to here with that).

Still looking at the other PR, but I'd recommend at least tweaking this description to reflect that this is already going to be a properly-formatted, non-empty string that gets created over in the cloud-environments repo. 

It may be worth refactoring eventually, such that that string formatting actually happens over here, just in case any community members happen to be using this Terraform code (either directly, or as inspiration) for their own deployments.

EDIT: OK, I've read through the other PR, so I get why this is just a string here - the logic for creating this peer list is also needed for some of our private instances over there. I feel like I want to refactor that, but I'm not sure what a good way to do that would be 😦 . I'd still like to tweak this `description` field, but other than that, let's keep going with the current implementation.

terraform/instances.tf
reset: Oh there's no difference, I could just totally pass it as a flag above. This is just an artifact from the previous form that this resource took. I'll simplify it :)

The variable is required so it can't be empty

terraform/variables.tf
reset: Actually I should change this to a list and not just a string. It *must* be a string to pass to the template resource but it doesn't need to be when we receive it at this point

reset:
@christophermaier addressed your comments here and in the companion PR. Gonna push this through and we can revisit the duplication of bootstrap scripts - we gotta figure that one out

Issue #1938, opened 2017-03-10
RFC: add a `--var(iable)` flag to `hab sup start`


bixu:
I've run into a number of cases when I really wanted a syntax like `hab sup start --var port=22` when running a package - cases where, for various reasons, reading in or setting environment variables is a bit awkward (for example variable names that include `.` characters, which can't be handled by Bash).

mwrock:
Have you tried this syntax:

```
HAB_MYTUTORIALAPP='message = "Habitat rocks!"' hab start <origin>/<packagename>
```

As mentioned on https://www.habitat.sh/docs/run-packages-apply-config-updates/ ?

bixu:
Sure,  but if my app name has a . it, Bash can't handle the env var I would try to set.

Also, for more than one or two key/value pairs, this syntax gets awkward and hard to read very quickly.

Passing key/values to Docker is much nicer with -- flags, and was the inspiration for the idea.

eeyun:
In our discussion during triage we all felt like this Issue really touches on something deeper in our UX and usability. Even if perhaps the implementation suggested here isn't what we go with I think we're all in agreement that theres some pain around this.

smurawski:
Since it can be tough in some orchestration systems to pass environment variables with valid TOML (which requires line breaks).  We could look at some convention for taking JSON and converting that to TOML (we should have the conversion functions already in the codebase), but I'm sure there are scenarios with valid json that isn't valid toml, etc..

bodymindarts:
I agree that there is an issue here.
I had a discussion this morning where we were wishing there would be a way to specify individual settings via env var like:
```
HAB_SERVICE_USER="foo"
HAB_SERVICE_PASS="bar"
```

Either way this is implemented it is clear that needing to pass in a toml file as 1 variable has its disadvantages.

bodymindarts:
@bixu @eeyun 

I have written a little alternative entrypoint script that can be used to pass in individual setting via environment variables.
It can be used by mounting the script into the container and then running docker with `--entrypoint /path/to/hab_entry.sh`
See more info [here](https://github.com/starkandwayne/habitat-plans/tree/master/scripts)

eeyun:
I'm glad that theres a thing that folks can use on the interim but I sort of feel like since compose files are the primary interface for interacting with Swarm, we might want to consider a native feature to resolve this as opposed to an external bolt on.

eeyun:
[...]

PR #2887, opened 2017-08-08
added json support to env var configuration injection

eeyun:
Ok, this is a tough one and at first glance it feelsbadman. But this is the outcome of https://github.com/habitat-sh/habitat/issues/1938 . For history and explanation you can read read the lengthy backscroll on the issue.

That being said for the sake of anyone following along and for the sake of historical context I'm going to put the TL;DR here:

12 Factor application patterns include a specific API for application configuration which is the ENVVVAR. As habitat has adopted a design that makes the 12 Factor approach to apps easier we've included a granular way to  configure a habitat package via a habitat prefixed environment variable `HAB_PKGNAME`. 

As users have started to discover more workflow and patterns around building and testing their apps as habitat packages some pain points have arrived around this API. What we've discovered is that there's some real pain around utilizing it to do more than set a single value. The reason being the syntax is obtuse and not easily read due to only accepting TOML as our configuration values. TOML is excellently readable in a file format, but as it is line-break delimited injecting configuration values here feels obtuse and somewhat arcane.

This problem gets exacerbated further by other developer and ops tooling around the package. The prime example being docker-compose which is a TML format and doesn't support multi-line envvars. You run into similar issues with things like kubernetes pod definitions. There are workarounds in that you can mount a file that contains your configuration into each container that you run but this adds to a negative experience for first-time users who are expecting a more natural ENVVAR api.

Through lots of discussion on the above RFC I think that we've come to the conclusion that providing JSON as an optional input type for this specific API is the least gross method for correcting the problem currently.

Signed-off-by: eeyun <ihenry@chef.io>

### DIFF
--- a/components/sup/src/error.rs
+++ b/components/sup/src/error.rs
@@ -111,6 +111,7 @@ pub enum Error {
     BadPackage(PackageInstall, hcore::error::Error),
     BadSpecsPath(PathBuf, io::Error),
     BadStartStyle(String),
+    BadEnvConfig(String),
     ButterflyError(butterfly::error::Error),
     DepotClient(depot_client::Error),
     EnvJoinPathsError(env::JoinPathsError),
@@ -200,6 +201,9 @@ impl fmt::Display for SupError {
                 )
             }
             Error::BadStartStyle(ref style) => format!("Unknown service start style '{}'", style),
+            Error::BadEnvConfig(ref pkg) => {
+                format!("Unable to find valid TOML or JSON in HAB_{} ENVVAR", pkg)
+            }
             Error::ButterflyError(ref err) => format!("Butterfly error: {}", err),
             Error::ExecCommandNotFound(ref c) => {
                 format!("`{}' was not found on the filesystem or in PATH", c)
@@ -339,6 +343,7 @@ impl error::Error for SupError {
             Error::BadPackage(_, _) => "Package was malformed or contained malformed contents",
             Error::BadSpecsPath(_, _) => "Unable to create the specs directory",
             Error::BadStartStyle(_) => "Unknown start style in service spec",
+            Error::BadEnvConfig(_) => "Unknown syntax in Env Configuration",
             Error::ButterflyError(ref err) => err.description(),
             Error::ExecCommandNotFound(_) => "Exec command was not found on filesystem or in PATH",
             Error::TemplateFileError(ref err) => err.description(),
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -25,6 +25,7 @@ use std::result;
 use ansi_term::Colour::Purple;
 use hcore::crypto;
 use serde::{Serialize, Serializer};
+use serde_json;
 use toml;
 
 use super::Pkg;
@@ -174,11 +175,24 @@ impl Cfg {
             .replace("-", "_");
         match env::var(&var_name) {
             Ok(config) => {
-                let toml = toml::de::from_str(&config).map_err(|e| {
-                    sup_error!(Error::TomlParser(e))
-                })?;
-                self.environment = Some(toml::Value::Table(toml));
-
+                match toml::de::from_str(&config) {
+                    Ok(toml) => {
+                        self.environment = Some(toml::Value::Table(toml));
+                        return Ok(());
+                    }
+                    Err(err) => debug!("Attempted to parse env config as toml and failed {}", err),
+                }
+                match serde_json::from_str(&config) {
+                    Ok(json) => {
+                        self.environment = Some(toml::Value::Table(json));
+                        return Ok(());
+                    }
+                    Err(err) => debug!("Attempted to parse env config as json and failed {}", err),
+                }
+                self.environment = None;
+                Err(sup_error!(
+                    Error::BadEnvConfig(package.name.to_ascii_uppercase())
+                ))
             }
             Err(e) => {
                 debug!(
@@ -187,9 +201,9 @@ impl Cfg {
                     e
                 );
                 self.environment = None;
+                Ok(())
             }
-        };
-        Ok(())
+        }
     }
 }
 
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/config.rs
```diff
@@ -371,6 +380,14 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
     }
 }
 
+// fn is_environment_toml(config: &str) -> bool {
```
mwrock: I'm assuming this is something that got left by accident.

eeyun: oh whoops! let me clean that up real quick.

eeyun:
I'm going to correct the commented nonsense and add docs to the PR before we merge. Going to add a DO NOT MERGE flag in the meantime

eeyun:
@mwrock updated  to include docs as well as removed the errant commented function

eeyun:
I must have the wrong version of rustfmt or something. Any idea what version we're supposed to be using?

eeyun:
Corrected Rustfmt issue

components/sup/src/manager/service/config.rs
```diff
@@ -174,11 +175,20 @@ impl Cfg {
...
             Ok(config) => {
-                let toml = toml::de::from_str(&config).map_err(|e| {
-                    sup_error!(Error::TomlParser(e))
-                })?;
-                self.environment = Some(toml::Value::Table(toml));
-
+                match toml::de::from_str(&config) {
+                    Ok(toml) => self.environment = Some(toml::Value::Table(toml)),
+                    Err(e) => {
+                        outputln!("{}", sup_error!(Error::TomlParser(e)));
```
reset: I wouldn't output that there was a parser error here since we're going to try and fallback to `JSON`. We can think of this instead like a "hint" system where we aren't sure what your intent was - we don't want to warn or error because we don't actually know what the intent was.

I would remove the Json/Toml parser errors here entirely and instead use early returns to short circuit the function. If we don't early return out, have the last line of this return an error like `Error::BadEnvConfig` which states that we were unable to read valid TOML or JSON and print out the contents that we did receive. You can store the contents in the enum variant like this `Error::BadEnvConfig(String)`

components/sup/src/error.rs
```diff
@@ -505,6 +508,12 @@ impl From<toml::de::Error> for SupError {
     }
 }
 
+impl From<serde_json::Error> for SupError {
```
reset: Per my comment below, I'd remove this trait implementation and the JsonParser error completely

eeyun:
@reset i've gone ahead and cleaned up the commit history and added changes in the vein of what you asked before. Let me know if theres more that I should do or if I should change this implementation

components/sup/src/error.rs
```diff
@@ -159,6 +160,7 @@ pub enum Error {
     TomlEncode(toml::ser::Error),
     TomlMergeError(String),
     TomlParser(toml::de::Error),
+    JsonParser(serde_json::Error),
```
reset: We don't need this error anymore

components/sup/src/manager/service/config.rs
```diff
@@ -174,11 +175,20 @@ impl Cfg {
             .replace("-", "_");
         match env::var(&var_name) {
             Ok(config) => {
-                let toml = toml::de::from_str(&config).map_err(|e| {
-                    sup_error!(Error::TomlParser(e))
-                })?;
-                self.environment = Some(toml::Value::Table(toml));
-
+                match toml::de::from_str(&config) {
```
reset: The "short circuit"/"early return" pattern I was thinking of would make this function look a bit like this:

```rust
fn load_environment(&mut self, package: &Pkg) -> Result<()> {
    let var_name = format!("{}_{}", ENV_VAR_PREFIX, package.name)
        .to_ascii_uppercase()
        .replace("-", "_");
    match env::var(&var_name) {
        Ok(config) => {
            match toml::de::from_str(&config) {
                Ok(toml) => {
                    self.environment = Some(toml::Value::Table(toml));
                    return Ok(());
                },
                Err(err) => debug!("Attempted to parse env config as toml and failed, {}", err),
            }
            match serde_json::from_str(&config) {
                Ok(json) => {
                    self.environment = Some(toml::Value::Table(json));
                    return Ok(());
                },
                Err(err) => debug!("Attempted to parse env config as json and failed, {}", err),
            }
            self.environment = None;
            Err(Error::BadEnvConfig(package.name.to_ascii_uppercase()))
        }
        Err(e) => {
            debug!(
                "Looking up environment variable {} failed: {:?}",
                var_name,
                e
            );
            self.environment = None;
            Ok(())
        }
    }
}
```

I recommend this pattern for this function becuase it let's you "fall through" and then eventually fail

eeyun:
Per @reset in slack. Changes implemented self merge iminent

eeyun:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/263135756).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/263135756).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/263135756) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/263135756).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/263135756).

scotthain:
YAAAZZZ!!! 😍

PR #2893, opened 2017-08-08
Extract rumor "heat" tracking to separate module

christophermaier:
This extracts the old `RumorList` struct, renames it to `RumorHeat`,
and simplifies its API to better reflect what its doing in such a way
that calling code reads more clearly.

Calling it a "rumor list" was misleading because a) it's not a list!
and b) it gave no indication that it was actually being used to track
the "heat" of a rumor. The API the struct exposed was also not as
descriptive as it could be, it returned more data than necessary, and
it also had more functions than we truly needed!

Particular changes to note:

* the `take` function was removed; it was only used in tests, and the
  tests were rewritten to not need it.
* the `take_by_kind` function was removed; it was only called in one
  place, so the implementation was basically just moved to that one
  call site.
* `insert` was renamed `start_hot_rumor` to better reflect what it
  actually did (in particular, calling it a second time didn't just
  "insert" a rumor, it effectively reset the "heat" information if the
  rumor was already being tracked).
* The `rumors` function was renamed to `currently_hot_rumors` to
  better reflect what it was doing and make calling code read
  better. Additionally, it now returns a `Vec<RumorKey>` instead of a
  `RumorVec`, which was a vector of tuples that paired a `RumorKey`
  with the current "heat" of that rumor. It turns out, this additional
  "heat" information was never used at call sites; it was really a
  purely internal detail. All call sites were correspondingly changed.
* `update_heat` was renamed to `cool_rumors` to better express what it
  was doing.
* Tests were revamped.

### DIFF
--- a/components/butterfly/src/rumor/heat.rs
+++ b/components/butterfly/src/rumor/heat.rs
@@ -1,0 +1,349 @@
+// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! In Butterfly, as in life, new rumors are "hot", but they get less
+//! exciting the more you hear them. For a given rumor, we keep track
+//! of how many times we've sent it to each member. Once we've sent
+//! that member the rumor a maximum number of times, the rumor has
+//! "cooled off". At that point we'll stop sending that rumor to the
+//! member; by now they will have heard it!
+//!
+//! Note that the "heat" of a rumor is tracked *per member*, and is
+//! not global.
+
+// Standard Library
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+
+// Internal Modules
+use rumor::RumorKey;
+
+// TODO (CM): Can we key by member instead? What do we do more frequently?
+// TODO (CM): Might want to type the member ID explicitly
+// TODO (CM): what do we do with rumors that have officially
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

components/butterfly/src/rumor/heat.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// For each rumor given, "cool" the rumor for the given member by
+    /// incrementing the count for how many times it has been sent
+    /// out. As a rumor cools, it will eventually cross a threshold
+    /// past which it will no longer be gossipped to the member.
+    ///
+    /// Call this after sending rumors out across the network.
+    ///
+    /// **NOTE**: "cool" in the name of the function is a *verb*; you're
+    /// not going to get a list of cool rumors from this.
+    pub fn cool_rumors(&self, id: &str, rumors: &Vec<RumorKey>) {
```
reset: I love the name of this function

reset: I would recommend just a slice (`&[]`) here instead of `&Vec`

reset:
@christophermaier great refactor here, I think `RumorHeat` is a just fine name but since you changed the name because it didn't appropriately represent the structure it's implemented with do you think it makes sense for the new name to reflect that? `RumorHeatMap`? It's cool to leave it as `RumorHeat` as well

christophermaier:
@reset I actually did name it `RumorHeatMap` in an earlier pass at this refactoring 😄 

I ended up changing it to remove any kind of data structure in the name because in the end, that's just an implementation detail. If the API it provided was more obviously "map-like", then it would have been OK, but as things currently stand, I felt "Map" was just unnecessary and got in the way.

components/butterfly/src/rumor/heat.rs
christophermaier: Roger, I'll take a look at tweaking this return value.

reset:
@christophermaier that makes perfect sense to me!

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/262463875).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/262463875).

PR #2897, opened 2017-08-08
Don't upload to "stable" channel by default

christophermaier:
When packages are uploaded, they *always* get added to the `unstable`
channel. However, by passing the `--channel` option to `hab pkg
upload`, users may specify an additional release channel to put the
package into.

Previously, we erroneously defaulted the value of this option to
`stable`, meaning that unless the user specified something else, every
package would be uploaded to both unstable `unstable` *and* `stable`
channels.

The fix turned out to be pretty easy. We were already passing around
an `Option` type for the channel name; now, that is simply a `None` if
the user doesn't supply a value. Docstrings, comments, and variable
names were changed to reflect this new state of affairs and make it
more clear what the desired behavior is.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -245,7 +245,9 @@ pub fn get() -> App<'static, 'static> {
                     "Use a specific Depot URL (ex: http://depot.example.com/v1/depot)")
                 (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for the Depot")
                 (@arg CHANNEL: --channel -c +takes_value
-                    "Upload to the specified release channel (default: unstable)")
+                    "Additional release channel to upload package to. \
+                     Packages are always uploaded to `unstable`, regardless \
+                     of the value of this option. (default: none)")
                 (@arg HART_FILE: +required +multiple {file_exists}
                     "One or more filepaths to a Habitat Artifact \
                     (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
--- a/components/hab/src/command/pkg/upload.rs
+++ b/components/hab/src/command/pkg/upload.rs
@@ -29,20 +29,23 @@
 //! complex than just latest version.
 //!
 
+// Standard Library
 use std::path::{Path, PathBuf};
 
-use common::ui::{Status, UI};
+// External Libraries
+use hyper::status::StatusCode;
+use retry::retry;
+
+// Local Dependencies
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/upload.rs
```diff
@@ -211,16 +219,16 @@ fn upload_into_depot(
...
+    if package_uploaded && additional_release_channel.is_some() {
+        let channel_str = additional_release_channel.unwrap();
         ui.begin(format!(
             "Promoting {} to channel '{}'",
             ident,
             channel_str
         ))?;
 
-        if channel_str != "stable" && channel_str != "unstable" {
+        if channel_str != STABLE_CHANNEL && channel_str != UNSTABLE_CHANNEL {
```
reset: Yes

components/hab/src/cli.rs
```diff
@@ -245,7 +245,9 @@ pub fn get() -> App<'static, 'static> {
                     "Use a specific Depot URL (ex: http://depot.example.com/v1/depot)")
                 (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for the Depot")
                 (@arg CHANNEL: --channel -c +takes_value
-                    "Upload to the specified release channel (default: unstable)")
+                    "Additional release channel to upload package to. \
```
reset: I thought about this a bit and I think this is the right move for *this* command. I was in the middle of making channel required instead of optional for all of the API points in the client (hab, supervisor, etc) but I think that's just appropriate for read operation. For write I think this is completely correct to make it optional. I also like that we're now telling somebody in the client that we automatically place in unstable regardless. Good work man

christophermaier: @reset So really, this is optional just in terms of the CLI itself; if it's present, we just chain a promote operation on after the upload. I didn't take a full pass throughout the codebase, but based on what we call in the course of _this_ code path, it looks like channel is required elsewhere.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/262469824).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/262469824).

PR #2913, opened 2017-08-11
[sup] fix env variable name in sup error

srenatus:
For a service called foo-bar, with a bad config in `HAB_FOO_BAR`, this
would log the error

    hab-sup(CF)[src/manager/service/config.rs:193:20]: Unable to find valid TOML or JSON in HAB_FOO-BAR ENVVAR

with this change, the mentioned "ENVVAR" should be HAB_FOO_BAR.

⚠️ I haven't been able to verify this change locally. It builds (so what's left to ask for?), but `cargo test` in `component/sup` (via `make shell`) fails 5 of 5 tests. Also, I wasn't able to figure out how to have the actual hart use _my_ sup instead of that core one. That said, I didn't try too hard, either, as the change _looks pretty obvious_...

### DIFF
--- a/components/sup/src/error.rs
+++ b/components/sup/src/error.rs
@@ -201,8 +201,8 @@ impl fmt::Display for SupError {
                 )
             }
             Error::BadStartStyle(ref style) => format!("Unknown service start style '{}'", style),
-            Error::BadEnvConfig(ref pkg) => {
-                format!("Unable to find valid TOML or JSON in HAB_{} ENVVAR", pkg)
+            Error::BadEnvConfig(ref varname) => {
+                format!("Unable to find valid TOML or JSON in {} ENVVAR", varname)
             }
             Error::ButterflyError(ref err) => format!("Butterfly error: {}", err),
             Error::ExecCommandNotFound(ref c) => {
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -190,9 +190,7 @@ impl Cfg {
                     Err(err) => debug!("Attempted to parse env config as json and failed {}", err),
                 }
                 self.environment = None;
-                Err(sup_error!(
-                    Error::BadEnvConfig(package.name.to_ascii_uppercase())
-                ))
+                Err(sup_error!(Error::BadEnvConfig(var_name)))
             }
             Err(e) => {
                 debug!(


thesentinels:

Thank you for contributing!

srenatus:
🏓 @eeyun 😃

eeyun:
Bahah i must have been drunk when I wrote that error. Come on Ian. This
should be fine. There's not necessarily a need to change the variable name
in the error function itself but switching to actually using var_name
instead of package.name should be fine.

srenatus:
@eeyun I'm hijacking my own PR there, but: I've just seen

```
[src/manager/service/config.rs:193:20]: Unable to find valid TOML or JSON in HAB_FOO ENVVAR
```

when passing JSON with a number in it:
```
    - HAB_FOO={"postgres":{"host":"postgres","port":5432,"username":"foo"}}
```

it works if I make it `"port":"5432"`...

eeyun:
@srenatus can you verify, is the above example you posted via docker compose? If so I wonder if we can't test with the native envvar syntax that looks like this:
```
- HAB_FOO='{"postgres":{"host":"postgres","port":5432,"username":"foo"}}'
```
My guess is that not having the wrapping single tics is going to complicate things. But thats also further complicated by the compose syntax which is an extra layer of abstraction here. What it boils down to either way is that we take the entire contents of the envvar as a string and send it through the serde library to parse into KV pairs. I wonder if theres something odd happening with the DC abstraction.

srenatus:
@eeyun you've been right, my context is still docker-compose. I've tried single quotes and it seems to make things worse, neither `"port":"5432"` nor `"port":5432` work when everything is put into `'...'`. Actually, nothing seems to work when using single quotes, so, this has nothing to do with the int value, I think.

christophermaier:
@srenatus @eeyun The change in this PR looks good; it seems like the quote issue that @srenatus mentioned a few days ago is a separate issue. What do you say to merging this and opening a new issue around the quotes?

srenatus:
@christophermaier I'd be happy if this got merged. I'll create the issue.

christophermaier:
@srenatus Sounds good; thanks in advance for creating that issue, and also for fixing this one 😄 

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/265156598).

srenatus:
@christophermaier ⏩ https://github.com/habitat-sh/habitat/issues/2947

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/265156598).

Issue #2916, opened 2017-08-11
Origin key import fails If there is a ring key of the same name in the key cache


cwebberOps:
OS: macOS 10.12.6

## Repro Steps
1. Generate an origin key
```
$ hab origin key generate key-issue
» Generating origin key for key-issue
★ Generated origin key pair key-issue-20170811154838.
```
2. Generate a ring key
```
$ hab ring key generate key-issue
» Generating ring key for key-issue
★ Generated ring key pair key-issue-20170811154852.
```
3. Run a build
```
$ RUST_BACKTRACE=1 RUST_LOG=debug HAB_ORIGIN=key-issue hab studio build
DEBUG:habitat_common::ui: InputStream(stdin): { is_a_terminal(): true }
DEBUG:habitat_common::ui: OutputStream(stdout): { is_colored(): true, supports_color(): true, is_a_terminal(): true }
DEBUG:habitat_common::ui: OutputStream(stderr): { is_colored(): true, supports_color(): true, is_a_terminal(): true }
DEBUG:hab::command::studio::enter: Setting HAB_CACHE_KEY_PATH=/Users/cwebber/.hab/cache/keys
DEBUG:hab::command::studio::enter: Setting ARTIFACT_PATH=/Users/cwebber/.hab/cache/artifacts
DEBUG:hab::command::studio::docker: Running command: "/usr/local/bin/docker" "images" "habitat-docker-registry.bintray.io/studio:0.29.1" "-q"
DEBUG:hab::command::studio::docker: Found Studio Docker image locally.
DEBUG:hab::command::studio::docker: Setting container env var: "HAB_ORIGIN"='key-issue'
DEBUG:habitat_core::os::process::imp: Calling execvp(): ("/usr/local/bin/docker") ["run", "--rm", "--tty", "--interactive", "--privileged", "--env", "HAB_ORIGIN=key-issue", "--volume", "/Users/cwebber/projects/customer_portal:/src", "--volume", "/Users/cwebber/.hab/cache/keys:/hab/cache/keys", "--volume", "/Users/cwebber/.hab/cache/artifacts:/hab/cache/artifacts", "--volume", "/var/run/docker.sock:/var/run/docker.sock", "habitat-docker-registry.bintray.io/studio:0.29.1", "build"]
   hab-studio: Destroying Studio at /hab/studios/src (default)
   hab-studio: Creating Studio at /hab/studios/src (default)
   hab-studio: Importing key-issue secret origin key
✗✗✗
[...]

PR #2946, opened 2017-08-16
only get key revisions for a requested key type

mwrock:
Currently when we pull revisions for a key, we do not discrimitate for a particular type (signing, user, ring, etc) but simply grab all revisions that match the key name. In reality keys of different types can have the same name.

When it comes to the secret keys, we could simply check the file names and rework the `check_filename` logic to ignore secret key filenames with a suffix that does not match the requested key type. However, all public keys have the same extention `.pub` and changing extensions at this point would introduce serious compatibility issues.

The only sure way to validate that a key file is of a certain type is to check its content. In particular, the first line which should begin with the key type string. This pr does that.

### DIFF
--- a/components/core/src/crypto/keys/mod.rs
+++ b/components/core/src/crypto/keys/mod.rs
@@ -17,7 +17,7 @@ use std::fs;
 use std::fmt;
 use std::fs::File;
 use std::io::prelude::*;
-use std::io::BufWriter;
+use std::io::{BufReader, BufWriter};
 use std::path::{Path, PathBuf};
 use std::result;
 use std::str::FromStr;
@@ -48,6 +48,16 @@ enum KeyType {
     Sig,
     Box,
     Sym,
+}
+
+impl fmt::Display for KeyType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            KeyType::Box => write!(f, "box"),
+            KeyType::Sig => write!(f, "sig"),
+            KeyType::Sym => write!(f, "sym"),
+        }
+    }
 }
 
 #[derive(Debug, Eq, PartialEq)]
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

mwrock:
fixes #2916

components/core/src/crypto/keys/mod.rs
```diff
@@ -578,18 +601,33 @@ mod test {
...
         assert_eq!(2, revs.len());
     }
 
     #[test]
+    fn get_key_revisions_can_only_return_keys_of_specified_type() {
+        let cache = TempDir::new("key_cache").unwrap();
+        let _ = SigKeyPair::generate_pair_for_origin("foo", cache.path());
+        let revs = super::get_key_revisions("foo", cache.path(), None, &KeyType::Sig).unwrap();
+        assert_eq!(1, revs.len());
+        // we need to wait at least 1 second between generating keypairs to ensure uniqueness
```
reset: I bet this was a hair pulling moment

mwrock: totally! would they be unique? would they be the same? It was like an eternity passed in that one second.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/264977770).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/264977770).

components/core/src/crypto/keys/mod.rs
```diff
@@ -275,6 +286,17 @@ where
                 continue;
             }
         };
+
+        let file = File::open(p.path())?;
```
georgemarshall: This change is causing issues with system metadata files in the directory. For example in OS X, if there is a `.DS_Store` file, it will cause the following panic.
```
✗✗✗
✗✗✗ stream did not contain valid UTF-8
✗✗✗
```

This seems to result from opening all files in the directory and attempting to decode them blindly.

PR #2999, opened 2017-08-23
include channel of service in census entry

### LINKED ISSUE #2982: Add an event message that exposes the channel a service is associated with
It would be nice to have the channel information for a service in the census message.

mwrock:
fixes #2982 


thesentinels:

Thank you for contributing!

components/eventsrv-protocol/build.rs
```diff
@@ -26,7 +26,6 @@ fn generate_protocols() {
     let cmd = Command::new(format!("{}/bin/protoc", prefix))
         .arg("--rust_out")
         .arg(out_dir)
-        .arg("--doc_out=markdown,index.md:doc")
```
mwrock: I removed this because it was raising errors and none of the other protocol generating crates had it.

reset: This generates documentation for the protocols - I don't think I moved the needle enough on automation which is why you were getting the error. It's fine to remove for now!

components/butterfly/protocols/swim.proto
```diff
@@ -51,6 +51,7 @@ message Service {
   optional string pkg = 9;
   optional bytes cfg = 10;
   optional SysInfo sys = 12;
+  optional string channel = 13;
```
reset: I don't think we need to include the channel the service is leveraging in this rumor. This would spread the information to other Supervisors but that's not a requirement for the EventSrv protocol

reset:
@mwrock this looks solid - just the one comment about modifying the service rumor

components/butterfly/protocols/swim.proto
mwrock: I added it here because the censusmember gets its data from the service rumor so I think we need this. but i could be missing something!

mwrock: even if different supervisors have different channels, they should be different members so the data should be correct

reset: The CensusMember is a starting point for reporting information about a service to EventSrv. It doesn't need to be a 1:1 mapping, though. What I'm saying is that it doesn't make sense to gossip this information with other Supervisors, it just makes sense to ship the information up to the EventSrv. How we build the EventSrv message might need to be refactored a bit to be populated by some additional fields

bodymindarts: I agree with leaving it out of the gossip if it is not required there.
We should have a clear segregation of layers and adapters between them to serve information in a easily consumable format to the consuming layer.

mwrock: If we leave it out of the gossip, it won't work. The channel will always be empty.

mwrock: I missed @reset 's last message I'll look at refactoring and having something else feed the event server.

mwrock: @elliott-davis does the channel need to be included in the census entry message? Would it suffice to create a new event server message that captures the channel of the service group? That would allow us to leave the current census entry as is.

elliott-davis: @mwrock The more I thought about it after reading this conversation I don't think it does. I think you should be able to query the depot for the channel info based on the pkg identifier or ship it as a new message with the pkg identifier attached

mwrock: Yeah I'm gonna scrap this PR and send the channel/service information in a different message since the channel is not included in census data and not intended to be gossiped as such

mwrock:
closing this because I think the requirement of including channel in the census message is incorrect and should be sent via a separate message.

PR #3001, opened 2017-08-23
Updated post-run hook behavior 

jamessewell:
Changed post-run hook to only run after a successful health check, and re-run till success.

This is needed to ensure that the application has started before more actions are taken. 

Without this behavior many plans push 'run once after start' tasks into the health check hook with a run once Bash block - which isn't ideal. Some plans also block in post-run, which is also not ideal.

Also fixes the health_check flag, which is never updated (updates only go to the cache previously).


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -85,6 +85,7 @@ pub struct Service {
     config_renderer: CfgRenderer,
     health_check: HealthCheck,
     initialized: bool,
+    post_run: bool,
```
christophermaier: This indicates whether the last run of the `post_run` hook was successful, correct?

If so, I'd recommend having the name of this field reflect that.

components/sup/src/manager/service/mod.rs
```diff
@@ -485,11 +487,13 @@ impl Service {
 
     fn post_run(&mut self) {
         if let Some(ref hook) = self.hooks.post_run {
-            hook.run(
+            self.post_run = hook.run(
                 &self.service_group,
                 &self.pkg,
                 self.svc_encrypted_password.as_ref(),
-            );
+            ).0 == 0;
```
christophermaier: I'd prefer adding functions to the `ExitCode` struct that allow us to ask if it was successful or not, instead of doing direct (and relatively opaque) manipulations like this.

It would also let you just store the post-run hook result directly, instead of keeping a boolean, which could yield more expressive code elsewhere.

components/sup/src/manager/service/mod.rs
```diff
@@ -485,11 +487,13 @@ impl Service {
...
+        } else {
+            self.post_run = true;
```
christophermaier: This seems potentially confusing, since `self.post_run` doesn't just reflect whether the post-run hook succeeded, since it's `true` even if you don't have a post-run hook.

Not sure what would be a better formulation at this point; I'm really just thinking out loud here.

UPDATE: perhaps instead of keeping the result of the post-run hook around, which might be confusing since it's just really needed when we start up the service, we should instead track whether the service has been "fully initialized" (or whatever we want to call it). That way, the intention is clear, whether there's a post-run hook or not.

components/sup/src/manager/service/mod.rs
```diff
@@ -618,8 +622,9 @@ impl Service {
             self.initialize();
             if self.initialized {
                 self.start(launcher);
-                self.post_run();
             }
+        } else if self.health_check == HealthCheck::Ok && !self.post_run {
```
christophermaier: I think pulling this conditional out into a function named to capture the intention ("keep running a post-run hook if we're healthy and haven't run the post-run hook yet") will help to clarify the desired behavior, and make this code read more easily.

components/sup/src/manager/service/mod.rs
```diff
@@ -711,6 +716,7 @@ impl Service {
             }
         };
         self.last_health_check = Instant::now();
+        self.health_check = check_result;
```
christophermaier: heh, yeah, this seems like something to keep track of 😄

components/sup/src/manager/service/hooks.rs
```diff
@@ -54,7 +54,7 @@ where
 }
 
 #[derive(Debug, Copy, Clone)]
-pub struct ExitCode(i32);
+pub struct ExitCode(pub i32);
```
christophermaier: Not quite sure this needs to be public; see comment below on adding functions to this struct instead.

components/sup/src/manager/service/mod.rs
jamessewell: I ended up creating a `success` function for the `health_check` `enum` - happy to remove this and create the specific combination function if you think that is better.

I think it's pretty readable now though.

components/sup/src/manager/service/mod.rs
jamessewell: I didn't go down the full initialized route - I feel like `mod.rs` and `hook.rs` need a bit of refactoring - a state machine would fix these kind of issues I think.

I want to have a go at this - but it will take some time!

components/sup/src/manager/service/mod.rs
jamessewell: I ended up on post_run_exit_code - happy to change as this seems a bit weak.

christophermaier:
@jamessewell I think this may need to be rebased on top of the current master to address the sporadic tests failures you've seen in this PR.

reset:
@jamessewell could you rebase against master? It looks like you did a merge of master into your branch instead

jamessewell:
Cleaned up and moved to PR [3721](https://github.com/habitat-sh/habitat/pull/3721)

PR #3018, opened 2017-08-26
send aggregated service and census data to event server

mwrock:
This sends data from both the `Service` struct iand the `CensusMember` to the event server. The actual message is now a `ServiceUpdate` instead of a `CensusEntry`. The data is intended to represent a snapshot of key service information. Both the `Service` and `CensusMember` include key data points valuable to an event subscriber. This also exposes the `CensusMember` fields as public. This makes sense since this struct is really intended to expose key census data and consumers will need direct access to this data.


thesentinels:

Thank you for contributing!

mwrock:
fixes #2982

components/eventsrv-protocol/protocols/event.proto
```diff
@@ -81,44 +81,43 @@ message CensusEntry {
...
+  optional string topology = 12;
+  /// Update strategy used by the service
+  optional string update_strategy = 13;
+  /// Service group application name
+  optional string application = 14;
+  /// Service group environment name
+  optional string environment = 15;
   /// `true` if this service instance is the leader when in a leader topology
-  optional bool leader = 8;
+  optional bool leader = 16;
```
raskchanky: I was told once upon a time that new fields should always be added at the end of a protobuf message because you don't want to change the existing field order.  Is that still the case or did that change?

christophermaier: It's not that the physical order in the definition is important, it's the field numbers themselves. If those change then I think you get into situations where you lose your backward-compatibility.

TL;DR: `leader` should stay as field 8.

mwrock: Since the very message itself is changing from a CensusEntry to a ServiceUpdate I did not consider compatibility since that alone will break it. @reset and I discussed that a breaking change is acceptable here.

christophermaier: Ah, so it is! Yeah, this is fine; sorry!

raskchanky: Ah, thanks for clarifying.  Makes sense.

reset: Thank you all for being on guard about this! In this case, yeah, it's a do-over so we can break :)

components/sup/src/census.rs
```diff
@@ -468,27 +466,26 @@ impl CensusGroup {
...
+    pub follower: bool,
+    pub update_leader: bool,
+    pub update_follower: bool,
+    pub election_is_running: bool,
+    pub election_is_no_quorum: bool,
+    pub election_is_finished: bool,
+    pub update_election_is_running: bool,
+    pub update_election_is_no_quorum: bool,
+    pub update_election_is_finished: bool,
+    pub alive: bool,
```
reset: The health portion here (`alive`, `suspect`, `confirmed`, `departed`) should remain private

reset:
@mwrock one super small nit-pick but this is 100% on point aside from that. Excellent cleanup!

tenor-130581358

components/sup/src/census.rs
mwrock: ok. I'll need to remove that info from the `ServiceUpdate` message as well unless we think those fields are going to be needed soon by our subscriber. Currently they are not.

mwrock: ignore that. I can't read.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/269792037).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/269792037).

PR #3022, opened 2017-08-28
[sup] Don't set ownership when rendering hooks.

fnichol:
This change simplifies the writing out of rendered hooks by leaving the
owner/group to the user executing the Supervisor process (currently this
is root). The services running should not need write access to the hook
files but only permission to read and execute the hooks, which the
`set_permissions()` call allows.

Adding the ownership call creates a situation where any code testing
this functionality must be run either as the package service user
(usually `hab`) or as a root user which can set ownership on files. This
resolves an issue where several hook tests in the unit test suite cannot
be performed without running as root or under `sudo`.

Signed-off-by: Fletcher Nichol <fnichol@nichol.ca>

PR #3023, opened 2017-08-28
[sup/functional] Support running functionals as a non-root user.

fnichol:
## [sup/functional] Support running functionals as a non-root user.

For the current functional tests which exercise hook file creation, the
service user running the service is not critical and not the subject
under test. Therefore, when preparing a fixture package, the following
package metafiles will be created *unless* one is present in the fixture
directory:

* `SVC_USER`: defaults to the current user running the tests
* `SVC_GROUP`: defaults to the current user's primary group running the
  tests

This approach will pass the service user/group assertions that the
Supervisor makes when creating the service directories without changing
production/runtime behavior.

As a result, and with the inclusion of #3022, all doc, unit, and
functional tests for the Supervisor can be run with a non-root user.

## [sup/functional] Print sup output in tests if `--nocapture` is set.

[sup/functional] Print sup output in tests if `--nocapture` is set. …This change helps to better debug when functional supervsior tests are failing. Existing behavior is preserved when running the tests; that is to not output supervisor messages. However, if a developer wishes to see more test output using the standard `--nocapture` flag, then the supervisor output will be emitted.  To see the difference you can run:      cd components/sup     cargo test -- --nocapture

tenor-152562062

### DIFF
--- a/components/sup/tests/utils/mod.rs
+++ b/components/sup/tests/utils/mod.rs
@@ -14,11 +14,14 @@
 
 //! Utility functions for testing a supervisor
 
-use std::fs;
+use std::fs::{self, File};
+use std::io::Write;
 use std::path::Path;
 use std::string::ToString;
 use std::thread;
 use std::time::Duration;
+
+use hcore::os::users;
 
 pub mod fixture_root;
 pub mod hab_root;
@@ -75,6 +78,7 @@ pub fn setup_package_files<O, P, S>(
     let expanded_fixture_dir = fixture_root.expanded_package_dir(&package_name);
     let hab_pkg_path = hab_root.pkg_path(&origin_name, &package_name);
     copy_dir(&expanded_fixture_dir, &hab_pkg_path);
+    write_default_svc_user_and_group_metafiles(&hab_root, &origin_name, &package_name);
 }
 
 /// Recursively copy the contents of `source_dir` into `dest_dir`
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/sup/tests/utils/mod.rs
```diff
@@ -120,3 +125,46 @@ where
         }
     }
 }
+
+fn write_default_svc_user_and_group_metafiles<P>(hab_pkg_path: P)
```
christophermaier: Maybe add the detail from the PR message as a docstring on this function?

components/sup/tests/utils/mod.rs
```diff
@@ -120,3 +125,46 @@ where
...
+where
+    P: AsRef<Path>,
+{
+    let svc_user_metafile = hab_pkg_path.as_ref().join(MetaFile::SvcUser.to_string());
```
christophermaier: Given that other metadata file path information is encapsulated as functions of `HabRoot`, what do you think about adding a `svc_user_file` and `svc_group_file` to that module and calling them here?

(Aside: at some point, I think I'd like something like `HabRoot` to be part of the "real" code)

components/sup/tests/utils/mod.rs
```diff
@@ -120,3 +125,46 @@ where
...
+        write_metafile(
+            svc_group_metafile,
+            users::get_current_groupname()
+                .expect("Could not determine current groupname")
+                .as_str(),
+        );
+    }
+}
+
+fn write_metafile<P>(metafile: P, content: &str)
```
christophermaier: What do you think about taking a `ToString` here instead of a `&str`? 

One of the things I've been trying to do is make these testing functions as general as can be, thus the liberal use of things like `AsRef<Path>` and `ToString` in the type signatures. I was basically getting tired of functions that were too narrowly scoped, causing me to have to write unnecessarily noisy test code like `do_the_thing(String::from("blah"))` instead of just `do_the_thing("blah")`.

components/sup/tests/utils/mod.rs
fnichol: Good call!

components/sup/tests/utils/mod.rs
fnichol: I refactor as per your suggestion. It made the calling function require more context and I'm not sure I like this version better, but it puts the metafile-computing logic in with `HabRoot` which makes sense.

components/sup/tests/utils/mod.rs
fnichol: I would have used `ToString`, but in this case we don't need the extract allocation of the file contents, only a borrow. I could update the function sig. to take a `Vec<u8>` which is what is being written, but I went with the easiest to get the write descriptor closed.

christophermaier:
@thesentinels approve

thesentinels:

fnichol:
@christophermaier Nice, thanks! I just realized I needed the `--nocapture` functionality to pass my next one 😄

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/269717131).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/269717131).

PR #3028, opened 2017-08-29
Read HAB_NON_ROOT when installing package to allow for non-root usage.

bodymindarts:
Setting the environment variable `HAB_NON_ROOT` will enable the `hab` cli to install packages without root permissions.

There is still the issue that when running `HAB_NON_ROOT=true hab start x/y` the env-var doesn't get propagated from the hab cli to `hab-sup` and won't take effect if the package hasn't already been installed.

I'm submitting this PR to get feedback and start discussion about the current implementation and expected behaviour.

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -72,19 +72,14 @@ where
     P1: AsRef<Path> + ?Sized,
     P2: AsRef<Path> + ?Sized,
 {
-    if !am_i_root() {
-        if let Some(_) = env::var_os("HAB_NON_ROOT") {
-            ui.warn("Allowing install by non-root user!")?;
-            ui.br()?;
-        } else {
-            ui.warn(
-                "Installing a package requires root or administrator privileges. Please retry \
+    if env::var_os("HAB_NON_ROOT").is_none() && !am_i_root() {
+        ui.warn(
+            "Installing a package requires root or administrator privileges. Please retry \
                    this command as a super user or use a privilege-granting facility such as \
                    sudo.",
-            )?;
-            ui.br()?;
-            return Err(Error::RootRequired);
-        }
+        )?;
+        ui.br()?;
+        return Err(Error::RootRequired);
     }
 
     let cache_key_path = cache_key_path(Some(fs_root_path.as_ref()));
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -35,6 +35,7 @@
 //! * Unpack it
 //!
 
+use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -72,13 +73,18 @@ where
     P2: AsRef<Path> + ?Sized,
 {
     if !am_i_root() {
-        ui.warn(
-            "Installing a package requires root or administrator privileges. Please retry \
+        if let Some(_) = env::var_os("HAB_NON_ROOT") {
+            ui.warn("Allowing install by non-root user!")?;
+            ui.br()?;
+        } else {
+            ui.warn(
+                "Installing a package requires root or administrator privileges. Please retry \
                    this command as a super user or use a privilege-granting facility such as \
                    sudo.",
-        )?;
-        ui.br()?;
-        return Err(Error::RootRequired);
+            )?;
+            ui.br()?;
+            return Err(Error::RootRequired);
+        }
     }
 
     let cache_key_path = cache_key_path(Some(fs_root_path.as_ref()));
--- a/components/core/src/package/archive.rs
+++ b/components/core/src/package/archive.rs
@@ -224,7 +224,6 @@ impl PackageArchive {
         let writer = writer::Disk::new();
         let mut extract_options = ExtractOptions::new();
         extract_options.add(ExtractOption::Time);
-        extract_options.add(ExtractOption::Owner);
         extract_options.add(ExtractOption::Permissions);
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

fnichol:
Will take a look shortly!

bodymindarts:
Ready for review.
FYI @reset I investigated wrapping the `am_i_root` function as we discussed but after looking at each individual call site I don't think it is appropriate and I would leave the PR as is.

www/source/docs/reference/environment-vars.html.md
```diff
@@ -25,6 +25,7 @@ This is a list of all environment variables that can be used to modify the opera
 | `HAB_STUDIO_SUP` | build system | no default | Used to customize the arguments passed to an automatically launched Supervisor, or to disable the automatic launching by setting it to `false`, `no`, or `0`. |
 | `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | supervisor | 60000 | Frequency of milliseconds to check for updates when running with an [update strategy](/docs/run-packages-update-strategy) |
 | `HAB_USER` | supervisor | no default | User key to use when running with [service group encryption](/docs/run-packages-security/#service-group-encryption) |
+| `HAB_NON_ROOT` | supervisor | no default | Enable the supervisor to install and run packages as non-root user. *Warning* it is highly reckomended not to use this setting except under very specific circumstances due to associated security risks. |
```
christophermaier: Spelling: "reckomended" -> "recommended"

Adding an example of when using this variable *would* be appropriate would help users figure out if they should be using it or not.

Also, doesn't this environment variable just govern installation of packages, rather than installation *and* running as a non-root user?

components/core/src/package/archive.rs
```diff
@@ -224,7 +224,6 @@ impl PackageArchive {
         let writer = writer::Disk::new();
         let mut extract_options = ExtractOptions::new();
         extract_options.add(ExtractOption::Time);
-        extract_options.add(ExtractOption::Owner);
```
christophermaier: Why was this removed?

components/common/src/command/package/install.rs
```diff
@@ -72,13 +73,18 @@ where
     P2: AsRef<Path> + ?Sized,
 {
     if !am_i_root() {
-        ui.warn(
-            "Installing a package requires root or administrator privileges. Please retry \
+        if let Some(_) = env::var_os("HAB_NON_ROOT") {
+            ui.warn("Allowing install by non-root user!")?;
```
christophermaier: It might be useful to be explicit in the warning message that this is because `HAB_NON_ROOT` was set; that way it is absolutely clear from the message what is happening and _why_ it is happening.

components/core/src/package/archive.rs
reset: I actually probably never meant to extract the owner, this would preserve the user who built it. This was never a problem before because we always build software as root and run the Supervisor as root

components/common/src/command/package/install.rs
reset: @christophermaier I think we can just remove the warning actually. If you set this env var you know full well what you're doing and it's documented in the docs above ;)

www/source/docs/reference/environment-vars.html.md
bodymindarts: Thanks for this. I'll add an example when our cf-exporter gets merged. Then the example is more concrete and less hypothetical

components/common/src/command/package/install.rs
christophermaier: Ah, I had a bit of a brain-fart here... this is UI messaging, not logging. In that case, then yeah, I'm with you 100% @reset

bodymindarts: I removed this in a later commit

christophermaier:
@bodymindarts Did you want to change that docstring to add a concrete example in this PR, or were you saying that when the code for your concrete example is available, you'll add that?

If it's the latter, we can go ahead and merge this.

components/common/src/command/package/install.rs
christophermaier: @bodymindarts I like the new code... very concise 👍

bodymindarts:
The latter. Once cf-expert I can use that as a concrete example.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/270448502).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/270448502).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/270448502) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/270468007).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/270468007).

PR #3029, opened 2017-08-29
Add origins sidebar and package count

elliott-davis:
tenor-21751941


thesentinels:

Thank you for contributing!

elliott-davis:
screen shot 2017-08-29 at 10 56 57 am

components/builder-web/app/origins-page/OriginsPageComponent.ts
```diff
@@ -17,62 +17,84 @@ import {acceptOriginInvitation, fetchMyOriginInvitations, fetchMyOrigins}
...
-                        </a>
+        <div class="page-body">
+            <div class="page-body--main">
+                <div *ngIf="ui.loading">&nbsp;</div>
+                <div *ngIf="!ui.loading">
+                    <p *ngIf="ui.errorMessage">
+                        Failed to load origins: {{ui.errorMessage}}
                     </p>
+                    <div *ngIf="origins.size === 0 && !ui.errorMessage">
+                        <p><b>You don't currently have any origins. Let's add one now.</b></p>
```
cnunciato: `<strong>` and `em` would be better for these.

components/builder-web/app/origins-page/OriginsPageComponent.ts
```diff
@@ -17,62 +17,84 @@ import {acceptOriginInvitation, fetchMyOriginInvitations, fetchMyOrigins}
...
-                </ul>
-            </div>
-        </div>
+      </div>
     </div>`,
 })
 
 export class OriginsPageComponent implements OnInit {
-    constructor(private store: AppStore) { }
+    constructor(private store: AppStore, private router: Router) { }
```
cnunciato: Technically this is using a "private" member, but the linter isn't picking it up because it's not in a template string.   Probably better to either delegate to a public getter or call a public `navigate` handler that calls the pribate `router` instance.

components/builder-web/package.json
```diff
@@ -31,14 +31,14 @@
     "travis": "npm run test-unit"
   },
   "dependencies": {
-    "@angular/animations": "^4.3.4",
+    "@angular/animations": "https://registry.npmjs.org/@angular/animations/-/animations-4.3.5.tgz",
```
cnunciato: Why this approach and not just a version number pin? I'm sure there's a reason, I'm probably just behind on my reading. 😄

cnunciato:
I think there's supposed to be hover behavior on the origin-list items. (When I run this locally, the line items don't look clickable.)

components/builder-web/package.json
cnunciato: You probably want to update the shrinkwrap as well, otherwise this won't take hold. (I'm not seeing the progress indicator locally and I suspect that's why.)

components/builder-web/app/actions/origins.ts
```diff
@@ -163,6 +165,22 @@ export function inviteUserToOrigin(username: string, origin: string, token: stri
     };
 }
 
+export function fetchOriginsPackageCount(origins) {
+    return dispatch => {
+        origins.forEach(origin => {
+            depotApi
+                .getUnique(origin)
```
raskchanky: This can be updated to use the package_stats API now:

`/v1/depot/pkgs/origins/:origin/stats`

components/builder-web/app/app.module.ts
```diff
@@ -59,6 +61,7 @@ import { SharedModule } from "./shared/shared.module";
 @NgModule({
     imports: [
         BrowserModule,
+        CdkTableModule,
```
cnunciato: We can probably drop this guy now, at least until we need it.

PR #3037, opened 2017-08-29
[sup] Fix hook table unit test which relied on lazy static value.

fnichol:
This change fixes a unit test that would not consistently pass or fail
when run as a non-root user. Prior to this change, the
`HookTable::load()` function computed its hook path using the
`sup::fs::svc_hooks_path()` which uses a `lazy_static` constant to
determine the root filesystem path for all future calls.

Rust unit test suites run many tests concurrently in threads so that
ordering is nondeterministic. Since all threads share the same
constants, the first test to fire code which calls-and-sets the
`core::fs::FS_ROOT_PATH` constant will "win" and all other threads will
use this value. The test in question attempted to set the filesystem
root to a temporary path and so would sometimes "win" and sometimes
"lose".

This change changes the signature of `HookTable::load()` to take the
path explicitly, leading to easer test setup and a more explicit
dependency injection for the production Supervisor code.

Also, this test shouldn't fail randomly any more. Not bad, eh?

---

Ready? Ready?

It's a HOOK. TABLE.

tenor-151990844

Ha!

### DIFF
--- a/components/sup/src/manager/service/hooks.rs
+++ b/components/sup/src/manager/service/hooks.rs
@@ -759,29 +759,29 @@ pub struct HookTable {
 
 impl HookTable {
     /// Read all available hook templates from the table's package directory into the table.
-    pub fn load<T>(service_group: &ServiceGroup, templates: T) -> Self
+    pub fn load<P, T>(service_group: &ServiceGroup, templates: T, hooks_path: P) -> Self
     where
+        P: AsRef<Path>,
         T: AsRef<Path>,
     {
         let mut table = HookTable::default();
-        let hooks = fs::svc_hooks_path(service_group.service());
         if let Some(meta) = std::fs::metadata(templates.as_ref()).ok() {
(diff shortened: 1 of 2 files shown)


christophermaier:
tenor-207598660

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/269785906).

reset:
@fnichol oh it was _not_ lost on me that we were building a captain hook table ;). Great find man!

tenor-201254152

fnichol:
Wow, gifs on point! Bangarang indeed!

fnichol:
Travis is green!

tenor-143062372

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/269785906).

PR #3101, opened 2017-09-06
Add owner_name to packages table and all APIs

mwrock:
Note that this adds `OriginPackageDetail` to `OriginPackageListResponse` instead of:

1. Adding `owner_name` to `OriginPackageIdent` because the owner is not logically part of the ident
2. Replacing the `idents` field of `OriginPackageListResponse` in order to maintain backward compat

In a future release we can remove the `idents` property of `OriginPackageListResponse`.

Also note that a future PR is forthcoming that will handle adding the `owner_name`s to the existing rows in `origin_packages` based on `owner_id`.


thesentinels:

Thank you for contributing!

mwrock:
related to #3052

mwrock:
Oh I didn't notice that was how we were getting channel and target data for the list responses. I'm very concerned this is making our DB lookups very chatty and stands as a potential scalability and performance risk. This means that a single page of 50 package results produces 150 requests. I think a denormalized approach is much safer.

I'm not crazy about the `OriginPackageDetail` name. I was originally going to call it something like `OwnedPackageIdent` but that was horrible and limited the message to only accomodate the extra owner info. `OriginPackageDetail` would allow us to add more detail. I'd argue it would be worth adding target and chanel to that message. Target is already in the package table and we could get channel info via a join or a view.

chefsalim:
Good point on the fan-out, and I agree that should get addressed as well at some point, however I think we have a lot of headroom at the moment to defer that for now.  Instead, we could save a lot of code and process (including the migration of usernames to the table).

PR #3137, opened 2017-09-13
Refactoring builder services

chefsalim:
This change starts the work to eliminate the dependency on a user auth token
for the builder worker service. It modifies some of the internal APIs to not
require a session id. It also adds channel support to the Job object, and
moves part of the promotion code from the scheduler to the worker.

A couple of features are eliminated (for now - some code preserved):
1) the ability to group promote automatically by the scheduler
2) the ability to upload and promote based on a TOML file

tenor-234667977


thesentinels:

Thank you for contributing!

components/builder-worker/src/error.rs
```diff
@@ -82,6 +86,12 @@ impl From<git2::Error> for Error {
     }
 }
 
+impl From<bldr_core::Error> for Error {
+    fn from(err: bldr_core::Error) -> Error {
+        Error::BuilderCore(err)
+    }
+}
+
```
raskchanky: I've heard @reset say before that this is a pattern we should avoid going forward, and instead use things like `map_err` at the call site to convert errors from one type to another.

components/builder-worker/src/server.rs
```diff
@@ -75,6 +77,12 @@ impl Server {
                 self.fe_sock.connect(&queue)?;
             }
         }
+
+        {
+            let cfg = self.config.read().unwrap();
+            Broker::run(Self::net_ident(), cfg.route_addrs())
+        };
+
```
raskchanky: I might be mistaken, but don't you need to `join` on the broker after telling it to run?  All the other servers do that, e.g.

https://github.com/habitat-sh/habitat/blob/master/components/builder-jobsrv/src/server/mod.rs#L228
https://github.com/habitat-sh/habitat/blob/master/components/builder-api/src/server.rs#L56
https://github.com/habitat-sh/habitat/blob/master/components/builder-depot/src/server.rs#L2215
https://github.com/habitat-sh/habitat/blob/master/components/builder-scheduler/src/server/mod.rs#L213
https://github.com/habitat-sh/habitat/blob/master/components/builder-admin/src/server.rs#L55

Curiously enough, sessionsrv doesn't `join` like the rest.  I read the documentation for `std::thread::JoinHandle` which is what `Broker::run()` returns, and I gather that `join` just waits for the thread to finish.  What are the consequences of not calling `join` on the `Broker`?

components/builder-core/src/api.rs
```diff
@@ -380,11 +401,13 @@ fn do_group_promotion(
...
-            )?
-            {
-                return Err(Error::OriginAccessDenied);
+            if let Some(session_id) = session_id_opt {
+                if !check_origin_access(session_id, origin).map_err(
+                    Error::NetError,
+                )?
+                {
+                    return Err(Error::OriginAccessDenied);
+                }
```
raskchanky: If I'm following this correctly, if no session_id is passed in, then we skip the check to make sure the thing making this request has permission to access the origin in question.  Is that because the thing making the request will be another builder service?  My impression reading this code is that we want to perform this check when there is a user making this request, and thus, has a session id, but builder services don't have a session and should be implicitly trusted, which is why it's safe to skip this check here.  Is that assumption correct or have I misread this?

chefsalim: Yes, that's exactly right and that is the intent.

components/builder-worker/src/error.rs
chefsalim: Correct - this is not used, I just added it to stay consistent with the rest of the errors.

components/builder-worker/src/server.rs
chefsalim: Since we go into an infinite loop right after this, we don't (actually, cannot or even want to) do a join on the broker.  It's probably not the most optimal design, but I didn't want to start refactoring this part of the code.

raskchanky: That explains the difficulty I had when I tried to add the `Broker` to this code during the docker integration work.  Thanks for the explanation 👍

raskchanky:
tenor-166439264

raskchanky:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/275147160).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/275147160).

PR #3143, opened 2017-09-14
Download keys for verification when updating services

### LINKED ISSUE #3121: Service updater should download public keys and verify new packages
The hab sup service updater currently installs packages, however it does not verify them. The side effect of this is that the public key associated with the package does not get downloaded, which can cause the install to eventually fail.

The fix is to call verify_package similar to how the hab pkg install command does.

christophermaier:
Previously we wouldn't download keys, which meant that we could fail
to verify an updated service package if we didn't happen to already
have the key with which it was signed in our local cache.

The implementation is admittedly awful, being lifted (almost verbatim)
from `common::command::package::install`; however, we have package
installation logic scattered around in several places in the codebase,
and I am in the middle of overhauling and consolidating all of it
right now. This patch should fix this bug now, and the proper solution
will be along Real Soon Now (TM).

(I'm not making a corresponding change to
`sup::manager::self_updater`, because that is only ever updating
`core/hab-sup`, which we should always have the keys for anyway. In
any event, that package installation path will also be subsumed in the
aforementioned refactoring, so it should be a moot point soon anyway.)

Fixes #3121

### DIFF
--- a/components/sup/src/manager/service_updater.rs
+++ b/components/sup/src/manager/service_updater.rs
@@ -19,7 +19,7 @@ use std::thread;
 use std::time::Duration;
 
 use butterfly;
-use common::ui::{Status, UI};
+use common::ui::{Coloring, Status, UI};
 use depot_client;
 use env;
 use hcore::package::{PackageIdent, PackageInstall};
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/275482372).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/275482372).

PR #3177, opened 2017-09-17
Add jobs dashboard

chefsalim:
This change adds the support files for the Builder Jobs dashboard to Terraform.  This is the first part of the change, to be accompanied by a corresponding change to the cloud environments repo.

tenor-217990642

### DIFF
--- a/terraform/files/postgres.yaml
+++ b/terraform/files/postgres.yaml
@@ -1,0 +1,40 @@
+init_config:
+
+instances:
+  - host: localhost
+    port: 5432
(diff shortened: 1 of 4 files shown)


terraform/files/postgres.yaml
```diff
@@ -0,0 +1,40 @@
+  1 init_config:
```
reset: Are these line numbers supposed to be here?

seasalim: argh - good catch!

PR #3193, opened 2017-09-18
Add Dockerfile for building a self contained Builder

reset:
tenor-61819063

### DIFF
--- a/BLDR-Dockerfile
+++ b/BLDR-Dockerfile
@@ -1,0 +1,68 @@
+FROM alpine:3.6
+MAINTAINER The Habitat Maintainers <humans@habitat.sh>
+
+ARG APP_HOSTNAME=localhost
+ARG GITHUB_ADDR=github.com
+ARG GITHUB_API_URL=https://api.github.com
+ARG GITHUB_WEB_URL=https://github.com
+ARG GITHUB_CLIENT_ID UNDEFINED
+ARG GITHUB_CLIENT_SECRET UNDEFINED
+ARG GITHUB_ADMIN_TEAM 0
+ARG GITHUB_WORKER_TEAM 0
+ARG WORKER_AUTH_TOKEN UNDEFINED
+
+ENV HAB_DEPOT_CHANNEL unstable
+ENV RUST_LOG info
+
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

BLDR-Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+ENV RUST_LOG info
+
+COPY support/builder/hab-entrypoint.sh /usr/local/bin/hab-entrypoint.sh
+COPY terraform/scripts/install_base_packages.sh /tmp/install_base_packages.sh
+COPY terraform/scripts/foundation.sh /tmp/foundation.sh
+
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
+RUN addgroup -S hab && adduser -S -G hab hab
+
+RUN apk update
```
fnichol: We could set up a virtual package group [like this](https://github.com/bscott/habitat-docker/blob/e4a7c9cd5175b0fa85701602e46ce164e5e00eb6/Dockerfile#L5-L9) so that we can delete them after we are done with them [like this](https://github.com/bscott/habitat-docker/blob/e4a7c9cd5175b0fa85701602e46ce164e5e00eb6/Dockerfile#L15).

The only catch is that those packages would still exist as lower level image layers, which is why it might be useful to run the `apk update` through to `rm -Rf hab_builder_bootstrap*` as a single `RUN` command.

BUILDER_CONTAINER.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. Retrieve the container ID of the running Builder container
+```
+$ docker ps
+```
+
+1. Import certificate
+```
+$ docker exec ${instance-id} hab pkg binlink core/openssl
+$ docker exec ${instance-id} openssl s_client -showcerts -connect ${SRV_ADDR}:443 </dev/null 2>/dev/null|openssl x509 -outform PEM >> /usr/local/share/ca-certificates/github.crt
```
fnichol: You could call `hab pkg exec core/openssl openssl s_client -showcerts -connect ${SRV_ADDR}:443 …` if you want to save on binlink'ing openssl.

PR #3198, opened 2017-09-18
Unify population of the census with health state reconciliation

christophermaier:
Previously, we would populate our census from `ServiceRumor`s in one
step and update census member health from Butterfly data in another
step. We only processed our `ServiceRumor`s if we got a new one since
last time we reconciled state; similar logic holds for processing of
the Butterfly membership list for health.

However, because both pieces of data are necessary to create a fully
initialized census member (they have an indeterminate health state --
neither alive, suspect, confirmed, nor departed -- until reconciled
with Butterfly data), and because there is no guarantee of the order
in which we'll receive data, we can get into a state where we've got a
stable Butterfly membership list, but are still building out our
census membership. In this case, we'll think that we've already
reconciled all our Butterfly health information (that list isn't
changing!), resulting in census members that have indeterminate
health. Until the Butterfly-layer data changes, this data will remain
out-of-sync, causing problems in the ring. Due to the nature of the
bug, these problems will manifest randomly (sometimes everything comes
up fine, other times one or more services will fail to start because
they are binding to something that's "dead").

To address this, the logic of `update_from_member_list` is pulled into
`update_from_service_store`, with the new function being named
`populate_census` to better reflect what it is doing. Additionally, if
either the service rumors or the Butterfly membership list have
changed since the last time we processed them, we update our census
representation accordingly. `update_from_member_list` was removed as a
stand-alone function.

### DIFF
--- a/components/sup/src/census.rs
+++ b/components/sup/src/census.rs
@@ -78,10 +78,11 @@ impl CensusRing {
         service_file_rumors: &RumorStore<ServiceFileRumor>,
     ) {
         self.changed = false;
-        self.update_from_service_store(service_rumors);
+
+        self.populate_census(service_rumors, member_list);
         self.update_from_election_store(election_rumors);
         self.update_from_election_update_store(election_update_rumors);
-        self.update_from_member_list(member_list);
+
         self.update_from_service_config(service_config_rumors);
         self.update_from_service_files(service_file_rumors);
     }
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

adamhjk:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/277006803).

christophermaier:
For reference, a follow-up issue for the broader refactoring @adamhjk mentioned is #3200

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/277006803).

PR #3219, opened 2017-09-19
Refactor Builder development environment

reset:
This change greatly simplifies setting up a development environment for Habitat and Builder by fully leveraging the Habitat Studio and the unstable package channel. The new instructions are to obtain a minimal toolchain on your host, enter the Habitat Studio, and read on the on screen instructions.

I have removed/altered quite a bit of documentation that has grown over time as new people came into the project. It's important that in the future we don't add band-aid instructions to our documentation. A lot of the instructions which are being removed were symptoms of a broken or less than ideal Habitat implementation. Since we are the creators of Habitat we must also be the primary consumers of Habitat. If you find something broken, fix it or open an issue in the tracker if you aren't able to at the moment. Point the documentation to the issue and the issue to the documentation.

I have also removed a lot of "preference" documentation where we included solutions to the same problem but in different flavours. We should avoid adding this kind of documentation, we need to provide a golden path and that path must be driven by Habitat. Our tools are at the place where this is possible but only if we don't stop now.

tenor-97994400


thesentinels:

Thank you for contributing!

christophermaier:
tenor-53946355

.studiorc
```diff
@@ -0,0 +1,288 @@
...
+export GITHUB_API_URL
+export GITHUB_WEB_URL
+export GITHUB_CLIENT_ID
+export GITHUB_CLIENT_SECRET
+export GITHUB_ADMIN_TEAM
+export GITHUB_WORKER_TEAM
+export WORKER_AUTH_TOKEN
+export SSL_CERT_FILE
+
+function init-datastore {
```
fnichol: You can drop the `function` from these…………functions.

.studiorc
```diff
@@ -0,0 +1,288 @@
...
+function configure {
+  sup-term
+  sleep 2
+  /src/support/builder/config.sh
+  sleep 2
+  sup-run
+}
+
+function install-packages {
+  hab pkg install core/cacerts core/busybox-static core/hab-launcher core/postgresql -b -c stable
```
fnichol: There's going to be a race with `core/cacerts`, `core/busybox-static`, `core/hab-launcher` when the Supervisor is being installed and started in the background. We should be able to skip these unless I'm mistaken.

Could this be for the `-b` (i.e. `--binlink`) feature on install? We'll want to be careful about binlinking all of busybox over what's already on `PATH`. Maybe a call to binlink on certain programs?

Having `psql` around is pretty great though!

.studiorc
```diff
@@ -0,0 +1,288 @@
...
+}
+
+function build-habitat {
+  build-hab
+  build-launcher
+  build-sup
+}
+
+function build-hab {
+  build "components/hab"
```
fnichol: Assuming that we're entering the Studio from the root of the project, you should be safe to call `build "/src/components/hab` so that these functions can be run from any directory.

.studiorc
```diff
@@ -0,0 +1,288 @@
...
+  * stop-builder - stop all Builder services
+  * start-{svc} - start a specfic Builder service. See the
+    above list of build targets for available services
+  * stop-{svc} - stop a specific Builder service
+DOCS
+}
+
+install-packages
+
+HAB_DEPOT_CHANNEL=unstable
```
fnichol: maybe pull this list to the top to make it clear what gets set?

.studiorc
```diff
@@ -0,0 +1,288 @@
...
+APP_HOSTNAME=localhost:3000
+GITHUB_API_URL=https://api.github.com
+GITHUB_WEB_URL=https://github.com
+GITHUB_CLIENT_ID=0c2f738a7d0bd300de10
+GITHUB_CLIENT_SECRET=438223113eeb6e7edf2d2f91a232b72de72b9bdf
+GITHUB_ADMIN_TEAM=1995301
+GITHUB_WORKER_TEAM=1995301
+SSL_CERT_FILE=$(hab pkg path core/cacerts)/ssl/cert.pem
```
fnichol: Cannot wait to purge this from the Earth once we have the next `core/openssl` rebuilt.

.studiorc
```diff
@@ -0,0 +1,288 @@
...
+
+init-datastore
```
fnichol: Maybe wrap in a check for `HAB_STUDIO_SUP=false` for these actions as a way to skip them if you don't need builder running (a way to save ram/cpu when working only on sup/hab/launcher).

raskchanky:
My primary concern with this is going back to the incredibly slow Docker dev env.  Does this mitigate that somehow?  I've been working in a totally self-contained VM with no file sharing for a few months now and I can't imagine going back to a development environment with file sharing, especially a Docker based one that's incredibly slow.  I like everything else about this though.

reset:
@raskchanky You're totally right, I added a note to the instructions which suggests somebody use a VM or a Linux machine to get around IO performance issues

reset:
@fnichol and I had a discussion about how we could improve upon this even more. Let's hold off and ensure that at a minimum you can run `cargo build`, `cargo check`, and `cargo test` inside the studio. The goal here is to ensure all workstation setup is encapsulated inside the Studio using the very same environment which we use to build the software.

In the mean time, I encourage especially the front-end developers to give this dev environment a shot. It's already a hell of a lot better for you :)

elliott-davis:
This doesn't seem like it will affect anyone who has an established dev workflow but would greatly help the UI devs move faster. Any chance we could rebase and merge this ?

reset:
@elliott-davis I rebased and pushed. I think we're close on being able to use this but I'm not 100% sure that it wouldn't affect @raskchanky and a few other's workflow that might still be using the Procfile. Right now I enter a dev shell and then docker exec into it as many times as I need another service to run. What we really need is some Rust Scaffolding and to have Scaffolding populate the environment

reset:
Completed by @elliott-davis!

PR #3230, opened 2017-09-20
Fix out of order messaging in RouteConn

reset:
Requests sent by RouteConn to another node would go through a bad
code path not shared by RouteClient. This change hits a refactor
that I had forgot to make before merging things in from my previous
net refactor.

RouteConn now leverages RouteClient for requests instead of
RouteReqConn. Both RouteReqConn and RouteRepConn have been removed
in favor of RouteConn and RouteClient.

RouteConn is a full duplex connection to the RouteSrv where requests
and responses can be received at once while RouteClient is single
duplex and can only make a request and wait for a response.

tenor-207894026

### DIFF
--- a/components/net/src/conn/mod.rs
+++ b/components/net/src/conn/mod.rs
@@ -18,8 +18,6 @@
 
 mod error;
 
-use std::ops::{Deref, DerefMut};
-
 use protobuf;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #3239, opened 2017-09-20
Remove all lines between Builder & Depot

reset:
* Hab CLI has had it's documentation updated to reference Builder
  instead of the Depot. This is primarily present in any commands
  which accept a `--url` and/or a `--channel` flag
* `HAB_DEPOT_URL` renamed to `HAB_BLDR_URL`. We will still honor
  `HAB_DEPOT_URL` but it is no longer present in documentation
* `HAB_DEPOT_CHANNEL` renamed to `HAB_BLDR_CHANNEL`. This is
  still supported the same as `HAB_DEPOT_URL`
* The value of `--url` or `HAB_BLDR_URL` no longer expects a path
  portion of the URL. For example, you no longer need to specify
  `--url http://localhost:9636/v1/depot`. This was leaking the API
  version and it's API in the CLI causing issues for clients which
  were to be configured to speak to a non-public Builder endpoint
* Large portions of the documentation have been updated to reference
  Builder as an entity. Documentation about the "Depot component"
  has not been altered but it should be given a refactor
* Builder-Web configuration no longer expects a url containing a
  path. This is identicle to the changes to the value of `--url` on
  in the CLI
* Renamed "Depot" to "Builder" on Builder-Web

tenor-69835060


thesentinels:

Thank you for contributing!

components/builder-web/app/BuilderApiClient.ts
```diff
@@ -19,7 +19,7 @@ import { GitHubApiClient } from "./GitHubApiClient";
 
 export class BuilderApiClient {
     private headers;
-    private urlPrefix: string = config["habitat_api_url"];
+    private urlPrefix: string = `${config["habitat_api_url"]}/v1`;
```
mgamini: Maybe `/v1` should just be in the config?

components/builder-web/habitat.conf.sample.js
```diff
@@ -4,7 +4,7 @@ habitatConfig({
     // of the habitat repo, this will be localhost (if running Docker for Mac or
     // Linux) or the result of `$(docker-machine ip default)` if using Docker
     // in a virtual Machine.
-    habitat_api_url: "http://localhost:9636/v1",
+    habitat_api_url: "http://localhost:9636",
```
mgamini: Oh. This was on purpose. Why?

components/builder-web/app/BuilderApiClient.ts
reset: It's on purpose because this doesn't belong in the config. This is part of the "protocol" and not part of the configuration. The client should hardcode which protocol it expects as it is _not_ actually configurable.

For example, say we had added a `v2`. Well, I could configure this client to talk to `v2` but it would have absolutely no idea how to handle it :)

reset:
@fnichol per our conversation on the walk out of Rakuten, take a peek at this post-merge for me. I'm going to get this in now so I can get moving on testing in acceptance!

www/source/docs/internals-bootstrapping.html.md
```diff
@@ -113,7 +113,7 @@ In this stage, we rebuild all the base packages needed by Habitat using the tool
 export STUDIO_TYPE=default
 export HAB_STUDIO_ROOT=/hab/studios/stage2
 export HAB_ORIGIN=core
-export HAB_DEPOT_URL=http://127.0.0.1:9632/v1
+export HAB_BLDR_URL=http://127.0.0.1:9632/v1
```
fnichol: Drop the `/v1`?

fnichol:
That’s looking pretty complete to me

www/source/docs/internals-bootstrapping.html.md
reset: Holy crap, good catch. These are ancient docs, that port doesn't even get used anymore :). I'll get this in the next one

components/sup/tests/fixtures/compilation/config-changes-hooks-do-not.spec
```diff
@@ -1,6 +1,6 @@
 ident = "sup-integration-test/config-changes-hooks-do-not"
 group = "default"
-depot_url = "http://hab.sup.test/v1/depot"
```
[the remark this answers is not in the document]
reset: Got it in a future one

components/hab/src/cli.rs
```diff
@@ -83,8 +83,8 @@ pub fn get() -> App<'static, 'static> {
                 (about: "Reads a stdin stream containing plain text and outputs \
                     an encrypted representation")
                 (aliases: &["e", "en", "enc", "encr", "encry"])
-                (@arg DEPOT_URL: -u --url +takes_value {valid_url}
-                    "Use a specific Depot URL (ex: http://depot.example.com/v1/depot)")
+                (@arg BLDR_URL: -u --url +takes_value {valid_url}
+                    "Specify an alternate Builder endpoint (ex: https://bldr.habitat.sh)")
```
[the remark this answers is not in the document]
reset: Gonna get this and the others in the next one, great idea

PR #3242, opened 2017-09-20
Migrate Service Spec files using an older format

reset:
This change will rewrite any loaded spec file from spec watcher which
will ensure that spec files are "migrated" between each Supervisor
release. Spec files are only written on `hab svc load`, so it is
important for us to migrate them as they enter the system on read
or they will never get updated.

@fnichol #3239 this one too :)


thesentinels:

Thank you for contributing!

fnichol:
That’s a good call. If we need something like these spec files into the future they could be pushed even further to a binary format

PR #3248, opened 2017-09-21
Add --force option to hab pkg binlink

echohack:
This PR adds additional safety to the `hab pkg binlink` command.

Currently when running `hab pkg binlink core/busybox-static` the command will automatically overwrite any existing files with their busybox-static counterparts. Overwriting is potentially dangerous, so we should skip overwriting those binaries, and tell the user to run the command again with `-f` to overwrite.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -26,6 +26,10 @@ pub fn get() -> App<'static, 'static> {
         .about("Alias for 'config apply'")
         .aliases(&["ap", "app", "appl"])
         .setting(AppSettings::Hidden);
+    let alias_binlink = sub_pkg_binlink()
+        .about("Alias for 'pkg binlink'")
+        .aliases(&["b", "bi", "bin", "binl", "binli", "binlin"])
+        .setting(AppSettings::Hidden);
     let alias_install = sub_pkg_install()
         .about("Alias for 'pkg install'")
         .aliases(&["i", "in", "ins", "inst", "insta", "instal"])
@@ -172,16 +176,7 @@ pub fn get() -> App<'static, 'static> {
             (about: "Commands relating to Habitat packages")
             (aliases: &["p", "pk", "package"])
             (@setting ArgRequiredElseHelp)
-            (@subcommand binlink =>
-                (about: "Creates a symlink for a package binary in a common 'PATH' location")
-                (aliases: &["bi", "bin", "binl", "binli", "binlin"])
-                (@arg PKG_IDENT: +required +takes_value
-                    "A package identifier (ex: core/redis, core/busybox-static/1.42.2)")
-                (@arg BINARY: +takes_value
-                    "The command to symlink (ex: bash)")
-                (@arg DEST_DIR: -d --dest +takes_value
-                    "Sets the destination directory (default: /bin)")
-            )
+            (subcommand: sub_pkg_binlink())
             (@subcommand config =>
                 (about: "Displays the default configuration options for a service")
                 (aliases: &["conf", "cfg"])
@@ -435,6 +430,7 @@ pub fn get() -> App<'static, 'static> {
             )
         )
         (subcommand: alias_apply)
+        (subcommand: alias_binlink)
         (subcommand: alias_install)
         (subcommand: alias_run())
         (subcommand: alias_setup)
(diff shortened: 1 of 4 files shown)


components/hab/src/cli.rs
```diff
@@ -172,16 +176,7 @@ pub fn get() -> App<'static, 'static> {
             (about: "Commands relating to Habitat packages")
             (aliases: &["p", "pk", "package"])
             (@setting ArgRequiredElseHelp)
-            (@subcommand binlink =>
```
reset: Why the switch from macro -> builder pattern? We prefer the macro syntax and only fallback to builder when we need it to dynamically generate things or load CLI bits from other crates

components/hab/src/cli.rs
```diff
@@ -435,6 +430,7 @@ pub fn get() -> App<'static, 'static> {
             )
         )
         (subcommand: alias_apply)
+        (subcommand: alias_binlink)
```
reset: We try to keep the root level aliasing to a minimum. If you think we should add this one to the root we should probably cover it in another PR and have a discussion there

fnichol: Agreed, this can stay only under `hab pkg` for now, otherwise we start to massively widen the first-level subcommand space.

components/hab/src/cli.rs
```diff
@@ -577,6 +573,24 @@ fn sub_pkg_build() -> App<'static, 'static> {
...
+            "A package identifier (ex: core/redis, core/busybox-static/1.42.2)")
+        (@arg BINARY: +takes_value
+            "The command to symlink (ex: bash)")
+        (@arg DEST_DIR: -d --dest +takes_value
+            "Sets the destination directory (default: /bin)")
+    );
+    sub.arg(
+        Arg::with_name("FORCE")
+            .help("Overwrites existing files")
+            .short("f")
```
fnichol: A small suggestion: would we be able to use `-F` rather than the lowercase form? I see little `f` and think of “file”.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -53,21 +54,35 @@ pub fn start(
...
     match fs::read_link(&dst) {
         Ok(path) => {
-            if path != src {
+            if force && path != src {
                 fs::remove_file(&dst)?;
                 filesystem::symlink(&src, &dst)?;
+                ui.end(&ui_symlinked)?;
+            } else if path != src {
+                ui.warn(
+                    format!("Skipping symlink because {} already exists at {}. Use -f to overwrite",
```
fnichol: It would be a bit clearer to tell the user to use `--force` (i.e. the long form) in this message.

Additionally, with the `has pkg install` command, this would be reported and there currently isn’t such a flag which would be confusing. I guess we’d default to non-forcing on package install but tell the user to explicitly call `hab pkg binlink` in that case?

echohack:
Thanks for the reviews <3

- I reverted to the macro form. It's the right way to do it.
- I exposed the --force flag on `hab pkg install`. This was already wired up, I just forgot to expose it. So the behavior for `hab pkg install -b origin/pkg_name` will be the consistent with regular binlink. Now you can `hab pkg install -bf origin/pkg_name`. That's nice!
- telling the user --force adds more clarify for sure. let's do that.

- I opted to keep `-f` instead of `-F`. We don't have many capital flags under hab commands (just -V and -D), and I think we should use the lowercase space first.

Most interesting is how `ln` works:

```
     -F    If the target file already exists and is a directory, then remove it
           so that the link may occur.  The -F option should be used with either
           -f or -i options.  If none is specified, -f is implied.  The -F option
           is a no-op unless -s option is specified.

     -f    If the target file already exists, then unlink it so that the link may
           occur.  (The -f option overrides any previous -i options.)
```

Based on this, I think `-f` makes the most sense, because it will be familiar to those who've used `ln`.

reset:
@thesentinels approve

tenor-105757474

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/279603620).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/279603620).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/279603620) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/279603620).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/279603620).

PR #3265, opened 2017-09-21
Add config option for tuning HTTP handler thread count

reset:
Our front-end servers have been bumped up significantly in production which no longer requires the hard setting of 128. This number is actually lower than what we can now support!

* Add `http.handler_count` config option to API/Admin
* Add `handler_count()` to GatewayCfg
* Set default handler count to `num_cpu() * 8`. This is a sensible
  default for most occasions and can now be easily overidden


thesentinels:

Thank you for contributing!

PR #3296, opened 2017-09-23
Schedule a registration retry OnError from RouteSrv

reset:
The current behaviour is to shutdown a backend service connected
to RouteSrv if RouteSrv sends a registration conflict error. This
happens when you attempt to register to the RouteSrv and it rejects
your registration.

This change keeps the server running and replies to a REG_NOT_FOUND
with a registration message


thesentinels:

Thank you for contributing!

PR #3300, opened 2017-09-23
Improve dead peer detection in RouteSrv and App

reset:
* Add Router to active router's table in `App` on-connect instead
  of when the App is started. This will ensure the connection is
  complete before attempting to route messages to a Router which
  we have not completed a connection to
* Remove Router/App if we attempt to route and receive a
  HostUnreachable error
* Turned on zmq socket option "router_mandatory" on both sides of
  the App->RouteSrv connection to ensure we receive a HostUnreachable
  error on send failure
* Identify Router socket with socket::srv_ident


thesentinels:

Thank you for contributing!

PR #3324, opened 2017-09-25
Add nginx status and logging for builder API

chefsalim:
Update the nginx configuration for builder API to add an internal status endpoint, along with custom nginx call logging.

tenor-209923620

### DIFF
--- a/components/builder-api-proxy/config/nginx.conf
+++ b/components/builder-api-proxy/config/nginx.conf
@@ -56,6 +56,10 @@ http {
   proxy_cache_path {{pkg.svc_var_path}}/cache levels=1:2 keys_zone=my_cache:10m max_size=10g inactive=60m use_temp_path=off;
   proxy_cache_key "$scheme$proxy_host$uri$is_args$args $http_user_agent";
 
+  log_format nginx '\$remote_addr - \$remote_user [\$time_local] '
+                   '"\$request" \$status \$body_bytes_sent \$request_time '
+                   '"\$http_referer" "\$http_user_agent"';
+
   upstream backend {
     server 127.0.0.1:{{bind.http.first.cfg.port}};
   }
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

nellshamrell:
👍

components/builder-api-proxy/config/nginx.conf
```diff
@@ -70,11 +74,21 @@ http {
       rewrite ^(.*)$ https://$host$1 permanent;
     }
 
+    access_log {{pkg.svc_path}}/logs/host.access.log nginx;
```
reset: Will this rotate appropriately? What we really want to do is ensure that logs from journald get into a log aggregator like Sumo Logic (unless data dog also does that)

reset: Or maybe since it's the access log we really don't care for it in an aggregator?

chefsalim: The Datadog agent will be able to parse it and we'll get the stats that way. The rotation is something I will figure out shortly - this is likely the first of a few tweaks on this.

fnichol: Yeah, we'll want to be careful about rotation or max file size. We've already learned that high traffic directly affects syslog file sizes which isn't surprising.

fnichol: Yuck, looks like nginx doesn't handle this itself: https://www.nginx.com/resources/wiki/start/topics/examples/logrotation/

chefsalim: Yeah, I'm leaning toward using logrotate - it looks fairly straightforward to wire up.

reset: The other option is to rely solely on your aggregator. I did this on my last large project and it worked swimmingly

chefsalim: Yeah the main reason atm to log to file with nginx is to get the response times

fnichol:
Would it be possible to also include #3063 in this PR?

PR #3331, opened 2017-09-26
Fix Terraform resource churn on API instances

reset:
Move EBS volume definition to a separate volume block. For the time
being we will specify this volume manually with an ID into a
variable. This isn't a long term solution, eventually we will move
the package data to an S3 bucket and not need this volume anymore.
For now, let's prevent future resource churn which occurs when we
manually mount volumes to our API instances

### DIFF
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -23,12 +23,6 @@ resource "aws_instance" "api" {
     agent       = "${var.connection_agent}"
   }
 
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

reset:
It runs on acceptance and it'll be part of a migration on live which shouldn't be too painful

PR #3371, opened 2017-09-27
Automatically determine jobsrv key_dir in Habitat plan

reset:
Use `{{pkg.svc_files_path}}` to determine the path for JobSrv's key_dir configuration setting

tenor-173587473

### DIFF
--- a/components/builder-jobsrv/habitat/config/config.toml
+++ b/components/builder-jobsrv/habitat/config/config.toml
@@ -1,3 +1,5 @@
+key_dir = "{{pkg.svc_files_path}}"
+
 {{~#eachAlive bind.router.members as |member|}}
 [[routers]]
 host = "{{member.sys.ip}}"


thesentinels:

Thank you for contributing!

reset:
Kind of, we were actually just configuring it by the ring but no value was set in `default.toml` and the default value in the config is actually bad - we should revisit all of them and set them to `/tmp` or a blank string and have the config fail validation if it hasn't been set

fnichol:
I used a trick in the Ruby Scaffolding to have an `init` hook check for a required value and die if not present. Like:

```sh
if [ -z "{{cfg.really_important}}" ]; then # if value is empty/unset
  echo "Required config setting 'really_important' is not set, aborting..."
  exit 1
fi
```

reset:
That would work for sure but since we control the software underneath it's better just to fix it there. It was a corner that I cut when I wrote all the config file parsing logic. Any application that takes input has to validate it and we just aren't doing that right now 😞

PR #3397, opened 2017-09-29
[Builder] Drop `/v1` from `app_url` when building dev container.

fnichol:
I believe this was missed in #3239, or was merged afterwards--something like that. This is used when building the Builder development container.

@reset Does this look right to you?


thesentinels:

Thank you for contributing!

reset:
Yep this is right! There might be some additional issues, I have a branch stashed with some changes if you're looking to get it going

PR #3410, opened 2017-09-29
Add advanced plan writing guide

tashimi:
Signed-off-by: Tasha Drew <tasha.drew@gmail.com>

### DIFF
--- a/www/data/docs_sidebar.yml
+++ b/www/data/docs_sidebar.yml
@@ -59,6 +59,8 @@ sidebar_links:
       link: "#ecs-and-habitat"
     - title: "Apache Mesos and DC/OS"
       link: "#mesos-dcos"
+    - title: "Advanced Plan.sh Writing"
+      link: "#planwriting-guide"
     - title: Iterative Development
       link: "#iterative-development"
     - title: "Binary Wrapper Packages"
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

reset:
tenor-198956901

themightychris:
`The value of name should exactly match the name of the project it represents ` @tashimi could you elaborate on what, if any, are the exceptions to this? e.g. projects with explicit caps or spaces/hyphens in the same? A lot of xorg projects have weird but significant caps and i've seen people do a few different things with them

tashimi:
@themightychris These are guide rails rather than absolutes, but I would say do what you need to do to clearly communicate the original source of the project and software so that a user can read your plan and be sure it represents what they're looking for, and that they don't need to go build a new one because it already exists

themightychris:
@tashimi none of the core plans have caps though, could these docs say for sure one way or the other whether caps are intended to be ok? It will be valuable if there's ever an issue with caps being preserved to be able to say "that needs to be fixed" or "you never should have started using caps in your plan names"

See https://bldr.habitat.sh/#/pkgs/xorg

reset:
@themightychris using capital letters is fine and we'll always preserve capitalization for a display name. Packages are stored in a case-insensitive way so `MYPACKAGE` and `mypackage` should be the same package. We haven't put in the work yet to ensure search is case-insensitive but I'm sure we'll take a pass at search in the future

PR #3411, opened 2017-09-30
Temporarily use old & new forms of `hab pkg binlink`.

### LINKED ISSUE #3323: Account for binlink --force change throughout codebase
Now that #3248 is merged we still need to audit where `hab pkg binlink` gets called and potentially update calls where `--force` behavior is required. Locations such as the `install.sh` hab installer, provisioning, etc.

fnichol:
This change calls the old behavior (to force override existing targets)
and the new behavior (an explicit force behavior via flag) in a couple
parts of the codebase. The temporary fixes can be removed after a short
period has elapsed when versions of Habitat 0.33.2 and older are
considered "really old".

Note that in the places we're updating, we expect that the binlink target is updated, especially in installation scripts. Otherwise with the new behavior, the command will warn, not overwrite the target and a system would continue calling the previous version of `hab`.

Closes #3323

### DIFF
--- a/terraform/scripts/install_base_packages.sh
+++ b/terraform/scripts/install_base_packages.sh
@@ -122,4 +122,11 @@ done
 
 # Now we ensure that the hab binary being used on the system is the
 # one that we just installed.
-${hab_bootstrap_bin} pkg binlink core/hab hab
+#
+# TODO fn: The updated binlink behavior is to skip targets that already exist
+# so we want to use the `--force` flag. Unfortunetly, old versions of `hab`
+# don't have this flag. For now, we'll run with the new flag and fall back to
+# running the older behavior. This can be removed at a future date when we no
+# lnger are worrying about Habitat versions 0.33.2 and older. (2017-09-29)
+${hab_bootstrap_bin} pkg binlink core/hab hab --force \
+  || ${hab_bootstrap_bin} pkg binlink core/hab hab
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

fnichol:
We'll want this PR merged into master before the next release is fully published to Bintray. However, it doesn't need to roll into the release itself.

terraform/scripts/install_base_packages.sh
```diff
@@ -122,4 +122,11 @@ done
...
 # one that we just installed.
-${hab_bootstrap_bin} pkg binlink core/hab hab
+#
+# TODO fn: The updated binlink behavior is to skip targets that already exist
+# so we want to use the `--force` flag. Unfortunetly, old versions of `hab`
+# don't have this flag. For now, we'll run with the new flag and fall back to
+# running the older behavior. This can be removed at a future date when we no
+# lnger are worrying about Habitat versions 0.33.2 and older. (2017-09-29)
+${hab_bootstrap_bin} pkg binlink core/hab hab --force \
+  || ${hab_bootstrap_bin} pkg binlink core/hab hab
```
christophermaier: This is probably not needed here, since we'll be running this on newly-provisioned instances. Probably won't hurt anything, though.

fnichol: I'm pretty sure you're right, but didn't want to take the chance if we needed this in an emergency.

fnichol:
@christophermaier I agree, I'm hoping within a month, or a certain number of releases (2-4) this should be irrelavent and would only apply to those trying to install a pinned version of Habitat. As the Chef cookbook does this now, I wanted to ensure both modes work for now.

PR #3424, opened 2017-09-30
[jobsrv] Export `log_port` for discovery by workers.

fnichol:
It appears that we were missing the jobsrv's worker log ingestion port as exported config which allows workers to update if this port number changes from its default. This change would require both `builder-jobsrv` and `builder-worker` to be built and deployed together.


thesentinels:

Thank you for contributing!

reset:
tenor-200958404

PR #3428, opened 2017-09-30
persist github installation id for private repos and clone with its auth token

mwrock:
This backs out the username/personal access token persistence and replaces it using github application access tokens.


thesentinels:

Thank you for contributing!

mwrock:
Just need to test end to end.

components/builder-protocol/protocols/originsrv.proto
```diff
@@ -341,8 +341,7 @@ message OriginProject {
   optional uint64 owner_id = 7;
   optional string vcs_type = 8;
   optional string vcs_data = 9;
-  optional string vcs_auth_token = 10;
-  optional string vcs_username = 11;
+  optional uint64 vcs_installation_id = 10;
```
reset: We can't reuse field numbers so you'd want to make this `12`

reset: It's intended that this protocol message is polymorphic across many different types of VCS authentication. I don't think I'd suggest the extra work right now since we aren't trying to support more than one VCS, but installation_id is a thing specific to GitHub. Go with what you got but we'll maybe want to revisit this later

components/builder-worker/habitat/config/config.toml
```diff
@@ -3,6 +3,8 @@ auto_publish = {{cfg.auto_publish}}
 data_path = "{{pkg.svc_data_path}}"
 bldr_channel = "{{cfg.bldr_channel}}"
 bldr_url = "{{bind.depot.first.cfg.url}}"
+github_app_id = 5565
```
reset: We'll want to make this configurable since we'll have an app_id for acceptance, too

components/builder-originsrv/src/data_store.rs
```diff
@@ -180,31 +176,23 @@ impl DataStore {
...
+        let install_id: Option<i64> = {
+            if project.has_vcs_installation_id() {
+                Some(project.get_vcs_installation_id() as i64)
             } else {
                 None
             }
         };
         let rows = conn.query(
-            "SELECT * FROM insert_origin_project_v2($1, $2, $3, $4, $5, $6, $7, $8)",
+            "SELECT * FROM insert_origin_project_v2($1, $2, $3, $4, $5, $6, $7)",
```
reset: We'll need to create a new stored procedure here (`_v3`). The function signature of a stored procedure needs to remain immutable

components/builder-worker/src/config.rs
```diff
@@ -68,6 +72,10 @@ impl Default for Config {
             bldr_channel: String::from("unstable"),
             bldr_url: url::default_bldr_url(),
             jobsrv: vec![JobSrvAddr::default()],
+            github_app_id: 5565,
+            github_private_key_path: PathBuf::from(
+                "/hab/svc/hab-jobsrv/files/github_app_private_key.pem",
```
reset: You probably want to set this default to `/hab/svc/builder-worker/files/github_app_private_key.pem` or just `/tmp`

components/builder-worker/src/vcs.rs
```diff
@@ -12,161 +12,80 @@
...
-                git2::Repository::clone(&(self.url()?).as_str(), path)?;
+                git2::Repository::clone(&(self.url(token)?).as_str(), path)?;
                 Ok(())
             }
             _ => panic!("Unknown vcs type"),
         }
     }
 
-    pub fn url(&self) -> Result<Url> {
+    pub fn url(&self, token: Option<String>) -> Result<Url> {
```
reset: This can be `token: Option<&str>` if you call `token.as_ref()` on line 63. You're only using the ref on line 82 below so you don't need a fully allocated String

reset:
tenor-162581707

Can't wait for this to land!

fnichol:
@mwrock Looks like you'll need to rebase on top of my work, sorry 'bout that!

components/builder-protocol/protocols/originsrv.proto
mwrock: yeah I think we will need to make deeper changes to make the polymorphism more natural but its really too early until we have at least 2 VCSs.

mwrock: is it ok to break back compat here since we have not deployed the above 2 fields to prod?

components/builder-originsrv/src/data_store.rs
mwrock: since v2 has not been deployed to prod, do we need a v3?

components/builder-worker/src/config.rs
mwrock: oops. thats what I meant to do

components/builder-originsrv/src/data_store.rs
fnichol: this could still break acceptance and development environments, no?

reset: Yeah, it'll still break acceptance unless we do some manual surgery on it to fix it. We can, and should, go through and remove old stored procedures. This will get cleaned up eventually when no other servers rely on it.

components/builder-protocol/protocols/originsrv.proto
reset: No, we need to consider acceptance production as well. Theoretically it's ok because it would only cause a super small outage lasting less then a few minutes, but it's best to just get in the Habitat of versioning the protocols and stored procedures properly

mwrock:
Note that I added the `vnd.github.machine-man-preview` media `ACCEPT` header to all github api calls. This supports GH application integration and I confirmed that it also works fine with normal oauth calls. It was far easier to pass this along on all calls because depending on if we are authenticating on behalf of an app installation, we need to call the same endpoints for plan.sh validation.

components/builder-originsrv/src/migrations/origin_projects.rs
```diff
@@ -116,11 +116,7 @@ pub fn migrate(migrator: &mut Migrator) -> SrvResult<()> {
                  $$ LANGUAGE plpgsql VOLATILE"#)?;
     migrator.migrate(
         "originsrv",
-        r#"ALTER TABLE IF EXISTS origin_projects ADD COLUMN IF NOT EXISTS vcs_auth_token text;"#,
-    )?;
-    migrator.migrate(
-        "originsrv",
-        r#"ALTER TABLE IF EXISTS origin_projects ADD COLUMN IF NOT EXISTS vcs_username text;"#,
+        r#"ALTER TABLE IF EXISTS origin_projects ADD COLUMN IF NOT EXISTS vcs_installation_id bigint;"#,
```
reset: This can't be removed until insert_origin_project_v1 & v2 are removed

reset:
@mwrock one last comment in the migrations and a rebase and then we should merge this one in!

reset:
@thesentinels approve

tenor-133958603

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/282144068).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/282144068).

Issue #3342, opened 2017-09-26
Integrate new docker publish into builder-worker


reset:
Now that we've finished the new `hab pkg export docker` command we'll need to wire this into the Builder-Worker. Creds will need to be recorded on the origin and assigned to the project through the public API for them to be made available to the worker.

Related: #3259, #3109

PR #3432, opened 2017-10-01
[builder-worker] Preparatory refactoring for exporters in workers

fnichol:
## [build] Refactor & centralize common Rust build script behavior

This change extracts common code used in `build.rs` scripts to set
program versions for Habitat (userland) and Builder (hosted) components,
as well as generating protocol stubs.

While it is possible to use external crates in build scripts, it
involves adding a `build-dependencies` section in every Cargo.toml.
Instead, we're opting for a slightly "clever" of inlining Rust code at
compile time via the `include!()` macro. Currently, the two inlined
"libraies" are:

* `components/libbuild.rs` - contains common versioning logic for both
Builder and Habitat artifacts
* `components/libbuild-protocols.rs` - contains common logic to generate
Rust stubs for protocol files.

Additionally, the following refactorings we undertaken:

* Rename `components/bldr-build.rs` to `components/build-builder.rs` and
update call sites in related `Cargo.toml` files
* Rename `components/build.rs` to `components/builder-habitat.rs` and
update call sites in related `Cargo.toml` files

---

So...why?

While working on `builder-worker` I need to add more build-time
computing of dependent artifacts which led me to customize yet another
local `build.rs` file. After surveying the rest of the exceptional
components, it became clear that some consistency would help new
future components. Using the compile time inling trick, we get reuse
through composition when a local component needs specialization. Which
is needed in the immediate future work.

## [builder-worker] Refactor runner in preparation for exporters.

This change is primarily structural refactorings with the essential
behavior preserved, aside from an explicit crash when the config
`auto_token` is empty or not present.

Among some of the changes:

* Compile in the exact Studio package dependency when building under
Habitat, otherwise fall back to look for the latest installed
`core/hab-studio`. The command resolving is in a `lazy_static!` and so
is invariant once the program starts, no matter what other Studio
packages may be installed in the future.
* Reference Habitat-known environment variables from the relevant
constants.
* Normalize some of the logging output in the `runner` module.
* Extract Studio-spawning code into a `runner::studio` module. Future
exporter work will follow this pattern.
* Set `log_path` to `{{pkg.svc_var_path}}` when running in a package

References #3342


thesentinels:

Thank you for contributing!

fnichol:
Ugh, this'll hit *every* Rust component in CI. Wee!

tenor-262484108

components/libbuild.rs
```diff
@@ -0,0 +1,97 @@
+#[allow(dead_code)]
```
reset: Way nicer than copy pasting build scripts everywhere :)

components/builder-worker/src/runner/mod.rs
```diff
@@ -46,30 +40,28 @@ use zmq;
...
 use self::log_pipe::LogPipe;
 use self::postprocessor::post_process;
+use self::studio::Studio;
 use self::workspace::Workspace;
 use config::Config;
 use error::{Error, Result};
 use retry::retry;
 use vcs;
 
+// TODO fn: copied from `components/common/src/ui.rs`. As this component doesn't currently depend
```
reset: Yeah this probably does belong in core and not common

components/builder-worker/habitat/config/config.toml
```diff
@@ -1,6 +1,7 @@
 auth_token = "{{cfg.auth_token}}"
 auto_publish = {{cfg.auto_publish}}
 data_path = "{{pkg.svc_data_path}}"
+log_path = "{{pkg.svc_var_path}}"
```
reset: Good catch!

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/281975341).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/281975341).

PR #3436, opened 2017-10-01
Store Session in memory instead of database

reset:
This change will keep all sessions in memory stored on the heap
instead of in a database row. Sessions should be considered transient
data which exists until the point of expiration or until the server
is restarted, whichever comes first.

We were previously storing sessions in memory in Redis so there was
no reason to store them in memory in the application server. This is
no longer the case with PostgreSQL being the backend for sessionsrv.

tenor-106273665


thesentinels:

Thank you for contributing!

components/builder-sessionsrv/src/data_store.rs
```diff
@@ -59,13 +57,27 @@ impl DataStore {
...
+            "SELECT * FROM get_account_by_name_v1($1)",
+            &[&msg.get_name()],
+        )?;
+        if rows.len() != 0 {
+            let row = rows.get(0);
+            Ok(self.row_to_account(row))
+        } else {
+            let rows = conn.query(
+                "SELECT * FROM select_or_insert_account_v1($1, $2)",
+                &[&msg.get_name(), &msg.get_email()],
```
christophermaier: Would this call by itself get you what you want?

raskchanky:
The first thing that pops into my head is: what happens when we have multiple sessionsrvs?  That's the classic reason for putting sessions into a datastore vs in memory, so you avoid the problem of multiple requests from the same person getting routed to different servers, and losing their session.

My gut says that because their token will remain the same between requests, and that's the routing key for `SessionGet` messages, we've sidestepped that problem here, and even with multiple sessionsrvs, people won't get accidentally routed to the wrong server and lose their session.

Is that correct?

reset:
@raskchanky that's exactly correct! We use deterministic routing to determine which servers host which shards. We won't have every shard in every database server pretty soon so you wouldn't be able to just ask any SessionSrv for your session anyway

raskchanky:
@reset Ah, right, good point.  In that case, merge when Travis is green.

components/builder-sessionsrv/src/data_store.rs
reset: Ummm, why yes! Hahaha. I actually wrote a TODO above this reminding myself to write this function that already exists that I'm already calling :). Thanks for the catch man

raskchanky:
tenor-32590983

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/282042370).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/282042370).

PR #3442, opened 2017-10-02
Process GitHub Push Hook notifications

reset:
This commit changes how we make authenticated requests to the GitHub API
on behalf of the user and adds the ability to process GitHub Webhook
notifications for Ping and Push.

In order to support Ping and Push notifications for Private repositories
we needed to switch to integrating with GitHub using a full GitHub App
instead of just an OAuth App. This PR changes the github-api-client to
only make requests which a GitHub App could make and signs the commits
appropriately.

The authorization process has changed to a per repo basis and no longer
requires a user to setup a personal access token.

tenor-169787551

### DIFF
--- a/components/builder-admin/src/server/mod.rs
+++ b/components/builder-admin/src/server/mod.rs
@@ -35,7 +35,7 @@ impl HttpGateway for AdminSrv {
 
     fn add_middleware(config: Arc<Self::Config>, chain: &mut iron::Chain) {
         chain.link(persistent::Read::<GitHubCli>::both(
-            GitHubClient::new(&*config),
+            GitHubClient::new(config.github.clone()),
         ));
     }
 
@@ -50,7 +50,7 @@ impl HttpGateway for AdminSrv {
     }
 
     fn router(config: Arc<Self::Config>) -> Router {
-        let admin = Authenticated::new(&*config).require(privilege::ADMIN);
+        let admin = Authenticated::new(config.github.clone()).require(privilege::ADMIN);
         router!(
             status: get "/status" => status,
             search: post "/search" => XHandler::new(search).before(admin.clone()),
--- a/components/builder-api/src/github.rs
+++ b/components/builder-api/src/github.rs
@@ -1,0 +1,222 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::str::FromStr;
+
+use bodyparser;
+use bldr_core::build_config::{BLDR_CFG, BuildCfg};
+use github_api_client::GitHubClient;
+use hab_core::package::Plan;
+use http_gateway::http::controller::*;
+use iron::status;
+use persistent;
+use protocol::scheduler::{Group, GroupCreate};
+
+use error::Error;
+use headers::*;
+use types::*;
+
+pub enum GitHubEvent {
+    Push,
+    Ping,
+}
+
+impl FromStr for GitHubEvent {
+    type Err = Error;
+
+    fn from_str(event: &str) -> Result<Self, Self::Err> {
+        match event {
+            "ping" => Ok(GitHubEvent::Ping),
+            "push" => Ok(GitHubEvent::Push),
+            _ => Err(Error::UnknownGitHubEvent(event.to_string())),
+        }
+    }
+}
+
+enum HandleResult<T> {
+    Ok(T),
+    Err(Response),
+}
+
+pub fn handle_event(req: &mut Request) -> IronResult<Response> {
+    let event = match req.headers.get::<XGitHubEvent>() {
+        Some(&XGitHubEvent(ref event)) => {
+            match GitHubEvent::from_str(event) {
+                Ok(event) => event,
+                Err(err) => return Ok(Response::with((status::BadRequest, err.to_string()))),
+            }
+        }
+        _ => return Ok(Response::with(status::BadRequest)),
+    };
+    match event {
+        GitHubEvent::Ping => Ok(Response::with(status::Ok)),
+        GitHubEvent::Push => handle_push(req),
+    }
+}
+
+fn handle_push(req: &mut Request) -> IronResult<Response> {
+    // JW TODO: THIS NEEDS TO BE AUTHENTICATED
+    // use sender.login to get user ID to get account & see if they have access to origin
+    let hook = match req.get::<bodyparser::Struct<GitHubWebhookPush>>() {
+        Ok(Some(hook)) => hook,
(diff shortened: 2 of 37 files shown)


thesentinels:

Thank you for contributing!

components/builder-api/src/github.rs
```diff
@@ -0,0 +1,222 @@
...
+        _ => return Ok(Response::with(status::BadRequest)),
+    };
+    match event {
+        GitHubEvent::Ping => Ok(Response::with(status::Ok)),
+        GitHubEvent::Push => handle_push(req),
+    }
+}
+
+fn handle_push(req: &mut Request) -> IronResult<Response> {
+    // JW TODO: THIS NEEDS TO BE AUTHENTICATED
```
reset: I'm going to handle decrypting the private key before we push this to live

components/builder-api/src/github.rs
```diff
@@ -0,0 +1,222 @@
...
+        plans.retain(|plan| match cfg.get(&plan.name) {
+            Some(project) => hook.changed().iter().any(|f| project.triggered_by(f)),
+            None => false,
+        })
+    }
+    build_plans(req, plans)
+}
+
+fn build_plans(req: &mut Request, plans: Vec<Plan>) -> IronResult<Response> {
+    // JW TODO: Validate that this repository is where these plans belong. You could theoretically
```
reset: I'll also handle this before we push to live

components/builder-protocol/src/originsrv.rs
```diff
@@ -980,6 +980,38 @@ impl Serialize for OriginProject {
     }
 }
 
+// impl Routable for ProjectStateSet {
```
christophermaier: Did you mean to add this?

components/github-api-client/src/types.rs
```diff
@@ -0,0 +1,392 @@
...
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(default)]
+pub struct GitHubAuthor {
+    /// Public name of commit author
+    pub name: String,
+    /// Public email of commit author
+    pub email: String,
+    /// Display name of commit author
+    pub username: String,
+}
```
christophermaier: Are the values of these just `""` if the user doesn't have them set? Should these be `Option<String>` (at least, `email` and `name`)? Does it matter much for our usecase?

(Ditto for the other types in this module that are `String` instead of `Option<String>`.)

components/github-api-client/src/types.rs
```diff
@@ -0,0 +1,392 @@
...
+pub struct SearchItem {
+    pub name: String,
+    pub path: String,
+    pub sha: String,
+    pub url: String,
+    pub git_url: String,
+    pub html_url: String,
+    pub repository: Repository,
+    pub score: f64,
+}
```
christophermaier: I'm amazed there's not a library for a Github client in Rust yet!

components/builder-admin/src/server/mod.rs
```diff
@@ -35,7 +35,7 @@ impl HttpGateway for AdminSrv {
 
     fn add_middleware(config: Arc<Self::Config>, chain: &mut iron::Chain) {
         chain.link(persistent::Read::<GitHubCli>::both(
-            GitHubClient::new(&*config),
+            GitHubClient::new((*config).github.clone()),
```
christophermaier: Rust question time: What's the significance of the parentheses around `*config`?

components/builder-protocol/src/originsrv.rs
reset: No, it was left over from a really old branch. Thanks!

components/github-api-client/src/types.rs
reset: When we use `Option<String>` it's when the API can send `null` for a field. I should probably remove `serde(default)` from these structs which is how you'll end up with an empty string instead of a runtime error

components/builder-admin/src/server/mod.rs
reset: Actually now that I go to type this out it might be overkill. I was dereferencing the `Arc` to get at the inner value but `.github` will do the same thing

components/github-api-client/src/types.rs
reset: There are a few but nothing with enough coverage. The good thing is that we are only going to be using a portion of the GitHub API so we can create a well tested sub set and not need to rely on an upstream dependency

christophermaier: Admittedly, that _is_ pretty nice. 👍

components/builder-admin/src/server/mod.rs
christophermaier: Cool.

christophermaier:
tenor-58292932

Looks good; feel free to merge whenever you like.

reset:
@thesentinels approve

thesentinels:

chefsalim:
tenor-146797911

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/282354706).

components/builder-protocol/protocols/originsrv.proto
```diff
@@ -341,7 +341,7 @@ message OriginProject {
   optional uint64 owner_id = 7;
   optional string vcs_type = 8;
   optional string vcs_data = 9;
-  optional uint64 vcs_installation_id = 12;
+  optional uint32 vcs_installation_id = 12;
```
mwrock: I think a 32 bit is plenty but curious why the change here since it seems our convention is 64.

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/282354706).

components/builder-protocol/protocols/originsrv.proto
reset: Our convention is 64 but their API is sending 32

components/github-api-client/src/client.rs
```diff
@@ -27,42 +24,66 @@ use hyper::net::HttpsConnector;
...
-            web_url: config.github_web_url().to_string(),
-            client_id: config.github_client_id().to_string(),
-            client_secret: config.github_client_secret().to_string(),
-            app_id: config.github_app_id(),
-            app_private_key_path: config.github_app_private_key_path(),
+            url: config.url,
+            web_url: config.web_url,
+            client_id: config.client_id,
+            client_secret: config.client_secret,
+            app_private_key: config.app_private_key,
```
mwrock: did we intend to hard code the app id?

components/github-api-client/src/client.rs
```diff
@@ -258,286 +259,94 @@ impl GitHubClient {
...
+where
+    T: ToString,
+{
+    let mut payload = jwt::Payload::new();
+    let header = jwt::Header::new(jwt::Algorithm::RS256);
+    let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+    let expiration = now + Duration::from_secs(10 * 60);
+    payload.insert("iat".to_string(), now.as_secs().to_string());
+    payload.insert("exp".to_string(), expiration.as_secs().to_string());
+    payload.insert("iss".to_string(), 5565.to_string());
```
mwrock: did we intend to hard code the app id?

components/builder-protocol/protocols/originsrv.proto
mwrock: Yeah I think that was my fault. Thats the type I used in our frank fork. Should be fine but in retrospect I'd have used u64.

components/github-api-client/src/client.rs
reset: Absolutely not, I've got this fixed in another branch ;)

PR #3443, opened 2017-10-02
[builder-worker] Add initial support for Docker exporting.

fnichol:
This change adds support for exporting a built package into a Docker
image. The behavior is very much MVP as it lacks support for pushing to
Docker Hub still lacking (due to credentials), and cleaning up the
Docker images layers.

Here are some things to note:

* The Docker exporting shares the same log stream as the Studio build so
these will show up one after the other with a very simple section
starting/ending line.
* The exporting logic is handled with `hab-pkg-export-docker`.
* The worker will spawn its own isolated Docker Engine instance only to
be used by the current job and will later be destroyed after being
terminated. It should have no inbound network access and only runs for
the duration of the export process. The worker handles process
management so there is no additonal provisioning or Habitat services
needed on the worker.
* Currently, the behavior is to create a Docker image if the package is
found to be "runnable" without checking any `Job` data.
* No Docker Hub credentials are checked for and as a result, the image
is not currently pushed.

This exporting behavior is behind a small feature flag which is off by
default. Adding the feature token in the workers config will enable like
so:

```
features_enabled = "DOCKER"
```

Toggling this on at runtime under Habitat supervision can be
accomplished with `hab apply`:

```
echo 'features_enabled = "DOCKER"' \
  | hab apply builder-worker.default $(date +%s)
```

And turned off by removing the `DOCKER` token in the string:

```
echo 'features_enabled = ""' \
  | hab apply builder-worker.default $(date +%s)
```

---

Also…

## [builder-worker] Add behavioral feature flagging using config value

This change adds feature flags to the worker which can be enabled by
setting the `features_enabled` string field in the config as a comma
separated list of uppercase string tokens such as `"FOO,BAR"`.

References #3342

tenor-5976805

### DIFF
--- a/components/builder-worker/src/runner/docker.rs
+++ b/components/builder-worker/src/runner/docker.rs
@@ -1,0 +1,199 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fs::{self, File};
+use std::io::Write;
+use std::path::PathBuf;
+use std::process::{Child, Command, ExitStatus, Stdio};
+
+use hab_core::env;
+use hab_core::fs as hfs;
+use hab_core::os::process::{self, Signal};
+
+use error::Result;
+use runner::log_pipe::LogPipe;
+use runner::{NONINTERACTIVE_ENVVAR, RUNNER_DEBUG_ENVVAR};
+use runner::workspace::Workspace;
+
+lazy_static! {
+    /// Absolute path to the Docker exporter program
+    static ref DOCKER_EXPORTER_PROGRAM: PathBuf = hfs::resolve_cmd_in_pkg(
+        "hab-pkg-export-docker",
+        include_str!(concat!(env!("OUT_DIR"), "/DOCKER_EXPORTER_PKG_IDENT")),
+    );
+
+    /// Absolute path to the Dockerd program
+    static ref DOCKERD_PROGRAM: PathBuf = hfs::resolve_cmd_in_pkg(
+        "dockerd",
+        include_str!(concat!(env!("OUT_DIR"), "/DOCKER_PKG_IDENT")),
+    );
+}
+
+const DOCKER_HOST_ENVVAR: &'static str = "DOCKER_HOST";
+
+pub struct DockerExporter<'a> {
+    workspace: &'a Workspace,
+    bldr_url: &'a str,
+}
+
+impl<'a> DockerExporter<'a> {
+    /// Creates a new Docker exporter for a given `Workspace` and Builder URL.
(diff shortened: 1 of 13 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/src/runner/docker.rs
```diff
@@ -0,0 +1,197 @@
...
+    ///
+    /// # Errors
+    ///
+    /// * If the child process can't be spawned
+    /// * If the calling thread can't wait on the child process
+    /// * If the `LogPipe` fails to pipe output
+    pub fn export(&self, log_pipe: &mut LogPipe) -> Result<ExitStatus> {
+        let dockerd = self.spawn_dockerd()?;
+        let exit_status = self.run_export(log_pipe);
+        self.teardown_dockerd(dockerd)?;
```
reset: I would warn that this failed but I wouldn't crash the worker because of it. Ideally we pick this up in our log dashboard and the investigate why our disk is filling up but jobs are still completing :)

fnichol: Whoops, that was what I intended before I shunted it all to a method. Will update.

components/builder-worker/src/runner/docker.rs
```diff
@@ -0,0 +1,197 @@
...
+        let mut cmd = Command::new(&*DOCKER_EXPORTER_PROGRAM);
+        cmd.current_dir(self.workspace.root());
+        cmd.arg("--base-pkgs-url");
+        cmd.arg(&self.bldr_url);
+        cmd.arg("--url");
+        cmd.arg(&self.bldr_url);
+        cmd.arg("--tag-latest");
+        cmd.arg("--tag-version");
+        cmd.arg("--tag-version-release");
+        // TODO fn: enable push once passwords are gettable
```
reset: We should pipe these creds into the job spec sent to the worker. I guess we haven't done that yet?

fnichol: That's what I need to determine now: what if any of this info is being sent down the job spec pipe. Once I can consume, this gets wired right quick.

components/builder-worker/src/runner/docker.rs
```diff
@@ -0,0 +1,197 @@
...
+        cmd.arg(self.dockerd_sock());
+        cmd.arg("--pidfile");
+        cmd.arg(root.join("var/run/docker.pid"));
+        cmd.arg("--data-root");
+        cmd.arg(root.join("var/lib/docker"));
+        cmd.arg("--exec-root");
+        cmd.arg(root.join("var/run/docker"));
+        cmd.arg("--config-file");
+        cmd.arg(daemon_json);
+        // TODO fn: Hard-coding this feels wrong. I'd like the {{pkg.svc_run_group}} for this
```
reset: I would think you'd want to spawn it as the same user/group as the worker itself or perhaps it's a configuration on the worker for what user you'd like to run Docker as?

fnichol: It sets the group ownership of the `docker.sock` which determines access. I'm hopeful if I plumb though the runtime variable this should work--even though that group is most likely `root`.

reset:
@fnichol I'd fix the one issue I commented above with stopping execution early. Other than that this is a :fire: PR

fnichol:
tenor-93705072

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/282128518).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/282128518).

PR #3452, opened 2017-10-02
Composite Plans!

### LINKED ISSUE #2902: Composite Plan/Package/Service
## User Story

As a plan author, I need a way to represent a service which is made up of one or more processes, so I can interact with a collection of services in the same Supervisor as one

## Background

We currently have one kind of plan which is represented by a `plan.sh` or a `plan.ps1`. A plan outputs a package which, when installed, can be started by a Supervisor to start a service. It can be said that a plan/package is a 1:1 mapping to a service.

Some services require another to run and we need a way to represent this "sidecar" pattern. An example of this is an API server and an API proxy.

## Changes to Plan-Build

We need a way to represent that the plan is a composite plan. We could do this in a few different ways such as:

* Add a new key/value setting which plan-build reads such as `pkg_composite = true`
* Use a different filename than `plan.sh` such as `composite-plan.sh`

We also do not need to allow any additional artifacts aside from the `plan.sh` in a composite plan. There's no need for the additional artifacts - read on for details.

### Promise Mapping

Once we've indicated that we're building a composite plan we also need a way to configure some mappings for any promises that the package exposes (aka, `exports`/`binds`). We also need a way to note that an export of a service of the composite plan satisfies a bind of another. For example, `builder-api` needs to satisfy the bind for `builder-api-proxy`.

We need to verify that there are no collisions in the key space of exports or binds. If one package in the composite exports `hello` and a second also does, we will need to fail the build (and should fail before any work is done)

We will want to record this in a new METADATA file indicating:

* That this is a composite package
* What the promise mappings are
* Which promise mappings self satisfy service members

## Changes to Hab CLI

All operations with the CLI should be atomic across all service members of the composite package/service. For example: installing a composite package should install all packages it points to. Another way to think about a composite package is as a "pointer package" (sometimes referred to in other systems as a virtual package).

We may want to consider representing a composite service in the status output of running services as something like:

```
$ hab svc status
core/builder-api/4567/20170727171959, state:up, time:PT1.870780743S, pid:9199, group:builder-api.acceptance, style:persistent
  - core/builder-api-srv/4567/20170727171959
  - core/builder-api-proxy/4355/20170706225642
```

All `hab pkg` and `hab svc` commands should be refreshed to support these changes

### Config Apply

Since we want a composite service to be addressable as a service group itself, we will also need  to make some alterations to `hab config apply`.

Currently we run `hab config apply <SERVICE_GROUP> <VERSION_NUMER> [file]` which takes a toml file representing a single service's configuration. We should allow the toml file to be split into multiple sections, one for each service in the composite service:

```toml
[builder-api-srv]
  [http]
  listen = "0.0.0.0"
  port = 9636
[builder-api-proxy]
  worker_rlimit_nofile = 8192
  worker_processes = "auto"

  [events]
  worker_connections = 8000
```

## Changes to Butterfly / Supervisor

With all of the above changes we may also need to add a new rumor type, `CompositeServiceRumor`, which contains the metadata for promise mappings.

The config apply changes described above may also result in the need for us to generate a new type of rumor for `CompositeServiceConfigRumor`. This would contain the combined configuration. If we decide to go this route, we will need to make sure that we reject config changes to the individual service groups which make up the composite service group - contains a combined 

Lastly, we should not generate a service directory at `/hab/svc/{name}` for a composite service.

christophermaier:
Composite Plans!

Read the commit messages for extensive notes and background, as well as discussion of current limitations and possible future improvements.

tenor-143623621

Closes #2902


thesentinels:

Thank you for contributing!

fnichol:
Awesome! Given the size and scope, this could take me a day or two to go through. This will make a lot of people happy!

christophermaier:
Ah, I forgot to mention that the proposed idea of subdividing composites into "sets" (e.g., create an entire composite for Habitat Builder, and have a "front-end" set of services and a "back-end" set of services, which could be addressed separately, is not present in this code. I demoed creating a composite package with that metadata, but given the amount of complexity already present in just dealing with composites themselves, adding sets right now didn't feel like the right choice.

jamessewell:
Amazing! 

Can a composite be made up of (some|all) composites?

christophermaier:
@jamessewell No, just services at the moment.

As I discuss in #3593, though, this is really just a first offering for this feature; we're interested in evolving this with community input.

jamessewell:
I'll give it a go today!

reset:
This is a big one. I didn't see anything wrong in my initial review, the work is solid. I'm going to merge this for now since I'm getting on a plane which will leave me with plenty of time to begin work on some changes to the Supervisor. I'd rather build all of it off of this and fix anything that's out of place as I go. Merging and taking responsibility if it breaks anything for the next 8 hours while I'm in the air ;).

reset:
tenor-261990503

johnjelinek:
LGTM 👏 👏 👏

PR #3481, opened 2017-10-03
Don't cache response for integration names

chefsalim:
We need to set the no-cache for one of the depot API calls.

tenor-262581493


thesentinels:

Thank you for contributing!

components/builder-http-gateway/src/http/helpers.rs
```diff
@@ -58,6 +58,12 @@ pub fn get_authenticated_session(req: &mut Request) -> Option<Session> {
     }
 }
 
+pub fn dont_cache_response(response: &mut Response) {
```
fnichol: Good idea!

PR #3492, opened 2017-10-04
Fix false not found error when streaming log output

reset:
A check was added to retrieve a package from a job to see if it had
the appropriate visibility before deciding to send job logs to the
requestor. The package is not created until the very end of the job
so we aren't able to check what the visibility is.

This change is still less than ideal, but it fixes the streamin logs.
We check to see what the default visibility setting is on the origin
before deciding to check if the user is a member of that origin if
the visibility is private.

We need to change projects to have a visibility setting that they
inherit from their origin and then make modifications to this change
to check the project instead of the origin.

tenor-142749779


thesentinels:

Thank you for contributing!

PR #3514, opened 2017-10-04
correct the private key path in the dev doc

mwrock:

### DIFF
--- a/BUILDER_DEV.md
+++ b/BUILDER_DEV.md
@@ -9,7 +9,7 @@ This document outlines the steps to start and run a Builder environment for deve
 1. The sample commands below use the 'httpie' tool. Install it if not present on your system (https://github.com/jkbrzt/httpie).
 1. A recent version of the habitat cli should also be installed in your dev environment (https://www.habitat.sh/docs/get-habitat/)
 1. Ensure that `sudo -E` functions correctly. Some OSes don't respect the -E flag. If this is the case, remove the `secure_path` section of your sudoers file.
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

BUILDER_DEV.md
```diff
@@ -9,7 +9,7 @@ This document outlines the steps to start and run a Builder environment for deve
 1. The sample commands below use the 'httpie' tool. Install it if not present on your system (https://github.com/jkbrzt/httpie).
 1. A recent version of the habitat cli should also be installed in your dev environment (https://www.habitat.sh/docs/get-habitat/)
 1. Ensure that `sudo -E` functions correctly. Some OSes don't respect the -E flag. If this is the case, remove the `secure_path` section of your sudoers file.
-1. Copy `habitat-builder-dev.2017-10-02.private-key.pem` from 1Password into `.secrets/builder-github-app.pem`
+1. Copy `habitat-builder-dev.2017-10-02.private-key.pem` from 1Password into `/src/.secrets/builder-github-app.pem`
```
chefsalim: Alternatively I think we can just set the builder-api config file to point to this - I'm testing that out

chefsalim: Also there are other changes to the builder api config file below - we need to remove the other github client id and secret config.

mwrock: there are several services that need to point here. `/src/.secrets/builder-github-app.pem` is the github config's default.

mwrock: We actually still need the other GH config properties.

reset:
I wrote this with the idea that you'd be outside the container but if this is more clear, let's do it

Issue #3513, opened 2017-10-04
Users removed from an origin must also be removed from acount_origins in sessionsrv


mwrock:
User removal requires surgery in sessionsrv in addition to the work already done in originsrv. Otherwise that user can still see the origin.

PR #3518, opened 2017-10-04
removed members should also be removed from account_origins

mwrock:


thesentinels:

Thank you for contributing!

components/builder-depot/src/server.rs
```diff
@@ -473,17 +473,28 @@ pub fn origin_member_delete(req: &mut Request) -> IronResult<Response> {
...
         Err(err) => return Ok(render_net_error(&err)),
     }
-    request.set_account_name(account_name.to_string());
+    session_request.set_account_id(session.get_id());
+    origin_request.set_account_name(account_name.to_string());
+
+    if let Err(err) = route_message::<AccountOriginRemove, NetOk>(req, &session_request) {
+        error!("Error deleting origin_account from sessionsrv, {}", err);
+        return Ok(render_net_error(&err));
+    }
```
raskchanky: I wonder if it'd be smart to only run the delete from the sessionsrv if the delete from the originsrv succeeded, or vice versa.  The point being, ensure that either both succeed or neither succeeds.  Right now, though it's unlikely, you could have the delete from sessionsrv succeed and the delete from originsrv fail, and now they're out of sync.

Thoughts?

mwrock: i thought about this. I think its more important to have the sessionsrv delete succeed and originsrv fail. The reason being that if for some reason a partial deletion occured and only sessionsrv succeeded. At least the removed user would not be able to see the origin anymore.  The weirdneww would be that the user would still show up as a member of the origin, but a "reremoval" could be attempted. The opposite could be worse: the member could still see the origin but legit origin members would not see that user as a member.

raskchanky: I think my point was more along the lines of only attempting the second delete if the first one succeeds, regardless of which one you do first.  We don't have a way of doing a distributed commit with this stuff, but at least that way, if the first one fails, you don't even try the second delete, and just abort completely.

mwrock: oh that should already be setup. If the first fails it does `return Ok(render_net_error(&err))`

raskchanky: Haha, no idea how I missed that.  👍

mwrock:
related to #3513

components/builder-depot/src/server.rs
reset: @mwrock this is a good starting place and better than having nothing, but what we want to do is use the `async_job` feature of the db crate that @adamhjk added. When you have a database transaction that crosses a shard partition you need to use the retry logic in the server of record to ensure the other bits are cleaned up

With that said, I think it's ok to merge this in as is for now but we should add a comment and circle around back

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/284462079).

components/builder-depot/src/server.rs
mwrock: ok thanks. I'll look for a usage example

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/284462079).

PR #3519, opened 2017-10-04
Set cookie domain to all hostname elements excluding the first

reset:
Previously we were setting cookies to always be the last two elements
of a domain so `bldr.habitat.sh` would be `.habitat.sh` and
`bldr.acceptance.habitat.sh` would also be `.habitat.sh`. This change
will leave the former while the latter becomes `.acceptance.habitat.sh`


thesentinels:

Thank you for contributing!

PR #3535, opened 2017-10-05
[plan-build] Update `_latest_installed_package()` to use `hab pkg path`.

fnichol:
This change deprecates the old implementation of the function which
determines the latest installed package on disk which is used to resolve
dependencies. The original implementation used a combination of `find`,
`sort` (with the `--version-sort` flag), and `head`. The updated
implementation simply uses `hab pkg path` as it provides the same basic
logic. However, the Rust implementation uses the same version string
sorting as the Supervisor and should be considered more correct. As the
GNU version of sort is no longer necessary, the explicit check in
`_find_system_commands()` can be dropped for `sort`.

tenor-191331462

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -542,7 +542,6 @@ _ensure_origin_key_present() {
 # command:
 #
 # * `$_hab_cmd` (hab cli for signing, hashing, and possibly installing)
-# * `$_sort_cmd` (GNU version from coreutils)
 # * `$_wget_cmd` (wget on system)
 # * `$_shasum_cmd` (either gsha256sum or sha256sum on system)
 # * `$_tar_cmd` (GNU version of tar)
@@ -558,17 +557,6 @@ _ensure_origin_key_present() {
 # If the commands are not found, `exit_with` is called and the program is
 # terminated.
 _find_system_commands() {
-  if $(sort --version 2>&1 | grep -q 'GNU coreutils'); then
-    _sort_cmd=$(command -v sort)
-  else
-    if $(/usr/bin/sort --version 2>&1 | grep -q 'GNU coreutils'); then
-      _sort_cmd=/usr/bin/sort
-    else
-      exit_with "We require GNU sort to find the latest package; aborting" 1
-    fi
-  fi
-  debug "Setting _sort_cmd=$_sort_cmd"
-
   if exists wget; then
     _wget_cmd=$(command -v wget)
     if [[ "${HAB_NONINTERACTIVE:-}" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

fnichol:
I'm 80-90% sure I got this right and it'll end up being more correct in the long run, but it's worth a second (or third) look to make sure I haven't forgotten any edge cases.

I stumbled across an issue when trying to build dependent packages from habitat-sh/core-plans#839 which rev'd a version of Go from `1.9` (note the lack of patch number) to `1.9.1`. When I ran the build for `core/nats-streaming-server`, I saw that the build program resolved the `core/go/1.9` dependency rather than the `core/go/1.9.1` dependency (both were installed on disk).

Looking at the implementation of our "find latest installed" function, it actually gave the wrong answer to what I think most people would expect. That is, it determined that `"1.9"` was newer than `"1.9.1"`.

Substituting in our `hab pkg path` implementation which uses the `core` library's version sorting logic, I got the answer I expected: `"1.9.1"` being the newest.

I'm guessing this isn't the only time where we haven't resolved to what we expected so I suspect this change fixes more bugs than it introduces. I'm only cautious as that function implementation has been exercised thousands and thousands of time and "mostly works right™".

Anyway, gut check?

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -607,7 +595,7 @@ _find_system_commands() {
   else
     exit_with "We require xz to compress artifacts; aborting" 1
   fi
-  debug "Setting _hab_cmd=$_hab_cmd"
+  debug "Setting _xz_cmd=$_xz_cmd"
```
christophermaier: Good copy-pasta catch!

christophermaier:
@thesentinels approve
tenor-188666007

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/283699156).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/283699156).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/283699156) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

christophermaier:
I think Travis may have gotten confused with all the merges that have happened this morning. Will try again in a bit.

fnichol:
Rebasing!

PR #3536, opened 2017-10-05
Fix session hijacking issue after changing to GitHub App authentication

reset:
Session hijacking came back after we switched authentication methods
from a GitHub OAuth App to a GitHub App. OAuth applications use
personal access tokens which can be used to identify an incoming user
and verify them with GitHub. These are static and cryptographically
secure so they _can_ be used as a session token to identify a user,
but this approach was just a stop gap.

GitHub Apps use the same base16 encoded string for their token, but
they are generated each time the user authenticates. They also appear
to be reused which is the source of our problem. If you log out and
log back in you may be assigned an auth token which another user
had in the past. Since we only time sessions out every day, each
login would result in a new session and you may have been assigned a
key which was already assigned this day.

In this commit instead of using an OAuth key directly to identify
our sessions we instead generate our own opaque token which contains
identifiable information plus the extern oauth token. We will never
issue identical tokens since the information contained is unique to
the user it was issued to.

tenor-200955590


thesentinels:

Thank you for contributing!

components/builder-protocol/protocols/net.proto
```diff
@@ -22,7 +23,7 @@ enum ErrCode {
   ENTITY_CONFLICT = 9;
   SOCK = 10;
   DATA_STORE = 11;
-  AUTH_SCOPE = 12;
+  BAD_TOKEN = 12;
```
reset: This is fine, we weren't using it

components/builder-protocol/protocols/sessionsrv.proto
```diff
@@ -94,27 +95,28 @@ message Session {
...
-  optional string token = 1;
-  optional uint64 owner_id = 2;
-  optional OAuthProvider provider = 3;
+  optional string oauth_token = 6;
 }
 
 message SessionCreate {
   optional string token = 1;
-  optional uint64 extern_id = 2;
+  optional uint32 extern_id = 2;
```
reset: This change is fine, we don't want the old authentication flow to work anyway at all. It was a security issue

PR #3537, opened 2017-10-05
Move VCS related code back to worker crate from bldr core

reset:
Moving the VCS module into the bldr core crate caused a libgit
dependency on each Builder server which isn't necessary. This change
decouples the VCS and Job modules and moves the VCS module back to
the worker crate


thesentinels:

Thank you for contributing!

PR #3568, opened 2017-10-06
[ci] Purge pre-existing origin keys before building.

fnichol:
This ensures that: a) the origin key is always fresh, b) there is only
one present, c) and import errors colliding with existing files should
be avoided.

This should allow us to publish build artifacts again more consistently.


thesentinels:

Thank you for contributing!

PR #3574, opened 2017-10-06
Key sessions on extern-id & provider instead of opaque token

reset:
* Add type for SessionKey - (u32, proto::OAuthProvider) - and key
  all sessions on this value instead of the opaque token. This
  will reduce the amount of sessions generated by avoiding a new
  session creation on every login/logout action
* Add debug logging for session get and session create to help
  diagnose acceptance environment issues

tenor-24160646

### DIFF
--- a/components/builder-sessionsrv/src/server/handlers.rs
+++ b/components/builder-sessionsrv/src/server/handlers.rs
@@ -14,11 +14,10 @@
 
 use std::env;
 
-use base64;
 use hab_net::app::prelude::*;
 use hab_net::privilege::{self, FeatureFlags};
 
-use protocol::{message, net};
+use protocol::net;
 use protocol::sessionsrv as proto;
 
 use super::{ServerState, Session};
@@ -106,7 +105,7 @@ pub fn session_create(
     state: &mut ServerState,
 ) -> SrvResult<()> {
     let mut msg = req.parse::<proto::SessionCreate>()?;
-    let mut session = Session::default();
+    debug!("session-create, {:?}", msg);
     let mut flags = FeatureFlags::default();
     if env::var_os("HAB_FUNC_TEST").is_some() {
         flags = FeatureFlags::all();
@@ -119,23 +118,11 @@ pub fn session_create(
     account_req.set_email(msg.take_email());
 
     match conn.route::<proto::AccountFindOrCreate, proto::Account>(&account_req) {
-        Ok(mut account) => {
-            let mut token = proto::SessionToken::new();
-            token.set_account_id(account.get_id());
-            token.set_extern_id(msg.get_extern_id());
-            token.set_provider(msg.get_provider());
-            token.set_token(msg.get_token().to_string().into_bytes());
-            let bytes = message::encode(&token)?;
-            let encoded_token = base64::encode(&bytes);
-
-            session.set_id(account.get_id());
-            session.set_email(account.take_email());
-            session.set_name(account.take_name());
-            session.set_token(encoded_token);
-            session.set_flags(flags.bits());
-            session.set_oauth_token(msg.take_token());
+        Ok(account) => {
+            let session = Session::build(msg, account, flags)?;
             {
-                state.sessions.write().unwrap().replace(session.clone());
+                debug!("issuing session, {:?}", session);
+                state.sessions.write().unwrap().insert(session.clone());
             }
             conn.route_reply(req, &*session)?;
         }
@@ -154,9 +141,10 @@ pub fn session_get(
     state: &mut ServerState,
 ) -> SrvResult<()> {
     let msg = req.parse::<proto::SessionGet>()?;
-    let encoded = message::encode(msg.get_token())?;
-    let base64_token = base64::encode(&encoded);
-    match state.sessions.read().unwrap().get(base64_token.as_str()) {
+    match state.sessions.read().unwrap().get(&(
+        msg.get_token().get_extern_id(),
+        msg.get_token().get_provider(),
+    )) {
         Some(session) => {
             if session.expired() {
                 let err = NetError::new(ErrCode::SESSION_EXPIRED, "ss:session-get:1");
--- a/components/builder-sessionsrv/src/server/mod.rs
+++ b/components/builder-sessionsrv/src/server/mod.rs
@@ -21,9 +21,11 @@ use std::ops::{Deref, DerefMut};
 use std::sync::RwLock;
 use std::time::{Duration, Instant};
 
+use base64;
 use github_api_client::GitHubClient;
 use hab_net::app::prelude::*;
-use protocol::sessionsrv as proto;
+use hab_net::privilege::FeatureFlags;
+use protocol::{message, sessionsrv as proto};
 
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

components/builder-sessionsrv/src/server/handlers.rs
```diff
@@ -106,7 +105,18 @@ pub fn session_create(
...
+        if let Some(session) = state.sessions.read().unwrap().get(&(
+            msg.get_extern_id(),
+            msg.get_provider(),
+        ))
+        {
+            debug!("found existing session, {:?}", session);
+            conn.route_reply(req, &**session)?;
+            return Ok(());
+        }
+    }
```
christophermaier: Why is this chunk of code in its own block? Is it for interacting with the `state`?

components/builder-sessionsrv/src/server/mod.rs
```diff
@@ -64,30 +66,51 @@ lazy_static! {
...
 
-#[derive(Clone)]
+pub type SessionKey = (u32, proto::OAuthProvider);
+
+#[derive(Clone, Debug)]
 pub struct Session {
     pub created_at: Instant,
+    pub key: SessionKey,
     inner: proto::Session,
+    token: proto::SessionToken,
```
christophermaier: We're going to need to get some documentation strings in here (i.e., the codebase, not just this particular site) soon so we can keep straight exactly what each component is.

adamhjk:
Should u33 be u64?

components/builder-sessionsrv/src/server/handlers.rs
chefsalim: yeah probably to not hold on to the read lock on sessions..

components/builder-sessionsrv/src/server/mod.rs
chefsalim: yeah.. personally I'd prefer cleaning up naming where needed vs. more documentation...

components/builder-sessionsrv/src/server/handlers.rs
```diff
@@ -119,23 +129,11 @@ pub fn session_create(
...
-            session.set_email(account.take_email());
-            session.set_name(account.take_name());
-            session.set_token(encoded_token);
-            session.set_flags(flags.bits());
-            session.set_oauth_token(msg.take_token());
+        Ok(account) => {
+            let session = Session::build(msg, account, flags)?;
             {
-                state.sessions.write().unwrap().replace(session.clone());
+                debug!("issuing session, {:?}", session);
```
chefsalim: should this debug go inside of Session::build ?

components/builder-sessionsrv/src/server/mod.rs
christophermaier: Yeah, either way is fine; I just want to reduce the "bus number" for this code soon 😄

components/builder-sessionsrv/src/server/mod.rs
```diff
@@ -64,30 +66,51 @@ lazy_static! {
...
+        flags: FeatureFlags,
+    ) -> SrvResult<Self> {
+        let mut session = proto::Session::new();
+        let mut token = proto::SessionToken::new();
+        token.set_account_id(account.get_id());
+        token.set_extern_id(msg.get_extern_id());
+        token.set_provider(msg.get_provider());
+        token.set_token(msg.get_token().to_string().into_bytes());
+
+        session.set_id(account.get_id());
```
chefsalim: rename session "id" field to "account_id"?

components/builder-sessionsrv/src/server/mod.rs
```diff
@@ -112,13 +135,13 @@ impl Hash for Session {
...
     {
-        self.inner.get_token().hash(state);
+        self.key.hash(state);
     }
 }
 
 impl PartialEq for Session {
     fn eq(&self, other: &Session) -> bool {
-        self.inner.get_token() == other.inner.get_token()
+        self.key == other.key
```
chefsalim: if I was paranoid I might add some debugging here as well :)

components/builder-sessionsrv/src/server/mod.rs
raskchanky: Yeah, agreed

raskchanky: Although that will break a ton of other code that all references `session.get_id()`.  😞

raskchanky: On second thought, maybe it's worth leaving it so we're not changing too many things at once.

components/builder-sessionsrv/src/server/handlers.rs
reset: Yep, it's because we're entering a critical section of memory and want to release it ASAP

components/builder-sessionsrv/src/server/handlers.rs
reset: I think it's best to debug in control flow and not units, that way you tell the story of what's happening in the order in which it's happening

reset:
@adamhjk the GitHub API and from what I can tell OAuth in general uses 32bit integers because 64bit integers can't be represented unless they're converted first to a string and then parsed in JSON

components/builder-sessionsrv/src/server/mod.rs
reset: Yeah we can take a pass at all of the message names/IDs if we'd like just before launch. It's all typed so it should be an easy enough refactor. We have dead messages, fields that no longer are needed, and at this point we don't need to be backwards compatible because we're going to take an outage when we deploy to prod anyway

mwrock:
Oh interesting. I thought they were u32 becauls of https://github.com/habitat-sh/frank_jwt/blob/habitat/src/lib.rs#L141

components/builder-sessionsrv/src/server/mod.rs
reset: Yeah I want to clean up the nomenclature here just as soon as we stabilize sessions but the gist is this:

* Session - contains all information you need to provide base operability upon login. It should include things like the origin's you're part of, your name, your id, and your token
* SessionToken - opaque piece of identifiable data which is transient and lasts for 1 day. It's packed with information for our server to identify you upon a login. You should never give somebody your session token because they can _become you_

reset:
@mwrock yes, they're u32 ;)

reset:
I thought about the changes I had made this morning and decided _not_ to include the operation of checking for a session before generating a new one. This is an over optimization to try and prevent a session hijack which should be fixed in another way. It also introduces a caching issue where the user has a different OAuth token and their session contains another

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/284333887).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/284333887).

PR #3584, opened 2017-10-06
Generate a txn-id when generating new synchronous requests

reset:
This change activates the usage of the `id` field in a network
message's TXN message part. We generate a non-zero transaction id
and include it as part of the message. When we receive a response
to our message we ensure that the txn-id is what we sent.

We should never receive messages out of order or a response for
somebody else's request, but if we do, this will prevent any data
corruption.

tenor-42229730


thesentinels:

Thank you for contributing!

chefsalim:
tenor-210074606

PR #3614, opened 2017-10-07
Remove expired Sessions on SessionGet and strict match on token

reset:
* When we get an expired Session in SessionGet we now remove it
  before returning the handler
* Additionally check if the user's given token matches in entirity
  and return session expired if they do not match

tenor-4172597


thesentinels:

Thank you for contributing!

PR #3632, opened 2017-10-08
Handle missing error messages

cnunciato:
Slight logic error here — what we should be doing is acting on the presence of a payload, rather than looking for an error message; in development, 404s are resolved with a message of `"Not Found"`, but in prod, 404 messages are `""`, which is falsy in JavaScript.


thesentinels:

Thank you for contributing!

reset:
tenor-169742788

PR #3636, opened 2017-10-09
Support multiple sessions for the same user

reset:
We previously would enforce a single session for a logged in user.
This is a bad experience for a web application and was a reflex
of mine from working on game authentication systems where you can
only have one active session.

This commit validates sessions appropriately and protects against
forgery while also allowing a user to have multiple active sessions.

tenor-88996589


thesentinels:

Thank you for contributing!

PR #3638, opened 2017-10-09
adds in blog post about node scaffolding

nellshamrell:

This adds in a blog post (which will also be copied to the official Chef blog) that highlights the node scaffolding - where it is now and what is coming soon - along with how to use it with Builder.


reset:
tenor-185847963

PR #3639, opened 2017-10-09
Check for origin access before allowing updates to an origin

reset:
tenor-66397537

### DIFF
--- a/components/builder-depot/src/server.rs
+++ b/components/builder-depot/src/server.rs
@@ -71,16 +71,18 @@ struct OriginUpdateReq {
 const ONE_YEAR_IN_SECS: usize = 31536000;
 
 pub fn origin_update(req: &mut Request) -> IronResult<Response> {
+    let session = req.extensions.get::<Authenticated>().unwrap().clone();
     let mut request = OriginUpdate::new();
-
-    let origin = {
+    {
         let params = req.extensions.get::<Router>().unwrap();
-
         match params.find("name") {
-            Some(o) => o.to_string(),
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-depot/src/handlers/integrations.rs
```diff
@@ -79,6 +79,10 @@ pub fn fetch_origin_integration_names(req: &mut Request) -> IronResult<Response>
         Ok(p) => p,
         Err(st) => return Ok(Response::with(st)),
     };
+    let session = req.extensions.get::<Authenticated>().unwrap().clone();
```
chefsalim: we already check origin access in the validate_params helper

components/builder-depot/src/handlers/integrations.rs
```diff
@@ -106,6 +110,10 @@ pub fn create_origin_integration(req: &mut Request) -> IronResult<Response> {
...
```
chefsalim: same as above

components/builder-depot/src/handlers/integrations.rs
```diff
@@ -155,6 +163,10 @@ pub fn delete_origin_integration(req: &mut Request) -> IronResult<Response> {
...
```
chefsalim: same as above

components/builder-depot/src/handlers/integrations.rs
reset: Ah of course! Updated the PR :)

PR #3642, opened 2017-10-09
Expose builder-web feature flag configuration

reset:
Add a new toml map `web_feature_flags` which can be used to set
feature flags for the UI. In the future, we should hook the feature
flags up to the contents of the session's feature flags. This is
good enough for now.


thesentinels:

Thank you for contributing!

PR #3643, opened 2017-10-09
Accumulate all repositories via pagination

### LINKED ISSUE #3628: Paginate repositories when retrieving the contents of a file
https://github.com/habitat-sh/habitat/blob/b85af98e238cfdc269bfec72b5f3156aec5d3806/components/builder-api/src/github.rs#L139-L146

The handler for retrieving contents from a file will sometimes 404 because the matching repository could not be found. This happens when you have a number of repositories in your organization/user. We need to handle pagination of API results: https://developer.github.com/v3/apps/installations/#list-repositories-accessible-to-the-user-for-an-installation

christophermaier:
Previously, we were only hitting the first page of repository results
from Github, which meant (in effect), that Builder would only be able
to build the first 30 repositories in your account, sorted
alphabetically :)

Now, we traverse the URLs in the Link header of responses, allowing us
to accumulate *all* the repositories in an account.

Fixes #3628


thesentinels:

Thank you for contributing!

PR #3652, opened 2017-10-09
quick fix for download buttons

echohack:
The download buttons currently are not very visible. This change moves them from the `button outline` class to the `button cta` class which makes the buttons and text visible.

Before:

Before

After:

After


thesentinels:

Thank you for contributing!

reset:
tenor-236787487

PR #3669, opened 2017-10-09
On repo fetch failure, return an object with empty array

cnunciato:
We were returning only an array, so the code that expects to loop through the results of the `repos` property was failing.


thesentinels:

Thank you for contributing!

chefsalim:
tenor-234701163

PR #3690, opened 2017-10-09
Refactor logging statements in builder-depot crate

reset:
This is some of the oldest code we've got and it's about time we take
a pass and refresh the logging output. All logs that you want to wake
you up in the middle of the night are set to `error`. Anything you want
an hourly aggregate on are set to `warn`. Anything that's interesting
is set to `info`. All else is `debug` or `trace` for network messages

tenor-101149005

### DIFF
--- a/components/builder-depot/src/handlers/integrations.rs
+++ b/components/builder-depot/src/handlers/integrations.rs
@@ -21,7 +21,7 @@ use http_gateway::http::helpers;
 use iron::status::{self, Status};
 
 use protocol::originsrv::*;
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
Overall looks good especially getting rid of the error 500s.  Maybe a bit aggressive in completely removing so much error output (eg, more could be left as debug instead) but not a blocker for merge.

reset:
@chefsalim my thought here was that we're already logging on the originating server (or should be if we aren't!)

PR #3697, opened 2017-10-10
Remove builder-web package and move it's contents into builder-api-proxy

reset:
Builder-Web can't be run as a service, so it can't be configured or bound
to which means it can only be consumed as a library dependency.

Additionally, we need to pull latest of builder-web regardless of which
channel we're building against. We don't want to consume it as a package
dep because we'll always pull the stable version of it.

tenor-66064989


thesentinels:

Thank you for contributing!

PR #3698, opened 2017-10-10
Fix reading config file for builder-api-proxy

reset:
* chown the config directory so the nginx user can serve the config
* set sensible default settings for builder-api-proxy

tenor-92778419


thesentinels:

Thank you for contributing!

PR #3715, opened 2017-10-10
Don't strip Builder binaries

reset:
This change will ensure that debugging symbols are left in all
binaries in a Builder hart which are now included in our release
builds

@fnichol I know we discussed using alternative flags to the `strip` command but I think we just don't want to touch any binaries at all since we're not interested in the size of any of the Builder harts


thesentinels:

Thank you for contributing!

chefsalim:
tenor-58003889

PR #3717, opened 2017-10-10
Add 'remove' subcommand to hab pkg 

smacfarlane:
This is a first pass at adding a `pkg remove` subcommand to hab.  It's functional in its current state, so I wanted to get additional :eyes: on it.  

The approach I took is to first verify the package is installed. If more than one matches the identifier given, I assume the user meant the latest. I chose to walk all installed packages, using the IDENT file to identify a package. I then iterate on that packages TDEPS to see if it references the specific version/release of the package we're trying to remove.  If any are found, then the user is notified what packages depend on the package they tried to delete. 

A few questions I have are:
Is this the right flow and experience for the user?
Are there edge cases I haven't thought of?
Is there a better way to identify/get installed packages? 
How do I test the behavior? 

TODO:
* [x] Should return non-zero exit status if remove fails.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -239,6 +239,10 @@ pub fn get() -> App<'static, 'static> {
                     (ex: core/busybox-static/1.24.2/20160708162350)")
                 (@arg FULL_PATHS: -p "Show full path to file")
             )
+            (@subcommand remove => 
+                (about: "Remove a package from the system")
+                (@arg PKG_IDENT: +required +takes_value "Package to remove")
+            )
             (@subcommand search =>
                 (about: "Search for a package in Builder")
                 (@arg SEARCH_TERM: +required +takes_value "Search term")
--- a/components/hab/src/command/pkg/remove.rs
+++ b/components/hab/src/command/pkg/remove.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashSet;
 use std::path::Path;
 use std::fs;
 use std::fs::File;
@@ -22,7 +21,7 @@ use std::cmp::Ordering;
 
 use walkdir::WalkDir;
 
-use error::Result;
+use error::{Error, Result};
 use hcore::fs::PKG_PATH;
 use hcore::package::{PackageIdent, PackageInstall};
 use common::ui::{Status, UI};
@@ -32,7 +31,7 @@ pub fn start(ui: &mut UI, pkg_name: &PackageIdent, fs_root_path: &Path) -> Resul
     let package_ident = package_install.ident();
     let pkg_root = fs_root_path.join(PKG_PATH);
 
-    let mut dependent_packages = HashSet::new();
+    let mut dependent_packages = Vec::new();
 
     ui.begin(format!(
         "Checking for installed packages that depend on {}",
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

fnichol:
I'd like to get a chance to talk this over with @christophermaier and @reset before merging. We've talked before about behavior in the Supervisor which could prune old service packages after updates. Exposing a command like this in operator land could result in unexpected surprises so I just want to be careful here.

In short, because of the nature of our packages, there could be infinite "distros" present on disk at the same time--it just depends on your current point of view. Here's an example which might help:

img_1374

There's 4 point of views of the same set of packages. I'd argue that the Supervisor's point of view is number one (under most circumstances) which is why I'd like to think about how a Supervisor could handle some of this. In the point of view (1), the Supervisor might care to delete package `A'`.

smacfarlane:
Thanks @fnichol, that diagram is really helpful.   I was approaching this from the perspective of an operator (or cm tool) that wants to safely remove a package, forgetting that the supervisors opinion on removal is as, if not more, important.   

Would it be accurate to say that a package is safe to remove IFF no installed packages depend on it AND it's not loaded by the supervisor?  If that is true, I can re-work this to include verification that the target package is not loaded by a supervisor. 

Using the above diagram, this is the behavior I would expect: 
(Reading the arrows as "A has TDEP D", "F has TDEP A' and D" )
```
hab pkg remove A 
-- Fails, A is loaded by a supervisor

hab pkg remove D
-- Fails, D is loaded by a supervisor, A depends on D

hab pkg remove F
-- Succeeds, F ( and A'? ) is removed.

hab pkg remove B
-- Succeeds, B (and C?) are removed

hab svc unload E
hab pkg remove E
-- Succeeds, package E is removed
```

christophermaier:
@smacfarlane Based on some talks with @reset and @fnichol, I think the way we want to initially proceed with this is from the point-of-view of the Supervisor. The idea is that you could start the Supervisor with a flag that indicates how many previous versions of the services it is running you want to keep. As services are updated (either right after, or on some periodic interval), the Supervisor would take a look at what it knows about, inspect the dependency graph, and remove any packages it no longer cares about. (We would also need to specifically whitelist things like `core/hab`, `core/hab-sup`, `core/hab-butterfly`, `core/hab-launcher`, etc., since those won't be service dependencies).

In the case where you've got a Habitat Supervisor running some number of services, but you also have additional Habitat packages / utilities installed on the same machine (e.g., `jq` and other things), we may want to additionally keep any additional top-level packages and their dependencies, in addition to whatever the supervisor is running. I'm not sure how common that scenario will be in practice; perhaps it could be specified with an additional option (`--keep-non-service-packages`??).

For a user-facing command, though, I think the behavior you describe makes sense. It might be nice to provide to the user for confirmation a list of packages that will be deleted (which might be overridden with a `--yes` option, or something similar).

I think that whatever path we take to get there, in the end we'd like the Supervisor to be able to prune things on a rolling basis, and also provide a (safe!) way for users to carry out manual pruning as well. All this logic should be in one central place, though, with different behaviors based on how it ends up being invoked.

One point of clarification, though: When you say `hab pkg delete A`, what behavior do you have in mind for `origin/name`, `origin/name/version`, and `origin/name/version/release` identifiers as values for `A`?

smacfarlane:
Thanks @christophermaier, that approach makes a lot of sense.  

The behavior I would expect is for it to act like `hab pkg install`,  where version/release would be expanded to the latest installed.  One thing I haven't developed a strong opinion on yet is if it should also attempt to remove everything in the TDEPS of the specified package, if they were no longer in use by other packages. I'm leaning toward yes on that, since it would give us a mirror of the `hab pkg install` command and that feels right. 

Should I go ahead and close this out?

christophermaier:
@smacfarlane I agree, re: TDEPS deletion; as a user, I think I'd be surprised with anything else.

I'll go ahead and close this out. Thanks for taking a go at it, though!

PR #3719, opened 2017-10-11
Fix GH Hook timeouts and auto-build with a forked repo

### LINKED ISSUE #3637: GitHub Webhook fails to auto-build forked repositories
https://github.com/habitat-sh/habitat/blob/b85af98e238cfdc269bfec72b5f3156aec5d3806/components/builder-api/src/github.rs#L206-L218

We leverage the code search API in the GitHub notification hook. This API endpoint doesn't allow searching of forks unless an additional key is set, but that only will search forks which have a higher star rating than their source.

We need to find another way to search for plan.sh files in a repository or come up with another way to determine which projects are all associated with the given repository.

reset:
It is impossible to guarantee that we can search a repository for
all of it's Plan Files so we must pivot to a mechanism where we
best guess where a plan file is and then ask the user to provide
a configuration file (`.bldr.toml`) which provides us a mapping of
plan file locations and any additional files/directories which, when
modified, will cause an auto-build for a project.

With this commit we no longer search the contents of a repository to
find all plan files and then attempt to read them in. We will now
look for a `.bldr.toml` and if one isn't found fallback to a default.

* With this change we now auto-build not only if a file at a trigger
  path is changed but also if the project's value for plan_path changes
* Don't sarch for `.bldr.toml`. It will only ever be available at
  the root of a repo so we shouldn't/don't need to search for it.
  If it's not in the root theres no config.
* Gather paginated results when searching for code and return them
  as a list of search items.

Closes #3637

### DIFF
--- a/.bldr.toml
+++ b/.bldr.toml
@@ -1,7 +1,8 @@
 [builder-admin-proxy]
-paths = ["components/builder-admin-proxy/*"]
+plan_path = "components/builder-admin-proxy"
 
 [builder-admin]
+plan_path = "components/builder-admin/habitat"
 paths = [
   "components/builder-admin/*",
   "components/builder-http-gateway/*",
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

Issue #3561, opened 2017-10-05
net:route:3 / Resource temporarily unavailable errors


chefsalim:
Seeing the following errors on acceptance:

Oct 05 19:41:51 ip-10-0-0-136 hab[32474]: builder-api.acceptance(O): ERROR:habitat_depot::server: Unable to check package dependency: origin: "core" name: "nats-streaming-server" version: "0.3.6" release: "20171005193914", err: NetError(code: BUG msg: "net:route:3")
Oct 05 19:41:52 ip-10-0-0-136 hab[32474]: builder-api.acceptance(O): WARN:habitat_depot: Package not found at "/hab/svc/builder-api/data/pkgs/54/5f/core-libsodium-1.0.13-20171005193911-x86_64-linux.hart"
Oct 05 19:41:56 ip-10-0-0-136 hab[32474]: builder-api.acceptance(O): ERROR:habitat_net::conn: [err: BUG, msg: net:route:3], Connection send/recv error, Resource temporarily unavailable

reset:
We increased the socket send/recv timeouts from 5s to 10s and it alleviated the problem almost completely. I further increased the timeout to 15s. We can continue to tweak as needed, but this most certainly was a timeout occurring when reaching out to a remote service (like GItHub)

PR #3725, opened 2017-10-11
Improve "app" configuration for all Builder services

reset:
This is a lot of general code health updates but they primarily
came from tracking down the socket issues we see under load. I found
that we were locking worker-count to num-cpu and a number of
our instances only have 2 cpus.

* Changed AppCfg from a trait to a struct. It is now the same
  mechanism to configure the app portion of a Builder server config
  as it is the database, net, permissions, etc
* Simplify traits for defining App ServerState and an App Dispatcher
* Set default worker-count to num-cpu * 8. It was previously capped
  at num-cpu due to a very old way we were starting the database
  pool
* worker-count is now a configurable from the plan

tenor-145140668

Another attempt at fixing #3561

### DIFF
--- a/components/net/src/app/config.rs
+++ b/components/net/src/app/config.rs
@@ -14,11 +14,13 @@
 
 use std::fmt;
 use std::net::{IpAddr, Ipv4Addr};
+use std::str::FromStr;
 
 pub use core::config::ConfigFile;
-pub use protocol::sharding::{ShardId, SHARD_COUNT};
 use num_cpus;
 use protocol::routesrv::DEFAULT_ROUTER_PORT;
+use protocol::sharding::{SHARD_COUNT, ShardId};
+use toml;
 
 use socket::ToAddrString;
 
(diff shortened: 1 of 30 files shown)


thesentinels:

Thank you for contributing!

components/net/src/app/config.rs
```diff
@@ -53,26 +55,85 @@ impl fmt::Display for RouterAddr {
...
     ///
     /// A value of `None` indicates that this is not a sharded service.
-    fn shards(&self) -> Option<&[ShardId]>;
-
+    #[serde(default = "AppCfg::default_shards")]
+    pub shards: Option<Vec<ShardId>>,
     /// Count of Dispatch workers to start and supervise.
-    fn worker_count(&self) -> usize {
-        Self::default_worker_count()
+    #[serde(default = "AppCfg::default_worker_count")]
```
reset: The reason for this entire PR is pretty much this right here

reset:
Thumbs approved from @chefsalim :)

PR #3732, opened 2017-10-11
Fix missing router config from services

reset:
Change the [[routers]] syntax for configuring an array of hashes under
the app config key for all Builder services. This syntax isn't valid when
used in conjunction with a map itself


thesentinels:

Thank you for contributing!

PR #3738, opened 2017-10-11
Fix routing issues to non-sharded services

reset:
* The JobSrv and Scheduler are not sharded services but we haven't piped it
into the protocol yet to identify this. If a JobSrv or Scheduler registers
with a RouteSrv with just one shard, even though it only hosts that shard,
we'll get routing errors.
* Fix a crash when a connection to a RouteSrv has not been established
but the application begins to try and forward messages. There may be
a more ideal way of handling this than to just drop the message on the
floor, but this will do for now.


thesentinels:

Thank you for contributing!

PR #3750, opened 2017-10-12
Fix the --no-color flag, add a --json flag

jamessewell:
Currently this covers the following:
- all launcher output (uses raw arg processing, should this use clap?)
- all hap-sup run output

--no-color and --json are only allowed for `hab sup run` (--no-color
was on all hab sup previously, which let colour leak in from other
components which were called)

My thinking is that this is only really for sending actual sup output
to a logging system / file. Happy to change if needed.

no-color does two things (probably up for debate):
- no color in preamble etc..
- strip all color from output (could be hab messages, could be service output)

The JSON output currently has these fields:
- content
- logkey
- preamble
- file (verbose only)
- line (verbose only)
- column (verbose only)

Potentially preamble could be split to service and group?

Signed-off-by: James Sewell <james.sewell@gmail.com>

### DIFF
--- a/components/core/src/output.rs
+++ b/components/core/src/output.rs
@@ -30,9 +30,12 @@ use std::fmt;
 
 use ansi_term::Colour::{White, Cyan, Green};
 
+use regex::Regex;
+
 use PROGRAM_NAME;
 
 static mut VERBOSE: AtomicBool = ATOMIC_BOOL_INIT;
+static mut JSON: AtomicBool = ATOMIC_BOOL_INIT;
 // I am sorry this isn't named the other way; I can't get an atomic initializer that defaults to
 // true. Them's the breaks.
 static mut NO_COLOR: AtomicBool = ATOMIC_BOOL_INIT;
@@ -44,9 +47,17 @@ pub fn is_verbose() -> bool {
 
 /// Turn verbose output on or off.
 pub fn set_verbose(booly: bool) {
-    unsafe {
-        VERBOSE.store(booly, Ordering::Relaxed);
-    }
+    unsafe { VERBOSE.store(booly, Ordering::Relaxed) }
+}
+
+/// True if JSON output is enabled
+pub fn is_json() -> bool {
+    unsafe { JSON.load(Ordering::Relaxed) }
+}
+
+/// Turn JSON output on or off.
+pub fn set_json(booly: bool) {
+    unsafe { JSON.store(booly, Ordering::Relaxed) }
 }
 
 /// True if color is enabled
@@ -78,6 +89,7 @@ pub struct StructuredOutput<'a> {
     content: &'a str,
     pub verbose: Option<bool>,
     pub color: Option<bool>,
+    pub json: Option<bool>,
 }
 
 impl<'a> StructuredOutput<'a> {
@@ -99,6 +111,7 @@ impl<'a> StructuredOutput<'a> {
             content: content,
             verbose: None,
             color: None,
+            json: None,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

bixu:
This looks exactly like what @mspringfeldt and I have been waiting for.

components/sup/src/main.rs
```diff
@@ -206,9 +218,6 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
             (@arg PEER: --peer +takes_value +multiple
                 "The listen address of an initial peer (IP[:PORT])")
             (@arg PERMANENT_PEER: --("permanent-peer") -I "If this Supervisor is a permanent peer")
-            (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with[peer]
```
reset: I think you've got a rebase issue here, removing this was probably not intended :)

bixu:
@jamessewell, I think you missed https://github.com/habitat-sh/habitat/blob/master/CONTRIBUTING.md#signing-your-commits

Looking forward to using this in a release :D

jamessewell:
OK lots of git-guesses here - hopefully this has worked and is rebased now.

components/core/src/output.rs
```diff
@@ -99,65 +111,115 @@ impl<'a> StructuredOutput<'a> {
...
             verbose: None,
             color: None,
+            json: None,
         }
     }
 }
 
 // If we ever want to create multiple output formats in the future, we would do it here -
 // essentially create a flag we check to see what output you want, then call a different formatting
 // function. Viola!
```
christophermaier: We can delete this comment now 😃

components/core/src/output.rs
```diff
@@ -99,65 +111,115 @@ impl<'a> StructuredOutput<'a> {
...
+          stdfmt(self, f)
+        }
+    }
+}
+
+fn stdfmt(output: &StructuredOutput, f: &mut fmt::Formatter) -> fmt::Result {
+    let verbose = output.verbose.unwrap_or(is_verbose());
+    let color = output.color.unwrap_or(is_color());
+
+    let re = Regex::new(r"\x1b\[[0-9;]*[a-zA-Z]").unwrap();
```
christophermaier: The intention of this is to strip out any embedded ANSI color codes in output destined for a "no color" formatter, correct?

If so, is this because there are still places in Habitat's code that inappropriately embed color codes, or because output from user services may include these codes? If it's the former, we should make an issue to find those places and eliminate the color codes once and for all. In either case, we should document the intention behind this regex.

Additionally, we'll want to use the `lazy_static` macro to ensure this regex gets compiled only once, which will improve the performance of this, as it will be called _a lot_.

(Actually, we already have some functionality for this in another corner of the codebase! See https://github.com/habitat-sh/habitat/blob/db42e8970682f0a803649b0bbda9483928c70a21/components/builder-protocol/src/jobsrv.rs#L203-L218)

components/core/src/output.rs
```diff
@@ -99,65 +111,115 @@ impl<'a> StructuredOutput<'a> {
...
-            }
+            write!(f, "{}({}): {}", output.preamble, output.logkey, re.replace_all(output.content, ""))
         }
     }
 }
 
+fn jsonfmt(output: &StructuredOutput, f: &mut fmt::Formatter) -> fmt::Result {
+    let verbose = output.verbose.unwrap_or(is_verbose());
+    let color = output.color.unwrap_or(is_color());
+    let re = Regex::new(r"\x1b\[[0-9;]*[a-zA-Z]").unwrap();
```
christophermaier: Similar feedback re: regexes

components/sup/src/main.rs
```diff
@@ -115,15 +116,27 @@ fn start() -> Result<()> {
...
+    match app_matches.subcommand() {
+      (_, Some(m)) => {
+
+        if m.is_present("VERBOSE") {
+            hcore::output::set_verbose(true);
+        }
+      }
+      _ => unreachable!(),
+    } 
+
```
christophermaier: Previously, each `sub_` function checked for `VERBOSE` and `NO_COLOR`, but if I'm reading correctly, now they'll only check for `VERBOSE`. It looks like only `sub_run` now checks for `NO_COLOR`, and it is also the only one that checks `JSON`. 

Was this intentional?

components/core/src/output.rs
jamessewell: This is the intention, yes.

The Habitat colour thing is an odd issue - I had actually chased them down to remove but then you don't get them with the default behavior (highlighted words etc..). I think this is nice for console output. As there is a single field for Habitat and service outputs (it could be one or the other) it becomes difficult.

The other option would be to make it so there was a 'habitat' created output string, which was separate from the service created output string.

I was anticipating some discussion on if we should even be stripping codes from service output, I think yes (or don't use the option) but I'm sure a lot of people would say no (I would argue if you don't want it don't use it).

components/core/src/output.rs
jamessewell: Will do

components/sup/src/main.rs
jamessewell: This is intentional. 

My thinking is that VERBOSE can  be set on anything, so I pulled it up above the `sub_` calls, so we only need it once.

`sub_run` is the only one which gets `no_color` and `json` as these are really log style formatters for a headless mode - which is what `hab sup run` is.

Currently there is no way to even make `hab sup start` etc... colourless and/or json as it mixes UI calls with logging calls on stdout.

bixu:
Is there anything @mspringfeldt and I can do to help get this tested/merged/released?

christophermaier:
@jamessewell FYI, we needed to move forward with a fix for `--no-color`, which is at https://github.com/habitat-sh/habitat/pull/4334. It turned out to be broader than stripping the codes using a regex.

jamessewell:
That's fine I've had 0 habitat time sadly - hopefully that changes over
Christmas

christophermaier:
@jamessewell No worries; just wanted to keep you in the loop! Best of luck with Christmas-time Habitat Hacking 😄

elliott-davis:
Hey @jamessewell!

I'm going to close this for now but please feel free to re-open if you get some more time to hack on it!

jamessewell:
I thought this was fixed elsewhere?

I have to admit I haven’t had time to look!

elliott-davis:
Maybe? Some of the coloring stuff rings a bell. I just closed since you'd mentioned you were busy and I was doing some PR house keeping

christophermaier:
@elliott-davis @jamessewell Yeah, the coloring bit got fixed in #4334 but not the JSON bit. That may fall under #4346 now (or at least, it's in the same general area of more flexible logging).

bixu:
Would love to see this promising work around JSON-flavored logging in the Supervisor.

christophermaier:
JSON logging was added in #5033

PR #3760, opened 2017-10-12
Experimental scheduler group abort functionality

chefsalim:
Adding some initial functionality to stop scheduled group jobs. This is currently primarily for development purposes, eventually there will be a more full fledged ability to stop/restart groups and jobs.

tenor-155592577


thesentinels:

Thank you for contributing!

components/builder-depot/src/server.rs
```diff
@@ -1103,6 +1103,30 @@ fn get_schedule(req: &mut Request) -> IronResult<Response> {
     }
 }
 
+// TODO (SA) This is an experiemental dev-only function for now
+fn abort_schedule(req: &mut Request) -> IronResult<Response> {
+    let group_id = {
```
reset: This should be an authenticated request even if it is experimental so people don’t cancel each other’s jobs :)

reset:
This will be killer to get in!

components/builder-depot/src/server.rs
chefsalim: ah yeah - thanks for the catch!

chefsalim: I've made it require  worker privilege for now - getting the right access check is one of the things that will need to be done right later.

reset:
@chefsalim looks good to me!

PR #3766, opened 2017-10-12
Assorted rumor refactorings

### LINKED ISSUE #3200: Refactor rumor processing in the Supervisor
The bug in #3198 is a concrete example of a more pervasive issue in the Supervisor: the totality of rumors are needed to create a coherent internal representation of the supervision ring, but each type of rumor is processed individually, without consideration of changes in other rumor types. As seen in that bug, processing rumors "out-of-sequence" could result in inconsistent internal states that were difficult to recover from. Apart from this, the code to handle the creation of the supervisor's internal state is not as clear as it might be, since single logical tasks are often spread out across multiple functions that interact through state modification.

christophermaier:
The main things this PR does are as follows:

* Rebuild the entire census when *any* new rumors are present, in order to be more internally consistent.
* Permit binds to be initially satisfied only when there is at least one live group member

See individual commits for more detail.

Fixes #3200 
(at least in spirit; there's always more refactoring that can be done!)

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -406,7 +406,7 @@ impl Service {
             "Service update failed; unable to find own service group",
         );
         let cfg_updated = self.cfg.update(census_group);
-        if cfg_updated || census_ring.changed {
+        if cfg_updated || census_ring.changed() {
             let (reload, reconfigure) = {
                 let ctx = self.render_context(census_ring);
                 let reload = self.compile_hooks(&ctx);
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -366,7 +366,17 @@ impl Service {
     fn all_binds_satisfied(&self, census_ring: &CensusRing) -> bool {
         let mut ret = true;
         for ref bind in self.binds.iter() {
-            if census_ring.census_group_for(&bind.service_group).is_none() {
+            if let Some(group) = census_ring.census_group_for(&bind.service_group) {
+                if group.members().iter().all(|m| !m.alive()) {
```
reset: If we're splitting hairs, and we aren't but I'm going to 😅 , you can use `.any()` to abort the iterator on the first match instead of iterating to the end

reset:
@christophermaier excellent refactor, this will make this much less error prone!

tenor-98854792

PR #3771, opened 2017-10-13
Add a kubernetes exporter

### LINKED ISSUE #3500: Export package to kubernetes
Implement `hab pkg export kubernetes <pkg_name>` command, that will generate the docker image and the manifest.

The manifest file would only contain description for one service group. An example of possible generated file could be https://github.com/kinvolk/habitat-operator/blob/master/examples/standalone/habitat.yml, with the difference that the generated file would be heavily commented and all the possible fields would be specified and commented out.

krnowak:
It only prints the manifest, we leave generating the docker images to the docker exporter.

Fixes #3500.

### DIFF
--- a/components/hab/src/command/pkg/export/kubernetes.rs
+++ b/components/hab/src/command/pkg/export/kubernetes.rs
@@ -1,0 +1,95 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ffi::OsString;
+
+use common::ui::UI;
+
+use error::Result;
+
+const EXPORT_CMD: &'static str = "hab-pkg-export-kubernetes";
+
+pub fn start(ui: &mut UI, args: Vec<OsString>) -> Result<()> {
+    inner::start(ui, args)
+}
+
+#[cfg(target_os = "linux")]
+mod inner {
+    use std::ffi::OsString;
+    use std::path::PathBuf;
+    use std::str::FromStr;
+
+    use common::ui::UI;
+    use hcore::crypto::{init, default_cache_key_path};
+    use hcore::env as henv;
+    use hcore::fs::find_command;
+    use hcore::os::process;
+    use hcore::package::PackageIdent;
+
+    use error::{Error, Result};
+    use exec;
+    use super::EXPORT_CMD;
+
+    const EXPORT_CMD_ENVVAR: &'static str = "HAB_PKG_EXPORT_KUBERNETES_BINARY";
+    const EXPORT_PKG_IDENT: &'static str = "core/hab-pkg-export-kubernetes";
+    const EXPORT_PKG_IDENT_ENVVAR: &'static str = "HAB_PKG_EXPORT_KUBERNETES_PKG_IDENT";
+
+    pub fn start(ui: &mut UI, args: Vec<OsString>) -> Result<()> {
+        let command = match henv::var(EXPORT_CMD_ENVVAR) {
+            Ok(command) => PathBuf::from(command),
+            Err(_) => {
+                init();
+                let ident = match henv::var(EXPORT_PKG_IDENT_ENVVAR) {
+                    Ok(ref ident_str) => PackageIdent::from_str(ident_str)?,
+                    Err(_) => PackageIdent::from_str(&format!("{}/{}", EXPORT_PKG_IDENT, "0.1.0"))?,
+                };
+                let cmd = exec::command_from_min_pkg(
+                    ui,
+                    EXPORT_CMD,
+                    &ident,
+                    &default_cache_key_path(None),
+                    0,
+                )?;
+                PathBuf::from(cmd)
+            }
+        };
+        if let Some(cmd) = find_command(&command) {
+            Ok(process::become_command(cmd, args)?)
+        } else {
+            Err(Error::ExecCommandNotFound(command))
+        }
+    }
+}
+
+#[cfg(not(target_os = "linux"))]
+mod inner {
+    use std::env;
+    use std::ffi::OsString;
+
+    use common::ui::UI;
+
+    use error::{Error, Result};
+    use super::EXPORT_CMD;
+
+    pub fn start(ui: &mut UI, _args: Vec<OsString>) -> Result<()> {
+        let cmd = EXPORT_CMD.replace("hab", "").replace("-", " ");
+        ui.warn(format!(
+            "Running 'hab {}' on this operating system is not yet supported. \
+            Try running this command again on 64-bit Linux.",
+            &cmd
+        ))?;
+        ui.br()?;
+        Err(Error::SubcommandNotSupported(cmd.to_string()))
+    }
+}
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -666,6 +666,9 @@ fn exec_subcommand_if_called(ui: &mut UI) -> Result<()> {
         ("pkg", "export", "cf") => {
             command::pkg::export::cf::start(ui, env::args_os().skip(4).collect())
         }
+        ("pkg", "export", "kubernetes") => {
+            command::pkg::export::kubernetes::start(ui, env::args_os().skip(4).collect())
+        }
         ("run", _, _) => command::launcher::start(ui, env::args_os().skip(1).collect()),
         ("stu", _, _) | ("stud", _, _) | ("studi", _, _) | ("studio", _, _) => {
             command::studio::enter::start(ui, env::args_os().skip(2).collect())
--- a/components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
+++ b/components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
@@ -25,9 +25,11 @@ spec:
   service:
     ## Habitat topology of the service.
     topology: {{service_topology}}
+{{#if service_group}}
     ## Habitat service group name, a logical grouping of services
     ## with the same package.
-    group: default
+    group: {{service_group}}
+{{/if}}
     ## Name of the configuration secret. Edit the Kubernetes Secret
     ## at the top.
     #configSecretName: user-toml-secret
--- a/components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
+++ b/components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
@@ -1,0 +1,33 @@
+---
+## Secret for initial configuration.
+#apiVersion: v1
+#kind: Secret
+#metadata:
+#  name: user-toml-secret
+#type: Opaque
+#data:
+#  # Each configuration item needs to be encoded in base64.
+#  # Plain text content of the secret: "port = 4444"
+#  user.toml: cG9ydCA9IDQ0NDQ=
+#---
+apiVersion: habitat.sh/v1
+kind: Habitat
+metadata:
+  ## Name of the Habitat service.
+  name: {{metadata_name}}
+spec:
+  ## Name of the Habitat service package exported as a Docker image.
+  image: {{image}}
+  ## Number of desired instances.
+  count: {{count}}
+  ## An object containing parameters that effects how the Habitat service
+  ## is executed.
+  service:
+    ## Habitat topology of the service.
+    topology: standalone
+    ## Habitat service group name, a logical grouping of services
+    ## with the same package.
+    group: default
+    ## Name of the configuration secret. Edit the Kubernetes Secret
+    ## at the top.
+    #configSecretName: user-toml-secret
--- a/components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
+++ b/components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
@@ -30,6 +30,8 @@ spec:
     ## with the same package.
     group: {{service_group}}
 {{/if}}
+{{#if config_secret_name}}
     ## Name of the configuration secret. Edit the Kubernetes Secret
     ## at the top.
-    #configSecretName: user-toml-secret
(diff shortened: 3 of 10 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/export/kubernetes.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
krnowak: There is a bit of annoying code duplication happening here as it is the same as `docker.rs` in this directory with basically `s/docker/kubernetes/` applied.

krnowak: Another small difference is with version - currently the kubernetes exporter follows the version of the Habitat operator. I'm not sure whether to go this way or to follow the versioning of hab CLI.

components/hab/src/main.rs
```diff
@@ -642,6 +642,9 @@ fn exec_subcommand_if_called(ui: &mut UI) -> Result<()> {
         ("pkg", "export", "cf") => {
             command::pkg::export::cf::start(ui, env::args_os().skip(4).collect())
         }
+        ("pkg", "export", "kubernetes") => {
```
krnowak: As stated in commit message - this won't work with `hab package export kubernetes`. Also, I was considering adding a shortcut like `hab pkg export k8s`. What do you think?

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+    // clap_app!() ensures that we do have the mandatory args so unwrap() is fine here
+    let pkg_ident_str = m.value_of("PKG_IDENT").unwrap();
+    let pkg_ident = match PackageIdent::from_str(pkg_ident_str) {
+        Ok(pi) => pi,
+        Err(e) => return Err(format!("{}", e)),
+    };
+    let image = m.value_of("IMAGE").unwrap_or(pkg_ident_str);
+
+    println!(
+        r###"---
```
christophermaier: Might be good to use a separate template for this, e.g.

* https://github.com/habitat-sh/habitat/blob/64c6f5c2fb012bfc7c724eab3120d8f5240288a6/components/pkg-export-docker/defaults/Dockerfile.hbs
* https://github.com/habitat-sh/habitat/blob/64c6f5c2fb012bfc7c724eab3120d8f5240288a6/components/pkg-export-docker/src/docker.rs#L444

Also, it looks like you're sending this all to standard output; is that intentional?

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+    let m = cli().get_matches();
+    debug!("clap cli args: {:?}", m);
+    let count = m.value_of("COUNT").unwrap_or("1");
+    // clap_app!() ensures that we do have the mandatory args so unwrap() is fine here
+    let pkg_ident_str = m.value_of("PKG_IDENT").unwrap();
+    let pkg_ident = match PackageIdent::from_str(pkg_ident_str) {
+        Ok(pi) => pi,
+        Err(e) => return Err(format!("{}", e)),
+    };
+    let image = m.value_of("IMAGE").unwrap_or(pkg_ident_str);
```
christophermaier: I think this default value will cause issues, since a Habitat identifier isn't always a valid Docker image identifier.

components/pkg-export-kubernetes/src/main.rs
krnowak: I think that printing the manifest to the standard output has its uses - you can redirect the output to the new file or to append it to an already existing one, since a single yaml file can contain many k8s manifests. Also, you can do something like `hab pkg export kubernetes foo/bar | kubectl apply -f -`.

zeenix: Perhaps we should have a '--output' parameter to control this. If the parameter is not set, we output to stdout.

components/pkg-export-kubernetes/src/main.rs
zeenix: So if things remain as they are, this argument should be mandatory?

zeenix:
Just FYI, I am taking over the kubernetes exporter tool work from @krnowak, who will focus on other habitat tasks. I'll update this PR soon with first version of the tool that only create the manifest file. I'll then work on changing it to also take care of the docker image generation as well using the crate/lib provided by docker exporter.

zeenix:
Updated the PR with latest work and I think it's good to go as the first phase. Next I add docker image generation..

components/pkg-export-kubernetes/plan.sh
```diff
@@ -0,0 +1,79 @@
...
+  # Used to find libgcc_s.so.1 when compiling `build.rs` in dependencies. Since
+  # this used only at build time, we will use the version found in the gcc
+  # package proper--it won't find its way into the final binaries.
+  export LD_LIBRARY_PATH=$(pkg_path_for gcc)/lib
+  build_line "Setting LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+}
+
+do_build() {
+  pushd $PLAN_CONTEXT > /dev/null
+  cargo build ${build_type#--debug} --target=$rustc_target --verbose
```
christophermaier: Why `${build_type#--debug}`? I only see that you set it to `--release`. How does it ever get to be `--debug`?

components/pkg-export-kubernetes/src/main.rs
christophermaier: We might construct a value for `IMAGE` from the origin and package name of the given identifier.

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+fn start(_ui: &mut UI) -> result::Result<(), String> {
+    let m = cli().get_matches();
+    debug!("clap cli args: {:?}", m);
+    let count = m.value_of("COUNT").unwrap_or("1");
+    let topology = m.value_of("TOPOLOGY").unwrap_or("standalone");
+    let (group, group_commentout) = match m.value_of("GROUP") {
+        Some(g) => (g, ""),
+        None => ("default", "#"),
```
christophermaier: This value shouldn't ever be commented out though, correct? If you don't specify a group, it defaults to "default".

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,167 @@
...
+    };
+    let (config_secret_name, config_secret_name_commentout) =
+        match m.value_of("CONFIG_SECRET_NAME") {
+            Some(s) => (s, ""),
+            None => ("user-toml-secret", "#"),
```
christophermaier: The "comment out" string should be `"# "` and not `"#"` (but see earlier comment about flow control in the template)

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,167 @@
...
+        };
+    let (ring_secret_name, ring_secret_name_commentout) = match m.value_of("RING_SECRET_NAME") {
+        Some(s) => (s, ""),
+        None => ("SECRET_NAME", "#"),
```
christophermaier: Same for this comment string

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,167 @@
...
+        (@arg IMAGE: --("image") +takes_value
+            "Image of the Habitat service exported as a Docker image")
+        (@arg OUTPUT: -o --("output") +takes_value
+            "Name of manifest file to create. Pass '-' for stdout (default: -)")
+        (@arg COUNT: --("count") +takes_value {valid_natural_number}
+            "Count is the number of desired instances")
+        (@arg TOPOLOGY: -t --("service-topology") +takes_value
+            "A topology describes the intended relationship between peers \
+             within a service group. Specify either standalone or leader \
+             topology (default: standalone)")
```
christophermaier: We could validate this topology value

components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
```diff
@@ -0,0 +1,36 @@
...
+#apiVersion: v1
+#kind: Secret
+#metadata:
+#  name: user-toml-secret
+#type: Opaque
+#data:
+#  # Each configuration item needs to be encoded in base64.
+#  # Plain text content of the secret: "port = 4444"
+#  user.toml: cG9ydCA9IDQ0NDQ=
+#---
```
christophermaier: What's the intention of including this commented block?

components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
```diff
@@ -0,0 +1,36 @@
...
+    topology: {{service_topology}}
+    ## Habitat service group name, a logical grouping of services
+    ## with the same package.
+    {{service_group_commentout}}group: {{service_group}}
+    ## Name of the configuration secret. Edit the Kubernetes Secret
+    ## at the top.
+    {{config_secret_name_commentout}}configSecretName: {{config_secret_name}}
+    ## The name of the Kubernetes Secret that contains the ring key, which
+    ## encrypts the communication between Habitat supervisors.
+    {{ring_secret_name_commentout}}ringSecretName: {{ring_secret_name}}
```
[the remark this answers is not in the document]
zeenix: Oh I didn't know of that. Thanks. I'll look into that.

components/pkg-export-kubernetes/src/main.rs
zeenix: It's not the value only that's being commented out here, the whole line. Default is "default" yes.

components/pkg-export-kubernetes/src/main.rs
zeenix: I actually went with "#" but IIRC yamllint didn't like that.

components/pkg-export-kubernetes/src/main.rs
zeenix: sure but wouldn't we end up with pkg_ident_str anyway? I think i'll need to change this logic in v2 of this exporter when I add docker image generation.

components/pkg-export-kubernetes/plan.sh
```diff
@@ -0,0 +1,79 @@
...
+pkg_build_deps=(
+  core/musl core/zlib-musl core/xz-musl core/bzip2-musl core/libarchive-musl
+  core/openssl-musl core/libsodium-musl
+  core/coreutils core/rust core/gcc
+)
+pkg_bin_dirs=(bin)
+
+bin=$_pkg_distname
+
+_common_prepare() {
```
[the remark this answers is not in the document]
krnowak: Frankly speaking - this is a copy pasta of the plan.sh from the new docker exporter. I took it because the kubernetes exporter is a similar kind of component to docker exporter. So that's probably why this plan is highly magical (to me at least).

I'd be grateful, if somebody had a look at the plan and tell me what stuff there is really required and what can be dropped.

components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
krnowak: To make it easy to add an initial user.toml to the service for users not well versed in kubernetes. Maybe this could be later dropped in favor of some flag where you pass contents of the user.toml and it computes base64 value out of it and puts in the manifest.

components/pkg-export-kubernetes/src/main.rs
zeenix: Actually in the next PR i'm going to submit for also generating docker image, switches away from clap_app! and hence makes it possible to specify possible values for this option and I just made sure it does that. So if it's OK with you, i won't be changing this here.

components/pkg-export-kubernetes/src/main.rs
zeenix: I remembered incorrectly, yamllint actually warned about lack of space. The reason now I recall was that I just followed what was already being done by @krnowak and I'm guessing he went for this approach because that's a convention for commenting out directives in config files. If you look at various .conf files under /etc, you'll see this convention being followed by most. However, if you feel strong about it, I'm fine with changing it.

components/pkg-export-kubernetes/plan.sh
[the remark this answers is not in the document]
christophermaier: @krnowak Since I made this comment, I was poking around in a bunch of our other plans and found these same patterns. I have the same questions, but now no longer about this PR 😄 

This is fine for now. We (the Habitat team) need to take a pass through all our plans to modernize them anyway, so we can do the cleanup then.

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    let image = m.value_of("IMAGE").unwrap_or(pkg_ident_str);
+
+    let json = json!({
+        "metadata_name": pkg_ident.name,
+        "image": image,
+        "count": count,
+        "service_topology": topology,
+        "add_service_group": add_group,
+        "service_group": group,
+        "add_config_secret_name": add_config_secret_name,
```
christophermaier: Can the template operate on whether `service_group` is a `Some` or `None`, instead of having to pass a boolean in as well?

zeenix: didn't realize you can pass an option type to json. I'll try.. would be much better indeed.

zeenix: it works \o/

christophermaier:
@krnowak @zeenix Thanks for this!

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/300280247).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/300280247).

thesentinels:
:broken_heart: Travis CI reports [this PR failed to pass the test suite](https://travis-ci.org/habitat-sh/habitat/builds/300280247).

The next step is to examine the [job](https://travis-ci.org/habitat-sh/habitat/builds/300280247) and figure out why. If it is transient, you can try re-triggering the Travis CI Job - if it passes, this PR will be automatically merged. If it is not transient, you should fix the issue and update this pull request, and issue `approve` again. If you believe it will never pass, and you are feeling :godmode:, you can issue a `force` to merge this PR anyway.

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/300280247).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/300280247).

PR #3772, opened 2017-10-13
Pass `HAB_AUTH_TOKEN` through to the studio

elliott-davis:
tenor-256554556

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -938,6 +938,11 @@ chroot_env() {
   if [ -n "${HAB_CONFIG_EXCLUDE:-}" ]; then
     env="$env HAB_CONFIG_EXCLUDE=$HAB_CONFIG_EXCLUDE"
   fi
+  # If a Habitat Auth Token is set, then propagate it into the Studio's
+  # environment.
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/studio/bin/hab-studio.sh
```diff
@@ -1011,6 +1016,9 @@ report_env_vars() {
   if [ -n "${HAB_ORIGIN:-}" ]; then
     info "Exported: HAB_ORIGIN=$HAB_ORIGIN"
   fi
+  if [ -n "${HAB_AUTH_TOKEN:-}" ]; then
+    info "Exported: HAB_AUTH_TOKEN=$HAB_AUTH_TOKEN"
```
stevendanna: Do we think it could be worth redacting this value from the logs?  My concern would be users who are sharing build logs with each other to help solve problems, not realizing they are revealing their auth token.

fnichol: This could be a case where we could display: `Exported: HAB_AUTH_TOKEN=<redacted>` to make it clear that a value cross the Studio barrier but that its value is sensitive.

elliott-davis:
@fnichol I think I already addressed that comment. Was there something else?

reset:
tenor-53422288

PR #3777, opened 2017-10-13
Fix handler crash on account creation when duplicate email is present

reset:
We key on account name and not email. A long while ago when the account table
was first created it wasn't clear if we would key on email or account because
we weren't sure which we could be guaranteed to get. It's now clear that name
is what we can guarantee and we will perform email verification steps to try
our best to obtain an email.

This change fixes a crash in the session server when a user without an email
tries to create an account in a shard where another user has already created
an account without an email.

tenor-125031684


thesentinels:

Thank you for contributing!

raskchanky:
tenor-257649768

PR #3781, opened 2017-10-13
Return None/Empty Vec when remote GitHub entity is not found

reset:
Return None or an empty vector when GitHub returns a 404 response on
retrieving an entity or set of entities. Previously we would consider
this an error and we would warn/error spam our server output when there
was actually nothing wrong.

tenor-44784701


thesentinels:

Thank you for contributing!

raskchanky:
tenor-134880289

chefsalim:
tenor-232931134

PR #3785, opened 2017-10-15
Add Sumologic to Terraform

chefsalim:
Update Terraform to add the Sumologic agent to nodes (only builder-workers for this change, more will be added later).

tenor-86562023

### DIFF
--- a/terraform/instances.tf
+++ b/terraform/instances.tf
@@ -254,6 +254,11 @@ resource "null_resource" "datastore_provision" {
   }
 
   provisioner "file" {
+    content     = "${data.template_file.sumo_sources_worker.rendered}"
+    destination = "/home/ubuntu/sumo_sources_worker.json"
+  }
+
+  provisioner "file" {
     content     = "${data.template_file.sup_service.rendered}"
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

terraform/instances.tf
```diff
@@ -254,6 +254,11 @@ resource "null_resource" "datastore_provision" {
   }
 
   provisioner "file" {
+    content     = "${data.template_file.sumo_sources_worker.rendered}"
+    destination = "/home/ubuntu/sumo_sources_worker.json"
```
reset: I think you want to write this to `/hab/svc/sumologic/files/builder-worker.json`. We hard configure [syncSource](https://github.com/habitat-sh/core-plans/blob/master/sumologic/config/user.properties#L4) to be the files directory of the sumologic service

chefsalim: Yeah I thought about doing that write directly.  The current design is that we'll use a 'hab file upload' to populate it into that directory, so if we make any changes to the config, we can push up that change like any other config.

elliott-davis: oh nice! Carry on!

PR #3789, opened 2017-10-16
Fix issue configuring worker bldr_url when first bind member is dead

reset:
This commit is two fold, first it reverts a previous change to fix a race condition when communicating with the worker. This change appears to cause the worker to not pull jobs from the job server.

Second, `.first` doesn't guarantee it will return an alive member. Only alive members have gossip configuration applied in a service group. Using the `.first` helper with a bind can result in a stale value. This commit ensures that we find an alive member to pull our depot configuration from for the worker.


thesentinels:

Thank you for contributing!

fnichol:
Yeah, the `config.toml` makes sense.

PR #3790, opened 2017-10-16
Removed old maintenance file and updated links in Contributing

eeyun:


thesentinels:

Thank you for contributing!

tashimi:
Hey @eeyun - 

We need to decide if we're going to continue to invest in the wiki & forums, and if they are providing enough value to have the side effect of making the information in them less discoverable than if they were in the repos. 

I see you adding some links to the wiki here, which is a step in the right direction for discoverability - but there are many other docs (e.g. Readmes) that provide no information about the wiki, and a casual surfer through the Habitat repo will never find it. 

There is also a question around offline users, i.e. working on an airplane, who may not think to clone a separate wiki ahead of time, and then be cut off from information they could use. 

I'd like to split this PR into two, one updating the roles doc, which makes sense, and then have a larger conversation about how to use our different communication channels effectively, how to consolidate into as few as possible, how to ensure information is highly discoverable, and where we want to invest.

bdangit:
Information in a wiki is not as discoverable as just finding all the `*.md|*.txt* files inside a git repo.  However, the information that is inside the wiki does appear more related to those working behind the scenes to make habitat a better place.  Being blocked by a PR to merge in new content or new found wealth to improve the human side of operations is a silly thing to have.

tashimi:
@bdangit I understand your philosophy there, but it doesn't preclude having the expected `*.(md|txt)` files and linking to the wiki location so the information is discoverable. Having great and easy to contribute docs that people can't find or don't know exists is a worse problem than needing to submit a PR. 

I'd also like to request next time you choose an alternate word than "silly," something along the lines of "as a developer, it works better for me when" is significantly friendlier and more informative language.

eeyun:
All good dialogue here. I'd argue that Discoverability  != Usability , and that's not just meant to be pedantic but I think it's important to think about WRT to onboarding contributors/maintainers. The differences IMO are - With the wiki, if you want any project documentation, you'll have an index of all of it and search functionality as well as all of the trappings of treating those docs as a repository behind a single URL, vs having a bunch of `RANDOM.md` in the root of our repo that are not organized in any useful manner for passersby. 

That being said - we already had this conversation several months ago when we decided to start using the Wiki. I'm happy to re-evaluate that previous decision but I think if we are going to Opt to migrate back to `RANDOM.md` pattern then this documentation needs to either 

  - Be split into it's own repository, or
  - Be organized into a directory where we store all of it and that documentation needs to have useful filenames so that folks understand the content in each file. 

For now I'm going to Open another Issue as an RFC and we can get some dialogue happening on in order to make the decision by committee. If we can't get @habitat-sh/habitat-core-maintainers to weigh in, then we'll continue forward with the previously determined stance of using the wiki. Does  everybody feel ok with that?

PR #3791, opened 2017-10-16
Add initial_file_event argu to FileWatcher::run

asymmetric:
The additional parameter serves to control whether an event should be
emitted if, at the time the FileWatcher starts, the watched file is
already present.

In the case of the PeerWatcher, this is exactly the behaviour we want,
because in Kubernetes, when a pod starts its volumes (and the peer file
among them) have already been mounted.


thesentinels:

Thank you for contributing!

asymmetric:
I ended up going this route, instead of adding an additional function, as it seems cleaner to me.

LMKWYT.

asymmetric:
Sorry, wrong repo :)

PR #3593, opened 2017-10-07
Composites Blog Post

christophermaier:

PR #3793, opened 2017-10-16
blog about configuration plans

thommay:
We covered these in a session at the seattle community summit, and it seemed a good time to write them up.

### DIFF
--- a/www/source/blog/2017-10-15-Configuration-Plans.html.md
+++ b/www/source/blog/2017-10-15-Configuration-Plans.html.md
@@ -1,0 +1,119 @@
+---
+title: Configuration Plans
+date: 2017-10-15
+author: Thom May
+tags: [plans, configuration, glossary]
+category: Packaging
+classes: body-article
+---
+
+One of the hardest problems in computer science is naming things. But one of the
+most important things for a healthy community is shared language. As the
+Habitat community grows, we'll try to identify patterns and practices
+that help us succeed, and give them names so that we can all talk about
+them.
+
+One pattern we use a lot is what we're calling a Configuration Plan. As
+we share our packages, we've been seeing that our colleagues and
+community members use those packages in ways that we could not have
+anticipated. So what to do? We could try and add templates to
+support all possible uses, but that will rapidly become untenable. What
+good packages should do instead is provide well documented entry points, so
+that consumers of our package can easily supply their own configuration
+in a new plan -- a Configuration Plan.
+
+## Building a Configuration Plan
+
+Let's have a look at an example. There are actually a fair number of
+live examples in the habitat repository - check out the [Builder API
+Proxy](https://github.com/habitat-sh/habitat/tree/master/components/builder-api-proxy), which is a Configuration Plan for nginx. But for now, let's build our own.
+
+I've created a tiny binary that simply reads a file and prints it out.
+You can [see the
+source](https://github.com/thommay/configuration_plan_demo), and you can
+try running it by entering a studio and running the demo.
+
+```bash
+$ hab studio enter
+(studio) $ hab svc start thom/configuration_plan_demo
+(studio) $ sup-log
+```
+
+You'll see the demo print out the configuration file name, and then the
+contents of the file.
+
+```
+configuration_plan_demo.default(O): Filename is: /hab/svc/configuration_plan/config/output.txt
+configuration_plan_demo.default(O): The original plan
+
+### Making a new plan
+Now, let's create a Configuration Plan that provides a new file for the
+demo. First, we'll create a new plan:
+
+```bash
+$ hab plan init -o <your_origin> configuration_plan
+$ cd configuration_plan
+```
+
+We now have a `configuration_plan` directory, containing a `plan.sh`,
+a `config` dir, and a `hooks` dir.
+
+Let's create that configuration file!
+
+```bash
+$ echo 'a really great demo' > config/my_configuration_plan.txt
+```
+
+Open up your `plan.sh`, and replace it with the following, being sure to
+set your origin correctly.
+
+```bash
+pkg_origin=your_origin
+pkg_name=configuration_plan
+pkg_version=0.1.0
+pkg_deps=(thom/configuration_plan_demo)
+pkg_svc_run="configuration_plan_demo ${pkg_svc_config_path}/my_configuration_plan.txt"
+
+do_build() {
+  return 0
+}
+
+do_install() {
+  return 0
+}
+
+do_unpack() {
+  return 0
+}
+```
+
+Because we're shipping nothing but configuration, we tell Habitat that
+we don't need to unpack, build or install. Instead, we simply depend on
+`thom/configuration_plan_demo`, and then ensure that we use our new
+configuration file.
+
+### Running the plan
+Now we can build and run our plan. Enter a studio as usual, run `build`
+to build your plan, and then install it. On my machine, I ran:
+
+```bash
+$ hab studio enter
+(studio) $ build
+(studio) $ hab pkg install /src/results/thom-configuration_plan-0.1.0-20171016162630-x86_64-linux.hart
+(studio) $ hab svc start thom/configuration_plan
+```
+
+By running `sup-log`, we should see our demo start and use the new
+config file.
+
+```
+configuration_plan.default(O): Filename is: /hab/svc/configuration_plan/config/my_configuration_plan.txt
+configuration_plan.default(O): a really great demo
+```
+
+## Conclusion
+
+Rather than needing to build our own specific package, Configuration Plans
+allow us to tailor widely used applications to our exact purposes. They
+also free the maintainer of the application to provide up to date
+packages, without trying to support every use case themselves.


thesentinels:

Thank you for contributing!

thommay:
@eeyun bloggage about configuration plans

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+(studio) $ sup-log
+```
+
+You'll see the demo print out the configuration file name, and then the
+contents of the file.
+
+```
+configuration_plan_demo.default(O): Filename is: /hab/svc/configuration_plan/config/output.txt
+configuration_plan_demo.default(O): The original plan
+
```
mwrock: looks like you are missing the closing ```

christophermaier:
@thommay This is nice timing, because I have a pending blog post that uses this same pattern! https://github.com/habitat-sh/habitat/pull/3593

I can reword mine to refer to yours and publish either with or after for some nice synergy / reinforcement :)

It could also serve as a slightly more real-world example of the idea, which might address the feedback @mwrock had (though mine is still kind of a toy example).

It might also be nice to call back to @cnunciato's earlier post that also illustrates this idea (before it had a name), which I shamelessly copied for my post :smiley: https://www.habitat.sh/blog/2017/08/Packaging-a-Website-with-Nginx-and-Habitat/

There is one wrinkle that more widespread usage of this pattern might expose more, which you can see in the nginx config file from my blog post. Because `core/nginx` compiles nginx with paths specific to `core/nginx` (assuming that *that* is the package that's going to be running, at the end of the day), I had to override some things in my config that I didn't really need for my functionality, but had to do in order to get the binary to run. This isn't a problem for you to fix, of course. It may just be a shortcoming of the `core/nginx` package itself. I just realized that I hadn't really brought this up elsewhere, though, and thought I'd share, since this post reminded me of it 😄 

Nice post!

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+  return 0
+}
+
+do_install() {
+  return 0
+}
+
+do_unpack() {
+  return 0
+}
```
fnichol: You shouldn't need to override `do_unpack()` unless I've missed something?

reset:
@thommay great article man! We definitely are seeing this pattern emerge and it's because of a big misstep with `core/nginx` in general - it shouldn't be a runnable service. We've avoided giving names to these patterns because we intend on fixing the experience in the software instead of making a companion guide on how to be successful with Habitat. I think this is a great article and you should definitely kick it out, but I would instead rephrase this as "How you use the NGINX plan" instead of giving it a generic pattern name. This just simply *is* how you use the NGIXN plan :)

eeyun:
I second the opinion of using NGINX to tell this story I think. But based on the discussions we had at summit I don't think I agree that we should rephrase from defining a pattern here. Likely if this pattern is in fact something people are doing (which they definitely are) then our build-time features will provide better experiences when enacting this pattern.

I could be convinced otherwise, but I think "putting a name to a face" is just fine when we're seeing this pop up a bunch. That being said I think this is a pattern we want to encourage people to follow. In doing so, the bar for what gets accepted into core simultaneously raises, and gets easier for contributors to write. (This is a super digression but I figure since we're talking about it here, we might as well talk about it here :) )

mwrock:
I agree with @eeyun that this pattern is still a viable pattern in spite of the unideal nginx case. I have bumped into other scenarios where this fits when I want to leverage an existing core service plan but basically want to hijack its configuration without requiring the use of `hab config apply` or feeding in toml to the `hab start`. An example is https://github.com/mwrock/national-parks where I create a `np-mongodb` plan to depend on `core/mongodb` and add in my own config as well as a `post-run` hook to seed the data.

reset:
@mwrock we had a good discussion at dinner with @thommay about this and a video chat with @eeyun before hand. There's definitely something here but there may be levels of when we suggest something like this. For example, it's a 1:1 ratio with NGINX. We shouldn't even let NGINX be started without you depending upon it and starting it yourself.

There's probably a middle level where it makes sense for somebody to do this but also the software without your own config is useful - perhaps HAProxy is a good example of this.

Databases would be a different level. I actually would strongly suggest we _never_ recommend for people to use a pattern like this when consuming a database. There is a whole lot of work that goes into those particular plans to get databases to replicated, be highly available, etc. It was a huge mistake for me to wrap the postgresql plan and call it `builder-datastore` but I had done it to protect the live service against upstream changes during a time of big churn. Now that things are stable it's going to take a downtime for us to switch to the new postgresql plan.

It may be a thing that we're missing a feature to apply our own configuration override with a package in an easy way, but doing it this way is more or less vendoring a plan without recourse to receive upstream changes

eeyun:
Yep theres a smell of feature afoot. I'm on board with getting this merged with the caveat that we add some notes when/why you would do this.

thommay:
I'm going to rewrite this next week to get more of a concept in about when you should use this pattern, how you get from "i just need a new attribute" to "my whole config has to change" and so on.

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+```
+configuration_plan_demo.default(O): Filename is: /hab/svc/configuration_plan/config/output.txt
+configuration_plan_demo.default(O): The original plan
+
+### Making a new plan
+Now, let's create a Configuration Plan that provides a new file for the
+demo. First, we'll create a new plan:
+
+```bash
+$ hab plan init -o <your_origin> configuration_plan
```
burtlo: When you run `hab plan init` I'm certain that it will already use their origin by default if they have gone and done the setup.

www/source/blog/2017-10-15-Configuration-Plans.html.md
burtlo: As long as you don't have a `pkg_source` the `do_download` and `do_unpack` will be ignored.

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Building a Configuration Plan
+
+Let's have a look at an example. There are actually a fair number of
+live examples in the habitat repository - check out the [Builder API
+Proxy](https://github.com/habitat-sh/habitat/tree/master/components/builder-api-proxy), which is a Configuration Plan for nginx. But for now, let's build our own.
+
+I've created a tiny binary that simply reads a file and prints it out.
+You can [see the
+source](https://github.com/thommay/configuration_plan_demo), and you can
```
burtlo: In the realm of naming hard things: would you consider a different name for this package?  Something that perhaps denotes that it's the application with with the default/standard configuration.

burtlo: Or call it what it does: config file reader/reporter?

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+(studio) $ hab svc start thom/configuration_plan_demo
+(studio) $ sup-log
+```
+
+You'll see the demo print out the configuration file name, and then the
+contents of the file.
+
+```
+configuration_plan_demo.default(O): Filename is: /hab/svc/configuration_plan/config/output.txt
+configuration_plan_demo.default(O): The original plan
```
burtlo: Perhaps "This is the configuration content defined in the original plan"

www/source/blog/2017-10-15-Configuration-Plans.html.md
burtlo: What's the actual naming convention here when someone creates a new plan that wraps your plan and provides new configuration.

* Would you name it the same? Essentially you are using it in the same way. But I could see for some applications (looking at you *nginx*) that when given different configuration files completely change.

* Would you name it what it becomes? In this case if you were to give it a whole new configuration and it performs in a different capacity would you call it something different? I think when we were building some learning content around this we defaulted to call our configuration plan version of `core/httpd` the `learn-chef/webapp` as it now behaved in way that served our web content.

> An aside, we also did something which was probably an anti-pattern and packaged our website with the updated config files. I saw in the nginx blog post the concept of storing the site in a completely different package and installing it this webapp one

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+$ cd configuration_plan
+```
+
+We now have a `configuration_plan` directory, containing a `plan.sh`,
+a `config` dir, and a `hooks` dir.
+
+Let's create that configuration file!
+
+```bash
+$ echo 'a really great demo' > config/my_configuration_plan.txt
```
burtlo: Could the demo content be something that counters each other in a person's mind to reinforce the point that this for situations when you want to replace the existing configuration?

Perhaps the original configuration states:
> I love Mondays and anchovies!
Perhaps the updated configuration states:
> I love Fridays and pizza!

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+
+Open up your `plan.sh`, and replace it with the following, being sure to
+set your origin correctly.
+
+```bash
+pkg_origin=your_origin
+pkg_name=configuration_plan
+pkg_version=0.1.0
+pkg_deps=(thom/configuration_plan_demo)
+pkg_svc_run="configuration_plan_demo ${pkg_svc_config_path}/my_configuration_plan.txt"
```
burtlo: Perhaps people get it, but it may be useful to point out why you can use the `configuration_plan_demo` executable because you are defining a runtime dependency above with the `pkg_deps`. So you are getting the built executable on the path.

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+
+```bash
+$ echo 'a really great demo' > config/my_configuration_plan.txt
+```
+
+Open up your `plan.sh`, and replace it with the following, being sure to
+set your origin correctly.
+
+```bash
+pkg_origin=your_origin
```
burtlo: For sure they should replace it. If they don't, it won't cause a problem when you attempt to build it in the studio.

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+configuration file.
+
+### Running the plan
+Now we can build and run our plan. Enter a studio as usual, run `build`
+to build your plan, and then install it. On my machine, I ran:
+
+```bash
+$ hab studio enter
+(studio) $ build
+(studio) $ hab pkg install /src/results/thom-configuration_plan-0.1.0-20171016162630-x86_64-linux.hart
```
burtlo: The build process will automatically install after the build so you don't have to install it. I still feel like I should mention it when talking to new Habitat users. It tells you that in the output. But who reads all that output at first.

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+(studio) $ hab svc start thom/configuration_plan
+```
+
+By running `sup-log`, we should see our demo start and use the new
+config file.
+
+```
+configuration_plan.default(O): Filename is: /hab/svc/configuration_plan/config/my_configuration_plan.txt
+configuration_plan.default(O): a really great demo
```
burtlo: I think that this simplified demo begs the question, why wouldn't I ask the template maintainer to add a template variable for this line of text. It definitely works as a demo as you are pointing to a different file path in the service run operation in the plan and that is not something that you can override. But in some ways the pointing to a different file almost, without the configuration defaults, doesn't feel like your are updating configuration.

I guess what I'm trying to say is that it feels almost a little too trivial.

burtlo: Could it remain trivial and at the end here some more examples could be provided like you mentioned above (like the nginx)?

I know that I have used this pattern when the `core/haproxy` package was not doing what I wanted and then configuration was broken for me at the time of using it (https://github.com/learn-chef/hab-two-tier/blob/master/haproxy/config/haproxy.conf). I also did this with`core/httpd` (https://github.com/learn-chef/hab-two-tier/tree/master/webapp) because the default behavior was not working correctly.

> The httpd example is a pretty gnarly one. If you look within the conf I use the `pkgPathFor "core/httpd"` for the ServerRoot because the original httpd requires that you have a few directories. It was either that or make a bunch of symlinks.

www/source/blog/2017-10-15-Configuration-Plans.html.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+## Conclusion
+
+Rather than needing to build our own specific package, Configuration Plans
```
burtlo: This article made me think of a flow of control:

* If you need to change how the original package is built you should... *fork the plan*
* If you need to change how the original service runs you should  *make a configuration/hooks plan*.
* If you need to change the configuration default values the service uses when it starts up you should *make a new defaults plan*.  
* If you are ok with starting the original default values and changing them when running you should *create a new defaults file and `hab config apply`*

burtlo: Thinking about this and writing more it makes me wonder if calling this 'configuration plan' does not describe that you can change how the application runs not how it is built. I guess what I mean is that because you can change the configuration, hooks, defaults , service users, etc. it really encompasses a whole lot more when you repackage this plan.

burtlo: And after rebuilding my httpd example I see that you *have* to define all the run things like default values and hooks.

I guess within your a hook, if you wanted to use the original, your could always say:
```
# habitat/hooks/init
source {{pkgPathFor "core/httpd"}}/hooks/init
# ... all the things that I may need to do ... 
```

christophermaier:
@thommay Looking through the discussion on this, as well as your last update, is it safe to say that we should close this out, pending a revised article / approach to this issue?

christophermaier:
Closing this out for now, pending a rewrite as suggested by @thommay above.

PR #3795, opened 2017-10-16
Add Studio dev env improvements and docs

elliott-davis:
tenor-25213883

### DIFF
--- a/BUILDER_DEV.md
+++ b/BUILDER_DEV.md
@@ -213,6 +213,30 @@ HTTP/1.1 201 Created
 }
 ```
 
+## Habitat Studio Based Development Environment
+This will start everything to run builder in a studio except the web app.
+
+#### Prerequisites
+1. Ensure you have a copy of the [package archive](http://nunciato-shared-files.s3.amazonaws.com/pkgs.zip) in the habitat directory
+1. Copy `habitat-builder-dev.2017-10-02.private-key.pem` from 1Password into `/src/.secrets/builder-github-app.pem`
+
+#### Notes for OS X
+* Enusre you have direnv installed and configured
+* The `.envrc` file will override several Habitat environment variables based on what is in your `cli.toml` file
+
+#### Notes for Vagrant VM
+* `cd /vagrant`
+* `direnv allow`
+
+1. `hab studio enter` (This will download and install and start the builder service)
+1. `sl` Wait until you see the origin service start
+1. `origin`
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

eeyun:
Can I ask, could this go into the wiki instead of in the root of the repository, or is this something that needs to live inside the source code?

elliott-davis:
@eeyun I moved it into to BUILDER_DEV. I'm also happy to add it to the wiki if you think it'd be useful there

BUILDER_DEV.md
```diff
@@ -213,6 +213,17 @@ HTTP/1.1 201 Created
...
 
+## Habitat Studio Based Development Environment
+This will start everything to run builder in a studio except the web app
+
+1. `hab studio enter` (This will download and install and start the builder service)
+1. `export HAB_AUTH_TOKEN=<my_token>` (Unless you have one set in your cli.toml)
+1. `sl` Wait until you see the origin service start
+1. `origin`
+1. `keys`
+1. `load_packages`
```
cnunciato: If this relies on an actual set of packages somewhere, should we explain where they should go, and link to the archive somehow?

BUILDER_DEV.md
```diff
@@ -213,6 +213,17 @@ HTTP/1.1 201 Created
...
+### Web development
+This is a great environment to use for development on builder-web. Just make sure you're running builder-web locally. Everything should follow the standard web setup.
```
cnunciato: We should probably link to the builder-web readme here.

cnunciato:
Nice! Only other suggestions I'd make, then, would be to mention that `direnv` is required (installed and sourced into your shell), a mention of how cli.toml is used, and maybe a tweak to honor environment variables if they're set. Otherwise:

tenor-167359958

eeyun:
I totally wrote a response to this and never posted it because I am a 100% interrupt driven worker :)

I'm happy to have it where you've got it now. My preference would be that we expose more of this type of documentation into the [wiki](https://github.com/habitat-sh/habitat/wiki) moving forward. Just so we can centralize and clean out the `RANDOM.md` files spread throughout the repository. There's (currently) no obligation to do this but I think so long as our README.md makes mention of the wiki and links to wiki articles we'll be providing a better contributor experience by going that path. That, and it's still super easy to clone and change the wiki via CLI if that's your bag.

PR #3814, opened 2017-10-18
Add an alternative path for user.toml

krnowak:
This commit makes the supervisor to look for user configuration in two
locations: `/hab/svc/<pkgname>/user.toml` and
`/hab/user/<pkgname>/config/user.toml`. If the first path is missing, it
will try to load the configuration from the second one.

The reason is to work around an issue in Kubernetes/Docker/Linux
kernel.

In the Habitat operator we use the Kubernetes Secret feature for
putting user configuration as an initial configuration and as a source
of configuration updates. Initially it worked fine, until we
discovered that Kubernetes actually mounts a directory at
`/hab/svc/<pkgname>` to put `user.toml` there, so all the previous
contents there were effectively hidden.

We worked it around with a SubPath feature of Kubernetes, which
basically instead of mounting a directory at `/hab/svc/<pkgname>`,
bind-mounts some file on the host to the `/hab/svc/<pkgname>/user.toml`
path inside the Kubernetes pod.

This unfortunately breaks updating the configuration via the
Kubernetes Secret. This is because Kubernetes tries hard to make the
change of the Secret atomic. The atomicity of the change is achieved
with renaming. Doing the rename changes the inode of the target file
and since the bind mounts are based on inodes, the change is not
reflected in the pod.

It may take some time before we come up with a solution for this
problem and the solution to be released, so for now we would like to
add the second path for the user configuration. The new
`/hab/user/<pkgname>/config` directory can only contain `user.toml` file,
so it is safe to mount a directory here with our own copy of the file,
and we can stop using SubPath feature, so the updates should work
again.

Signed-off-by: Krzesimir Nowak <krzesimir@kinvolk.io>

Also, there seem to be an agreement that this file should be moved out of the `/hab/svc` directory, so users do not need to mess with the service specific files there.

### DIFF
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -467,8 +467,6 @@ mod test {
     use toml;
     use tempdir::TempDir;
 
-    use hcore::package::{PackageIdent, PackageInstall};
-
     use super::*;
     use error::Error;
 
@@ -763,20 +761,14 @@ mod test {
 
     #[test]
     fn serialize_config() {
-        let pkg_id = PackageIdent::new("testing", "testing", Some("1.0.0"), Some("20170712000000"));
-        let pkg_install = PackageInstall::new_from_parts(
-            pkg_id.clone(),
-            PathBuf::from("/tmp"),
-            PathBuf::from("/tmp"),
-            PathBuf::from("/tmp"),
-        );
-        let pkg = Pkg::from_install(pkg_install).expect("Could not create package!");
         let concrete_path = TempDir::new("habitat_config_test").expect("create temp dir");
-
-        let mut cfg = Cfg::new(&pkg, Some(&concrete_path.as_ref().to_path_buf()))
-            .expect("Could not create config");
-
-        let default_toml = "shards = []\n\n[datastore]\ndatabase = \"builder_originsrv\"\npassword = \"\"\nuser = \"hab\"\n";
+        let pkg = TestPkg::new(&concrete_path);
+        let mut cfg = Cfg::new(&pkg, None).expect("Could not create config");
+        let default_toml = "shards = []\n\n\
+                            [datastore]\n\
+                            database = \"builder_originsrv\"\n\
+                            password = \"\"\n\
+                            user = \"hab\"\n";
 
         cfg.default = Some(toml::Value::Table(
             toml::de::from_str(default_toml).unwrap(),
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -23,6 +23,7 @@ use std::path::{Path, PathBuf};
 use std::result;
 
 use ansi_term::Colour::Purple;
+use fs;
 use hcore::crypto;
 use serde::{Serialize, Serializer};
 use serde::ser::SerializeMap;
@@ -64,7 +65,8 @@ impl Cfg {
             &package.path,
         );
         let default = Self::load_default(&pkg_root)?;
-        let user = Self::load_user(&package)?;
+        let user_config_path = Self::determine_user_config_path(&package);
+        let user = Self::load_user(&user_config_path)?;
         let environment = Self::load_environment(&package)?;
         return Ok(Self {
             default: default,
@@ -165,8 +167,36 @@ impl Cfg {
         Self::load_toml_file(config_from, "default.toml")
     }
 
-    fn load_user(package: &Pkg) -> Result<Option<toml::Value>> {
-        Self::load_toml_file(&package.svc_path, "user.toml")
+    fn determine_user_config_path(package: &Pkg) -> PathBuf {
+        let recommended_dir = fs::user_config_path(&package.name);
+        let recommended_path = recommended_dir.join("user.toml");
+        if recommended_path.exists() {
+            return recommended_dir;
+        }
+        debug!(
+            "'user.toml' at {} does not exist",
+            recommended_path.display()
+        );
+        let deprecated_dir = package.svc_path.clone();
+        let deprecated_path = deprecated_dir.join("user.toml");
+        if deprecated_path.exists() {
+            outputln!(
+                "The user configuration location at {} is deprecated, \
+                 consider putting it in {}",
+                deprecated_path.display(),
+                recommended_path.display(),
+            );
+            return deprecated_dir;
+        }
+        debug!(
+            "'user.toml' at {} does not exist",
+            deprecated_path.display()
+        );
+        recommended_dir
+    }
+
+    fn load_user<T: AsRef<Path>>(path: T) -> Result<Option<toml::Value>> {
+        Self::load_toml_file(path, "user.toml")
     }
 
     fn load_environment(package: &Pkg) -> Result<Option<toml::Value>> {
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -44,6 +44,34 @@ static ENV_VAR_PREFIX: &'static str = "HAB";
 /// for a single service.
 static TOML_MAX_MERGE_DEPTH: u16 = 30;
 
+/// Trait for getting paths to directories where various configuration
+/// files are expected to be.
+pub trait PackageConfigPaths {
+    /// Get name of the package (basically name part of package ident.
+    fn name(&self) -> String;
+    /// Get path to directory which holds default.toml.
+    fn default_config_dir(&self) -> PathBuf;
+    /// Get recommended path to directory which holds user.toml.
+    fn recommended_user_config_dir(&self) -> PathBuf;
+    /// Get deprecated path to directory which holds user.toml.
+    fn deprecated_user_config_dir(&self) -> PathBuf;
+}
+
+impl PackageConfigPaths for Pkg {
+    fn name(&self) -> String {
+        self.name.clone()
+    }
+    fn default_config_dir(&self) -> PathBuf {
+        self.path.clone()
+    }
+    fn recommended_user_config_dir(&self) -> PathBuf {
+        fs::user_config_path(&self.name)
+    }
+    fn deprecated_user_config_dir(&self) -> PathBuf {
+        self.svc_path.clone()
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct Cfg {
     /// Default level configuration loaded by a Package's `default.toml`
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/package.rs
```diff
@@ -101,6 +101,8 @@ pub struct Pkg {
     pub svc_run: PathBuf,
     pub svc_user: String,
     pub svc_group: String,
+    pub user_path: PathBuf,
```
krnowak: I'm not sure if adding new fields here create some serialization problem. Like during the update of supervisor - we serialize this struct, store it on the disk, update the supervisor, deserialize the struct from the disk and we get empty fields for `user_path` and `user_config_path`. Is this something to be defensive about (by detecting if the fields are empty, for instance)?

krnowak:
CC: @fnichol, @christophermaier, @LiliC, @asymmetric

We talked about it on the last meeting.

krnowak:
Some food for thought:
- Should we prefer the new path over the old one?
  - If so, then it means that if both `/hab/svc/<pkgname>/user.toml` and `/hab/user/<pkgname>/config/user.toml` files exist, we read the latter, not the former, like the patch currently does.
- Should the path preference be persistent over the service's lifetime?
  - This is to clarify whether Supervisor should read the less preferred path after the more preferred path was removed sometime during the service's lifetime (so the service already got some config from the file in the more preferred location) or it should just think that there is no user configuration at all.
- Should we deprecate the `/hab/svc/<pkgname>/user.toml` path?
  - If so, we probably should print a warning about the deprecated path, when reading config from there.
- When implementing the user.toml watcher, should we watch for both old and new paths?
  - Doing this may be a bit complicated.
  - Since this new path and the user.toml watcher are new features, maybe only watching the new path would be a good option.

asymmetric:
I personally think that the user-config watcher should only watch the new path (`/hab/user/foo/config/user.toml`).

The reason is that the user-config watcher is a new feature, so this won't break anybody's workflow.

I also think the old location should be deprecated with a warning, and removed in future releases. The `user.toml` (or its location) hasn't been widely documented yet, so I think now is the right time.

components/sup/src/manager/service/config.rs
```diff
@@ -43,6 +43,50 @@ static ENV_VAR_PREFIX: &'static str = "HAB";
...
+
+    fn steal_path(self) -> PathBuf {
+        match self {
+            UserConfigPath::Recommended(p) |
+            UserConfigPath::Deprecated(p) => p,
+        }
+    }
+}
+
+pub trait PackageForConfig {
```
christophermaier: I'm not sure the name of this trait accurately captures the intention behind the trait; what does `PackageForConfig` mean?

components/sup/src/manager/service/config.rs
```diff
@@ -170,8 +215,54 @@ impl Cfg {
...
+        for ucp in ucps {
+            match File::open(ucp.get_path()) {
+                Ok(file) => {
+                    if let UserConfigPath::Deprecated(_) = ucp {
+                        outputln!(
+                            "The {} at {} is deprecated, consider putting it in {}",
+                            "user configuration location",
+                            ucp.get_path().display(),
+                            recommended_path.display(),
+                        );
```
christophermaier: This format string has three placeholders, but only two values are supplied.

components/sup/src/manager/service/config.rs
```diff
@@ -56,17 +100,20 @@ pub struct Cfg {
...
         let pkg_root = config_from.and_then(|p| Some(p.as_path())).unwrap_or(
-            &package.path,
+            package.path(),
         );
         let mut cfg = Cfg::default();
         cfg.load_default(&pkg_root)?;
-        cfg.load_user(&package)?;
-        cfg.load_environment(&package)?;
+        cfg.load_user(package)?;
+        cfg.load_environment(package)?;
```
christophermaier: Ah, thanks for cleaning up those unnecessary `&`s 👍

components/sup/src/manager/service/config.rs
```diff
@@ -170,8 +215,54 @@ impl Cfg {
...
+                Err(e) => {
+                    debug!(
+                        "Failed to open 'user.toml', {}, {}",
+                        ucp.get_path().display(),
+                        e,
+                    );
+                }
+            }
+        }
+        // No user config found at all. Next time try loading only the
```
christophermaier: When is "next time"? As far as I can see, this is only ever called when creating a new `Cfg` struct.

I presume this is laying the groundwork for refreshing the `user.toml` content based on file watch events... in that case, is the intention to never change which `user.toml` path is being looked at? If this starts up and a `user.toml` in the new location is not present, but one at the old path is present, won't this always look at the old path, even if a new one shows up later? If that's the case, it may simplify this code and make it easier to reason about if we split this function into two functions: one that does the initial determination of which path we're going to look at, and the other that does the actual consumption of content. Actually, in that case, we could probably just use the original code after we figured out which path to consult.

components/sup/src/manager/service/package.rs
christophermaier: Hrm... that's a good question. I wonder if these really need to be public members of the `Pkg` struct at all; maybe they should just stay function calls (which you use with your `PackageForConfig` trait). You could just use the `fs::user_path(&package.ident.name)`, etc. implementations directly in your trait implementation for `Pkg`, get the exact same behavior, and avoid any serialization issues.

components/sup/src/manager/service/config.rs
krnowak: It has three values, first is the `"user configuration location"` string. I did it that way, otherwise the line was over 100 chars long and rustfmt was bailing out… I'll split it out to a separate function so I'll have less indentation and more space for the string.

components/sup/src/manager/service/config.rs
krnowak: Yeah, we basically want to stick to a single path, which is determined at the time when we load the config for the first time.

I'll have a look at splitting the functions. I also had an idea of cleaning up the load_{default,user,environment} functions, so they don't take the `self` parameter. This would be to avoid using `Cfg::default`, which forces the `user_config_parameter` to be an `Option`, because it's value is set sometime later in the `new` function.

components/sup/src/manager/service/config.rs
```diff
@@ -43,6 +43,50 @@ static ENV_VAR_PREFIX: &'static str = "HAB";
...
+
+impl UserConfigPath {
+    fn get_path(&self) -> &PathBuf {
+        match self {
+            &UserConfigPath::Recommended(ref p) |
+            &UserConfigPath::Deprecated(ref p) => p,
+        }
+    }
+
+    fn steal_path(self) -> PathBuf {
```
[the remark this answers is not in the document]
krnowak: Ah, the `From<UserConfigPath>` looks like the way to go, thanks.

components/sup/src/manager/service/config.rs
krnowak: I'll try to come up with some better name for it. And some docs. Probably will take me more time than addressing the rest of the issues here. ;)

This trait was meant to be basically a subset of the `Pkg` struct, so I could mock it in tests.

components/sup/src/manager/service/config.rs
christophermaier: Ah, you're right... I totally missed that.

You can also split long strings up, like we do here: https://github.com/habitat-sh/habitat/blob/cc079b61663b826a4da9e88939eaeaffdf3acb3e/components/sup/src/main.rs#L160-L162

krnowak: Ah, nice. That's exactly what I need. Thanks!

krnowak:
Updated the PR. I dropped the user config path persistence for now, I'll make it a part of the follow up PR with the user config watching functionality. But I have refactored `Cfg` a bit, so it does not use `Cfg::default()` in the constructor.

components/sup/src/manager/service/config.rs
```diff
@@ -63,11 +63,16 @@ impl Cfg {
...
+        let default = Self::load_default(&pkg_root)?;
+        let user = Self::load_user(&package)?;
+        let environment = Self::load_environment(&package)?;
+        return Ok(Self {
+            default: default,
+            user: user,
+            gossip: None,
+            environment: environment,
+            gossip_incarnation: 0,
+        });
```
christophermaier: I like how this ended up!

components/sup/src/manager/service/config.rs
```diff
@@ -651,6 +655,138 @@ mod test {
...
+            file.flush().expect("flush changes in toml file");
+
+            file.seek(SeekFrom::Start(0)).expect(
+                "seek in toml file to start for checking",
+            );
+            let mut buf = String::new();
+            file.read_to_string(&mut buf).expect(
+                "read value from toml file",
+            );
+            assert_eq!(buf, self.text);
```
christophermaier: I think this implementation is more complex than it needs to be for these tests; you probably just need a `write_all` and a `flush`. The `assert_eq!` is basically testing that Rust's I/O code is working.

Beyond that, though, I think there's a larger opportunity to simplify this test code. There's a lot of repetition in the test cases themselves around opening a file, creating a representation of the desired TOML, and then writing it. In the end, you just want one function that takes a path and some content, and writes that data to the right place. That one function could hide all the details of file opening, formatting, and writing, which are really incidental to the tests themselves. When you're reading the tests, you just want one line that says "OK, I'm creating this config file at this location"; you don't want to worry about the details of how that's done.

krnowak: I added this assert_eq, because earlier I had a `reset` function that cleared a file and wrote new contents there. So I truncated it and then wrote new contents only to find out that I got a bunch of zeros at the beginning of the file, because truncating is not seeking.

Anyway, I'll have a look how to make it simpler.

krnowak:
Updated. The tests are now indeed simpler.

christophermaier:
tenor-143582723
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/298127532).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/298127532).

PR #3820, opened 2017-10-18
Add user to nginx config to avoid permissin issues on upgrade

elliott-davis:
tenor-80232704

### DIFF
--- a/components/builder-api-proxy/config/nginx.conf
+++ b/components/builder-api-proxy/config/nginx.conf
@@ -1,5 +1,6 @@
 daemon off;
 pid {{pkg.svc_var_path}}/pid;
+user hab hab;
 
 worker_processes  {{cfg.nginx.worker_processes}};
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
I'm not quite sure how this will fix the issue if the supervisor is setting directory permissions to root?

reset:
I don't believe this is the fix that we're looking for - this is a long time bug that came about over a year and a half ago when we merged some work to execute run hooks as `pkg_svc_user`. @fnichol and I discussed this recently and there's a few options here:

1. Consider switching back to forcing people to use `chpst` (change process state) if they define a custom run hook instead of just using `pkg_svc_run`. We would need to change a number of plans and educate people in the docs about this behaviour
1. Add an additional variable to a plan to define which user should run the hook. `pkg_svc_init_user` or something like along those lines. This user would be who runs the hook but isn't who owns the data in the directories

I'm going to close this for now but thanks for the stab at it @elliott-davis

PR #3827, opened 2017-10-19
Consolidate builder job server and scheduler services

chefsalim:
Dim the lights, pour a glass of a fine beverage of your choice, put on some music, and put your feet up. Are you resting comfortably? Then let's begin.

In the bad old days there were monolithic services. These giant monsters quickly became incomprehensible and unmanageable, and devoured most engineers who dared to try and constrain them. Those days thankfully passed, and we moved on to the micro-services era. Now services were small and easy enough to understand, and even refactor in a single sitting. But, like Captain Ahab's peg holes drilled into the deck of the Pequod, they became numerous and comfortable crutches which we could use even when they were not needed. Such is the sad case of the builder-scheduler service - it came into being as an offshoot of the builder-jobsrv, had a nice independent existence, and is now heading back to its origins. Thankfully, this story has a happy ending. The combined job-srv and scheduler are still small and nimble, and as a bonus eliminate a class of issues that came with a protocol that is ultimately more efficient when not decoupled via a network boundary.

---

There's no getting around the fact that this is a large PR. However, the good news is that it is _mostly_ (but not completely) a mechanical integration of the scheduler into the job server. The main things in this PR are:

* Renaming of the scheduler protobuf types. These now fit into the Job protobuf hierarchy, and also are prefixed more consistently to indicate their taxonomy. For example, what was previously just a Package, is now a JobGraphPackage. This lets us more easily distinguish it from other Package types, such as OriginPackages.

* Minor clean-up of the DB artifacts. Since we are moving the scheduler into the jobsrv DB, this is an easy opportunity to jettison un-needed functions and consolidate added columns into the original table. 

* Addition of some new tests for the scheduler datastore.

* Removal of the external network boundary between scheduler and job server communication. The scheduler and worker manager threads use internal notifications to trigger work and status, and use the DB as a shared store to pass around the jobs and job status. This means we no longer have to persist job status notifications that have any integration credentials, and no longer have to have the route any of those calls.

* Since we have moved the scheduler data to job server, we need to re-build the graph - this is simply done by resetting the scheduler_sync status on the origin packages table.

* The builder-scheduler component is deleted, and removed from other artifacts that referenced it.

* The BUILDER_DEV and other docs, as well as the Makefile and Procfile, and several other related artifacts were updated to remove the scheduler references.

* Terraform was updated to remove the scheduler, and also to move the DataDog dashboard script to the job server.

---

tenor-69111944


thesentinels:

Thank you for contributing!

raskchanky:
I haven't reviewed this yet, but I wish that every PR description was as good as this.  10/10, would read again.

reset:
The PR message alone is worth a GitHub Emmy

builder/plan.sh
```diff
@@ -1,7 +1,7 @@
 pkg_origin="core"
 pkg_name="builder"
 pkg_type="composite"
-pkg_version="0.1.0"
+pkg_version="0.1.1"
```
reset: I don't think you need to bump this. We should just be using the same technique to determine version as the other components (`git rev-list master --count`)

components/builder-graph/src/error.rs
```diff
@@ -33,12 +32,10 @@ pub enum Error {
     DbTransaction(postgres::error::Error),
     HabitatCore(hab_core::Error),
     IO(io::Error),
-    PackageInsert(postgres::error::Error),
```
reset: I love to see unused errors getting cleaned up. I really wish the compiler found this for us!

components/builder-jobsrv/habitat/default.toml
```diff
@@ -1,4 +1,5 @@
 log_level = "info"
+log_path = "/tmp"
```
reset: Should this be configurable? For things like this I typically hardcode it in the `config.toml` as `{{pkg.svc_log_path}}` like we did above for `key_dir = "{{pkg.svc_files_path}}"`

components/builder-jobsrv/src/migrations/mod.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Chef Software Inc. and/or applicable contributors
+// Copyright (c) 2016-2017 Chef Software Inc. and/or applicable contributors
```
reset: The date the file was originally authored is the right pattern for the license. We've got a number of files that have a range because of a commit that was done in early 2017 but I've slowly been setting it back to the single date as I modify files. I probably should just go through the whole codebase and fix it.

components/builder-jobsrv/src/server/scheduler.rs
```diff
@@ -22,15 +22,16 @@ use hab_net::conn::RouteClient;
 use hab_net::socket::DEFAULT_CONTEXT;
 use zmq;
 
-use protocol::jobsrv::{self, Job, JobSpec};
-use protocol::originsrv::*;
-use protocol::scheduler as proto;
+use protocol::jobsrv;
+use protocol::originsrv;
 use data_store::DataStore;
-use error::{SrvResult, SrvError};
```
reset: Ugh, I knew not refactoring this in one of the services would eventually bite me :). It's cool to change this all to `Result` but as a pattern I had switched all but the JobSrv to `SrvResult` to clean up the need to `use std::result` and then address a regular result with `result::Result`. If you want to clean it up here by all means, if not we can get it eventually to look like the other servers

reset:
@chefsalim nothing catches my eye really here, it's a really large housecleaning task done well. If you could rebase to fix the conflicts from master let's get this merged in!

components/builder-jobsrv/habitat/default.toml
chefsalim: Yeah, it's configurable right now because we still haven't resolved the directory permissions issue for the datadog agent (and potentially logrotate) to have access to files that are within the svc hierarchy.  Once those issues are resolved we can go back to hard coding this.

components/builder-jobsrv/src/migrations/mod.rs
chefsalim: Sure, easy enough to fix.

chefsalim:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/290136622).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/290136622).

PR #3857, opened 2017-10-22
Some cleanup and additional debugging to jobsrv

chefsalim:
This change has some fixes and cleanup for the builder job server:
1. Change the ZMQ server socket for the worker manager and scheduler from DEALER to ROUTER. This is because a DEALER-DEALER can only work if there is a single client peer, but with the the worker-manager and scheduler becoming peers to each other, the DEALER-DEALER was causing deadlocks.

2. Fix a case where a stored procedure was being called with an incorrect name.

3. Fix a case where skipping a project was incorrectly erroring out the entire group status.

4. Add some additional debugging statements and clean up some left over cases where we were still referring to the scheduler database.

5. Clean up the main thread loops to ensure we don't inadvertently exit the main loop if an error is encountered.

tenor-205865182

### DIFF
--- a/components/builder-graph/README.md
+++ b/components/builder-graph/README.md
@@ -32,7 +32,7 @@ Example run:
 ```
 $ bldr-graph
 
-Connecting to builder_scheduler
+Connecting to builder_jobsrv
 Building graph... please wait.
 OK: 1224 nodes, 3537 edges (PT1.758762699S sec)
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

reset:
tenor-220579327

PR #3879, opened 2017-10-24
Queue builds instead of rejecting

chefsalim:
This change adds build job queueing functionality to the job server. If a build job for a project is already in progress, it queues the request for later dispatch after the in-progress job completes. The queue size is 1, so multiple queue requests will only a single additional build job into the queue.

tenor-267715188


reset:
@chefsalim looks great!

tenor-117244531

PR #3898, opened 2017-10-25
User-config Watcher

### LINKED ISSUE #2805: Configuration updates in a Kubernetes cluster
Kubernetes is responsible for delivering configuration updates to pods in the form of [`Secrets`](https://kubernetes.io/docs/concepts/configuration/secret/)/[`ConfigMaps`](https://kubernetes.io/docs/tasks/configure-pod-container/configmap/).

The Supervisor will watch for the file, which location you pass in as an argument to a flag when starting the service (for example: `... --config-file <dir>/<filename>`). If a change was made to the file, the supervisor will react in the same way as it does now with receiving a config update over gossip. We would possibly need to disable config update gossip part of the supervisor. 

As it needs to contain service group, version and the actual configuration data we are open for suggestions as to what format we want this file in. Maybe we can go with `toml` format as this is already common in Habitat. 

The solution does not depend on Kubernetes at all, it can be for use cases outside of the Kubernetes world.

We are open to any ideas and suggestions on the details of this implementation.

asymmetric:
The User-config Watcher uses the FileWatcher to keep track of the `user.toml` file.

When a service is started, the UCW starts a worker thread that watches the file for changes, removal or creation. When either of those events happen, the worker notifies the Watcher via a channel.

The Manager can signal the Worker to stop running by using another channel.

Closes #2805.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -146,12 +146,12 @@ pub struct Manager {
     launcher: LauncherCli,
     services: Arc<RwLock<Vec<Service>>>,
     updater: ServiceUpdater,
-    watcher: SpecWatcher,
+    peer_watcher: Option<PeerWatcher>,
+    spec_watcher: SpecWatcher,
     organization: Option<String>,
     self_updater: Option<SelfUpdater>,
     service_states: HashMap<PackageIdent, Timespec>,
     sys: Arc<Sys>,
-    peer_watcher: Option<PeerWatcher>,
 }
 
 impl Manager {
@@ -318,12 +318,12 @@ impl Manager {
             events_group: cfg.eventsrv_group,
             launcher: launcher,
             services: services,
-            watcher: SpecWatcher::run(&fs_cfg.specs_path)?,
+            peer_watcher: peer_watcher,
+            spec_watcher: SpecWatcher::run(&fs_cfg.specs_path)?,
             fs_cfg: Arc::new(fs_cfg),
             organization: cfg.organization,
             service_states: HashMap::new(),
             sys: Arc::new(sys),
-            peer_watcher: peer_watcher,
         })
     }
 
@@ -528,7 +528,7 @@ impl Manager {
     }
 
     pub fn run(&mut self) -> Result<()> {
-        self.start_initial_services_from_watcher()?;
+        self.start_initial_services_from_spec_watcher()?;
 
         outputln!(
             "Starting gossip-listener on {}",
@@ -564,7 +564,7 @@ impl Manager {
                 self.shutdown();
                 return Ok(());
             }
-            self.update_running_services_from_watcher()?;
+            self.update_running_services_from_spec_watcher()?;
             self.update_peers_from_watch_file()?;
             self.check_for_updated_packages();
             self.restart_elections();
@@ -682,7 +682,7 @@ impl Manager {
             active_services.push(service.spec_ident.clone());
         }
 
-        for loaded in self.watcher
+        for loaded in self.spec_watcher
             .specs_from_watch_path()
             .unwrap()
             .values()
@@ -790,7 +790,7 @@ impl Manager {
         // add services that are not active but are being watched for changes
         // These would include stopped persistent services or other
         // persistent services that failed to load
-        for down in self.watcher
+        for down in self.spec_watcher
             .specs_from_watch_path()
             .unwrap()
             .values()
@@ -891,8 +891,8 @@ impl Manager {
         release_process_lock(&self.fs_cfg);
     }
 
-    fn start_initial_services_from_watcher(&mut self) -> Result<()> {
-        for service_event in self.watcher.initial_events()? {
+    fn start_initial_services_from_spec_watcher(&mut self) -> Result<()> {
+        for service_event in self.spec_watcher.initial_events()? {
             match service_event {
                 SpecWatcherEvent::AddService(spec) => {
                     if spec.desired_state == DesiredState::Up {
@@ -906,7 +906,7 @@ impl Manager {
         Ok(())
     }
 
-    fn update_running_services_from_watcher(&mut self) -> Result<()> {
+    fn update_running_services_from_spec_watcher(&mut self) -> Result<()> {
         let mut active_specs = HashMap::new();
         for service in self.services
             .read()
@@ -917,7 +917,7 @@ impl Manager {
             active_specs.insert(spec.ident.name.clone(), spec);
         }
 
-        for service_event in self.watcher.new_events(active_specs)? {
+        for service_event in self.spec_watcher.new_events(active_specs)? {
             match service_event {
                 SpecWatcherEvent::AddService(spec) => {
                     if spec.desired_state == DesiredState::Up {
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -22,6 +22,7 @@ mod service_updater;
 mod spec_watcher;
 mod file_watcher;
 mod peer_watcher;
+mod user_config_watcher;
 mod sys;
 
 use std::collections::HashMap;
@@ -35,6 +36,8 @@ use std::thread;
 use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
+use std::mem;
+use std::ops::DerefMut;
 
 use butterfly;
 use butterfly::member::Member;
@@ -59,6 +62,7 @@ use self::service::{DesiredState, Pkg, ProcessState, StartStyle};
 use self::service_updater::ServiceUpdater;
 use self::spec_watcher::{SpecWatcher, SpecWatcherEvent};
 use self::peer_watcher::PeerWatcher;
+use self::user_config_watcher::UserConfigWatcher;
 use VERSION;
 use error::{Error, Result, SupError};
 use config::GossipListenAddr;
@@ -148,6 +152,7 @@ pub struct Manager {
     updater: ServiceUpdater,
     peer_watcher: Option<PeerWatcher>,
     spec_watcher: SpecWatcher,
+    user_config_watcher: UserConfigWatcher,
     organization: Option<String>,
     self_updater: Option<SelfUpdater>,
     service_states: HashMap<PackageIdent, Timespec>,
@@ -320,6 +325,7 @@ impl Manager {
             services: services,
             peer_watcher: peer_watcher,
             spec_watcher: SpecWatcher::run(&fs_cfg.specs_path)?,
+            user_config_watcher: UserConfigWatcher::new(),
             fs_cfg: Arc::new(fs_cfg),
             organization: cfg.organization,
             service_states: HashMap::new(),
@@ -520,6 +526,16 @@ impl Manager {
                 0,
             );
         }
+
+        if let Err(e) = self.user_config_watcher.add(&service) {
+            outputln!(
+                "Unable to start UserConfigWatcher for {}: {}",
+                service.spec_ident,
+                e
+            );
+            return;
+        }
+
         self.updater.add(&service);
         self.services
             .write()
@@ -566,6 +582,7 @@ impl Manager {
             }
             self.update_running_services_from_spec_watcher()?;
             self.update_peers_from_watch_file()?;
+            self.update_running_services_from_user_config_watcher();
             self.check_for_updated_packages();
             self.restart_elections();
             self.census_ring.update_from_rumors(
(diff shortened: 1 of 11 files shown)


thesentinels:

Thank you for contributing!

asymmetric:
Oh, seems like we have to fix a couple failing tests. On it.

krnowak:
Updated, but is based on https://github.com/habitat-sh/habitat/pull/3814, so this PR needs to wait until #3814 is merged.

components/sup/src/manager/user_config_watcher.rs
```diff
@@ -40,8 +41,8 @@ impl Serviceable for Service {
         &self.pkg.name
     }
 
-    fn path(&self) -> &Path {
-        &self.pkg.user_config_path
+    fn path(&self) -> UserConfigPath {
+        self.cfg.user_config_path.clone()
```
asymmetric: Why clone here? Wouldn't it be better to let the caller do it, if they need to?

components/sup/src/manager/service/mod.rs
```diff
@@ -407,29 +412,46 @@ impl Service {
...
+            }
         }
-        cfg_updated
+
+        let (cfg_changed, reload, reconfigure) = {
+            let ctx = self.render_context(census_ring);
+
+            if cfg_updated_from_rumors || census_ring.changed() {
+                let config_changed = cfg_updated_from_rumors || self.user_config_updated;
+                let needs_reload = self.compile_hooks(&ctx) || self.user_config_updated;
```
christophermaier: I think compiling the hooks is all you need here... if a change to a user config file resulted in the compiled output of the templates, then that should be the only thing you care about, right?

components/sup/src/manager/service/mod.rs
```diff
@@ -407,29 +412,46 @@ impl Service {
...
+                let needs_reconfigure = self.compile_configuration(&ctx);
+
+                (config_changed, needs_reload, needs_reconfigure)
+            } else if self.user_config_updated {
+                (true, true, self.compile_configuration(&ctx))
+            } else {
+                (false, self.needs_reload, self.needs_reconfiguration)
```
christophermaier: I believe that `needs_reload` and `needs_reconfiguration` are only set inside this function, so I think we can collapse this branch of the `if/else`, too.

components/sup/src/manager/user_config_watcher.rs
```diff
@@ -0,0 +1,394 @@
...
+use manager::service::Service;
+use manager::service::UserConfigPath;
+
+static LOGKEY: &'static str = "UCW";
+
+// This trait exists to ease the testing of functions that receive a Service. Creating Services
+// requires a lot of ceremony, so we work around this with this trait.
+pub trait Serviceable {
+    fn name(&self) -> &str;
+    fn path(&self) -> &UserConfigPath;
```
christophermaier: Might be good to call this `user_config_path` instead of just `path` for clarity.

components/sup/src/manager/mod.rs
```diff
@@ -879,20 +904,30 @@ impl Manager {
...
-        for mut service in services.drain(..) {
+        // The problem we're trying to work around here by adding this block is that `write`
+        // creates an immutable borrow on `self`, and `self.remove_service` needs `&mut self`.
+        // The solution is to introduce the block to drop the immutable borrow before the call to
+        // `self.remove_service`, and use `mem::swap` to move the services to a variable defined
+        // outside the block while we have the lock.
+        {
+            let mut services = self.services.write().expect("Services lock is poisoned!");
+            mem::swap(services.deref_mut(), &mut svcs);
+        }
```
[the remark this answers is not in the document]
asymmetric: `remove_service` took `&self` before, rather than `&mut self`.

Now it takes `&mut self` because it calls `self.user_config_watcher.remove(service)`, which stops the watcher thread for that service and removes it from the `states` HashMap. Therefore, we have a clash between the above, immutable borrow, and the mutable one below. Hence the block.

Previou

components/sup/src/manager/service/mod.rs
krnowak: I'm not sure I understand, but I'll bite - `needs_reload` and `needs_reconfiguration` are set to false also in `reload` and `reconfigure` functions, which are also gated on `needs_reload` and `needs_reconfiguration` being true. The `(false, self.needs_reload, self.needs_reconfiguration)` tuple here is to try to avoid changing value any of the members, so we don't accidentally set them from `true` to `false` because some config we don't care about has changed, but the reloading/reconfiguring didn't yet happen.

I haven't investigated if this can be the case, though.

components/sup/src/manager/mod.rs
```diff
@@ -947,21 +982,38 @@ impl Manager {
...
+                    outputln!(
+                        "Tried to remove service for {} but could not find it running, skipping",
+                        &spec.ident
+                    );
+                    return Ok(());
+                }
+            };
+
+            service = services.remove(services_idx);
+        }
```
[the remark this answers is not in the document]
krnowak: The explanation is basically the same as above. Obtaining the `services` variable involves taking a mutable ref on `self`. Also, `remove_service` requires mutable ref on `self`. So we needed to limit the scope of the `services` variable for mutable ref on `self` to go away before we could acquire another one by running the `remove_service` function.

components/sup/src/manager/service/mod.rs
asymmetric: Couldn't a reload be necessary if e.g. a port was changed in the configuration of a DB?

asymmetric:
@cm we've tried to address your comments, could you PTALA?

components/sup/src/manager/service/mod.rs
christophermaier: That should be taken care of when we generate a render context and compile the hooks. If the change to the user config resulted in different configs being rendered to disk, then that would be accounted for with the earlier code.

components/sup/src/manager/service/mod.rs
christophermaier: See reply above... I don't think additional if/else comparisons should be needed if new user config is accounted for when we generate our template rendering context.

components/sup/src/manager/service/mod.rs
asymmetric: Do you mean that the `reconfigure` hook would be run whenever the configuration in a `user.toml` has changed?
What if that hook isn't defined?

In general, I'm not sure I understand the connection between a change in the `user.toml` and hooks.
Couldn't there be an update to a `user.toml` file that doesn't produce any change in hooks?

christophermaier: `user.toml` is one layer of a service's configuration (https://github.com/habitat-sh/habitat/blob/e2a946fafab06f7f4cfe8b5b4e883b88d69f356b/components/sup/src/manager/service/config.rs#L79-L80)

This, in turn, is part of the `RenderContext` (https://github.com/habitat-sh/habitat/blob/e2a946fafab06f7f4cfe8b5b4e883b88d69f356b/components/sup/src/templating/context.rs#L56-L63) that is created for templating both hooks and configuration.

When [hooks](https://github.com/habitat-sh/habitat/blob/e2a946fafab06f7f4cfe8b5b4e883b88d69f356b/components/sup/src/manager/service/mod.rs#L617) and [configuration](https://github.com/habitat-sh/habitat/blob/e2a946fafab06f7f4cfe8b5b4e883b88d69f356b/components/sup/src/manager/service/mod.rs#L598) are rendered according to the new data present in the `RenderContext`, we return a boolean that reflects whether any _content_ of the templated files changed. If they do, then we reconfigure or reload as needed.

If data changed, but the output in the templated files remained the same, then there is no reason to restart. In the extreme case, imagine adding comments to your `user.toml` file. The toml file certainly changed, but it will have no effect on the hooks or configuration of the service, by definition.

asymmetric:
@christophermaier  I followed your advice and removed the redundant checks, which also results in a cleaner implementation. Let me know what you think.

components/sup/src/manager/user_config_watcher.rs
```diff
@@ -0,0 +1,395 @@
...
+                    return Ok(());
+                }
+            };
+            // Establish bi-directional communication with the worker by creating two channels.
+            // The sync_channel's buffer size is 1 because we want to use it as a boolean, i.e. we
+            // are not interested in the events themselves, but only whether at least one has
+            // happened.
+            let (events_tx, events_rx) = sync_channel(1);
+            let (running_tx, running_rx) = channel();
+            let (watching_tx, watching_rx) = sync_channel(1);
```
christophermaier: Out of curiosity, why did you choose to go with channels with a buffer size of 1 to act as a boolean, rather than using an `AtomicBool`?

components/sup/src/manager/mod.rs
```diff
@@ -855,6 +873,13 @@ impl Manager {
                 err
             );
         }
+
+        if let Err(_) = self.user_config_watcher.remove(service) {
+            debug!(
+                "Error stopping user-config watcher thread for service {}",
+                service
```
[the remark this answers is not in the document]
krnowak: Added.

components/sup/src/manager/mod.rs
```diff
@@ -947,21 +982,38 @@ impl Manager {
...
-            None => {
-                outputln!(
-                    "Tried to remove service for {} but could not find it running, skipping",
-                    &spec.ident
-                );
-                return Ok(());
+
+        for service in services.iter_mut() {
+            if self.user_config_watcher.have_events_for(service) {
+                outputln!("Reloading service {}", &service.spec_ident);
```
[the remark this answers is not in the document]
krnowak: Replaced the message with "User configuration of the service {} changed".

components/sup/src/manager/service/config.rs
```diff
@@ -43,6 +44,32 @@ static ENV_VAR_PREFIX: &'static str = "HAB";
...
+}
+
+impl UserConfigPath {
+    pub fn get_path(&self) -> &PathBuf {
+        match self {
+            &UserConfigPath::Recommended(ref p) |
+            &UserConfigPath::Deprecated(ref p) => p,
+        }
+    }
+}
```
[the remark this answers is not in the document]
krnowak: Done. Ampersands strike back, though. ;)

`let user = Self::load_user(self.user_config_path.get_path())?;`

is now

`let user = Self::load_user(&self.user_config_path)?;`

components/sup/src/manager/user_config_watcher.rs
krnowak: Uh, I guess that it has some Golang background, but I'll leave this question to @asymmetric - he wrote that code. ;)

krnowak:
Review issues addressed, rebased to current master and resolved the conflict.

components/sup/src/manager/service/config.rs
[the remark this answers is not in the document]
krnowak: I was wondering if I also should implement the `Deref` trait, so we can get the `&Path` too for scenarios like `user_config_path().join("foo")`. Or is doing `user_config_path().as_ref().join("foo")` fine enough? For now I went with the second option (in user config watcher tests).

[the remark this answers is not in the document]
christophermaier: I think using `as_ref()` for now is fine; we can revisit later if necessary.

components/sup/src/manager/user_config_watcher.rs
christophermaier: @asymmetric any input on this one? ☝️

christophermaier:
@krnowak In the future, I think it would be easier to review if you could just add new commits when addressing feedback, instead of folding the changes into the commits that are already there. Otherwise, a reviewer almost has to go over the entire PR again to look for what's different. When everything is good, you can go back and squash things as needed. 

Other than that, I just had the remaining question about channel vs. AtomicBool... after that, I think we can get this merged in.

components/sup/src/manager/user_config_watcher.rs
asymmetric: Sorry for the delay, I was on holiday. Mostly because I found easier to reason about this problem using message passing, rather than sharing data.

If I remember correctly, there were also issues around the fact that the Arcs are fields of the `WorkerState` struct, and borrowing one field for reading locked the whole `struct`, making it hard to modify the other field in the same context.

asymmetric:
@cm addressed your comment, let me know if there's anything else.

components/sup/src/manager/user_config_watcher.rs
christophermaier: I think there are techniques that we could use to deal with the borrowing, and ultimately, I think something like an `AtomicBool` may be simpler and easier to reason about.

That being said, we can probably tackle that later. I don't think we need to hold this feature up any more.

christophermaier:
@asymmetric @krnowak If you all could rebase this on current master, we can go ahead and get it merged in.

asymmetric:
@cm done!

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/329953283).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/329953283).

PR #3908, opened 2017-10-26
Fix the ability to disable non-core builds

reset:
This commit ensures that the `non_core_builds_enabled` config field
for the builder-api disables the ability to build and will hide the
build button when disabled


thesentinels:

Thank you for contributing!

PR #3910, opened 2017-10-26
Fix a few broken links to Slack.

frewsxcv:
Looks like there's and invalid certificate so stick to HTTP. And there's
a couple instances that need protocols.

Hi from Rust Belt Rust! 👋


cnunciato:
Thanks for the contribution, @frewsxcv! 

Looks like we just need you to sign your commit with a Developer Certificate of Origin. Here's a bit of info about how to do that for us: https://github.com/habitat-sh/habitat/blob/master/CONTRIBUTING.md#signing-your-commits

frewsxcv:
sorry for the delay, should be signed-off now!

PR #3926, opened 2017-10-27
Start studio process in builder-worker with `airlock` 

reset:
Studios will be started in a more difficult to escape jail by using
the airlock program. The studio is also now given it's own user and
group for the airlock.

Companion to #3925


thesentinels:

Thank you for contributing!

PR #3931, opened 2017-10-27
Set appropriate default log_path for jobsrv in Habitat plan

reset:
Set log_path to the service's var directory which the jobsrv user will have access to read and write from


thesentinels:

Thank you for contributing!

chefsalim:
This will break our dashboard since datadog agent will not have access to it :(

reset:
@chefsalim datadog should be running as the `hab` user which would have access, yeah?

PR #3932, opened 2017-10-27
Use stable packages except for when installing builder specifics

reset:
We want to use the stable package channel except for when we're
installing the builder packages themselves.

Additionally this installs a required user for the builder-worker
to run


thesentinels:

Thank you for contributing!

PR #3940, opened 2017-10-27
Use studio user's home dir for storing private keys in worker

reset:
Prior the worker would download keys into the root user's directory
and those keys would be imported into the studio. By changing which
user is running the studio we need to install the keys into a more
appropriate location

### DIFF
--- a/components/builder-worker/src/runner/mod.rs
+++ b/components/builder-worker/src/runner/mod.rs
@@ -31,7 +31,7 @@ use bldr_core::job::Job;
 use bldr_core::logger::Logger;
 use chrono::UTC;
 use depot_client;
-use hab_core::crypto;
+use hab_core::fs::CACHE_KEY_PATH;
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/builder-worker/src/runner/mod.rs
```diff
@@ -208,7 +207,7 @@ impl Runner {
                 self.depot_cli.fetch_origin_secret_key(
                     self.job().origin(),
                     &self.config.auth_token,
-                    &crypto::default_cache_key_path(None),
+                    &*studio::STUDIO_HOME.lock().unwrap(),
```
fnichol: maybe something like? `&*studio::STUDIO_HOME.lock().unwrap().join(format!(".{}", …::CACHE_KEY_PATH))`

fnichol:
tenor-208179446

PR #3944, opened 2017-10-27
Keep environment when executing Studio from within worker

reset:
We should keep the environment when running the Studio from the
worker so we are able to keep the important bits setup by Habitat's
packaging. The environment will be clean once the program is being
built in the Studio, so there is no specific reason to clear the env
in the first place


thesentinels:

Thank you for contributing!

PR #3948, opened 2017-10-28
Builder Worker airlock fixes

reset:
* Fix the chance that the src directory will fall outside of the airlock by first changing directory to the src tree before entering the studio
* Studio is locked to the version of backline it was built against at runtime. This change will allow us to use whatever version of backline is located on the machine instead when starting a studio from withint he Worker


thesentinels:

Thank you for contributing!

PR #3955, opened 2017-10-30
Remove raw conversion of io::Error to a non-descript Error in Worker

reset:
This removes the catch all From<io::Error> trait implementation for
errors in the Worker. This catch all masks an actual error and makes
it difficult to determine what went wrong


thesentinels:

Thank you for contributing!

fnichol:
That's much nicer--should help to track down those IO errors with context.

PR #3972, opened 2017-11-01
[plan-build] Extract tar with ownership of current user.

fnichol:
This change will ignore user and group metadata in source tarballs when
extracting them in `do_unpack()` or when using `unpack_file()`. Trying
to set unknown uid and gid numbers to files can cause issues and
shouldn't be required when building the software.


thesentinels:

Thank you for contributing!

PR #4001, opened 2017-11-03
Ignore non UTF-8 files while looking for key pairs

### LINKED ISSUE #3833: `hab studio enter` fails in the presence of .DS_store files
Hello - thanks for reporting an issue with Habitat.

In order to help us troubleshoot the issue, please be sure to include the following information if applicable:

- [x] The OS: macOS High Sierra
````
$ hab origin key export --type secret <origin>
✗✗✗
✗✗✗ stream did not contain valid UTF-8
✗✗✗

$ rm ~/.hab/cache/keys/.DS_store

$ hab origin key export --type secret <origin>
SIG-SEC-1
<origin>-<datestamp>
````

### PR DESCRIPTION
**Pull request by rsertelon**
I chose to go with ignoring hidden files to fix #3833

### DIFF
--- a/components/core/src/crypto/keys/mod.rs
+++ b/components/core/src/crypto/keys/mod.rs
@@ -290,7 +290,11 @@ where
         let file = File::open(dir_entry.path())?;
         let mut reader = BufReader::new(file);
         let mut buf = String::new();
-        reader.read_line(&mut buf)?;
+
+        if let Err(e) = reader.read_line(&mut buf) {
+            debug!("Invalid content: {}", e);
+            continue;
+        }
 
         if !buf.starts_with(&key_type.to_string().to_uppercase()) {
             debug!(

### CODE REVIEW

**Comment by thesentinels** [comment]

Thank you for contributing!

**Comment by nellshamrell** [comment]
I'm trying to think if there would be any circumstances we would want to include hidden files...

**Comment by rsertelon** [comment]
Yeah, I thought of including key files only instead of excluding hidden ones; but I don't know if looking for a pattern or for `*.pub` and `*.sig.key` would be enough...

**Comment by rsertelon** [comment]
Oh I see there's a lint.sh script executed by travis. will fix if given solution is enough for you.

File: components/core/src/crypto/keys/mod.rs [hunk 1]
```diff
@@ -273,6 +273,16 @@ where
             }
         };
 
+        // Ignore hidden files such as .DS_store
+        if dir_entry
+            .file_name()
+            .to_str()
+            .map(|s| s.starts_with("."))
```
christophermaier: Given that we have all the key name formatting information already expressed in code, I think that leveraging that code here makes more sense than carving out exceptions. There can be many other exceptions we might need to add later, for instance. 

Ensuring that there is one place where "what it means to be a key file" is represented will make the code easier to reason about and modify going forward.

rsertelon:
@christophermaier did a lot better I think. Not sure this is the rust way, but it works ;) A little help is always welcome, thanks again!

components/core/src/crypto/keys/mod.rs
```diff
@@ -287,6 +291,20 @@ where
...
+        };
+
+        debug!("checking file: {}", &filename);
+        if !check_filename(keyname, filename, &mut candidates, pair_type) {
+            continue;
+        }
+
         let file = File::open(dir_entry.path())?;
         let mut reader = BufReader::new(file);
         let mut buf = String::new();
```
christophermaier: I can't comment on the line I want since it's not in the diff... Line 311 in this patch:

```rust
reader.read_line(&mut buf)?;
```

This is where the error is coming from, because `read_line` [can fail when the content is not UTF-8](https://doc.rust-lang.org/1.21.0/std/io/trait.BufRead.html#method.read_line). By explicitly handling this error (instead of passing it out using the `?` operator), we'll fix this immediate issue due to MacOS `.DS_Store` files, but will also handle any other file that happens to have non-UTF-8 data.

Moving the filename checking to before the read call, as done in this PR, is a welcome change (it allows you to catch things earlier on), but handling this error is the core of the issue, and will also make Habitat more robust.

rsertelon: How could I write this exactly? With a match that would lead to a continue if errored?

christophermaier: @rsertelon Yes, exactly! You're in a loop populating the vector of candidates right now, but this error is just for one of them. The code currently bails out of everything if one file is bad, which is why this is a bug in the first place 😄 . You just want to say "yup, that's a bad file", and go on to process the rest of your files, and ultimately return a list of candidates.

Once we get into that candidate processing loop, we should be handling all possible errors.

rsertelon: Fixed the error handling, but not sure about the implementation (I don't like returning the contents of `Ok()`)

components/core/src/crypto/keys/mod.rs
```diff
@@ -287,10 +291,32 @@ where
...
-        reader.read_line(&mut buf)?;
+        let read_result = reader.read_line(&mut buf);
+        
+        match read_result {
```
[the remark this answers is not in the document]
rsertelon: yes, but the result is never actually used isn't it? Though about replacing `read_result` by `reader.read_line(&mut buf)` but felt weird :) Also, maybe I should just use `is_err()` on `read_result` in this case?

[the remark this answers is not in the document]
christophermaier: Argh, yeah, you're right... I was focusing on the wrong thing. Something like this should do what you want:

```rust
if let Err(e) = reader.read_line(&mut buf) {
    // error handling and continue call
}
```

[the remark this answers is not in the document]
rsertelon: This one is super cool! Implementing... :construction_worker_man:

rsertelon:
@christophermaier done :)

christophermaier:
@rsertelon Nice!

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/302707146).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/302707146).

PR #4005, opened 2017-11-04
Uses tabwriter to format hab svc status output

### LINKED ISSUE #3859: want formatted output for `hab svc status`
The current view is not pretty-printed (aligned into columns), which means that for more than 3 or so listed services the output becomes increasingly hard to read.

rsertelon:
> Should close #3859

This pull request leverages crate tabwriter to output `hab svc status` with nice columns:

```
rsertelon@rsertelon-desktop:~/sources/habitat (formatted-output-hab-svc-status=)$ hab svc status
core/redis/3.2.4/20170514150022, state:up, time:PT1366.889919553S, pid:9813, group:redis.default, style:transient
rsertelon/monit/5.24.0/20171022142238, state:up, time:PT1060.891684845S, pid:10744, group:monit.default, style:transient
rsertelon/mosquitto/1.4.14/20171027184938, state:up, time:PT930.893769325S, pid:11315, group:mosquitto.default, style:transient
```
becomes
```
rsertelon@rsertelon-desktop:~/sources/habitat (formatted-output-hab-svc-status *)$ target/debug/hab-sup status
core/redis/3.2.4/20170514150022            standalone  state:up  time:PT597.121500061S  pid:9813   group:redis.default      style:transient
rsertelon/monit/5.24.0/20171022142238      standalone  state:up  time:PT291.124764752S  pid:10744  group:monit.default      style:transient
rsertelon/mosquitto/1.4.14/20171027184938  standalone  state:up  time:PT161.128453720S  pid:11315  group:mosquitto.default  style:transient
```

The code has been modified in place for this particular command. It might be interesting to try to use tabwriter with the `outputln!` macro (or new equivalent one) so that we can use its column output almost anywhere "for free". If you think it might be worth it, I'm keen on opening a new issue to track this.

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -26,6 +26,7 @@ extern crate libc;
 extern crate clap;
 extern crate time;
 extern crate url;
+extern crate tabwriter;
 
 use std::collections::{HashMap, HashSet};
 use std::collections::hash_map::Entry;
@@ -52,6 +53,7 @@ use hcore::package::metadata::{BindMapping, PackageType};
 use hcore::service::{ApplicationEnvironment, ServiceGroup};
 use hcore::url::{bldr_url_from_env, default_bldr_url};
 use launcher_client::{LauncherCli, ERR_NO_RETRY_EXCODE, OK_NO_RETRY_EXCODE};
+use tabwriter::TabWriter;
 use url::Url;
 
 use sup::VERSION;
@@ -61,7 +63,7 @@ use sup::feat;
 use sup::command;
 use sup::http_gateway;
 use sup::http_gateway::ListenAddr;
-use sup::manager::{Manager, ManagerConfig};
+use sup::manager::{Manager, ManagerConfig, ServiceStatus};
 use sup::manager::service::{DesiredState, ServiceBind, Topology, UpdateStrategy};
 use sup::manager::service::{CompositeSpec, ServiceSpec, StartStyle};
 use sup::util;
@@ -851,24 +853,38 @@ fn sub_status(m: &ArgMatches) -> Result<()> {
                     process::exit(2);
                 }
             };
-
-            for spec in specs {
-                let status = Manager::service_status(&cfg, &spec.ident)?;
-                outputln!("{}", status);
-            }
+            print_statuses(specs.iter()
+                                .filter_map(|spec| Manager::service_status(&cfg, &spec.ident).ok())
+                                .collect::<Vec<ServiceStatus>>())?;
         }
         None => {
-            let statuses = Manager::status(&cfg)?;
-            if statuses.is_empty() {
-                println!("No services loaded.");
-                return Ok(());
-            }
-            for status in statuses {
-                println!("{}", status);
-            }
+            print_statuses(Manager::status(&cfg)?)?;
         }
     }
     Ok(())
+}
+
+fn print_statuses(statuses: Vec<ServiceStatus>) -> Result<()> {
+    if statuses.is_empty() {
+        println!("No services loaded.");
+        return Ok(());
+    }
+    let titles = vec!("package", "type", "state", "uptime (s)", "pid", "group", "style");
+    let mut tw = TabWriter::new(io::stdout());
+    write!(tw, "{}\n", titles.join("\t"));
+    for status in statuses {
+        write!(tw, "{}\t{}\t{}\t{}\t{}\t{}\t{}\n",
+            status.pkg.ident,
+            status.composite.unwrap_or("standalone".to_string()),
+            status.process.state,
+            status.process.elapsed.num_seconds(),
+            status.process.pid.map(|p| p.to_string()).unwrap_or("<none>".to_string()),
+            status.service_group,
+            status.start_style
+        );
+    }
+    tw.flush()?;
+    return Ok(());
 }
 
 fn sub_stop(m: &ArgMatches) -> Result<()> {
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
@rsertelon Sounds good; best of luck with the new job!

rsertelon:
@christophermaier here are my thoughts in response to yours :)

* Completely agree for column names, I didn't want to break too much at a time but clearly this would be a super plus!
* Oh, good catch, didn't see this one.
* Yep, definitely.

I'll try to organize it a bit more like you suggested, again, thanks for the advice and help!

rsertelon:
Output is a bit better now:

```
package                               type        state  uptime  pid     group              style
core/mosquitto/1.4.14/20171117213807  standalone  up     4471s   20118   mosquitto.default  transient
core/redis/3.2.4/20170514150022       standalone  down   4421s   <none>  redis.default      persistent
core/rabbitmq/3.6.12/20171025231625   standalone  down   4420s   <none>  rabbitmq.default   persistent
```

However, I'm pretty sure rust code can be improved.

I've added a `TableWriter` struct which wraps the `tabwriter::TabWriter` from `tabwriter` crate to provide an API closer to what we want: have a title line and then content lines.

Points that I don't like as is (but I'm still learning rust, and need guidance):
* No error management, I feel like it'll blow up in our faces somehow ;) (especially the `panic!()`)
* method `TableWriter::add_line(cols: Vec<String>)` cannot be made more generic (like `Vec<ToString>`) due to compilation requirements, do its usage feels too cumbersome (lots of `to_string()`)
* finally, the `print_statuses()` method is used at two places, in cases a bit different and the flow of the program around this method feels a bit odd (preceding `process::exit(2)` allows us to reuse the method, otherwise, it would print two different messages if not existing service).

components/sup/src/main.rs
```diff
@@ -851,26 +852,39 @@ fn sub_status(m: &ArgMatches) -> Result<()> {
...
     Ok(())
 }
 
+fn print_statuses(statuses: Vec<ServiceStatus>) -> Result<()> {
+    if statuses.is_empty() {
+        println!("No services loaded.");
+        return Ok(());
+    }
+    let mut tw = cli::TableWriter::new(7);
+    tw.set_titles(vec!("package", "type", "state", "uptime", "pid", "group", "style"));
```
christophermaier: I'd declare the header vector first, then use its length to create the `TableWriter`, instead of having the magic number `7` in the code.

components/sup/src/main.rs
```diff
@@ -851,26 +852,39 @@ fn sub_status(m: &ArgMatches) -> Result<()> {
...
+    for status in statuses {
+        tw.add_line(vec!(
+            status.pkg.ident.to_string(),
+            status.composite.unwrap_or("standalone".to_string()),
+            status.process.state.to_string(),
+            status.process.elapsed.num_seconds().to_string() + "s",
```
christophermaier: It may be easier for any downstream Unix pipeline commands to parse the data if we exclude the "s" suffix here. 

We might want to incorporate that into the column header instead. What are the separator characters here, tabs or spaces? I wouldn't want a header like `"uptime (sec)"` to throw off the column alignment due to the space.

components/sup/src/util/cli.rs
```diff
@@ -0,0 +1,55 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern crate tabwriter;
+
+use std::io::{self, Write, Stdout};
+use self::tabwriter::TabWriter;
+
+pub struct TableWriter {
```
christophermaier: I'm not sure that a `TableWriter` struct necessarily buys us much here, and isn't quite general purpose (for example, nothing requires that you call `set_titles` first, and nothing prevents you from calling it after you've already added some content lines).  The `panic!` call is also problematic, as you pointed out.

It may just be easier to remove this struct and just assemble the table directly in `print_statuses`.

components/sup/src/main.rs
rsertelon: Yep this feels better indeed.

components/sup/src/main.rs
rsertelon: Ok I can move it. Makes sense. The separator for columns is `\t` so we're good.

components/sup/src/util/cli.rs
rsertelon: The idea here was to finish the job later actually ;) needed a first look on the overall changes. I effectively want to forbid titles after lines. And to create he basis of a utility struct that could be used in other places of the app.

christophermaier: I think the simpler way is good for now. We can employ the "Rule of 3" for further abstraction: when we find ourselves implementing the same pattern the third time, we should have a better idea what the real abstraction should be.

rsertelon: Good for me! I'll change this then ;)

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/305532225).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/305532225).

Issue #3978, opened 2017-11-01
Allow packages to define runtime environment variables


christophermaier:
In some cases, it can be advantageous to allow packages to define environment variables that can be used to construct a runtime environment for a service. One example of this could be a dependency that creates Python libraries which need to be in the `PYTHONPATH` of a consuming service. By allowing packages to set this information, we can automatically compose a suitable runtime environment on behalf of a user (similar to what we do to generate a `PATH` that points to all the binaries a service depends on), which will simplify service hooks and unlock new usage patterns. 

This can be seen as an extension / refinement of ideas presented in:
  - https://forums.habitat.sh/t/allow-env-vars-to-be-passed-from-plan-to-runtime/264/6
  - https://github.com/habitat-sh/habitat/issues/949
  - https://github.com/habitat-sh/habitat/issues/1579
  - https://github.com/habitat-sh/habitat/pull/1830

cc: @georgemarshall @bixu

georgemarshall:
Also https://github.com/habitat-sh/habitat/pull/3171

christophermaier:
Closed by #4067

PR #4067, opened 2017-11-15
Specify runtime environments for packages

### LINKED ISSUE #949: Automatic Environment Variable Dependency Propagation
I had to set the Git config environment variable `GIT_SSL_CAINFO` to use `go get` in a plan that depends on `core/go` (cacerts), would it be possible to automatically inject this?

e.g.:

``` shell
do_prepare() {
    export GIT_SSL_CAINFO=$(pkg_path_for cacerts)/ssl/certs/cacert.pem
    build_line "Setting GIT_SSL_CAINFO=$GIT_SSL_CAINFO"

    # XXX(fujin): Figure out GOPATH convention for Habitat
    export GOPATH=${pkg_prefix}
    go get -u github.com/FiloSottile/gvt
}
```

### LINKED ISSUE #1579: RFC: a mechanism to declare runtime environment variables at build time

christophermaier:
This PR allows plan authors to specify additional runtime and buildtime environment variables for their plans, using the new `set_runtime_env`, `push_runtime_env`, `set_buildtime_env`, and `push_buildtime_env` helper functions inside the new `do_setup_environment` plan callback function.

Runtime environments are layered together from direct dependencies, allowing (for example) the automatic generation of `PYTHONPATH`, `GEM_PATH`, etc., based on a package's dependencies. This generated runtime environment (generated at buildtime) is used by the Supervisor when running services, so authors will have less manual environment management to do in their hooks. Provenance of environment values is tracked in the new `RUNTIME_ENVIRONMENT_PROVENANCE` and `BUILDTIME_ENVIRONMENT_PROVENANCE` metadata files, which can aid in debugging any environment-related issues.

Buildtime environments function in a similar way, but are composed of the runtime environments of any build dependencies. They are not incorporated into the runtime environment seen by the Supervisor.

Read individual commit comments for further details.

Closes #949
Closes #1579
Closes #1830

cc: @georgemarshall @robbkidd @bixu

### DIFF
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -109,6 +109,8 @@ pub enum Error {
     InvalidServiceGroup(String),
     /// Occurs when an origin is in an invalid format
     InvalidOrigin(String),
+    /// Occurs when an OsString path cannot be converted to a String
+    InvalidPathString(ffi::OsString),
     /// Occurs when making lower level IO calls.
     IO(io::Error),
     /// Errors when joining paths :)
@@ -130,8 +132,6 @@ pub enum Error {
     NoOutboundAddr,
     /// Occurs when a call to OpenDesktopW fails
     OpenDesktopFailed(String),
-    /// Occurs when dealing an OsString cannot be converted to a String
-    OsString(ffi::OsString),
     /// Occurs when a suitable installed package cannot be found.
     PackageNotFound(package::PackageIdent),
     /// When an error occurs parsing an integer.
@@ -328,6 +328,9 @@ impl fmt::Display for Error {
                     origin
                 )
             }
+            Error::InvalidPathString(ref s) => {
+                format!("Could not generate String from path: {:?}", s)
+            }
             Error::IO(ref err) => format!("{}", err),
             Error::JoinPathsError(ref err) => format!("{}", err),
             Error::LogonTypeNotGranted => {
@@ -347,7 +350,6 @@ impl fmt::Display for Error {
             Error::MetaFileIO(ref e) => format!("IO error while accessing MetaFile: {:?}", e),
             Error::NoOutboundAddr => format!("Failed to discover this hosts outbound IP address"),
             Error::OpenDesktopFailed(ref e) => format!("{}", e),
-            Error::OsString(ref s) => format!("Failed to convert to String: {:?}", s),
             Error::PackageNotFound(ref pkg) => {
                 if pkg.fully_qualified() {
                     format!("Cannot find package: {}", pkg)
@@ -476,6 +478,7 @@ impl error::Error for Error {
                 "Origins must begin with a lowercase letter or number.  \
                     Allowed characters include a - z, 0 - 9, _, and -. No more than 255 characters."
             }
+            Error::InvalidPathString(_) => "Failed to convert an OsString Path to a String",
             Error::IO(ref err) => err.description(),
             Error::JoinPathsError(ref err) => err.description(),
             Error::LogonTypeNotGranted => {
@@ -490,7 +493,6 @@ impl error::Error for Error {
             Error::MetaFileIO(_) => "MetaFile could not be read or written to",
             Error::NoOutboundAddr => "Failed to discover the outbound IP address",
             Error::OpenDesktopFailed(_) => "OpenDesktopW failed",
-            Error::OsString(_) => "Failed to convert an OsString to a String",
             Error::PackageNotFound(_) => "Cannot find a package",
             Error::ParseIntError(_) => "Failed to parse an integer from a string!",
             Error::PermissionFailed(_) => "Failed to set permissions",
@@ -514,12 +516,6 @@ impl error::Error for Error {
 impl From<env::JoinPathsError> for Error {
     fn from(err: env::JoinPathsError) -> Self {
         Error::JoinPathsError(err)
-    }
-}
-
-impl From<ffi::OsString> for Error {
-    fn from(s: ffi::OsString) -> Self {
-        Error::OsString(s)
     }
 }
 
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
 use std::error;
+use std::ffi;
 use std::io;
 use std::fmt;
 use std::num;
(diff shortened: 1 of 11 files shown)


thesentinels:

Thank you for contributing!

components/plan-build/bin/environment.sh
```diff
@@ -0,0 +1,467 @@
...
+    [GOPATH]=":"
+
+    # Java
+    [CLASSPATH]=";"
+
+    # NodeJS
+    [NODE_PATH]=":"
+
+    # Python
+    [PYTHONHOME]=":"
```
georgemarshall: `PYTHONHOME` should be removed since it is not joinable

robbkidd:
yessssss

rsertelon:
@christophermaier is #3978 still relevant then?

bixu:
@christophermaier, do we need `LD_LIBRARY_PATH` as an aggregate variable?

christophermaier:
@bixu Hrmm, that's an excellent question. I wonder if we should instead treat it like we do `LD_RUN_PATH` and `LD_FLAGS` and handle it via `pkg_lib_dirs`.

@fnichol Do you have any thoughts on that?

components/plan-build/bin/environment.sh
```diff
@@ -0,0 +1,466 @@
...
+declare -A __buildtime_environment_provenance
+
+# Layer all direct dependencies' environment together.
+#
+# Priority is last-one-wins, based on the order in which you define
+# the dependencies in your plan file.
+#
+# (I'm sure there are many other common variables we could add here;
+# PRs welcome!)
+declare -A -g __well_known_aggregate_env_vars=(
```
reset: Maybe these belong in scaffolding instead of the core of plan build. Probably safe for a first pass to leave them here for now

components/core/src/error.rs
```diff
@@ -501,6 +511,18 @@ impl error::Error for Error {
     }
 }
 
+impl From<env::JoinPathsError> for Error {
+    fn from(err: env::JoinPathsError) -> Self {
+        Error::JoinPathsError(err)
+    }
+}
+
+impl From<ffi::OsString> for Error {
```
reset: I'd avoid general conversions like this. They're pretty hard to track down as more and more people leverage them. We see this with the `io::Error` From implementation - it lacks the context of *where* and *why* this error occurred.

Instead I'd recommend an error such as "Error::BadEnvironmentString" or something related to what was being read at the time of failure.

reset:
@christophermaier some minor feedback inline but overall this looks perfect! Some samples/docs would be super nice additions before we get this merged in

robbkidd:
The scaffolds could maybe use this in the future to do their `scaffolding_env` settings!

HT154:
+1 for including `LD_LIBRARY_PATH`. It covers the case of `ctypes`-based python modules where libraries must be located and loaded at runtime. I'd like to avoid manually setting it in my run hook.

components/plan-build/bin/environment.sh
christophermaier: @reset interesting idea... could give us more flexibility in pushing updates independently of the rest of the build tooling, if nothing else.

christophermaier:
@bixu @HT154 Thinking a bit more... Including `LD_LIBRARY_PATH` could undermine some of the isolation and atomicity guarantees that Habitat provides at runtime. Would it be possible to formulate your usecase in another way, using features that Habitat already provides? 

Opening a new issue around this with further details about what you are trying to achieve would be useful, as I don't believe that we're going to address this in the current PR.

bixu:
@christophermaier, yes I can understand what you mean about breaking isolation guarantees. My feeling at the moment is that I (personally at least) don't yet see strong patterns for cases that have needed us to set `LD_LIBRARY_PATH`, so I think it's fine to wait for those to emerge. Setting `LD_LIBRARY_PATH` has always feel a bit hack-like to me. Perhaps when better patterns emerge we can find a way to guide users towards them.

christophermaier:
Going to work on a blog post introducing the feature; in the mean time, folks riding the bleeding edge can consult the commit message of https://github.com/habitat-sh/habitat/pull/4067/commits/13344a679155e5210dd58ecb9d94654f5ae676d3 for preliminary documentation.

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/309056113).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/309056113).

PR #4089, opened 2017-11-20
Fix paths inconsistencies errors

krnowak:
There were two bugs:
- repeated file disappeared events for the same file
- repeated unwatch actions for the same file

This happened when deleting a directory with watched files quickly, so
we were getting a batch of RemoveNotice events, which after 2 seconds
would be followed by a batch of Remove events.

The handling of RemoveNotice event involves traversing the chain
starting from the removed file until the end and generating the drop
watch action for every traversed item. This happens in the drop_common
function. So if we quickly removed a file and its parent directory
then drop_common would generate a drop watch event for the file and
then two drop watch events for the parent directory and the file
again.

Also, every time the function checks the real_file member to check if
we should generate the file disappearance action. So here is the
similar problem like with drop action generation above.

Since the chain is like a linked list, unwatching some item in the
middle of the chain will certainly make the real file to disappear. So
set the real_file to None every time an item is going to be
unwatched. Also, set the next member of the previous item in the chain
to None for every unwatched item, so we don't generate repeated drop
watch actions for the following NoticeRemove events.

### DIFF
--- a/components/sup/src/manager/file_watcher.rs
+++ b/components/sup/src/manager/file_watcher.rs
@@ -991,33 +991,45 @@ impl Paths {
     }
 
     fn drop_watch(&mut self, path: &PathBuf) -> Option<PathBuf> {
-        match self.paths.remove(path) {
-            Some(watched_file) => {
-                let common = watched_file.steal_common();
-                let dir_path = common.dir_file_name.directory;
-                let unwatch_directory = match self.get_mut_directory(&dir_path) {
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/305433825).

PR #2951, opened 2017-08-16
Corrects Binlink in Studio, adds HAB_BINLINK_DIR envvar

eeyun:
Signed-off-by: Ian Henry <ihenry@chef.io>

PR #4119, opened 2017-11-25
Binlink to /usr/local/bin by default

### LINKED ISSUE #1920: `hab pkg binlink` should default to /usr/local/bin
By default `hab pkg binlink` links into /bin. According to https://linux.die.net/man/7/hier /bin is for "executable programs which are needed in single user mode and to bring the system up or repair it.", /usr/bin is for programs installed by the system package manage, and /usr/local/bin is for locally installed programs.

If you binlink into /bin or /usr/bin you could have unexpected consequences resulting from operations your operating system does.

/usr/local/bin should be the default.

Perhaps also, it should test which of these directories exist, so that if you are on a more limited system where /usr/local does not exist, it would show /usr/bin as the default.

rsertelon:
If it is not in path, then try /usr/bin and if not, use /bin in last resort

Should close #1920

### DIFF
--- a/components/core/src/binlink.rs
+++ b/components/core/src/binlink.rs
@@ -14,8 +14,10 @@
 
 use env;
 
-/// Default Binlink Dir
-pub const DEFAULT_BINLINK_DIR: &'static str = "/bin";
+/// Default Binlink Dirs
+pub const DEFAULT_BINLINK_DIR_1: &'static str = "/usr/local/bin";
+pub const DEFAULT_BINLINK_DIR_2: &'static str = "/usr/bin";
+pub const DEFAULT_BINLINK_DIR_3: &'static str = "/bin";
 
 /// Binlink Dir Environment variable
 pub const BINLINK_DIR_ENVVAR: &'static str = "HAB_BINLINK_DIR";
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
@rsertelon This looks good to me, and tracks with the discussion in #1920. 

I'd like to run this by @fnichol before merging, though, since a lot of that discussion is old, and he's done a lot more work / thinking on binlinking since then.

@fnichol What do you think?

rsertelon:
@fnichol ? :)

fnichol:
I'm way late to this, so first off really sorry.

I'm not a big fan of the multiple-defaults approach because it can make it harder to reason about where it should go by…default (in a static, knowable way). From a support perspective, now we need to know about what's on your file system (i.e. which directories are at play) in addition to what's on `PATH`. Also, from a scripting perspective it's also harder to know if I binlink something just where it might go (I might have to "refind" this later with fallback if/elsif/else style checks).

The idea behind having it in `/bin` is partly that it's aggressive--we have high confidence that this will work and there are no more than one directory which needs to exist. In this way we are bucking standards such as [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) in the Linux world, but Habitat isn't Linux-specific.

To try and mitigate which default in a slightly more flexible way, the binlinking logic will honor the value of the `HAB_BINLINK_DIR` environment variable to override the default which we introduced in #2951 (note that an explicit flag will beat both the default and environment variable). We make use of this approach in a default Studio type so that binlinking gets on a user's PATH which is purpose-build set to `/hab/bin` (https://github.com/habitat-sh/habitat/blob/2d44cccdc63e2ae704d964ca21f41fe405b9fc93/components/studio/libexec/hab-studio-type-default.sh#L103).

I realize that this might cut across the grain and the implementation here is great, I'm more worried about supporting these multi-level decision trees in the long term. That's why I'm not positive on merging this right now.

rsertelon:
Hey @fnichol thanks for your comment on this one. I've got no problem not merging it ;) Can one of you close this PR and related issue then? :)

christophermaier:
Closing per discussion above.

PR #4143, opened 2017-11-28
Allow for running containers as a non-root user

### LINKED ISSUE #4137: Generalize Docker export to facilitate running as non-root users
Some environments (notably OpenShift) can constrain containers to run as non-root users for security purposes.

christophermaier:
In some situations, it can be useful to generate a customized Docker
container with control over the user ID of a service's main user, as
well as whether or not the Supervisor itself should run as root, or as
the service user.

This is motivated by a need to run in certain OpenShift environments.
Note, however, that this does not (yet?) make it possible to run in
_any_ OpenShift environment. In particular, this allows running in
environments where the user ID of the container user is defined (and
not random, as is the default for OpenShift Online, for
instance). Further work is needed to unlock *that* use case.

This commit introduces a number of changes in container generation to
make this possible:

1.) By specifying a value for the `--user-id` option, you can control
the ID for the service user. Previously, this was a hardcoded value of
`42`, which remains the default.

2.) By specifying the flag `--non-root`, the Supervisor will be run as
the service user. This may be used with or without `--user-id`, as
necessary.

3.) In order to support non-root user cases, the `/hab` directory tree
is no longer unconditionally owned by `root`. Instead, if `--non-root`
is specified, the service user owns the directory. The
`/hab/svc/SERVICE/{data,config}` volumes are also owned by this user,
as appropriate.

Be aware that because of this change of permissions, non-root
containers will be larger, due to the additional layer the `chown`
command creates.

4.) To support running as the service user, we no longer
unconditionally create a `hab` user and group. Instead, we consult the
service's `SVC_USER` and `SVC_GROUP` metadata. This allows services to
be run as their appropriate owner.

As an example, if you would like to generate a `core/redis` container
image with the Supervisor running as a non-root service user with a
UID of `12345`, you would run:

    hab pkg export docker core/redis --user-id=12345 --non-root

Fixes #4137

### DIFF
--- a/components/pkg-export-docker/src/build.rs
+++ b/components/pkg-export-docker/src/build.rs
@@ -39,6 +39,7 @@ use util;
 const DEFAULT_HAB_IDENT: &'static str = "core/hab";
 const DEFAULT_LAUNCHER_IDENT: &'static str = "core/hab-launcher";
 const DEFAULT_SUP_IDENT: &'static str = "core/hab-sup";
+const DEFAULT_USER_ID: u32 = 42;
 
 /// The specification for creating a temporary file system build root, based on Habitat packages.
 ///
@@ -64,6 +65,10 @@ pub struct BuildSpec<'a> {
     /// A list of either Habitat Package Identifiers or local paths to Habitat Artifact files which
     /// will be installed.
     pub idents_or_archives: Vec<&'a str>,
+    /// Numeric user ID of the user / group
+    pub user_id: u32,
+    /// Run the container as a non-root user?
+    pub non_root: bool,
 }
 
 impl<'a> BuildSpec<'a> {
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

components/pkg-export-docker/src/build.rs
```diff
@@ -494,6 +521,10 @@ impl BuildRootContext {
         self.rootfs.as_ref()
     }
 
+    pub fn primary_user_id(&self) -> u32 {
+        if self.non_root { self.user_id } else { 0 }
```
reset: Is root's UID always 0? Should we do a syscall to figure it out instead?

reset:
@christophermaier this looks good to me!

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/315369922).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/315369922).

Issue #4126, opened 2017-11-27
implement docker exporter for windows containers


mwrock:
Currently one can only export docker containers on linux. If one is on windows and is running docker in "windows container mode," then the docker exporter should produce a windows container for the package specified.

mwrock:
closed by #4147

PR #4147, opened 2017-11-29
Implement docker exporter for windows containers

mwrock:
This will cause `hab pkg export docker <ident>` to build a windows based docker image when run on windows. The `<ident>` will need to be a windows targeted package. Also the docker server running on the Windows machine will need to be in "windows container mode" otherwise we throw an error asking the user to switch to that mode.

Note this PR fixes an edge case supervisor bug that is not an edge case in a container. More on that later.

There are a couple key differences in the image creation for windows:

1. We do not wrap the `entrypoint` in a separate script but instead run `hab` as the entrypoint. We do this because the script is essentially unnecesary and we remove a process layer by doing so.
2. We do not need to add file entries for users, groups and networking because our base windows image defaults should be what we want.

### Supervisor fix

The supervisor will panic if it starts within 30 seconds of system boot. This is because rust's `Instance` struct is based on a counter that starts at 0 from boot time. When the supervisor initializes `last_health_check` to 30 seconds prior to `Now()` that will cause an overflow if `Now()` is less than 30 seconds from boot. To fix this we change `last_health_check` to an `option` and initialize it to `None`.

Besides the supervisor fix, this code should have no effect on Linux exports.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -37,11 +37,11 @@ environment:
 
   matrix:
     - hab_build_action: "test;build"
-      hab_components: "core;http-client;builder-protocol;builder-depot-client;common;win-users;sup;hab-butterfly"
+      hab_components: "core;http-client;builder-protocol;builder-depot-client;common;win-users;sup;hab-butterfly;pkg-export-docker"
 
     - hab_build_action: "package"
       hab_exe_version: "%24latest"
-      hab_components: "hab;plan-build-ps1;studio;sup;hab-butterfly;bintray-publish"
+      hab_components: "hab;plan-build-ps1;studio;sup;hab-butterfly;pkg-export-docker;bintray-publish"
 
 build_script:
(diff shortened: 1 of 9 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
Looks reasonable.. would be good to rebase and combine the commits before merge.

mwrock:
closes #4126

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/309160304).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/309160304).

PR #4169, opened 2017-11-30
Don't abort hab-studio.sh if `umount` command fails

baumanj:
This should be applied to other cleanup activities as well, but I wanted to get some feedback on the approach before I do more of it.

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -1050,14 +1050,25 @@ cleanup_studio() {
   unmount_filesystems
 }
 
+# **Internal**  Run a command which may fail without aborting whole script
+try() {
+  echo "TRYing '$@'"
+  if "$@"; then
+    status=$?
+  else
+    status=$?
+    echo "Warning: '$@' failed with status $status"
+  fi
+}
+
 # **Internal** Kills a Launcher process, if one exists.
 kill_launcher() {
   local pid_file
   pid_file="$HAB_STUDIO_ROOT/hab/sup/default/LOCK"
 
   if [ -f $pid_file ]; then
-    $bb kill $($bb cat $pid_file) \
-      && $bb rm -f $pid_file
+    try $bb kill $($bb cat $pid_file) \
+      && try $bb rm -f $pid_file
   fi
 }
 
@@ -1071,8 +1082,18 @@ chown_artifacts() {
       -a -d "$ARTIFACT_PATH" \
   ]; then
     local artifact_path_owner
-    artifact_path_owner="$($bb stat -c '%u:%g' $ARTIFACT_PATH)"
-    $bb chown -R "$artifact_path_owner" "$ARTIFACT_PATH"
+    artifact_path_owner="$(try $bb stat -c '%u:%g' $ARTIFACT_PATH)"
+    try $bb chown -R "$artifact_path_owner" "$ARTIFACT_PATH"
+  fi
+}
+
+# **Internal** Unmount mount point if mounted
+# Don't abort script on umount failure
+# ARGS: [umount_options] <mount_point>
+try_umount() {
+  eval _mount_point=\$$# # getting the last arg is surprisingly hard
+  if $bb mount | $bb grep -q "on $_mount_point type"; then
+    try $bb umount $*
   fi
 }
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

baumanj:
If this looks good, I'd add similar things for `kill_launcher` and `chown_artifacts`.

components/studio/bin/hab-studio.sh
```diff
@@ -1076,6 +1076,35 @@ chown_artifacts() {
   fi
 }
 
+# **Internal**  Run a command which may fail without aborting whole script
+try() {
+  local status
+  if [[ $- == *e* ]]; then
```
fnichol: You'd need to use `if [ $- = *e* ]; then` here to make sure it would work outside of Bash--I *think* the wildcard match would still work though.

components/studio/bin/hab-studio.sh
```diff
@@ -1076,6 +1076,35 @@ chown_artifacts() {
...
+    # Disable `set -e` as briefly as possible to avoid masking other errors
+    set +e
+    $*
+    status=$?
+    set -e
+  else
+    $*
+    status=$?
+  fi
```
fnichol: Would something like this work to avoid the `set -e` check?

```sh
if $*; then
  status=$?
else
  status=$?
fi
```

components/studio/bin/hab-studio.sh
```diff
@@ -1076,6 +1076,35 @@ chown_artifacts() {
...
+  if [ $status -ne 0 ]; then
+    echo "Warning: '$*' failed with status $status"
+  fi
+}
+
+# **Internal** Unmount mount point if mounted
+# Don't abort script on umount failure
+# ARGS: [umount_options] <mount_point>
+try_umount() {
+  eval _mount_point=\$$# # getting the last arg is surprisingly hard
```
fnichol: Wow! That ought to work as far as I know. You could also use `awk` for this with

```sh
_mount_point="$(echo "$@" | awk '{print $NF}')"
```

components/studio/bin/hab-studio.sh
```diff
@@ -1091,43 +1120,23 @@ unmount_filesystems() {
...
-  fi
+  try_umount $v -l $studio_artifact_path
 
-  if $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/run type"; then
-    $bb umount $v $HAB_STUDIO_ROOT/run
-  fi
+  try_umount $v $HAB_STUDIO_ROOT/run
 
-  if $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/sys type"; then
-    if [ -z "${KRANGSCHNAK+x}" ]; then
```
fnichol: For the moment we'll need to preserve this if/else case with the slightly different `umount` args. It's a temporary fix to handle `/sys` differently if it was set up in a user namespace or not.

components/studio/bin/hab-studio.sh
```diff
@@ -1076,6 +1076,35 @@ chown_artifacts() {
...
+    echo "Warning: '$*' failed with status $status"
+  fi
+}
+
+# **Internal** Unmount mount point if mounted
+# Don't abort script on umount failure
+# ARGS: [umount_options] <mount_point>
+try_umount() {
+  eval _mount_point=\$$# # getting the last arg is surprisingly hard
+  if $bb mount | $bb grep "on $_mount_point type"; then
```
fnichol: I think you'd still want `grep -q` to suppress the grep matches if any are found.

components/studio/bin/hab-studio.sh
baumanj: Good catch; I didn't mean to change that.

components/studio/bin/hab-studio.sh
baumanj: I don't think that works with [single brackets](http://www.tldp.org/LDP/abs/html/comparison-ops.html#SCOMPARISON1), unfortunately. But, I think your later comment obviates the need to mess with `-e`.

components/studio/bin/hab-studio.sh
baumanj: It's strange; I think this was my first approach, but I couldn't get it working right. I tried again and it seems correct. I changed `$*` to `"$@"` to make `shellcheck` happy, but it seems to work correctly in `sh`:
```sh
➤ sh -l
$ good() { return 0; }
$ bad()  { return 7; }
$ try() {
>     if "$@"; then
>         status=$?
>     else
>         status=$?
>         echo "Warning: '$*' failed with status $status"
>     fi
> }
$ try true
$ try false
Warning: 'false' failed with status 1
$ try touch foo
$ try touch /nonexist/bar
touch: cannot touch '/nonexist/bar': No such file or directory
Warning: 'touch /nonexist/bar' failed with status 1
```

Thanks for pointing this out; it's much nicer!

components/studio/bin/hab-studio.sh
baumanj: Yeah, it's surprising to me that there's not an easier way. I'll stick with this unless you think the `awk` version would be more easily understood by most.

baumanj:
I incorporated your feedback, @fnichol. I did it as a separate commit to make it easy to see. I intentionally left off the signoff to make sure I remember to squash them before merge.

I'm going to work on applying the `try` pattern to the rest of the cleanup stuff.

fnichol:
That's an interesting GitHub hack (leaving off the DCO line)--I might use that in future PR rebases

components/studio/bin/hab-studio.sh
fnichol: I originally was trying a variant of `if ! $*` but I think that flipped the exit code for the body of the conditional. Interesting stuff

baumanj:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/311411809).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/311411809).

PR #4174, opened 2017-12-01
adding all tdeps to be direct deps of hab-backline to fix studio chroot

mwrock:
This fixes the studio chroot where `hab pkg exec` is failing to run bash. The new environment variable resolution does not include tdeps in `pkg exec` which leaves out bash from the `hab-backline` package used by the studio.

To at least get things running again, this adds all deps as direct deps.


thesentinels:

Thank you for contributing!

Issue #3521, opened 2017-10-05
[STABILITY] Permission denied issue on builder-api-proxy directory


chefsalim:
We need to fix the issue where the supervisor sets the service directories to be owned by root, when they should be owned by the 'hab' user (nginx scenario).

chefsalim:
Re-opening as the hook fix does not seem to work.
Also we should handle the permission change for the builder api data folder.

chefsalim:
Re-opening as this is still occuring

tashimi:
pretty sure this issue is still happening, and known to be so; Chris had to fix the permissions in acceptance in builder-api-proxy manually today

burtlo:
Thank you very much for fixing this!

PR #4193, opened 2017-12-01
Ensure permissions are set on the config directory

christophermaier:
NOTE: This is a temporary bandaid for the micro-outages we've been
having with the API server every time the service restarts. The real
fix lies in how the Supervisor starts services (and more specifically,
which user they are started by). Addressing that should (partially)
unlock new behaviors for the Supervisor (like the ability to more
sanely run as a non-root user).

Fixes: #3521


thesentinels:

Thank you for contributing!

PR #4281, opened 2017-12-09
Updates Segment tag

### LINKED ISSUE #3887: Update Segment tag.
Please update the current Segment snippet for www and the builder web app with the attached version.
[habitat-segment-chef-key.txt](https://github.com/habitat-sh/habitat/files/1412089/habitat-segment-chef-key.txt)

rsertelon:
Didn't know how to make it a one liner easily, if someone has a tip to do so, might be better I think ;)

Should close #3887

### DIFF
--- a/components/builder-web/index.html
+++ b/components/builder-web/index.html
@@ -40,7 +40,61 @@
     })(window,document,'script','dataLayer','GTM-5VR24H');</script>
     <!-- End Google Tag Manager -->
     <!-- Segment -->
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

cnunciato:
Slim cares about indentation, so this fails when I attempt to run it. (You can do that also by running `make run` from within the `www` directory, though in order to get useful failure info, you'll need to run `bundle exec middleman build --verbose`). I suspect we'll want to use an `.erb` template for this instead.

This script also assumes the presence of jQuery, which is all right for `www`, but not for Builder (we don't bundle jQuery there). I've messaged @hamburglar in the associated issue to see if we can get a non-jQuery version of it.

cnunciato:
Actually now that I run this, even with the ERB template, I'm still seeing complaints about `jQuery` missing in the console. Maybe a load-ordering thing.

cnunciato:
Either this script is problematic or we need to do some additional configuration at the receiving end in order to get things to line up. Let's wait to hear back from Jason and then we'll figure out next steps.

Issue #4207, opened 2017-12-04
Ability to "test drive" releases before publish


mwrock:
Currently our travis and appveyor builds publish artifacts to stable as bits are being built and finally publish the studio image and cli to bintray at the very end. We then "sanity check" the release before officially announcing it. This usually entails downloading from bintray and entering a studio. While this  has historically been a smooth process, we have recently experienced pain if a release includes a defect. By the time we discover the defect, the new release is already in the hands of users. Additionally, if the release itself fails, we end up in a state where some artifacts are published and others are not.

What we want is to have an atomic means of releasing all artifacts and a way to manually check the release before it is published to our stable builder and bintray channels.

One straightforward path to getting this functionality before building our pipeline into builder would be to have travis/appveyor releases publish to a new builder channel "release" and NOT publish the cli to bintray at all. So once the release build completes, if it failed, no users are impacted. If it succeeds, we can `hab pkg install` the `hab` package from the `release` channel and test that cli and its studio. Once we deem this release acceptable for public consumption, we should be able to run a script that promotes the release packages to stable and publishes the cli to bintray. Eventually we could automate the "manual" testing as well.

mwrock:
This will create a new builder channel on the fly: `rc-[VERSION]` and will upload all packages to that channel. We will add a new bintray repo called `release-candidate` and publish the cli to that repo. Studio images will continue to be published to stable - thats ok because you can only get that studio using its correspondingly versioned cli.

Once an RC is validated, a deployer will invoke a bash script which:
* iterates all packages in the rc channel and promotes to `stable`
* downloads the rc cli for all platforms from bintray and uploads/publishes to bintray's `stable` repo

mwrock:
fixed by #4297

PR #4297, opened 2017-12-12
release should stage all packages in an RC channel and not publish the cli which is done from a makefile script

mwrock:
See the release.md change for a full description of what this includes.

### DIFF
--- a/support/ci/deploy.sh
+++ b/support/ci/deploy.sh
@@ -2,26 +2,29 @@
 
 set -e
 
+# now do the linux unstable build
+HAB_VERSION=$(cat VERSION)
+BOOTSTRAP_DIR=/root/travis_bootstrap
+TEST_BIN_DIR=/root/hab_bins
+TRAVIS_HAB=${BOOTSTRAP_DIR}/hab
+HAB_DOWNLOAD_URL="https://api.bintray.com/content/habitat/stable/linux/x86_64/hab-%24latest-x86_64-linux.tar.gz?bt_package=hab-x86_64-linux"
+BINTRAY_REPO=unstable
+CHANNEL=unstable
+
+export HAB_ORIGIN=core
+
 # fail fast if we aren't on the desired branch or if this is a pull request
 if [ -n "$TRAVIS_BRANCH" || -n "$TRAVIS_PULL_REQUEST" ]; then
-  if  [[ "${TRAVIS_BRANCH}" != "$(cat VERSION)" && ("${TRAVIS_PULL_REQUEST}" != "false" || "${TRAVIS_BRANCH}" != "master") ]]; then
+  if  [[ "${TRAVIS_BRANCH}" != "$HAB_VERSION" && ("${TRAVIS_PULL_REQUEST}" != "false" || "${TRAVIS_BRANCH}" != "master") ]]; then
       echo "We only publish on successful builds of master."
       exit 0
   fi
 fi
 
-# now do the linux unstable build
-BOOTSTRAP_DIR=/root/travis_bootstrap
-TEST_BIN_DIR=/root/hab_bins
-TRAVIS_HAB=${BOOTSTRAP_DIR}/hab
-HAB_DOWNLOAD_URL="https://api.bintray.com/content/habitat/stable/linux/x86_64/hab-%24latest-x86_64-linux.tar.gz?bt_package=hab-x86_64-linux"
-export HAB_ORIGIN=core
-
-BINTRAY_REPO=unstable
-CHANNEL=unstable
-if [ "$(cat VERSION)" == "$TRAVIS_TAG" ]; then
+if [ "$HAB_VERSION" == "$TRAVIS_TAG" ]; then
   BINTRAY_REPO=stable
-  CHANNEL=stable
+  CHANNEL=rc-$HAB_VERSION
+  export HAB_BLDR_CHANNEL=$CHANNEL
 fi
 
 mkdir -p ${BOOTSTRAP_DIR}
@@ -81,5 +84,9 @@ env HAB_ORIGIN= ${TRAVIS_HAB} studio run sh -c \'rm -f /hab/cache/keys/*-*.sig.k
 if [ -n "$BINTRAY_USER" ]; then
   echo "Publishing hab to $BINTRAY_REPO"
   env  HAB_BLDR_CHANNEL=$CHANNEL ${TRAVIS_HAB} pkg exec core/hab-bintray-publish publish-studio
-  env  HAB_BLDR_CHANNEL=$CHANNEL ${TRAVIS_HAB} pkg exec core/hab-bintray-publish publish-hab -r $BINTRAY_REPO $RELEASE
+  if [ "$HAB_VERSION" == "$TRAVIS_TAG" ]; then
+    env  HAB_BLDR_CHANNEL=$CHANNEL ${TRAVIS_HAB} pkg exec core/hab-bintray-publish publish-hab -s -r $BINTRAY_REPO $RELEASE
+  else
+    env  HAB_BLDR_CHANNEL=$CHANNEL ${TRAVIS_HAB} pkg exec core/hab-bintray-publish publish-hab -r $BINTRAY_REPO $RELEASE
+  fi
 fi
--- a/support/ci/promote_rc.sh
+++ b/support/ci/promote_rc.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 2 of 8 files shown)


thesentinels:

Thank you for contributing!

mwrock:
closes #4207

support/ci/promote_rc.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+if [ -z ${BINTRAY_KEY+x} ]; then
+  echo "BINTRAY_KEY var is unset"
+  exit 1
+fi
+
+HAB_VERSION=$(cat VERSION)
+PROMOTE_CHANNEL=stable
+RC_CHANNEL=rc-$HAB_VERSION
+RC_PKGS_JSON=$(curl https://bldr.habitat.sh/v1/depot/channels/core/$RC_CHANNEL/pkgs | jq -r '.data' | jq -c '.[]')
```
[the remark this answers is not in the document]
christophermaier: Digging further, it looks like this will return all releases in the channel, not just the latest ones. In the ideal case, there will only be a single release for each package in this release channel, but if we have issues with the release and end up having to re-run part or all of it, we can end up with multiple releases for a given package.

Ideally there'd be a flag for this in the API, but barring that, we can probably leverage a bit of shell scripting to take care of things for us, likely with a bit of `awk` and `sort`. It may be easier to use `jq` to get the data into a form that's easier for those tools to interact with:

```sh
curl https://bldr.habitat.sh/v1/depot/channels/core/stable/pkgs | \
jq -cr '.data |
      map ([.origin +"/"+ .name +"/"+ .version +"/"+ .release, .origin, .platforms[0]]) |
      map(join(" ")) |
      .[]'
```
That would yield something like: 
```
core/7zip/16.04/20170209152224 core x86_64-windows
core/R/3.4.0/20170623211014 core x86_64-linux
core/R/3.4.2/20171011183236 core x86_64-linux
core/R/3.4.2/20171013091650 core x86_64-linux
core/R core/R/3.4.2/20171014220034 core x86_64-linux
...
```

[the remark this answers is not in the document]
christophermaier: Though I doubt we'll be running afoul of the pagination limit anytime soon, it may be prudent to at least check the result that comes back to ensure that we are indeed staying on one page of results.

christophermaier:
Also, when we're building packages, we'll want to be pulling preferentially from the `release-[VERSION]` channel, to ensure that we're using the packages from this release as dependencies; otherwise, we'll be depending on that last stable releases.

mwrock:
I don't think we can set `HAB_BLDR_CHANNEL` to the release channel because it will fail to find alot of the core deps (rust, glibc, etc). Because we are building the final binaries locally, even though we are on the stable channel, it should use the locally built version over the last stable version.

mwrock:
ok I addressed feedback @christophermaier

support/ci/promote_rc.sh
```diff
@@ -0,0 +1,83 @@
...
+declare -A pkg_releases
+for pkg in $RC_PKGS_JSON_DATA
+do
+    ORIGIN=$(echo $pkg | jq -r ".origin")
+    NAME=$(echo $pkg | jq -r ".name")
+    VERSION=$(echo $pkg | jq -r ".version")
+    RELEASE=$(echo $pkg | jq -r ".release")
+    TARGET=$(echo $pkg | jq -r ".platforms[0]")
+    IDENT=$ORIGIN/$NAME/$VERSION/$RELEASE
+    if [[ -z ${pkg_releases[$NAME-$TARGET]+x} || "${pkg_releases[$NAME-$TARGET]}" < "$IDENT" ]]; then
```
christophermaier: This logic merits a comment explaining why it's here. If we ever add options to the API to just return the latest releases from a channel, then it all goes away.

support/ci/promote_rc.sh
```diff
@@ -0,0 +1,83 @@
...
+        pkg_releases[$NAME-$TARGET]=$IDENT
+
+        if [ "$NAME" == "hab" ]; then
+            case $TARGET in
+                *-linux)
+                LINUX_RELEASE=$RELEASE
+                ;;
+                *-darwin)
+                DARWIN_RELEASE=$RELEASE
```
christophermaier: I don't think we actually host darwin packages in Builder yet, so this script is always going to exit on line 71

support/ci/promote_rc.sh
```diff
@@ -0,0 +1,83 @@
...
+HAB_VERSION=$(cat VERSION)
+PROMOTE_CHANNEL=stable
+RC_CHANNEL=rc-$HAB_VERSION
+RC_PKGS_JSON=$(curl https://bldr.habitat.sh/v1/depot/channels/core/$RC_CHANNEL/pkgs)
+RC_PKGS_JSON_DATA=$(echo $RC_PKGS_JSON | jq -c -r '.data | .[]')
+TOTAL_COUNT=$(echo $RC_PKGS_JSON | jq -r '.total_count')
+END=$(echo $RC_PKGS_JSON | jq -r '.range_end')
+
+if [ $(($END+1)) !=  $TOTAL_COUNT ]; then
+  echo "There are multipe pages of releases. Consider deleting the channel and then rebuild."
```
christophermaier: Maybe add a blurb about our default page size being 50 right now, and we are in no way expecting there to be anywhere _near_ that many packages in a normal release channel.

Just so Future Us knows why we didn't bother implementing pagination here.

support/ci/promote_rc.sh
```diff
@@ -0,0 +1,83 @@
...
+fi
+
+declare -A pkg_releases
+for pkg in $RC_PKGS_JSON_DATA
+do
+    ORIGIN=$(echo $pkg | jq -r ".origin")
+    NAME=$(echo $pkg | jq -r ".name")
+    VERSION=$(echo $pkg | jq -r ".version")
+    RELEASE=$(echo $pkg | jq -r ".release")
+    TARGET=$(echo $pkg | jq -r ".platforms[0]")
```
christophermaier: As an aside... why is `platforms` an array here (and not a scalar `platform`) if we only build releases for a single platform?

christophermaier:
@mwrock We can specify `HAB_BLDR_CHANNEL` when we're building a package:

https://github.com/habitat-sh/habitat/blob/6a38cdcbdbd89368119962342b94fb9fb5a0d9f2/components/plan-build/bin/hab-plan-build.sh#L670-L681

The build process will try to resolve a dependency from that channel first, and if it can't be found, it'll try searching in a fallback channel, which is set to `stable`. 

That way we'll make sure to build using the right dependencies. We didn't have to worry about this previously because `hab-plan-build` always pulls from `stable`, and we'd immediately promote packages to `stable` as they got made during the release process.

support/ci/promote_rc.sh
mwrock: yeah that's a great question. I just dug into that service code and it looks like we use some generic logic to handle listing packages. Such lists can be as granular as listing info by release (as we do here) or might list info by version or even origin/name. At the release level, it is extremely unlikely if not impossible there would be more than a single platform but there may be multiple platforms at the version or name level.

mwrock:
ah you are right!

mwrock:
Addressed latest feedback @christophermaier

support/ci/deploy.sh
```diff
@@ -2,26 +2,29 @@
...
+      echo "We only publish on successful builds of master."
+      exit 0
+  fi
+fi
+
+if [ "$HAB_VERSION" == "$TRAVIS_TAG" ]; then
   BINTRAY_REPO=stable
-  CHANNEL=stable
+  CHANNEL=rc-$HAB_VERSION
+  export HAB_BLDR_CHANNEL=$CHANNEL
```
christophermaier: It may be a bit easier to reason about if the build command has this set explicitly, rather than exporting it.

support/ci/promote_rc.sh
christophermaier: ☝️

mwrock: I put the comment just above the loop

christophermaier: ah, so you did!

christophermaier:
@mwrock Looks good; merge at your leisure.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/316712715).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/316712715).

PR #4298, opened 2017-12-13
Kubernetes exporter crate & some other refactoring/cleanups

zeenix:
These patches lay the foundation for the Helm exporter, that I'll be working on next. Some of the clean-up and refactoring patches are not exactly needed for helm exporter but would be nice to have anyway.

### DIFF
--- a/components/common/src/ui.rs
+++ b/components/common/src/ui.rs
@@ -105,6 +105,28 @@ impl UI {
         Self::new(Shell::default_with(coloring, isatty))
     }
 
+    /// Creates a new default `UI` with a coloring strategy and tty hinting.
+    pub fn default_with_env() -> Self {
+        let isatty = if env::var(NONINTERACTIVE_ENVVAR)
+            .map(|val| val == "true")
+            .unwrap_or(false)
+        {
+            Some(false)
+        } else {
+            None
+        };
+        let coloring = if env::var(NOCOLORING_ENVVAR)
+            .map(|val| val == "true")
+            .unwrap_or(false)
+        {
+            Coloring::Never
+        } else {
+            Coloring::Auto
+        };
+
+        UI::default_with(coloring, isatty)
+    }
+
     /// Creates a new `UI` from generic `Read` and `Write` streams.
     ///
     /// The standard input stream needs to implement `Read` and both the standard output and
--- a/components/common/src/ui.rs
+++ b/components/common/src/ui.rs
@@ -108,7 +108,8 @@ impl UI {
     /// Creates a new default `UI` with a coloring strategy and tty hinting.
     pub fn default_with_env() -> Self {
         let isatty = if env::var(NONINTERACTIVE_ENVVAR)
-            .map(|val| val == "true")
+            // Keep string boolean for backwards-compatibility
+            .map(|val| val == "1" || val == "true")
             .unwrap_or(false)
         {
             Some(false)
@@ -116,7 +117,7 @@ impl UI {
             None
         };
         let coloring = if env::var(NOCOLORING_ENVVAR)
-            .map(|val| val == "true")
+            .map(|val| val == "1" || val == "true")
             .unwrap_or(false)
         {
             Coloring::Never
--- a/components/pkg-export-docker/src/build.rs
+++ b/components/pkg-export-docker/src/build.rs
@@ -88,7 +88,9 @@ impl<'a> BuildSpec<'a> {
             channel: m.value_of("CHANNEL").unwrap_or(&default_channel),
             base_pkgs_url: m.value_of("BASE_PKGS_BLDR_URL").unwrap_or(&default_url),
             base_pkgs_channel: m.value_of("BASE_PKGS_CHANNEL").unwrap_or(&default_channel),
-            idents_or_archives: m.values_of("PKG_IDENT_OR_ARTIFACT").unwrap().collect(),
+            idents_or_archives: m.values_of("PKG_IDENT_OR_ARTIFACT")
+                .expect("No package specified")
+                .collect(),
             user_id: match m.value_of("USER_ID") {
                 Some(i) => {
                     // unwrap OK because validation function ensures it
--- a/components/pkg-export-docker/src/cli.rs
+++ b/components/pkg-export-docker/src/cli.rs
@@ -152,13 +152,13 @@ impl<'a, 'b> Cli<'a, 'b> {
                     .short("i")
                     .value_name("USER_ID")
                     .validator(valid_user_id)
-                    .help("Specify the numeric ID of the service user and group (default: 42)")
+                    .help(
+                        "Specify the numeric ID of the service user and group (default: 42)",
+                    ),
             )
-            .arg(
-                Arg::with_name("NON_ROOT")
-                    .long("non-root")
-                    .help("Run the container as a non-root user (default: false)")
-            );
+            .arg(Arg::with_name("NON_ROOT").long("non-root").help(
+                "Run the container as a non-root user (default: false)",
+            ));
         Cli { app: app }
     }
 
@@ -247,8 +247,8 @@ impl<'a, 'b> Cli<'a, 'b> {
                  .value_name("REGISTRY_TYPE")
                  .help("Remote registry type, Ex: Amazon, Docker (default: docker)"))
             .arg(Arg::with_name("REGISTRY_URL")
-                 // This is not strictly a requirement but will keep someone from making a mistake when
-                 // inputing an ECR URL
+                 // This is not strictly a requirement but will keep someone from making a mistake
+                 // when inputing an ECR URL
                  .requires("REGISTRY_TYPE")
                  .long("registry-url")
                  .short("G")
--- a/components/pkg-export-docker/src/cli.rs
+++ b/components/pkg-export-docker/src/cli.rs
@@ -149,7 +149,7 @@ impl<'a, 'b> Cli<'a, 'b> {
             .arg(
                 Arg::with_name("USER_ID")
                     .long("user-id")
-                    .short("i")
+                    .short("s")
                     .value_name("USER_ID")
                     .validator(valid_user_id)
                     .help(
--- a/components/pkg-export-docker/src/cli.rs
+++ b/components/pkg-export-docker/src/cli.rs
@@ -43,6 +43,10 @@ where
     pub app: App<'a, 'b>,
 }
 
+pub struct PkgIdentArgOptions {
+    pub multiple: bool,
+}
+
 impl<'a, 'b> Cli<'a, 'b> {
     pub fn new(name: &str, about: &'a str) -> Self {
         Cli {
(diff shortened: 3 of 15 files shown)


thesentinels:

Thank you for contributing!

components/pkg-export-docker/src/cli.rs
```diff
@@ -243,6 +243,26 @@ impl<'a, 'b> Cli<'a, 'b> {
 
         Cli { app: app }
     }
+
+    pub fn add_pkg_ident_arg(self, multiple: bool) -> Self {
```
krnowak: Booleans stink. I'd prefer using an enum I think.

components/pkg-export-kubernetes/src/topology.rs
```diff
@@ -0,0 +1,45 @@
...
+use error::Error;
+
+#[derive(Debug)]
+pub enum Topology {
+    Standalone,
+    Leader,
+}
+
+impl FromStr for Topology {
+    type Err = Error;
```
asymmetric: For my understanding, what's the point of this type alias?

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -221,6 +221,7 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
         .add_builder_args()
         .add_tagging_args()
         .add_publishing_args()
+        .add_pkg_ident_arg(false)
```
asymmetric: It's not clear what `false` does here. Either an `Enum` as @krnowak suggested or a `struct` would be an improvement I think.

components/pkg-export-docker/src/lib.rs
```diff
@@ -179,3 +183,41 @@ pub fn export(ui: &mut UI, build_spec: BuildSpec, naming: &Naming) -> Result<Doc
 
     Ok(image)
 }
+
+/// Creates a build specification and naming policy from Cli arguments, and then exports a Docker
+/// image to a Docker engine from them.
+///
+/// # Errors
+///
+/// * If the actual import fails.
```
asymmetric: Can we remove the "If"s? Or add them for each point?

components/pkg-export-docker/src/cli.rs
zeenix: When you know that you are not going to have more than 2 options ever, booleans is the right way to go IMO. The underlying clap API is using boolean too and this is just being passed verbatim. Don't see the point to introducing an enum, just to be translated to a boolean.

components/pkg-export-docker/src/lib.rs
```diff
@@ -179,3 +183,41 @@ pub fn export(ui: &mut UI, build_spec: BuildSpec, naming: &Naming) -> Result<Doc
...
+    let docker_image = export(ui, spec, &naming)?;
+    docker_image.create_report(
+        ui,
+        env::current_dir()?.join("results"),
+    )?;
+
+    if matches.is_present("PUSH_IMAGE") {
+        let credentials = Credentials::new(
+            naming.registry_type,
+            matches.value_of("REGISTRY_USERNAME").unwrap(),
```
asymmetric: Using `expect` might be a bit nicer for the user here and below.

zeenix:
I don't think the failing CI check is related to my changes: https://ci.appveyor.com/project/chef/habitat/build/0.51.0-dev-20171213023217/job/pt58nn9vie79lh7j unless I missed something?

components/pkg-export-docker/src/cli.rs
asymmetric: I think it's rather for clarity when invoking the function, see [my comment](https://github.com/habitat-sh/habitat/pull/4298#discussion_r156634425).

zeenix: hmm.. OK. I'll change that but i think I'll go with two separate public methods instead.

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -127,7 +129,9 @@ fn gen_docker_img(ui: &mut UI, matches: &clap::ArgMatches) -> Result<()> {
 
 fn gen_k8s_manifest(_ui: &mut UI, matches: &clap::ArgMatches) -> Result<()> {
     let count = matches.value_of("COUNT").unwrap_or("1");
-    let topology = matches.value_of("TOPOLOGY").unwrap_or("standalone");
+    let topology: Topology = FromStr::from_str(
+        matches.value_of("TOPOLOGY").unwrap_or("standalone"),
```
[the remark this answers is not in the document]
zeenix: This is not new code as you can see from the diff. I recall an issue about the fact that we don't have "none" topology supported.

components/pkg-export-kubernetes/src/topology.rs
zeenix: We get the topology specified as string from commandline arg and we need to convert to the enum. Viceversa for the Display implementation below.

components/pkg-export-kubernetes/src/main.rs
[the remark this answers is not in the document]
asymmetric: To be fair I'm not sure if "none" is currently the default topology, see [this](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/mod.rs#L862-L866).

@fnichol could you please clarify this point for us? :)

components/pkg-export-docker/src/lib.rs
zeenix: actually it shouldn't never get a NONE here so unwrap() is fine. The dependencies setup with clap ensure that this arg is always specified if PUSH_IMAGE is.

zeenix: but I do like the idea of using expect() instead, just to be super safe. That'd be independent of these changes though.

zeenix: Pushed two additional commits and address this. I believe I've addressed all your comments now but let me know if I missed something.

zeenix:
I'll do a rebase now to fix the merge conflict. Is that OK?

components/pkg-export-kubernetes/src/manifest.rs
```diff
@@ -61,7 +61,9 @@ impl Manifest {
         );
         let ring_secret_name = matches.value_of("RING_SECRET_NAME").map(|s| s.to_string());
         // clap ensures that we do have the mandatory args so unwrap() is fine here
-        let pkg_ident_str = matches.value_of("PKG_IDENT_OR_ARTIFACT").unwrap();
+        let pkg_ident_str = matches.value_of("PKG_IDENT_OR_ARTIFACT").expect(
```
asymmetric: Why is this its own commit? The previous one seems to handle the same situation.

asymmetric:
One nit unaddressed (about the "If"s in comments). More importantly, I'm not sure what we should do with the topology.

@cm @fnichol maybe one of you wants to chime in? Is the default topology "standalone" or is it "none"?

zeenix:
@asymmetric 

> One nit unaddressed (about the "If"s in comments)

https://github.com/habitat-sh/habitat/pull/4298/commits/e3c01ec292ca19ff5744da9af0ec09fd561bf82f

about the topology, that's not relevant to this PR really so I don't think that should block this PR.

christophermaier:
According to https://github.com/habitat-sh/habitat/blob/232ca6d863e1cc5af14ae343e688c0cd6abb319a/components/sup/src/manager/service/mod.rs#L862-L866, the default topology is standalone

zeenix:
Since the current review comments have been addressed, I'll take this opportunity to rebase this branch ab bit.

zeenix:
Rebase done. Please review. While the rest isn't super urgent, these two patches are urgent IMO:

* https://github.com/habitat-sh/habitat/pull/4298/commits/69a6e8af052802b3b6751745ef14aec0c22af070
* https://github.com/habitat-sh/habitat/pull/4298/commits/493968404245acc7c5d8e45c2c633551c9337b2d

components/pkg-export-docker/src/cli.rs
```diff
@@ -149,7 +149,7 @@ impl<'a, 'b> Cli<'a, 'b> {
             .arg(
                 Arg::with_name("USER_ID")
                     .long("user-id")
-                    .short("i")
+                    .short("s")
```
christophermaier: Eek, good catch. I'm surprised clap doesn't catch this at compile-time...

components/common/src/ui.rs
```diff
@@ -105,6 +105,28 @@ impl UI {
         Self::new(Shell::default_with(coloring, isatty))
     }
 
+    /// Creates a new default `UI` with a coloring strategy and tty hinting.
+    pub fn default_with_env() -> Self {
```
christophermaier: Nice

components/pkg-export-kubernetes/src/main.rs
```diff
@@ -200,6 +200,7 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
         .add_builder_args()
         .add_tagging_args()
         .add_publishing_args()
+        .add_pkg_ident_arg(PkgIdentArgOptions { multiple: true })
```
christophermaier: I thought this was supposed to be `false` in this crate?

components/pkg-export-docker/src/build.rs
```diff
@@ -88,7 +88,9 @@ impl<'a> BuildSpec<'a> {
             channel: m.value_of("CHANNEL").unwrap_or(&default_channel),
             base_pkgs_url: m.value_of("BASE_PKGS_BLDR_URL").unwrap_or(&default_url),
             base_pkgs_channel: m.value_of("BASE_PKGS_CHANNEL").unwrap_or(&default_channel),
-            idents_or_archives: m.values_of("PKG_IDENT_OR_ARTIFACT").unwrap().collect(),
+            idents_or_archives: m.values_of("PKG_IDENT_OR_ARTIFACT")
+                .expect("No package specified")
```
christophermaier: I like this pattern of `.expect` rather than `.unwrap` for required CLI args.

components/pkg-export-kubernetes/src/cli.rs
```diff
@@ -0,0 +1,164 @@
...
+            .add_bind_args()
+    }
+
+    pub fn add_docker_args(self) -> Self {
+        let cli = docker::Cli { app: self.app };
+        let app = cli.add_base_packages_args()
+            .add_builder_args()
+            .add_tagging_args()
+            .add_publishing_args()
+            .add_pkg_ident_arg(docker::PkgIdentArgOptions { multiple: true })
```
christophermaier: I thought this was supposed to be false here?

zeenix: ah yes, good catch.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/316473821).

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/316473821).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/316473821).

PR #4306, opened 2017-12-14
Be more thorough in generating PATHs from older packages

christophermaier:
With the addition of the new runtime environment variables feature, it
has become apparent that our fallback logic for extracting `PATH`
variables from older packages built before version 0.50.3 was not as
complete as it needed to be.

Now, when we encounter a dependency that does not have a
`RUNTIME_ENVIRONMENT` metadata file, we not only take the `PATH`
metadata from that dependency, but also from its _transitive_
dependencies.

This should more fully replicate the older behavior of the Supervisor,
which assembled this information at runtime prior to the 0.50.3
release.

### DIFF
--- a/components/plan-build/bin/environment.sh
+++ b/components/plan-build/bin/environment.sh
@@ -219,10 +219,12 @@ __populate_environment_from_deps() {
             # well as empty entries sometimes. The current resolution
             # algorithm should filter these out immediately.
             warn "No RUNTIME_ENVIRONMENT metadata file found for ${dep_ident}; falling back to PATH metadata"
-
-            local old_path=$(_read_metadata_file_for "${path_to_dep}" "PATH")
+            local old_path=$(__assemble_legacy_path "${path_to_dep}")
             if [ -n "${old_path}" ]; then
+                warn "Final generated PATH for ${path_to_dep}: ${old_path}"
                 __push_env "${env_name}" "PATH" "${old_path}" ":" "${dep_ident}"
+            else
+                warn "No PATH metadata found for ${path_to_dep}"
             fi
         fi
     done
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/plan-build/bin/environment.sh
```diff
@@ -238,6 +238,54 @@ __populate_environment_from_deps() {
...
+#
+# Don't need to do this for packages that already have a
+# RUNTIME_ENVIRONMENT file, because this operation will have already
+# been done when those packages were built. This is to help preserve
+# backward compatibility.
+__assemble_legacy_path() {
+    local path_to_dep="${1}"
+    local path=""
+
+    warn "Assembling legacy path for ${path_to_dep}"
```
fnichol: These could probably be `build_line` instead since we've already warned and now we're doing the backwards compatible activity. Otherwise you start to accumulate a lot or `warns` which really only have one triggered source (the old package metadata).

components/plan-build/bin/environment.sh
```diff
@@ -238,6 +238,54 @@ __populate_environment_from_deps() {
...
+    local tdeps=$(_read_metadata_file_for ${path_to_dep} "TDEPS")
+    if [ -n "${tdeps}" ]; then
```
fnichol: Not critical, but perhaps a Bash `[[` test?

components/plan-build/bin/environment.sh
```diff
@@ -238,6 +238,54 @@ __populate_environment_from_deps() {
...
+# Don't need to do this for packages that already have a
+# RUNTIME_ENVIRONMENT file, because this operation will have already
+# been done when those packages were built. This is to help preserve
+# backward compatibility.
+__assemble_legacy_path() {
+    local path_to_dep="${1}"
+    local path=""
+
+    warn "Assembling legacy path for ${path_to_dep}"
+    local tdeps=$(_read_metadata_file_for ${path_to_dep} "TDEPS")
```
fnichol: I've learned that if you split the variable declaration and variable setting that you'll be able to propagate an error back on `tdeps=$(_read_metadata_file_for ${path_to_dep} "TDEPS")` if `_read_metadata_file_for()` fails, whereas `local tdeps=$(_read_metadata_file_for ${path_to_dep} "TDEPS")` always returns 0.

I know we have more than a few places with this issue--I only learned later about this and it's helped in some places.

components/plan-build/bin/environment.sh
christophermaier: Sounds good; will change

components/plan-build/bin/environment.sh
christophermaier: Oh, that's nice to know! I'll make those tweaks

components/plan-build/bin/environment.sh
christophermaier: Yeah, that's probably a good call... I think I've just been habitually sticking to `[`, but I see a lot of your code uses `[[`, and we _are_ explicitly using `bash` and not `sh`.

baumanj: Out of curiosity, what's the benefit of `[[` over `[` here? I know some things require it, but if we're not using them, it seems like writing it the more portable way has merit.

baumanj: …and if we are explicitly using `bash` and not `sh`, maybe a `#!/bin/bash` would be helpful?

christophermaier: @baumanj Probably consistency and keeping the mental overhead low. This code doesn't strictly require `[[`, but there are numerous other places in this package where we are explicitly relying on bash-only features, so portability across shells isn't really a concern (and if it were, there'd be a lot more to change than just the test procedures we use).

robbkidd:
yes, please

This fixes some trouble I'm seeing with the `[app]-console` failing to run with the Ruby scaffolding.

<details>

```
# hab pkg exec robbkidd/supermarket supermarket-console
Loading production environment (Rails 5.0.3)
bin/rails: No such file or directory - infocmp
/hab/pkgs/robbkidd/supermarket/3.1.50/20171215193359/app/vendor/bundle/ruby/2.4.0/gems/rb-readline-0.5.4/lib/rbreadline.rb:1815:in `get_term_capabilities': undefined method `split' for nil:NilClass (NoMethodError)
...
```

* `infocmp` is in the `core/ncurses` package.
* A `core/ncurses` _is_ in my package's TDEPS.
* ... but does not appear in the package's PATH (as described by this PR)

</details>

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/317106400).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/317106400).

PR #4315, opened 2017-12-15
Enable provision.sh to be run without sudo

baumanj:
This avoids the situation where the user's cargo directory is owned by root,
necessitating excessive sudo usage. Instead, add sudo to the specific commands
which require it.

### DIFF
--- a/support/linux/install_dev_9_linux.sh
+++ b/support/linux/install_dev_9_linux.sh
@@ -21,6 +21,6 @@ else
   sudo -E addgroup --system hab || true
 fi
 
-sudo -E sh /tmp/install.sh
-sudo -E hab install core/busybox-static core/hab-studio
-sudo -E rm -rf /tmp/install.sh
+sudo sh /tmp/install.sh
+sudo hab install core/busybox-static core/hab-studio
+sudo rm -rf /tmp/install.sh
--- a/support/linux/provision.sh
+++ b/support/linux/provision.sh
@@ -14,24 +14,30 @@ pushd /src
 cp components/hab/install.sh /tmp/
 sh support/linux/install_dev_0_ubuntu_latest.sh
 sh support/linux/install_dev_9_linux.sh
-echo 'eval "$(direnv hook bash)"' >> /root/.bashrc
-echo "source /src/support/linux/helpers.sh" >> /root/.bashrc
+# shellcheck disable=SC2016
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

support/linux/provision.sh
```diff
@@ -14,24 +14,28 @@ pushd /src
...
 
-hab pkg install core/postgresql core/shadow -c stable
-hab pkg install core/builder-datastore -c unstable
-hab pkg install core/docker
-hab pkg install core/hab-pkg-export-docker
+sudo hab pkg install core/postgresql core/shadow -c stable
+sudo hab pkg install core/hab-sup core/hab-launcher
+sudo hab pkg install core/builder-datastore
+sudo hab pkg install core/docker
+sudo hab pkg install core/hab-pkg-export-docker
```
baumanj: Added a couple additional packages here that logs were complaining about missing

support/linux/provision.sh
```diff
@@ -14,24 +14,28 @@ pushd /src
...
 
-hab pkg exec core/shadow groupadd --force krangschnak
-hab pkg exec core/shadow useradd --groups=tty --create-home -g krangschnak krangschnak || echo "User 'krangschnak' already exists"
+sudo hab pkg exec core/shadow groupadd --force krangschnak
+sudo hab pkg exec core/shadow useradd --groups=tty --create-home -g krangschnak krangschnak
```
baumanj: I removed the `|| echo "User 'krangschnak' already exists` because that message will be output by `useradd` anyway and it potentially masks other errors. If we want to never fail on this, we should check for the specific
 exit value (9: username already in use).

support/linux/install_dev_9_linux.sh
```diff
@@ -21,6 +21,6 @@ else
   sudo -E addgroup --system hab || true
 fi
 
-sudo -E sh /tmp/install.sh
-sudo -E hab install core/busybox-static core/hab-studio
-sudo -E rm -rf /tmp/install.sh
+sudo sh /tmp/install.sh
+sudo hab install core/busybox-static core/hab-studio
+sudo rm -rf /tmp/install.sh
```
baumanj: Removing the `-E` to prevent having a `HAB_BLDR_URL` set interfering. If we decide we ever need this script to install from somewhere other than the default builder, we can always make it more flexible, but I tend to think YAGNI.

reset:
tenor-161872093

PR #4334, opened 2017-12-18
Ensure `--no-color` applies to all Supervisor output

### LINKED ISSUE #997: hab-sup --no-color doesn't strip color from all output
`--no-color` cli arg parsing was fixed in https://github.com/habitat-sh/habitat/pull/996, but not all text is stripped of color.

christophermaier:
This should ensure that the `--no-color` option truly does eliminate all ANSI color codes from Supervisor output (note that this does _not_ apply to any ANSI codes that might be present in the output of any supervised services).

There are a few key points:

1) We were embedding color codes directly in log messages to emphasize certain parts. This was done in a way that completely bypassed any `--no-color` option; the color would _always_ be there.

2) The Launcher is responsible for routing standard output and error from supervised services, and uses our core output library. However, that library is currently driven by a global variable that is not set in the Launcher (because it's a completely separate process). Now, we ensure that the Launcher uses whatever colorization option is specified by the Supervisor

3) Some places in the Supervisor commands were using the `UI::default()` function to set up their outputs, but this does not honor the `--no-color` option. Now, all uses of `UI` in the Supervisor commands goes through a common setup function that _does_ honor this option.

Starting up the Supervisor with `--no-color` now should truly configure things so there is no colored output. Future work still exists, though, including but not limited to:

- Trying to unify our output logic in a single place 
- re-introducing "emphasis" coloring, but in a semantically-clean way
- Making `--no-color` the default behavior for non-interactive uses

Further details are in the commit messages. 

Note that for users to fully experience this, a new Launcher package will need to be built and promoted.

Fixes #997


thesentinels:

Thank you for contributing!

christophermaier:
cc: @bixu

components/launcher/src/main.rs
```diff
@@ -22,7 +23,21 @@ use launcher::server;
...
+    // Since we have access to all the arguments passed to the
+    // Supervisor here, we can simply see if the user requested
+    // `--no-color` and set our global variable accordingly.
+    //
+    // This does, of course, rely on the option name used here staying
+    // in sync with the name used in the Supervisor.
+    //
+    // There is currently no short option to check; just the long
+    // name.
+    if args.contains(&String::from("--no-color")) {
```
georgemarshall: This seems redundant and could be simplified

`if args.contains("--no-color") {`

christophermaier: @georgemarshall Sadly, no 😦 : https://github.com/rust-lang/rust/issues/42671

raskchanky: Wow

components/sup/src/main.rs
```diff
@@ -1622,3 +1579,41 @@ fn update_composite_service_specs(
...
+// the scope of change contained.
+fn ui() -> UI {
+    let coloring = if hcore::output::is_color() {
+        Coloring::Auto
+    } else {
+        Coloring::Never
+    };
+
+    let isatty =  if env::var(NONINTERACTIVE_ENVVAR)
+    // Keep string boolean for backwards-compatibility
```
fnichol: It's not exactly backwards-compatibility, it's current 😄

components/sup/src/manager/service/config.rs
```diff
@@ -372,7 +371,7 @@ impl CfgRenderer {
                     cfg_dest.display()
                 );
                 outputln!(preamble ctx.svc.group, "Updated {} {}",
-                          Purple.bold().paint(template.as_str()),
```
fnichol: Purple, wow, we were using all the colors!

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/318822488).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/318822488).

PR #4353, opened 2017-12-20
[builder-worker] Create network namespace only at server boot.

### LINKED ISSUE #4264: [STABILITY] workers losing ens4 network, not able to recover
We're seeing workers in prod (and acceptance) lose the ens4 network

```
Dec 08 01:23:45 ip-10-0-0-242 hab[18939]: builder-worker.live(O): FATAL: Could not find network interface ens4
```

Checking the IP address:
```
root@ip-10-0-0-242:~# ip address show dev ens4 | grep 'inet '
Device "ens4" does not exist.
```

We need to figure out a more robust way to apply the ens4 interface.

fnichol:
This change modifies where and when an optional network namespace is set
up for use when running Studio builds.

Prior to this change a network namespace would be created just before
each `studio build` invocation and destroyed immediately afterwards.
Now, a namespace will be created when the worker service starts and
will, by default, reuse an existing network namespace (the key signal
being the presence of an `airlock-ns/` directory). The network namespace
management continues to be provided by the Airlock program.

If for some reason the existing network namespace needs to destroyed on
each start/restart of the worker service, a new configuration option of:

```toml
recreate_ns_dir = true
```

can be set. The default behavior set to `false` which will reuse any
network namespaces found.

Note that the failure scenarios are different, but hopefully more
desirable:

* On a fresh installation, a network namespace will be created once, at
boot.
* If the network namespace creation fails, this happens at worker boot
and will sent the service in a failed/flapping state, loudly explaining
what is happening. Previously failures of this nature would have been
delayed until build time.
* If a worker is upgraded or restarted, it will reuse an existing
network namespace setup.
* If the network namespace is having issues, an operator can enable
network namespace recreation mode by setting the `recreate_ns_dir =
true` as above. The Supervisor would restart a worker service and the
destroy/create logic would be fired.

Finally, note that one major failure scenario is still unaddressed: if
the network interface's settings are not restored after a network
namespace teardown, then re-creating the namespace will fail. The good
news is that such a failure would effectively bring the worker offline
since it can't finish booting itself and would put itself out of the
ready pool of workers (rather than leaving a pool of dead workers as is
the case now). This final failure scenario can most likely be solved by
restarting the host's networking service. Sadly, this failure is highly
dependant on the cloud platform and operating system combination.

Closes #4264

### DIFF
--- a/components/builder-worker/src/runner/mod.rs
+++ b/components/builder-worker/src/runner/mod.rs
@@ -50,6 +50,7 @@ use self::docker::DockerExporter;
 use self::workspace::Workspace;
 use config::Config;
 use error::{Error, Result};
+use network::NetworkNamespace;
 use retry::retry;
 use vcs::VCS;
 
@@ -375,11 +376,11 @@ impl Runner {
     fn build(&mut self, log_pipe: &mut LogPipe) -> Result<PackageArchive> {
         log_pipe.pipe_buffer(b"\n--- BEGIN: Studio build ---\n")?;
 
-        let networking = match (
+        let network_namespace = match (
             self.config.network_interface.as_ref(),
             self.config.network_gateway.as_ref(),
         ) {
-            (Some(interface), Some(gateway)) => Some((interface.as_str(), gateway)),
+            (Some(_), Some(_)) => Some(NetworkNamespace::new(self.config.ns_dir_path())),
             (None, None) => None,
             (None, Some(_)) => return Err(Error::NoNetworkInterfaceError),
             (Some(_), None) => return Err(Error::NoNetworkGatewayError),
@@ -389,7 +390,7 @@ impl Runner {
             &self.config.bldr_url,
             &self.bldr_token,
             self.config.airlock_enabled,
-            networking,
+            network_namespace,
         ).build(log_pipe)?;
         log_pipe.pipe_buffer(b"\n--- END: Studio build ---\n")?;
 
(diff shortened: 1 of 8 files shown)


thesentinels:

Thank you for contributing!

fnichol:
Note that I still have to test this a bit, but that currently requires me to spin up an Ubuntu VM with user namespace support kernel-enabled. If someone beats me to trying this and it's good, that's cool too 😉

fnichol:
Assuming this goes green, we should be good for a review and merge. Finally!

components/builder-worker/src/runner/mod.rs
```diff
@@ -375,11 +376,11 @@ impl Runner {
     fn build(&mut self, log_pipe: &mut LogPipe) -> Result<PackageArchive> {
         log_pipe.pipe_buffer(b"\n--- BEGIN: Studio build ---\n")?;
 
-        let networking = match (
+        let network_namespace = match (
             self.config.network_interface.as_ref(),
             self.config.network_gateway.as_ref(),
         ) {
-            (Some(interface), Some(gateway)) => Some((interface.as_str(), gateway)),
+            (Some(_), Some(_)) => Some(NetworkNamespace::new(self.config.ns_dir_path())),
```
chefsalim: Ah much nicer :)

chefsalim:
tenor-253424191

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/330124398).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/330124398).

PR #4357, opened 2017-12-20
Fix ipc-channel dependency

asymmetric:
The previously referenced commit doesn't exist in the `ipc-channel`
repo, resulting in an error when building:

```
root@vagrant:/src# cargo build --verbose
    Updating git repository `https://github.com/habitat-sh/ipc-channel`
error: failed to load source for a dependency on `ipc-channel`
Caused by:
  Unable to update https://github.com/habitat-sh/ipc-channel?branch=hbt-windows#2d085faf
Caused by:
  [4/-3] revspec '2d085faf5bc053fb321e0d375b7328d023ea3436' not found
```

Ran this with `cargo update -p ipc-channel` from the root of the repo.

--

## Update

Just using the latest commit didn't work, a it introduced a newer version of a dependency, with its own incompatibilities.

I instead opted for hard-coding the second-to-last commit.

It might be better to just fix the errors and use the latest version though, but since this is breaking builds today, I thought a quickfix might be useful.

Let me know if I should squash this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,6 +131,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "byteorder 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

asymmetric:
Not sure why the build fails on Windows but not on Linux.

asymmetric:
Fixed in #4364.

PR #4364, opened 2017-12-20
[Cargo.lock] Temporarily pin ipc-channel to pre-upgrade commit.

fnichol:
This change will be updated/removed once our Cargo dependency update
changeset is merged. The commit sha is updated to the commit immediately
prior to the ipc-channel update commit in the `hbt-windows` branch of
the Habitat fork. In this way, master should be buildable in the
meantime.


fnichol:
A small note to folks: this doesn't 100% address builds on Windows, however it makes the build slightly better.

PR #4366, opened 2017-12-20
Help for privilege-dropping services

### LINKED ISSUE #3521: [STABILITY] Permission denied issue on builder-api-proxy directory

christophermaier:
ONCE MORE, WITH FEELING!

OK, here's the deal. Our `core/nginx` runs with a `SVC_USER` of `root`; all well and good. However, `nginx`'s worker processes generally like to run as a less-privileged user for security purposes. Our `core/nginx` plan is compiled to bake in the `hab` user and group for these workers.

The way the Supervisor works is that it sets up `/hab/svc/$SERVICE/` and all the various subdirectories within to be owned solely by the `SVC_USER` user, with permissions `700`. This causes problems for us in our `core/builder-api-proxy` package (and indeed, any service built with `core/nginx`) because these worker processes need access to those service directories. In the past, we've tried to work around this by adding `chown` commands in our `init`, `run`, and `reconfigure` hooks. This actually works most of the time; the Supervisor will set the ownership of the directories to `root`, but the hooks ensure that by the time `nginx` is started, the directories are owned by `hab`.

This works perfectly when the `core/builder-api-proxy` package is updated; where this breaks down is when the _Supervisor itself_ is updated. In this scenario, the Supervisor will shut down, but leave everything it is supervising up and running. When the new version of the Supervisor starts up, it runs through its permission-changing dance, but because it detects that the service is already running, it does _not_ run the service's `init` hook. Thus, every time the Supervisor updates itself, we have a brief API outage until the permissions can be changed back manually (restarting the `core/builder-api-proxy` service would also fix things).

While investigating this situation, we discovered that while we required a `SVC_GROUP` for our services, we weren't actually making use of it in any meaningful way. 

Whenever we provision systems to run Habitat, we always create a `hab` user in a `hab` group, since the vast majority of services end up using these (they're the default). We can take advantage of this by having the Supervisor set group permissions for a package's service directories, instead of completely locking the group out; we can similarly ensure that rendered configuration files are group-readable. With this in place, we can modify our `core/builder-api-proxy` package to use the `SVC_GROUP` of `hab`, and not `root` (this is actually how `core/nginx` is set up, too). This will ensure that the `nginx` worker processes, which are running as the `hab` user and group, will be able to access the directories and files they need.

By having the Supervisor handle this, we can remove the workarounds in `core/builder-api-proxy`'s hooks, and can actually remove most of the hooks! Without the extra code, the `run` hook is effectively the same as the one that `core/hab-plan-build` generates based on `pkg_svc_run` in `plan.sh`.

This approach should be useful for other services that behave in a similar way, such as those based on `core/httpd`. It still relies on out-of-band knowledge of the less-privileged user, though, since there is currently no plan metadata that exposes this. That being said, it seems a good step forward for Habitat, and also means that `SVC_GROUP` now has some concrete meaning within the system.

Fixes #3521


thesentinels:

Thank you for contributing!

raskchanky:
Rebasing this on master should clear the travis failure

components/builder-api-proxy/hooks/run
```diff
@@ -1,14 +0,0 @@
...
-if [ "\$(whoami)" = "root" ]; then
-  chown -R hab:hab {{pkg.svc_var_path}}
-  chown -R hab:hab {{pkg.svc_config_path}}
-  exec chpst \
-    -U {{pkg.svc_user}}:{{pkg.svc_group}} \
-    -u {{pkg.svc_user}}:{{pkg.svc_group}} \
-    ${pkg_svc_run} 2>&1
-else
-  exec ${pkg_svc_run} 2>&1
-fi
```
fnichol: This feels like win to me!

components/builder-api-proxy/plan.sh
```diff
@@ -6,7 +6,7 @@ pkg_license=("Apache-2.0")
 pkg_deps=(core/nginx core/curl core/coreutils)
 pkg_build_deps=(core/node8 core/gcc core/git core/tar core/phantomjs core/python2 core/make)
 pkg_svc_user="root"
-pkg_svc_group="root"
+pkg_svc_run="nginx -c ${pkg_svc_config_path}/nginx.conf"
```
fnichol: I believe this *should* work, although it might be better to try `pkg_svc_run="nginx -c {{pkg.svc_config_path}}/nginx.conf"` while does delay this to runtime but lets the Supervisor do the work (ex: https://github.com/fnichol/rocketry/blob/master/habitat/plan.sh#L16). The only catch (at the moment) is to add a fix to `do_build_service()` [like this](https://github.com/fnichol/rocketry/blob/master/habitat/plan.sh#L51-L57) until we fix the build program to put the `run` hook under `hooks/` (so that `run` always gets template-rendered).

However, if the `run` hook with this change works, ship it. It's already better in my book!

components/builder-api-proxy/hooks/init
```diff
@@ -1,5 +1,3 @@
 #!/bin/sh
 set -e
 mkdir -p {{pkg.svc_var_path}}/nginx
-chown -R hab:hab {{pkg.svc_var_path}}
-chown -R hab:hab {{pkg.svc_config_path}}
```
fnichol: This should be the sign we're winning

components/sup/src/manager/service/config.rs
```diff
@@ -42,6 +42,7 @@ static ENV_VAR_PREFIX: &'static str = "HAB";
 /// is deeper than this value crosses into overly complex territory when describing configuration
 /// for a single service.
 static TOML_MAX_MERGE_DEPTH: u16 = 30;
+pub const CONFIG_PERMISSIONS: u32 = 0o740;
```
fnichol: Good call!

I'm assuming this permission is on a directory entry? Is there a reason to avoid the execute bit for the group (i.e. `750` vs `740`)?

fnichol: Never mind, I opened up more context and saw that this is working on a file not directory, disregard last 😄

components/sup/src/manager/service/mod.rs
```diff
@@ -62,6 +62,9 @@ pub use self::supervisor::ProcessState;
 
 static LOGKEY: &'static str = "SR";
 
+pub const SVC_DIR_PERMISSIONS: u32 = 0o770;
```
fnichol: I feel like `7` (`rwx`) is probably right for group here, although on first look I expected to see `5` (`rx`). Is this because the priv-dropped user may need to write under these directories? I might have answered my own question.

Oh while we're doing this, it might be worth considering adding a sticky bit to the group which would preserve the group of created child directories (I believe `0o1770` would do the trick). I don't think this is critical, but maybe worth a comment or `TODO` if it looks like this would be handy in the future?

components/sup/src/manager/service/mod.rs
```diff
@@ -806,7 +809,7 @@ impl Service {
                       file.as_ref().display(), e);
             return false;
         }
-        if let Err(e) = set_permissions(&file, 0o640) {
+        if let Err(e) = set_permissions(&file, GOSSIP_FILE_PERMISSIONS) {
```
fnichol: I like consistent constants

components/builder-api-proxy/plan.sh
christophermaier: I was following the example of other plans that do something similar. I also compared the generated run hook to what we used to have and it was basically identical.

components/sup/src/manager/service/mod.rs
christophermaier: I actually did start this with lower permissions, but my `nginx` test case ended up complaining because it didn't have sufficient privileges along the entire directory path.

I'll look into the sticky bit solution.

christophermaier: @fnichol Actually, can you give a bit more idea of what you're thinking re: sticky bit? I'm not seeing how it would come into play here.

fnichol: Sure! If we had a service dropping privileges to a `alpha` user who has a primary group of `alpha` and a member of the `beta` group and that service had `pkg_svc_group=beta`, then any directories created under the svc directory would be `alpha:alpha` (not `alpha:beta` which would be preferred). A sticky bit (which affects the group ownership) on a parent directory keeps that same group ownerships for new child dirs (I think files as well). In that way, the svc group name would percolate down the child dirs

components/builder-api-proxy/plan.sh
fnichol: Awesome, I wasn’t sure either way

fnichol: Awesome, I wasn’t sure either way

components/sup/src/manager/service/mod.rs
christophermaier: Reading section 15.4.5 of The Linux Programming Interface says something different... that the sticky bit is a restricted deletion flag, meaning an unprivileged process can only delete a file if it has write permissions on the directory _and_ also owns the file in question.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/320440709).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/320440709).

PR #4378, opened 2017-12-22
redirect both stderr out stdout from sup to log in windows studio

### LINKED ISSUE #4362: Windows studio not sending supervisor stderr to log
Only stdout is being forwarded to the supervisor's log file. Any errors that occur bubble directly to the console which is disorienting.

mwrock:
It's unfortunate that we have to use such a verbose route to accomplish this but `2>&1` does not work with background jobs and `Start-Process` does not allow both out and err to go to the same file.

fixes #4362 


thesentinels:

Thank you for contributing!

reset:
tenor-88744125

PR #4422, opened 2018-01-05
Update dev config script

chefsalim:
This updates the config script for the dev environment by adding a new key_dir config, and removing the deprecated auth_token config for the builder-worker.

tenor-123459764


reset:
tenor-103050292

PR #4434, opened 2018-01-09
Add a Helm exporter

zeenix:
Here is a PR for adding the first version of Helm exporter. In my limited testing, it works perfectly, except that the docker image push fails for some reason on me. Could have something to do with Fedora being weird about docker again, I'm not sure but the issue is reproducible with docker exporter too so irrelevant to these changes. I've a script (to follow, not to run) to test:

```
# Go to dir where the plan.sh lies
cd /home/zeenix/checkout/habitat-example-plans/mynodeapp/

# Build the plan
hab pkg build  -r ${PWD}../studio .

# Export the package
hab pkg export helm --push-image --username zeenix --password pass --registry-url docker.io --registry-type docker results/zeenix-testapp-*.hart

# Run operator
../habitat-operator/habitat-operator --kubeconfig /var/lib/kube-spawn/default/kubeconfig

# Initialize Helm (Tiller)
cat >/tmp/helm.yaml <<EOL
hi
hello
EOL

helm-exec init --service-account helm

# Install/deploy the app already! (but wait for the Helm Tiller pod to get ready)
helm install ./testapp

# Check the app running
kubectl get pods -w
kubectl port-forward PODNAME 8000 &
wget localhost:8000
cat index.html # You should see 'hello world'

# Clean-up after                                                                                                                       
helm delete HELM_APP_NAME # check `helm list` output for finding out the name
docker rmi IMAGE_ID
# Remove docker app from docker hub through web
hab studio -r /home/zeenix/checkout/habitat-example-plans/studio  rm .
```

### DIFF
--- a/.bldr.toml
+++ b/.bldr.toml
@@ -174,6 +174,7 @@ plan_path = "components/pkg-export-kubernetes"
 paths = [
   "components/core/*",
   "components/common/*",
+  "components/hab/*",
   "components/pkg-export-docker/*"
 ]
 
--- a/.bldr.toml
+++ b/.bldr.toml
@@ -177,6 +177,16 @@ paths = [
   "components/pkg-export-docker/*"
 ]
 
+[hab-pkg-export-helm]
+plan_path = "components/pkg-export-helm"
+paths = [
+  "components/core/*",
+  "components/common/*",
+  "components/hab/*",
+  "components/pkg-export-docker/*"
+  "components/pkg-export-kubernetes/*"
+]
+
 # NOTE: cfize has a dependency on hab-pkg-dockerize, but we are
 # *explicitly not building* `hab-pkg-dockerize` any more. Read more at
 # components/pkg-dockerize/README.md
--- a/components/hab/src/command/pkg/export/helm.rs
+++ b/components/hab/src/command/pkg/export/helm.rs
@@ -1,0 +1,35 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ffi::OsString;
+
+use common::ui::UI;
+
+use error::Result;
+
+const EXPORT_CMD: &'static str = "hab-pkg-export-helm";
+const EXPORT_CMD_ENVVAR: &'static str = "HAB_PKG_EXPORT_HELM_BINARY";
+const EXPORT_PKG_IDENT: &'static str = "core/hab-pkg-export-helm";
+const EXPORT_PKG_IDENT_ENVVAR: &'static str = "HAB_PKG_EXPORT_HELM_PKG_IDENT";
+
+pub fn start(ui: &mut UI, args: Vec<OsString>) -> Result<()> {
+    ::command::pkg::export::export_common::start(
+        ui,
+        args,
+        EXPORT_CMD,
+        EXPORT_CMD_ENVVAR,
+        EXPORT_PKG_IDENT,
+        EXPORT_PKG_IDENT_ENVVAR,
+    )
+}
--- a/components/pkg-export-helm/plan.sh
+++ b/components/pkg-export-helm/plan.sh
@@ -1,0 +1,80 @@
+pkg_name=hab-pkg-export-helm
+_pkg_distname=$pkg_name
+pkg_origin=core
+pkg_version="0.1.0"
+pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+pkg_license=('Apache-2.0')
+pkg_deps=(core/docker)
+pkg_build_deps=(
+  core/musl core/zlib-musl core/xz-musl core/bzip2-musl core/libarchive-musl
+  core/openssl-musl core/libsodium-musl
+  core/coreutils core/rust core/gcc core/make
+)
+pkg_bin_dirs=(bin)
+
+bin=$_pkg_distname
+
+_common_prepare() {
+  do_default_prepare
+
+  # Can be either `--release` or `--debug` to determine cargo build strategy
+  build_type="--release"
+  build_line "Building artifacts with \`${build_type#--}' mode"
+
+  # Used by the `build.rs` program to set the version of the binaries
+  export PLAN_VERSION="${pkg_version}/${pkg_release}"
+  build_line "Setting PLAN_VERSION=$PLAN_VERSION"
+
+  if [ -z "$HAB_CARGO_TARGET_DIR" ]; then
+    # Used by Cargo to use a pristine, isolated directory for all compilation
+    export CARGO_TARGET_DIR="$HAB_CACHE_SRC_PATH/$pkg_dirname"
+  else
+    export CARGO_TARGET_DIR="$HAB_CARGO_TARGET_DIR"
+  fi
+  build_line "Setting CARGO_TARGET_DIR=$CARGO_TARGET_DIR"
+}
+
+do_prepare() {
+  _common_prepare
+
+  export rustc_target="x86_64-unknown-linux-musl"
+  build_line "Setting rustc_target=$rustc_target"
+
+  la_ldflags="-L$(pkg_path_for zlib-musl)/lib -lz"
+  la_ldflags="$la_ldflags -L$(pkg_path_for xz-musl)/lib -llzma"
+  la_ldflags="$la_ldflags -L$(pkg_path_for bzip2-musl)/lib -lbz2"
+  la_ldflags="$la_ldflags -L$(pkg_path_for openssl-musl)/lib -lssl -lcrypto"
+
+  export LIBARCHIVE_LIB_DIR=$(pkg_path_for libarchive-musl)/lib
+  export LIBARCHIVE_INCLUDE_DIR=$(pkg_path_for libarchive-musl)/include
+  export LIBARCHIVE_LDFLAGS="$la_ldflags"
+  export LIBARCHIVE_STATIC=true
+  export OPENSSL_LIB_DIR=$(pkg_path_for openssl-musl)/lib
+  export OPENSSL_INCLUDE_DIR=$(pkg_path_for openssl-musl)/include
+  export OPENSSL_STATIC=true
+  export SODIUM_LIB_DIR=$(pkg_path_for libsodium-musl)/lib
+  export SODIUM_STATIC=true
+
+  # Used to find libgcc_s.so.1 when compiling `build.rs` in dependencies. Since
+  # this used only at build time, we will use the version found in the gcc
+  # package proper--it won't find its way into the final binaries.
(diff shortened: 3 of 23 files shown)


thesentinels:

Thank you for contributing!

components/pkg-export-kubernetes/src/manifestjson.rs
```diff
@@ -0,0 +1,86 @@
...
+use manifest::Manifest;
+use error::Error;
+
+// Kubernetes manifest template
+const MANIFESTFILE: &'static str = include_str!("../defaults/KubernetesManifest.hbs");
+const BINDFILE: &'static str = include_str!("../defaults/KubernetesBind.hbs");
+
+pub struct ManifestJson {
+    pub main: Value,
+    pub binds: Vec<Value>,
```
christophermaier: Do these need to be public?

components/pkg-export-kubernetes/src/bind.rs
```diff
@@ -0,0 +1,56 @@
...
+        })
+    }
+}
+
+pub fn parse_bind_args(matches: &ArgMatches) -> Result<Vec<Bind>, Error> {
+    let mut binds = Vec::new();
+
+    if let Some(bind_args) = matches.values_of("BIND") {
+        for arg in bind_args {
+            let b = Bind::from_str(arg)?;
```
christophermaier: Maybe just use `arg.parse()?` instead?

components/pkg-export-kubernetes/src/bind.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::str::FromStr;
+
+use error::Error;
+
+#[derive(Debug, Clone)]
+pub struct Bind {
+    pub name: String,
+    pub service: String,
+    pub group: String,
+}
```
christophermaier: Naming question for you... in the interest of having `Bind` mean only one thing in the Habitat codebase, is there another more descriptive name we could use here (like `K8sBind`, `KubeBind`, etc.)?

I know we're in the Kubernetes exporter and all, but before I got to this commit of the PR, for a second I was wondering why you'd need to be using `core::metadata::Bind`.

components/pkg-export-kubernetes/src/manifest.rs
```diff
@@ -82,10 +82,7 @@ impl Manifest {
             None => pkg_ident.origin + "/" + &pkg_ident.name,
         };
 
-        let binds: Vec<String> = match matches.values_of("BIND") {
-            Some(binds) => binds.map(|s| s.to_string()).collect(),
-            None => Vec::new(),
-        };
+        let binds = bind::parse_bind_args(&matches)?;
```
christophermaier: I'm not sold on the utility of pushing CLAP awareness further down into the `Bind` type. Is there something additional you're shooting for by doing that?  Seems like using the parse method in here with the prior implementation would be a good separation of concerns.

components/pkg-export-helm/src/chartfile.rs
```diff
@@ -0,0 +1,53 @@
...
+use export_docker::Result;
+
+// Keep the default version in main::cli() in sync with this one
+pub const DEFAULT_VERSION: &'static str = "0.0.1";
+
+// Helm chart file template
+const CHARTFILE: &'static str = include_str!("../defaults/HelmChartFile.hbs");
+
+pub struct ChartFile {
+    json: Value,
```
christophermaier: What's the reason for using JSON types directly to store the information, rather than creating full types and serializing them out to JSON separately? Seems like you forego type safety this way and make the code a little harder to reason about.

components/pkg-export-helm/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+        valid_version("1.1.10".to_owned()).unwrap();
+        valid_version("1.10.1".to_owned()).unwrap();
+        valid_version("10.1.1".to_owned()).unwrap();
+
+        assert!(valid_version("1".to_owned()).is_err());
+        assert!(valid_version("1.1".to_owned()).is_err());
+        assert!(valid_version("X".to_owned()).is_err());
+        assert!(valid_version("1.2.Z".to_owned()).is_err());
+        assert!(valid_version("1.1.1.1".to_owned()).is_err());
+        assert!(valid_version("٣.7.৬".to_owned()).is_err());
```
christophermaier: Excellent test string 😄

components/pkg-export-helm/src/values.rs
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern crate habitat_pkg_export_docker as export_docker;
+
+use std::io::Write;
+
+use export_docker::Result;
+
+pub struct Values {
```
christophermaier: Naming and typing questions... I was initially thinking that this was another case of using JSON types directly.

What's the benefit of this type versus a plain `HashMap`? Is insertion order important?

components/hab/src/command/pkg/export/helm.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
christophermaier: 2018! 😄

components/pkg-export-helm/plan.sh
```diff
@@ -0,0 +1,80 @@
+pkg_name=hab-pkg-export-helm
```
christophermaier: Don't forget to add this package to `.bldr.toml` and to [this line](https://github.com/habitat-sh/habitat/blob/b4fa6df76d4c4f8f7f87e1898563eda52db16d0a/.travis.yml#L256) of `.travis.yml` to ensure that things are properly picked up in the builds

components/pkg-export-kubernetes/src/manifestjson.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl ManifestJson {
+    pub fn new(manifest: &Manifest) -> Self {
```
christophermaier: What's the advantage of implementing things this way, versus implementing JSON serialization on Manifest itself?

components/pkg-export-kubernetes/src/manifestjson.rs
zeenix: Yes, at least helm exporter needs to use them in chart.rs to split the manifest into a manifest template and values files.

zeenix: but I'm ok with adding simple getter methods if that's more appropriate.

components/hab/src/command/pkg/export/helm.rs
zeenix: The file was created in 2017 as you can see in the author date on the commit. :)

components/pkg-export-helm/plan.sh
zeenix: Sure.

components/pkg-export-helm/src/chartfile.rs
zeenix: Hmm.. sure. That would also mean new() not having to return Result.

components/pkg-export-helm/src/main.rs
zeenix: Yeah but the credit goes to @krnowak for that one. :)

components/pkg-export-helm/src/values.rs
zeenix: I actually went with HashMap approach at first but I wanted to have custom methods on it that makes it simpler to add entries and then also a method to generate the file. I could just have this as a container of HashMap and get rid of ValuesEntry private type below but I'm a bit unsure if that approach is better than the current very explicit one. It'd only save us 4 LOCs above.

components/pkg-export-kubernetes/src/bind.rs
zeenix: I think in Rust (like most modern languages with the concept of module hierarchy), there is very little for ambiguity and adding such prefixes seems very much redundant to me. If you are in the code of the k8s exporter itself, then the context is very much obvious if you only read/write "Bind (even if sometimes it takes a few seconds before you recall what the current context is :-)). If you are using this API from another crate, you either have to use the module prefix with the type or if it's ambiguous, you must create an alias (use x.y.Bind as XyBind).

A k8s or kubernetes prefix would also be a bit misleading in this case since the bind is a concept of habitat, not k8s.

FWIW, on the lower (C) level (if you ever needed to interface with C for some reason), this would also imply double prefixes and much longer identifier names than they need to be.

components/pkg-export-kubernetes/src/bind.rs
zeenix: sure

components/pkg-export-kubernetes/src/manifestjson.rs
zeenix: As the commit log says:

>     pkg-export-k8s: Add ManifestJson
>
>     Separate out JSON representations of the manifest into its own
>     struct/module. This is to allow users of the API to be able to modify the
>     JSON before writing it to a file.

This was done for the Helm exporter in question.Please refer to chart.rs in this PR, where I am manually generating this struct from the `Manifest`.

These question give me good ideas for my upcoming PR to add docs and tests in k8s exporter crate. :)

raskchanky:
@zeenix You mind squashing these commits?

zeenix:
@raskchanky which ones? The only ones I plan to squash and make sense to squash IMO are the last 3 additional ones that I'm going to squash just after @christophermaier gives me a green signal that he's done with reviews.

.bldr.toml
```diff
@@ -177,6 +177,14 @@ paths = [
   "components/pkg-export-docker/*"
 ]
 
+[hab-pkg-export-helm]
+plan_path = "components/pkg-export-helm"
+paths = [
+  "components/core/*",
+  "components/common/*",
+  "components/pkg-export-kubernetes/*"
```
christophermaier: I think we want to add `components/pkg-export-docker/*` here, too. If some file there changes, but nothing directly in `pkg-export-kubernetes`, then `pkg-export-kubernetes` will be built, but this won't.

Tracing up the dependency tree, it looks like we should also add `components/hab/*` for this and `[hab-pkg-export-docker]` as well.

components/pkg-export-kubernetes/src/bind.rs
```diff
@@ -29,7 +29,7 @@ impl FromStr for Bind {
 
     fn from_str(bind_spec: &str) -> Result<Self, Error> {
         let split: Vec<&str> = bind_spec.split(":").collect();
-        if split.len() != 3 {
+        if split.len() != 3 || split[0] == "" || split[1] == "" || split[2] == "" {
```
christophermaier: Nice catch.

components/pkg-export-kubernetes/src/manifestjson.rs
christophermaier: Public is fine.

christophermaier:
@zeenix one tweak to `.bldr.toml`, along with whatever squashes you want to make, then I think we're good to go.

components/pkg-export-kubernetes/src/bind.rs
zeenix: thanks, i caught this when I added a testcase for a later PR. :)

zeenix:
@christophermaier Done. :)

christophermaier:
@zeenix nice work :smile: 

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/328074036).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/328074036).

PR #4435, opened 2018-01-09
Group ID defaults to User ID in exported Docker containers

christophermaier:
Prior to the `--group-id` option (in #4310), the behavior was to have
the `SVC_GROUP`'s GID be the same as the `SVC_USER`'s UID, but that
was broken when the new option was added.

Now, the GID will be the same as the UID, unless specifically
overridden.


thesentinels:

Thank you for contributing!

PR #4448, opened 2018-01-17
Docs/test additions to k8s exporter & various fixes

zeenix:
The main goal of this PR is to add docs/test additions to k8s exporter but during our internal review at Kinvolk, we discovered quite a few issues that I fixed along the way. Since github doesn't have the concept of dependencies between PRs, I have put all the work in the same PR. I have however, kept each logical change in it's own commit, that should make it very easy to review if you review one commit at a time.


thesentinels:

Thank you for contributing!

components/pkg-export-kubernetes/src/bind.rs
```diff
@@ -27,6 +28,32 @@ pub struct Bind {
 impl FromStr for Bind {
     type Err = Error;
 
+    /// Create a `Bind` from a string.
+    ///
+    /// # Errors
+    ///
+    /// * `bind_spec` is not in the form `"name:service:group"`
```
christophermaier: If this is a Habitat bind, shouldn't it be `name:service.group`?

Also, if it is a Habitat bind, you might want to use the type from the Supervisor (like you do for `Topology`... there are other parts of a Bind that might get used which would break this.

(which might lend more weight to the "pull the common types up to a shared crate" argument below)

components/pkg-export-kubernetes/src/manifest.rs
```diff
@@ -24,7 +24,7 @@ use rand::Rng;
 
 use export_docker::Result;
 
-use topology::Topology;
+use habitat_sup::manager::service::Topology;
```
christophermaier: OK for now, but I wonder if longer-term we might want to pull this type up to a common / core crate instead of depending directly on the supervisor crate.

components/pkg-export-kubernetes/src/bind.rs
zeenix: Yeah we've had this typo for a bit now even though you pointed out in a previous review. Fortunately converting the code to make use of use `habitat_sup::manager::service::ServiceBind` will automatically solve the typo.

zeenix:
@christophermaier Thanks. I pushed the patch to ditch the custom `Bind` type here but I've now also a bunch of other fixes to k8s and docker in a branch on top of these changes. Do you want me to push them here as well so we deal with them all at once?

components/pkg-export-kubernetes/src/manifestjson.rs
```diff
@@ -59,8 +59,8 @@ impl ManifestJson {
         for bind in &manifest.binds {
             let json = json!({
                 "name": bind.name.clone(),
-                "service": bind.service.clone(),
-                "group": bind.group.clone(),
+                "service": bind.service_group.service().to_owned(),
+                "group": bind.service_group.group().to_owned(),
```
christophermaier: Does the bind need to be broken out into its constituent parts, or can it just be added to the JSON using its own string rendering?

There are other parts of a bind; this implementation is going to omit them entirely, and will also need to be updated when / if the parts of a bind change in the future. Admittedly, these other parts of a bind are not 100% fleshed out at the moment, but they are there.

christophermaier:
@zeenix as long as those other changes are thematically related to what's already here, go ahead and include them. Otherwise, they should be in their own PR.

components/pkg-export-kubernetes/src/manifestjson.rs
zeenix: > Does the bind need to be broken out into its constituent parts, or can it just be added to the JSON using its own string rendering?

The hHbitat k8s operator keeps them divided in their API so unless that changes, we have to have them separated in here too.

> There are other parts of a bind; this implementation is going to omit them entirely, and will also need to be updated when / if the parts of a bind change in the future. Admittedly, these other parts of a bind are not 100% fleshed out at the moment, but they are there.

That will also have to change in the habitat (k8s) operator side.

zeenix:
@christophermaier they are not exactly related except that they are mostly fixes to k8s, like some of the patches here already. No worries, I'll file another PR once this is merged.

christophermaier:
We can merge once the Habitat repository split is finalized (I don't anticipate any problems, but it's nice to reduce the amount of things in flight at once).

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/333303494).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/333303494).

PR #4457, opened 2018-01-18
add 2013 C runtime to hab binaries

mwrock:
We are currently unable to run the hab.exe binary on a clean windows machine because it now has a dep on the 2013 C runtime likely brought in by an upgraded crate.


thesentinels:

Thank you for contributing!

mwrock:
Tested running the artifacts in a clean machine which previously broke and it now successfully runs.

Issue #4172, opened 2017-12-01
Tar exporter: the next generation


tashimi:
The tar exporter is due for a re-write in the style of the new docker exporter. 

The resultant exporter should: 
- be written in rust
- follow the patterns of the docker exporter work by @fnichol 
- include `hab-launcher` in exports 
- follow a standard way across all new exporters as to where generated exports are stored (per #4104)

nellshamrell:
Also related (defines need for hab-launcher in exports) #4039

nellshamrell:
I think I've got a good handle on how the Docker exporter works now - details are here https://gist.github.com/nellshamrell/f38630f98c7b09d86607529e20208f45

nellshamrell:
This includes a abbreviated explanation for how the current generate tar exporter works:

https://gist.github.com/nellshamrell/e9163880ef79f99e1fb17b0b7af1a66e

nellshamrell:
Breakdown of work as I see it right now:

- [ ]  Create pkg-export-tar rust crate - should be BOTH a lib and a binary
- main.rs should call lib
- lib should include TarBuildRoot (modeled after DockerBuildRoot)
- lib should export new tarball (modeled after how the docker image is exported)
- [ ] Create hab/src/command/pkg/export/tar.rs - model after hab/src/command/pkg/export/docker.rs
- [ ] Wire hab pkg export tar to call hab/src/command/pkg/tar.rs

nellshamrell:
Managed to isolate the silent failures I was seeing when the builder keys for the package could not be accessed:

- unset local HAB_BLDR_URL (will use public builder as default)
- hab install core/test-app
- copy test-app artifact to directory for habitat/components/pkg-export-tar
- run $ cargo run test-app-artifact
- tars package, includes test-app in tarball

- set HAB_BLDR_URL to http://localhost:9636
- hab install core/nginx-cc-demo
- copy nginx-cc-demo to directory for habitat/components/pkg-export-tar
- run $ cargo run nginx-cc-demo artifact
- tars package, includes nginx-cc-demo in tarball

- make sure local builder is running
- set HAB_BLDR_URL to http://localhost:9636
- hab install core/scaffolding-node
- copy scaffolding-node to directory for habitat/components/pkg-export-tar
- run $ cargo run scaffolding-node-artifact
- tars package, includes scaffolding-node

- unset local HAB_BLDR_URL (will use public builder as default)
[...]

PR #4468, opened 2018-01-20
Tar Exporter - The Next Generation

nellshamrell:
This is a partial fix for #4172, rewriting the tar exporter in Rust.  #4172 also requests some new functionality.  I prefer to separate refactoring and new behavior so this PR just focuses on re-writing the exporter in Rust.  I will follow this one up with a separate PR that adds the new behavior.

### DIFF
--- a/components/pkg-export-tar/Cargo.toml
+++ b/components/pkg-export-tar/Cargo.toml
@@ -1,23 +1,20 @@
 [package]
-name = "habitat_pkg_export_docker"
+name = "habitat_pkg_export_tar"
 version = "0.0.0"
-authors = ["Fletcher Nichol <fnichol@habitat.sh>"]
-build = "build.rs"
+authors = ["Nell Shamrell <nellshamrell@gmail.com>"]
 workspace = "../../"
 
 [lib]
-name = "habitat_pkg_export_docker"
+name = "habitat_pkg_export_tar"
 
-[[bin]]
-name = "hab-pkg-export-docker"
-path = "src/main.rs"
-doc = false
+[[bin]]                                                                
+name = "hab-pkg-export-tar"                                        
+path = "src/main.rs"                                  
+doc = false  
 
 [dependencies]
 base64 = "*"
 clap = { version = "*", features = ["suggestions", "color", "unstable"] }
-clippy = { version = "*", optional = true }
-chrono = "*"
 env_logger = "*"
 hab = { path = "../hab" }
 habitat_common = { path = "../common" }
@@ -25,16 +22,11 @@ habitat_core = { git = "https://github.com/habitat-sh/core.git" }
 handlebars = { version = "*", default-features = false }
 lazy_static = "*"
 log = "*"
-rusoto_core = "*"
-rusoto_credential = "*"
-rusoto_ecr = "*"
-serde = { version = "*", features = ["rc"] }
-serde_json = "*"
-tempdir = "*"
+mktemp = "*"
+serde = { version = "1.0", features = ["rc"] }
+serde_json = "1.0"
 url = "*"
 failure = { git = "https://github.com/withoutboats/failure.git" }
 failure_derive = { git = "https://github.com/withoutboats/failure_derive.git" }
-
-[features]
-default = []
-functional = []
+tempdir = "*"
+tar = "*"
--- a/components/pkg-export-tar/build.rs
+++ b/components/pkg-export-tar/build.rs
@@ -1,0 +1,6 @@
+// Inline common build behavior
+include!("../libbuild.rs");
+
+fn main() {
+    habitat::common();
+}
--- a/components/pkg-export-tar/plan.sh
+++ b/components/pkg-export-tar/plan.sh
@@ -1,0 +1,79 @@
+pkg_name=hab-pkg-export-tar
+_pkg_distname=$pkg_name
+pkg_origin=core
+pkg_version=$(cat "$PLAN_CONTEXT/../../VERSION")
+pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+pkg_license=('Apache-2.0')
+pkg_deps=(core/coreutils core/findutils core/gawk core/grep core/bash core/tar core/gzip core/hab)
+pkg_build_deps=(
+  core/musl core/zlib-musl core/xz-musl core/bzip2-musl core/libarchive-musl
+  core/openssl-musl core/libsodium-musl
+  core/coreutils core/rust core/gcc core/make
+)
+
+pkg_bin_dirs=(bin)
+
+bin=$_pkg_distname
+
+_common_prepare() {
+  do_default_prepare
+
+  # Can be either `--release` or `--debug` to determine cargo build strategy
+  build_type="--release"
+  build_line "Building artifacts with \`${build_type#--}' mode"
+
+  # Used by the `build.rs` program to set the version of the binaries
+  export PLAN_VERSION="${pkg_version}/${pkg_release}"
+  build_line "Setting PLAN_VERSION=$PLAN_VERSION"
+
+  if [ -z "$HAB_CARGO_TARGET_DIR" ]; then
+    # Used by Cargo to use a pristine, isolated directory for all compilation
+    export CARGO_TARGET_DIR="$HAB_CACHE_SRC_PATH/$pkg_dirname"
+  else
+    export CARGO_TARGET_DIR="$HAB_CARGO_TARGET_DIR"
+  fi
+  build_line "Setting CARGO_TARGET_DIR=$CARGO_TARGET_DIR"
+}
+
+do_prepare() {
+  _common_prepare
+
+  export rustc_target="x86_64-unknown-linux-musl"
+  build_line "Setting rustc_target=$rustc_target"
+
+  la_ldflags="-L$(pkg_path_for zlib-musl)/lib -lz"
+  la_ldflags="$la_ldflags -L$(pkg_path_for xz-musl)/lib -llzma"
+  la_ldflags="$la_ldflags -L$(pkg_path_for bzip2-musl)/lib -lbz2"
+  la_ldflags="$la_ldflags -L$(pkg_path_for openssl-musl)/lib -lssl -lcrypto"
+
+  export LIBARCHIVE_LIB_DIR=$(pkg_path_for libarchive-musl)/lib
+  export LIBARCHIVE_INCLUDE_DIR=$(pkg_path_for libarchive-musl)/include
+  export LIBARCHIVE_LDFLAGS="$la_ldflags"
+  export LIBARCHIVE_STATIC=true
+  export OPENSSL_LIB_DIR=$(pkg_path_for openssl-musl)/lib
+  export OPENSSL_INCLUDE_DIR=$(pkg_path_for openssl-musl)/include
+  export OPENSSL_STATIC=true
+  export SODIUM_LIB_DIR=$(pkg_path_for libsodium-musl)/lib
+  export SODIUM_STATIC=true
+
+  # Used to find libgcc_s.so.1 when compiling `build.rs` in dependencies. Since
+  # this used only at build time, we will use the version found in the gcc
+  # package proper--it won't find its way into the final binaries.
+  export LD_LIBRARY_PATH=$(pkg_path_for gcc)/lib
+  build_line "Setting LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+
+  PLAN_TAR_PKG_IDENT=$(pkg_path_for tar | sed "s,^$HAB_PKG_PATH/,,")
+  export PLAN_TAR_PKG_IDENT
+  build_line "Setting PLAN_TAR_PKG_IDENT=$PLAN_TAR_PKG_IDENT"
+}
+
+do_build() {
+  pushd $PLAN_CONTEXT > /dev/null
+  cargo build ${build_type#--debug} --target=$rustc_target --verbose
+  popd > /dev/null
+}
+
+do_install() {
+  install -v -D $CARGO_TARGET_DIR/$rustc_target/${build_type#--}/$bin \
+    $pkg_prefix/bin/$bin
+}
--- a/components/pkg-export-tar/src/build.rs
+++ b/components/pkg-export-tar/src/build.rs
@@ -1,0 +1,254 @@
+// Copyright (c) 2016-2018 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fs as stdfs;
+#[cfg(target_os = "linux")]
+use std::os::unix::fs::symlink;
+#[cfg(target_os = "windows")]
+use std::os::windows::fs::symlink_dir as symlink;
+use clap;
+use common;
+use common::command::package::install::InstallSource;
+use common::ui::{UI, Status};
+use tempdir::TempDir;
+use std::path::Path;
+use hcore::package::PackageIdent;
+use error::Result;
+use hcore::PROGRAM_NAME;
+use hcore::fs::{CACHE_ARTIFACT_PATH, CACHE_KEY_PATH, cache_artifact_path, cache_key_path};
+
+use super::{VERSION, BUSYBOX_IDENT};
+
+use rootfs;
+
+// Much of this functionality is duplicated (or slightly modified)
+// from the Docker exporter. This needs to be abstacted out in
+// the future for use with further exporters.
+// https://github.com/habitat-sh/habitat/issues/4522
+const DEFAULT_HAB_IDENT: &'static str = "core/hab";
+const DEFAULT_LAUNCHER_IDENT: &'static str = "core/hab-launcher";
+const DEFAULT_SUP_IDENT: &'static str = "core/hab-sup";
+
+/// The specification for creating a temporary file system build root, based on Habitat packages.
+///
+/// When a `BuildSpec` is created, a `BuildRoot` is returned which can be used to produce exported
+/// images, archives, etc.
+#[derive(Debug)]
+pub struct BuildSpec<'a> {
+    /// A string representation of a Habitat Package Identifer for the Habitat CLI package.
+    pub hab: &'a str,
+    /// A string representation of a Habitat Package Identifer for the Habitat Launcher package.
+    pub hab_launcher: &'a str,
+    /// A string representation of a Habitat Package Identifer for the Habitat Supervisor package.
+    pub hab_sup: &'a str,
+    /// The Builder URL which is used to install all service and extra Habitat packages.
+    pub url: &'a str,
+    /// The Habitat release channel which is used to install all service and extra Habitat
+    /// packages.
+    pub channel: &'a str,
+    /// The Builder URL which is used to install all base Habitat packages.
+    pub base_pkgs_url: &'a str,
+    /// The Habitat release channel which is used to install all base Habitat packages.
+    pub base_pkgs_channel: &'a str,
+    /// A Habitat Package Identifer or local path to a Habitat Artifact file which
+    /// will be installed.
+    pub ident_or_archive: &'a str,
+}
+
+impl<'a> BuildSpec<'a> {
+    /// Creates a `BuildSpec` from cli arguments.
+    pub fn new_from_cli_matches(
+        m: &'a clap::ArgMatches,
+        default_channel: &'a str,
+        default_url: &'a str,
+    ) -> Self {
+
+        BuildSpec {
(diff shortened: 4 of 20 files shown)


thesentinels:

Thank you for contributing!

nellshamrell:
Note for deploying - we will need to build the hab-pkg-export-tar package and upload it to the core-origin on the public builder.

christophermaier:
@nellshamrell You'll want to add an entry for this package in `.bldr.toml`, and the [_RUST_HAB_BIN_COMPONENTS](https://github.com/habitat-sh/habitat/blob/6678d9a07aa29b4ccb2d375333a557c3a9533e4e/.travis.yml#L15) and [PACKAGES](https://github.com/habitat-sh/habitat/blob/6678d9a07aa29b4ccb2d375333a557c3a9533e4e/.travis.yml#L256) variables, in `.travis.yml`, to ensure this code gets built, tested, packaged, and released along with everything else.

components/pkg-export-tar/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+path = "src/main.rs"                                  
+doc = false  
+
+[dependencies]
+base64 = "*"
+clap = { version = "*", features = ["suggestions", "color", "unstable"] }
+env_logger = "*"
+hab = { path = "../hab" }
+habitat_common = { path = "../common" }
+habitat_core = { path = "../core" }
```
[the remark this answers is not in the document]
nellshamrell: 👍

components/pkg-export-tar/plan.sh
```diff
@@ -0,0 +1,87 @@
...
+_pkg_distname=$pkg_name
+pkg_origin=core
+pkg_version=$(cat "$PLAN_CONTEXT/../../VERSION")
+pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+pkg_license=('Apache-2.0')
+pkg_deps=(core/coreutils core/findutils core/gawk core/grep core/bash core/tar core/gzip core/hab)
+pkg_build_deps=(
+  core/musl core/zlib-musl core/xz-musl core/bzip2-musl core/libarchive-musl
+  core/openssl-musl core/libsodium-musl
+  core/coreutils core/rust core/gcc core/make
```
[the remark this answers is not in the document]
nellshamrell: musl are needed, alas.

components/pkg-export-tar/src/build.rs
```diff
@@ -0,0 +1,928 @@
...
+
+/// An enum of service and library Habitat packages.
+///
+/// A package is considered a service package if it contains a runnable service, via a `run` hook.
+#[derive(Debug)]
+enum PkgIdentType {
+    /// A service package which contains a runnable service.
+    Svc(SvcIdent),
+    /// A library package which does not contain a runnable service.
+    Lib(PackageIdent),
```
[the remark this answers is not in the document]
nellshamrell: The original ticket (#4172) requested the addition of the hab-launcher, but I wanted to get the rewrite done, replicating all the old functionality, and merged before adding new functionality.

nellshamrell:
Ty so much for all the feedback, @christophermaier!  I think I've yanked out all the Docker specific cruft at this time, will look at refactoring this to use the BuildRoot, tar-rs, and more to make this more platform agnostic.

components/pkg-export-tar/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+fn tar_name(hart_to_package: &str) -> String {
+    let path = Path::new(hart_to_package);
+    if path.is_file() {
+        let ident = PackageArchive::new(path).ident().unwrap();
+        format_tar_name(ident)
+    } else {
+        let pkg_path_command = Command::new("hab")
```
christophermaier: In this branch, we're presumably dealing with a package identifier. We've actually got code to handle the resolution of a full identifier from a partial one in our core install logic. It will pull down the package in question from Builder if it doesn't already exist in the cache, which the current studio-based implementation in this PR does not do, but it does line up with the behavior in the rest of the project.

Also, just from a separation-of-concerns viewpoint, this function is doing several things: figuring out whether we're dealing with a hart file or an identifier, figuring out what full identifier to use, and then formatting that identifier into a tar file name. 

There's also a hidden, assumed behavior based on how it's called, where `hart_to_package` could actually be a string of multiple identifiers to resolve, but I don't think this code actually handles that behavior. 

And actually, now that I type that, any "multiple" package export would only be named for a single package (presumably the first one), which is potentially confusing... do we really want to take more than one hart file or identifier as an argument to this exporter? (This thought actually applies to our other exporters as well, because I think they have the same assumptions.) @fnichol do you have any thoughts on this point?

christophermaier:
@nellshamrell It'd also be good to add docs on the new environment variables to https://github.com/habitat-sh/habitat/blob/master/www/source/partials/docs/_reference-environment-vars.html.md.erb

components/pkg-export-tar/src/lib.rs
nellshamrell: I had not thought to try it with more than one hart file or identifier.  Does the current tar exporter allow for this?  @fnichol would love your thoughts as well :)

components/pkg-export-tar/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        println!("Please see the above error for details.");
+    }
+}
+
+fn tar_command(temp_dir_path: &PathBuf, hart_to_package: &str) {
+    let tarball = File::create(tar_name(&hart_to_package)).unwrap();
+    let mut tar_builder = Builder::new(tarball);
+
+    let mut hab_pkgs_buf = temp_dir_path.clone();
+    hab_pkgs_buf.push("hab/pkgs");
```
christophermaier: Does this work properly on Windows, or would something like hab_pkgs_buf.push("hab").push("pkgs") be more cross-platform?

As an aside, a lot of times when writing functions to deal with file paths in Rust, it's customary to accept an AsRef<Path> instead of something more concrete like a Path or PathBuf. Not only is that more flexible in terms of what inputs the function can accept, but it would let you do things like tmp_dir_path.as_ref().join("hab").join("pkgs") which gets you a PathBuf in the end, in a cross-platform way, and lets you dispense with the extra tmp_dir_path.clone() bit here (that happens implicitly in the join call).

components/pkg-export-tar/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    tar_builder.append_dir_all("hab/pkgs", hab_pkgs_buf);
```
christophermaier: How does this tar crate deal with Unix vs. Windows paths? Does it do any internal translation of path separators?

mwrock: Windows can handle `/` in paths. So this should work fine. The times that causes problems are when application logic does not expect `/` which is rare and likely not the case here. That said, we should for sure give this a test run on a windows package on windows for sanity.

christophermaier: @mwrock Good to know... sounds like I need to update my Windows knowledge from the Windows98 era 😄

components/pkg-export-tar/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        let ident_regex = Regex::new(r"[\w-]+/([\w-]+/[\d/.]+/\d+$)").unwrap();
+        let ident_captures = ident_regex.captures(edited_pkg_path).unwrap();
+        let ident_string = &ident_captures[0];
+
+        let ident = PackageIdent::from_str(ident_string).unwrap();
+        format_tar_name(ident)
+    }
+
+}
+
```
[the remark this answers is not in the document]
nellshamrell: this has been fixed!

fnichol:
I'll give this a spin this afternoon barring anything big. Nice work!

components/pkg-export-tar/build.rs
```diff
@@ -0,0 +1,20 @@
...
+    habitat::common();
+    write_tar_pkg_ident();
+}
+
+fn write_tar_pkg_ident() {
+    let ident = match env::var("PLAN_TAR_PKG_IDENT") {
+        // Use the value provided by the build system if present
+        Ok(ident) => ident,
+        // Use the latest installed package as a default for development
+        _ => String::from("core/tar"),
```
christophermaier: Is this still needed if we're using `tar-rs`?

components/pkg-export-tar/src/build.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// The Habitat release channel which is used to install all service and extra Habitat
+    /// packages.
+    pub channel: &'a str,
+    /// The Builder URL which is used to install all base Habitat packages.
+    pub base_pkgs_url: &'a str,
+    /// The Habitat release channel which is used to install all base Habitat packages.
+    pub base_pkgs_channel: &'a str,
+    /// A list of either Habitat Package Identifiers or local paths to Habitat Artifact files which
+    /// will be installed.
+    pub idents_or_archives: Vec<&'a str>,
```
christophermaier: If multiple artifacts are not supported, we should make the name of this field singular, and remove the `Vec` from the type, just to make it super clear.

components/pkg-export-tar/src/build.rs
```diff
@@ -0,0 +1,262 @@
...
+            launcher,
+            busybox,
+        })
+    }
+
+    fn install_user_pkgs<P: AsRef<Path>>(
+        &self,
+        ui: &mut UI,
+        rootfs: P,
+    ) -> Result<Vec<PackageIdent>> {
```
christophermaier: Ditto w/r/t taking a single package for export.

components/pkg-export-tar/src/build.rs
```diff
@@ -0,0 +1,262 @@
...
+use hcore::fs::{CACHE_ARTIFACT_PATH, CACHE_KEY_PATH, cache_artifact_path, cache_key_path};
+
+use super::{VERSION, BUSYBOX_IDENT};
+
+use rootfs;
+
+const DEFAULT_HAB_IDENT: &'static str = "core/hab";
+const DEFAULT_LAUNCHER_IDENT: &'static str = "core/hab-launcher";
+const DEFAULT_SUP_IDENT: &'static str = "core/hab-sup";
+
```
christophermaier: It's probably worth adding some comments here about the intended future consolidation and finalization of the `BuildRoot` abstraction across exporters, so future devs have some more context on what otherwise looks like plain duplication of this code from the Docker exporter.

A similar note over in the Docker exporter (even if only to point to the comment here) could also be helpful, just to tie everything together.

components/pkg-export-tar/src/build.rs
```diff
@@ -0,0 +1,262 @@
...
+#[derive(Debug)]
+struct BasePkgIdents {
+    /// Installed package identifer for the Habitat CLI package.
+    pub hab: PackageIdent,
+    /// Installed package identifer for the Supervisor package.
+    pub sup: PackageIdent,
+    /// Installed package identifer for the Launcher package.
+    pub launcher: PackageIdent,
+    /// Installed package identifer for the Busybox package.
+    pub busybox: Option<PackageIdent>,
```
christophermaier: Why is this an `Option`?

components/pkg-export-tar/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+    export_tar::export_for_cli_matches(ui, &m)
+}
+
+fn cli<'a, 'b>() -> App<'a, 'b> {
+    let name: &str = &*PROGRAM_NAME;
+    let about = "Creates a tar package from a Habitat package";
+    Cli::new(name, about)
+        .add_base_packages_args()
+        .add_builder_args()
+        .add_pkg_ident_arg(PkgIdentArgOptions { multiple: true })
```
christophermaier: This should probably be `multiple: false`, or we should just not use the `PkgIdentArgOptions` pattern from the Docker exporter.

components/pkg-export-tar/src/rootfs.rs
```diff
@@ -0,0 +1,100 @@
...
+    fs::create_dir_all(root.join("etc"))?;
+
+    fs::create_dir_all(root.join("root"))?;
+    filesystem::chmod(root.join("root").to_string_lossy().as_ref(), 0o0750)?;
+    fs::create_dir_all(root.join("tmp"))?;
+    filesystem::chmod(root.join("tmp").to_string_lossy().as_ref(), 0o1777)?;
+    fs::create_dir_all(root.join("var/tmp"))?;
+    filesystem::chmod(root.join("var/tmp").to_string_lossy().as_ref(), 0o1777)?;
+
+    Ok(())
```
christophermaier: I don't think we need `/bin`, `/etc`, `/tmp`, or `/var` in a tar export, assuming that a user will be expanding the tarball onto an existing Linux system, rather than having to "carry one around with us" in a container scenario.

(In a future with a finalized `BuildRoot` abstraction, the Docker exporter would probably implement some kind of `BuildRoot` trait that specializes this behavior, whereas the tar exporter would probably just use a default no-op implementation.)

components/pkg-export-tar/src/cli.rs
```diff
@@ -0,0 +1,153 @@
...
+
+pub struct PkgIdentArgOptions {
+    pub multiple: bool,
+}
+
+impl<'a, 'b> Cli<'a, 'b> {
+    pub fn new(name: &str, about: &'a str) -> Self {
+        Cli {
+            app: clap_app!(
+            (name) =>
```
elliott-davis: Is it possible to use the same Clap syntax as below when initializing the `clap_app`?

components/pkg-export-tar/plan.sh
```diff
@@ -0,0 +1,87 @@
...
+  popd > /dev/null
+}
+
+do_install() {
+  install -v -D $CARGO_TARGET_DIR/$rustc_target/${build_type#--}/$bin \
+    $pkg_prefix/bin/$bin
+}
+
+do_strip() {
+  if [[ "$build_type" != "--debug" ]]; then
```
elliott-davis: Can we ever get here given: https://github.com/habitat-sh/habitat/pull/4468/files#diff-d606dc6e0f433bdd6d4a44b053202e31R22

components/pkg-export-tar/plan.sh
```diff
@@ -0,0 +1,87 @@
...
+  if [ -z "$HAB_CARGO_TARGET_DIR" ]; then
+    # Used by Cargo to use a pristine, isolated directory for all compilation
+    export CARGO_TARGET_DIR="$HAB_CACHE_SRC_PATH/$pkg_dirname"
+  else
+    export CARGO_TARGET_DIR="$HAB_CARGO_TARGET_DIR"
+  fi
+  build_line "Setting CARGO_TARGET_DIR=$CARGO_TARGET_DIR"
+}
+
+do_prepare() {
```
elliott-davis: Most of this code is shared across all of the Rust exporters. I think it might be a good opportunity to abstract it. It would probably look pretty similar to https://github.com/habitat-sh/builder/blob/master/support/ci/builder-base-plan.sh

nellshamrell:
Workflow with the "old" tar exporter:

1) Export hart package as tarball
2) Upload tarball to VM
3) Unpack tarball to /
4) Add a service file for hab sup run with peers etc. (Service file for systemd)
5) Run hab sup load for the service.

components/pkg-export-tar/src/build.rs
nellshamrell: Agreed!  Just created an issue around refactoring this duplicate logic into it's own crate https://github.com/habitat-sh/habitat/issues/4522, will add some comment notes with links to this issue to both the tar and docker exporters.

components/pkg-export-tar/plan.sh
nellshamrell: Good point!  Due to the time pressure of what we need to do for the on-prem builder in the next couple of months, I'd like to delay doing this abstraction for now.  I have created an issue around abstracting this logic here https://github.com/habitat-sh/habitat/issues/4522

components/pkg-export-tar/src/build.rs
nellshamrell: Fixed!

components/pkg-export-tar/src/build.rs
nellshamrell: Fixed!

components/pkg-export-tar/src/main.rs
nellshamrell: Fixed!

components/pkg-export-tar/src/rootfs.rs
nellshamrell: Fixed!

components/pkg-export-tar/plan.sh
nellshamrell: Good point, removed it.

components/pkg-export-tar/src/cli.rs
nellshamrell: I'm not sure what you mean by that, could you expand a bit?

components/pkg-export-tar/plan.sh
christophermaier: Indeed! I stumbled across this myself, resulting in https://github.com/habitat-sh/habitat/pull/4524

@nellshamrell I agree with you 100% on pushing the abstraction off a bit... I was tempted to do it myself, but we have bigger fish to fry at the moment.

components/pkg-export-tar/src/lib.rs
```diff
@@ -53,17 +53,17 @@ pub fn export_for_cli_matches(ui: &mut UI, matches: &clap::ArgMatches) -> Result
...
     Ok(())
 }
 
 #[allow(unused_must_use)]
-fn tar_command(temp_dir_path: &Path, builder_idents: Vec<PackageIdent>) {
+fn tar_command(temp_dir_path: &Path, pkg_ident: PackageIdent) {
     // We currently support only exporting one tar package at a time
     // So we grab the first value from the Vec of idents
-    let pkg_ident = builder_idents[0].clone();
+    let pkg_ident = pkg_ident.clone();
```
christophermaier: I don't think you need this line at all anymore.

Also, the comment immediately above no longer applies.

components/pkg-export-tar/src/build.rs
christophermaier: Never mind... I see the platform-dependence of this above.

components/pkg-export-tar/src/cli.rs
```diff
@@ -0,0 +1,153 @@
...
+        } else {
+            "A Habitat package identifier (ex: acme/redis) and/or filepath to a Habitat Artifact \
+            (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)"
+        };
+
+        let app = self.app.arg(
+            Arg::with_name("PKG_IDENT_OR_ARTIFACT")
+                .value_name("PKG_IDENT_OR_ARTIFACT")
+                .required(true)
+                .multiple(options.multiple)
```
christophermaier: I think we could probably just remove this `.multiple` call, as well as the copy of `PkgIdentArgOptions` in this crate, in the interest of keeping everything here as simple and focused as possible.

components/pkg-export-tar/src/rootfs.rs
```diff
@@ -0,0 +1,92 @@
...
+/// # Errors
+///
+/// * If files and/or directories cannot be created
+/// * If permissions for files and/or directories cannot be set
+pub fn create<T>(root: T) -> Result<()>
+where
+    T: AsRef<Path>,
+{
+    let root = root.as_ref();
+    fs::create_dir_all(root.join("root"))?;
```
christophermaier: Do we still need to create a `/root` directory?

Also, I think most all the tests below will fail, since they're still expecting a lot of extra Linux OS stuff.

nellshamrell: Fixed!

components/pkg-export-tar/src/cli.rs
nellshamrell: Fixed!

components/pkg-export-tar/src/cli.rs
```diff
@@ -130,7 +130,6 @@ impl<'a, 'b> Cli<'a, 'b> {
             Arg::with_name("PKG_IDENT_OR_ARTIFACT")
                 .value_name("PKG_IDENT_OR_ARTIFACT")
                 .required(true)
-                .multiple(options.multiple)
```
christophermaier: We don't need the `PkgIdentArgOptions` argument to this function, either.

christophermaier:
tenor-230497150

components/pkg-export-tar/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    tar_command(builder_dir_path, pkg_ident, hab_pkg);
+    Ok(())
+}
+
+#[allow(unused_must_use)]
+fn tar_command(temp_dir_path: &Path, pkg_ident: PackageIdent, hab_pkg: &str) {
+    // We currently support only exporting one tar package at a time
+    // So we grab the first value from the Vec of idents
+    let pkg_ident = pkg_ident.clone();
```
christophermaier: I don't think you need this line at all anymore.

Also, the comment immediately above no longer applies.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/337657272).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/337657272).

PR #4480, opened 2018-01-23
Simplify port handling in butterfly server

krnowak:
The butterfly server allowed to specify a separate port number for
swim connections and a separate port for gossip connections. The only
serious user of the butterfly server, the supervisor, was passing the
same number in for both. It works, because even if butterfly is still
opening two ports, one is a UDP port and the other is TCP port.

Another reason for this change is to simplify the discovery of the
exposed address and port for the supervisor being behind a NAT. This
feature may come in future work revolving around the spanning ring.

Signed-off-by: Krzesimir Nowak <krzesimir@kinvolk.io>

### DIFF
--- a/components/butterfly-test/src/lib.rs
+++ b/components/butterfly-test/src/lib.rs
@@ -54,16 +54,13 @@ impl Suitability for NSuitability {
 
 pub fn start_server(name: &str, ring_key: Option<SymKey>, suitability: u64) -> Server {
     SERVER_PORT.compare_and_swap(0, 6666, Ordering::Relaxed);
-    let swim_port = SERVER_PORT.fetch_add(1, Ordering::Relaxed);
-    let gossip_port = SERVER_PORT.fetch_add(1, Ordering::Relaxed);
-    let listen_swim = format!("127.0.0.1:{}", swim_port);
-    let listen_gossip = format!("127.0.0.1:{}", gossip_port);
+    let swim_gossip_port = SERVER_PORT.fetch_add(1, Ordering::Relaxed);
+    let listen_swim_gossip = format!("127.0.0.1:{}", swim_gossip_port);
     let mut member = Member::default();
-    member.set_swim_port(swim_port as i32);
-    member.set_gossip_port(gossip_port as i32);
+    member.set_swim_port(swim_gossip_port as i32);
+    member.set_gossip_port(swim_gossip_port as i32);
     let mut server = Server::new(
-        &listen_swim[..],
-        &listen_gossip[..],
+        &listen_swim_gossip[..],
         member,
         Trace::default(),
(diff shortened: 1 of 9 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
Looks good to me, though I'm interested if @reset and/or @adamhjk have any objections: would this violate any hidden assumptions or fundamental design constraints that you envisioned earlier in the life of the project?

reset:
@krnowak I don't see how removing the port flexibility from Butterfly will simplify the discovery of an exposed port of a Supervisor behind a NAT. What were your thoughts there?

krnowak:
First thing to note is that the discovery is about Supervisor behind a NAT learning about the address under which it is exposed outside.

Let's assume we start a Supervisor A behind a NAT and Supervisor B outside the NAT. A has an IP AN_IP behind the NAT. B's IP address is not important here. B is told to join the ring by talking to A, but B can't talk to A by using AN_IP - A needs to be exposed first. Let's assume we did that, so the address outside NAT is AX_IP. A opens two ports: SWIM_PORT and GOSSIP_PORT. Let's assume that we expose the same port numbers outside the NAT, so AX_IP:SWIM_PORT gets forwarded to AN_IP:SWIM_PORT, and the same goes for GOSSIP_PORT.

The idea behind the discovery was that A does not need to know about AX_IP and the exposed ports - it can learn about it when B starts talking to it. The talking starts with sending a Ping message. I was thinking about adding a "destination" field to the message, from which A could learn the address that was used to talk to it (AX_IP:SWIM_PORT).

Now the problem is that that way A discovers it's exposed address for SWIM communication, but not for gossip communication. So it has discovered the exposed communication channels only partially. And usually the Ack message contains both SWIM and gossip addresses to talk to (inside a Member struct). So it leads to a bit of a chicken and egg problem - B normally should learn about A's gossip address from A's Ack message, but A could discover the gossip address through some (gossip?) message it receives from B.

Granted, this is a hypothetical problem - from Supervisor's point of view, both SWIM_PORT and GOSSIP_PORT are always the same. But butterfly code has no such assumption and it would be the code of butterfly that does the discovery and this is where this assumption would need to be coded for the discovery to work (so A is never in state that it knows only exposed SWIM address, but not exposed gossip address, so it does not send half-filled Ack messages).

So my idea was to just use a single port number, since the only real user of butterfly was already assuming this.

Another idea could be to have the "destination" field in a Ping message to contain AX_IP, SWIM_PORT and GOSSIP_PORT. And, to be sincere, this occurred to me while writing this post. ;)

christophermaier:
Closing this for now, pending the coming Supervisor / Butterfly rewrite. We can revisit after that lands.

PR #4497, opened 2018-01-25
Introduce network abstraction for butterfly

krnowak:
This PR adds a network abstraction to be used by butterfly server and client. The reason for introducing it is to be able to add a different implementation of the abstraction for testing purposes. That testing implementation should make mocking some scenarios way easier than it is currently possible. One of those scenarios would be talking to the butterfly server which is behind a NAT.

The PR also contains some cleanups I had to make for rustfmt to stop complaining.

### DIFF
--- a/components/butterfly/src/network.rs
+++ b/components/butterfly/src/network.rs
@@ -1,0 +1,80 @@
+// Copyright (c) 2018 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! The Butterfly network abstraction.
+//!
+//! The abstraction provides communication channels for sending SWIM
+//! and gossip messages.
+
+use std::fmt::Debug;
+use std::marker::Send;
+use std::net::SocketAddr;
+
+use error::Result;
+
+// TODO(krnowak): See a TODO about Debug for Network trait below.
+/// A trait for types used for sending SWIM messages.
+pub trait SwimSender: Send + Debug {
+    /// Send a SWIM message (as bytes) to the given address. The
+    /// returned value holds a number of bytes sent.
+    fn send(&self, buf: &[u8], addr: SocketAddr) -> Result<usize>;
+}
+
+/// A trait for types used for receiving SWIM messages.
+pub trait SwimReceiver: Send {
+    /// Receive a SWIM message (as bytes) from the channel. The
+    /// returned value holds the size and an address from where the
+    /// bytes came.
+    fn receive(&self, buf: &mut [u8]) -> Result<(usize, SocketAddr)>;
+}
+
+/// A trait for types used for sending gossip messages (rumors).
+pub trait GossipSender {
+    /// Send a rumor (as bytes).
+    fn send(&self, buf: &[u8]) -> Result<()>;
+}
+
+/// A trait for types used for receiving gossip messages (rumors).
+pub trait GossipReceiver {
+    /// Receive a rumor (as bytes).
+    fn receive(&self) -> Result<Vec<u8>>;
+}
+
+// TODO(krnowak): Not sure if this static lifetime specifier here is a
+// correct thing to do. It is either here on in several other places
+// where generic type N constrained to being an implementation of the
+// Network trait is used (trace, expire, inbound, outbound and so
+// on). I added it here, because Network is exclusively used by the
+// butterfly component.
+//
+// Same for Debug - Network is used by the butterfly component only so
+// I add it here to save me some typing.
+/// A trait for types used to provide SWIM and gossip communication
+/// channels.
+pub trait Network: Send + Sync + Debug + 'static {
+    type SwimSender: SwimSender;
+    type SwimReceiver: SwimReceiver;
+    type GossipSender: GossipSender;
+    type GossipReceiver: GossipReceiver;
+
+    fn get_swim_addr(&self) -> SocketAddr;
+    fn get_swim_port(&self) -> u16;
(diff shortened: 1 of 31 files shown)


thesentinels:

Thank you for contributing!

krnowak:
The failure on appveyor seems to be unrelated to my change:
```
'"C:\hab\pkgs\core\hab-studio\0.53.0-dev\20180125090333\bin\powershell/powershell.exe"' is not recognized as an internal or external command,
operable program or batch file.
Command exited with code 1
```

Looks like the final slash in the powershell path should be a backslash.

mwrock:
The forward slash is ok. The powershell binary name has changed in the latest package. If you rebase against latest master then that should fix this.

krnowak:
Let's see. Rebased.

components/butterfly/src/network.rs
```diff
@@ -0,0 +1,82 @@
...
+/// A trait for types used to provide SWIM and gossip communication
+/// channels.
+pub trait Network: Send + Sync + Debug + 'static {
+    type SwimChannel: SwimChannel;
+    type GossipSender: GossipSender;
+    type GossipReceiver: GossipReceiver;
+
+    fn get_swim_addr(&self) -> SocketAddr;
+    fn get_swim_port(&self) -> u16;
+    fn get_swim_channel(&self) -> Result<Self::SwimChannel>;
```
christophermaier: If you're already using `SocketAddr`, why add another trait function to get just the port? Consumers could just as easily call `.port()` on the address they get.

components/butterfly/src/network.rs
```diff
@@ -80,3 +88,219 @@ pub trait Network: Send + Sync + Debug + 'static {
...
+    }
+}
+
+impl Network for RealNetwork {
+    type SwimChannel = SwimUdpSocket;
+    type GossipReceiver = GossipZmqSocket;
+    type GossipSender = GossipZmqSocket;
+
+    fn get_swim_addr(&self) -> SocketAddr {
+        self.swim_addr.clone()
```
christophermaier: Does this need to be a `clone`?

components/butterfly/src/network.rs
```diff
@@ -80,3 +88,219 @@ pub trait Network: Send + Sync + Debug + 'static {
...
+    }
+
+    fn get_swim_port(&self) -> u16 {
+        self.swim_addr.port()
+    }
+
+    fn get_swim_channel(&self) -> Result<SwimUdpSocket> {
```
christophermaier: The `get` naming implies (to me, anyway) that the type is giving you access to some data it owns. These `get` functions, though, are more like constructors. Naming them accordingly could help clarify the code.

components/butterfly/src/network.rs
```diff
@@ -0,0 +1,82 @@
...
+// where generic type N constrained to being an implementation of the
+// Network trait is used (trace, expire, inbound, outbound and so
+// on). I added it here, because Network is exclusively used by the
+// butterfly component.
+//
+// Same for Debug - Network is used by the butterfly component only so
+// I add it here to save me some typing.
+/// A trait for types used to provide SWIM and gossip communication
+/// channels.
+pub trait Network: Send + Sync + Debug + 'static {
```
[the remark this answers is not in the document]
krnowak: I was basically getting these errors:

```
error[E0310]: the parameter type `N` may not live long enough
   --> src/server/mod.rs:252:14
    |
126 | impl<N: Network> Server<N> {
    |      -- help: consider adding an explicit lifetime bound `N: 'static`...
...
252 |             .spawn(move || {
    |              ^^^^^
    |
note: ...so that the type `[closure@src/server/mod.rs:252:20: 255:14 server_a:server::Server<N>, swim_receiver:<N as network::Network>::SwimReceiver, swim_sender_a:<N as network::Network>::SwimSender, tx_outbound:std::sync::mpsc::Sender<(std::net::SocketAddr, message::swim::Swim)>]` will meet its required lifetime bounds
   --> src/server/mod.rs:252:14
    |
252 |             .spawn(move || {
    |              ^^^^^

error[E0310]: the associated type `<N as network::Network>::SwimReceiver` may not live long enough
   --> src/server/mod.rs:252:14
    |
252 |             .spawn(move || {
    |              ^^^^^
    |
    = help: consider adding an explicit lifetime bound `<N as network::Network>::SwimReceiver: 'static`...
note: ...so that the type `[closure@src/server/mod.rs:252:20: 255:14 server_a:server::Server<N>, swim_receiver:<N as network::Network>::SwimReceiver, swim_sender_a:<N as network::Network>::SwimSender, tx_outbound:std::sync::mpsc::Sender<(std::net::SocketAddr, message::swim::Swim)>]` will meet its required lifetime bounds
   --> src/server/mod.rs:252:14
    |
252 |             .spawn(move || {
    |              ^^^^^
```

So I either would have to add this static lifetime in every place I use the `Network` trait or its associated types or add it just for the `Network` trait. I guess the reason is that `Network` trait is used in the detached thread that lives as long as the process itself.

components/butterfly/src/network.rs
krnowak: Ah, didn't notice that `SocketAddr` implements `Copy` trait…

components/butterfly/src/network.rs
krnowak: It was mostly because to avoid clone, but I didn't notice before that `SocketAddr` implements the `Copy` trait. Will drop the port getters.

krnowak:
@cm: Sorry for the forced push - I didn't notice your review and I pushed an updated version of the abstraction. There were some changes I had to make to get my channel-based Network implementation working (because of rust being anal with traits and whatnot).

Basically what has changed is as follows:

- I splitted `SwimChannel` into `SwimReceiver` and `SwimSender`.
- I dropped the `TryClone` trait - instead we just use `Network::get_swim_sender` every time we need a sender of SWIM messages for different threads.
- `Inbound` now takes both `SwimReceiver` and `SwimSender`.
- I added a workaround for the "too many open files" issue.

Not sure I understand what do you mean by abstraction being shallow (probably it's about that it still uses `SocketAddr` for addressing) - I explained that in the commit message.

The implementation of the Network trait for the NAT testing is here: https://github.com/kinvolk/habitat/tree/krnowak/butterfly-nat-test, but I still need to clean this up.

krnowak:
Pushed some more commits to address review issues you have raised so far. Also, this hopefully fixes the CI build.

christophermaier:
Closing this for now, pending the coming Supervisor / Butterfly rewrite. We can revisit after that lands.

PR #4532, opened 2018-02-02
fix hab plan build for kubernetes exporter

mwrock:
Since `pkg-exporter-kubernetes` added a dependency on the supervisor, it fails to compile via pkg build because of its deps on ipc-channel and zeromq. We could certainly make that compile but we don't want the exporters to depend on the supervisor.

Because its just `Topology` and `ServiceBind` used, we are going to duplicate what we need here. Ideally these should be extracted to a common location.


thesentinels:

Thank you for contributing!

mwrock:
Not entirely sure why this started happening but pretty sure it is related to the repo split and cargo caching.

components/pkg-export-kubernetes/plan.sh
```diff
@@ -6,9 +6,9 @@ pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_license=('Apache-2.0')
 pkg_deps=(core/docker)
 pkg_build_deps=(
-  core/musl core/zlib-musl core/xz-musl core/bzip2-musl core/libarchive-musl
-  core/openssl-musl core/libsodium-musl
-  core/coreutils core/rust core/gcc core/make
+  core/zlib core/xz core/bzip2 core/libarchive
```
fnichol: We'll need to move some of these into `pkg_deps` so they install at runtime. The list will end up looking pretty similar to `hab-sup`'s dep list.

components/pkg-export-kubernetes/plan.sh
```diff
@@ -37,23 +37,24 @@ _common_prepare() {
 do_prepare() {
   _common_prepare
 
-  export rustc_target="x86_64-unknown-linux-musl"
```
fnichol: We can update this entire section to look a lot more like `./components/sup/plan.sh` as they are both dynamically linked against glibc. As a result, most of the linking here won't be needed.

christophermaier:
This looks good in the near term to get packages building; I've added a follow-on issue (#4546) for a longer-term solution.

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/337624421).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/337624421).

PR #4540, opened 2018-02-03
fallback to older powershell binary name if newer one is not installed

mwrock:
I realized that the newer windows launcher (not yet released) is not backwards compatible with supervisor versions 0.52.0 and under. This fixes that by falling back to the older powershell binary name if the newer one (`pwsh.exe`) which is pulled in via supervisors versions 0.53.0 and later does not exist.


reset:
Great catch!

christophermaier:
@mwrock 👍

PR #4565, opened 2018-02-06
Download on-demand packages from the `stable` channel.

### LINKED ISSUE #4539: Latest `core/hab-launcher` is installed, not latest stable hab-launcher
If you don't have `core/hab-launcher` on your machine when you start the supervisor, it will install one for you. However, it does not install the latest stable release of the launcher, it returns the latest release, regardless of channel.

The proximal cause is [here](https://github.com/habitat-sh/habitat/blob/8dbb0fc5ea7030e8e3d8e545e8b56a857a8748e4/components/hab/src/exec.rs#L74)... we should be using the stable channel. Ultimately, the use of an `Option` for this should be eliminated altogether, but we should examine the other callers of this function to ascertain the overall impact of this change, and proceed accordingly.

This only affects the first time a supervisor is started without a launcher present. Thereafter, we'll continue to use the version on disk.

christophermaier:
The functionality of the `hab` binary is split across several Habitat
packages, each of which are downloaded on-demand as needed when not
already found on disk.

As originally coded, however, there was no channel restriction applied
to this download operation. This meant that if no package was already
installed on disk, then the most recently built package was
downloaded, rather than the most recent _stable_ package. The original
code likely predated the introduction of channels. Because most
packages downloaded in this way were using an `origin/package/version`
identifier, and because they were only downloaded when not already
present, this behavior escaped notice until recently.

Now, we pull these on-demand packages from the `stable` channel.

Because the version is included in almost all these on-demand package
downloads (`core/hab-launcher` is the sole exception), and because we
currently tie all versions together based on the version of the `hab`
binary itself, this can cause issues for people developing Habitat. In
particular, there won't ever be `stable` releases of anything with a
`*-dev` version. In this case, an environment variable escape hatch is
provided with `HAB_INTERNAL_BLDR_CHANNEL`, which can be set to
`unstable` to gain access to development packages. This override will
_only_ apply to these Habitat-specific packages, and will not apply to
e.g., services you're trying to run or packages you're trying to
export. For those, the existing `HAB_BLDR_CHANNEL` environment
variable still applies.

This override should only be needed by Habitat developers, and should
ideally remain unknown and unnecessary for everyday uses.

Fixes #4539

### DIFF
--- a/components/hab/src/exec.rs
+++ b/components/hab/src/exec.rs
@@ -16,7 +16,8 @@ use std::path::{Path, PathBuf};
 
 use common;
 use common::ui::{Status, UI};
-use hcore;
+use hcore::{self, channel};
+use hcore::env as henv;
 use hcore::fs::{self, cache_artifact_path};
 use hcore::package::{PackageIdent, PackageInstall};
 use hcore::url::default_bldr_url;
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
I think we will want to make sure to set `HAB_INTERNAL_BLDR_CHANNEL` in our travis unstable builds. I *think* that would ensure unstable releases are unstable all the way through but we should validate after merge that and make sure that it the case with the unstable cli and studio image.

reset:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/338256879).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/338256879).

Issue #1180, opened 2016-08-22
Corporate proxy not working on Mac


hiucimon:
Trying to build fails behind corp proxy. Using same versions of MacOS, Habitat, and Docker on my home Mac with identical plan.sh without corp proxy works clean. We want very much to start a pilot for Habitat but need to have a few builds working first.

OSX 10.11.6 (15G31)
```
f45c89caabd3:ServiceMix ndb338$ RUST_LOG=debug RUST_BACKTRACE=1 hab studio enter
DEBUG:hab::config: Loading CLI config from /Users/ndb338/.hab/etc/cli.toml
DEBUG:hab::command::studio: Setting default origin hiucimon via CLI config
DEBUG:hab::command::studio::inner: Propagating environment variable into container: HAB_ORIGIN=hiucimon
DEBUG:hab::exec: Calling execv: ("/usr/local/bin/docker") ["run", "--rm", "--tty", "--interactive", "--privileged", "--env", "HAB_ORIGIN=hiucimon", "--volume", "/var/run/docker.sock:/var/run/docker.sock", "--volume", "/Users/ndb338/.hab/cache/keys:/hab/cache/keys", "--volume", "/Users/ndb338/Documents/Habitat/ServiceMix:/src", "habitat-docker-registry.bintray.io/studio", "enter"]
   hab-studio: Creating Studio at /hab/studios/src (default)
   hab-studio: Importing hiucimon secret origin key
» Importing origin key from standard input
★ Imported secret origin key hiucimon-20160615201715.
   hab-studio: Entering Studio at /hab/studios/src (default)
   hab-studio: Exported: HAB_ORIGIN=hiucimon
   hab-studio: Exported: http_proxy=http://ndb338:mypasswd@proxy.kdc.capitalone.com:8099
   hab-studio: Exported: https_proxy=http://ndb338:mypasswd@proxy.kdc.capitalone.com:8099

[1][default:/src:0]# build
   : Loading /src/plan.sh
   servicemix: Plan loaded
   servicemix: hab-plan-build setup
   servicemix: Using HAB_BIN=/hab/pkgs/core/hab/0.7.0/20160614230104/bin/hab for installs, signing, and hashing
   servicemix: Resolving dependencies
» Installing core/node
✗✗✗
✗✗✗ connection refused
✗✗✗
   servicemix: WARN No installed packages of 'core/node' were found
ERROR: Resolving 'core/node' failed, should this be built first?

[...]

PR #4567, opened 2018-02-07
Add build-run function and alias to studio

onlyhavecans:
One thing I find myself doing a lot when testing habitat and services in the studio is the constant unloading and starting of packages after a fresh build. This could be far more efficent and the env file generated was pointed out to me so I built a spike on adding a function to make this far easier in the studio.

I think this improved dev cycle makes testing more approachabe and delightful to someone already not having a good time figuring out why a thing doesn't work well.

I'm not a huge fan of the name `build-run` but I figure we can pick a color once the bikeshed was built.

Signed-off-by: David Aronsohn <WagThatTail@Me.com>

### DIFF
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -152,6 +152,7 @@ sup-run() {
   echo "    Running: hab sup run \$*"
   hab sup run \$* > /hab/sup/default/sup.log 2>&1 &
   echo "    * Use 'hab svc start' & 'hab svc stop' to start and stop services"
+  echo "    * Use 'build-run' to build and start the service"
   echo "    * Use 'sup-log' to tail the Supervisor's output (Ctrl+c to stop)"
   echo "    * Use 'sup-term' to terminate the Supervisor"
   if [[ -z "\${HAB_STUDIO_SUP:-}" ]]; then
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

onlyhavecans:
I am very confused how my change broke this but am open for feedback
 

```
   Compiling rusoto_credential v0.10.0
error[E0599]: no method named `to_ascii_lowercase` found for type `std::string::String` in the current scope
   --> c:\cargo\registry\src\github.com-1ecc6299db9ec823\rusoto_credential-0.10.0\src\profile.rs:178:46
    |
178 |         let lower_case_line = unwrapped_line.to_ascii_lowercase().to_string();
    |                                              ^^^^^^^^^^^^^^^^^^
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
            candidate #1: `use std::ascii::AsciiExt;`
error: Could not compile `rusoto_credential`.
warning: build failed, waiting for other jobs to finish...
error: build failed
Invoke-Build : Cargo build failed!
At C:\hab\pkgs\core\hab-studio\0.54.0-dev\20180206161953\bin\hab-plan-build.ps1:962 char:11
+     try { Invoke-Build } finally { Pop-Location }
+           ~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
+ FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Invoke-Build
 ```

robbkidd:
That's the error #1180 aims to resolve. Unrelated to your change, I think.

tashimi:
Re-running appveyor

eeyun:
Ok, this might not be a popular opinion (and if it's not thats totally fine, I just figured I'd share a quick thought here). I'm always a little bit gun-shy of adding new aliases to our kit, for a couple of different reasons. I think a lot of the time what we've tried to do with the functions, commands, and flags is to keep them as clean and generic as possible. 

I think something like this could be useful to people, but I don't know if it belongs in the core build-program. What's great about habitat's studio environment is that you could add this function to a .studiorc file and it would make it available within the studio.

If I'm a minority in this thought i'm happy to merge this PR :) just wanted to share my thoughts.

elliott-davis:
I'm with @eeyun on this one. You should be able to accomplish the same behavior by using `.studiorc` to define a function but I don't think this should be available globally.

Issue #4586, opened 2018-02-08
Gossip arbitrary metadata about services


jaym:
The example use-case I'm thinking for this is gluing together something like prometheus and the services that need monitoring. Binding to service groups seems tedious. If services could gossip metadata like "I'm a thing that needs to be monitored", the prometheus could render a config based off things based on that metadata.

eeyun:
We've decided this is a feature that we want to implement. That being said, we still need to have a discussion around what the implementation of this feature might look like. 

It will likely require quite a bit of thought to get the right shape. So for now we're going to keep the `S-needs-discussion` label and determine the appropriate path going forward.

baumanj:
Confirmed valid in v0.57

adamhjk:
I'm :-1: on having this be arbitrary. I think the pattern we are looking for is one that is declarative, not one that involves having a random lookup. 

What about a reverse binding? Where rather than declaring who fulfils that binding when prometheus starts, you get it from the other service? `--bind-to promethus.prod:services` or something like that, where you get the aggregate of all bound services?

jaym:
I think this all came about because the first thing I tried didn't work:
```
hab svc load jaym/prometheus --channel unstable --bind targets:hello-service.default --bind targets:world-service.default
```
It happily accepted that, but it didn't merge the service groups. I checked and the latest version of hab still behaves that way. From what I understand, this means that the core plan for prometheus is not really that useful if you have more than one service group. To make it work, you'd need to create a new plan, and replace the config for a section for every service group you have.

I did a little demo here of what I thought would solve my problem: https://github.com/jaym/habitat-selector-binds/blob/master/.studiorc. The idea is that when starting something like prometheus, I want to tell it to bind targets(services) to things that implement a certain interface. In the case of the demo, that was things with metrics-http-port. But this could be expanded to say things that implement the prometheus interface and things that are in a certain service group, etc. I think this approach allows services to glue themselves together a bit more.

In the example of metrics, the thing that jumps out to me is that if you wanted to try swapping in a different metrics poller thing, it would require updating all the bind-to for each running service. That being said, if you gave me `--bind-to`, I would use it over creating another prometheus service package with a  place to bind each of my service groups.

adamhjk:
[...]

PR #4604, opened 2018-02-12
[prototype] Selector binds

jaym:
This is a crack at seeing if #4586 would work the way I wanted.

In hab master, you can specify a binding by `--bind name:service_group`.
This PR introduces a new way, `--bind name:[selector]`. A selector can
select multiple service groups based on labels, which I currently
emulated by using the exported config values. There are problems with
doing that in that I wanted the select service groups and not individual
members, so it doesn't quite fit. Also because it means members within
the same service group could have different labels.

Below is an example of how I use it:

```bash
echo "Starting prometheus with selector binding:"
echo 'hab svc start jaym/prometheus --channel unstable --bind "targets:[metric-http-port]"'
hab svc start jaym/prometheus --channel unstable --bind "targets:[metric-http-port]" > /dev/null

echo "Starting hello-service"
hab svc start jaym/hello-service --channel unstable > /dev/null

echo "Starting world-service"
hab svc start jaym/world-service --channel unstable > /dev/null
```

In this example, there are 3 services: prometheus, hello-service, and
world-service. hello-service and world-service expose metrics endpoints.
Prometheus is started with a selector binding that says find me
everything with the label `metric-http-port`. So, prometheus comes up...
nothing is monitored. `hello-service` then comes up, and hab matches
`[metric-http-port]` the the `hello-service.default` service group,
rerendering the prometheus config. Then world-service comes up, hab sees
that `[metric-http-port]` also matches world-service. It adds the member
to the bind group, and the prometheus config gets rerendered. It now has
targets for both the hello-service member and world-service member

https://github.com/jaym/habitat-selector-binds has a studiorc that can run this demo

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -64,8 +64,8 @@ use sup::command;
 use sup::http_gateway;
 use sup::http_gateway::ListenAddr;
 use sup::manager::{Manager, ManagerConfig, ServiceStatus};
-use sup::manager::service::{DesiredState, ServiceBind, Topology, UpdateStrategy};
-use sup::manager::service::{CompositeSpec, ServiceSpec, StartStyle};
+use sup::manager::service::{DesiredState, ServiceBind, Topology, UpdateStrategy, };
+use sup::manager::service::{CompositeSpec, ServiceSpec, StartStyle, SpecifiedServiceBind};
 use sup::util;
 
 /// Our output key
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

elliott-davis:
hey @jaym is this still a thing you'd like to pursue?

elliott-davis:
Talked to @jaym about that and he's fine with closing this for now.

PR #4611, opened 2018-02-13
Helm exporter to add operator to exported charts

zeenix:
From the last commit's log:

    Add dependency to the Habitat operator chart in the generated charts. This
    automates the installation of the operator into the Kubernetes cluster and
    is installed as part of the generated chart for the user.
    
    Unless '--download-deps' CLI option is passed to the exporter, user of the
    exporter has to manually run
    
    $ helm repo add habitat-operator \
      https://kinvolk.github.io/habitat-operator/helm/charts/stable/
    $ helm dep up PATH_OF_GENERATED_CHART_DIR
    
    before they can install the chart or package it.
    
    Before the first time the exporter is run, Helm needs to be setup, which
    is just a matter of `helm init -c` if one needs to setup Helm for the
    purpose of creating charts only and not actually installing them on a
    Kubernetes cluster.

### DIFF
--- a/components/pkg-export-helm/defaults/HelmDeps.hbs
+++ b/components/pkg-export-helm/defaults/HelmDeps.hbs
@@ -1,0 +1,4 @@
+dependencies:
+  - name: habitat-operator
+    version: {{operator_version}}
+    repository: {{operator_repo_url}}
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

PR #4627, opened 2018-02-16
pkg-export-k8s: Adapt to Habitat Operator API changes

zeenix:
* Use the new API version: v1beta1
* Separate name for Habitat service/package than the Kubernetes resource name.
  Unless the user sets the Kubernetes resource name using the newly added
  CLI option, its derived from the Habitat service/package name.

**WARNING: BREAKING CHANGE:** This change makes the generated Kubernetes manifests work with latest Habitat Operator (>= 0.5) but breaks support for older operator versions.


thesentinels:

Thank you for contributing!

components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
```diff
@@ -24,6 +24,8 @@ spec:
   ## An object containing parameters that affects how the Habitat service
   ## is executed.
   service:
+    ## Name of the Habitat service.
```
LiliC: Not sure if I wouldn't say "Name of the Habitat pkg" here. As this is what we use to form the path to place the `user.toml` file. Not sure which one would make it more clear, service or pkg.

christophermaier: Judging by the rest of the code, service name sounds like the right thing to call this; package name implies (to me) at least an `origin/name` identifier, whereas we currently refer to services just by their names.

christophermaier:
@zeenix Are you in the process of writing up some documentation on the Kubernetes exporter? While I don't think it needs to block merging this, it will be nice to have a place where we can talk about uses of the exporter, and how it relates to the Habitat Operator. That will help users be better able to make sense of how changes like this PR may (or may not) affect them.

zeenix:
@christophermaier I think @asymmetric is looking into that but for the moment, we do have a blog post with an asciinema demo: https://kinvolk.io/blog/2017/12/introducing-the-habitat-kubernetes-exporter/ . I will put it on top of my todo to ensure we get that documenation in place very soon.

components/pkg-export-kubernetes/defaults/KubernetesManifest.hbs
zeenix: Since I called it a `Habitat service` in the comment that @LiliC was talking about, I'm wondering if you want me to change the name of the variables/fields from `hab_name` to `service_name`?

christophermaier: @zeenix Yeah, that'd probably make things more consistent and easier to keep straight.

zeenix: It's done. Sorry, i ammended the patch already so it was a forced push. Hope that's ok in this simple case. I only changed the name of the fields/variables.

christophermaier: @zeenix It's OK here. 

Also, the rename makes it much clearer.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/343987251).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/343987251).

PR #4628, opened 2018-02-16
pkg-export-helm: Depend on Habitat operator chart

zeenix:
Apart from some minor fixes/improvments to the Helm exporter code, the main change in this PR is the latest patch:
```
    Add dependency to the Habitat operator chart in the generated charts. This
    automates the installation of the operator into the Kubernetes cluster and
    is installed as part of the generated chart for the user.

    Unless '--download-deps' CLI option is passed to the exporter, user of the
    exporter has to manually run

    $ helm repo add habitat-operator \
      https://kinvolk.github.io/habitat-operator/helm/charts/stable/
    $ helm dep up PATH_OF_GENERATED_CHART_DIR

    before they can install the chart or package it.

    Before the first time the exporter is run, Helm needs to be setup, which
    is just a matter of `helm init -c` if one needs to setup Helm for the
    purpose of creating charts only and not actually installing them on a
    Kubernetes cluster.
```

**PLEASE NOTE:** While this change does not require https://github.com/habitat-sh/habitat/pull/4627 at build or runtime, the resulting Helm chart from the exporter will not work without that PR since we are fetching the latest release of Habitat Opertor here and the API and CRD version has changed.

If you would like to try these changes without https://github.com/habitat-sh/habitat/pull/4627, you will need to change the version of the operator in `requirements.yaml` in the generated chart directory to `0.4.0` manually and run `helm dup up CHARTDIR` before installing the chart.

### DIFF
--- a/components/pkg-export-helm/defaults/HelmChartFile.hbs
+++ b/components/pkg-export-helm/defaults/HelmChartFile.hbs
@@ -1,5 +1,5 @@
 name: {{name}}
 version: {{version}}
-{{#if description}}
+{{~ #if description}}
 description: {{description}}
-{{/if}}
+{{/if ~}}
--- a/components/pkg-export-helm/src/deps.rs
+++ b/components/pkg-export-helm/src/deps.rs
@@ -1,0 +1,103 @@
+// Copyright (c) 2018 Chef Software Inc. and/or applicable contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use failure::SyncFailure;
+use handlebars::Handlebars;
+use clap;
+use std::io::Write;
+use std::path::Path;
+use std::process::Command;
+
+use common::ui::{UI, Status};
+use export_docker::Result;
+use error::Error;
+
+pub const DEFAULT_OPERATOR_VERSION: &'static str = "0.5.1";
+pub const OPERATOR_REPO_URL: &'static str = "https://kinvolk.github.io/habitat-operator\
+                                             /helm/charts/stable/";
+
+// Helm requirements.yaml template
+const DEPSFILE: &'static str = include_str!("../defaults/HelmDeps.hbs");
+
+pub struct Deps {
+    operator_version: String,
+    update: bool,
+}
+
+impl Deps {
+    pub fn new_for_cli_matches(matches: &clap::ArgMatches) -> Self {
+        Deps {
+            operator_version: matches
+                .value_of("OPERATOR_VERSION")
+                .unwrap_or(DEFAULT_OPERATOR_VERSION)
+                .to_owned(),
+            update: matches.is_present("DOWNLOAD_DEPS"),
+        }
+    }
+
+    pub fn generate(&mut self, write: &mut Write) -> Result<()> {
+        let out = self.into_string()?;
+        write.write(out.as_bytes())?;
+
+        Ok(())
+    }
+
+    pub fn download<P: AsRef<Path>>(&self, dir: P, ui: &mut UI) -> Result<()> {
+        if !self.update {
+            return Ok(());
+        }
+
+        Command::new("helm")
+            .arg("repo")
+            .arg("add")
+            .arg("habitat-operator")
+            .arg(OPERATOR_REPO_URL)
+            .spawn()
+            .map_err(|_| Error::HelmLaunchFailed)
+            .and_then(|mut c| if !c.wait()
+                .map_err(|_| Error::HelmLaunchFailed)?
+                .success()
+            {
+                Err(Error::HelmNotSetup(
+                    String::from("Failed to update chart dependencies"),
+                ))?
+            } else {
+                Ok(())
+            })?;
+
+        ui.status(Status::Downloading, "dependencies")?;
+
+        Command::new("helm")
+            .arg("dep")
+            .arg("up")
+            .arg(dir.as_ref().as_os_str())
+            .spawn()?
+            .wait()
+            .map(|_| ())
+            .map_err(From::from)
+    }
+
+    // TODO: Implement TryInto trait instead when it's in stable std crate
+    fn into_string(&self) -> Result<String> {
+        let json = json!({
+            "operator_version": self.operator_version,
+            "operator_repo_url": OPERATOR_REPO_URL,
+        });
+
+        Handlebars::new()
+            .template_render(DEPSFILE, &json)
+            .map_err(SyncFailure::new)
+            .map_err(From::from)
+    }
+}
--- a/components/pkg-export-helm/src/main.rs
+++ b/components/pkg-export-helm/src/main.rs
@@ -87,8 +87,8 @@ fn cli<'a, 'b>() -> clap::App<'a, 'b> {
                 .short("V")
                 .long("version")
                 .validator(valid_version)
-                // Keep the default version here in sync with chartfile::DEFAULT_VERSION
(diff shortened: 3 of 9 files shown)


thesentinels:

Thank you for contributing!

components/pkg-export-helm/src/main.rs
```diff
@@ -84,6 +84,7 @@ fn cli<'a, 'b>() -> clap::App<'a, 'b> {
         .arg(
             Arg::with_name("VERSION")
                 .value_name("VERSION")
+                .short("v")
```
christophermaier: We use `V` elsewhere for version, and `v` for `verbose`; we should remain consistent here, too.

zeenix: Sure, fixed.

components/pkg-export-helm/src/deps.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn download<P: AsRef<Path>>(&self, dir: P, ui: &mut UI) -> Result<()> {
+        if !self.update {
+            return Ok(());
+        }
+        ui.status(Status::Downloading, "dependencies")?;
+
+        // Ignore the results cause we only install the repo as a courtesy to try to ensure that
+        // the actual `helm dep up` command afterwards succeeds. Currently `helm` doesn't mind if
+        // the repo is already added but we can't rely on that behaviour in future `helm` versions.
+        Command::new("helm")
```
christophermaier: What does the error message look like if a user tries to run this without `helm` being present? Can the exporter check that early on and give an actionable error message to the user if it's not there, along with whatever setup instructions they might need to perform? (e.g. `helm init -c`, as you mention in your commit message)

christophermaier: Also, the docker exporter uses a Habitat docker package; should we do something similar for Helm?

components/pkg-export-helm/src/deps.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn into_string(&self) -> Result<String> {
+        let json = json!({
+            "operator_version": self.operator_version,
+            "operator_repo_url": OPERATOR_REPO_URL,
+        });
+
+        let r = Handlebars::new().template_render(DEPSFILE, &json).map_err(
+            SyncFailure::new,
+        )?;
+        let s = r.lines().filter(|l| *l != "").collect::<Vec<_>>().join(
```
christophermaier: Does this indicate something that should be fixed in the template, rather than here in code?

components/pkg-export-helm/src/deps.rs
```diff
@@ -0,0 +1,99 @@
...
+    operator_version: String,
+    update: bool,
+}
+
+impl Deps {
+    pub fn new_for_cli_matches(matches: &clap::ArgMatches) -> Self {
+        Deps {
+            operator_version: matches
+                .value_of("OPERATOR_VERSION")
+                .unwrap_or(DEFAULT_OPERATOR_VERSION)
```
christophermaier: Given that we define a default, is this `unwrap_or` needed?

components/pkg-export-helm/src/main.rs
```diff
@@ -96,6 +96,18 @@ fn cli<'a, 'b>() -> clap::App<'a, 'b> {
...
+            Arg::with_name("OPERATOR_VERSION")
+                .value_name("OPERATOR_VERSION")
+                .long("operator-version")
+                .validator(valid_version)
+                .help("Version of the Habitat operator to set as dependency")
+                .default_value(deps::DEFAULT_OPERATOR_VERSION),
+        )
+        .arg(Arg::with_name("DOWNLOAD_DEPS").long("download-deps").help(
+            "Whether to download dependencies. The Kubernetes Habitat Operator is the only \
+             dependancy currently. (default: no)",
```
christophermaier: Can we just use `.default_value(false)` and have that do the right thing?

components/pkg-export-helm/src/deps.rs
zeenix: I am not sure I understand the question.  `unwrap_of` gives you the default here if user didn't pass that argument on the CLI.

components/pkg-export-helm/src/deps.rs
zeenix: I haven't checked but since `spawn` method returns an `std::io::Result`, I bet the error is of `std::io::ErrorKind::NotFound`. Regardless of the exact error, exporter can surely do that. :)

For the dependency, AFAICT helm already has a package but not in 'core' but rather 'inro' origin (no idea what that origin is about). Can we depend on that or do we need to move/create a new one to/for 'core'?

components/pkg-export-helm/src/deps.rs
zeenix: Here it's just a copy&paste from how other parts of k8s and helm exporter are rendering but I originally wrote this code to avoid empty lines in the generated files that appear because of variables not getting substituted. I'll check if it's needed here and remove. Then I'll check the other uses and see if i can remove this filtering from other places too.

components/pkg-export-helm/src/main.rs
zeenix: `default_value` expects an `&str` so that wont' work.

components/pkg-export-helm/src/deps.rs
zeenix: So turns out that if the `helm` binary is not installed, we get error from the `spawn()` call but if it's not initialized, we don't get an error but can check the exit status of the child process from the return value of the wait() call. My upcoming patch handles these two cases separately to give helpful messages to user:

```
Ω Creating file `testapp-0.2.0-20180118130531/requirements.yaml`
✗✗✗
✗✗✗ Failed to launch `helm` command. Please ensure Helm is installed.
✗✗✗
```

```
Ω Creating file `testapp-0.2.0-20180118130531/requirements.yaml`
Error: open /home/zeenix/.helm/repository/repositories.yaml: no such file or directory
✗✗✗
✗✗✗ Failed to update chart dependencies. Please ensure Helm is initialized. For only setting up the Helm client, please run `helm init -c`.
✗✗✗
```

christophermaier: I really like these error messages.

As for a helm package, I don't think `core` packages should be depending on non-`core` packages. We can look into a `core/helm` package as a separate issue; I think having these error messages will be very helpful in the meantime.

components/pkg-export-helm/src/deps.rs
christophermaier: Ugh, sorry about that... I think my brain must have been off when I typed that. I read it and understood it as `unwrap`, even though I typed `unwrap_or`. 

You are, of course, correct.

tenor-151105279

components/pkg-export-helm/src/main.rs
christophermaier: _That's_ unfortunate 😦

components/pkg-export-helm/defaults/HelmChartFile.hbs
```diff
@@ -1,5 +1,5 @@
 name: {{name}}
 version: {{version}}
-{{#if description}}
+{{~ #if description}}
```
christophermaier: Oh nice... TIL about the `~` operator 👍

christophermaier: Also, if this was a copy-paste from other exporters, it'd be nice to go back and make this change there, too.

zeenix: Yeah, PR coming up for k8s change very soon. Stay tuned. :)

components/pkg-export-helm/src/main.rs
zeenix: yeah :(

components/pkg-export-helm/src/deps.rs
zeenix: No worries, I wouldn't add an unwrap() call without a comment explaining why it's OK. Now a days, I try to use expect() instead of unwrap() so it's easy to trace the reasons.

zeenix:
@christophermaier Squashed commits.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/343987316).

PR #4638, opened 2018-02-20
Use the service name for the Helm chart name

christophermaier:
Don't use metadata name, so as not to run afoul of Helm conventions.

### DIFF
--- a/components/pkg-export-helm/src/chart.rs
+++ b/components/pkg-export-helm/src/chart.rs
@@ -45,7 +45,7 @@ impl<'a> Chart<'a> {
 
         let name = matches
             .value_of("CHART")
-            .unwrap_or(&manifest.metadata_name)
+            .unwrap_or(&manifest.service_name)
             .to_string();
         let version = matches.value_of("VERSION");
         let description = matches.value_of("DESCRIPTION");


thesentinels:

Thank you for contributing!

zeenix:
@christophermaier i'm confused. Didn't we decide not to do this today?

christophermaier:
@zeenix slight change of plans... see Slack discussion (tl;dr - removing the version info from the name _was_ the proper decision)

zeenix:
@christophermaier thanks, i saw the discussion after. Sure, please go ahead but unfortunately this change would mean that exporter will behave differently in the next release to one in 0.54 but it will behave the same as 0.53.

tashimi:
@christophermaier @zeenix let's pause on this and I will open a new issue describing the behavior I'd like to see so we can work on it from a different angle. Suggest closing this out.

christophermaier:
@tashimi Sounds good; closing this one out.

zeenix:
@tashimi no hurry on the issue opening but just wanted to share some important findings before I forget:

Helm requires the chart directory name to be the same as the chart name in the metadata:

```
 $ helm package nginx-1.11.10-20180216165808
Error: directory name (nginx-1.11.10-20180216165808) and Chart.yaml name (nginx) must match
```

This fact combined with the fact that Helm distinguishes between chart and app version (and hence also does our helm exporter), means that when you create a package to distribute, you get two versions in the package name:

```
$ helm package nginx-1.11.10-20180216165808
Successfully packaged chart and saved it to: /home/zeenix/checkout/habitat/components/pkg-export-helm/nginx-1.11.10-20180216165808-0.0.1.tgz
```
Notice the "0.0.1" towards the end (which is the default chart version assumed by the exporter unless you specify one explicitly with "--version" CLI option).

I probably makes sense to use the Habitat package version as the chart version by default but then we'll have even more redundancy and final package name will look bad, e.g `nginx-1.11.10-20180216165808-1.11.10-20180216165808.tgz`, which I'd guess is not something we want.

tashimi:
@zeenix sounds good! cc @fnichol fyi

PR #4641, opened 2018-02-21
Make HAB_STUDIO_BINARY env var override work to run alternate studio …

baumanj:
…code

This will now successfully set up a new studio as well as allow a previously
created studio to be entered while running new studio code. The prompt will
be updated to reflect the studio code being run.

See https://github.com/habitat-sh/habitat/issues/4531

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -631,14 +631,14 @@ alias la="ls -al"
 
 # Set a prompt which tells us what kind of Studio we're in
 if [ "${HAB_NOCOLORING:-}" = "true" ]; then
-  PS1='[\#]['${STUDIO_TYPE:-unknown}':\w:`echo -n $?`]\$ '
+  PS1='[\#]'${HAB_STUDIO_BINARY+[HAB_STUDIO_BINARY]}[${STUDIO_TYPE:-unknown}':\w:`echo -n $?`]\$ '
 else
   case "${TERM:-}" in
   *term | xterm-* | rxvt | screen | screen-*)
-    PS1='\[\e[0;32m\][\[\e[0;36m\]\#\[\e[0;32m\]]['${STUDIO_TYPE:-unknown}':\[\e[0;35m\]\w\[\e[0;32m\]:\[\e[1;37m\]`echo -n $?`\[\e[0;32m\]]\$\[\e[0m\] '
+    PS1='\[\e[0;32m\][\[\e[0;36m\]\#\[\e[0;32m\]]${HAB_STUDIO_BINARY+[\[\e[1;31m\]HAB_STUDIO_BINARY\[\e[0m\]]}['${STUDIO_TYPE:-unknown}':\[\e[0;35m\]\w\[\e[0;32m\]:\[\e[1;37m\]`echo -n $?`\[\e[0;32m\]]\$\[\e[0m\] '
     ;;
   *)
-    PS1='[\#]['${STUDIO_TYPE:-unknown}':\w:`echo -n $?`]\$ '
+    PS1='[\#]'${HAB_STUDIO_BINARY+[HAB_STUDIO_BINARY]}[${STUDIO_TYPE:-unknown}':\w:`echo -n $?`]\$ '
     ;;
   esac
 fi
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

fnichol:
I'd like to test this out first to understand what I would see on the prompt. On first look, this would be useful although I'm not sold on outputing more information to the prompt (we don't do this for overriding Supervisor behavior, using origin keys, etc.). I should have a chance mid-day to build and test this and will update.

baumanj:
Here's what it would look like, @fnichol:
```
➤ hab studio enter
…
[1][default:/src:0]# 
```
```
➤ env HAB_STUDIO_BINARY=$HOME/habitat/components/studio/bin/hab-studio.sh hab studio enter
…
[1][HAB_STUDIO_BINARY:/home/jbauman/habitat/components/studio/bin/hab-studio.sh][default:/src:0]# 
```
I'm definitely open to rethinking or removing the changes to the prompt, but what about the core functionality for making the `HAB_STUDIO_BINARY` override work? That is: using `hab studio version` and `hab pkg path` to resolve the necessary `version` and `libexec_path` variables. Is relying on the installed `hab` for those things appropriate?

components/studio/bin/hab-studio.sh
```diff
@@ -991,6 +991,11 @@ chroot_env() {
   if [ -n "${HAB_UPDATE_STRATEGY_FREQUENCY_MS:-}" ]; then
     env="$env HAB_UPDATE_STRATEGY_FREQUENCY_MS=$HAB_UPDATE_STRATEGY_FREQUENCY_MS"
   fi
+  # If a Habitat studio binary is set, then propagate it into the Studio's
+  # environment.
+  if [ -n "${HAB_STUDIO_BINARY:-}" ]; then
```
christophermaier: How will this behave if `HAB_STUDIO_BINARY` is given as an absolute path (i.e., a binary that exists at a path on the host system that's not at that location inside the studio)? 

Is that a valid concern? I think I've lost track at which Inception level we may be on at this point in the code :smile:

christophermaier:
@baumanj I think the overall approach is probably sound, though I haven't yet run it to try and break it 😄 

As for the prompt, maybe we could just add a single glyph of some kind to indicate that you're running with an overridden studio, and then you can consult your `env` to discover the concrete details. That would provide a useful visual indication of being in a non-standard state, while not eating up a large portion of the line with prompt.

components/studio/bin/hab-studio.sh
baumanj: I've actually only tested it with `HAB_STUDIO_BINARY` set to an absolute path:
```
11:13:10 AM jbauman@ubuntu:~/some-non-hab-source-dir
➤ env HAB_STUDIO_BINARY=$HOME/habitat/components/studio/bin/hab-studio.sh hab studio enter
```
But it does seem to work if the binary is not at a location inside the studio as above.

christophermaier: Oh, that makes total sense then; that binary determines what studio is run, so it's by definition relative to the workstation, and not the inside of the studio. 

I was one level of Inception too deep 😆

baumanj:
I considered what single glyph made sense to alert users they were using an overridden studio binary and ultimately decided that a single glyph was too abstruse. But I agree that the whole path is to verbose, so hopefully this is the happy, usable middle path:
```
➤ hab studio enter
…
[1][default:/src:0]# 
…
➤ env HAB_STUDIO_BINARY=$HOME/habitat/components/studio/bin/hab-studio.sh hab studio enter
…
[1][HAB_STUDIO_BINARY][default:/src:0]# echo $HAB_STUDIO_BINARY
/home/jbauman/habitat/components/studio/bin/hab-studio.sh
```

christophermaier:
Hrmm... Just looking at the output, I don't think I'd have guessed that `HAB_STUDIO_BINARY` in the prompt meant that something had been overridden, though to be fair, you'd need extra information looking at a single glyph, too. I'm not sure I have a better suggestion, though! Also, if you're even seeing this, you're already likely to know what you're doing, almost by definition.

I'm 👍 on this. It's great to have an indicator that you're running in a non-standard state, which is incredibly useful on it's own.

PR #4652, opened 2018-02-21
comment pkg_source in init template to prevent build error

mwrock:
Ideally, you should be able to run `build` against a freshly initialized plan without error. Currently because `pkg_source` is set to a dummy URL by default, `build` will break. This comments that line like we do for other "non buildable" settings.


thesentinels:

Thank you for contributing!

chefsalim:
Seems wierd - did that used to work?

Also just remove the offending line instead of commenting it?

mwrock:
I don;t think that ever worked. The current tutorials get away with it because they use scaffolding but not the new asp.net one. We do want it commented so that new users see all settings that are available.

Issue #4170, opened 2017-11-30
Make shell scripts like hab-studio.sh shellcheck compliant and possibly integrate shellcheck into automation


baumanj:
[`shellcheck`](https://www.shellcheck.net) is a static analysis tool that could help avoid subtle bugs. It's especially useful to highlight usage of non-POSIX features (e.g., `local`) that may be supported on some `sh` implementations but not others.

baumanj:
https://github.com/koalaman/shellcheck/wiki/TravisCI

Or once hab is installed we could use https://bldr.habitat.sh/#/pkgs/alasconnect/shellcheck/latest

baumanj:
[Shellcheck will catch that](https://github.com/koalaman/shellcheck/wiki/SC1017)!

```
➤ cat -v /tmp/test.sh 
#!/bin/sh^M
^M
echo "Hello, World"⏎                                                                                                                                                                                        

➤ shellcheck /tmp/test.sh 

In /tmp/test.sh line 1:
#!/bin/sh
         ^-- SC1017: Literal carriage return. Run script through tr -d '\r' .


In /tmp/test.sh line 2:

^-- SC1017: Literal carriage return. Run script through tr -d '\r' .
```

PR #4661, opened 2018-02-22
Clean up hab-studio.sh linter errors and reduce duplication

baumanj:
- printf -> echo: https://github.com/koalaman/shellcheck/wiki/SC2059

- Dedup VERBOSE -> -v handling

- $* -> "$@": https://github.com/koalaman/shellcheck/wiki/SC2048

- Remove local as it's not POSIX

- Dedup source_studio_type_config

- Enforce variables are set after sourcing type-specific config, e.g.:
  "$studio_env_command" -> "${studio_env_command:?}"
  "$studio_build_environment" -> "${studio_build_environment?}"

- Add double quotes: https://github.com/koalaman/shellcheck/wiki/SC2086

- Dedup root filesystem creation

- Dedup chroot_env

- Remove superfluous return

- Simplify secret printing/redaction in report_env_vars

- [ p -a q ] -> [ p ] && [ q ]: https://github.com/koalaman/shellcheck/wiki/SC2166

See https://github.com/habitat-sh/habitat/issues/4170

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -45,6 +45,20 @@ if [ -n "${DEBUG:-}" ]; then
   export DEBUG
 fi
 
+# **Internal** Set the verbose flag (i.e. `-v`) for any coreutils-like commands
+# if verbose mode was requested. This definition has to occur before the usage
+# of $v without quotes in the file to not trip the linter. If quotes are added
+# around $v, it causes failures in non-verbose mode. For example:
+#   $bb mkdir -p "$v" "$HAB_STUDIO_ROOT"/dev
+# Results in: mkdir: can't create directory '': No such file or directory
+set_v_flag() {
+  if [ -n "${VERBOSE:-}" ]; then
+    v="-v"
+  else
+    v=
+  fi
+}
+
 # ## Help/Usage functions
 
 # **Internal** Prints help and usage information. Straight forward, no?
@@ -386,11 +400,11 @@ new_studio() {
 
   # Mount filesystems
 
-  $bb mkdir -p "$v" "$HAB_STUDIO_ROOT"/dev
-  $bb mkdir -p "$v" "$HAB_STUDIO_ROOT"/proc
-  $bb mkdir -p "$v" "$HAB_STUDIO_ROOT"/sys
-  $bb mkdir -p "$v" "$HAB_STUDIO_ROOT"/run
-  $bb mkdir -p "$v" "$HAB_STUDIO_ROOT"/var/run
+  $bb mkdir -p $v "$HAB_STUDIO_ROOT"/dev
+  $bb mkdir -p $v "$HAB_STUDIO_ROOT"/proc
+  $bb mkdir -p $v "$HAB_STUDIO_ROOT"/sys
+  $bb mkdir -p $v "$HAB_STUDIO_ROOT"/run
+  $bb mkdir -p $v "$HAB_STUDIO_ROOT"/var/run
 
   # Unless `$NO_MOUNT` is set, mount filesystems such as `/dev`, `/proc`, and
   # company. If the mount already exists, skip it to be all idempotent and
@@ -398,37 +412,37 @@ new_studio() {
   if [ -z "${NO_MOUNT}" ]; then
     if ! $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/dev type"; then
       if [ -z "${KRANGSCHNAK+x}" ]; then
-        $bb mount "$v" --bind /dev "$HAB_STUDIO_ROOT"/dev
+        $bb mount $v --bind /dev "$HAB_STUDIO_ROOT"/dev
       else
-        $bb mount "$v" --rbind /dev "$HAB_STUDIO_ROOT"/dev
+        $bb mount $v --rbind /dev "$HAB_STUDIO_ROOT"/dev
       fi
     fi
 
     if ! $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/dev/pts type"; then
-      $bb mount "$v" -t devpts devpts "$HAB_STUDIO_ROOT"/dev/pts -o gid=5,mode=620
+      $bb mount $v -t devpts devpts "$HAB_STUDIO_ROOT"/dev/pts -o gid=5,mode=620
     fi
     if ! $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/proc type"; then
-      $bb mount "$v" -t proc proc "$HAB_STUDIO_ROOT"/proc
+      $bb mount $v -t proc proc "$HAB_STUDIO_ROOT"/proc
     fi
     if ! $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/sys type"; then
       if [ -z "${KRANGSCHNAK+x}" ]; then
-        $bb mount "$v" -t sysfs sysfs "$HAB_STUDIO_ROOT"/sys
+        $bb mount $v -t sysfs sysfs "$HAB_STUDIO_ROOT"/sys
       else
-        $bb mount "$v" --rbind /sys "$HAB_STUDIO_ROOT"/sys
+        $bb mount $v --rbind /sys "$HAB_STUDIO_ROOT"/sys
       fi
     fi
     if ! $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/run type"; then
-      $bb mount "$v" -t tmpfs tmpfs "$HAB_STUDIO_ROOT"/run
+      $bb mount $v -t tmpfs tmpfs "$HAB_STUDIO_ROOT"/run
     fi
     if [ -e /var/run/docker.sock ]; then
       if ! $bb mount | $bb grep -q "on $HAB_STUDIO_ROOT/var/run/docker.sock type"; then
         $bb touch "$HAB_STUDIO_ROOT"/var/run/docker.sock
-        $bb mount "$v" --bind /var/run/docker.sock "$HAB_STUDIO_ROOT"/var/run/docker.sock
+        $bb mount $v --bind /var/run/docker.sock "$HAB_STUDIO_ROOT"/var/run/docker.sock
       fi
     fi
 
     if [ -h "$HAB_STUDIO_ROOT/dev/shm" ]; then
-      $bb mkdir -p "$v" "$HAB_STUDIO_ROOT/$($bb readlink "$HAB_STUDIO_ROOT"/dev/shm)"
+      $bb mkdir -p $v "$HAB_STUDIO_ROOT/$($bb readlink "$HAB_STUDIO_ROOT"/dev/shm)"
     fi
 
     # Mount the `$ARTIFACT_PATH` under `/hab/cache/artifacts` in the Studio,
@@ -436,9 +450,9 @@ new_studio() {
     if [ -z "${NO_ARTIFACT_PATH}" ]; then
       studio_artifact_path="${HAB_STUDIO_ROOT}${HAB_CACHE_ARTIFACT_PATH}"
       if ! $bb mount | $bb grep -q "on $studio_artifact_path type"; then
-        $bb mkdir -p "$v" "$ARTIFACT_PATH"
-        $bb mkdir -p "$v" "$studio_artifact_path"
-        $bb mount "$v" --bind "$ARTIFACT_PATH" "$studio_artifact_path"
+        $bb mkdir -p $v "$ARTIFACT_PATH"
+        $bb mkdir -p $v "$studio_artifact_path"
+        $bb mount $v --bind "$ARTIFACT_PATH" "$studio_artifact_path"
       fi
     fi
   fi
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/studio/bin/hab-studio.sh
```diff
@@ -920,104 +843,65 @@ sanitize_secrets() {
 # to pass into the studio
 load_secrets() {
   sanitize_secrets
-  echo $($bb env | $bb awk -F '=' '/^HAB_STUDIO_SECRET_/ {gsub(/HAB_STUDIO_SECRET_/, "", $0); print}')
+  $bb env | $bb awk -F '=' '/^HAB_STUDIO_SECRET_/ {gsub(/HAB_STUDIO_SECRET_/, ""); print}'
```
baumanj: I thought this would be of interest to you, @elliott-davis. I changed this because it tripped [SC2005](https://github.com/koalaman/shellcheck/wiki/SC2005) and [SC2046](https://github.com/koalaman/shellcheck/wiki/SC2046). The former is self-explanatory, the latter was complaining because it though the `$0` was referring to the shell variable. But since [`gsub` doesn't require that argument](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html):
```
gsub(ere, repl[, in])
Behave like sub (see below), except that it shall replace all occurrences of the regular expression (like the ed utility global substitute) in $0 or in the in argument, when specified.
```
I just removed it.

components/studio/bin/hab-studio.sh
```diff
@@ -1063,11 +947,9 @@ report_env_vars() {
...
   fi
 
-  if [ "$(load_secrets)" ]; then
-    for secret in $(load_secrets); do
-      info "$(echo $secret | $bb awk -F= '{print "Exported:",$1"=[redacted]"}')"
-    done
-  fi
+  for secret_name in $(load_secrets | $bb cut -d = -f 1); do
+    info "Exported: $secret_name=[redacted]"
+  done
```
baumanj: Also interested in your input here, @elliott-davis. I think this is a bit simpler without sacrificing safety.

components/studio/bin/hab-studio.sh
```diff
@@ -1220,7 +1087,7 @@ unset PATH
 
 # The root path of the Habitat file system. If the `$HAB_ROOT_PATH` environment
 # variable is set, this value is overridden, otherwise it is set to its default
-: ${HAB_ROOT_PATH:=/hab}
+: "${HAB_ROOT_PATH:=/hab}"
 # The root path containing all locally installed packages
 HAB_PKG_PATH=$HAB_ROOT_PATH/pkgs
```
baumanj: This is the only remaining linter error:
[`HAB_PKG_PATH` appears unused. Verify it or export it.](https://github.com/koalaman/shellcheck/wiki/SC2034)

I'm hesitant to remove it, but I don't see how it's used at all. Thoughts?

fnichol:
I’ll need to fully review this change set which won’t happen until at least Monday before I can give more feedback.

components/studio/bin/hab-studio.sh
elliott-davis: Oh TIL! Thanks @baumanj

components/studio/bin/hab-studio.sh
elliott-davis: :+1: - Plus it's one less call to awk

elliott-davis:
From a code perspective this looks good to me - I'd like to take some time to functionally review it before I give it the ol' :+1:

components/studio/bin/hab-studio.sh
mwrock: That variable is used inside of a `plan.sh` @baumanj see https://www.habitat.sh/docs/reference/#plan-variables

baumanj: Yeah, but I think in that instance it's set here in [hab-plan-build.sh](https://github.com/habitat-sh/habitat/blob/master/components/plan-build/bin/hab-plan-build.sh#L351). As best I can tell, the only environment variables that pass from `hab-studio.sh` into subcommands are those that are included by [`chroot_env`](https://github.com/habitat-sh/habitat/blob/master/components/studio/bin/hab-studio.sh#L928), but `HAB_PKG_PATH` is not among those. Am I missing something, @mwrock ?

components/studio/bin/hab-studio.sh
```diff
@@ -356,17 +338,22 @@ subcommand_run() {
...
-# **Internal** Creates a new Studio.
-new_studio() {
-  # Check if a pre-existing Studio configuration is found and use that to
-  # determine the type
+# **Internal** Check if a pre-existing Studio configuration is found and use
+# that to determine the type and assign it to STUDIO_TYPE
+source_studio_type_config() {
   if [ -s "$studio_config" ]; then
+    # shellcheck disable=1090
     . "$studio_config"
```
baumanj: It turns out [`hab-studio-type-default.sh`](https://github.com/habitat-sh/habitat/blob/master/components/studio/libexec/hab-studio-type-default.sh#L6) references a positional parameter, which means that we have to pass those args into `source_studio_type_config` or things will very subtly break. I think we should look into changing that as I think it's a fairly brittle and confusing abstraction.

components/studio/bin/hab-studio.sh
mwrock: ugh. apologies. I was thinking of plan build. not studio.

baumanj: Hey, thanks for looking! I'm like 95% sure it should be removed, but I'll wait for @fnichol and/or @elliott-davis to weigh in.

components/studio/bin/hab-studio.sh
baumanj: I made some changes to try to evaluate this later and give an error at the more appropriate time. See:
1. https://github.com/habitat-sh/habitat/pull/4661/files#diff-b102b5e067962e9cce22e9e8d3541fefR591
2. https://github.com/habitat-sh/habitat/pull/4661/files#diff-b102b5e067962e9cce22e9e8d3541fefR727
3. https://github.com/habitat-sh/habitat/pull/4661/files#diff-98ec64e3f381a5ccde4f4506db6aae9cR6

Though maybe we should just rework this so that `studio_build_command` doesn't need to reference `$1`, as that's super confusing.

baumanj: This is moot now; I reworked it to be simpler: https://github.com/habitat-sh/habitat/pull/4661/commits/ae4e0d69078762aadaa685d3bbe03686e7a9e4c8

components/studio/bin/hab-studio.sh
```diff
@@ -844,14 +730,14 @@ info() {
   fi
 
   if [ "${HAB_NOCOLORING:-}" = "true" ]; then
-    printf -- "   ${program:-unknown}: $1\n"
+    echo "   ${program:-unknown}: $1"
```
fnichol: Using a raw `echo` breaks the ANSI coloring, requiring the `-e` flag which isn't always available--one reason why I used `printf` originally in `info()`, `exit_with()`, and `ensure_root()`

components/studio/bin/hab-studio.sh
```diff
@@ -844,14 +730,14 @@ info() {
...
   else
     case "${TERM:-}" in
       *term | xterm-* | rxvt | screen | screen-*)
-        printf -- "   \033[1;36m${program:-unknown}: \033[1;37m$1\033[0m\n"
+        echo "   \033[1;36m${program:-unknown}: \033[1;37m$1\033[0m"
```
fnichol: Sorry, this is the more problematic call

components/studio/bin/hab-studio.sh
```diff
@@ -861,27 +747,24 @@ info() {
...
-  local msg
-  local fatal
-
   warn_msg="Running Habitat Studio requires root or administrator privileges. \
 Please retry this command as a super user or use a privilege-granting facility such as sudo."
   fatal_msg="Root or administrator permissions required to complete operation"
 
   if [ "${HAB_NOCOLORING:-}" = "true" ]; then
-    printf -- "∅ $warn_msg\n\n✗✗✗\n✗✗✗ $fatal_msg\n✗✗✗\n"
+    printf -- "∅ %s\n\n✗✗✗\n✗✗✗ %s\n✗✗✗\n" "$warn_msg" "$fatal_msg"
```
fnichol: 👍 yeah!

components/studio/bin/hab-studio.sh
```diff
@@ -895,18 +778,18 @@ Please retry this command as a super user or use a privilege-granting facility s
 # ```
 exit_with() {
   if [ "${HAB_NOCOLORING:-}" = "true" ]; then
-    printf -- "ERROR: $1\n"
+    echo "ERROR: $1"
   else
     case "${TERM:-}" in
       *term | xterm-* | rxvt | screen | screen-*)
-        printf -- "\033[1;31mERROR: \033[1;37m$1\033[0m\n"
+        echo "\033[1;31mERROR: \033[1;37m$1\033[0m"
```
fnichol: Same ANSI coloring issue as above

components/studio/bin/hab-studio.sh
fnichol: Yeah that's better. In the later case shellcheck most likely threw a false positive as that token was single-quoted, and on purpose. That being said, nice to drop the `$0` if that's possible!

components/studio/bin/hab-studio.sh
```diff
@@ -895,18 +778,18 @@ Please retry this command as a super user or use a privilege-granting facility s
...
         ;;
       *)
-        printf -- "ERROR: $1\n"
+        echo "ERROR: $1"
         ;;
     esac
   fi
-  exit $2
+  exit "$2"
```
fnichol: While shellcheck leads you to do this, in this case `$2` would be an integer which doesn't necessarily need to be quoted.

components/studio/bin/hab-studio.sh
```diff
@@ -1118,56 +956,45 @@ chown_artifacts() {
...
 }
 
 # **Internal** Sets the `$libexec_path` variable, which is the absolute path to
 # the `libexec/` directory for this software.
 set_libexec_path() {
-  local bb
-  local p
   # First check to see if we have been given a path to a `busybox` command
-  if [ -n "${BUSYBOX:-}" -a -x "${BUSYBOX:-}" ]; then
+  if [ -n "${BUSYBOX:-}" ] && [ -x "${BUSYBOX:-}" ]; then
```
fnichol: Yep, that's the safer version!

components/studio/bin/hab-studio.sh
fnichol: Yep, it's used in the `baseimage` and `bare` Studio type implementations (found this with a grep through the codebase). I'd add an ignore for that one.

components/studio/bin/hab-studio.sh
```diff
@@ -1236,7 +1064,7 @@ hab="$libexec_path/hab"
 # The author of this program
 author='@author@'
 # The short version of the program name which is used in logging output
-program=$($bb basename $0)
+program=$($bb basename "$0")
```
fnichol: If `$0` is quoted, won't shellcheck also complain about `$($bb basename "$0")` being unquoted?

components/studio/libexec/hab-studio-profile.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+record_build() {
+  build_command_name=$1
+  plan_context=$2
+  session=$plan_context
+  record "$session" "$build_command_name" "$plan_context"
+}
+
+cd /src || echo "ERROR: cd to /src failed"
```
fnichol: Why eat this failure? If changing directory to `/src` fails something is terribly wrong with the Studio setup.

components/studio/libexec/hab-studio-profile.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
baumanj: Isn't it? I thought we only loaded this when we executed

```bash
studio_enter_command="$HAB_ROOT_PATH/bin/hab pkg exec core/hab-backline bash --login +h"
```
or
```bash
studio_run_command="$HAB_ROOT_PATH/bin/hab pkg exec core/hab-backline bash --login"
```
In any case, [it makes use of `[[` and `=~`](https://github.com/habitat-sh/habitat/blob/master/components/studio/bin/hab-studio.sh#L669) which are a bash-only constructs.

components/studio/libexec/hab-studio-profile.sh
baumanj: Just [to satisfy the linter](https://github.com/koalaman/shellcheck/wiki/SC2164), really. I could disable that warning or the best thing might be to `|| exit`. What do you think?

components/studio/bin/hab-studio.sh
baumanj: Ah, good catch. The linter complained about the use of [variables in the format string](https://github.com/koalaman/shellcheck/wiki/SC2059), but we can change this to
```bash
        printf -- "   \033[1;36m%s: \033[1;37m$1\033[0m\n" "${program:-unknown}"
```
While we're here though, should it be printing all the args and not just the first?
```bash
        printf -- "   \033[1;36m%s: \033[1;37m$*\033[0m\n" "${program:-unknown}"
```

components/studio/bin/hab-studio.sh
baumanj: Yeah, it's just easier to add the quotes than the `disable`, I figure. I mean, there's no harm, right?

Plus, if someone misused `exit_with` and did something like:
```bash
status="*" # <== assume a logic error leads to something like this
exit_with "bad news" "$status"
```
I'd say we don't want to expand that. This error:
```
ERROR: bad newslogout
bash: exit: *: numeric argument required
```
Is more understandable to me than:
```
ERROR: bad newslogout
bash: exit: builder: numeric argument required
```

components/studio/bin/hab-studio.sh
baumanj: It doesn't. I'm not exactly sure why. It would be fine to quote the `$(…)` as well, but I haven't been quoting where `shellcheck` doesn't seem to thing it's necessary. Quoting both is probably better for correctness. It's not likely that `$0` would contain a any `$IFS` characters, but you never know.

elliott-davis:
It looks like `hab studio rm` doesn't preserve the new line behavior

```elliott@elliott-MacBookPro:/src/builder$ hab studio rm
   hab-studio.sh: Destroying Studio at /hab/studios/src--builder ()elliott@elliott-MacBookPro:/src/builder$```

**Comment by elliott-davis** [comment]
same newline behavior change here: old
```
elliott@elliott-MacBookPro:/src/builder$ hab studio enter
[sudo hab-studio] password for elliott: 
   hab-studio.sh: Creating Studio at /hab/studios/src--builder (default)
   hab-studio.sh: Importing 'edavis' secret origin key
★ Imported secret origin key edavis-20180223212513.
```
new
```
   hab-studio.sh: Creating Studio at /hab/studios/src--builder (default)   hab-studio.sh: Importing 'edavis' secret origin key» Importing origin key from standard input
★ Imported secret origin key edavis-20180223212513.
```

**Comment by baumanj** [comment]
Ah good catch. I unintentionally removed a newline in the `info` function. Will fix.

File: components/studio/libexec/hab-studio-profile.sh [hunk 14]
```diff
@@ -0,0 +1,72 @@
...
+      ownership=$($bb stat -c '%u:%g' /src)
+      $bb chown -R "$ownership" "/src/results" || true
+    fi
+    unset BUSYBOX LOGDIR name ownership
+
+    $bb script -c "$bb env -i $env $cmd $*" -e "$log"
+  ); return $?
+}
+
+record_build() {
```
elliott-davis: Would you mind dropping an underscore here to signal that it's internal. Adding the leading underscore also preserves tab completion on `record`

components/studio/libexec/hab-studio-profile.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+_record_build() {
+  build_command_name=$1
+  plan_context=$2
+  session=$plan_context
+  record "$session" "$build_command_name" "$plan_context"
+}
+
+cd /src || exit 1 # If you can't cd to /src, the studio is totally borked
```
elliott-davis: Could we change this to `cd /src || echo 'Failed to cd to /src. Exiting Studio!' && exit 1`? It's pretty jarring to get ejected without warning.

baumanj: If `cd /src` fails you'll already get an error message
```
bash: cd: /src: No such file or directory
```
but I'm happy to add an additional "Exiting Studio!" print if you think that would make things clearer.

elliott-davis:
It wasn't immediately clear that I got booted out of the studio when I was
testing this so I think a point of clarity would be good

baumanj:
OK, output should look like this now if the `cd /src` fails:
```
bash: cd: /foo: No such file or directory
Setup failed; exiting studio!
```

PR #4681, opened 2018-02-27
Fix Helm chart naming

zeenix:
Please read https://github.com/habitat-sh/habitat/pull/4638 for rationale for these changes, especially https://github.com/habitat-sh/habitat/pull/4638#issuecomment-367686986.

*WARNING:* This changes the behavior of the Helm exporter. After these changes, the generated chart source directory will no longer have version and release suffix in the name. e.g if before these changes, the exporter produced a directory named `nginx-1.11.10-20180216165808`, it will now create it as `nginx`.

### DIFF
--- a/components/pkg-export-helm/src/chart.rs
+++ b/components/pkg-export-helm/src/chart.rs
@@ -45,7 +45,7 @@ impl<'a> Chart<'a> {
 
         let name = matches
             .value_of("CHART")
-            .unwrap_or(&manifest.metadata_name)
+            .unwrap_or(&manifest.pkg_ident.name)
             .to_string();
         let version = matches.value_of("VERSION");
         let description = matches.value_of("DESCRIPTION");
--- a/components/pkg-export-helm/src/chart.rs
+++ b/components/pkg-export-helm/src/chart.rs
@@ -80,7 +80,7 @@ impl<'a> Chart<'a> {
 
         let mut values = Values::new();
         values.add_entry("metadataName", &manifest.metadata_name);
-        values.add_entry("serviceName", &manifest.service_name);
+        values.add_entry("serviceName", &manifest.pkg_ident.name);
         values.add_entry("imageName", &manifest.image);
         values.add_entry("instanceCount", &manifest.count.to_string());
         values.add_entry("serviceTopology", &manifest.service_topology.to_string());
--- a/components/pkg-export-helm/src/chart.rs
+++ b/components/pkg-export-helm/src/chart.rs
@@ -47,7 +47,12 @@ impl<'a> Chart<'a> {
             .value_of("CHART")
             .unwrap_or(&manifest.pkg_ident.name)
             .to_string();
-        let version = matches.value_of("VERSION");
+        let pkg_version = manifest.pkg_ident.version.clone();
+        let version = matches.value_of("VERSION").or(
+            pkg_version.as_ref().map(|s| {
+                s.as_ref()
+            }),
+        );
         let description = matches.value_of("DESCRIPTION");
         let chartfile = ChartFile::new(&name, version, description);
         let deps = Deps::new_for_cli_matches(&matches);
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

zeenix:
Reviewers, just FYI github for some reason lists the patches in the wrong order. :(

components/pkg-export-helm/src/main.rs
```diff
@@ -66,6 +66,12 @@ fn export_for_cli_matches(ui: &mut UI, matches: &clap::ArgMatches) -> Result<()>
     Ok(())
 }
 
+lazy_static! {
+    pub static ref VERSION_HELP: String =
+        format!("Version of the chart to create (default: {{pkg_version}} if available, or {})",
+                chartfile::DEFAULT_VERSION);
+}
+
```
christophermaier: Why is `lazy_static` needed here?

components/pkg-export-helm/src/chart.rs
```diff
@@ -47,7 +47,12 @@ impl<'a> Chart<'a> {
             .value_of("CHART")
             .unwrap_or(&manifest.pkg_ident.name)
             .to_string();
-        let version = matches.value_of("VERSION");
+        let pkg_version = manifest.pkg_ident.version.clone();
+        let version = matches.value_of("VERSION").or(
```
christophermaier: Totally minor, but could you use `or_else` here instead, and only only clone a version when necessary?

components/pkg-export-helm/src/chart.rs
```diff
@@ -47,7 +47,12 @@ impl<'a> Chart<'a> {
...
+            pkg_version.as_ref().map(|s| {
+                s.as_ref()
```
christophermaier: Are you taking a reference to a reference here?

components/pkg-export-helm/src/main.rs
zeenix: because that's the only way to create a static string from two strings in Rust currently AFAIK. The `Arg::help()` we pass this to, expects a static string.

components/pkg-export-helm/src/chart.rs
zeenix: not exactly. Getting an `&str` from an `&String`.

components/pkg-export-helm/src/chart.rs
zeenix: I first tried to keep it simple and not clone but not cloning meant that we'd have to borrow `manifest` and borrow it again before the first borrow is finished in this line:

```
    Ok(Self::new_for_manifest(manifest, name, chartfile, deps, ui))
```

zeenix: why prefer `or_else` btw? It only seems to make the call more complex.

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/346832474).

components/pkg-export-helm/src/chart.rs
zeenix: Actually we're moving the `manifest` while borrowed if we don't clone the version string:

```
error[E0505]: cannot move out of `manifest` because it is borrowed
  --> components/pkg-export-helm/src/chart.rs:59:35
   |
51 |             manifest.pkg_ident.version.as_ref().map(|s| {
   |             -------------------------- borrow of `manifest.pkg_ident.version` occurs here
...
59 |         Ok(Self::new_for_manifest(manifest, name, chartfile, deps, ui))
   |                                   ^^^^^^^^ move out of `manifest` occurs here

```

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/346832474).

PR #4692, opened 2018-02-28
allow supervisor to start services as pkg_svc_user

### LINKED ISSUE #4685: windows packages are not honoring `pkg_svc_user`
Last year we added functionality to start Habitat services under alternate credentials in #2648. It is with extreme embarrassment that I discover that we never actually honored the configured `pkg_svc_user` . We tested this functionality against the "default" hab user which is the user "hab" if it exists. Since that user is not necessarily the logged in user, it validated the basic supervisor functionality which had all of the hard work fortunately.

So now we need to make sure that the windows implementation of `get_user_and_group` does not simply return the default user. In fact, this may be as simple of removing the platform branching alltogether and just using one implementation - what is currently the linux implementation.

mwrock:
fixes #4685 

Note that this mainly moves existing code around. It pulls tyhe unix specific `get_user_and_group` function into `package` which is the only place that calls that function and removes the no-op windows implementation so that windows follows through with the same behavior which is to read the package metadata for the svc_user.

This also changes the windows `assert_pkg_user_and_group` implementation which was once a no-op to actually validate the user so that is the user does not exist on the system, a propper error is returned by the supervisor.

Note that this PR is "enhanced" by https://github.com/habitat-sh/core/pull/10 which will not require users to also specify a `pkg_svc_group` in their plans if also specifying a `pkg_svc_user`.

I have tested this with alternate users, bogus users and non admin users and things are working well.


thesentinels:

Thank you for contributing!

christophermaier:
tenor-170912738

Looks good!

@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/347508673).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/347508673).

components/sup/src/manager/service/package.rs
```diff
@@ -145,3 +149,57 @@ impl Pkg {
...
+/// if not, we'll try and use hab/hab.
+/// If hab/hab doesn't exist, try to use (current username, current group).
+/// If that doesn't work, then give up.
+fn get_user_and_group(pkg_install: &PackageInstall) -> Result<(String, String)> {
+    if let Some((user, group)) = get_pkg_user_and_group(&pkg_install)? {
+        Ok((user, group))
+    } else {
+        let defaults = default_user_and_group()?;
+        Ok(defaults)
+    }
```
baumanj: I'm late to the party, but wanted to review for the sake of my `rust` learning. Could this be written as:
```rust
    Ok(get_pkg_user_and_group(&pkg_install)?.unwrap_or(
        default_user_and_group()?,
    ))
```

components/sup/src/sys/unix/users.rs
```diff
@@ -16,7 +16,6 @@ use hcore::os::users;
 use hcore::package::PackageInstall;
```
baumanj: I think this is no longer referenced and can be removed

components/sup/src/sys/windows/users.rs
```diff
@@ -12,19 +12,20 @@
...
+/// Windows does not have a concept of "group" in a Linux sense
+/// So we just validate the user
+pub fn assert_pkg_user_and_group(user: &str, _group: &str) -> Result<()> {
+    match users::get_uid_by_name(user) {
+        Some(_) => Ok(()),
+        None => Err(sup_error!(Error::Permissions(format!(
+            "Package requires user {} to exist, but it doesn't",
+            user
+        )))),
+    }
```
baumanj: I think this is fine. Again, for the sake of learning, what about
```rust
    users::get_uid_by_name(user)
        .ok_or(sup_error!(Error::Permissions(format!(
            "Package requires user {} to exist, but it doesn't",
            user
        ))))
        .map(|_| ())
```

components/sup/src/manager/service/package.rs
raskchanky: It's hard to know without compiling, because sometimes rustc insists on the creation of temporary variables in order to satisfy its borrow checking logic.  On the surface they seem the same, though to my eye, Matt's code is easier to understand.

components/sup/src/sys/windows/users.rs
raskchanky: Just a personal preference, but I think Matt's code in this case is easier to understand.

components/sup/src/manager/service/package.rs
baumanj: Oh, I compiled it. I was asking if this was any different logically.

 Good to know your readability preference, though.

components/sup/src/sys/windows/users.rs
baumanj: Yeah, I'm not a fan of the need for `.map(|_| ())`. Since `Result<()>` is a common idiom, it would be nice if there were a more idiomatic way to get there.

mwrock:
Yeah I'll admit that I fairly blindly just moved this code in the PR from one module to another without a very critical eye. I probably was not the original author of most of it.

PR #4706, opened 2018-03-05
Encrypt and upload secrets.

elliott-davis:
Blocked by: https://github.com/habitat-sh/builder/pull/232


thesentinels:

Thank you for contributing!

components/builder-depot-client/src/lib.rs
```diff
@@ -398,6 +398,50 @@ impl Client {
...
+        let sbody = serde_json::to_string(&body)?;
+        let res = self.add_authz(self.0.post(&path), token)
+            .body(&sbody)
+            .header(Accept::json())
+            .header(ContentType::json())
+            .send()?;
+
+        if res.status != StatusCode::Created {
+            return Err(err_from_response(res));
+        };
```
baumanj: Is that `;` doing something here?

components/builder-depot-client/src/lib.rs
```diff
@@ -398,6 +398,50 @@ impl Client {
         )
     }
 
+    /// Create secret for an origin
+    ///
+    /// # Failures
+    ///
+    /// * Remote Builder is not available
+    pub fn create_origin_secret(&self, origin: &str, token: &str, key: &str, secret: &str) -> Result<()> {
```
baumanj: How does one update a secret? Presumably by calling this function again? I'm wondering if there's utility in having there be any distinction in the interface about whether there was already a value for the provided `key` (such as a `Result<bool>` return) or requiring something like an `overwrite: bool` parameter. I don't actually know; just throwing that idea out there.

components/hab/src/cli.rs
```diff
@@ -264,6 +264,46 @@ pub fn get() -> App<'static, 'static> {
...
+                (@subcommand delete => 
+                    (about: "Delete a secret for your origin")
+                    (@arg KEY_NAME: +required +takes_value
+                        "The name of the variable key to be injected into the studio.")
+                    (@arg BLDR_URL: -u --url +takes_value {valid_url}
+                        "Specify an alternate Builder endpoint. If not specified, the value will \
+                         be taken from the HAB_BLDR_URL environment variable if defined. (default: \
+                         https://bldr.habitat.sh)")
+                    (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
+                    (@arg ORIGIN: +takes_value
```
baumanj: I wonder if it might be a bit more usable to make `ORIGIN` an option (requiring `-o, --origin`)
```
hab origin secret delete foo bar
```
is less obvious to me than
```
hab origin secret delete --origin bar foo
```
And generally, I think making the purpose of arguments that will often be omitted explicit reduces cognitive load. This goes for all the `hab origin secret` subcommands

components/hab/src/cli.rs
```diff
@@ -264,6 +264,46 @@ pub fn get() -> App<'static, 'static> {
...
+                        "The origin for which the secret will be generated. Default is from 'HAB_ORIGIN' or cli.toml")
+                )
+                (@subcommand delete => 
+                    (about: "Delete a secret for your origin")
+                    (@arg KEY_NAME: +required +takes_value
```
baumanj: Should we add aliases to be consistent with the other commands?

components/hab/src/cli.rs
```diff
@@ -264,6 +264,46 @@ pub fn get() -> App<'static, 'static> {
                     (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
                 )
             )
+            (@subcommand secret =>
+                (about: "Commands related to secret management")
+                (@setting ArgRequiredElseHelp)
+                (@subcommand generate =>
```
baumanj: The name `generate` implies to me that I don't have to come up with the value. To me `add` is the more natural counterpart to `delete`.

components/builder-depot-client/src/lib.rs
elliott-davis: right now there will be a unique constraint on the key name per origin. Users will have to delete and recreate a key if the want to reuse it. This behavior should be consistent with Travis CI

components/hab/src/cli.rs
elliott-davis: I talked with @ryankeairns about this and he wanted the UX to be consistent with `origin key generate`. Previously I had `origin secret create` to follow the CRUD pattern.

components/hab/src/cli.rs
elliott-davis: I don't think so. Personally I'd like to remove the other aliases as well. I find they often lead to typo bugs and provide an inconsistent user experience.

components/hab/src/cli.rs
elliott-davis: I agree with the idea of having origin be a flag but in an effort to remain consistent with the existing UX I left it as a positional argument. I think a larger refactor is necessary to make origin a flag everywhere to keep everything consistent

components/hab/src/command/origin/secret/generate.rs
```diff
@@ -0,0 +1,60 @@
...
+        Err(_) => {
+            debug!("Didn't find public encryption key in cache path");
+            download_public_encryption_key(ui, &depot_client, origin, token, cache)?;
+            BoxKeyPair::get_latest_pair_for(origin, cache)?
+        }
+    };
+
+    ui.status(Status::Encrypting, format!("value for key {}.", key))?;
+    let encrypted_secret_bytes = encryption_key.encrypt(secret.as_bytes(), None)?;
+    let encrypted_secret_string = String::from_utf8_lossy(&encrypted_secret_bytes);
```
baumanj: If this is arbitrary bytes, I worry about storing them in a `str`, which guarantees utf8 validity. Can we switch to either base64 encoding or using a `Vec<u8>` in `create_origin_secret`?

components/builder-depot-client/src/lib.rs
baumanj: That sounds good. Where is the the `depot/origins/{}/secret` implementation? Also, might be worth a note in the help.

components/hab/src/cli.rs
baumanj: I 💯 on-board with consistency, but in this case I'd say it's a different operation and so should have a different name to make that clear. Not a hill I'm looking to die on or anything, it just struck me as unexpected. I'm interested what @ryankeairns thinks.

components/hab/src/cli.rs
baumanj: I'm on-board with removing all the aliases, but I think we should do that as a separate change rather than introducing just one new command that doesn't follow the pattern and leads to an inconsistent user experience. Not a big deal; do what feels right to you.

components/hab/src/cli.rs
baumanj: I looked at the rest of the UX when there are multiple positional parameters, we *do* make origin an option:
```
    hab bldr job promote [FLAGS] [OPTIONS] <GROUP_ID> <CHANNEL>
…
OPTIONS:
…
    -o, --origin <ORIGIN>      Limit the promotable packages to the specified origin
```
When the origin is the *only* possible positional argument, it's fine, but in this case I think it's the UX consistent behavior to make it an option requiring a flag.

components/hab/src/cli.rs
ryankeairns: Just so I'm following... generate generates a new key; are we also creating/generating a new secret?

components/hab/src/cli.rs
ryankeairns: My .02 is that I also prefer the clarity of using the -o or --origin flag... that also helps with not having concern your self with the order of the arguments - its explicit that the thing following the flag is in fact the origin name, in other words.

components/hab/src/cli.rs
baumanj: @elliott-davis can correct me if I'm wrong, but what we're doing here is asking builder to store an encrypted key/value pair that we are supplying. Builder isn't generating a value for us.

components/builder-depot-client/src/lib.rs
elliott-davis: https://github.com/habitat-sh/builder/pull/232

components/hab/src/cli.rs
ryankeairns: I don't want to beat this to death (too late!), but perhaps `upload` is the verb then.

baumanj:
All the changes look look great! I'm still a unclear on https://github.com/habitat-sh/habitat/pull/4706#discussion_r172613781. Your thoughts?

components/hab/src/command/origin/secret/generate.rs
elliott-davis: I think I can switch `from_utf8_lossy` to `from_utf8` to get a utf8 guarantee? What do you think?

elliott-davis: Done :tada:

baumanj: I think that you're going to get errors at the rate which a random series of bytes do not make valid UTF-8, right? I don't know what that frequency is off the top of my head, but if there a reason why you wouldn't?

elliott-davis: I'm not sure I follow the random series of bytes bit. We are taking a user supplied string, converting it to bytes, sending it to encrypt which returns a base64 encoded series of bytes, then we convert that back into a string for a json payload. Does that match up with what you were thinking?

baumanj: For posterity, the confusion was mine. I did not realize that `BoxKeyPair::encrypt` returned already base64-encoded data. I assumed since it's return type was `Result<Vec<u8>>` that it was raw bytes. This is pretty misleading, so I filed an issue: https://github.com/habitat-sh/core/issues/15

components/builder-depot-client/src/lib.rs
```diff
@@ -391,13 +410,87 @@ impl Client {
...
+        }
+
+        let mut encoded = String::new();
+        res.read_to_string(&mut encoded)?;
+        debug!("Response body: {:?}", encoded);
+        let secret_keys: Vec<String> = serde_json::from_str::<Vec<OriginSecret>>(&encoded)?
+            .into_iter()
+            .map(|m| m.into())
+            .map(|s: originsrv::OriginSecret| s.get_name().to_string())
+            .collect();
```
raskchanky: I love this kind of chaining.  So good.

components/hab/src/command/bldr/channel/destroy.rs
```diff
@@ -18,18 +18,18 @@ use common::ui::{Status, UI};
...
-    ui.status(Status::Deleting, format!("channel {}.", channel))?;
+    ui.status(Status::Deleting, format!("secret key {}.", key))?;
 
     depot_client
-        .delete_channel(origin, channel, token)
+        .delete_origin_secret(origin, token, key)
         .map_err(Error::DepotClient)?;
 
-    ui.status(Status::Deleted, format!("channel {}.", channel))?;
+    ui.status(Status::Deleted, format!("secret key {}.", key))?;
```
raskchanky: Hmm, I'm a little confused.  Why was `components/hab/src/command/bldr/channel/destroy.rs` changed to destroy origin secrets instead?  Wouldn't it be better to leave the channel stuff alone and put the destroying secrets stuff into its own file?

components/hab/src/command/origin/secret/delete.rs
```diff
@@ -0,0 +1,35 @@
...
+    ui.status(Status::Deleting, format!("secret {}.", key))?;
+
+    depot_client
+        .delete_origin_secret(origin, token, key)
+        .map_err(Error::DepotClient)?;
+
+    ui.status(Status::Deleted, format!("secret {}.", key))?;
+
+    Ok(())
+}
```
raskchanky: Right, here's the file I was thinking would be present for deleting secrets.  I'm not sure what's happening in the channels file I commented on above.

components/hab/src/command/origin/secret/upload.rs
```diff
@@ -0,0 +1,73 @@
...
+    let encrypted_secret_bytes = encryption_key.encrypt(secret.as_bytes(), None)?;
+    let encrypted_secret_string = match String::from_utf8(encrypted_secret_bytes) {
+        Ok(string_from_bytes) => string_from_bytes,
+        Err(_) => {
+            return Err(Error::ArgumentError(
+                "Failed to convert encrypted bytes to string",
+            ))
+        }
+    };
+    ui.status(Status::Encrypted, format!("{}=[REDACTED].", key))?;
```
raskchanky: The `[REDACTED]` is awesome.

raskchanky:
I left a few comments, but on the whole this looks terrific!

components/hab/src/cli.rs
```diff
@@ -264,6 +264,46 @@ pub fn get() -> App<'static, 'static> {
                     (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
                 )
             )
+            (@subcommand secret =>
+                (about: "Commands related to secret management")
+                (@setting ArgRequiredElseHelp)
+                (@subcommand upload => 
+                    (about: "Create and upload a secret for your origin")
+                    (@arg KEY_NAME: +required +takes_value
+                        "The name of the variable key to be injected into the studio.")
```
fnichol: Would a user provide the key prefix or not here? Maybe an example in the usage like we do in a few other places to clarify?

components/hab/src/cli.rs
```diff
@@ -264,6 +264,46 @@ pub fn get() -> App<'static, 'static> {
...
+                    (@arg SECRET: +required +takes_value
```
fnichol: Potential future idea (which might be terrible): I wonder what this would look like if you optionally allowed for `stdin` of the secret value so that you could pipe it in like `command-that-gets-secret | hab origin secret AWS_SECRET`. That's just freethinking though…

components/hab/src/command/bldr/channel/destroy.rs
fnichol: I had the same thought!

components/hab/src/command/origin/key/download.rs
```diff
@@ -137,9 +137,10 @@ fn handle_encryption(
         return Ok(());
     }
 
-    ui.begin(
-        format!("Downloading public encryption origin key for {}", origin),
-    )?;
+    ui.begin(format!(
+        "Downloading public encryption origin key for {}",
```
fnichol: I'm guessing this is simply `rustfmt` reformatting, but now that I read this message, it doesn't really make sense: origin keys aren't about encrypting anything, they're all about signing and verifying things (namely packages). I wonder if taking out the word "encryption" in the origin keys messaging might help to distinguish these 2 concepts in the future.

components/hab/src/command/bldr/channel/destroy.rs
elliott-davis: :whoa: that's definitely a mistake.

components/hab/src/command/origin/key/download.rs
elliott-davis: This function is handling the case specific to origin encryption keys (they were a new concept introduced in a previous PR)

raskchanky:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/353059833).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/353059833).

PR #4713, opened 2018-03-06
use join for x-plat friendly root paths

mwrock:
This ensures that `SVC_ROOT` and `SVC_USER` file path slashes are OS native. Specifically making them backslashes on Windows.

While almost all apps on windows handle forward slashes just as well as backslashes there are a few (like SQL Server 2016 (unlike 2017)) that may not understand forward slashes. This should fix those cases.

Related to https://github.com/habitat-sh/core/pull/12

### DIFF
--- a/components/sup/src/fs.rs
+++ b/components/sup/src/fs.rs
@@ -18,11 +18,11 @@ use hcore::fs::FS_ROOT_PATH;
 lazy_static! {
     /// The root path containing all runtime service directories and files
     pub static ref SVC_ROOT: PathBuf = {
-        Path::new(&*FS_ROOT_PATH).join("hab/svc")
+        Path::new(&*FS_ROOT_PATH).join("hab").join("svc")
     };
 
     pub static ref USER_ROOT: PathBuf = {
-        Path::new(&*FS_ROOT_PATH).join("hab/user")
+        Path::new(&*FS_ROOT_PATH).join("hab").join("user")
     };
 }
 


thesentinels:

Thank you for contributing!

baumanj:
Looks correct to me! I wonder if there are other places we need to address such as https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L80

mwrock:
yeah good call @baumanj . I'll see if I can scour them all up. I have a workaround in the affected plan to tide us over in the meantime.

mwrock:
I took a look and this is the only one in this repo that would affect what gets leaked out of the `{{pkh.svc_*_path}}` variables. The only other path that the supervisor exposes to templates is `PKG_PATH` which is defined in `core` and can be handled in the other PR.

There are certainly other paths with forward slashes constructed as the one you noted but all of those paths are consumed internally and can thus be used with confidence since we don't have arbitrary code that is prejudiced against forward slashes.

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/350081085).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/350081085).

baumanj:
Thanks for taking the extra time and looking into the other possible places, @mwrock !

PR #4738, opened 2018-03-12
Eeyun/sudo fix

### LINKED ISSUE #4716: hab-studio on linux should not require root/sudo when using docker
I want to use docker to enter a studio on a linux system to avoid root/sudo requirements but when I run `hab studio enter -D` I still get prompted by sudo for my password. After some [discussion in the forums](https://forums.habitat.sh/t/how-to-avoid-sudo-when-entering-hab-studio-on-linux/429) it was agreed that this is a bug since my user is a member of the `docker` group and already has appropriate permissions to start a docker container without requiring root/sudo. Can we please get `hab studio enter -D` to work on a linux system for a non-root user?

eeyun:
Fixes #4716 but requires https://github.com/habitat-sh/core/pull/16 to be merged as well.


thesentinels:

Thank you for contributing!

PR #4771, opened 2018-03-16
[common] Add offline mode support for install logic.

fnichol:
While this change started out with the goal of adding offline support to the install logic, I found a few opportunities to set the table for this work with a few refactorings. Then by the end, there were a couple more easy wins, all of which exclusively affect this part of the codebase. As a result, I'm hoping we can look at all of this together and if need be, to look at the specific refactorings, each of which are in their own commit. Since I don't want to bury the lead, the commit message for the offline mode is presented first, with the other commit messages following, however the true ordering is in the commits themselves (if you need the narrative of "what I did and when").

tenor-60645185

---

## [common] Support an "offline" mode for the install logic.

tenor-9852479

This change introduces a new mode for the install logic to use in
situations where contacting a Builder API is not possible or not
allowed. As this code is internal business logic, no external behavior
has (yet) been exposed, which will happen in future work. For now, the
groundwork is laid and is ready to be experimented with, mostly likely
using feature toggles in the `hab` CLI and the Supervisor.

Rationale
---------

There are circumstances and situations where a healthy internet
connection isn't available and others where it is not permitted. For
example:

* devices operating in environments where and internet connection
is supported over a cellular or mesh network
* demonstrating an aspect of Habitat in a hotel, conference, or
internet-less cafe
* in a completely air-gapped environment with no ingress or egress
communication allowed
* a completely offline build environment

An implementation of the install logic that cannot talk to a Builder API
would have to derive its own solution for a "latest" package and have
different failure scenarios, therefore it seemed reasonable to make the
"offline" behavior explicit and opt-in. That way, any tools or components
using this new logic could offer the end user an explicit opt-in to this
behavior so it's understood by all why certain errors must result in an
install failure, etc.

Network Calls in Online Mode
----------------------------

In an effort to understand what an offline implementation would look
like, we should first describe all points at which the normal "online"
logic contacts Builder API endpoints over the network. Then for an
"offline" behavior,  it's a matter of determining what is the reasonable
equivalent and when we must loudly fail.

Currently, the following network calls are made (in order) to a single
Builder API instance (sometimes referred to as the `HAB_BLDR_URL`):

1. If the given package identifier is not fully qualified (referred
hereafter as a "fuzzy" package identifier), we call the `/latest`
endpoint (for example `GET depot/channels/acme/stable/pkgs/foo/latest`
in the case of trying to get the latest for an `acme/foo` package in the
`stable` channel) in order to get back a fully-qualified package
identifier. Note that if the given identifier is fully-qualified we skip
the network call as there can only be *one* possible match for a
fully-qualified identifier.
2. If the target package is not already installed (that is, an extracted
package living in a `/hab/pkgs/<origin>/<name>/<version>/<release>`
path) and if the target artifact (that is, a file ending in `.hart`) is
not already present in the artifact cache, then we call the package
download endpoint (for example `GET
depot/pkgs/acme/foo/1.2.3/20180316084443/download` in the case of trying
to download the `acme/foo/1.2.3/20180316084443` release) in order to get
the artifact into the artifact cache. Note that if the package is either
installed or the artifact is cached then we skip this network call since
artifacts and packages are immutable and will never change (therefore
artifacts are infinitely cacheable).
3. If the target package is not installed, the `TDEPS` metadata file
will be read and used to ensure every corresponding artifact is cached
or installed. For each of these fully-qualified package identifiers, the
same determination is done as in step 2 above.
4. For each package dependency in the target package's `TDEPS` metadata
file, and for the target package itself, the origin's public signing key
is checked to make sure it's cached in the key cache. Each package is
signed with an origin key with a specific revision and so for any
missing public key-with-revision, we call the origin key download
endpoint (for example `GET depot/origins/acme/keys/20180316084443` in
the case of trying to download the `acme-20180316084443.pub` public
origin key) in order to get the public key into the key cache. Note that
if the public key is already in the cache then we skip this network call
since a key-with-revision is also immutable and will never change.

Offline Mode Alternatives to Network Calls
------------------------------------------

Given the network calls above, here is what the offline strategy does,
using the same case numbering for clarity:

1. (resolve fuzzy identifiers) We can only look at what is present on
disk, so there are 3 steps: i) find the latest locally installed
package, if any (no matching packages may be found) and record the
fully-qualified package identifier, ii) find the latest cached artifact,
if any (again, no matching artifacts may be found) and record the
fully-qualified package identifier, and iii) take the latest winner
between the latest installed package and the latest cached artifact. If
we can determine a winner, we can return that fully-qualified package
identifier and continue the logic. If no installed package or cached
artifact can be found, we must return an error reporting that in offline
mode we can't resolve this fuzzy package identifier as no local
candidates were found (offline failure 1a).
2. (download target package) Now that we have a fully-qualified package
identifier, we need to find either this package installed locally or the
corresponding cached artifact. If either are found, the logic continues
as normal. If neither are found, we must return an error reporting that
in offline mode we don't have the package locally and can't continue
(offline failure 2a).
3. (download TDEPS) The same determination must be made for each
runtime-dependent package which is the contents of the `TDEPS` metadata
file in the target package. For each package, step 2 is followed with
the same offline failure 2a returned if any dependency package cannot be
found locally.
4. (download origin key-with-revision) Finally, we need to ensure that
every related package artifact has its public origin key present in the
key cache. As in step 2, if the key is not already present in the key
cache, we must return an error reporting that in offline mode we don't
have the specific origin key and can't continue (offline failure 4a).

Clarifying "Locally Installed Wins"
-----------------------------------

tenor-130410691

In working on this feature, it became clear that an always-present
underlying assumption has been at play with Habitat packages that was at
odds with the flow of this install logic before this change set. It's
the idea that (to date), the concept of what is the "latest" package
includes looking at locally installed packages on disk. This means that
even if the "latest" version is downloaded from a Builder API, it might
be older than a version already installed on a system. Most of the
Habitat tooling and software will favor this newer version and
shadow/mask out the older version--all assuming we're talking about a
fuzzy package identifier.

An example might help make this more concrete. We'll assume we have a
host system running Habitat that has a version of an `acme/foo` package
already installed: `acme/foo/5.0/20180310120105`. We'll also assume that
this package is a service (i.e. we can run `hab start acme/foo`) and
that it contains a standalone binary program called `foobar`. So on
this system when we run `hab start acme/foo`, `hab pkg exec acme/foo
foobar`, or `hab pkg binlink acme/foo foobar`, we know this will resolve
to the `acme/foo/5.0/20180310120105` release of `acme/foo`.

If we were to install an older version of `acme/foo`, let's say
`acme/foo/2.0/20160614130203`, we know that `hab start acme/foo`, `hab
pkg exec acme/foo foobar`, and `hab pkg binlink acme/foo foobar` would
all still resolve to the `acme/foo/5.0/20180310120105` release of
`acme/foo`. In other words, `acme/foo/2.0/20160614130203` would be
"masked" or "shadowed" by `acme/foo/5.0/20180310120105`. If we truly
desired to target the older version of `acme/foo`, we'd need to either
further constrain our fully package identifier (i.e., use `acme/foo/2.0`
instead of `acme/foo`), or provide the fully qualified package
identifier to be 100% explicit.

Now let's say that this local release `acme/foo/5.0/20180310120105` is
presently installed because it is a yet unpublished artifact that is
being tested or that it is a newly built version that is being used to
build another package. Additionally, let's assume that our associated
Builder API has `acme/foo/3.0/20171113025501` as the latest package in
its "stable" channel. The current behavior of the build program would
still pick the local version as "latest", despite downloading
`acme/foo/3.0/20171113025501` and then immediately not use it (again,
this version is older and gets masked out by the newer locally installed
`acme/foo/5.0/20180310120105`).

(Sidenote) While this "locally install wins" behavior might seem strange
in the context on building on first glance, it helped when bootstrapping
the Habitat package system, helps in iterating builds in development,
quickly testing dependencies when regression testing, and has been the
mechanism by which every package has been built in Habitat to date.
There's a good case to be made for altering the concept of "latest" in
the build case, although more work would need to happen to make this
fully realized. However, the work in this changeset shouldn't interfere
with some future "Builder API always wins" semantics. Rather, more
tooling is required like a standalone `hab pkg latest`-style command
which would return a fully qualified package identifier given optional
constraints such as: factoring in locally installed pacakges (or not),
factoring in cached artifacts (or not), exclusively only using the
Builder API result, etc. To make a long story short, this distinction is
on our radar and may be in future work.

In order to make the "locally install wins" a bit more consistent, there
are 2 edge case updates made to the install logic in the default/online
case (again, this assumes we are installing from a fuzzy package
identifier that needs to be resolved):

1. The latest version of an installed package matching the identifier is
determined (if any) before calling out to the Builder API endpoint for
"/latest". If the locally installed version is newer than the latest
available in Builder (again, this request is always scoped to a release
channel), then a status is reported to the user that a newer version was
found locally and will be used. Finally, this is considered a successful
installation in terms of the libarary logic and this would result in a
"clean install" from `hab pkg install`. That is, an exit code of zero
would result on Unix-derived systems. This is a change from the previous
behavior where the known-older version would be downloaded, installed
and then not used as the older version would be masked out. This change
is made partly to save on download bandwith and to inform the end user
that the "latest" is already present locally and did not come from the
Builder API.
2. The second case relates to the first. In this case we still have a
locally installed package matching the identifier, but now there is no
package present in Builder for the specified release channel. Prior to
this change the user would get a warning with a list of package/channel
matches that a user could opt to try instead. This change will use the
locally installed package version (if found) and will report to the user
that it found a version of the package locally and could not find one
remotely. As before, if a locally installed package is not present, the
user would still be presented with the warning and package/channel
recommendations.

Additional Changes
------------------

In the course of supporting this mode, the following
changes/updates/fixes were made:

* A `UI.begin()` message is now displayed to a user when installing a
package from an artifact file. It will look like `Installing
./results/core-wget-static-1.19.1-20180316084443-x86_64-linux.hart` to
make it clearer that we are install from a local artifact and
downloading an artifact remotely.

## [common] Explicitly report latest version checking in install logic.

tenor-188429238

This change adds back channel information back to a user, but in a more
correct location. Prior to this change (and a previous refactoring), the
install logic would report a line like the following:

```
Installing core/redis from channel 'stable'
```

however, no installing would necessarily happen from the stable channel.
Rather, we check for the latest version of a package scoped to a channel
and use that information to determine next steps.

This change adds a status line specifically mapping to the Builder API
call where we check for the latest version and get back the fully
qualified package identifier in a JSON response.

## [common] De-emphasize "channelness" in the install logic.

tenor-252849323

This is an internal refactoring of the install logic to introduce a
simple type representing a release channel. Future work could and should
expand on this type and ultimately extract it into the `core` crate for
more general use.

The goal with this change was to reduce the channel-obsession present in
this module and attempt to keep work related with channels scoped to the
activity of fetching the latest package version and to display channel
recommendations when a package was not found the requested channel. The
internal `InstallTask` type now takes a non-optional `Channel` allowing
the rest of the code to count on a set channel.

## [common] Add `FullyQualifiedPackageIdent` to install logic.

tenor-182726176

This change is an internal refactoring of the install logic which
introduces a new type called `FullyQualifiedPackageIdent` which
represents...a fully qualified package identifier. This allows any code
which requires such an identifier to use a type which guarantees these
promises.

The type uses an internal `Cow` of a `PackageIdent` which allows the
user of this type to hold borrowed and owned idents without additional
cloning. This type helped to update several functions and methods with
clearer expectations and led to less `.is_some()`-style checking.

## [common] Use token if present when installing from archive.

tenor-160359392

This is a reasonably simple fix to the install-from-hart logic to
include an optional Builder API token if provided, allowing any private
dependent packages to be queried and downloaded.

## [common] Canonicalize paths when copying artifact to cache.

tenor-210335693

This change is intended to close any remaining edge conditions occurring
when a user tries to install a `.hart` file directly from the artifact
cache directory. For example:

```
hab pkg install /hab/cache/artifacts/core-redis-*.hart
```

Without a sufficient guard, the install logic will effectively try to do
the following:

```
cp /hab/cache/artifacts/artifact.hart /hab/cache/artifacts/artifact.hart
```

That is, the source file will be copied over and onto itself, usually
resulting in a zero-sized file--obviously not what was expected.

There is currently a check for this in the install logic, however both
the source and target paths are not canonicalized (dereferencing
symlinks, relative paths, etc.). This change fixes that and adds a debug
line in both the `true` and `false` arms of the check for debugging
completeness.

## [hab] Add feature toggle support to the `hab` CLI.

tenor-191424445

This is the same implementation as we have for the Supervisor. An
environment variable prefixed with `HAB_FEAT_` set to true may enable a
feature toggle. To list the current toggles in effect, you can export
`HAB_FEAT_LIST=true` before and `hab` subcommand (including asking for
the usage).

## [hab] Expose offline install behind feature flag.

tenor-92581868

This change exposes the `InstallMode::Offline` behavior out to the `hab
pkg install` subcommand. To activate the behavior, you need to set the
feature toggle environment variable `HAB_FEAT_OFFLINE_INSTALL` to
`true`. This will expose an `--offline` flag in the `hab pkg install`.

Note that the name of this flag is still subject to
change/renaming/removal and should not be considered stable.

For example:

```
 # Set the feature flag for offline installs
> export HAB_FEAT_OFFLINE_INSTALL=true

 # See the `--offline` flag behind the feature flag
> hab pkg install --help
∅ Enabling feature: OfflineInstall
hab-pkg-install
Installs a Habitat package from Builder or locally from a Habitat Artifact

USAGE:
    hab pkg install [FLAGS] [OPTIONS] <PKG_IDENT_OR_ARTIFACT>...

FLAGS:
    -b, --binlink    Binlink all binaries from installed package(s)
    -f, --force      Overwrite existing binlinks
        --offline    Install packages in offline mode

OPTIONS:
    -u, --url <BLDR_URL>       Specify an alternate Builder endpoint. If not specified, the value will be taken from the
    -c, --channel <CHANNEL>    Install from the specified release channel (default: stable)

ARGS:
    <PKG_IDENT_OR_ARTIFACT>...    One or more Habitat package identifiers (ex: acme/redis) and/or filepaths to a
                                  Habitat Artifact (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)

 # Populate the artifact cache using an online install
> sudo -E hab pkg install core/wget-static
∅ Enabling feature: OfflineInstall
» Installing core/wget-static
☁ Determining latest version of core/wget-static in the 'stable' channel
↓ Downloading core/wget-static/1.18/20170514150832
    1.19 MB / 1.19 MB | [========================================] 100.00 % 3.29 MB/s
☛ Verifying core/wget-static/1.18/20170514150832
✓ Installed core/wget-static/1.18/20170514150832
★ Install of core/wget-static/1.18/20170514150832 complete with 1 new packages installed.

 # Remove the installed package
> sudo rm -rf /hab/pkgs/core/wget-static

 # Install the same package, this time without contacting a Builder API
> sudo -E hab pkg install --offline core/wget-static
∅ Enabling feature: OfflineInstall
» Installing core/wget-static
☁ Determining latest version of core/wget-static locally installed or cached (offline)
☛ Verifying core/wget-static/1.18/20170514150832
✓ Installed core/wget-static/1.18/20170514150832
★ Install of core/wget-static/1.18/20170514150832 complete with 1 new packages installed.

 # Try the install again, and see that the package is already installed
> sudo -E hab pkg install --offline core/wget-static
∅ Enabling feature: OfflineInstall
» Installing core/wget-static
☁ Determining latest version of core/wget-static locally installed or cached (offline)
→ Using core/wget-static/1.18/20170514150832
★ Install of core/wget-static/1.18/20170514150832 complete with 0 new packages installed.
```


thesentinels:

Thank you for contributing!

chefsalim:
@fnichol Just a couple of notes - 1) there are also now a couple of cards for supporting this behavior  in hab supervisor and studio so hopefully some of those needs we'll get for free.  2) Also, I believe we will want to attempt to check the cache even if a fully qualified ident is not specified.

For reference:
https://app.zenhub.com/workspace/o/habitat-sh/habitat/issues/4762
https://app.zenhub.com/workspace/o/habitat-sh/habitat/issues/4763
https://app.zenhub.com/workspace/o/habitat-sh/habitat/issues/4764
https://app.zenhub.com/workspace/o/habitat-sh/habitat/issues/4765

fnichol:
1) Agreed that this will help those issues and the broader feature of running parts of Habitat with and without a Builder/network connectivity. This PR won't go beyond preparing the install logic and exposing `hab pkg install` to the first pass of this. That should allow me (and others) to get an idea about how some parts might work, if we chose the right name, how this would interact with the Studio, etc.

2) This change actually does that! Normally with an internet connection, our logic makes a "latest_pkg_ident_for" call to the Builder API, and this offline logic replaces that with looking for the latest version matching that ident that is either a) an installed package (i.e. the latest `PackageInstall`) or b) a cached artifact (i.e. the latest `PackageArchive`). A nice side effect of this work, is that the same check will be performed (most likely) before making the network call to the Builder API to ensure that even if no published package exists, a locally found version will still count as successful.

I still have notes to write around this, clearly, but it's a great start.

components/common/src/command/package/install.rs
```diff
@@ -229,6 +260,7 @@ impl<'a> InstallTask<'a> {
     ) -> Result<Self> {
         Ok(InstallTask {
             depot_client: Client::new(url, product, version, Some(fs_root_path))?,
+            channel: channel,
```
baumanj: Can this (and the lines around it) make use of [field init shorthand](https://doc.rust-lang.org/book/second-edition/ch05-01-defining-structs.html#using-the-field-init-shorthand-when-variables-and-fields-have-the-same-name)?

components/common/src/command/package/install.rs
```diff
@@ -220,15 +314,19 @@ struct InstallTask<'a> {
...
         product: &str,
         version: &str,
         fs_root_path: &'a Path,
         artifact_cache_path: &'a Path,
         key_cache_path: &'a Path,
     ) -> Result<Self> {
         Ok(InstallTask {
+            install_mode: install_mode,
             depot_client: Client::new(url, product, version, Some(fs_root_path))?,
+            channel: channel,
```
baumanj: Can this (and the lines around it) make use of [field init shorthand](https://doc.rust-lang.org/book/second-edition/ch05-01-defining-structs.html#using-the-field-init-shorthand-when-variables-and-fields-have-the-same-name)?

raskchanky: It's definitely clever, but imo, field init shorthand makes the code harder to read.  Just a personal preference, though.

components/common/src/command/package/install.rs
```diff
@@ -185,6 +269,7 @@ pub fn start<P1, P2>(
     fs_root_path: P1,
     artifact_cache_path: P2,
     token: Option<&str>,
+    install_mode: &InstallMode,
```
baumanj: Is there a need for `install_mode` to be a reference rather than a `Copy` type? Since it's a small simple type, using a reference implies to me that it's modified somewhere and we want those modifications to be shared by the various places references are held, but I don't see that that is the case.

components/common/src/ui.rs
```diff
@@ -74,10 +75,11 @@ impl Status {
             Status::Deleting => ('☒', "Deleting".into(), Colour::Green),
             Status::Demoted => ('✓', "Demoted".into(), Colour::Green),
             Status::Demoting => ('→', "Demoting".into(), Colour::Green),
-            Status::Determining => ('→', "Determining".into(), Colour::Green),
+            Status::Determining => ('☁', "Determining".into(), Colour::Green),
```
baumanj: I love the ☁️ as the glyph for determining, but I wonder if it might be misleading since we use `Status::Determining` in some offline contexts.

components/common/src/command/package/install.rs
```diff
@@ -339,52 +373,136 @@ impl<'a> InstallTask<'a> {
...
+            match self.latest_installed_or_cached(&ident) {
+                Ok(i) => Ok(i),
+                Err(Error::PackageNotFound) => {
+                    return Err(Error::OfflinePackageNotFound(ident.clone()));
+                }
+                Err(e) => return Err(e),
+            }
+        } else {
+            // Otherwise, we're online and we have a fuzzy package identifier. Now we can find the
+            // latest identifier from any installed pacakges and from a Builder API.
```
baumanj: "pacakges"

components/common/src/command/package/install.rs
```diff
@@ -339,52 +373,136 @@ impl<'a> InstallTask<'a> {
...
+        token: Option<&str>,
+    ) -> Result<FullyQualifiedPackageIdent> {
+        if ident.fully_qualified() {
+            // If we have a fully qualified package identifier, then our work is done--there can
+            // only be *one* package that satisfies a fully qualified identifier.
+
+            FullyQualifiedPackageIdent::from(ident)
+        } else if self.is_offline() {
+            // If we can't contact a Builder API, then we'll find the latest installed package or
+            // cached artifact that satisfies the fuzzy package identifer.
```
baumanj: "identifer"

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
     /// Adapter function to retrieve an installed package given an
     /// identifier, if it exists.
-    fn installed_package(&self, ident: &PackageIdent) -> Option<PackageInstall> {
-        PackageInstall::load(ident, Some(self.fs_root_path)).ok()
+    fn installed_package(&self, ident: &FullyQualifiedPackageIdent) -> Option<PackageInstall> {
+        PackageInstall::load(ident.as_ref(), Some(self.fs_root_path)).ok()
+    }
+
+    /// Checks for the latest installed package or cached artifact that matches a given package
+    /// identifier and returns a fully qualified package identifer if a match exists.
```
baumanj: "identifer"

components/common/src/command/package/install.rs
```diff
@@ -154,6 +157,87 @@ impl AsRef<PackageIdent> for InstallSource {
...
+}
+
+impl<'a> fmt::Display for Channel<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Represents a fully-qualified Package Identifier, meaning that the normally optional version and
+/// release package coordinates are guarenteed to be set. This fully-qualified-ness is checked on
```
baumanj: "guarenteed"

components/common/src/command/package/install.rs
baumanj: If that's the team consensus, that would be fine. As with most things style, I think the lion's share of the benefit comes from consistency which reduces cognitive load.

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
+        debug!("offline mode: winner: {:?}", &latest);
+
+        Ok(latest)
     }
 
-    // TODO (CM): This could return a plain bool IF we could ensure
-    // above that the package identifier is FULLY QUALIFIED
-    fn is_artifact_cached(&self, ident: &PackageIdent) -> Result<bool> {
-        Ok(self.cached_artifact_path(ident)?.is_file())
+    fn latest_installed_ident(&self, ident: &PackageIdent) -> Result<FullyQualifiedPackageIdent> {
```
baumanj: Since we're not using the `Err` for anything that I can see (nor `?` shortcuts), would it be more evocative to make the return an `Option` rather than a `Result`? To me the answer to the question `latest_installed_ident` may well be `None`, but I don't see that as an error _per se_.

If this would be too disruptive, it's not a big deal, but I mention it because I think there is value in the distinction between something that may be absent in normal behavior versus an error state.

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
+            if ident.version.is_none() {
+                ident.version = Some(String::from("_VERSION_"));
+            }
+            if ident.release.is_none() {
+                ident.release = Some(String::from("_RELEASE_"));
+            }
+            match ident.archive_name() {
+                Some(name) => name.replace("_VERSION_", "?*").replace("_RELEASE_", "?*"),
+                None => return Err(Error::PackageNotFound),
+            }
```
baumanj: I must be missing something. Why wouldn't this be
```rust
            if ident.version.is_none() {
                ident.version = Some(String::from("?*"));
            }
            if ident.release.is_none() {
                ident.release = Some(String::from("?*"));
            }

            match ident.archive_name() {
                Some(name) => name
                None => return Err(Error::PackageNotFound),
            }
```
?

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
+        };
+        let glob_path = self.artifact_cache_path.join(filename_glob);
+        let glob_path = glob_path.to_string_lossy();
```
baumanj: There definitely shouldn't be any non-unicode sequences here, right? Would
```rust
        let glob_path = glob_path.to_str().unrwap();
```
be safer and obviate the need for the `as_ref()`s later?

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
+                if latest.is_empty() {
+                    latest.push((artifact_ident, artifact));
+                } else {
+                    if artifact_ident > latest[0].0 {
+                        let _ = latest.pop();
+                        latest.push((artifact_ident, artifact));
+                    }
+                }
+            }
+        }
```
baumanj: I'm confused by the use of the `Vec` here. Would an `Option` not be preferable?
```rust
        let mut latest: Option<(PackageIdent, PackageArchive)> = None;
        for …
        {
            let mut artifact = PackageArchive::new(&file);

            if let Ok(artifact_ident) = artifact.ident() {
                if artifact_ident.origin == ident.origin && artifact_ident.name == ident.name {
                    if latest.is_none() || artifact_ident > latest.as_ref().unwrap().0 {
                        latest = Some((artifact_ident, artifact));
                    }
                }
            }
        }
```

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
+
+        if latest.is_empty() {
+            Err(Error::PackageNotFound)
+        } else {
+            Ok(FullyQualifiedPackageIdent::from(
+                latest.pop().unwrap().1.ident()?,
+            )?)
+        }
```
baumanj: Related to [the previous comment](https://github.com/habitat-sh/habitat/pull/4771/files#r175818297), an `Option` would allow you to avoid some of the `pop().unrwap()`-type stuff:
```rust
        latest.map_or(Err(Error::PackageNotFound), |mut l| {
            Ok(FullyQualifiedPackageIdent::from(l.1.ident()?)?)
        })
```
though, if you find the `if`/`else` easier to read, that still saves the `pop()`:
```rust
        if latest.is_none() {
            Err(Error::PackageNotFound)
        } else {
            Ok(FullyQualifiedPackageIdent::from(
                latest.unwrap().1.ident()?,
            )?)
        }
```

raskchanky:
Is this PR even ready for review at this point?  The description is a placeholder and all the commits say "wip".

baumanj:
Why create a PR if not for feedback?

raskchanky:
I've done it in the past to allow easy collaboration between multiple people, like when @cnunciato and I were working on oauth stuff.  Ultimately though, it doesn't really matter why someone creates a PR.  I thought @fnichol 's description made it pretty clear that this was a work in progress.  I think feedback is great, but my main concern is we've approved changes that aren't in their final form.

baumanj:
Fair point. I'm curious, do "approved" reviews get reset if the PR is changed?

raskchanky:
I don't believe so, but I'm not 100% sure.

components/common/src/command/package/install.rs
fnichol: Yeah, I could have but I tend to preserve the same style of the surrounding code where possible. In this case I did consider it, but I didn't want to change the entire function call to make it more modern (and consistent) and didn't want to have 2 styles going on (older and modern but inconsistent).

If this was all new code however, I'd likely do just that!

components/common/src/command/package/install.rs
fnichol: Yep, a copy would work just as well, but I was trying to keep the function arguments more consistent in that they are almost all immutable borrows. Currently the `enum` is zero-sized, but at some future point it might contain something and I figured I may as well leave it in a borrow-first state.

I also would read this type in a function as a sign that I shouldn't try to change this value downstream, whereas with a copied value it's a little more open to interpretation.

But yes, the copy would work fine here as well.

components/common/src/ui.rs
fnichol: I looked at where this type is used and it looks like it was added for Builder job status subcommands. I was thinking of less an internet cloud, but rather a thought bubble representing thinking about something 😄 

One takeaway when I changed this was to notice that some of the other types of iconography could be better represented (like promoted (up?), demoted (down?)). I might look at that afterwards as a separate change (it's on my whiteboard right now and doesn't hurt anything in the meantime).

components/common/src/command/package/install.rs
```diff
@@ -502,43 +625,139 @@ impl<'a> InstallTask<'a> {
...
+                    pkg_install
+                } else {
+                    pkg_artifact
+                }
+            }
+            (Err(_), Err(_)) => return Err(Error::PackageNotFound),
+        };
+        debug!("offline mode: winner: {:?}", &latest);
+
+        Ok(latest)
```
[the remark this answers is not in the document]
fnichol: I think that should be the same. The reason I'm saying "think" however is that it takes me a lot longer to reason through the compound conditional than to scan through the `match` arms. That's probably a personal style thing though--for me I tend to use `match`ing where possible because the compiler makes sure I don't miss any arms (it helped in an earlier version of this code where I was dealing with `Option`s and had to convert to `Result`s). Same for that inner `if`/`else`--I'm trying to get a fully balanced conditional so that I can see every condition has been dealt with.

components/common/src/command/package/install.rs
fnichol: Oh, this one! Yes maybe I should update that--I was being needlessly careful. When I saw the lack of validation present in `PackageIdent` I got nervous that at some future time we might want to apply rules to the `version` and `release` fields in a package identifier--like a release string should only have a set number of numeric characters. It was with this thought in mind that I put what I though would be a safer token to be replaced in the struct and then replaced over it once I had a `String` to work with.

I think I'll update to your suggestion and leave a note that if any future compile failures or issues arise it might be from stuffing arbitrary strings into these fields.

components/common/src/command/package/install.rs
fnichol: The `match` in the method is where we're handling any `Err`s from `PackageInstall::load()` (which returns a `Result<PackageInstall>`).

components/common/src/command/package/install.rs
fnichol: There's a super low chance of hitting any non-Unicode chars, correct. I usually use the [`PathBuf.to_string_lossy()`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.to_string_lossy) in these situations as the result is a `Cow<str>` which delays any further allocations until necessary. If we hit a non-Unicode character, it's true that this code would add those strange `�` characters in the resulting string, but we also wouldn't thread panic and in the worst case not find a cached artifact. Since we don't allow non-Unicode characters in the package identifiers (in the `PackageInstall` `FromStr` impl), we're left with the filesystem path having one of these characters. This is one place where I'd rather let this code work through rather than panicking and having to cope with a source location and an unknown unwrap call.

components/common/src/command/package/install.rs
fnichol: Haha, yeah I agree! I was getting pretty tired by the time I finished this method up and wasn't totally happy with it. I started with an `Option<(PackageIdent, PackageArchive)>` but either got into code obsession with `None` or an ownership issue. What I really wanted was something you could do in C: allocate the memory for the slot, and keep the "winner" as you go. That's what led me to a memory location of size 1, which was a Vec with stack semantics 😄 

I'll give this a few minutes to rework and see if I can get it clearer!

components/common/src/ui.rs
baumanj: That sounds like a good approach. If we want this to be a thought balloon, there is an emoji for that: 💭. That sounds like a good change to make.

Or if we really want to go all-in: 🤔

👍 on a separate change to clean these up more generally. It may seem like a trivial thing, but I feel it has an impact un usability.

components/common/src/command/package/install.rs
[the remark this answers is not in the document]
christophermaier: 👍 on using match... it's a lot easier to reason through, and you can be sure that the compiler has your back.

Incidentally, you'll find this same pattern of matching against a tuple in Erlang code in place of complex if/else logic, and for the same reason. It takes a little getting used to at first, but I think it's ultimately clearer.

components/common/src/command/package/install.rs
christophermaier: I'm mildly nervous about stuffing regex strings into something that should be a version or a release. Could we not just assemble the regex string directly?

components/common/src/command/package/install.rs
baumanj: Right, I'm wasn't suggesting changing the return type of `PackageInstall::load`, just changing the return type of `latest_installed_ident`. I tried it out a bit and I don't think it's actually much of a win. And being consistent with always using `Result<PackageInstall>` is probably best.

components/common/src/command/package/install.rs
christophermaier: Can you use `Option` with a helping of `get_or_insert` and/or `take`? I think that'd effectively allow you to do what you want, with clearer semantics than the `Vec` approach.

Also, since `PackageArchive` can give you an identifier, do you need to keep both? Would it be enough to just hang on to the `PackageArchive`?

components/common/src/command/package/install.rs
christophermaier: Oooh, I like that.

components/common/src/command/package/install.rs
baumanj: Would the code I suggested work?

components/common/src/command/package/install.rs
```diff
@@ -298,7 +298,7 @@ where
 
     match *install_source {
         InstallSource::Ident(ref ident) => task.from_ident(ui, ident.clone(), token),
-        InstallSource::Archive(ref local_archive) => task.from_archive(ui, local_archive),
+        InstallSource::Archive(ref local_archive) => task.from_archive(ui, local_archive, token),
```
christophermaier: With hindsight, this is completely obvious; nice catch!

components/common/src/command/package/install.rs
```diff
@@ -807,14 +807,29 @@ impl<'a> InstallTask<'a> {
         ident: &FullyQualifiedPackageIdent,
         artifact_path: &Path,
     ) -> Result<()> {
-        let cache_path = self.cached_artifact_path(ident);
+        // Canonicalize both paths to ensure that there aren't any symlinks when comparing them
+        // later.
+        let artifact_path = artifact_path.canonicalize()?;
+        let cache_path = self.cached_artifact_path(ident).canonicalize()?;
```
christophermaier: tenor-48006762
... to canonicalize both paths!

components/hab/src/main.rs
```diff
@@ -942,7 +946,10 @@ fn install_sources_from_matches(matches: &ArgMatches) -> Result<Vec<InstallSourc
 }
 
 fn enable_features_from_env(ui: &mut UI) {
-    let features = vec![(feat::List, "LIST")];
+    let features = vec![
+        (feat::List, "LIST"),
+        (feat::OfflineInstall, "OFFLINE_INSTALL"),
+    ];
```
christophermaier: Is there an alternative way to represent the features we're using? We've got the bitfield declared in one place, and the "meaning" of the flags defined here. Might be nice to have, say, a `features.rs` module where everything is consolidated in one easy-to-understand place.

components/common/src/command/package/install.rs
christophermaier: @baumanj Maybe... not sure if the code issues / ownership issues @fnichol ran into would be bypassed by your implementation. At first glance, though, I _think_ it'd work. 

I may have been trying to be too clever with the `get_or_insert`/`take` stuff 😄

components/common/src/command/package/install.rs
```diff
@@ -523,12 +529,24 @@ impl<'a> InstallTask<'a> {
...
-                .show_package(ident, channel, token)?
+                .show_package(ident, Some(self.channel.0), token)?
+                .into(),
+        )
+    }
+
+    fn fetch_latest_pkg_ident_in_channel_for(
+        &self,
+        ident: &PackageIdent,
+        channel: &str,
```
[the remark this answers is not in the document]
fnichol: Oh yeah, forgot! Will rebase this in.

components/common/src/command/package/install.rs
```diff
@@ -359,7 +368,7 @@ impl<'a> InstallTask<'a> {
         };
 
         for channel in channels {
-            match self.fetch_latest_pkg_ident_for(ident, Some(&channel), token) {
+            match self.fetch_latest_pkg_ident_in_channel_for(ident, &channel, token) {
```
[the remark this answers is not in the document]
fnichol: The `self.depot_client.list_channels(ident.origin(), false)` would return a `Vec<String>` but I could map/convert it. Will try…

components/common/src/command/package/install.rs
```diff
@@ -339,6 +325,29 @@ impl<'a> InstallTask<'a> {
...
+    fn recommend_channels(
+        &self,
+        ui: &mut UI,
+        ident: &PackageIdent,
+        token: Option<&str>,
+    ) -> Result<()> {
+        if let Ok(recommendations) = self.get_channel_recommendations(&ident, token) {
+            if !recommendations.is_empty() {
+                ui.warn(
+                    "The package does not have any versions in the specified channel.",
```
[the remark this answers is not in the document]
fnichol: I think I dealt with this in another commit; it displays `"No releases of {} exist in the '{}' channel"`

components/common/src/command/package/install.rs
```diff
@@ -177,6 +178,52 @@ impl<'a> fmt::Display for Channel<'a> {
     }
 }
 
+/// Represents a fully-qualified Package Identifier, meaning that the normally optional version and
+/// release package coordinates are guaranteed to be set. This fully-qualified-ness is checked on
+/// construction and as the underlying representation is immutable, this state does not change.
+struct FullyQualifiedPackageIdent<'a> {
```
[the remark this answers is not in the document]
fnichol: woot!

components/common/src/command/package/install.rs
```diff
@@ -177,6 +178,52 @@ impl<'a> fmt::Display for Channel<'a> {
...
+    // The ident is a struct field rather than a "newtype" struct to ensure its value cannot be
+    // directly accessed
+    ident: Cow<'a, PackageIdent>,
+}
+
+impl<'a> FullyQualifiedPackageIdent<'a> {
+    fn from<I>(ident: I) -> Result<Self>
```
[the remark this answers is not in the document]
fnichol: Good idea

components/common/src/command/package/install.rs
```diff
@@ -177,6 +178,52 @@ impl<'a> fmt::Display for Channel<'a> {
...
+        } else {
+            Err(Error::HabitatCore(
+                hcore::Error::FullyQualifiedPackageIdentRequired(
+                    ident.to_owned().to_string(),
+                ),
+            ))
+        }
+    }
+
+    fn archive_name(&self) -> String {
```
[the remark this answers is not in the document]
fnichol: Yeah--it probably makes sense to look at `PackageIdent` as an enum in a future world meaning some of the methods would sort themselves better.

components/common/src/command/package/install.rs
```diff
@@ -380,10 +427,10 @@ impl<'a> InstallTask<'a> {
     /// Given an archive on disk, ensure that it is properly installed
     /// and return the package's identifier.
     fn from_archive(&self, ui: &mut UI, local_archive: &LocalArchive) -> Result<PackageInstall> {
-        let ref ident = local_archive.ident;
-        match self.installed_package(ident) {
+        let ident = FullyQualifiedPackageIdent::from(&local_archive.ident)?;
```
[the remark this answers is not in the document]
fnichol: Yeah!

components/common/src/command/package/install.rs
```diff
@@ -422,10 +469,16 @@ impl<'a> InstallTask<'a> {
...
-                    if self.installed_package(dependency).is_some() {
+                    if self.installed_package(&FullyQualifiedPackageIdent::from(dependency)?)
+                        .is_some()
+                    {
                         ui.status(Status::Using, dependency)?;
                     } else {
-                        artifacts_to_install.push(self.get_cached_artifact(ui, dependency, token)?);
+                        artifacts_to_install.push(self.get_cached_artifact(
+                            ui,
+                            &FullyQualifiedPackageIdent::from(dependency)?,
```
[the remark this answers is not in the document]
fnichol: I added a note for this

components/common/src/command/package/install.rs
```diff
@@ -508,55 +561,52 @@ impl<'a> InstallTask<'a> {
...
     }
 
     /// Retrieve the identified package from the depot, ensuring that
     /// the artifact is cached locally.
-    fn fetch_artifact(&self, ui: &mut UI, ident: &PackageIdent, token: Option<&str>) -> Result<()> {
+    fn fetch_artifact(
+        &self,
+        ui: &mut UI,
+        ident: &FullyQualifiedPackageIdent,
+        token: Option<&str>,
```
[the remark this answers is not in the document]
fnichol: Yes, totally agree!

components/common/src/command/package/install.rs
fnichol: Figured I'd go full paranoid here

fnichol:
I don't know why, but I'm getting a bad CI build and it's failing to load the logs. I'll give Travis a sec, but am hoping too see the .1 and .2 Rust jobs come back green.

fnichol:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/357265707).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/357265707).

PR #4786, opened 2018-03-20
Pin core/docker to 17.09.9 for cfize and dockerize

elliott-davis:
We must pin here since these two packages are dependent. If Cfize builds, it pulls a newer version of core/docker than the current hab-pkg-dockerize has so we get a dep conflict.


thesentinels:

Thank you for contributing!

christophermaier:
tenor-122936424

PR #4797, opened 2018-03-22
always use stable windows-service and fail build on failure

mwrock:
The `windows-service` which hosts the supervisor as a windows service should always be pulled from the stable channel. Also, if the install fails of the studio build fails, we should completely fail the build.

### DIFF
--- a/components/studio/build-docker-image.ps1
+++ b/components/studio/build-docker-image.ps1
@@ -23,14 +23,24 @@ try {
     $env:HAB_BINLINK_DIR = $null
     
     info "Installing and extracting initial Habitat packages"
+    # Generally, we want to use packages that are aligned with the channel
+    # set when building the image which is usually unstable or an rc channel
+    # in our pipelines. Because the windows-service package is maintained
+    # in another repo and on a slower deployment cadence, we will always
+    # install from its stable channel
     if(!$InstallHarts) {
         $InstallHarts = @(
             "core/hab-studio",
             "core/hab-sup",
-            "core/windows-service"
+            "core/windows-service --channel stable"
         )
     }
-    $InstallHarts | % { hab pkg install $_ }
+    $InstallHarts | % {
+        hab pkg install $_
+        if ($LASTEXITCODE -ne 0) {
+            Write-Error "hab install failed for $_, aborting"
+        }
+    }
     $studioPath = hab pkg path core/hab-studio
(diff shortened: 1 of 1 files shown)


components/studio/build-docker-image.ps1
```diff
@@ -27,10 +27,15 @@ try {
         $InstallHarts = @(
             "core/hab-studio",
             "core/hab-sup",
-            "core/windows-service"
```
christophermaier: Do `hab-studio` and `hab-sup` not also need to be pulled from `stable`, too? If not, what's the difference between how these packages are used?

mwrock: No we want those versions to align with whatever channel we are building (unstable or rc). `windows-service` is different because it is maintained in a separate repo and changes very infrequently. Until this last release it was not used by any of our core code but was instead a sort of "add-on" that windows users could use to harness the launcher/supervisor as a windows service.

One option would be to bring it in to this repo but I don't see that as particularly high value yet.

christophermaier: Ah, gotcha. That makes sense. Could you basically paste that response into the code as a comment so we keep that rationale around? I could see someone accidentally changing that in some future refactoring, absent that information.

components/studio/build-docker-image.ps1
```diff
@@ -23,14 +23,23 @@ try {
     $env:HAB_BINLINK_DIR = $null
     
     info "Installing and extracting initial Habitat packages"
+    # Generally, we want to use packages that are aligned with the channel
+    # set when building the image which is usually unstable or an rc channel
+    # in our pipelines. Because the windows-service package is maintained
+    # in another repo and on a slower deployment cadence
```
christophermaier: Looks like you left off the last bit of this sentence.

mwrock: oopsies. added it

christophermaier:
tenor-74768137

mwrock:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/habitat/builds/357053138).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/habitat/builds/357053138).

PR #4813, opened 2018-03-27
Make `sed -i` work on both linux and Mac OS

baumanj:
Because the Mac OS version of sed's -i option requires an argument but linux's
does not, the way to get inline ovewrite behavior on both is to supply a zero-
length argument with no spaces after the -i.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -52,10 +52,10 @@ If you happen to be aware of features or bugs and feel there are user impacting 
     $ git checkout -b <branch>
     ```
 
-1. Remove the `-dev` suffix from the version number found in the `VERSION` file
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

baumanj:
> I would prefer this change to only encapsulate changes that could be run on a mac and linux.

Which ones are those, @elliott-davis? Looking just now, I kind of thought all of these could be, but if I'm mistaken, let me know.

elliott-davis:
@baumanj it's the two that touch `plan-build`

baumanj:
Ah, I did not realize that would never run natively on Mac OS.

PR updated to remove it.

PR #4823, opened 2018-03-28
Formalize template rendering context

### LINKED ISSUE #4761: Formalize the contents of the template rendering context
In the course of working on https://github.com/habitat-sh/habitat/pull/4760, a number of shortcomings in the way we populate our template rendering context became obvious.

First, it is basically a JSON serialization of a number of internal Rust data types, which are spread out across several modules. Making a change to the public fields of these types will immediately change the contents of the rendering context, limiting our ability to refactor them. Furthermore, these types often are being used for other purposes, which unnecessarily couples those purposes to template rendering. Additionally, it's not exactly clear what the use of some of the data is for templating, which is probably a byproduct of the serialization of internal data types, rather than ones specifically dedicated to templating.

Second, since we're just relying on a basic JSON serialization of these various types, we end up with a number of inconsistencies in the data we present. The data itself is not inconsistent, but the naming of semantically identical information is different, or it is accessed via a different data path, or two objects that should be identical have a different set of fields, depending on where they were sourced from. 

Third, documenting the contents is a bit cumbersome, as it's basically done manually and in plain text on the website.

I propose that we explicitly control what data is put into the rendering context, instead of just serializing internal types. This would allow us to curate a consistent collection of information, with a structure that makes sense for ease of templating (this is particularly important given our choice of Mustache / Handlebars as a templating language). With this level of control over the contents of the rendering context, we could also document it using something like JSON Schema, which would be much more complete and structured than our current ad-hoc, free-form documentation. That would also open the door to easy automated testing and validation of our templating code.

christophermaier:
Previously, the data we made available to users in Habitat templates
was based directly on the JSON serialization of several internal data
structures. This needlessly coupled internal workings of Habitat with
the public interface we expose to users. This often made it difficult
to know exactly what data was presented to users, and made it
difficult to safely change that data.

Here, we explicitly decouple the internal data structures of the
Supervisor from what is exposed to users. To do this in an efficient
and clear way, we introduce a number of "templating proxies", which
effectively wrap internal data structures, holding on to their
relevant data using `Cow` types to reduce memory footprint (this also
makes it easier to instantiate the templating proxies in tests, since
we can simply provide owned copies of the important data). Not only
does this give us a clear, typed way to refer to the data, but it
allows us to implement `Serialize` on these types, which lets us
control what is exposed to users in a very fine-grained and targeted
way. For instance, `{{svc.me.persistent}}` should really be
`{{svc.me.permanent}}`; before this change, we would need to introduce
another field to `census::CensusMember` named `permanent`, and it
would be unclear why. Now, however, we can simply modify the
templating proxy's `Serialize` implementation to render that same
field multiple times under different names, making schema evolution
and deprecation much easier to reason about. Note, however, that no
user-facing changes are introduced in this commit; it is a pure
refactoring.

By clearly separating out the templating data, we now have an
opportunity to revisit some of the internal data structures that were
previously serving "double duty"; it may be possible to simplify them
now.

Additionally, this commit introduces a JSON Schema definition of the
entire render context. This is currently used in tests to validate our
code, but could be used in the future as a source for generated
documentation on the website, and possibly as part of a future
template validation tool.

Fixes #4761

### DIFF
--- a/components/sup/src/manager/service/package.rs
+++ b/components/sup/src/manager/service/package.rs
@@ -33,13 +33,22 @@ const PATH_KEY: &'static str = "PATH";
 static LOGKEY: &'static str = "PK";
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct Env(pub HashMap<String, String>);
+pub struct Env(HashMap<String, String>);
 
 impl Deref for Env {
     type Target = HashMap<String, String>;
 
     fn deref(&self) -> &Self::Target {
         &self.0
+    }
+}
+
+// Convenience implementation for ease of setting up test instances;
+// real code should use `Env::new`
+#[cfg(test)]
+impl From<HashMap<String, String>> for Env {
+    fn from(inner_map: HashMap<String, String>) -> Self {
+        Env(inner_map)
     }
 }
 
--- a/components/sup/src/manager/service/package.rs
+++ b/components/sup/src/manager/service/package.rs
@@ -33,7 +33,7 @@ const PATH_KEY: &'static str = "PATH";
 static LOGKEY: &'static str = "PK";
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct Env(HashMap<String, String>);
+pub struct Env(pub HashMap<String, String>);
 
 impl Deref for Env {
     type Target = HashMap<String, String>;
@@ -79,16 +79,6 @@ impl Env {
     }
 }
 
-// NOTE: This is exposed to users in templates. Any public member is
-// accessible to users, so change this interface with care.
-//
-// User-facing documentation is available at
-// https://www.habitat.sh/docs/reference/#template-data; update that
-// as required.
-//
-// TODO (CM): move templatable content into a distinct type to
-// separate these concerns; Pkg is currently also used for
-// non-templating purposes, as well.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Pkg {
     #[serde(deserialize_with = "deserialize_using_from_str",
--- a/components/sup/src/templating/context.rs
+++ b/components/sup/src/templating/context.rs
@@ -141,13 +141,13 @@ impl<'a> RenderContext<'a> {
 struct SystemInfo<'a> {
     version: Cow<'a, String>,
     member_id: Cow<'a, String>,
-    ip: IpAddr,
+    ip: Cow<'a, IpAddr>,
     hostname: Cow<'a, String>,
-    gossip_ip: IpAddr,
-    gossip_port: u16,
-    http_gateway_ip: IpAddr,
-    http_gateway_port: u16,
-    permanent: bool,
+    gossip_ip: Cow<'a, IpAddr>,
+    gossip_port: Cow<'a, u16>,
+    http_gateway_ip: Cow<'a, IpAddr>,
+    http_gateway_port: Cow<'a, u16>,
+    permanent: Cow<'a, bool>,
 }
 
 impl<'a> SystemInfo<'a> {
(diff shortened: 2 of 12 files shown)


components/sup/src/manager/service/package.rs
```diff
@@ -33,7 +33,7 @@ const PATH_KEY: &'static str = "PATH";
 static LOGKEY: &'static str = "PK";
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct Env(HashMap<String, String>);
+pub struct Env(pub HashMap<String, String>);
```
reset: Scanning other changes I can't see why this was made public. Btw, this struct does implement `Deref<Target = HashMap<String, String>>` so you could call `deref()` to get at it's insides if you need it!

christophermaier: It was actually the opposite; I have a `HashMap`, but I needed to turn it into an `Env` for testing, without having to mock out an installed package with a `RUNTIME_ENVIRONMENT` file.

Should have implemented `From`, I suppose.

reset: That's what I'd suggest, yep!

components/sup/src/templating/context.rs
```diff
@@ -95,81 +112,805 @@ impl<'a> RenderContext<'a> {
...
-    pub update_election_is_finished: bool,
-    pub me: SvcMember<'a>,
-    pub first: SvcMember<'a>,
-    pub members: Vec<SvcMember<'a>>,
-    pub leader: Option<SvcMember<'a>>,
-    pub update_leader: Option<SvcMember<'a>>,
+struct SystemInfo<'a> {
+    version: Cow<'a, String>,
+    member_id: Cow<'a, String>,
+    ip: IpAddr,
```
baumanj: Why are the `IpAddr`s, ports and `permanent` not `Cow` types?

components/sup/src/templating/context.rs
```diff
@@ -95,81 +112,805 @@ impl<'a> RenderContext<'a> {
...
+        }
+    }
+}
+
+impl<'a> Serialize for Package<'a> {
+    fn serialize<S>(&self, serializer: S) -> result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(20))?;
```
baumanj: I'm guessing there's not any good way to derive this `20` or even catching that it needs updating if the `Package` struct changes.

components/sup/src/templating/context.rs
```diff
@@ -95,81 +112,805 @@ impl<'a> RenderContext<'a> {
...
-            members: census_group
+            service_group: Cow::Borrowed(&census_group.service_group),
+            election_status: Cow::Borrowed(&census_group.election_status),
+            update_election_status: Cow::Borrowed(&census_group.update_election_status),
+            members: Cow::Owned(
+                census_group
+                    .members()
+                    .iter()
+                    .map(|m| SvcMember::from_census_member(m))
+                    .collect::<Vec<SvcMember<'a>>>(),
```
baumanj: This seems to build with just
```rust
                    .collect(),
```
Presumably the `from_census_member` is a strong enough hint for the type-inference to work. Is it valuable to be more explicit here, or would it be preferable to change to the simpler version?

components/sup/src/templating/context.rs
reset: I think it's because `IpAddr`, integers, and bool all implement `Copy` so this satisfied the compiler just fine. @cm I don't think memory really matters here since we're talking 1 to ~16 bytes per each of these types, but you do mean to point to something allocated elsewhere.

components/sup/src/templating/context.rs
christophermaier: This code went through several iterations before I landed on the final implementation... one of those previous iterations required the type hint, but I guess the final one doesn't!

Thanks for catching this; I'll remove it.

components/sup/src/templating/context.rs
christophermaier: I don't _think_ so, but I'll review the Serde docs again to make sure.

components/sup/src/templating/context.rs
christophermaier: Yeah, they all implement `Copy`... I'll make them `Cow`s for consistency (and future safety).

components/sup/src/templating/context.rs
baumanj: Yeah, I looked a bit and it wasn't terribly clear. It said that some serializers needed to know the size _a priori_, but what happened if it was not required and potentially differed from the number actually added seemed ambiguous.

My guess is that it's just a hint and if it's too low we'll just incur an additional alloc, but I think to really know we'd have to use my old master's advice: "use the source, Luke"

christophermaier: Yeah, it sounds like "some serializers" need the hint, but others' don't. Apparently, the JSON serializer is one of the latter, as I can pass `None` as the hint, and all the tests still pass.

If we were just doing a straight serialization of the struct, then we could just use the number of fields in the struct (though really, we'd probably just derive the serialization at that point). However, I envision adding new serialized fields that may be derived from existing fields (e.g., renaming of some fields, while keeping the old one for backwards compatibility, etc.), which means the value won't be so easily computable, apart from a programmer counting.

These types are pretty strongly JSON focused now, seeing as how they're ultimately validated by JSON schema :smile:... we could just drop the size hint for these implementations. I know that we _do_ add the hint in our other `Serialize` implementations, though... probably a good practice to keep them, just for future-proofing's sake; I don't know what formats we may end up wanting to serialize to in the future.

baumanj:
babies-everywhere

`s/babies/cows/`

New changes look great!

PR #4830, opened 2018-03-29
[studio] Ensure that cleanup is performed on `studio new`.

fnichol:
This change addresses a bug where a user runs `hab studio new` and has
mounted filesystems persisting after the command exits. This is unlike
the other commands, namely `hab studio enter`, `hab studio build`, and
`hab studio run`.

The mechanism in the other cases is invoking the `cleanup_studio()`
function in a `trap` statement, but this was missed for the `new`
subcommand. While looking into why, it became clearer that the reason
for this was the trap was being called just before invoking the `chroot`
call to start running the jail. The current call sites also wouldn't
catch any setup issues during the `new_studio()` call, which all of the
above subcommands call. Therefore, it was easier to pull the `trap`
calls up into the subcommand functions consistently which also addresses
the `new` subcommand.

Note that the `rm` subcommand isn't updated here because it calls the
same `cleanup_studio()` function directly.

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -257,6 +257,8 @@ subcommand_new() {
   # Shift off all parsed token in `$*` so that the subcommand is now `$1`.
   shift "$((OPTIND - 1))"
 
+  trap cleanup_studio EXIT
+
   new_studio
 }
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

Issue #4824, opened 2018-03-28
Automatically generate template render context documentation


christophermaier:
#4823 introduced a JSON Schema definition of our rendering context. It would be great if we could replace the hand-maintained documentation at https://www.habitat.sh/docs/reference/#template-data with documentation extracted from the schema itself.

Bonus points for adding examples to the schema.

cnunciato:
I'll bet we could generate the hints for the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=cnunciato.habitat) as part of this, too -- I'll take a look at that as well.

PR #4836, opened 2018-03-29
Expose bind leader directly

### LINKED ISSUE #4127: RFE: In the template syntax, I want to know what topology a given service is in
Currently, you have at the time of template writing whether or not you're binding to a service that is in a topology (e.g. `leader`) or not.   when writing your template, you either choose something like this for finding the leader:

```
{{~ #if bind.database}}
  {{~ #eachAlive bind.database.members as |member|}}
    {{~ #if member.leader}}
PGHOST="{{member.sys.ip}}"
    {{~ /if}}
  {{~ /eachAlive}}
{{~ /if}}
```

but if the service isn't in a topology, `member.leader` will always be false and instead users are instructed to rely on the `@first` or `@last` helpers.   There's no way to detect if a given service is in a topology or not, so you have to know this in advance and that makes reuse difficult.

Ideally, I would want to do something like this in my template:

```
{{~#if bind.database}}
  {{~#if bind.database.topology?}}
     {{~#findLeader bind.database.members as |member|}}
       PGHOST="{{member.sys.ip}}"
     {{~/findLeader}}
  {{else}}
     # loop through via #eachAlive
  {{~/if}}
{{~/if}}
```

or an even more terse way would be something like:
```
  {{~#if bind.database.topology?}}
     PGHOST={{bind.database.leader.sys.ip}}
  {{else}}
  {{~/if}}
```

pie in the sky, but a "just do the right thing" pointer that would choose the leader the in a leader topology or any alive member at random in a non-leader topology would be the nicest from a user perspective

christophermaier:
Previously, determining the leader of a service group was
complicated. If the service was running in a leader topology, the
`{{bind.<BIND>.first}}` key would be the leader, but we didn't
directly indicate if that was the case (you could have queried to see
if `{{bind.<BIND>.first.leader}}` was `true`, but that leads to some
needlessly verbose templates.

Now, we introduce an explicit `leader` key, which will be non-null if
the service group has a leader. Furthermore, the use of `first` is
deprecated, since the meaning of it is not deterministic (it could be
a non-alive member of the service group, for instance, which is not
what anyone wants). The `first` key will remain in rendering contexts
for the foreseeable future, but users are encouraged to use
`{{bind.<BIND>.leader}}` if they want the leader, and
`{{bind.<BIND>.members[0]}}` otherwise, pending future refactorings.

Documentation is updated to reflect this, and a new `$deprecated` and
`$since` key are used in our JSON Schema to indicate when an old field
is deprecated, and when a new one is introduced. Our JSON Schema
parser is configured to allow non-standard keywords like this through
its processing, and this provides some additional hooks that a future
automated documentation generator can use (see #4824)

Fixes #4127

### DIFF
--- a/components/sup/doc/render_context_schema.json
+++ b/components/sup/doc/render_context_schema.json
@@ -498,7 +498,16 @@
                 "properties": {
                     "first": {
                         "description": "The first member of this service group. If the group is running in a leader topology, this will also be the leader.",
+                        "$deprecated": "Since 0.56.0; if you want the leader, use `leader` explicitly. 'first' isn't deterministic, either, so you can just use `members[0]` instead",
                         "$ref": "#/definitions/svc_member"
+                    },
+                    "leader": {
+                        "description": "The current leader of this service group, if running in a leader topology",
+                        "$since": "0.56.0",
+                        "oneOf": [
+                            { "$ref": "#/definitions/svc_member" },
+                            { "type": "null" }
+                        ]
                     },
                     "members": {
                         "description": "All members of the service group, across the entire ring. Includes all liveness states!",
@@ -510,6 +519,7 @@
                 },
                 "required": [
                     "first",
+                    "leader",
                     "members"
                 ],
                 "additionalProperties": false
--- a/components/sup/src/templating/context.rs
+++ b/components/sup/src/templating/context.rs
@@ -748,6 +748,37 @@ two = 2
 
     ////////////////////////////////////////////////////////////////////////
 
+    /// Create a basic SvcMember struct for use in tests
+    fn default_svc_member<'a>() -> SvcMember<'a> {
+        let ident = PackageIdent::new("core", "test_pkg", Some("1.0.0"), Some("20180321150416"));
+        SvcMember {
+            member_id: Cow::Owned("MEMBER_ID".into()),
+            pkg: Cow::Owned(Some(ident)),
+            application: Cow::Owned(None),
+            environment: Cow::Owned(None),
+            service: Cow::Owned("foo".into()),
+            group: Cow::Owned("default".into()),
+            org: Cow::Owned(None),
+            persistent: Cow::Owned(true),
+            leader: Cow::Owned(false),
+            follower: Cow::Owned(false),
+            update_leader: Cow::Owned(false),
+            update_follower: Cow::Owned(false),
+            election_is_running: Cow::Owned(false),
+            election_is_no_quorum: Cow::Owned(false),
+            election_is_finished: Cow::Owned(false),
+            update_election_is_running: Cow::Owned(false),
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/doc/render_context_schema.json
```diff
@@ -478,8 +478,17 @@
                 "properties": {
                     "first": {
                         "description": "The first member of this service group. If the group is running in a leader topology, this will also be the leader.",
+                        "$deprecated": "Since 0.56.0; if you want the leader, use `leader` explicitly. 'first' isn't deterministic, either, so you can just use `members[0]` instead",
```
baumanj: I'm not up on the whole handlebars world, but I remember seeing something about the syntax changing to something like `members.[0]`. Does that apply here?

baumanj:
leader
Obscure?

irvingpop:
hey @christophermaier I'm thinking we should include a generic example in the documentation, like what's in the `postgresql` plan README here:  https://github.com/irvingpop/habitat-core-plans/blob/master/postgresql/README.md#binding

If I'm interpreting this PR correctly, would this be the right syntax?

```
{{~ #if bind.database}}
  {{~ #if bind.database.leader}}
PGHOST="{{bind.database.leader.sys.ip}}"
  {{else}}
PGHOST="{{bind.database.members[0].sys.ip}}"
  {{~ /if}}
{{~ /if}}
```

right?

christophermaier:
@baumanj _Manos: The Templates of Fate_

christophermaier:
@irvingpop Yup, you've got the idea 👍

irvingpop:
After sleeping on this, I'm wondering if we could use this opportunity to improve the UX a bit  -  since this if/else statement will wind up in everyone's plans - would you consider some ideas to make it more terse and expressive?

some ideas in no particular order:
- `bind.BIND.leader` always returns something - if not in a topology, then a random alive member
- same as above, but call it `bind.BIND.auto` - the idea is to try to do the right thing in any scenario

some users may also prefer going the other way and wanting more control:
- `bind.BIND.topology` - returns the topology type
-  `bind.BIND.followers`,  `bind.BIND.anyfollower` 
- `bind.BIND.members.count`,  also `newest`, `oldest`, etc - should be self explanatory and only give you Alive members
- `bind.BIND.members.dead` - enumerate the list of departed members that the ring still knows about

christophermaier:
Just to make things visible... @irvingpop and I will be chatting more about ideas for improving template data next week. It's fine for this PR to stay open until at least then.

christophermaier:
Had a chat with @irvingpop and captured ideas for the future in #https://github.com/habitat-sh/habitat/issues/4872.

Nothing affects the correctness or suitability of this PR, so it's open again to comment and merging.

PR #4841, opened 2018-03-29
Update hab bldr job status

chefsalim:
This change fixes a break in the 'hab bldr job status' command, where it was not possible to query the job groups for an origin because we were always making the group id required, and also getting status for large groups was timing out.

In addition, it adds some other improvements:
* Default job group status for origins to a sane (small) number (10) 
* Add a --limit parameter for when querying all the job groups in an origin
* Add a --showjobs flag for when querying by group id 
* Job groups for origins will not show detailed job status (perf feature. Use the group id and --showjobs flag to get the details)
* Update the display to show tab-aligned tabular format

These changes are backward compatible with the current services, however there is also a corresponding PR for the builder services where we will see the perf improvements.  

tenor-155810919

### DIFF
--- a/components/hab/Cargo.toml
+++ b/components/hab/Cargo.toml
@@ -28,6 +28,7 @@ retry = "*"
 serde = "*"
 serde_json = "*"
 serde_derive = "*"
+tabwriter = "1"
 toml = { version = "*", default-features = false }
 url = "*"
 walkdir = "*"
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -149,16 +149,20 @@ pub fn get() -> App<'static, 'static> {
                     (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
                 )
                 (@subcommand status =>
+                    (about: "Get the status of one or more job groups")
+                    (aliases: &["stat", "statu"])
                     (@group status =>
                         (@attributes +required)
-                        (@arg GROUP_ID: +required +takes_value
+                        (@arg GROUP_ID: +takes_value
                             "The group id that was returned from \"hab bldr job start\" \
                             (ex: 771100000000000000)")
                         (@arg ORIGIN: -o --origin +takes_value
-                            "You can see the status of every group in an origin by providing this value")
+                            "Show the status of recent job groups created in this origin (default: 10 most recent)")
                     )
-                    (about: "Get the status of a job group")
-                    (aliases: &["stat", "statu"])
+                    (@arg LIMIT: -l --limit +takes_value {valid_numeric::<usize>}
+                        "Limit how many job groups to retrieve, ordered by most recent (default: 10)")
+                    (@arg SHOW_JOBS: -s --showjobs
+                        "Show the status of all build jobs for a retrieved job group")
                     (@arg BLDR_URL: -u --url +takes_value {valid_url}
                         "Specify an alternate Builder endpoint. If not specified, the value will \
                          be taken from the HAB_BLDR_URL environment variable if defined. (default: \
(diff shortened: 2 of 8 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -803,3 +807,10 @@ fn valid_url(val: String) -> result::Result<(), String> {
         Err(_) => Err(format!("URL: '{}' is not valid", &val)),
     }
 }
+
+fn valid_numeric<T: FromStr>(val: String) -> result::Result<(), String> {
```
elliott-davis: I think you might be able to omit this function and use the [value_t](https://docs.rs/clap/2.31.2/clap/macro.value_t.html) macro when parsing.

components/hab/src/main.rs
```diff
@@ -523,7 +523,13 @@ fn sub_bldr_job_status(ui: &mut UI, m: &ArgMatches) -> Result<()> {
     let url = bldr_url_from_matches(m);
     let group_id = m.value_of("GROUP_ID");
     let origin = m.value_of("ORIGIN");
-    command::bldr::job::status::start(ui, &url, group_id, origin)
+    let limit = m.value_of("LIMIT")
+        .unwrap_or("10")
+        .parse::<usize>()
+        .unwrap();
```
baumanj: Per @elliott-davis's comment above, it seem like it would be nice to omit the checking in the CLI macro in favor of doing it all here. We could make this
```rust
let limit = m.value_of("LIMIT").unwrap_or("10").parse::<usize>()?;
```
if we add
```rust
impl From<num::ParseIntError> for Error {
    fn from(err: num::ParseIntError) -> Self {
        Error::ParseIntError(err)
    }
}
```
to [errors.rs](https://github.com/habitat-sh/habitat/blob/SA/update-job-status/components/hab/src/error.rs). It already has a `ParseIntError` in the enum, so I'm not sure why we don't have this.

baumanj:
This looks really good!

One thing I noticed in terms of changed behavior:
```
☁ Determining status of Job Group 947929271496884224

Status for Job Group 947929271496884224 (baumanj/minimal-habitat-app): Complete
Created at: 2018-03-28 20:16:05.998406 +00:00

Reverse dependencies:
                                                   Skipped
```
Is now
```
☁ Determining status of job group 947929271496884224
CREATED AT                        GROUP ID            STATUS    IDENT
2018-03-28T20:16:05.998406+00:00  947929271496884224  Complete  baumanj/minimal-habitat-app
```
which looks way nicer, but I'm wondering about why the `Reverse dependencies` info is gone.

Also, I think it might be nice to have a blank line between `☁ Determining` and the rest of the output like there used to be.

baumanj:
tabs

components/hab/Cargo.toml
```diff
@@ -28,6 +28,7 @@ retry = "*"
 serde = "*"
 serde_json = "*"
 serde_derive = "*"
+tabwriter = "1"
```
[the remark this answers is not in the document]
chefsalim: hmm, not clear what is missing - looks ok to me

components/hab/src/cli.rs
chefsalim: Yeah, It's also nice to have the validation up front, similar to the url validation, so it seemed better to follow that pattern

components/hab/src/main.rs
chefsalim: At some point, we decided to move away from using the From pattern for errors, and use map_err instead for more transparency.  Either would be fine, it's nice to have the check up front though so I'l just leave it this way for now.

chefsalim:
The reason for removing the Reverse Dependencies by default is that it is not a lightweight operation.  It's removed completely when getting group status for origins. When getting the group status for a specific group, you can use the --showjobs param to get the reverse dependency/jobs status. I'll add back a blank line to separate the output a bit more.

PR #4843, opened 2018-03-30
prefix pkgPathFor with FS_ROOT

mwrock:
I have never had to use `pkgPathFor` in a windows hook but ran accross a scenario where I need it. Without `FS_ROOT`, non studio supervisors and containerized studios work fine. However, a local studio simply emits a path beginning with `/hab/pkg...` and the local studio is rooted in `c:\hab\studios\...` so the path is not valid. We have managed to catch most scenarios where this is a problem (for example https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/package.rs#L83) but this one slipped.

This should not have any negative side effects for Linux because in Linux studios, the `FS_ROOT` is always `/`.


thesentinels:

Thank you for contributing!

PR #4844, opened 2018-03-30
Revert to boolean-based health in CensusMember

christophermaier:
This fixes a subtlety introduced with the addition of a `Health` field
to the `CensusMember` struct (PR #4823). Previously, there was a small
window where the health of a `CensusMember` was in an indeterminate
state, before it's health was reconciled with Butterfly data. By using
the `Health` enum, all new `CensusMember`s were created as `Alive`.

Arguments could be made for the correctness or incorrectness of this
decision, and we're all agreed that having four booleans to represent
this data is not what we ultimately want. However, since we don't want
to introduce subtle bugs into this part of the code in the course of
solidifying our template data structures, we're backing this change
out for now.

### DIFF
--- a/components/sup/src/census.rs
+++ b/components/sup/src/census.rs
@@ -349,7 +349,7 @@ impl CensusGroup {
     /// you have no alive peers.
     pub fn previous_peer(&self) -> Option<&CensusMember> {
         let alive_members: Vec<&CensusMember> =
-            self.population.values().filter(|cm| cm.alive).collect();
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
@baumanj I don't know that we need a card for cleaning up these 4 booleans, specifically... I think that will happen naturally in the course of tidying up the internal data structures we've got.

PR #4864, opened 2018-04-05
[common] Fix artifact install canonicalizing when dst doesn't exist.

fnichol:
This fixes an issue when installing an artifact directly from disk, for
example:

hab pkg install /path/to/artifact.hart

A previous refactoring dereferenced all relative paths and symlinks when
copying the source (in the above exaple `/path/to/artifact.hart`) into
the artifact cache (i.e. `/hab/cache/artifacts/artifact.hart`). This
refactoring used `PathBuf#canonicalize` from the Rust stdlib, which
returns an error when the file representing the `PathBuf` does not exist
on disk (and also if the file cannot be read due to permissions, etc.).

This change first creates the artifact cache directory, then
canonicalizes the artifact cache *directory* `PathBuf` before appending
on the file name, which allows us to still canonicalize a path that we
know exists.

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -816,8 +816,9 @@ impl<'a> InstallTask<'a> {
         // Canonicalize both paths to ensure that there aren't any symlinks when comparing them
         // later.
         let artifact_path = artifact_path.canonicalize()?;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

Issue #4863, opened 2018-04-05
Health checks NOT reporting to Census


afiune:
While trying to add a `health_check` hook for my micro-service I have detected that Census always reports the health of the service as `Unknown` even though the hook exits with exit code 0, 1 or 2. (look at the docs [here](https://www.habitat.sh/docs/reference/#health_check))

A quick way to reproduce this is by using the elasticsearch core-plan, enter the studio, start the service and check Census. It will always report its health as `Unknown`. (here the ES [hook](https://github.com/habitat-sh/core-plans/blob/master/elasticsearch/hooks/health_check))

Details of my workstation:

OS: Mac so I am running all these inside my studio
```
Linux c733a53a65df 4.9.60-linuxkit-aufs #1 SMP Mon Nov 6 16:00:12 UTC 2017 x86_64 GNU/Linux
```
Hab Version: `hab 0.53.0/20180205213018`

christophermaier:
@afiune As an alternative that works now, you can get a health check status from the following:

```sh
curl <SUPERVISOR_HOST>:9631/services/<SERVICE>/<GROUP>/health
```
If the status is `Ok` or `Warning`, you'll get an HTTP 200, while `Critical` translates to HTTP 503, and `Unknown` will be HTTP 500.

christophermaier:
@afiune I've created https://github.com/habitat-sh/habitat/pull/4873 to add the status of the service in the body returned by that HTTP call, which will allow you to distinguish between "OK" and "WARNING" states when you get an HTTP 200.

Ensuring that the health check status is properly reflected in the census endpoint of the Supervisor's HTTP gateway is a little more involved than I'd initially anticipated... it's work we should still do, but in the meantime, that PR should allow you to do what you need to do.

baumanj:
As of now, I don't see the service health in the census output at all. Is this still an issue?

Here's what I'm seeing on v0.57 today:
```
➤ curl -v http://localhost:9631/census | jq 

* TCP_NODELAY set
* connect to ::1 port 9631 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9631 (#0)
> GET /census HTTP/1.1
> Host: localhost:9631
> User-Agent: curl/7.55.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json
< Content-Length: 2593
< Date: Mon, 02 Jul 2018 21:41:19 GMT
[...]

PR #4873, opened 2018-04-06
Expose health-check status in HTTP gateway response

christophermaier:
We currently expose service health in the HTTP gateway at
`http://<SERVER>:9631/services/<SERVICE>/<GROUP>/health`. The HTTP
status of the response is derived from the health check status,
according to the following mapping:

OK, WARNING => HTTP 200
CRITICAL => HTTP 503
UNKNOWN => HTTP 500

The contents of the `stdout` and `stderr` streams of the health check
script are returned in the body of the response.

This works well, as long as you don't need to distinguish between the
OK and WARNING states.

Since there is no immediately obvious HTTP status that might
correspond with a WARNING state, this commit adds a `status` field to
the JSON response body, the value of which is either "OK", "WARNING",
"CRITICAL", or "UNKNOWN". Users that need to distinguish between OK
and WARNING can inspect the body of HTTP 200 responses to determine
which state the service is currently in.

See #4863 for some background on this.

### DIFF
--- a/components/sup/src/http_gateway.rs
+++ b/components/sup/src/http_gateway.rs
@@ -188,6 +188,7 @@ impl Server {
 
 #[derive(Default, Serialize)]
 struct HealthCheckBody {
+    status: String,
     stdout: String,
     stderr: String,
 }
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #4874, opened 2018-04-06
Generate template reference docs (and update them)

cnunciato:
This change adds a script for generating the template rendering context based on its JSON schema. To run it, install Habitat, then:

```
cd www
make template_reference
```

It also changes the layout of this section to use tables, for scannability, but I also think we could do some work to improve the usability of this page in general; it's pretty huge, and it feels like breaking things out into individual pages might actually be a better solution. 

Signed-off-by: Christian Nunciato <chris@nunciato.org>


thesentinels:

Thank you for contributing!

christophermaier:
tenor-224848591

PR #4875, opened 2018-04-06
[pkg-export-docker] Only require a registry url with amazon/azure types

fnichol:
This change set fixes the above bug but also refactors a CLI-specific type and extracts it to a proper library type which resolves a UX issue in how the CLI interprets the value of `--registry-type`.

## [pkg-export-docker] Extract `RegistryType` to a non-CLI type.

This is a refactoring change which removes the CLI-only enum
`RegistryType` and makes it a proper business logic type in the root of
the `export_docker` module.

The underlying reason for this was twofold:

1) The default `Debug` and `Display` derivations would return the enum
variants as capitalized strings, and not lowercased. This causes the CLI
usage to look incorrect as we're looking for values like `azure` and
not `Azure`. Controlling the `FromStr` and `Display` implementations
yields a much better result.
2) This type was being imported by the rest of the module and used in
the other types and business logic which breaks the CLI/business logic
separation of concerns. If we need a type in non CLI-parsing code, it
shouldn't live in the CLI-parsing code.

## [pkg-export-docker] Only require a registry url with amazon/azure types.
This fixes a bug where if the user provides `--registry-type docker` on
the command line, the CLI parsing demoands that a `--registry-url` be
provided (which it shouldn't).

The only time that `--registry-url` is required is if `--registry-type`
is provided and if the value is either `amazon` or `azure`.

### DIFF
--- a/components/pkg-export-docker/src/cli.rs
+++ b/components/pkg-export-docker/src/cli.rs
@@ -15,25 +15,18 @@
 #![cfg_attr(feature="clippy", feature(plugin))]
 #![cfg_attr(feature="clippy", plugin(clippy))]
 
-use clap::{App, Arg};
 use std::path::Path;
 use std::result;
 use std::str::FromStr;
 
+use clap::{App, Arg};
 use hcore::package::PackageIdent;
 use url::Url;
 
+use RegistryType;
+
 /// The version of this library and program when built.
 pub const VERSION: &'static str = include_str!(concat!(env!("OUT_DIR"), "/VERSION"));
-
-arg_enum!{
-    #[derive(Debug)]
-    pub enum RegistryType {
-        Amazon,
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

elliott-davis:
I'm torn on this. The logic was in place to force users that have private docker registries to still be able to supply a URL. I think they still can with PR but it's less in your face about it.

reset:
@elliott-davis I think that it's still optional that you can provide this though, it's just not required. I think that it's the right behaviour to assume a registry URL if one is not provided which is the canonical public one

elliott-davis:
@reset that sounds right. The behavior previously was, if you specified a URL, you must specify a type. Conversely, if you specify a type, you must specify a URL.

PR #4915, opened 2018-04-13
[common] Restore UI begin/end coloring and debugging capability.

fnichol:
This change fixes a few issues from the refactoring in
habitat-sh/habitat#4805, namely:

* Fixes the lack of coloring with `#begin()`, `#end()`, `#warn()`,
`#fatal()`.
* Fixes the coloring of `#end()` from Yellow back to Blue.
* Adds back UI debugging by implementing `Debug` trait on several inner
types, allowing a default `Debug` impl to be derived on `UI`. This
allows for better debugging in the future when running in various
environments.
* Exposes per-stream coloring and terminal detection so that display
methods can query the specific output stream's capabilities.

Additionally, several issues were also addressed:

* Fixes coloring on multiple lines in `#fatal()` (introduced in
habitat-sh/habitat#1322).
* Fixes a potential coloring issue that arises in the Builder's job
streams where an ANSI reset sequence is not printed before certain
newlines in `#fatal()` (used to print error messages when various CLI
commands fail). This change ensures that each line up to, but not
including the newline is colored.

### DIFF
--- a/components/common/src/ui.rs
+++ b/components/common/src/ui.rs
@@ -115,10 +115,15 @@ pub trait UIWriter {
     fn err(&mut self) -> &mut io::Write;
     /// IO Stream for sending normal or informational messages to.
     fn out(&mut self) -> &mut io::Write;
-    /// Messages sent to the IO streams will be formatted in color if true.
-    fn is_colored(&self) -> bool;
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

fnichol:
tenor-200993992

PR #4917, opened 2018-04-13
Validate configuration before applying to service group

reset:
Apply a basic validation against a running service's configuration
interface. Using `default.toml` (for now) we can message back to the
user that an unknown key was set. This validation can, and will be
expanded upon as we expand upon our configuration interface by
converting `default.toml` from a simple key value store to a markup
which will enable better validations. For example:

```
[my_value]
this = 1
```

Could become:

```
[my_value]
type = "integer"
required = true
default = 1
```

This is now possible because of the addition of the CtlGateway. With
it, we are able to communicate in a request response pattern with a
Supervisor that is aware of what it believes the state of the ring
is and it's own state.

* Send configuration through the Control Gateway and then into the
  gossip ring once validated
* Move `config apply` from hab-butterfly to hab-sup


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -826,6 +816,110 @@ impl Manager {
...
+                ErrCode::BadPayload,
+                format!(
+                    "Unable to decode configuration as {}, {}",
+                    opts.get_format(),
+                    e
+                ),
+            )
+        })?;
+        let service_group: ServiceGroup = opts.take_service_group().into();
+        for service in mgr.services.read().unwrap().iter() {
```
christophermaier: What if the `unwrap` panics?

christophermaier: Don't suppose this could one day be a `HashMap`-style interface?

components/sup/src/manager/mod.rs
```diff
@@ -826,6 +816,110 @@ impl Manager {
...
+            }
+            outputln!(
+                "Setting new configuration version {} for {}",
+                opts.get_version(),
+                service_group,
+            );
+            let mut client = butterfly::client::Client::new(
+                format!("127.0.0.1:{}", mgr.cfg.gossip_listen.port()),
+                mgr.cfg.ring_key.clone(),
+            ).unwrap();
```
christophermaier: Ditto, re: `unwrap`

components/sup/src/main.rs
```diff
@@ -162,6 +179,18 @@ fn cli<'a, 'b>() -> App<'a, 'b> {
             (about: "Start an interactive Bash-like shell")
             (aliases: &["b", "ba", "bas"])
         )
+        (@subcommand apply =>
+            (about: "Applies a configuration to a group of Habitat Supervisors")
+            (@arg SERVICE_GROUP: +required {valid_service_group}
+                "Target service group (ex: redis.default)")
+            (@arg VERSION_NUMBER: +required
+                "A version number (positive integer) for this configuration (ex: 42)")
```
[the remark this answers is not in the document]
reset: This first pass is as close a 1:1 as the previous behaviour. The original reason we chose this was so an external system could keep a version number and pass it here. Since we have no external system for that, we now have made it a habit of just using the date trick.

I'm hesitant to make this auto-generate a version number for you because it could theoretically be destructive. You'll notice that I changed the wording a bit here to apply to be a little less confusing. What we're doing is a PUT or a SET and not an _apply_ which, to me, insinuates we take your previous configuration into account in anyway.

components/sup/src/main.rs
```diff
@@ -437,6 +478,108 @@ fn sub_bash(m: &ArgMatches) -> Result<()> {
...
+                    }
+                    _ => {
+                        Err(SrvClientError::from(
+                            io::Error::from(io::ErrorKind::UnexpectedEof),
+                        ))
+                    }
+                },
+            )
+        })
+        .wait()?;
```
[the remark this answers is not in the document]
reset: I haven't implemented client timeouts yet aside from the initial handshake. Once you've made your handshake the client assumes everything is good. More to come!

components/sup/src/manager/mod.rs
reset: Then the Supervisor is 💯 fucked already. It means the lock is poisoned and we don't want to handle it, we want to crash

reset: Yep! I worked on killing `hab-butterfly` before phase 2 of the refactor because it was safest to work on during the review of phase 1. I'm finishing this work up and then working on killing spec files which will get our services to be stored as the appropriate data structure, a map 😉

components/sup/src/main.rs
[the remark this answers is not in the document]
christophermaier: Auto-generating the number seems like it might be a nice option to provide at some point. I get the compatibility-with-current-behavior argument, though.

Also, 100% behind removing "apply" wording; that has always bugged me, since it does imply that we take into account what's already there!

components/sup/src/ctl_gateway/server.rs
```diff
@@ -285,6 +285,27 @@ impl Future for SrvHandler {
                                             },
                                         )
                                     }
+                                    "SvcSetCfg" => {
+                                        let m = msg.parse::<protocols::ctl::SvcSetCfg>().unwrap();
```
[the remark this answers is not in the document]
reset: I was trusting the client to send the payload they say they are because they've already passed validation. I probably don't want to crash this thread though since I don't have a recovery mechanism. I'll just handle the decoding errors - good catch!

PR #4918, opened 2018-04-13
[hab-sup] Exit 0 when given --help or --version

stevendanna:
Previously, hab-sup --version would exit with 86. Now, we exit 0 for
both --version and --help.

Note, however, we will still exit with an error if the help is
displayed because of an error such as a bad subcommand or bad option.

You may ask, "Do we need to print the error in the case of
HelpDisplayed or VersionDisplayed?" In the case of HelpDisplayed the
answer is yes because the help output isn't actually output to stdout
but rather to a buffer that is contained in the error. In the case of
VersionDisplay the answer is "kinda". The version is printed to
stdout but without a newline, so printing the empty error prints a
newline, which is a bit dirty but kept the function simple.

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -114,7 +114,16 @@ fn start() -> Result<()> {
         Err(err) => {
             let out = io::stdout();
             writeln!(&mut out.lock(), "{}", err.message).expect("Error writing Error to stdout");
-            process::exit(ERR_NO_RETRY_EXCODE);
+            match launcher {
+                Some(_) => process::exit(ERR_NO_RETRY_EXCODE),
+                // If we weren't started by a launcher, exit 0 for
+                // help and version
+                None => match err.kind {
+                    clap::ErrorKind::HelpDisplayed => process::exit(0),
+                    clap::ErrorKind::VersionDisplayed => process::exit(0),
+                    _ => process::exit(ERR_NO_RETRY_EXCODE),
+                },
+            }
         }
     };
     match app_matches.subcommand() {


thesentinels:

Thank you for contributing!

reset:
@stevendanna the reason we exit with 86 is to signal to the launcher to not attempt to restart the Supervisor's process. Commands to the Supervisor binary are proxied through the launcher from the `hab` CLI

stevendanna:
@baumanj Thoughts on special casing `--help` and `--version` in the calling code?

Really, my concern here is I want to by able to query the version of hab-sup without ignoring the exit code.  Apologies for not testing all the way via `hab`, that was on my TODO list but I clearly forgot to do it :D.

baumanj:
> I want to by able to query the version of hab-sup without ignoring the exit code

What is the exit code getting you in this case? You could call something like
```bash
hab sup --version | cut -d' ' -f 2 | cut -d'/' -f 1
```
To extract the version number and check whether you have a reasonable value. I think you'd want that check regardless of the exit code. I'm not clear what behavior you'd want if the version number looked reasonable but the exit code wasn't what was expected.

stevendanna:
To be clear, I've already added the code I need to in my service.  I filed this since when writing that code I had to ignore the bad exit status.  If y'all don't care about --version and --help exiting with an non-zero exit code, it certainly isn't a hill I plan on fighting for.

stevendanna:
@tashimi @reset @baumanj No worries, I wasn't offended, I thought your communication is fine.  I was a bit terse when closing.  I poked at implementing this correctly a bit and found it was hard to test locally on OS X because the launcher early-exits on OSX.  Getting the full dev environment set up this week probably isn't in the cards at the moment, so it didn't seem worth the effort if this was behavior you didn't think you needed.

@reset That implementation seems good to me.  I can try to poke it at this weekend.

stevendanna:
I went to take a look at this today, but the behavior of `hab sup --version`  and `hab sup --help` has changed a bit on master.  `hab sup --version` is no longer a valid command.  `hab sup --help` outputs help but it is quiet different from the previous output and the current output of `hab-sup --help`.  I'm sure these aren't big problems to fix, but was wondering what the intended new behavior is with the rewrite.

baumanj:
Are you running something newer than 0.55.0? This seems to work for me at that release:
```
➤ hab sup --version
hab-sup 0.55.0/20180321222338
```

stevendanna:
@baumanj Yes, I'm using hab and hab-sup built from master.

baumanj:
This is what I get from master:
```
➤ target/debug/hab-sup --version
hab-sup 0.56.0-dev
```
Are you seeing something different?

baumanj:
Gotcha:
```
➤ target/debug/hab sup --version
error: Found argument '--version' which wasn't expected, or isn't valid in this context

USAGE:
    hab sup [FLAGS] [SUBCOMMAND]

For more information try --help
```
I'd say that's a definite bug since the `hab-sup` binary *does* still accept a `--version` flag.

stevendanna:
@baumanj No worries.  This isn't urgent for me.  If you consider `hab sup --version` not working to a bug, I'll track that down the next time I have a few minutes to take a look.

reset:
@stevendanna actually, I think I like this implementation just fine vs the env var one I suggested. I'd love to merge this in but there's a conflict. Would you mind rebasing, pushing, and I'll merge it in today?

baumanj:
@reset: what about the infinite loop potential I mentioned in https://github.com/habitat-sh/habitat/pull/4918#pullrequestreview-112604337 ?

Is that not a practical concern, or were there changes that obviate it that I'm not seeing?

reset:
@baumanj oh nope, that's right. Exactly why this wasn't going to work. I totally was just going through old PRs and looking at the code quickly. I hit this one and saw the changes, thought to myself "why would I suggest something more complicated?" and then left a comment. You're right, we do need the logic in the Launcher. My env var suggestion still stands ;)

stevendanna:
As a side note, it looks like that secondary issue I mentioned with hab sup version on master was solved at some point.

stevendanna:
@reset @baumanj Apologies for the very long delay in such a small PR.  

I've rebased this on master.  It turns out that hab-sup already knows whether it was started via the launcher or not, so I reused the launcher variable here to avoid adding another env var for the same purpose.

Anyway, let me know if that works for you.

baumanj:
I probably won't get to this until Monday, but it looks promising! Thanks for keeping on this.

baumanj:
This code it totally correct. One suggestion I'd make in terms of style/readability:
```rust
            let started_by_launcher = launcher.is_some();
            match err.kind {
                clap::ErrorKind::HelpDisplayed | clap::ErrorKind::VersionDisplayed
                    if !started_by_launcher =>
                {
                    println!("NOT STARTED BY LAUNCHER AND help or version");
                    process::exit(0)
                }
                _ => process::exit(ERR_NO_RETRY_EXCODE),
            }
```

reset:
@stevendanna thanks for following up on this one! Looks good to me

Issue #4066, opened 2017-11-15
MAX_FILE_UPLOAD_SIZE_BYTES seems too conservative


gfodor:
When uploading a file via `hab file upload` there is a strict limit of 4096 bytes. For our use case, we have a generic habitat package that provides a web server and we leave the SSL certificate to be part of the config files. So a user can install the package and then upload their SSL cert via `hab file upload`. However, when including the full trust chain, this file easily exceeds 4096 bytes.

Is there a strong reason this limit has been chosen so conservatively? It doesn't seem right for us to create a habitat package with SSL certificates baked in, and we definitely don't want to fork off and have to deploy these files using some other mechanism entirely. Is it possible to bump this limit?

gfodor:
FYI this is defined here: https://github.com/habitat-sh/habitat/blob/46a724604360e2ad6469c6a4578661b42fb77587/components/hab-butterfly/src/main.rs#L51

eeyun:
So I believe the reason why this is set this way is specifically because file uploads actually use the gossip for dissemination. I think the potential for destructive behavior inside a ring probably grows as the size of the file being gossiped' grows. That's rather vague I know. 

@fnichol @christophermaier @reset or @adamhjk might have some deeper details on testing involved or the actual raw data behind why 4096 was chosen. That being said even if we increase the file size for this upload I would assume we'd still have to keep it pretty small.

christophermaier:
Gossip is the reason the size is limited, but I don't know the details on why 4096 is the magic number. Given that it is gossipped, the size _will_ need to be capped at some point, though.

gfodor:
We're still using a custom fork of `hab` to be able to upload these keys and certs. Any hope on bumping it to say 16k?

reset:
It's 4k because it's meant for gossiping single files like a key, cert, etc. When we chose that byte count it was because we didn't believe those types of files would exceed it. It may be too conservative if you've got a cert chain that blew the cap. I'd be willing to bump it to 32k or so.

gfodor:
yep it's due to our cert chain, with the full cert coming in at about 7k. that'd be awesome! thanks :)

christophermaier:
With Habitat 0.56.0, the maximum size is [64K](https://github.com/habitat-sh/habitat/blob/676d024eb0b9baf1dcc7228e0dc2817560329509/components/sup-protocol/src/butterfly.rs#L16)

PR #4929, opened 2018-04-14
Finish remote supervisor command and control & move commands into hab binary

reset:
This PR accomplishes three major things:
1. Move all commands which go through target a remote supervisor through the CtlGateway from the `hab-sup` binary into the `hab` binary
1. Send and validate files through CtlGateway instead of directly into gossip ring
1. Send departure messages through CtlGateway instead of directly into the gossip ring

#### Details
* Move `hab depart` from `hab-butterfly` to `hab-sup`
* Remove entirely the `hab-butterfly` crate. All command and control
  will now be performed through the CtlGateway and this is no longer
  needed to directly connect command line clients to the Ring
* Add `file upload` to hab-sup binary
* Remove `file upload` from hab-butterfly binary
* Validate an uploaded file at the client and Supervisor layer
  before attempting to gossip file in the Supervisor ring. It's
  important that validation occurs here and not at the protocol
  level so we can _trust_ all data in the Ring as the protocol
  changes. We don't want to retroactively consider rumors "bad"
  but we will want to eventually phase them out.
* Bump allowed file size from 4k to 64k. 4k is a little too
  conservative. We want to stop people from gossiping RPMs and
  other large artifacts, but not their cert chain. This addresses #4066.


thesentinels:

Thank you for contributing!

components/hab/src/main.rs
```diff
@@ -119,6 +143,19 @@ fn start(ui: &mut UI) -> Result<()> {
                 _ => unreachable!(),
             }
         }
+        ("config", Some(m)) => {
+            match m.subcommand() {
+                ("apply", Some(m)) => sub_svc_set(m)?,
```
christophermaier: Should we change the name of this subcommand to `set`, with an alias of `apply`? That way we can start weaning folks over to the command with the better name.

Also, why `sub_svc_set` instead of `sub_config_set`?

components/hab/src/main.rs
```diff
@@ -119,6 +143,19 @@ fn start(ui: &mut UI) -> Result<()> {
...
+                ("show", Some(m)) => sub_svc_config(m)?,
+                _ => unreachable!(),
+            }
+        }
+        ("file", Some(m)) => {
+            match m.subcommand() {
+                ("upload", Some(m)) => sub_file_put(m)?,
```
christophermaier: Why `sub_file_put` instead of `sub_file_upload`?

components/hab/src/main.rs
```diff
@@ -706,6 +754,352 @@ fn sub_pkg_channels(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
+    set.set_service_group(service_group.into());
+    set.set_version(value_t!(m, "VERSION_NUMBER", u64).unwrap());
+    ui.begin(format!(
+        "Setting new configuration version {} for {}",
+        set.get_version(),
+        set.get_service_group(),
+    ))?;
+    ui.status(
+        Status::Creating,
+        format!("service configuration"),
```
christophermaier: Feels like this UI message should be something about validating the configuration, rather than creating it.

components/hab/src/main.rs
```diff
@@ -706,6 +754,352 @@ fn sub_pkg_channels(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
+                        Err(SrvClientError::from(
+                            io::Error::from(io::ErrorKind::UnexpectedEof),
+                        ))
+                    }
+                },
+            )
+        })
+        .wait()?;
+    ui.status(
+        Status::Applying,
```
christophermaier: This status might need to become `Setting`, depending on how we want to roll out the new emphasis on setting a config, rather than applying it.

Aaah... I see later on that this message is more about which supervisor you're connecting to; that was particularly confusing in the context of this particular command. This feels like a more appropriate use of the word "apply" (though we could perhaps use something like "Connecting")... we should probably accelerate the move to "set" when talking about configuration changes.

Incidentally, I notice that this `Status::Applying` isn't used in all commands... would it be worth including that feedback in all commands? If not, what's the criterion for inclusion/exclusion?

components/hab/src/main.rs
```diff
@@ -706,6 +754,352 @@ fn sub_pkg_channels(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
+                }
+                _ => {
+                    Err(SrvClientError::from(
+                        io::Error::from(io::ErrorKind::UnexpectedEof),
+                    ))
+                }
+            })
+        })
+        .wait()?;
+    ui.end("Applied configuration")?;
```
christophermaier: Ditto, re "set" vs. "apply"

components/hab/src/main.rs
```diff
@@ -706,6 +754,352 @@ fn sub_pkg_channels(ui: &mut UI, m: &ArgMatches) -> Result<()> {
...
+    Ok(())
+}
+
+fn sub_svc_config(m: &ArgMatches) -> Result<()> {
```
christophermaier: This just effectively pulls back the package's `default.toml`, as opposed to whatever is currently circulating in the ring, right? Wonder if it's worth changing the subcommand name to reflect that.

(Also plays into https://github.com/habitat-sh/habitat/issues/4827)

I also notice this function is called as `hab config show`, so I'd expect this to be `sub_config_show`, or something similar.

components/hab/src/main.rs
reset: Fixed it

components/hab/src/main.rs
reset: We :100:% need to revisit the UX of the CLI. I didn't want to change the nomenclature in this PR

components/hab/src/main.rs
reset: Precisely. This is a pretty old command that you couldn't even access without calling hab-sup directly. In the long time ago, this command was actually created so you could ask a _package_ and not a _service_ what it's configurables were. I think this needs to move to the pkg command and then this command needs to be expanded upon - we should be showing the entire config, not just defaults, including each section and the merged result.

PR #4953, opened 2018-04-17
Add `hab sup secret` subcommand

reset:
Add a subcommand for interacting with control gateawy secrets. Currently
there is only one subcommand for this subcommand, `generate`. This
command is used for generating a secret key which can then be written to
disk when provisioning a machine with a new Supervisor. This will help
ensure that there is an easy path for sharing a common secret key between all
Supervisors.


thesentinels:

Thank you for contributing!

PR #4962, opened 2018-04-19
Firmly establish role of PATH, RUNTIME_ENVIRONMENT, & RUNTIME_PATH metadata files

fnichol:
This changeset is a bit broad, but all related to the usage and interpretation of 3 metadata files in a pacakge:

* `PATH`
* `RUNTIME_ENVIRONMENT`
* `RUNTIME_PATH`

In the case of PATH, we introduced this really nice-to-have pre-computed runtime `$PATH` value which the Supervisor and other tooling can use without have to re-compute this statically-knowable answer at runtime. This leaves us with 2 notions of PATH: one is used by the build system and some tooling to answer the question: "where are the programs for this package?" and the other is used by runtime tools to answer the question: "what should I prepend to my $PATH before running a program
out of this package?". Notice that these 2 notions are independent of any other runtime environment variables that a Plan author could introduce such as `JAVA_HOME`, `PYTHONPATH`, `CA_CERT_FILE`, etc.

As a result, we split this notion of `PATH` into 3 parts:

1. Re-assert the role of the `PATH` metadata file which represents "where are the programs for this package?".
2. Build and export a value for the `$PATH` (`$env:PATH` on Windows) environment variable that is deprived from using the `$pkg_bin_dirs` Plan variable and the `PATH` metadata files of package dependencies.
3. Build a static "runtime path" value that gets written into a net-new `RUNTIME_PATH` metadata file and also gets written into a `PATH` entry in the `RUNTIME_ENVIRONMENT` metadata file only for backwards compatibility.

The following inlined commit messages add more context and detail:

## [plan-build] Treat PATH distinctly from RUNTIME_ENVIRONMENT.

This change further re-asserts how the PATH environment variable is used
and interpreted in difference places. In the context of the build
program, there are 3 primary places where we care about something called
`$PATH`:

1. The path elements in a package which contain programs, declared by a
Plan author in `$pkg_bin_dirs[@]` and written out to a `PATH` metadata
file.
2. The "runtime" value for `$PATH` which the Supervisor (and others) use
when starting a service. This is a combination of a package's own path
elements, and those of its direct and indirect runtime dependencies
(ordered in a particular way). It is written out to a `PATH` key in a
`RUNTIME_ENVIRONMENT` metadata file.
3. The value of `$PATH` which is exported for the build program to use
when building the package. This is only used in the running context of
the build program and contains build dependencies as well as runtime
dependencies.

This change deals with cases 2 and 3 and leaves 1 as-is.

2. Runtime Environment PATH
---------------------------

Prior to this change, the value of the `PATH` key in the
`RUNTIME_ENVIRONMENT` metadata file was calculated by pulling in the
`RUNTIME_ENVIRONMENT` from dependencies and layering them on themselves.
The issue was it was to use this information to derive a clearly ordered
path array that is in the following order:

1. All directories in this package which have program binaries (set by
the Plan author in `${pkg_bin_dirs[@]}`), in declared order
2. All path elements for each *direct* runtime dependency which has a
`PATH` metadata file, in the order present in `PATH`
3. All remaining path elements for each *indirect* (i.e. transitive)
dependency which has a `PATH` metadata file, in the order present in
`PATH`

Ideally, this collection should be a unique ordered set with no
duplicated path elements.

It ended up being easier to re-compute this runtime PATH ordering by
querying the direct and indirect dependencies directly and derive the
result without using the `RUNTIME_ENVIRONMENT` metadata files. This
seemed to suggest that the notion of "PATH" is more like the `CFLAGS`
and `LDFLAGS` metadata and less like the runtime environment
capabilities in that there are explicit Plan variables
(`$pkg_bin_dirs[@]`, `$pkg_lib_dirs[@]`, etc.) that map directly to
corresponding metadata files (`PATH`, `LDFLAGS`, etc.) which are used by
the build system to assemble environment variables to be used when
building a package.

This change changes how that `PATH` entry is computed--separately from
the values `RUNTIME_ENVIRONMENT` metadata files. Instead this value is
computed on its own (via `_set_runtime_env_path()`) after all other
environment variable loading is complete and set in the runtime
environment hash which will ultimately be written to the package's
`RUNTIME_ENVIRONMENT` metadata file.

As a result, most of the PATH-related code associated with reading these
environment metadata files was removed, and any `PATH` key entries are
skipped when loading in other package dependencies.

3. Build time value of `$PATH`
------------------------------

In a similar line of reasoning the case 2 above, the build time value of
`$PATH` which gets exported by the build program is calculated
independently of any `RUNTIME_ENVIRONMENT` metadata file values and is
created in the following specific order:

1. All directories in this package which have program binaries (set by
the Plan author in `${pkg_bin_dirs[@]}`), in declared order
2. All path elements for each *direct run* dependency which has a `PATH`
metadata file, in the order present in `PATH`
3. All path elements for each *direct build* dependency which has a
`PATH` metadata file, in the order present in `PATH`
4. All remaining path elements for each *indirect run* dependency which
has a `PATH` metadata file, in the order present in `PATH`
5. All remaining path elements for each *indirect build* dependency
which has a `PATH` metadata file, in the order present in `PATH`
6. The initial/original value of `$PATH` when the program started to
ensure that build program-related programs can still be found

As above, this collection should be a unique ordered set with no
duplicated path elements. The only exception is that we preserve the
initial/original value of `$PATH` to append it as-is (i.e. we don't
de-duplicate the original `$PATH`).

Finally, this task is broken out into a dedicated private phase in the
build (`_set_build_path()`) which happens directly after
`do_setup_environment_wrapper()` and before the `do_before()` build
phase to ensure that `do_before()` implementations have a correctly set
`$PATH`.

## [plan-build] Only create `PATH` metadata file if non-empty.

This change fixes an issue where all built packages contained a `PATH`
metadata file. If no program directories were present, then this file
would be empty. While not technically a bug (all code consuming these
metadata files should know how to deal with a non-existant or empty
file), it is much more intention-revealing to skip creating files if
they are empty.

The refactoring in the previous commit meant that the internal
`__process_path()` function was only used by `_render_metadata_PATH()`,
and so its implementation was inlined, removing this extra function.

## [hab,sup] Update to call `PackageInstall#environment_for_command()`.

References habitat-sh/core#25

## [plan-build] Generate `RUNTIME_PATH` metadata files where appropriate.

This change makes the build program produce `RUNTIME_PATH` metadata
files if a non-empty runtime path is computed.

As this change builds on previous work, part of this change involves
moving some functions and renaming them to further clarify their
responsibilities.

## [plan-build-ps1] Treat PATH distinctly from RUNTIME_ENVIRONMENT.

This change updates the PowerShell implementation of the build system to
re-assert how the PATH environment variable is used and interpreted in
different places. Additionally, this change brings back the `LIB_DIRS`
and `INCLUDE_DIRS` metadata files which a package will contain if it has
`$pkg_lib_dirs`/`$pkg_include_dirs` values. Lastly, there is fallback
logic to look in the `RUNTIME_ENVIRONMENT` for `LIB`/`INCLUDE` entries
if the corresponding `LIB_DIRS`/`INCLUDE_DIRS` metadata files aren't
present (this is to handle older built packages which are used as
dependencies in a build).

See the previous change to the Bash implementation for more details and
rationale. The changes here are designed to bring both implementations
in line with each other and continue to produce similar results.

tenor-4846503

### DIFF
--- a/components/plan-build/bin/environment.bash
+++ b/components/plan-build/bin/environment.bash
@@ -166,6 +166,14 @@ __populate_environment_from_deps() {
         if [ -f "${path_to_dep}/RUNTIME_ENVIRONMENT" ]; then
             while read -r line; do
                 IFS== read var val <<< "${line}"
+                # Any values of `PATH` are skipped as we will be computing the
+                # runtime path independently of the RUNTIME_ENVIRONMENT
+                # metadata files. Additionally, this acts as backwards
+                # compatibility for all `RUNTIME_ENVIRONMENT` files that
+                # contain a `PATH` key.
+                if [[ "${var}" == "PATH" ]]; then
+                  continue;
+                fi
 
                 if [ -n env["${var}"] ]; then
                     # There was a previous value; need to figure out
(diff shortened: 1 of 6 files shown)


thesentinels:

Thank you for contributing!

components/plan-build/bin/shared.bash
```diff
@@ -143,26 +143,26 @@ _render_metadata_LD_RUN_PATH() {
...
+    local dir
 
+    debug "Rendering PATH metadata file"
     # Contents of `pkg_bin_dirs` are relative to the plan root;
     # prepend the full path to this release so everything resolves
     # properly once the package is installed.
-    for bin in "${pkg_bin_dirs[@]}"; do
-        path+=("${pkg_prefix}/${bin}")
+    for dir in "${pkg_bin_dirs[@]}"; do
+        paths+=("${pkg_prefix}/${dir}")
```
christophermaier: These renamed variables... 💯 👍

components/plan-build/bin/environment.bash
```diff
@@ -204,80 +212,61 @@ __populate_environment_from_deps() {
...
+  # onto the result, assuming it hasn't already been added. In this way, all
+  # direct dependencies will match first and any programs that are used by a
+  # direct dependency will also be present on PATH, albeit at the very end of
+  # the PATH.
+  for dep in "${pkg_deps_resolved[@]}" "${pkg_tdeps_resolved[@]}"; do
+    if [[ -f "$dep/PATH" ]]; then
+      data="$(cat "$dep/PATH")"
+      data="$(trim "$data")"
+      while read -r entry; do
+        paths=($(_return_or_append_to_set "$entry" "${paths[@]}"))
```
christophermaier: We might want to filter out any elements in the `PATH` file that aren't rooted in the package it's coming from, to account for packages built when `PATH` contained the entire runtime path of the package.

components/plan-build/bin/environment.bash
```diff
@@ -204,80 +212,61 @@ __populate_environment_from_deps() {
...
+  # Return the elements of the result, joined with a colon
+  join_by ':' "${paths[@]}"
+}
 
-    # Finally, return the path
-    echo "$path"
+# **Internal**  Set the `PATH` key for the runtime environment if a computed
+# runtime path is necessary. If a package has no `${pkg_bin_dirs[@]}` elements
+# and has no runtime dependencies with `PATH` entries, then a computed runtime
+# path would be empty, meaning that no `PATH` value should be set.
```
christophermaier: Might be worth noting that we are only doing this to ensure that any services built with this version of the build program can still run in Supervisors between when `RUNTIME_ENVIRONMENT` was introduced and now, which de facto assume the presence of a `PATH` variable (and any non-services that you might run via `hab pkg exec`, I suppose). Supervisors after this time will use a separate `RUNTIME_PATH` file which will get layered on top (spoiler alert for those reading along at home 😜 )

components/plan-build/bin/environment.bash
```diff
@@ -463,12 +452,7 @@ do_setup_environment_wrapper() {
     # runtime path onto the front of the system path
     for k in ${!__runtime_environment[@]}; do
         local v="${__runtime_environment[${k}]}"
-        if [ "${k}" == "PATH" ]; then
-            build_line "Value of ${k} is ${!k}"
-            export "${k}"="$(push_to_path ${v} ${!k})"
-        else
-            export "${k}"="${v}"
-        fi
+        export "${k}"="${v}"
```
christophermaier: Is`__set_runtime_env_path` expected to run before or after `set_buildtime_env`? They operate on the same data structures, so it'd be nice to document the assumptions for the use of each.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1353,6 +1378,41 @@ _resolve_dependencies() {
...
+  for dir in "${pkg_bin_dirs[@]}"; do
+    paths+=("$pkg_prefix/$dir")
+  done
+
+  for dep in "${pkg_all_tdeps_resolved[@]}"; do
+    if [[ -f "$dep/PATH" ]]; then
+      data="$(cat "$dep/PATH")"
+      data="$(trim "$data")"
+      while read -r entry; do
+        paths=($(_return_or_append_to_set "$entry" "${paths[@]}"))
```
christophermaier: Same comment as above, re: filtering `PATH` metadata file entries from packages other than the current one.

components/plan-build/bin/environment.bash
fnichol: I was planning on changing this comment with the next commit to make this change.

Oh, and I forgot to call this out specially, but I dropped the provenance of `PATH` for `RUNTIME_ENVIRONMENT_PROVENANCE` as a first step to making this a backwards compatible change.

components/plan-build/bin/environment.bash
fnichol: That's a great, excellent point. Turns out this was easy and a regression of my own making here, so I'll rebase this into the commit and augment the comment above this code block to explain.

components/plan-build/bin/hab-plan-build.sh
fnichol: Same as above, will add this in!

components/plan-build/bin/environment.bash
fnichol: I also renamed `$dep` to `$dep_prefix` in order to better explain that this variable contains a prefix path and not a package identifier, which in turn makes the above suggestion clearer to implement.

components/plan-build/bin/environment.bash
fnichol: Addressed now.

components/plan-build/bin/environment.bash
```diff
@@ -488,7 +447,7 @@ do_setup_environment_wrapper() {
         fi
     done
 
-    __set_runtime_env_path
+    __back_compat_set_runtime_env_path
```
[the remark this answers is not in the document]
fnichol: I'll address this one too, great idea!

fnichol:
I'll get this rebased against master to clear the conflict and hopefully this will fix the DCO check not completing

PR #4967, opened 2018-04-19
Reuse recv buffer in SrvCodec

reset:
Previously we would allocate two new buffers for every message when
decoding a message sequence in SrvCodec. This change will allocate a
single buffer the size of our maximum allowed payload and re-use it
between each call to decode.


thesentinels:

Thank you for contributing!

PR #4969, opened 2018-04-20
Fix handling of environment-level configuration

### LINKED ISSUE #4439: Unable to parse TOML/JSON from ENV
I am attempting to use HAB_PACKAGENAME environment variables to override default.toml config options in my docker-packaged application, and no matter what content I put in the HAB_PACKAGENAME envvar, the hab sup reports that it cannot parse the content:
```
Unable to start finalze/active-oversight-api, hab-sup(CF)[src/manager/service/config.rs:248:20]: Unable to find valid TOML or JSON in HAB_ACTIVE_OVERSIGHT_API ENVVAR
```

I have tried working with habitat versions 0.50.3 and 0.51 with the same results. The same behavior has been observed on ECS running the latest docker-optimized Amazon Linux AMI as well as locally on a machine running Docker CE 17.12.0-ce on Ubuntu Linux 17.10

docker-compose snippet:
```
    ao-worker:
        image: "finalze/active-oversight-api"
        environment:
                - HAB_ACTIVE_OVERSIGHT_API="{'worker':'true'}"
        hostname: "ao-worker"
        links:
            - postgres
            - elasticsearch
            - redis
```

env content within the container:
```
HAB_ACTIVE_OVERSIGHT_API="{'worker':'true'}"
```

other env contents that have produced the same error:
```
HAB_ACTIVE_OVERSIGHT_API='worker="true"'
HAB_ACTIVE_OVERSIGHT_API="worker='true'"
HAB_ACTIVE_OVERSIGHT_API="worker=true"
HAB_ACTIVE_OVERSIGHT_API=""
HAB_ACTIVE_OVERSIGHT_API="{}"
HAB_ACTIVE_OVERSIGHT_API={}
```

christophermaier:
When we were trying to parse TOML from an environment variable, we
weren't properly type-hinting the result of that operation, which
meant Rust derived the type from the return value of our function. As
a result, we ended up with an error, because Serde was trying to
decode directly to an `Option<toml::value::Table>`. We now type things
correctly.

(By the way, the code was all completely valid Rust before; it's just
that Serde is such a flexible framework that can parse into arbitrary
types!)

Additionally, parsing JSON as a fall-back wasn't working because we
were expecting `serde_json` to parse a string into a TOML table
directly, which it can't do. Now, we transcode the JSON string into a
TOML string, and from there use the `toml` crate to parse into a
`toml::value::Table`.

Fixes #4439

### DIFF
--- a/components/sup/src/manager/service/config.rs
+++ b/components/sup/src/manager/service/config.rs
@@ -28,6 +28,7 @@ use hcore::fs::USER_CONFIG_FILE;
 use serde::{Serialize, Serializer};
 use serde::ser::SerializeMap;
 use serde_json;
+use serde_transcode;
 use toml;
 
 use super::Pkg;
@@ -132,7 +133,7 @@ impl Cfg {
         };
         let user_config_path = Self::determine_user_config_path(package);
         let user = Self::load_user(user_config_path.get_path())?;
-        let environment = Self::load_environment(package)?;
+        let environment = Self::load_environment(&package.name())?;
         return Ok(Self {
             default: default,
             user: user,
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/config.rs
```diff
@@ -938,4 +976,84 @@ mod test {
...
+
+        // Clean up the environment after ourselves
+        env::remove_var(key);
+
+        assert!(result.is_err(), "Expected an error: got {:?}", result);
+    }
+
+    #[test]
+    fn no_environment_config_is_fine() {
+        match Cfg::load_environment(&"omg-there-wont-be-an-environment-variable-for-this".to_string())
```
krnowak: `export HAB_OMG_THERE_WONT_BE_AN_ENVIRONMENT_VARIABLE_FOR_THIS='in_your_face="haha"'`

Probably putting `=` in the string would defeat me. ;)

PR #4973, opened 2018-04-20
This may not be for you; it is for applications

nathenharvey:
This is almost definitely an irrational pet peeve of mine that is not
rooted in any user research but I really dislike speaking to "you" in
our copy.  This reader might be a manager who does not actually run any
applications, for example.   Or an operations engineer who does not
build applications but does need to run them.

Additionally, I think this reads a bit better.  I do recognize that this
is a pet peeve of mine that I should maybe just let go.

Opening this PR as a way to discuss...

Signed-off-by: Nathen Harvey <nharvey@chef.io>

### DIFF
--- a/www/source/index.html.slim
+++ b/www/source/index.html.slim
@@ -8,7 +8,7 @@ section.home--hero
   .home--hero--content.row
     .columns.small-12.medium-6
       h1.home--hero--heading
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

Issue #4909, opened 2018-04-12
Give all topologies a "leader"


christophermaier:
Right now, the only topology that has a formal "leader" is the leader/follower topology.

However, it can be useful for *all* topologies to have a "leader" (or "distinguished member", or whatever you'd like to call it). This member could ensure that all members of the group are running the same version of the service in question, and could be the ones to download the new package from the Depot and distribute it to other group members. This would drastically cut down how many Supervisors need to open persistent connections with the Depot (or a future notifications service).

baumanj:
Valid as of v0.57

adamhjk:
This is already what happens if you specify an update strategy. It’s called the UpdateLeader. 

We could make sure it happens for all service groups, tho.

PR #5024, opened 2018-05-03
Allow services to start in the absence of their binds

christophermaier:
Previously, a service would not start until all service groups it was
bound to were present and alive in the census. In general, though,
well-behaved distributed services should be able to gracefully deal
with the absence of any of their dependencies. This behavior makes it
difficult to do that.

Here, we introduce a new `--binding-mode` option, which affects how
the absence of a bind influences service start up. The current
behavior, where start-up is blocked until all binds are present, is
used with `--binding-mode=strict`. For now, this will be the default
mode, to retain consistent behavior with currently-running services,
but eventually this will _not_ be the default, in favor of
`--binding-mode=relaxed`, described below.

A service starting in the relaxed binding mode can start even if none
of its binds are present. Plan authors should account for this in
their templated configuration and hook scripts by looking for the
presence of a particular bind in the templating data. As an example,
if your service can operate in the absence of its `database` bind, you
could have the following in a template:

```
{{ #if bind.database }}
  // database-specific configuration
{{ else }}
  // fallback configuration when no database is present
{{ /if }}
```

The `bind.database` structure will be empty until there are
active (live or suspect) census members present in the bound service
group.

In a change from previous behavior, should the number of active
members in the service group drop to 0, the bind will _not_ be present
in the templating data. In this way, your application can reconfigure
itself when a group goes away.

This change also introduces bind validation when determining if a bind
has been satisfied or not. Previously, mere presence was enough for a
bind to be considered satisfied. Now, the exports of the bound service
group are compared to what the bind expects. Only when all exports are
present is the bind considered satisfied. This is also reevaluated as
the census is updated. In order to achieve this, we need some idea of
what "the group" is exporting. In the absence of structures to
formally expose that, we simply ask either the leader or first active
member of the group what they export, and use that as a proxy for what
the group as a whole exports. #4909 has been suggested as a mechanism
to do this properly.

It should be noted that, just as in all other releases of Habitat to
date, no special action is taken when the bind of a service starting
in "strict" mode disappears. That is, the Supervisor does not halt the
service or do anything else, but the information would be reflected in
the census, and thus available when templating files. The removal of
the `bind.<MY_BIND>` map when the `<MY_BIND>` bind goes away should
make it easier to write templates that can react to this, however.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -727,7 +727,7 @@ dependencies = [
 [[package]]
 name = "habitat_core"
 version = "0.0.0"
-source = "git+https://github.com/habitat-sh/core.git#40451521e357548e68a2b3d845da850dcecd2a74"
+source = "git+https://github.com/habitat-sh/core.git#0e296163684e6f2ea7d9e692355107dab5b51188"
 dependencies = [
  "ansi_term 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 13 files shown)


thesentinels:

Thank you for contributing!

PR #5033, opened 2018-05-04
Add JSON structured logging to the Supervisor

christophermaier:
TL;DR - add `--json-logging` when you start your Supervisor and each
log line will be a JSON object.

While the core of this implementation is in the `habitat_core` crate,
governing how a `StructuredOutput` struct is emitted, there are some
interesting follow-on changes that were needed here, along with
several housekeeping tasks.

First, we had `-v` and `--no-color` as global options, which are no
longer necessary. Previously, there were several ways that a
Supervisor could start up, but now there is only one: `hab sup
run`. We also simply had these on commands where they simply were not
needed. These options ONLY govern how the Supervisor's output is
formatted, and not any client-side output.

Secondly, the Launcher is now "eavesdropping" on all the log
formatting flags. Previously, it was only paying attention to
`--no-color`, but not `-v`. This means that the `-v` flag for verbose
logging was effectively only half-implemented, since output from the
Launcher itself was never configured with it.

Most interestingly, the `CtlRequest` struct performs an interesting
double-duty; it's output is sent to both the client that issued the
request, as well as to the Supervisor's output streams. The
`CtlRequest` is currently configured to _always_ use color (as well as
to behave as though it is always connected to a terminal), but this
will result in ANSI color codes being output to the Supervisor logs,
regardlss of how the Supervisor itself is configured. The
implementation here continues to do double-duty, but strips the ANSI
codes out of the string it sends to the Supervisor, when
necessary. It's not yet clear that additional refactoring is
necessary, given that this is the only place in the code where we have
this situation. This at least makes the two output destinations
explicit, whereas before it was not very obvious.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -727,7 +727,7 @@ dependencies = [
 [[package]]
 name = "habitat_core"
 version = "0.0.0"
-source = "git+https://github.com/habitat-sh/core.git#0e296163684e6f2ea7d9e692355107dab5b51188"
+source = "git+https://github.com/habitat-sh/core.git#6a7cc6b5be2e13c55737eca4765ba35dc580d8bd"
 dependencies = [
  "ansi_term 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


thesentinels:

Thank you for contributing!

PR #5059, opened 2018-05-10
[ci] Pin rustfmt by pinning to Rust 1.25.0.

fnichol:
When we upgrade to Rust 1.26.0 (most likely after the next Habitat
release) we can drop these pins back to `stable` which will revert us
back to using Rust 1.26.0 and *also* `rustfmt` 0.4.1.


thesentinels:

Thank you for contributing!

PR #5064, opened 2018-05-11
Fix some command aliases

christophermaier:
Some aliases got broken. Now they're fixed. 💥


thesentinels:

Thank you for contributing!

fnichol:
Note that while I labelled it a bug (we're limited with out category buckets for good reason), it's a UX-preserving change and not a failure of logic or anything like that.

PR #5070, opened 2018-05-11
[plan-build.ps1] Use `RUNTIME_ENVIRONMENT` if necessary for run path.

fnichol:
This change adds one more backwards compatible case when building
Windows Habitat packages. Older packages have their `PATH` entries
exclusively in `RUNTIME_ENVIRONMENT` metadata files and so we need to
fallback and check these entries if no `PATH` metadata file is present
in dependent packages when computing the runtime path value.


thesentinels:

Thank you for contributing!

PR #5087, opened 2018-05-16
Run the best local candidate of a service first

christophermaier:
This is the previous behavior, inadvertently undone in the recent
Supervisor refactoring.

When a service is loaded, locally-installed packages are consulted
first, and the most suitable one is run if it exists. Otherwise, we'll
install the latest from Builder and run that.

Without this, you can get unexpectedly updating services, or attempts
to reach out to Builder when you may neither be able to, nor want
to. Docker containers are particularly important here, as without this
change, any exported container would always update itself to whatever
is latest in Builder whenever it starts.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -960,8 +960,13 @@ impl Manager {
             None => {
                 // We don't have any record of this thing; let's set it up!
                 //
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #5090, opened 2018-05-17
Disable validation of config/files

reset:
Disable (for now) validation of configuration and files when applying
configuration or uploading to a Supervisor. We aren't quite there just
yet - we need to be storing all loaded services in memory at all times
(not just when they are running).


thesentinels:

Thank you for contributing!

PR #5091, opened 2018-05-17
Disable interface validation when applying config

reset:
Temporarily disable configuration validation against a service's
configuration interface until we are propagating that interface via
butterfly and not relying on a service loaded in memory.


thesentinels:

Thank you for contributing!

PR #5096, opened 2018-05-17
Fix panic when an initial service fails to loaded

reset:
Allow errors when loading an initial service via `hab sup run` fails to
propagate to the CLI instead of panicking.

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -43,9 +43,9 @@ use clap::{App, ArgMatches};
 use common::command::package::install::InstallSource;
 use common::ui::{Coloring, NONINTERACTIVE_ENVVAR, UI};
 use hcore::channel;
-use hcore::crypto::{self, default_cache_key_path, SymKey};
 #[cfg(windows)]
 use hcore::crypto::dpapi::encrypt;
+use hcore::crypto::{self, default_cache_key_path, SymKey};
 use hcore::env as henv;
 use hcore::service::{ApplicationEnvironment, BindingMode, ServiceGroup};
 use hcore::url::{bldr_url_from_env, default_bldr_url};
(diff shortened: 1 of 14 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -664,7 +664,7 @@ impl Manager {
         });
         core.handle().spawn(ctl_handler);
         if let Some(svc_load) = svc {
-            Self::service_load(&self.state, &mut CtlRequest::default(), svc_load).unwrap();
+            Self::service_load(&self.state, &mut CtlRequest::default(), svc_load)?;
```
reset: This is the line you care about

reset:
Note to any reader: the new Rust (1.26.0) includes a new Rustfmt which automatically arranges use statements. That's why there's more churn here than a few lines.

chefsalim:
Fix looks good but should we hold off on the rustfmt churn until after the release when we're all on 1.26?

chefsalim:
Ah ok - cool

components/sup/src/manager/mod.rs
```diff
@@ -664,7 +664,16 @@ impl Manager {
         });
         core.handle().spawn(ctl_handler);
         if let Some(svc_load) = svc {
-            Self::service_load(&self.state, &mut CtlRequest::default(), svc_load).unwrap();
+            match Self::service_load(&self.state, &mut CtlRequest::default(), svc_load) {
+                Err(err) => {
+                    if err.get_code() == ErrCode::Conflict {
```
christophermaier: Doesn't this swallow the error when you try to load something that is already loaded, without passing `--force`?

reset: Ah, I can see a slight edge case here now that you point it out.

This only runs if you pass a package identifier to `hab sup run`. However, if you had previously ever run the Supervisor and loaded a service with the same name, we'd just use what you had specified for that service to run and ignore any additional flags that you passed to `hab sup run`.

I don't think people want to need to pass `--force` here and `hab sup run` is more of a do what I mean thing. If that's the case they'd probably be wondering why the other flags they passed didn't get picked up. I'll think a bit more about this

components/sup/src/main.rs
```diff
@@ -226,6 +224,10 @@ fn sub_run(m: &ArgMatches, launcher: LauncherCli) -> Result<()> {
         let svc = if let Some(pkg) = m.value_of("PKG_IDENT_OR_ARTIFACT") {
             let mut msg = protocol::ctl::SvcLoad::new();
             update_svc_load_from_input(m, &mut msg)?;
+            // Always force - running with a package ident is a "do what I mean" operation. You
+            // don't care if a service was loaded previously or not and with what options. You
+            // want one loaded right now and in this way.
+            msg.set_force(true);
```
reset: This is the magic

PR #5118, opened 2018-05-25
Allow `--remote-sup` to take a DNS name as an argument

christophermaier:
When wiring up Supervisors, the `--peer` option will accept DNS names
as well as bare IP addresses. Additionally, it will add the default
gossip port if a port is not supplied. The new `--remote-sup` option,
however, would _not_ do this; instead, it required a bare IP address
_and_ port, even if it was the default control port.

This commit gives the same flexibility of the `--peer` option to the
`--remote-sup` option.

Now, each of the following are possible:

    hab svc load core/redis --remote-sup=alpha.habitat.dev
    hab svc load core/redis --remote-sup=alpha.habitat.dev:9632
    hab svc load core/redis --remote-sup=172.18.0.2
    hab svc load core/redis --remote-sup=172.18.0.2:9632

Leaving the `--remote-sup` option off results in connecting to a
Supervisor running on the local host, as before.

If you supply something that doesn't resolve, you get an error message like this:

    hab svc load core/redis --remote-sup=300.300.300.300 # not a real IP address
    ✗✗✗
    ✗✗✗ Failed to resolve remote supervisor '300.300.300.300:9632': failed to lookup address information: Name or service not known
    ✗✗✗

Supplying a DNS name that doesn't resolve gives the same error, but it
takes a bit longer (a few seconds) to return, since it has to query
DNS.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -70,7 +70,7 @@ pub fn get() -> App<'static, 'static> {
                 (aliases: &["sh", "sho"])
                 (@arg PKG_IDENT: +required +takes_value
                     "A package identifier (ex: core/redis, core/busybox-static/1.42.2)")
-                (@arg REMOTE_SUP: --("remote-sup") -r +takes_value {valid_socket_addr}
+                (@arg REMOTE_SUP: --("remote-sup") -r +takes_value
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

PR #5129, opened 2018-05-29
Use octal for permissions of CTL_SECRET!

christophermaier:
Signed-off-by: Christopher Maier <cmaier@chef.io>

PR #5131, opened 2018-05-29
WIP: Make lint fix

baumanj:
It looks like getting clippy running as a cargo subcommand isn't too hard and allows the removal of a lot of superfluous code needed to make clippy an optional dependency. This uncovered at least [one serious correctness issue](https://github.com/habitat-sh/habitat/pull/5129), so perhaps it's worth examining running some lints in travis.


thesentinels:

Thank you for contributing!

reset:
@baumanj great catch here and it's nice to see we can remove some of the bits of clippy inside our modules. I think there's value in the code analysis tooling here for sure and agree we should hook this up Travis if we want to make it real.

It looks like there's a slight formatting issue which is what made the build fail. Could you reformat with the rustfmt included in Rust 1.26.0 or Rust 1.26.1? And if you could hit the Cargo.lock rebase that'd get this in shape!

PR #5136, opened 2018-05-30
Revert "Merge pull request #5126 from habitat-sh/prost-sup"

christophermaier:
This reverts commit 2c0d9f0bc570a7ec497593e6ed3ef282b7833ae0, reversing
changes made to 251fe99ff3a23b8f4c51220ffb681f2cd99360ce.

PR #5138, opened 2018-05-31
Switch from the `protobuf` crate to the `prost` crate

reset:
This reverts the revert from #5136 and fixes an issue where a default value of group would be a blank string and not `"default"`. In addition, the test suite for `sup-protocol` and `sup-client` will now run on the build server.


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/spec.rs
```diff
@@ -123,47 +122,36 @@ pub trait IntoServiceSpec {
 
 impl IntoServiceSpec for protocol::ctl::SvcLoad {
     fn into_spec(&self, spec: &mut ServiceSpec) {
-        spec.ident = self.get_ident().clone().into();
-        if self.has_group() {
-            spec.group = self.get_group().to_string();
+        spec.ident = self.ident.clone().unwrap().into();
+        spec.group = self.group.clone().unwrap_or(DEFAULT_GROUP.to_string());
```
reset: @christophermaier this is the line that you care about 😢

christophermaier: tenor-115949343

PR #5147, opened 2018-06-03
Improve error message when acting on a service that isn't loaded

reset:
This will clarify the error message returned when you attempt to start
or stop a service which is not loaded. The previous error didn't contain
any messaging around `loaded` vs `unloaded` state and just messaged that
a service could not be located.

In addition, all other messages have been updated to reflect the same
messaging when attempting to act on a service which has not been loaded.


reset:
@baumanj great comments here. I'm going to merge this as-is because it is a significant improvement upon what we currently have. Let's continue the conversation though

Issue #5151, opened 2018-06-04
hab-sup 0.56.0 crashes if hab-sup client is killed before response is sent


stevendanna:
The title is a best guess at the cause, here is behavior I've seen while reproducing #5150

1) Block the supervisor by loading a service with a blocking init script as in #5150 
2) Attempt to load another service, this command will not return. 
3) Interrupt (CTRL+C) the command you ran in (2)
4) Unblock the supervisor by unblocking the init script from the service in (1)

The supervisor will crash with:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SendError("...")', libcore/result.rs:945:5
stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: <unknown>
  15: <unknown>
  16: <unknown>
  17: <unknown>
  18: <unknown>
  19: <unknown>
  20: <unknown>
  21: <unknown>
  22: <unknown>
  23: <unknown>
  24: <unknown>
  25: <unknown>
  26: <unknown>
  27: <unknown>
  28: <unknown>
  29: <unknown>
  30: <unknown>
  31: __libc_start_main
  32: <unknown>
```

The backtrace isn't too useful.

reset:
@stevendanna thanks for reporting this one, we'll get it fixed up!

FYI: The backtrace isn't useful because we strip symbols in our release builds

Issue #5148, opened 2018-06-04
`hab sup status origin/service-name` can take down supervisor (Results::unwrap() called on an Err)


srenatus:
In a Vagrantbox (`bento/ubuntu-16.04`, 8G mem, 4 cpus), using *0.56.0*,

- having started the supervisor: `env RUST_LOG=debug RUST_BACKTRACE=1 hab sup run &`
- when I run `hab sup status chef/deployment-service` (a service that has been loaded), I see this most of the time,
  ```
  package                                       type        state  uptime (s)  pid     group
  chef/deployment-service/0.1.0/20180601231405  standalone  down   46          <none>  deployment-service.default
  ```
- *occasionally*, however, it kills the supervisor (I've re-run the above command a few times):

```
[84][default:/src:130]# hab sup status chef/deployment-serviceDEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop poll - Duration { secs: 1, nanos: 932912 }                                                                        [10/4717]
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop time - Instant { tv_sec: 1611, tv_nsec: 565357111 }
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop process - 0 events, Duration { secs: 0, nanos: 46455 }
DEBUG 2018-06-04T08:12:13Z: tokio::reactor: loop process - 1 events, 0.000s
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop poll - Duration { secs: 43, nanos: 90746836 }
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop time - Instant { tv_sec: 1611, tv_nsec: 628163489 }
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop process - 1 events, Duration { secs: 0, nanos: 19481 }
DEBUG 2018-06-04T08:12:13Z: tokio::reactor: loop process - 1 events, 0.000s
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: added a timeout: 0
DEBUG 2018-06-04T08:12:13Z: tokio::reactor: scheduling direction for: 0
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop poll - Duration { secs: 0, nanos: 1144 }
DEBUG 2018-06-04T08:12:13Z: tokio_core::reactor: loop time - Instant { tv_sec: 1611, tv_nsec: 628320137 }
[...]

PR #5166, opened 2018-06-07
Ignore Err result in send_msg rather than panicking

baumanj:
Why we're getting an error from protocol::codec::SrvMessage::start_send is
not clear, but since we can't do much, it's better to silently ignore the error

Resolves https://github.com/habitat-sh/habitat/issues/5148
Resolves https://github.com/habitat-sh/habitat/issues/5151


thesentinels:

Thank you for contributing!

PR #5177, opened 2018-06-07
Fix serialization problem where defaults were injected

reset:
Prost will incorrectly concat the default value specified in a protocol
file to the value of a field if the field is marked as required. This
change removes the default values that were placed in the protocol file
for `types.proto` to side step the issue, but we should follow up with
the creators of Prost to get this solved for real. We weren't using the
default value in the protocol file anyway, so this isn't much of a
change.

### DIFF
--- a/components/sup-protocol/protocols/types.proto
+++ b/components/sup-protocol/protocols/types.proto
@@ -35,7 +35,7 @@ enum BindingMode {
 
 message ApplicationEnvironment {
   required string application = 1;
-  required string environment = 2 [default = "default"];
+  required string environment = 2;
 }
 
 message PackageIdent {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
The default removal looks good... do you need to rebase or update the rustfmt changes?

reset:
@chefsalim the churn is all generated code - it looks like somebody ran a formatter on it which, when it's regenerated, won't be formatted in that way ;). We don't have a blacklist for things to not be formatted and I'm not sure if rustfmt supports it but that'd be nice!

chefsalim:
👍

PR #5179, opened 2018-06-08
 Switch to `prost` from `protobuf` crate in Butterfly

reset:
This is a sister change to #5126 which removed the protobuf crate and
replaced it with the prost crate in the Supervisor proper. These changes
are house cleaning work in preparation of improving Butterfly and the
Supervisor overall.

In addition to the refactor, this change will fix a number of
serialization/deserialization bugs which were introduced at some point
into Butterfly's types.

## The Approach

The goal of this PR is to coerce or validate protocol messages received on inbound sockets into a format the Butterfly server can use. All fields of our protocol messages are marked as `optional` which has been done to ensure we can add or remove fields as we see fit between versions. Older versions of the Supervisor are designed to communicate with newer versions, and vice-versa, at least in a degraded state.

Prior to this PR we had no mechanisms in place to coerce or validate messages. We would receive a message and, in the best case, drop the message on the floor if we couldn't decode it. However, a message which could be decoded but did not have a field we required to be set would instead have a bogus default value provided. This is a source of bugs from the `protobuf` crate and the primary reason I'm ripping this out. This brings me to the two approaches that I was teetering between on how to solve this problem.

1. Receive a protocol message, wrap it in a new-type, add a validate/coerce function which returns a Result, and then add accessor functions for every field you want to read/write to
1. Receive a protocol message and validate/coerce by moving each value into a field of a concrete type
1. Receive a protocol message, validate/coerce it, and then use `unwrap` in all areas of the code assuming the validation code has caught any issues

I took the second approach in this PR. Both Option 1 and 2 are pretty shit for boilerplate setup, but it's not so bad modifying existing messages and gives us a single place to coerce/validate. I'm not entirely sure that option 3 isn't the right approach, though. It would lead to the least amount of boilerplate and allocations, but the compiler won't catch everything since you'll be essentially asserting when you unwrap in areas deeper in the codebase (ie. not the inbound socket modules). Unwraps are looked down upon in the Rust community but I think this is something we may want to consider if these changes become too much of an aggravation to manage.


thesentinels:

Thank you for contributing!

christophermaier:
Update: I've finally gotten around to rebasing and evaluating this... things build fine on a workstation, but fail in the studio. After a long debugging session, I've tracked it down to the fact that the `protoc` binary that `prost-build` provides is dynamically linked, which of course means that it's not finding what it needs in our studio environment.

We can override this with our own `protoc` binary, which means we'll need to add that as a build dependency. As our package for that is rather old, it'll need to be updated.

I'm not quite sure how the other code that uses `prost` has actually built. 🤔

christophermaier:
@reset Quick question for you: I discovered that the protobuf code generation doesn't work properly within a Studio, partly because of 

https://github.com/habitat-sh/habitat/blob/df58a0cea922e6b0bd52be1b2111883a51e950d3/components/butterfly/build.rs#L22-L27

We can't rename the files in a studio because they're on different filesystems. Obviously, we can _copy_ the files, but digging a little further, I think I've got a simpler approach. Instead of moving the files into `src/generated`, from which they'll be `include!`d into the rest of our files, we could just directly include them from `OUT_DIR`, e.g.

```rust
include!(concat!(env!("OUT_DIR"), "/butterfly.swim.rs"));
```
This seems to work fine, and eliminates the need to have the `generated` directory altogether. Was there a particular goal you were trying to accomplish using that directory that we wouldn't get with just a direct `include!` from `OUT_DIR`?

Thanks!

reset:
This is a preference of mine to commit generated code into the codebase so I know exactly what has changed with a quick git diff. I know that there’s some other people on the project who have different feelings out this and don’t find the same value.

My feelings wouldn’t be hurt if you’d prefer to not commit the generated code :)

fnichol:
So I've been thinking about the relationship we have (in some parts of the codebase) between Protobuf-derrived types and pure internal Rust types and have some to a very similar place. That is, I'm favoring approach 2 as well 😄 

I like that we went with the second option there as it's closer to how I've been thinking about dealing with wire serialized/auto generated types. Possibly a missing piece here is that if `TryFrom` (https://doc.rust-lang.org/std/convert/trait.TryFrom.html) was stable, then this might be a more obvious solution, in that our coersion/conversion from protobuf representations to pure Rust types are fallible (and should be). I've seen in other codebases that authors implement a version of `TryFrom` for their own purposes and leave a note to swap over to `TryFrom` once it stabilizes in Rust stable.

---

On the question of committing the generated code into source control or not, it's dawned on me that we don't need 1 answer for all circumstances and perhaps shouldn't anyway. In this specific case, we are representing the wire serialized representation of a specific protocol *and* we want this representation to be a) stable and b) forwards compatible. For both of these reasons we don't expect much or any nontrivial changes in the Protobuf representations in normal development. Given this, I'd favor the build-determinism of seeing this code in the source tree, knowing that it shouldn't be prone to code churn. It's also one more step to *choose* to regenerate this code (using a Cargo feature flag, or simply by running the appropriate command and have it documented) which is not a bad thing here--we don't expect there to be high churn here and it should be given close scrutiny when changed anyway. Our approach could be entirely different in the Builder codebase, which is more about messaging between unlike microservices (and we expect much higher churn here).

I'm more convinced after re-reading the approach taken in the [raft crate](https://crates.io/crates/raft) where the generated code is a part of the [source code](https://github.com/pingcap/raft-rs/blob/master/src/eraftpb.rs) and regenerating the Rust code on Protobuf changes is simply documented in the [Readme](https://github.com/pingcap/raft-rs#developing-the-raft-crate). It's refreshingly simple and straight forward.

Issue #5172, opened 2018-06-07
`hab svc status <IDENT>` exit status is inconsistent


baumanj:
Note that `hab svc status` and `hab sup status` are aliases

If no services are loaded:
```
[74][default:/src:0]# hab svc status
No services loaded.
```
Then a query for a particular service's status returns 0 (success) even if it's not present:
```
[75][default:/src:0]# hab svc status habitat/builder-datastore
No services loaded.
[76][default:/src:0]# hab svc status habitat/builder-api
No services loaded.
[77][default:/src:0]#
#                 ^ exit status
```
But as soon as there's *something* loaded, querying for a non-loaded service, give an non-zero exit:
```
[77][default:/src:0]# hab svc load habitat/builder-datastore
The habitat/builder-datastore service was successfully loaded
[78][default:/src:0]# hab svc status habitat/builder-datastore
package                                        type        state  uptime (s)  pid    group
habitat/builder-datastore/7043/20180208190943  standalone  1      4           99047  builder-datastore.defaultdefault
[79][default:/src:0]# hab svc status habitat/builder-api
✗✗✗
✗✗✗ [Err: 2] habitat/builder-api not loaded.
✗✗✗
[80][default:/src:1]#
``` 

The expected behavior is that if a service is not loaded, `hab svc status <IDENT>` will always have a non-zero exit status.

I believe this is as of the 0.56.0 supervisor.

PR #5187, opened 2018-06-12
Make querying an unloaded service status always exit nonzero

baumanj:
Resolves https://github.com/habitat-sh/habitat/issues/5172

Previously, an early exit in the case of no loaded services caused a successful
exit even if there was an <IDENT> argument. This undermined the value of
`hab svc status <IDENT>` as a check on a service being loaded.

Also, this eliminates one cause of the crash avoided by
https://github.com/habitat-sh/habitat/pull/5166.
Calling CtlRequest::reply_complete multiple times resulted in an inconsistent
state for the request. That fix should remain, since it's preferable to not
crash, but it illustrates why the CtlRequest.tx debug output was confusing.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -270,22 +270,20 @@ impl Manager {
         opts: protocol::ctl::SvcStatus,
     ) -> NetResult<()> {
         let statuses = Self::status(&mgr.cfg)?;
-        if statuses.is_empty() {
-            req.reply_complete(net::ok());
-            return Ok(());
-        }
         if let Some(ident) = opts.ident {
             for status in statuses {
                 if status.pkg.ident.satisfies(&ident) {
                     let mut msg: protocol::types::ServiceStatus = status.into();
                     req.reply_complete(msg);
-                    break;
+                    return Ok(());
                 }
             }
             return Err(net::err(
                 ErrCode::NotFound,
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -270,22 +270,20 @@ impl Manager {
         opts: protocol::ctl::SvcStatus,
     ) -> NetResult<()> {
         let statuses = Self::status(&mgr.cfg)?;
-        if statuses.is_empty() {
-            req.reply_complete(net::ok());
-            return Ok(());
```
baumanj: This is what caused the zero exit status for `hab svc status <IDENT>` even if `<IDENT>` was not loaded.

components/sup/src/manager/mod.rs
```diff
@@ -270,22 +270,20 @@ impl Manager {
...
-        }
         if let Some(ident) = opts.ident {
             for status in statuses {
                 if status.pkg.ident.satisfies(&ident) {
                     let mut msg: protocol::types::ServiceStatus = status.into();
                     req.reply_complete(msg);
-                    break;
+                    return Ok(());
```
baumanj: Without this `return`, the case of a service whose status is present in `statuses` returned `Err`, but that was masked by the call to `reply_complete`.

PR #5220, opened 2018-06-19
Make launcher exit status indicate if supervisor exited with a fatal error

baumanj:
See https://github.com/habitat-sh/habitat/issues/5169


thesentinels:

Thank you for contributing!

mwrock:
I don't see anything here that would have a negative impact on Windows or even differ there @baumanj

christophermaier:
This looks good. I think we should hold off merging right this second, because the core plans refresh brought in a change to glibc that dropped support for older Linux kernels. Until the core plans team sorts that situation out, we'll need to pin back to older versions to ensure that the Launcher still works on the older kernels we support; otherwise the next Launcher release won't work on those kernels.

baumanj:
What should our strategy be here? I think we've finished the pinning work, but how do we move forward with launcher (or any other pinned `core` package) changes at this point?

christophermaier:
@baumanj There shouldn't be a problem. Let's get this rebased and merged.

PR #5227, opened 2018-06-19
Add force option to hab bldr job cancel

chefsalim:
This adds a force option to 'hab bldr job cancel' command, that will let us more easily script cancels (for example, if we are in a situation where we need to do bulk cancels).

tenor-114597218

### DIFF
--- a/components/hab/src/command/bldr/job/cancel.rs
+++ b/components/hab/src/command/bldr/job/cancel.rs
@@ -18,15 +18,17 @@ use common::ui::{Status, UIReader, UIWriter, UI};
 use error::{Error, Result};
 use {PRODUCT, VERSION};
 
-pub fn start(ui: &mut UI, bldr_url: &str, group_id: &str, token: &str) -> Result<()> {
-    // TODO (SA): Show all the in-progress builds that will get canceled
-    let question =
-        "If you choose to cancel a group build, \
-         all of the builds that are in progress will be canceled. Is this what you want?";
+pub fn start(ui: &mut UI, bldr_url: &str, group_id: &str, token: &str, force: bool) -> Result<()> {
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/bldr/job/cancel.rs
```diff
@@ -18,15 +18,17 @@ use common::ui::{Status, UIReader, UIWriter, UI};
...
-    // TODO (SA): Show all the in-progress builds that will get canceled
-    let question =
-        "If you choose to cancel a group build, \
-         all of the builds that are in progress will be canceled. Is this what you want?";
-
-    if !ui.prompt_yes_no(question, Some(true))? {
-        ui.fatal("Aborted")?;
-        return Ok(());
+pub fn start(ui: &mut UI, bldr_url: &str, group_id: &str, token: &str, prompt: bool) -> Result<()> {
+    if prompt {
```
christophermaier: I think it might be clearer to call it `force` rather than `prompt` and have the conditional be `if !force` _here_, rather than in `main.rs`... as the code currently exists, the call site in `main.rs` is a little confusing.

baumanj: Either that or in `main.rs` do something like:
```rust
let prompt = !m.is_present("FORCE");
```
Both seem fine to me, but I agree as written it's a bit confusing.

christophermaier: Yup, that's cool, too; just something that more clearly conveys the relationship between `--force` and prompting 👍

chefsalim:
Updated to remove the prompt param and just use force @cm @baumanj

PR #5237, opened 2018-06-21
Fix an erroneous shellcheck fix

christophermaier:
Quoting the string made `hab`try to install a single package named
`"core/hab core/hab-studio"`, which subsequently fails.
    
The proper fix here is to use an array variable to store the package
identifiers. We also use `$@` instead of `$*` to make the
corresponding change for any arguments passed in.


thesentinels:

Thank you for contributing!

christophermaier:
@baumanj Good call; that's a much better fix!

PR #5238, opened 2018-06-21
Pin everything to use pre-refresh core plans

christophermaier:
The refresh brought in a version of glibc that dropped support for
older 2.6 Linux kernels. That is currently being sorted out, but until
then, this pins all core dependencies to the pre-refresh versions to
ensure that our next release continues to run on older kernels.

While some dependencies could have been pinned at the version level, I
felt it prudent to be a pinning maximalist in this case. As a result,
everything except dependencies on actual Habitat packages is pinned
with a fully-qualified identifier. (If we pinned those, we wouldn't be
able to make a release!)

### DIFF
--- a/components/backline/plan.sh
+++ b/components/backline/plan.sh
@@ -6,16 +6,14 @@ pkg_license=('Apache-2.0')
 pkg_source=nosuchfile.tar.gz
 pkg_build_deps=()
 
-pkg_deps=(
-  core/hab-plan-build
-  core/diffutils
-  core/less
-  core/make
-  core/mg
-  core/util-linux
-  core/vim
-  core/ncurses
-)
+pkg_deps=(core/hab-plan-build
+          core/diffutils/3.3/20170513214227
+          core/less/481/20170513213936
+          core/make/4.2.1/20170513214620
+          core/mg/20160118/20170513220846
+          core/util-linux/2.27.1/20170513214732
+          core/vim/8.0.0004/20170513220742
+          core/ncurses/6.0/20170513213009)
 
 do_download() {
   return 0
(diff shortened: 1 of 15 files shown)


stevendanna:
@christophermaier Looking at the glibc 2.26 changelog I see:

> * Linux kernel 3.2 or later is required at runtime, on all architectures
  supported by that kernel.  (This is a change from version 2.25 only for
  x86-32 and x86-64.)

Is that the glibc change being referenced here?

christophermaier:
@stevendanna Yes, that's the change. As a team, we're currently planning a path forward that preserves backward compatibility so Habitat will continue to run on older kernels, while still allowing users on more modern kernels to move forward. In the meantime, however, this preserves the status quo.

@baumanj I went through Builder to find the last stable release before the base plans refresh. Sadly, not automated, though I have some ideas of possible ways to do that, should this ever be required in the future.

christophermaier:
Hold on merging... I'm investigating something at the moment.

stevendanna:
@christophermaier 

>  In the meantime, however, this preserves the status quo.

For the hab binary itself yes, but does this imply that all packages built with the new refresh also have the new linux kernel requirement?  I'm guessing this doesn't matter for many of our users, but it might be good for them to know in general.

christophermaier:
@stevendanna It does... we're working on messaging for that

christophermaier:
OK, I think we can merge. I discovered some oddities in how the ACI, Mesos, and Tar exporters are depending on hab packages, but it looks like they existed prior to this. Also, the eventsrv package doesn't build, but that also predates this. I'll do further investigation of those issues and follow up.

PR #5254, opened 2018-06-25
Simplify tar exporter dependencies

christophermaier:
After the tar exporter was rewritten in Rust, it became a
statically-compiled binary. However, we accidentally retained the
runtime dependencies of the older, shell-script-based exporter. These
dependencies are unnecessary, and can thus be removed.

Additionally, as the binary is statically compiled, we can remove the
pins altogether (these were too-conservatively added _en masse_
following the June 2018 base plans refresh that inadvertently dropped
support for older Linux kernels).


thesentinels:

Thank you for contributing!

PR #5261, opened 2018-06-26
Add a feature-flagged `--ignore-local` flag to `hab pkg install`

christophermaier:
Following the base plans refresh of June 2018, we discovered a small
bug in how our package installation logic could resolve dependencies
that are part of the Studio. In particular, this can occur when using
the `HAB_BLDR_CHANNEL` environment variable to specify an
other-than-stable channel to pull packages from.

There is logic in the Rust code that will end up using a
locally-installed package if something cannot be found in the
specified channel in Builder. If this is the case, the plan build
software will not fall back to look in the `stable` channel. This
would be an issue for us when building the next Habitat release,
because the first parts of our release are built with the
currently-stable Habitat, which was built before the base plans
refresh. Because of how we build using a release channel, we would end
up pulling in some dependencies based on what is installed in the
Studio, rather than allowing the build logic to fall back to pull from
stable. We can then run into issues later in the release process
building later packages that do end up pulling in newer dependencies;
we end up with multiple versions and thus cannot build.

For the next release, we will have these dependencies pinned
explicitly to ensure we get pre-refresh packages.

That brings us to this commit.

This introduces an `--ignore-local` flag for `hab pkg install` that
disables this "fallback to a local package" logic. Once this is
present in a stable `hab` release, we'll be able to build using
packages exclusively from Builder, as `hab-plan-build` has been
modified to use it.

This issue probably only shows up when you "straddle" a base plans
refresh like this.

Additionally, there may be better ways to ultimately handle this,
which is why this capability is currently behind a feature-flag. To
use it, you must set the environment variable `HAB_FEAT_IGNORE_LOCAL`
to `true`, which makes the `--ignore-local` option available on the
`hab pkg install` command. As such, this behavior *IS NOT* guaranteed
to remain in the `hab` CLI, pending a solidification of the desired
behavior.

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -178,6 +178,43 @@ impl Default for InstallMode {
     }
 }
 
+/// When querying Builder, we may not find a package that satisfies
+/// the desired package identifier, but we may have such a package
+/// already installed locally. In most cases, it should be fine for us
+/// to use the locally-installed package. However, it can cause issues
+/// when building packages using HAB_BLDR_CHANNEL, due to how the
+/// fallback logic in `hab-plan-build` is currently implemented.
+///
+/// This enum governs whether or not we should use a locally-installed
+/// package to satisfy a dependency, or if we should ignore it, thus
+/// giving the user the opportunity to try installing from another
+/// channel.
+///
+/// Usage of this is currently hidden behind the IGNORE_LOCAL feature
+/// flag, as there is still some question as to the best way to solve
+/// this.
+#[derive(Debug, Eq, PartialEq)]
+pub enum LocalPackageUsage {
+    /// Use locally-installed packages if they satisfy the desired
+    /// identifier, but a package cannot be found in Builder.
+    ///
+    /// This *may* be a different package than Builder may have found
+    /// in another channel.
+    Prefer,
+    /// Do not use locally-installed packages if a package cannot be
+    /// found in Builder.
+    Ignore,
+}
+
+impl Default for LocalPackageUsage {
+    /// The default behavior is to use locally-installed packages if
+    /// they can satisfy the desired identifier, and if no more
+    /// suitable package could not be found in Builder.
+    fn default() -> Self {
+        LocalPackageUsage::Prefer
+    }
+}
+
 /// Represents a release channel on a Builder Depot.
 // TODO fn: this type could be further developed and generalized outside this module
 struct Channel<'a>(&'a str);
@@ -280,6 +317,7 @@ pub fn start<U, P1, P2>(
     artifact_cache_path: P2,
     token: Option<&str>,
     install_mode: &InstallMode,
+    local_package_usage: &LocalPackageUsage,
 ) -> Result<PackageInstall>
 where
     U: UIWriter,
@@ -298,6 +336,7 @@ where
 
     let task = InstallTask::new(
         install_mode,
+        local_package_usage,
         url,
         channel,
         product,
@@ -315,6 +354,7 @@ where
 
 struct InstallTask<'a> {
     install_mode: &'a InstallMode,
+    local_package_usage: &'a LocalPackageUsage,
     depot_client: Client,
     channel: Channel<'a>,
     fs_root_path: &'a Path,
@@ -326,6 +366,7 @@ struct InstallTask<'a> {
 impl<'a> InstallTask<'a> {
     fn new(
         install_mode: &'a InstallMode,
+        local_package_usage: &'a LocalPackageUsage,
         url: &str,
         channel: Channel<'a>,
         product: &str,
@@ -336,6 +377,7 @@ impl<'a> InstallTask<'a> {
     ) -> Result<Self> {
         Ok(InstallTask {
             install_mode: install_mode,
+            local_package_usage: local_package_usage,
             depot_client: Client::new(url, product, version, Some(fs_root_path))?,
             channel: channel,
             fs_root_path: fs_root_path,
@@ -457,6 +499,7 @@ impl<'a> InstallTask<'a> {
             // package is installed as a signal that it can satisfy a "latest" answer. Checking for
             // any cached artifacts is too aggressive in this case: if you really want that cached
             // version to win--install it first, then it will be counted.
+
             let latest_local = self.latest_installed_ident(&ident);
 
             ui.status(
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -390,7 +392,7 @@ function _Set-HabBin {
 
 function _install-dependency($dependency) {
   if (!$env:NO_INSTALL_DEPS) {
-    & $HAB_BIN install -u $env:HAB_BLDR_URL --channel $env:HAB_BLDR_CHANNEL $dependency
+    & $HAB_BIN install -u $env:HAB_BLDR_URL --channel $env:HAB_BLDR_CHANNEL --ignore-local $dependency
     if ($LASTEXITCODE -ne 0 -and ($env:HAB_BLDR_URL -ne $FALLBACK_CHANNEL)) {
       Write-BuildLine "Trying to install '$dependency' from '$FALLBACK_CHANNEL'"
       & $HAB_BIN install -u $env:HAB_BLDR_URL --channel $FALLBACK_CHANNEL $dependency
```
mwrock: I believe you need `--ignore-local` here too

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -86,6 +86,8 @@ $env:HAB_BLDR_URL = "$script:HAB_BLDR_URL"
 if (!(Test-Path Env:\HAB_BLDR_CHANNEL)) {
     $env:HAB_BLDR_CHANNEL = "stable"
 }
+# Enable the `--ignore-local` option for `hab pkg install`
+$env:HAB_FEAT_IGNORE_LOCAL = "true"
```
mwrock: this would "work" but the powershell convention would be `$env:HAB_FEAT_IGNORE_LOCAL = $true`

components/common/src/command/package/install.rs
```diff
@@ -467,51 +511,63 @@ impl<'a> InstallTask<'a> {
...
+                    if self.ignore_locally_installed_packages() {
+                        // This is the behavior that is currently feature-flagged
+                        self.recommend_channels(ui, &ident, token)?;
+                        ui.warn(format!(
+                            "Locally-installed package '{}' would satisfy '{}', \
+                             but we are ignoring that as directed",
+                            local.as_ref(),
+                            &ident,
+                        ))?;
+                        Err(Error::PackageNotFound("If you are attempting to install from a local depot with an upstream configured, try again in a few seconds.".to_string()))
```
baumanj: I think there's an error with `rustfmt` here. When I make changes to this line that should violate the standard (introducing random spaces) and are corrected elsewhere, `rustfmt` doesn't seem to notice.

I'd expect us to be consistent about whether long strings are split up and wrapped at column 100 like this:
```rust
                        Err(Error::PackageNotFound(
                            "If you are attempting to install from a local depot with an upstream \
                             configured, try again in a few seconds."
                                .to_string(),
                        ))
```

components/common/src/command/package/install.rs
```diff
@@ -467,51 +511,63 @@ impl<'a> InstallTask<'a> {
...
+                            ),
+                        )?;
+                        FullyQualifiedPackageIdent::from(local.as_ref().clone())
                     }
                 }
-                Err(_) => Ok(latest_remote),
+                (Err(_), Some(remote)) => Ok(remote),
+                (Err(_), None) => {
+                    self.recommend_channels(ui, &ident, token)?;
+                    Err(Error::PackageNotFound("If you are attempting to install from a local depot with an upstream configured, try again in a few seconds.".to_string()))
```
baumanj: Same comment as above.

components/hab/src/cli.rs
```diff
@@ -737,15 +737,25 @@ fn sub_pkg_install() -> App<'static, 'static> {
...
+    };
+    sub = if feat::is_enabled(feat::IgnoreLocal) {
+        sub.arg(
+            Arg::with_name("IGNORE_LOCAL")
+                .help("Do not use locally-installed packages when a corresponding package cannot be installed from Builder")
+                .long("ignore-local"),
+        )
+    } else {
+        sub
+    };
```
baumanj: Just out of curiosity is there a reason why
```rust
    sub = if feat::is_enabled(feat::OfflineInstall) {
        sub.arg(
            …
        )
    } else {
        sub
    };
```
rather than
```rust
    if feat::is_enabled(feat::IgnoreLocal) {
        sub = sub.arg(
            …
        )
    }
```
?

I understand if you're maintaining consistency with the existing pattern above, but I wanted to make sure there wasn't a deeper reason like a rust thing that I wasn't aware of.

christophermaier: Nah, no reason... your way is cleaner; will change.

components/common/src/command/package/install.rs
```diff
@@ -178,6 +178,44 @@ impl Default for InstallMode {
     }
 }
 
+/// When querying Builder, we may not find a package that satisfies
+/// the desired package identifier, but we may have such a package
+/// already installed locally. This enum governs whether or not we
+/// should use the local package to satisfy the requirement.
+///
+/// In most cases this should be fine (and was the behavior prior to
```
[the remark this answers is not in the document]
christophermaier: The identifier does not need to be a fully-qualified one, but yes, that's what "this" refers to.

Could you elaborate on how the wording is unclear? I'm not seeing it (but I wrote it, so I may be blind to the ambiguity).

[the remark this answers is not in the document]
baumanj: In the text prior to `this should be fine` we have three clauses `this` could refer to:

1. When querying Builder, we may not find a package that satisfies the desired package identifier
2. we may have such a package already installed locally.
3. This enum governs whether or not we should use the local package to satisfy the requirement.

And with the parenthetical `was the behavior`, it implies that `this` must refer to a behavior. But 1 and 2 describe situations, not behaviors and 3 describes a new behavior, not the prior one. So that's what made me confused.

If I'm understanding correctly, what `this` refers to is `if Builder doesn't find a package that satisfies the package identifier, but there is a locally installed package that does, we will use the locally installed package`. That is alluded to by the following text, but not present in the text that precedes the `this` on line 186.

Does that make sense? If not, let's jump on a zoom. I'll just be hanging out in my personal room since Slack seems to be down at the moment.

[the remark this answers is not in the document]
christophermaier: Ah, OK, I see... I'll reword it!

fnichol:
I guess what I'm seeing with the `_install_dependency()` logic is that it's avoiding a purely informational request to the API and rather performing that operation indirectly in an install task which we already know might fail (hence that fallback `||` logic). That's maybe not relevant to this particular change, but I'm more uneasy about the rest of that function now as a result.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -716,10 +716,18 @@ _resolve_dependency() {
 _install_dependency() {
     local dep="${1}"
     if [[ -z "${NO_INSTALL_DEPS:-}" ]]; then
-    $HAB_BIN install -u $HAB_BLDR_URL --channel $HAB_BLDR_CHANNEL "$dep" || {
+
+    # Enable --ignore-local if invoked with HAB_FEAT_IGNORE_LOCAL in
+    # the environment, set to either "true" or "TRUE" (features are
+    # not currently enabled by the mere presence of an environment variable)
+    if [[ "${HAB_FEAT_IGNORE_LOCAL:-}" = "true" ||
+              "${HAB_FEAT_IGNORE_LOCAL:-}" = "TRUE" ]]; then
```
baumanj: I think it's worth considering the value of considering any non-empty value here to be true. This still won't work with `t`, `T`, `1`, `yes`, etc. The code for doing it the other way is nice and simple:
```bash
    if [[ ${HAB_FEAT_IGNORE_LOCAL:-} ]]; then
```

christophermaier: I completely agree. This reflects the current behavior of how feature flags are enabled, though, so I wanted to stay consistent. I'm gonna file an issue to fix that up more broadly.

christophermaier: @baumanj See #5275

Issue #4762, opened 2018-03-15
[On-Prem Depot] Supervisor should respect HAB_BLDR_URL env


chefsalim:
In order to support workflow where we have an on-premise installation of Habitat Depot, we need the supervisor to be able to respect the HAB_BLDR_URL environment variable when downloading or looking for package updates.

baumanj:
The supervisor *does* respect the `HAB_BLDR_URL` environment variable. The code for that is [here](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/main.rs#L365-L370).

I added a print to dump the config and confirmed the correct behavior:
```
➤ env HAB_BLDR_URL="https://hab.foo" HAB_SUP_BINARY=/home/jbauman/habitat/target/debug/hab-sup  RUST_LOG= hab sup run
cfg: ManagerConfig { auto_update: false, custom_state_path: None, eventsrv_group: None, update_url: "https://hab.foo", update_channel: "stable", gossip_listen: GossipListenAddr(V4(0.0.0.0:9638)), ctl_listen: V4(127.0.0.1:9632), http_listen: ListenAddr(V4(0.0.0.0:9631)), gossip_peers: [], gossip_permanent: false, ring_key: None, name: None, organization: None, watch_peer_file: None }
```

It does look like the supervisor CLI help needs updating, though. I'll close this when that's complete.

PR #5263, opened 2018-06-27
Make all BLDR_URL CLI docs reference the HAB_BLDR_URL env var override

baumanj:
Resolves https://github.com/habitat-sh/habitat/issues/4762


thesentinels:

Thank you for contributing!

baumanj:
That sounds like a good idea, @elliott-davis. Why don't we do it in a separate PR? The point of this one was to address https://github.com/habitat-sh/habitat/issues/4762 (and a couple other places that had identical issues), so I'm trying to keep it a single logical change.

I'm very much in favor of reducing duplication like this. Last time I went down that road it was a major slog, but it sounds like you may be aware of a clap feature that I'm not.

PR #5298, opened 2018-07-06
Handle return case of Ok(None) for pkg.svc_user() and pkg.svc_group()

smacfarlane:
When troubleshooting habitat-sh/core-plans#1656 I discovered that the docker exporter isn't handling the case where a package is considered a service, but the SVC_USER / SVC_GROUP metadata isn't present.   In this case the return value is `Ok(None)`,  which is then unwrapped twice causing a panic.  

I chose to treat the `Ok(None)` case the same as the `Error` case, which is to turn it into the default user of "hab".  

I tested this by running the export against a package with known issues `core/postgresql95`,  a known good package `core/postgresql` and a combination of a non-service package with a service package `core/gcc-libs core/postgresql`.

### DIFF
--- a/components/pkg-export-docker/src/build.rs
+++ b/components/pkg-export-docker/src/build.rs
@@ -26,7 +26,9 @@ use common::command::package::install::{InstallMode, InstallSource, LocalPackage
 use common::ui::{Status, UIWriter, UI};
 use failure::SyncFailure;
 use hab;
+use hcore::error as hcoreerror;
 use hcore::fs::{cache_artifact_path, cache_key_path, CACHE_ARTIFACT_PATH, CACHE_KEY_PATH};
+use hcore::package::metadata::MetaFile;
 use hcore::package::{PackageArchive, PackageIdent, PackageInstall};
 use hcore::PROGRAM_NAME;
 use tempdir::TempDir;
@@ -525,12 +527,18 @@ impl BuildRootContext {
 
         let pkg = self.primary_svc()?;
         let user_name = match pkg.svc_user() {
-            Ok(Some(user)) => user, 
-            _ => String::from("hab")
+            Ok(Some(user)) => user,
+            Ok(None) => {
+                return Err(hcoreerror::Error::MetaFileNotFound(MetaFile::SvcUser).into());
+            }
+            Err(_) => String::from("hab"),
         };
         let group_name = match pkg.svc_group() {
-            Ok(Some(group)) => group, 
-            _ => String::from("hab")
+            Ok(Some(group)) => group,
+            Ok(None) => {
+                return Err(hcoreerror::Error::MetaFileNotFound(MetaFile::SvcGroup).into());
+            }
+            Err(_) => String::from("hab"),
         };
 
         // TODO: In some cases, packages based on core/nginx and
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
Hrmm... wonder if something went astray elsewhere; this seems inconsistent with what https://github.com/habitat-sh/core/blob/d754c88e6b4ca131c8447aeeaaa7eba83d6b7686/components/core/src/package/archive.rs#L182-L191 is saying.

christophermaier:
Looks like there should _always_ be an `SVC_USER` and `SVC_GROUP` metadata file if it's a service: https://github.com/habitat-sh/habitat/blob/b0db32b76b6d24e2c592e67ed522388f6522598e/components/plan-build/bin/hab-plan-build.sh#L1896-L1902

smacfarlane:
@christophermaier  Here's the changes,  existing behavior is preserved while giving the user a nicer error message when SVC_USER/GROUP metadata doesn't exist.    

It feels a little awkward returning an `Error` on `Ok(None)`, but turning and Error into `"hab"`,  but that is the current behavior, unless I'm reading it wrong :) .  Happy to make another update if want to change that.

christophermaier:
@smacfarlane I agree it does feel odd... I think the current behavior is to panic, though, which is also something we'd like to avoid.

I think ideally we'd have an "is this package a service?" check well before we got to this point. If it had worked like that, you would have quickly gotten an error back when you first tried to export the postgresql95 package; it would have said "Hey, that's not a service!".

With that in place, we could safely do the unwrap that we're currently doing, because we would have ensured that that it would never fail (i.e., that you couldn't ever get `Ok(None)` when grabbing the service user from metadata).

smacfarlane:
I was digging into this a bit more, and we do check if it is runnable,  but we rely on the PackageInstall implementation of `is_runnable()`  https://github.com/habitat-sh/core/blob/master/components/core/src/package/install.rs#L203 which checks for presence a run hook.  If we instead used PackageArchive's `is_a_service()` https://github.com/habitat-sh/core/blob/master/components/core/src/package/archive.rs#L186  we would have been spared this.  

In my digging I also noticed both PackageInstall and PackageArchive implement svc_user() with slightly different behavior.  

https://github.com/habitat-sh/core/blob/master/components/core/src/package/archive.rs#L252-L258
https://github.com/habitat-sh/core/blob/master/components/core/src/package/install.rs#L564-L570

Would it be better to close this out and make the appropriate changes in the core crate, adjust things to use the PackageArchive instead of PackageInstall, or will this be an acceptable band-aid?

christophermaier:
Pinging @fnichol on this ☝️

smacfarlane:
@christophermaier I horked something up in my commit.  If you'd have a moment to help me untangle this I'd appreciate it.

baumanj:
@smacfarlane: I'm guessing you rebased on a different branch than you intended or something. Can you identify which commits you want to keep?

smacfarlane:
@christophermaier   I think this is almost complete, but  I haven't had the cycles to do the validation on the changes made.     I'm going to close this out for now and revisit in the future when I have some more time.  Thanks for the assistance and feedback!

PR #5330, opened 2018-07-13
Update to allow postrun to only run after a successful HealthCheck

jamessewell:
Three changes:

- postrun removed from run on init to run exactly once after HealthCheck is OK
- re-run postrun after each service Start
- changed self.health_check so it gets assigned (it didn't before)

This does somewhat change the behaviour of postrun, which used to only
run at init - this part can be backed out if needed.

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -103,6 +103,7 @@ pub struct Service {
     pub pkg: Pkg,
     pub sys: Arc<Sys>,
     pub initialized: bool,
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

baumanj:
Did you mean to close this, @jamessewell ?

baumanj:
Oh, I see. You opened https://github.com/habitat-sh/habitat/pull/5331 instead. I'll try to look at that today or tomorrow.

Issue #5327, opened 2018-07-12
Take service readiness into account


christophermaier:
This is related to #5326 but somewhat distinct. If health checks can be incorporated into ongoing service state dissemination, they could also be used to gate when the Supervisor broadcasts that it's actually running that service. In other words, we would only introduce new service rumors for services that are healthy and running. This can be particularly useful for services with long startup times, and by extension, the services that depend on them via binds.

Kubernetes does this with their readiness probes.

If we do this, it's also worth considering if it's worth adding a distinct `readiness` hook.

mwrock:
Would/could this also impact rolling updates? I have a use case where it can take severla seconds for a run hook to actually get a service into a properly running state. Currently that means that rolling updates bring everyone down because while the code may be update one at a time, the rolling update does not wait for each member to be available.

Ideally a rolling update should move to the next member (or a member should only be marked `updated`) once the updated member is healthy.

christophermaier:
@mwrock Yes! I think this should affect rolling updates, too.

themightychris:
It would be amazing if habitat's lifecycle could remove the need to have sleep loops in any hooks waiting for a service to come up, they usually end up running similar code to the health check

christophermaier:
@themightychris Yup, I think that's the end goal here. The current structure of the code makes that a bit hard, but we've got a pathway to fixing that.

Issue #5326, opened 2018-07-12
Incorporate health into gossip


christophermaier:
Health checks currently do not participate in the Habitat network; they're just run on a timer on the side of the main service loop, and report information only via the HTTP gateway. Other services in the Habitat network that depend on an unhealthy service have no way of knowing whether that service is actually healthy.

We'll need to think about how to best do this; too eagerly broadcasting that a service is unhealthy could have cascading effects, particularly if the failing health check is only a transient issue; we should have some kind of threshold like "the last X health checks failed; it's officially Unhealthy". This is basically the Service-level analog of the SWIM suspicion mechanism.

We'll also need to think of how to best expose this in templating data so dependent services can take advantage of it. As stated in #5325, we currently conflate presence of a Supervisor with the presence/health of the services running on that Supervisor. In our templating data, we currently only present service group members that are either "alive" or "suspect" (these are, of course, Supervisor-level states, and not Service-level states); we can probably just flip this over to using presence (#5325) and / or health (this issue) and preserve the desired semantics (and actually be correct about it :smile: )

We may also want to expose some service-level runtime configuration options to control the frequency and threshold of checking: `hab svc load foo/bar --check-every=20s --unhealthy-after=3`, or similar. Right now, health checks occur with a hard-coded 30 second period. We _could_ add additional metadata to packages, but that may be to constraining; we'd likely still want to be able to override at runtime.

jamessewell:
This is a great idea - I think if Habitat wants health checks to feel like less of a bolt-on it needs something like this.

Could implementing this help with the ground work of allowing Health Checks to be used for triggering a Leader/Follower topology failover as well ([#3249](https://github.com/habitat-sh/habitat/issues/3249)?

christophermaier:
@jamessewell Yes, for sure this would be fundamental for #3249.

christophermaier:
Still needed.

Issue #5318, opened 2018-07-12
Add new hooks for more complete service lifecycle control


christophermaier:
In order to run effectively, services may need more control over their lifecycle than Habitat currently provides. 

By using new hooks that run at additional control points, we can provide users with Habitat packages that are better tuned to the specific needs of their services.

Some concrete examples are provided in the linked issues, but further research and community engagement should be done to determine the final suite of hooks to add.

themightychris:
I was just asking about something like this in Slack. Are you talking about defining a bunch upfront or letting users add their own arbitrary service hooks?

I'm picturing something like `hab pkg exec` but instead `hab svc exec` that takes a service identifier and runs a rendered hook of your own with proper environment

PR #5331, opened 2018-07-13
Update to allow postrun to only run after a successful HealthCheck

jamessewell:
Three changes:

- postrun removed from run on init to run exactly once after HealthCheck is OK
- re-run postrun after each service Start
- changed self.health_check so it gets assigned (it didn't before)

This does somewhat change the behaviour of postrun, which used to only
run at init - this part can be backed out if needed.

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -103,6 +103,7 @@ pub struct Service {
     pub pkg: Pkg,
     pub sys: Arc<Sys>,
     pub initialized: bool,
+    pub postrun: bool,
     pub user_config_updated: bool,
 
     #[serde(skip_serializing)]
@@ -186,6 +187,7 @@ impl Service {
                 fs::svc_hooks_path(&service_group.service()),
             ),
             initialized: false,
+            postrun: false,
             last_election_status: ElectionStatus::None,
             needs_reload: false,
             needs_reconfiguration: false,
@@ -259,6 +261,7 @@ impl Service {
         } else {
             self.needs_reload = false;
             self.needs_reconfiguration = false;
+            self.postrun = false;
         }
     }
 
@@ -680,7 +683,12 @@ impl Service {
     }
 
     fn post_run(&mut self) {
+        if self.postrun {
+            return;
+        }
+        self.postrun = true;
         if let Some(ref hook) = self.hooks.post_run {
+            outputln!(preamble self.service_group, "Execute Postrun hook");
             hook.run(
                 &self.service_group,
                 &self.pkg,
@@ -812,9 +820,11 @@ impl Service {
             self.initialize();
             if self.initialized {
                 self.start(launcher);
-                self.post_run();
             }
         } else {
+            if !self.postrun && self.health_check == HealthCheck::Ok {
+                self.post_run();
+            }
             self.check_process();
             match self.last_health_check {
                 Some(last_check) => {
@@ -902,6 +912,7 @@ impl Service {
             }
         };
         self.last_health_check = Some(Instant::now());
+        self.health_check=check_result;
         self.cache_health_check(check_result);
     }
 


thesentinels:

Thank you for contributing!

themightychris:
@jamessewell I just got up and running with this, looks good!

So far I've verified that:

- `post-run` doesn't execute until after the first successful `health_check`
- if `heatlh_check` fails, `post-run` doesn't run
- if `health_check` fails initially, and succeeds later, `post-run` does run

Some issues I see though:

1. It takes quite a while for the first `health_check` to run
   - With `post-run` deferred until then this means it takes quite a while for the service to be finished starting up, while the supervisor reports it as up
   - Perhaps an initial `health_check` could be run immediately after `run` like `post-run` used to?
2. `post-run` does not get re-run when a config change causes it to be recompiled and the service reloaded

jamessewell:
Hi Chris,

I think it’s actually doing a first health check quickly, which fails -
then it’s a 30sec standoff till the next one.

If you monitor the health check state it goes

- UNKNOWN
- FAILING
- OK

I’ll have a bit more of a poke - but the propsed changes below are much
better than this solution!

Cheers,
James Sewell

baumanj:
@jamessewell, are you still planning additional changes to this PR, or should it be considered final for review at this point?

baumanj:
I'm going to close this for now, so that our PR reminders don't think we're ignoring it. Whenever you're ready, feel free to reopen @jamessewell.

jamessewell:
I’m not really sure what to do about this one - I was hoping Chris would
chime back in.

It works, but the other proposed (larger) solution is better.

baumanj:
@christophermaier, can you weigh in?

Sorry if I was premature, @jamessewell, I was reading
> I’ll have a bit more of a poke
as an indication I should hold off on review until you had done more.

christophermaier:
Sorry @jamessewell ... I'll get around to reviewing this soon.

themightychris:
I was just wondering, would it make sense to leave the current `post-run` behavior alone and introduce this as a new hook, `post-up`/`post-available`/`post-online`/`post-healthy`?

This would have the benefit of definitely not breaking any existing services, and allows `post-run` code to play a role in getting a service into the healthy state which might be important

jamessewell:
That makes sense - although I do wonder what the point of the old post run
would be apart from backwards compatibility?

christophermaier:
@jamessewell Sorry it's taken a while to get to this!

I don't want to change the existing contract of `post-run` only running after the initial startup of the service, even though I'm not 100% convinced that it's current behavior is quite correct. There are a handful of non-trivial `post-run` hooks in the wild that this could negatively affect. There very well may be room for additional lifecycle hooks, though, as suggested above by @themightychris; I'd be very interested in your thoughts on some of those.

Also, given the relatively long time until a health-check initially fires, this could cause services to be in a potentially incomplete state for a long time. This wouldn't be a problem after #5327, and possibly #5326, are implemented, though, since services wouldn't be available to the rest of the network until they're healthy (and, presumably, after their `post-run` hook has been successful).

There's a lot of work currently planned around all the lifecycle hooks (#5318) (and I'm currently starting work on them), and I think this PR points out some additional real issues. As is, though, I think the potential for introducing additional instability and breakage is high, so I'm going to close this for now.

I appreciate the work and effort you've put in thus far, and apologize for taking as long as I have to give you some feedback on this.

themightychris:
@christophermaier deferring services being available to the rest of the network doesn't solve the use case here. All the related PRs you linked to are great and related, but they are far broader in scope than the use case at hand here:

- CI tests currently flag `sleep` in `post-run`
- There is a common need for packages to run hook code after a service comes up to complete configuration. These are things you might otherwise want to do in `init`, but need to issue commands to a running service rather than templating config files to affect
- The current ways to do that, by example/precedent/documentation, are extremely gnarly. I dug through habitat history and tried both on the way to habitat-sh/core-plans#1674:
  - [`init` hooks that run a service once, background it and store its pid, wait for it to come up, does its init, then kill it by pid and start it a second time](https://github.com/JarvusInnovations/habitat-core-plans/commit/89c0a9e)
  - [After `post-run` became a thing people started doing sleep loops there for this](https://github.com/JarvusInnovations/habitat-core-plans/compare/89c0a9e...4b73494)
- Both are blocked by CI policy now, and they probably should be
- Both tend towards replicating or stripping down the code in `health_check` 

I agree it's probably best not to change `post-run`, there's definitely a need for it and it would be bad to break all the existing ones, but a lot of those existing ones don't pass current standards anymore because they're working around not having a post-healthy hook

@jamessewell I think a very simple derivation of this PR would stop this gnarly from spreading and let packages that need this start passing CI again:

- Call a new `post-healthy` hook after the first `health_check` pass 
- Extra awesome sauce: when a `post-healthy` hook is present, accelerate initial `health_check` frequency to every 2 seconds up to a limit of like 5 tries

christophermaier:
Posted in Slack, but copied here for visibility / posterity:

> I think it _may_ be the right way to go. At this point, though, I think it might be a better idea to have some kind of RFC / broader discussion around exactly what the ideal set of lifecycle hooks should be, and really nail down their semantics with a lot of input from the broader community. I'd like to avoid a situation where we keep adding hooks to get past the problem-of-the-day (which may be due to current Supervisor implementation details) and end up with an ultimately confused and incoherent global picture of things. If that sounds useful, I'll start some discussions today to see what we can do to get that started.

Issue #4673, opened 2018-02-26
Clean up `/tmp/rust-ipc-socket.XXXXXX` files


christophermaier:
The Launcher and Supervisor processes use a pair of sockets to communicate with each other, which are written out to `/tmp/rust-ipc-socket.XXXXXX` files. Furthermore, each restart of the Supervisor generates a new pair of socket files. 

It appears that we don't always do a good job of cleaning up these files (if at all), resulting in some users reporting the accumulation of a large number of these files. We should do better cleanup.

In the meantime, only the two most recently-created socket files are necessary, so if you encounter this, feel free to remove any unneeded files. (However, if you happen to be running other Rust binaries on the same host that use the [ipc-channel](https://github.com/servo/ipc-channel) crate, some of those socket files might be in use by those binaries; in that case, proceed with caution.)

christophermaier:
Pinging @bdangit and @chrisortman since they've raised the issue in Slack before.

bdangit:
Hi @christophermaier,

Yes.  I spent sometime deleting the files in /tmp. There’s definiately a lot more than can fit within the standard `*` so I had to iterate to remove files this way: 

```bash
rm -rf /tmp/rust-ipc-socket.A*
rm -rf /tmp/rust-ipc-socket.B*
...
rm -rf /tmp/rust-ipc-socket.Z*
rm -rf /tmp/rust-ipc-socket.a*
rm -rf /tmp/rust-ipc-socket.b*
...
rm -rf /tmp/rust-ipc-socket.z*
rm -rf /tmp/rust-ipc-socket.0*
rm -rf /tmp/rust-ipc-socket.1*
...
rm -rf /tmp/rust-ipc-socket.9*

```

I’m sures some for loop we can create to remove them.  

I also do remember seeing an issue from that crate where they want to use some other method for creating the ipc-channel (one that doesn’t involve generating a tmp file that plagues shm technique.)

bdangit:
https://github.com/servo/ipc-channel/issues/125

Looks like memfd was considered but it’s a Linux only feature.

smartb-pair:
As requested by @christophermaier, wanted to leave some notes here about our experience with this behavior. Habitat version is `0.53.0`:

Total number of tempfiles:
```
$ find /tmp -name rust-* | wc -l
339298
```
First and last datestamps of files in `/tmp`:
```
$ find /tmp -name rust-* | xargs ls -l | head -n1
[...]

PR #5365, opened 2018-07-20
Remove rust-ipc-socket files when appropriate

baumanj:
Also add a test and some more logging

Resolves https://github.com/habitat-sh/habitat/issues/4673

### DIFF
--- a/components/launcher-client/src/client.rs
+++ b/components/launcher-client/src/client.rs
@@ -44,11 +44,13 @@ impl Drop for LauncherCli {
 }
 
 impl LauncherCli {
-    pub fn connect(pipe: String) -> Result<Self> {
-        let tx = IpcSender::connect(pipe).map_err(Error::Connect)?;
-        let (ipc_srv, pipe) = IpcServer::new().map_err(Error::BadPipe)?;
+    pub fn connect(pipe_to_launcher: String) -> Result<Self> {
+        debug!("LauncherCli::connect({})", pipe_to_launcher);
+        let tx = IpcSender::connect(pipe_to_launcher).map_err(Error::Connect)?;
+        let (ipc_srv, pipe_to_sup) = IpcServer::new().map_err(Error::BadPipe)?;
+        debug!("IpcServer::new() returned pipe_to_sup: {}", pipe_to_sup);
         let mut cmd = protocol::Register::new();
-        cmd.set_pipe(pipe);
+        cmd.set_pipe(pipe_to_sup);
         Self::send(&tx, &cmd)?;
         let (rx, raw) = ipc_srv.accept().map_err(|_| Error::AcceptConn)?;
         Self::read::<protocol::NetOk>(&raw)?;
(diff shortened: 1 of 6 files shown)


christophermaier:
@baumanj It looked like there may be a way to avoid using temp files for these pipes:

* https://github.com/servo/ipc-channel/issues/125
* https://github.com/servo/ipc-channel/pull/140

Would that work here?

baumanj:
I've spent a bit of time looking at the docs for [`IpcSharedMemory`](https://doc.servo.org/ipc_channel/ipc/struct.IpcSharedMemory.html) and [the test which uses it](https://github.com/servo/ipc-channel/blob/b225f6d1c03c32963139ca43ac10a19c866a642b/src/test.rs#L289), but I'm having a hard time seeing how it's supposed to work between processes.

With [`IpcOneShotServer`](https://doc.servo.org/ipc_channel/ipc/struct.IpcOneShotServer.html), `new` returns a `String` which identifies the pipe and can be passed from the launcher to the supervisor. `IpcSharedMemory` doesn't expose anything like that. Looking at the implementation, its [opaque backing store](https://github.com/dlrobertson/ipc-channel/blob/master/src/platform/unix/mod.rs#L667-L681) uses a file descriptor, so that couldn't be shared across separate processes as far as I can tell. And even if it could, that data is hidden from us. I could see where the `IpcSharedMemory` object itself could be cloned and shared across threads, but not processes, I don't think.

The documentation for this is rather lacking, but I can't really see a way to make this work for our purposes. Do you?

components/launcher-client/src/client.rs
```diff
@@ -40,7 +52,11 @@ impl LauncherCli {
         Self::send(&tx, &cmd)?;
         let (rx, raw) = ipc_srv.accept().map_err(|_| Error::AcceptConn)?;
         Self::read::<protocol::NetOk>(&raw)?;
-        Ok(LauncherCli { tx: tx, rx: rx })
+        Ok(LauncherCli {
```
christophermaier: It might be good to name the incoming `pipe` variable and the `pipe` variable returned by the `IpcServer::new()` call differently to avoid potential confusion with exactly which thing we're storing.

components/launcher/src/server/mod.rs
```diff
@@ -76,21 +85,31 @@ impl Server {
...
     #[allow(unused_must_use)]
     fn reload(&mut self) -> Result<()> {
         self.supervisor.kill();
         self.supervisor.wait();
-        let ((rx, tx), supervisor) = Self::init(&self.args, true)?;
+        let ((rx, tx), supervisor, pipe) = Self::init(&self.args, true)?;
         self.tx = tx;
         self.rx = rx;
         self.supervisor = supervisor;
+        self.remove_pipe();
```
christophermaier: A little bit of an explanatory comment about why we're "removing a pipe before saving a pipe" might allay potential confusion when reading this code for the first time.

christophermaier: Ooooh, I like that.

christophermaier:
@baumanj Looks good 👍

PR #5369, opened 2018-07-23
Avoid incomplete package installs

### LINKED ISSUE #4897: [hab] Interrupted pkg install can leave package in bad state
If `hab pkg install` is interrupted when unpacking a hartifact, the result can be a situation in which `hab` will report the package as installed despite the installation being unusable.  Reproducing this by hand requires sufficient luck:

```
# hab pkg install core/erlang
» Installing core/erlang from channel 'stable'
☛ Verifying core/erlang/20.0/20171014214153
→ Using core/acl/2.2.52/20170513213108
→ Using core/attr/2.4.47/20170513213059
→ Using core/cacerts/2017.09.20/20171014212239
→ Using core/gcc-libs/5.2.0/20170513212920
→ Using core/glibc/2.22/20170513201042
→ Using core/linux-headers/4.3/20170513200956
→ Using core/ncurses/6.0/20170513213009
→ Using core/openssl/1.0.2l/20171014213633
→ Using core/sed/4.2.2/20170513213123
→ Using core/zlib/1.2.8/20170513201911
^C
root@automate-deployment:/opt/automate-deployment# hab pkg exec core/erlang erl
✗✗✗
✗✗✗ `erl' was not found on the filesystem or in PATH
✗✗✗
```

Attempts to reinstall the package succeed:

```
# hab pkg install core/erlang
» Installing core/erlang from channel 'stable'
→ Using core/erlang/20.0/20171014214153
★ Install of core/erlang/20.0/20171014214153 complete with 0 new packages installed.
# hab pkg exec core/erlang erl
✗✗✗
✗✗✗ `erl' was not found on the filesystem or in PATH
✗✗✗
```

The user must remove the package contents from `/hab` and reinstall:

```
# rm -rf /hab/pkgs/core/erlang/
# hab pkg install core/erlang
» Installing core/erlang from channel 'stable'
☛ Verifying core/erlang/20.0/20171014214153
→ Using core/acl/2.2.52/20170513213108
→ Using core/attr/2.4.47/20170513213059
→ Using core/cacerts/2017.09.20/20171014212239
→ Using core/gcc-libs/5.2.0/20170513212920
→ Using core/glibc/2.22/20170513201042
→ Using core/linux-headers/4.3/20170513200956
→ Using core/ncurses/6.0/20170513213009
→ Using core/openssl/1.0.2l/20171014213633
→ Using core/sed/4.2.2/20170513213123
→ Using core/zlib/1.2.8/20170513201911
✓ Installed core/erlang/20.0/20171014214153
★ Install of core/erlang/20.0/20171014214153 complete with 1 new packages installed.
# hab pkg exec core/erlang erl
Erlang/OTP 20 [erts-9.0] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V9.0  (abort with ^G)
1>
User switch command
 --> q
```

My assumption is that the root cause is that the unpack function (https://github.com/habitat-sh/core/blob/master/components/core/src/package/archive.rs#L264) is unpacking directly to the final location and the other code in hab to detect if a package is installed doesn't do any verification of the contents.

One solution may be to unpack the archive into a temporary directory and then atomically rename() it once the unpack is complete.

stevendanna:
Resolves #4897 

Previously, `hab pkg install` would unpack a package directly into the
target directory. If `hab pkg install` was killed in the middle of
the installation, the system might be left in state where `hab` would
report the package as installed, even though it was unusable.

This changes the unpack process so that we first unpack into a
temporary directory and then rename the directory into place.

The trade-off here is that a killed install will leave behind a
temporary directory that would take up space on the user's disk.

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -49,14 +49,17 @@ use hcore;
 use hcore::crypto::keys::parse_name_with_rev;
 use hcore::crypto::{artifact, SigKeyPair};
 use hcore::fs::cache_key_path;
+use hcore::fs::pkg_install_path;
 use hcore::package::metadata::PackageType;
 use hcore::package::{Identifiable, PackageArchive, PackageIdent, PackageInstall, PackageTarget};
+use hcore::package::install::INSTALL_TMP_PREFIX;
 use hyper::status::StatusCode;
 
 use error::{Error, Result};
 use ui::{Status, UIWriter};
 
 use retry::retry;
+use tempdir::TempDir;
 
 pub const RETRIES: u64 = 5;
 pub const RETRY_WAIT: u64 = 3000;
@@ -702,9 +705,34 @@ impl<'a> InstallTask<'a> {
     where
         T: UIWriter,
     {
-        artifact.unpack(Some(self.fs_root_path))?;
-        ui.status(Status::Installed, artifact.ident()?)?;
+        let ident = &artifact.ident()?;
+        let real_install_path = &pkg_install_path(ident, Some(self.fs_root_path));
+        let real_install_base = real_install_path.parent().ok_or(
+            Error::PackageUnpackFailed(
+                "Could not determine parent directory for package unpack location".to_owned()
+            )
+        )?;
+        let temp_dir = self.temporary_unpack_directory(real_install_base, real_install_path)?;
+        let temp_install_path = &pkg_install_path(ident, Some(temp_dir.path()));
+        artifact.unpack(Some(temp_dir.path()))?;
+        fs::rename(temp_install_path, real_install_path)?;
+        #[cfg(unix)]
+        fs::File::open(real_install_base).and_then(|f| f.sync_all())?;
+        ui.status(Status::Installed, ident)?;
         Ok(())
+    }
+
+    fn temporary_unpack_directory(&self, base: &Path, full_path: &Path) -> Result<TempDir> {
+        fs::create_dir_all(base)?;
+        let temp_install_prefix = full_path.file_name()
+            .and_then(|f| f.to_str() )
+            .and_then(|dirname| Some(format!("{}-{}", INSTALL_TMP_PREFIX, dirname)))
+            .ok_or(
+                Error::PackageUnpackFailed(
+                    "Could not generate prefix for temporary unpack directory".to_owned()
+                )
+            )?;
+        Ok(TempDir::new_in(base, &temp_install_prefix)?)
     }
 
     /// Adapter function to retrieve an installed package given an
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -748,7 +748,7 @@ dependencies = [
 [[package]]
 name = "habitat_core"
 version = "0.0.0"
-source = "git+https://github.com/habitat-sh/core.git#a6f0fb04e692d82a2e1c84d596bec09bc3410518"
+source = "git+https://github.com/habitat-sh/core.git#91238adaec5713db8838151bc8be002f8d68dd3c"
 dependencies = [
  "ansi_term 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -804,7 +804,7 @@ dependencies = [
 [[package]]
 name = "habitat_http_client"
 version = "0.0.0"
-source = "git+https://github.com/habitat-sh/core.git#a6f0fb04e692d82a2e1c84d596bec09bc3410518"
+source = "git+https://github.com/habitat-sh/core.git#91238adaec5713db8838151bc8be002f8d68dd3c"
 dependencies = [
  "base64 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "habitat_core 0.0.0 (git+https://github.com/habitat-sh/core.git)",
@@ -963,7 +963,7 @@ dependencies = [
 [[package]]
 name = "habitat_win_users"
 version = "0.0.0"
-source = "git+https://github.com/habitat-sh/core.git#a6f0fb04e692d82a2e1c84d596bec09bc3410518"
+source = "git+https://github.com/habitat-sh/core.git#91238adaec5713db8838151bc8be002f8d68dd3c"
 dependencies = [
  "gcc 0.3.54 (registry+https://github.com/rust-lang/crates.io-index)",
  "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/components/common/Cargo.toml
+++ b/components/common/Cargo.toml
@@ -14,6 +14,7 @@ libc = "*"
 log = "*"
 pbr = "*"
 regex = "*"
+tempdir = "*"
 retry = "*"
 term = "*"
 time = "*"
@@ -29,9 +30,6 @@ path = "../builder-depot-client"
 [target.'cfg(windows)'.dependencies]
 kernel32-sys = "*"
 winapi = "0.2"
-
-[dev-dependencies]
-tempdir = "*"
 
 [features]
 default = []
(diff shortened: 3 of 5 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
@stevendanna Thanks for digging into this! I do think this is a problem worth worrying about; it probably doesn't happen very often, but if it does, it would be a bit of a head-scratcher unless you were very familiar with what's going on during an install.

As for coordination with `habitat-sh/core` crate updates, we generally merge those first, then merge the changes to the `habitat-sh/habitat` repo (along with a `cargo update` to ensure it's pointing at the newly-merged core changes).

Sadly, it doesn't look like we've got tests yet for `archive.unpack()`. There are some other tests in https://github.com/habitat-sh/core/blob/master/components/core/src/package/archive.rs as well as https://github.com/habitat-sh/core/blob/master/components/core/src/package/install.rs that you might take inspiration from.

In general, this approach looks good... we might want to think about some kind of "temporary installation cleanup" command in the future to deal with the potential for orphaned directories, but given the likely rarity of that situation, I don't think that would need to be included in this PR.

components/common/src/command/package/install.rs
```diff
@@ -702,11 +704,38 @@ impl<'a> InstallTask<'a> {
...
+        f.sync_all()?;
+        Ok(())
+    }
+
+    fn temporary_unpack_directory(&self, base: &Path, full_path: &Path) -> Result<TempDir> {
+        fs::create_dir_all(base)?;
+        let temp_install_prefix = full_path.file_name()
+            .and_then(|f| f.to_str() )
+            .and_then(|dirname| Some(format!(".hab-pkg-intall-{}", dirname)))
+            .ok_or_else(|| Error::PackageUnpackFailed("Could not generate prefix for temporary unpack directory".to_owned()))?;
```
baumanj: You can make this a bit simpler/shorter with `ok_or`:
```rust
            .ok_or(Error::PackageUnpackFailed("Could not generate prefix for temporary unpack directory".to_owned()))?;
```

components/common/src/command/package/install.rs
```diff
@@ -702,11 +704,38 @@ impl<'a> InstallTask<'a> {
     where
         T: UIWriter,
     {
-        artifact.unpack(Some(self.fs_root_path))?;
-        ui.status(Status::Installed, artifact.ident()?)?;
+        let ident = &artifact.ident()?;
+        let real_install_path = &pkg_install_path(ident, Some(self.fs_root_path));
+        let real_install_base = Path::new(real_install_path)
+            .parent()
+            .ok_or_else(|| Error::PackageUnpackFailed("Could not determine parent directory for package unpack location".to_owned()))?;
```
baumanj: This can be simplified a bit with `ok_or`:
```rust
        let real_install_base = Path::new(real_install_path).parent().ok_or(
            Error::PackageUnpackFailed(
                "Could not determine parent directory for package unpack location".to_owned(),
            ),
        )?;
```

components/common/src/command/package/install.rs
```diff
@@ -702,11 +704,38 @@ impl<'a> InstallTask<'a> {
...
+        fs::rename(temp_install_path, real_install_path)?;
+        self.fsync_dir(real_install_base)?;
+        ui.status(Status::Installed, ident)?;
         Ok(())
     }
 
+
+    fn fsync_dir(&self, dir: &Path) -> Result<()> {
+        // TODO(ssd) 2018-07-23: This works on Linux but I'm not sure
+        // how portable it is.
```
baumanj: I think you can remove this comment. There is [a Windows implementation](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/fs.rs#L277). In general, the rust documentation calls out if something is platform-specific.

components/common/src/command/package/install.rs
```diff
@@ -702,11 +704,38 @@ impl<'a> InstallTask<'a> {
...
+        let ident = &artifact.ident()?;
+        let real_install_path = &pkg_install_path(ident, Some(self.fs_root_path));
+        let real_install_base = Path::new(real_install_path)
+            .parent()
+            .ok_or_else(|| Error::PackageUnpackFailed("Could not determine parent directory for package unpack location".to_owned()))?;
+        let temp_dir = self.temporary_unpack_directory(real_install_base, real_install_path)?;
+        let temp_install_path = &pkg_install_path(ident, Some(temp_dir.path()));
+        artifact.unpack(Some(temp_dir.path()))?;
+        fs::rename(temp_install_path, real_install_path)?;
+        self.fsync_dir(real_install_base)?;
```
baumanj: Since the comment isn't necessary, in my opinion, I think you can remove this new function and just have this inline:
```rust
fs::File::open(real_install_base).and_then(|f| f.sync_all())?;
```

stevendanna:
Related core change: https://github.com/habitat-sh/core/pull/44

baumanj:
Thanks for the updates, @stevendanna! I think once [the related core change](https://github.com/habitat-sh/core/pull/44) merges and we run a `cargo update`, this can merge as well.

stevendanna:
Sounds good.  I've updated the core PR to use a constant and made use of that constant here, so the CI tests will fail until I update the Cargo.lock most likely, so it'll be easy to avoid forgetting :D

christophermaier:
@stevendanna I just merged your core PR; feel free to bump the dependency here, and then we'll get this merged!

Thank you very much for this; it's a nice feature that'll make things safer overall for everyone 👍

stevendanna:
Is anyone familiar with the appveyor tests able to point me in the right direction there?  I attempted to push a commit that set RUST_LOG=debug to the appveyor script, but that produced so much output that it kinda killed appveyor and I couldn't really find the error.

If I'm reading this powerhsell correctly, it looks like the package installation that is failing is happening with the /old/ hab version so I'm not sure how my changes would have caused it.

baumanj:
@mwrock is the expert in all things windows, so perhaps he can weigh in on the AppVeyor failure. My experience has always been that AppVeyor (or something about our setup) is flakey and when I've restarted tests that timed out they subsequently passed.

Is your original error a timeout or something more tractable, @stevendanna ?

stevendanna:
@baumanj If it was a timeout, I would be a bit less worried; unfortunately, it is a permission denied error:

```
✗✗✗
✗✗✗ Access is denied. (os error 5)
✗✗✗
```

which could plausibly be the result of my change.  Now, based on my reading of the powershell driving those tests, I don't think that is coming from a version of hab that shouldn't have my code in it; however, I could be reading the script incorrectly.

mwrock:
Taking a look now. After hab and the studio are build, the `hab.exe` originally obtained from bintray are swaped with the one being built. This failure seems to occur on the first `pkg install` after that swap so it seems to be directly impacted. I'll build locally and install a package to see if I can repro and pinpoint the problem.

components/common/src/command/package/install.rs
```diff
@@ -702,11 +705,35 @@ impl<'a> InstallTask<'a> {
...
+        let real_install_base = real_install_path.parent().ok_or(
+            Error::PackageUnpackFailed(
+                "Could not determine parent directory for package unpack location".to_owned()
+            )
+        )?;
+        let temp_dir = self.temporary_unpack_directory(real_install_base, real_install_path)?;
+        let temp_install_path = &pkg_install_path(ident, Some(temp_dir.path()));
+        artifact.unpack(Some(temp_dir.path()))?;
+        fs::rename(temp_install_path, real_install_path)?;
+        fs::File::open(real_install_base).and_then(|f| f.sync_all())?;
```
mwrock: Looks like this is the culprit and I think it is unnecessary. The error is likely occuring because `real_install_base` is a directory. I don't think a `sync` is necessary after a rename. Only after a `write`.

stevendanna: I think Sync is necessary on some unix file systems to ensure the rename operation is commited to disk.

PR #5375, opened 2018-07-24
Enable feature flags with any value except the empty string

### LINKED ISSUE #5275: Have feature flags enabled by presence of an environment variable
To enable a feature flag in `hab` right now, you must set the appropriate environment variable (say, `HAB_FEAT_OFFLINE_INSTALL` for the offline install feature) to either the string`"true"` or `"TRUE"`. This is too restrictive, in that we shouldn't really care what the value is, merely that it exists.

Alternatively, we could recognize values of `""`, `0`, `false`, `False`, `false`, etc., as signals to _not_ enable the flag, whereas any other value would be a signal to enable the flag.

(Be sure to check for instances of us using `HAB_FEAT_*` flags in e.g., release process code, and ensure that we simplify as appropriate.)

christophermaier:
Previously, a feature flag could only be enabled if its corresponding
environment variable was set to the literal string `"true"` or
`"TRUE"`.

In general, this is too restrictive, since you really only need the
existence of the variable to toggle the feature.

Now, if the environment variable is set to anything except the empty
string, it will be enabled.

That is,

``` sh
HAB_FEAT_IGNORE_LOCAL=yup hab pkg install core/redis --ignore-local
HAB_FEAT_IGNORE_LOCAL=true hab pkg install core/redis --ignore-local
HAB_FEAT_IGNORE_LOCAL=false hab pkg install core/redis --ignore-local
HAB_FEAT_IGNORE_LOCAL=1 hab pkg install core/redis --ignore-local
HAB_FEAT_IGNORE_LOCAL=0 hab pkg install core/redis --ignore-local
```

will all activate the `IGNORE_LOCAL` option for `hab pkg install`, but

``` sh
HAB_FEAT_IGNORE_LOCAL="" hab pkg install core/redis --ignore-local
HAB_FEAT_IGNORE_LOCAL= hab pkg install core/redis --ignore-local
```

will all fail, because the feature will _not_ be enabled.

This is consistent with our use of environment variables elsewhere, in
that the empty string is not considered to be a valid value for an
environment variable.

Fixes #5275

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -1297,9 +1297,11 @@ fn enable_features_from_env(ui: &mut UI) {
         (feat::IgnoreLocal, "IGNORE_LOCAL"),
     ];
 
+    // If the environment variable for a flag is set to _anything_ but
+    // the empty string, it is activated.
     for feature in &features {
         match henv::var(format!("HAB_FEAT_{}", feature.1)) {
-            Ok(ref val) if ["true", "TRUE"].contains(&val.as_str()) => {
+            Ok(_) => {
                 feat::enable(feature.0);
                 ui.warn(&format!("Enabling feature: {:?}", feature.0))
                     .unwrap();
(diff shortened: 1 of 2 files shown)


christophermaier:
@baumanj Nice catch; I'll get that fixed up.

christophermaier:
@baumanj Force-pushed a fix

PR #5383, opened 2018-07-25
Print the values of the feature environment variables rather than "true"

baumanj:
This can be useful for features whose env vars take on a meaningful value.
Furthermore, it's already printed which features are enabled.


thesentinels:

Thank you for contributing!

PR #5384, opened 2018-07-25
 Allow a crashed supervisor to be restarted by the launcher

baumanj:
The first commit just adds testability, so you can confirm what was broken and that the second commit fixes it.

There's some additional follow-up work to do, as I don't think we should endlessly loop for failures to start such as:
```
hab-sup(MR)[components/sup/src/manager/mod.rs:533:24]: Unable to read or write to data directory, /hab/sup/default/data, Permission denied (os error 13)
```
but that can come separately.

This PR is best viewed [with whitespace changes hidden](https://github.com/habitat-sh/habitat/pull/5384/files?w=1).

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -48,7 +48,7 @@ use hcore::crypto::dpapi::encrypt;
 use hcore::crypto::{self, default_cache_key_path, SymKey};
 use hcore::env as henv;
 use hcore::url::{bldr_url_from_env, default_bldr_url};
-use launcher_client::{LauncherCli, ERR_NO_RETRY_EXCODE, OK_NO_RETRY_EXCODE};
+use launcher_client::{LauncherCli, ERR_NO_RETRY_EXCODE};
 use protocol::{
     ctl::ServiceBindList,
     types::{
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -690,6 +691,22 @@ impl Manager {
...
         loop {
+            if feat::is_enabled(feat::TestExit) {
+                if let Ok(exit_file_path) = env::var("HAB_FEAT_TEST_EXIT") {
+                    if let Ok(mut exit_code_file) = File::open(&exit_file_path) {
+                        let mut buffer = String::new();
+                        exit_code_file
+                            .read_to_string(&mut buffer)
+                            .expect("couldn't read");
+                        if let Ok(exit_code) = buffer.lines().next().unwrap_or("").parse::<i32>() {
+                            fs::remove_file(&exit_file_path).expect("foo");
```
christophermaier: We could be a little more detailed than simply "foo" 😄

test/end-to-end/test_launcher_restarts_supervisor.sh
```diff
@@ -0,0 +1,63 @@
...
+
+TESTING_FS_ROOT=$(mktemp -d)
+export TESTING_FS_ROOT
+sup_log="$TESTING_FS_ROOT/hab/sup/default/sup.log"
+exit_file=$(mktemp)
+exit_code=${1:-1}
+
+mkdir -p "$(dirname "$sup_log")"
+echo -n "Starting launcher with root $TESTING_FS_ROOT (logging to $sup_log)..."
+env HAB_FEAT_TEST_EXIT="$exit_file" hab sup run &> "$sup_log" &
```
christophermaier: This will be using whatever hab, supervisor, and launcher packages happen to be installed on the system. Can we take advantage of the `HAB_LAUNCH_BINARY` and `HAB_SUP_BINARY` environment variables to control what gets run?

test/end-to-end/test_launcher_restarts_supervisor.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+set -eou pipefail
+
+wait_for_sup_to_start() {
+	until pgrep hab-sup &>/dev/null; do
```
christophermaier: It'd be good to have some kind of timeout here, in case something goes wrong.

test/end-to-end/test_launcher_restarts_supervisor.sh
```diff
@@ -0,0 +1,63 @@
...
+	echo "Failure! Supervisor process did not change"
+	exit 3
+elif [[ $launcher_pid != "$new_launcher_pid" ]]; then
+	echo "Failure! Launcher process changed unexpectedly"
+	exit 4
+else
+	echo "Success! Launcher restarted supervisor process"
+fi
+
+hab sup term
```
christophermaier: Ensuring this gets run in some kind of cleanup function, triggered by a `trap`, would ensure we don't inadvertently leave any stray processes around.

components/sup/src/main.rs
```diff
@@ -233,44 +233,41 @@ fn sub_run(m: &ArgMatches, launcher: LauncherCli) -> Result<()> {
     set_supervisor_logging_options(m);
 
     let cfg = mgrcfg_from_matches(m)?;
-    if Manager::is_running(&cfg)? {
-        process::exit(OK_NO_RETRY_EXCODE);
-    } else {
```
christophermaier: To clarify, this is what was ultimately causing the lack of restarting, right? The launcher would try to restart the supervisor, the new Supervisor process would see that there's already a Launcher running, and then shut itself down?

components/sup/src/manager/mod.rs
baumanj: Ha ha, I thought I updated that. Good catch.

test/end-to-end/test_launcher_restarts_supervisor.sh
baumanj: The idea is to pass those environment variables into she test script:
```
➤ env HAB_SUP_BINARY=(pwd)/target/debug/hab-sup HAB_LAUNCH_BINARY=(pwd)/target/debug/hab-launch ./test/end-to-end/test_launcher_restarts_supervisor.sh 
```
That makes it easy to go back and forth between the behavior of the installed system and the local changes.

christophermaier: Fair point; can we document that intention in the script?

test/end-to-end/test_launcher_restarts_supervisor.sh
baumanj: Since we can't actually run this as an automated test, I opted for the simpler route. I'm planning to get all the end-to-end test scripts running on Buildkite since they're not simple to get running in travis. Until that time, I'm trying to keep the scripts as simple as possible since it may make sense to [add logic like timeouts at a higher level](https://linux.die.net/man/1/timeout) rather than duplicating a bunch of logic in the test.

test/end-to-end/test_launcher_restarts_supervisor.sh
baumanj: That's a good call, will fix.

components/sup/src/main.rs
baumanj: Yup

test/end-to-end/test_launcher_restarts_supervisor.sh
christophermaier: Cool; add that as a comment and :shipit:

components/sup/src/main.rs
christophermaier: Nice sleuthing on that one 👏

baumanj:
Changes made

Issue #3060, opened 2017-08-31
Launcher should be more resilient to supervisor crashes


chefsalim:
Currently, if the supervisor crashes on start-up for some reason, the launcher can go into a tight loop restarting the supervisor.  We should have a better throttling, back-off and eventing strategy in the launcher for this scenario, as this scenario can (has) occur on acceptance, and cause a production incident.

PR #5387, opened 2018-07-26
If the supervisor terminates with an error result, don't restart it

baumanj:
Resolves https://github.com/habitat-sh/habitat/issues/3060

Exit ERR_NO_RETRY_EXCODE to communicate to the launcher that it shouldn't be
restarted. Otherwise errors on startup can lead to a tight loop of perpetual
restarts.

Abrupt exits due to panics and graceful supervisor exit for update will still
exit 0 and be restarted by the launcher.

Before:
```
➤ ./test/end-to-end/test_launcher_exits_on_supervisor_startup_failure.sh 
Setting /tmp/tmp.SGOnFKb6FC to read-only to induce supervisor start failure
Starting launcher with root /tmp/tmp.SGOnFKb6FC (logging to /tmp/tmp.n7NidR5L14)..........
Failure! Launcher failed to exit before timeout
```
After:
```
➤ env HAB_SUP_BINARY=(pwd)/target/debug/hab-sup ./test/end-to-end/test_launcher_exits_on_supervisor_startup_failure.sh 
Setting /tmp/tmp.Ci34qa8zrd to read-only to induce supervisor start failure
Starting launcher with root /tmp/tmp.Ci34qa8zrd (logging to /tmp/tmp.fDjHViBah9)....
Success! Launcher cleanly exited
```

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -74,13 +74,12 @@ static RING_KEY_ENVVAR: &'static str = "HAB_RING_KEY";
 
 fn main() {
     let result = start();
-    let exit_code;
-    match result {
+    let exit_code = match result {
+        Ok(_) => 0,
         Err(ref err) => {
             println!("{}", err);
-            exit_code = ERR_NO_RETRY_EXCODE;
+            ERR_NO_RETRY_EXCODE
         }
-        Ok(_) => exit_code = 0,
     };
     debug!("start() returned {:?}; Exiting {}", result, exit_code);
     process::exit(exit_code);
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -73,24 +73,17 @@ static RING_ENVVAR: &'static str = "HAB_RING";
 static RING_KEY_ENVVAR: &'static str = "HAB_RING_KEY";
 
 fn main() {
-    if let Err(err) = start() {
-        println!("{}", err);
-        match err {
-            SupError {
-                err: Error::ProcessLocked(_),
-                ..
-            }
-            | SupError {
-                err: Error::Departed,
-                ..
-            }
-            | SupError {
-                err: Error::ButterflyError(_),
-                ..
-            } => process::exit(ERR_NO_RETRY_EXCODE),
-            _ => process::exit(1),
+    let result = start();
+    let exit_code;
+    match result {
+        Err(ref err) => {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/main.rs
```diff
@@ -73,24 +73,17 @@ static RING_ENVVAR: &'static str = "HAB_RING";
...
+    let result = start();
+    let exit_code;
+    match result {
+        Err(ref err) => {
+            println!("{}", err);
+            exit_code = ERR_NO_RETRY_EXCODE;
         }
-    }
+        Ok(_) => exit_code = 0,
+    };
```
christophermaier: This whole bit can be written more simply as 
``` rust
let exit_code = match result {
    Ok(_) => 0,
    Err(ref err) => {
        ...
        ERR_NO_RETRY_EXCODE
    }
};
```

baumanj: Ooh, good call!

christophermaier:
@baumanj Looks good; ship it!

baumanj:
With regards to a more rust-idiomatic way of handling the "should the supervisor restart" issue, I think that's an interesting idea. As far as I can tell right now that distinction is encapsulated by the [`Result`](http://doc.rust-lang.org/1.21.0/std/result/enum.Result.html)  enum. We always restart in the `Ok` case and never in the `Err` case. If that needs changing in the future, perhaps we should think more about it.

To me, the trickiest part of the code after this change is that the launcher restarts both on a `0` (success) exit **and** all nonzero (failure) exits _except_ `ERR_NO_RETRY_EXCODE`. This leads to the slightly strange behavior of the launcher exiting `0` even when the supervisor exits due to `Err` such as `Unable to read or write to data directory, /hab/sup/default/data, Permission denied (os error 13)`. I think it may be worth a follow-up change to the launcher to make the behavior more intuitive.

christophermaier:
@baumanj Yeah, having the launcher return non-zero when the Supervisor errors out would be a nice follow-on addition.

baumanj:
🤣 I already fixed that with https://github.com/habitat-sh/habitat/commit/51f661749c004e26466acc552c4137d58c77b1c7

You'd think I'd remember my own fixes.

christophermaier:
Well-played, Past-@baumanj; well-played 😆

PR #5391, opened 2018-07-28
Release from Buildkite

christophermaier:
We're moving our release process from TravisCI over to Buildkite in order to get better control over things. This will make it possible to automate many more things than we were able to before, and should result in quicker, easier, less error-prone releases, which are good for everybody.

tenor-62182516


thesentinels:

Thank you for contributing!

.buildkite/Brewfile
```diff
@@ -0,0 +1,29 @@
...
+brew "hab"
+
+# Our builders currently have Bash 3.2.57 installed, but our Habitat
+# package build process requires a more modern version. The Linux
+# build currently uses Bash 4.3.42, and Homebrew currently brings in
+# 4.4.23.
+#
+# Until our builders have a modern Bash installed, we'll just bring our
+# own.
+brew "bash"
```
scotthain: We can assume that the habitat omnibus package has current bash in it in the future.

christophermaier: tenor-71154006

components/hab/mac/README.md
```diff
@@ -1,26 +1,74 @@
...
+
+This will install the toolchain in `/opt/hab-bundle`; this is where the build program is expecting to find binaries and libraries it needs.
+
+### Install Homebrew
+Follow the instructions at https://brew.sh.
+
+### Install Rust toolchain
+Follow the instructions at https://rustup.rs/
+
+### Install Homebrew prerequisites
```
fnichol: I think that in light of these changes we may want to consider updating our Mac/Development instructions to use similar deps when making development `cargo build`s. That is, use the same Omnibus/Homebrew combinations in development so that we more closely match a production build later on.

components/hab/mac/README.md
```diff
@@ -1,26 +1,74 @@
...
+brew install hab
+```
+
+### Install Builder Origin Keys
+
+To build a Habitat package, you must have the secret `core` origin key for signing that package. Obtaining the secret key requires a personal authentication token from Builder.
+
+``` sh
+sudo hab origin key download core
+sudo hab origin key download --secret --auth="${HAB_AUTH_TOKEN}" core
```
fnichol: While this won't bite us here, it might be more consistent in the future to use `sudo -E` as it will pass currently exported environment variables to the command to be run. I see that form used a bit later in the instructions which makes sense.

components/hab/mac/plan.sh
```diff
@@ -7,10 +7,6 @@ pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_deps=()
 pkg_build_deps=()
 
-nproc() {
-  sysctl -n hw.ncpu
-}
```
fnichol: I had to go back to remember why I added this. Answer: initially we had `cargo -j $(nproc) … build` and I removed this in bf5724191f3d94c70388040a9fb8331bcb6e1bd8#diff-53cf941dbc3096011996a1427420a6e4 for more consistency (I think that Cargo does this by default anyway these days). So, whoops!

support/ci/only_master.sh
```diff
@@ -14,12 +26,12 @@ echo "TRAVIS_PULL_REQUEST_BRANCH: ${TRAVIS_PULL_REQUEST_BRANCH}"
 #       for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG).
 #     TRAVIS_PULL_REQUEST:
 #       The pull request number if the current job is a pull request, “false” if it’s not a pull request.
-
 if [ "${TRAVIS_TAG}" == "$(cat VERSION)" ]; then
-    echo "Building a release for ${TRAVIS_TAG}"
+    echo "Looks like a release tag; not building in Travis. Please check out Buildkite instead."
+    exit 1
```
fnichol: Nice work!

components/hab/mac/README.md
scotthain: This is going to be a work in progress as we tweak the process and automate more bits - but yes you are totally correct.

components/hab/mac/README.md
scotthain: We should also try to reduce the use of `sudo -E` as much as possible, in case that hasn't be stated. I'm a big fan of explicit passing vs "passing the environment through"

.buildkite/pipeline.sh
```diff
@@ -0,0 +1,33 @@
...
+    if [[ "${FAKE_RELEASE_TAG}" ]]; then
+        echo "Using fake release tag '${FAKE_RELEASE_TAG}'"
+        release="${FAKE_RELEASE_TAG}"
+    elif [[ "${BUILDKITE_TAG}" ]]; then
+        echo "Using release tag '${BUILDKITE_TAG}'"
+        release="${BUILDKITE_TAG}"
+    fi
+    channel="rc-${release}"
+    echo "Release channel is '${channel}'"
+    echo -e "## Habitat Release _${release}_" | buildkite-agent annotate --context "release-manifest"
```
baumanj: Do we expect backslash escapes in `$release`? Just wondering why the `-e` flag for `echo` is being used.

.buildkite/pipeline.sh
```diff
@@ -0,0 +1,33 @@
...
+
+    # We'll record the explicit version being built for future access,
+    # even though 'cat'ing a file is pretty easy; we'll be sticking
+    # other things into the Buildkite metadata, so it'll all be uniform.
+    version=$(cat VERSION)
+    buildkite-agent meta-data set "version" "${version}"
+
+    buildkite-agent pipeline upload .buildkite/release_pipeline.yaml
+else
+    echo -e "--- :sparkles: Preparing for a CI run! :sparkles:"
```
baumanj: ditto

.buildkite/scripts/bintray_publish.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source .buildkite/scripts/shared.sh
+
+if [[ "${FAKE_RELEASE_TAG}" ]]; then
```
baumanj: I'd suggest `${FAKE_RELEASE_TAG:-}` if this isn't necessarily set, since `set -u` will fail on the unbound variable otherwise.

.buildkite/scripts/bintray_publish.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "--- :warning: FAKE_RELEASE_TAG was found in the environment! This isn't a \"real\" release!"
+    fake_release=1
+fi
+
+publish(){
+    url=${1}
+    if [ -z "${fake_release}" ]; then
```
baumanj: Similar to above, either make this `${fake_release:-}` or make sure the variable is always defined.

.buildkite/scripts/bintray_upload.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# TODO: bintray user = chef-releng-ops!
+
+channel=$(buildkite-agent meta-data get "release-channel")
+
+# TODO (CM): extract set_hab_binary function to a common library and
+# use it here
+
+echo "--- :habicat: Installing core/hab-bintray-publish from '${channel}' channel"
+sudo hab pkg install \
```
baumanj: Do we want `sudo -E` here as well as it is below?

.buildkite/scripts/bintray_upload_macos.sh
```diff
@@ -0,0 +1,44 @@
...
+# TODO: bintray user = chef-releng-ops!
+
+# We need to upload (but not publish) artifacts to Bintray right now.
+channel=$(buildkite-agent meta-data get "release-channel")
+
+# TODO (CM): extract set_hab_binary function to a common library and
+# use it here
+
+echo "--- :habicat: Installing core/hab-bintray-publish from '${channel}' channel"
+sudo hab pkg install \
```
baumanj: `-E`?

.buildkite/scripts/build_component.sh
```diff
@@ -0,0 +1,105 @@
...
+# The binlink dir is set by releng, but seems to be messing things up
+# for us in the studio.
+unset HAB_BINLINK_DIR
+export HAB_ORIGIN=core
+
+# Eww
+#
+# CI_OVERRIDE_CHANNEL is basically used to tell the studio which
+# hab/backline to grab
+if [[ "${new_studio}" == "true" ]]; then
```
baumanj: You could just
```bash
if $new_studio; then
```
`true` and `false` are POSIX commands. No need to even quote `$new_studio` since it's always set to a constant string.

.buildkite/scripts/build_component.sh
```diff
@@ -0,0 +1,105 @@
...
+        # buildkite-agent artifact upload "results/${pkg_artifact}"
+        buildkite-agent meta-data set "studio-version" "${pkg_ident}"
+        ;;
+    *)
+        ;;
+esac
+
+# TODO (CM): Replace "Linux" below with ${pkg_target:?} once that's in
+# hab-plan-build (see https://github.com/habitat-sh/habitat/pull/5373)
+echo -e "<br>* ${pkg_ident} (Linux)" | buildkite-agent annotate --append --context "release-manifest"
```
baumanj: `-e`?

.buildkite/scripts/build_mac_release.sh
```diff
@@ -0,0 +1,76 @@
...
+# need, because we're not pulling anything from Builder. Once we do,
+# we'll need to make sure we pull from the right channels.
+sudo -E "$(brew --prefix bash)/bin/bash" components/plan-build/bin/hab-plan-build.sh components/hab/mac
+source results/last_build.env
+
+echo "--- :buildkite: Annotating build"
+
+# TODO (CM): Replace "MacOS" below with ${pkg_target:?} once that's in
+# hab-plan-build (see https://github.com/habitat-sh/habitat/pull/5373)
+echo -e "<br>* ${pkg_ident:?} (MacOS)" | buildkite-agent annotate --append --context "release-manifest"
```
baumanj: Same question as before

.buildkite/scripts/build_mac_release.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Since we can't store MacOS packages in Builder yet, we'll store it
+# in Buildkite until we grab it later for upload to Bintray
+echo "--- :buildkite: Storing MacOS 'hab' artifact ${pkg_artifact:?}"
+buildkite-agent meta-data set "hab-artifact-macos" "${pkg_artifact:?}"
+buildkite-agent meta-data set "hab-release-macos" "${pkg_release:?}"
+(
+    cd results
+    buildkite-agent artifact upload "${pkg_artifact}"
+)
```
baumanj: Why execute in a subshell when there's nothing following this?

.buildkite/scripts/promote_launchers.sh
```diff
@@ -0,0 +1,34 @@
...
+resolve_launcher() {
+    local metadata_key="${1}"
+    local target="${2}"
+
+    if buildkite-agent meta-data exists "${metadata_key}"; then
+        launcher_ident=$(buildkite-agent meta-data get "${metadata_key}")
+    else
+        launcher_ident=$(latest_from_builder "${target}" stable hab-launcher)
+    fi
+    echo "${launcher_ident}"
```
baumanj: There's no need for the command substitution and `echo`. This could be simplified to:
```bash
    if buildkite-agent meta-data exists "${metadata_key}"; then
        buildkite-agent meta-data get "${metadata_key}"
    else
        latest_from_builder "${target}" stable hab-launcher
    fi
```

.buildkite/scripts/shared.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
```
baumanj: No `set -euo pipefail`?

.buildkite/pipeline.sh
christophermaier: @baumanj The `-e` is most likely leftovers from earlier attempts at the pipeline... I can likely remove all of them now.

.buildkite/scripts/bintray_publish.sh
christophermaier: Yup, fixed!

.buildkite/scripts/build_mac_release.sh
christophermaier: Just in case we add something afterwards... the artifact upload incorporates the path into the "key" under which the artifact is stored, so I need to change directories in order to not have `results/` tacked onto the front of it. This just makes sure we're back in the "right" directory if we add something following this.

.buildkite/scripts/shared.sh
christophermaier: This basically a library script, and doesn't actually run anything

.buildkite/scripts/promote_launchers.sh
christophermaier: Ah, nice catch... will change!

components/hab/mac/README.md
```diff
@@ -1,26 +1,74 @@
-# Building a hab Mac Binary
+# Building a `hab` MacOS Binary
```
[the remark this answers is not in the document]
christophermaier: Ah, good point! Will fix.

components/hab/mac/README.md
christophermaier: After thinking about it a bit, I think I'm also in favor of reducing `sudo -E` in favor of being super-explicit. Any `sudo -E`s that I added are probably bits of early "just make it work" code that I didn't fully clean up.

PR #5401, opened 2018-07-31
Add rootless studio to release

elliott-davis:

### DIFF
--- a/.buildkite/release_pipeline.yaml
+++ b/.buildkite/release_pipeline.yaml
@@ -158,6 +158,10 @@ steps:
 
   - wait
 
+  - label: ":docker: Upload (but don't promote!) containers to Docker Hub"
+    command: .buildkite/scripts/dockerhub_upload.sh
+    agents:
+      queue: habitat-release
   # TODO Create a JFrog emoji
   - label: ":linux: Upload (but don't publish!) artifacts to Bintray"
     command: .buildkite/scripts/bintray_upload.sh
--- a/.buildkite/scripts/dockerhub_publish.sh
+++ b/.buildkite/scripts/dockerhub_publish.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source .buildkite/scripts/shared.sh
+version=$(buildkite-agent meta-data get "version")
+
+if [[ "${FAKE_RELEASE_TAG}" ]]; then
+  # This overrides IMAGE_NAME in docker-base.sh
+  export IMAGE_NAME="habitat/fakey-mc-fake-face-studio"
+fi
+
+pushd ./components/rootless_studio >/dev/null
+./tag-docker-image.sh "${version}" "latest"
+./publish-to-dockerhub.sh "${version}" "latest"
+popd >/dev/null
--- a/.buildkite/scripts/dockerhub_upload.sh
+++ b/.buildkite/scripts/dockerhub_upload.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source .buildkite/scripts/shared.sh
+version=$(buildkite-agent meta-data get "version")
+
+if [[ "${FAKE_RELEASE_TAG}" ]]; then
+  # This overrides IMAGE_NAME in docker-base.sh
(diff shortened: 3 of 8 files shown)


thesentinels:

Thank you for contributing!

.buildkite/release_pipeline.yaml
```diff
@@ -158,6 +158,10 @@ steps:
 
   - wait
 
+  - label: ":docker: Upload Containers to Docker Hub"
+    command: .buildkite/scripts/dockerhub_upload.sh
+    agents:
+      queue: habitat-release
```
christophermaier: In the course of a normal release (not a "fake" one for testing), will the fact that this image is pushed out to DockerHub with its "real" tags cause any issues? I'm thinking of what happens if we push something out, then discover that it's buggy, but by then, someone out in the real world has already consumed it.

Could we have a "dev" repository to push the image to initially, then once we've approved the release and promoted Habitat packages to the stable channel, and published releases in Bintray, we could re-upload the image to a "real" repository? This would essentially be like creating an "unstable channel" in DockerHub.

(For the record, I think we should also do something like this for our images in Bintray)

.buildkite/scripts/dockerhub_upload.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source .buildkite/scripts/shared.sh
+version=$(buildkite-agent meta-data get "version")
+
+if [[ "${FAKE_RELEASE_TAG}" ]]; then
+  export IMAGE_NAME="habitat/fakey-mc-fake-face-studio"
```
christophermaier: Why not "Habby McHabFace"?

WHY NOT @elliott-davis ??

.buildkite/release_pipeline.yaml
elliott-davis: We currently have a release tag of something like `0.60.0-rc` right? If so, I'd say we could use that as the tag for rollback and latest should always point at stable. For example:

```
habitat/default-studio:latest -> habitat/default-studio:0.59.0
habitat/default-studio:0.60.0-rc
habitat/default-studio:0.60.0-dev
habitat/default-studio:0.59.0
```

.buildkite/scripts/dockerhub_upload.sh
elliott-davis: I am made of fail

.buildkite/release_pipeline.yaml
christophermaier: I don't think we've currently got anything like that for Docker images, though that'd also probably work. I don't have super strong feelings exactly which way we separate things, just that we _do_ separate things.

elliott-davis: The nice thing about doing it with tags over a repo is you won't have to rebuild the image which could potentially introduce drift

christophermaier: I don't think you need to rebuild the image, just re-tag it and push.

elliott-davis:
@christophermaier this should be good to go if you have a moment to give it some review

.buildkite/scripts/dockerhub_publish.sh
```diff
@@ -0,0 +1,16 @@
...
+
+source .buildkite/scripts/shared.sh
+version=$(buildkite-agent meta-data get "version")
+
+if [[ "${FAKE_RELEASE_TAG}" ]]; then
+  export IMAGE_NAME="habitat/fakey-mc-fake-face-studio"
+fi
+
+pushd ./components/rootless_studio >/dev/null
+docker pull "${IMAGE_NAME}:${version}"
```
christophermaier: Where does `IMAGE_NAME` come from if we're doing a real release?

.buildkite/scripts/dockerhub_upload.sh
christophermaier: Same question here, re: `IMAGE_NAME`

.buildkite/scripts/dockerhub_publish.sh
elliott-davis: https://github.com/habitat-sh/habitat/pull/5401/files#diff-0b94ebc237e784dd28240fc35ae17e26R4

PR #5404, opened 2018-08-01
Launcher simplify reload logic

baumanj:
This keeps the logic simpler and obviates the need to store the CLI args within the `Server` struct.

Also, add a environment variable `HAB_LAUNCH_SUP_RESTART_SLEEP_SECS` to govern
how long the supervisor waits before restarting the supervisor.

Tests confirm the behavior is still good:

```
➤ env RUST_LOG= HAB_SUP_BINARY=(pwd)/target/debug/hab-sup HAB_LAUNCH_BINARY=(pwd)/target/debug/hab-launch ./test/end-to-end/test_launcher_restarts_supervisor.sh
Starting launcher with root /tmp/tmp.t7NgS8xK8n (logging to /tmp/tmp.t7NgS8xK8n/hab/sup/default/sup.log)....
Launcher is process 5319, supervisor is process 5322
Forcing supervisor to exit /tmp/tmp.rmLDbtoijt...
Waiting for old supervisor process to exit....
Waiting for new supervisor process to start...
.....
Success! Launcher restarted supervisor process
```

```
➤ env RUST_LOG= TESTING_FS_ROOT=/tmp HAB_SUP_BINARY=(pwd)/target/debug/hab-sup HAB_LAUNCH_BINARY=(pwd)/target/debug/hab-launch ./test/end-to-end/test_socket_file_cleanup.sh
Starting launcher (logging to /tmp/tmp.FxsTH9FrTB)....
Waiting for launcher to exit...
Checking for socket files left behind...
Success! No socket file left behind
```

### DIFF
--- a/components/launcher/src/server/mod.rs
+++ b/components/launcher/src/server/mod.rs
@@ -43,7 +43,10 @@ use service::Service;
 use {SUP_CMD, SUP_PACKAGE_IDENT};
 
 const IPC_CONNECT_TIMEOUT_SECS: &'static str = "HAB_LAUNCH_SUP_CONNECT_TIMEOUT_SECS";
+const SUP_RESTART_SLEEP_SECS: &'static str = "HAB_LAUNCH_SUP_RESTART_SLEEP_SECS";
 const DEFAULT_IPC_CONNECT_TIMEOUT_SECS: u64 = 5;
+const DEFAULT_SUP_RESTART_SLEEP_SECS: u64 = 5;
+
 const SUP_CMD_ENVVAR: &'static str = "HAB_SUP_BINARY";
 static LOGKEY: &'static str = "SV";
 
@@ -61,25 +64,29 @@ pub struct Server {
     rx: Receiver,
     pipe: String,
     supervisor: Child,
-    args: Vec<String>,
 }
 
 impl Drop for Server {
     fn drop(&mut self) {
-        self.remove_pipe();
+        if fs::remove_file(&self.pipe).is_err() {
+            error!("Could not remove old pipe to supervisor {}", self.pipe);
+        } else {
+            debug!("Removed old pipe to supervisor {}", self.pipe);
+        }
     }
 }
 
 impl Server {
-    pub fn new(args: Vec<String>) -> Result<Self> {
-        let ((rx, tx), supervisor, pipe) = Self::init(&args, false)?;
+    // TODO: After https://github.com/habitat-sh/habitat/issues/5382, the
+    // `restart` argument can be removed from here and called functions
+    pub fn new(args: &Vec<String>, restart: bool) -> Result<Self> {
+        let ((rx, tx), supervisor, pipe) = Self::init(args, restart)?;
         Ok(Server {
             services: ServiceTable::default(),
             tx: tx,
             rx: rx,
             pipe: pipe,
             supervisor: supervisor,
-            args: args,
         })
     }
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/launcher/src/server/mod.rs
```diff
@@ -95,29 +102,6 @@ impl Server {
...
-            error!("Could not remove old pipe to supervisor {}", self.pipe);
-        } else {
-            debug!("Removed old pipe to supervisor {}", self.pipe);
-        }
-    }
-
-    #[allow(unused_must_use)]
-    fn reload(&mut self) -> Result<()> {
-        self.supervisor.kill();
-        self.supervisor.wait();
```
christophermaier: How is this handled in the new code... I'm missing it.

components/launcher/src/server/mod.rs
```diff
@@ -403,18 +387,31 @@ where
...
+                    debug!("tick() returned error {}. Attempting to restart server", e);
+                    break;
+                }
             }
-            Err(_) => while server.reload().is_err() {
-                thread::sleep(Duration::from_millis(1_000));
-            },
         }
+        error!(
+            "Supervised exited; will restart in {} secs...",
```
christophermaier: "Supervised" => "Supervisor"

christophermaier: Could also say "seconds", too.

components/launcher/src/server/mod.rs
baumanj: The server is now scoped within the `loop` on line 397, so when it restarts the old server is dropped (triggering the cleanup) and a new server is created.

components/launcher/src/server/mod.rs
baumanj: Ooh good catch on the typo, will fix.

But why the "secs negativity"? 😜

components/launcher/src/server/mod.rs
christophermaier: I don't think that's true... `supervisor` is an instance of [std::process::Child](https://doc.rust-lang.org/1.27.2/std/process/struct.Child.html), and the docs say:

> There is no implementation of Drop for child processes, so if you do not ensure the Child has exited then it will continue to run, even after the Child handle to the child process has gone out of scope.

baumanj: I'm not relying on the drop to terminate `supervisor: Child`. On the contrary, the outer loop is only ever restarted when `server.tick()` returns `Err`, which implies the supervisor has already exited.

The tests demonstrate this works and it doesn't leave an old supervisor process running:

```
➤ env RUST_LOG=1 HAB_FEAT_TEST_EXIT=/tmp/exit-file TESTING_FS_ROOT=(mktemp -d)  HAB_SUP_BINARY=(pwd)/target/debug/hab-sup HAB_LAUNCH_BINARY=(pwd)/target/debug/hab-launch hab sup run
DEBUG: setting custom filesystem root for testing only (TESTING_FS_ROOT='/tmp/tmp.4aM77rLXOEhab-sup(MN): Enabling feature: TestExit
DEBUG: setting custom filesystem root for testing only (TESTING_FS_ROOT='/tmp/tmp.4aM77rLXOE')
hab-sup(MR): Supervisor Member-ID 8da7aa59e93c4215b9406d8610aac2a3
hab-sup(MR): Starting ctl-gateway on 127.0.0.1:9632
```
In other shell:
```
➤ pgrep -afl  hab-
118443 /home/jbauman/habitat/target/debug/hab-launch run
118445 /home/jbauman/habitat/target/debug/hab-sup run

➤ echo "101" > /tmp/exit-file
```
```
hab-sup(MR): Simulating abrupt, unexpected exit with code 101
```
```
➤ pgrep -afl  hab-
118443 /home/jbauman/habitat/target/debug/hab-launch run

➤ pgrep -afl  hab-
118443 /home/jbauman/habitat/target/debug/hab-launch run
118661 /home/jbauman/habitat/target/debug/hab-sup run
```
```
hab-sup(MN): Enabling feature: TestExit
DEBUG: setting custom filesystem root for testing only (TESTING_FS_ROOT='/tmp/tmp.4aM77rLXOE')
hab-sup(MR): Supervisor Member-ID 8da7aa59e93c4215b9406d8610aac2a3
hab-sup(MR): Starting ctl-gateway on 127.0.0.1:9632
```
That said, there's no _harm_ in calling `kill/wait` once `supervisor: Child` has already exited, so I can add that to the `drop` implementation for robustness.

baumanj: The more I think about it, you have a really good point. While I don't think it's broken the way the code is structured now, it's certainly not as robust as it should be.

Also, just to note, in the update I made the [`wait`](http://doc.rust-lang.org/1.28.0/std/process/struct.Child.html#method.wait) into a [`try_wait`](http://doc.rust-lang.org/1.28.0/std/process/struct.Child.html#method.try_wait). I don't think the latter existed at the time this code was originally written, but we absolutely don't want the `drop` function to block.

PR #5414, opened 2018-08-03
[ci] Reimplement lint.sh program for better resilience.

fnichol:
This change reimplements the `support/ci/lint.sh` program which tries to
ensure that any Rust code contributions were prepared with the `rustfmt`
tool. The CI invocation of this programs is identical to the prior
implementation, but there are several advantages to this new approach:

* Drop reliance on flags and options in `rustfmt` to better bridge any
future changes to the program. This program attempts to use `rustfmt`
closer to how a code editor would interact with the formatter.
* Update the exit code checking of `rustfmt`, which has slightly changed
over the last 12+ months.
* Drop any remaining references to `rustfmt` version checks. Our current
approach is to use the version of `rustfmt` which ships as a component
in Rustup alongside a Rust toolchain.
* Expose explicit modes of operation via flags: `all`, `cached`,
`unstaged`, `files`, and `git`.
* Add a `--files` option, allowing easier local lint checking on
specific files.
* Add self-documenting usage via a `--help` flag.

### DIFF
--- a/support/ci/lint.sh
+++ b/support/ci/lint.sh
@@ -1,141 +1,361 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-# Fail if there are any unset variables and whenever a command returns a
-# non-zero exit code.
-set -eu
+main() {
+  # Fails on unset variables and whenever a command returns a non-zero exit
+  # code, inside or outside of a pipeline
+  set -euo pipefail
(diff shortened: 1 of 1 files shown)


baumanj:
I assume these 3 PRs are identical? I'm really wanting a fix for https://github.com/habitat-sh/habitat/issues/4852

PR #5418, opened 2018-08-07
Resurrect enough of the old macOS build to work in TravisCI

christophermaier:
We removed too much! In order to continue getting macOS builds of
`hab` when we merge code to master, we still need to have this code
around.

I've consolidated things into a `travis` subdirectory, though, which
should make it easy to extract once we're completely in Buildkite.

Paths have been modified where necessary to account for the additional
level of directory structure.


thesentinels:

Thank you for contributing!

PR #5421, opened 2018-08-07
ensure RUNTIME_ENVIRONMENT does not include studio prefix

mwrock:
This is a regression that snuck in to 0.56.0 that prepends the full studio path to the runtime dependency paths. So the deps can't be located in a supervisor because it will be rooted to `c:\`.


thesentinels:

Thank you for contributing!

PR #5423, opened 2018-08-08
Provide a way to communicate with one's own Butterfly server

### LINKED ISSUE #5257: supervisor manager don't use gossip_listen address from config
When starting the supervisor with the `--listen-gossip` option binding it to a specific `ip:port` (eg: `192.168.0.10:9638`) then any `hab apply config` fail.

More detail can be found at https://forums.habitat.sh/t/hab-0-57-0-hab-config-apply-issue/609

From what I found in the [supervisor manager code](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L873) the adress is not taken from the config but hardcoded to `127.0.0.1`.

The same issue is present at:

- https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L911
- https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L1259

christophermaier:
Previously, when attempting to apply configuration, add files, or
depart a Supervisor (i.e., any command needing to interact directly
with the gossip layer), the operations would only work if the
Supervisor was listening on the default gossip address, which is
`0.0.0.0`. This is listening on _all_ interfaces, and thus, the
hard-coded loopback address `127.0.0.1` worked fine.

However, when listening on a non-default address, using
`--listen-gossip`, this would fail, since Butterfly would no longer be
listening at `127.0.0.1`.

Here, we provide a new `local_addr` function for the
`GossipListenAddr` struct. This function provides an address at which
the Butterfly server may be contacted locally, taking into account the
address it is listening at. If it is using the default `0.0.0.0` (the
"unspecified address"), it will return an address at `127.0.0.1`. On
the other hand, if a user has configured a different address, that
will be used directly.

Fixes #5257

### DIFF
--- a/components/sup/src/config.rs
+++ b/components/sup/src/config.rs
@@ -37,6 +37,24 @@ static LOGKEY: &'static str = "CFG";
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct GossipListenAddr(SocketAddr);
 
+impl GossipListenAddr {
+    /// Generate an address at which a server configured with this
+    /// GossipListenAddr can communicate with itself.
+    ///
+    /// In particular, a server configured to listen on `0.0.0.0` vs
+    /// `192.168.1.1` should be contacted via `127.0.0.1` in the
+    /// former case, but `192.168.1.1` in the latter.
+    pub fn local_addr(&self) -> Self {
+        let mut addr = self.clone();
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

fnichol:
Post-facto 👍

PR #5430, opened 2018-08-09
Make docker studio behave like chroot studio

elliott-davis:
This means that a docker studio is no longer completely ephemeral. It will persist between exits.


thesentinels:

Thank you for contributing!

components/hab/src/command/studio/docker.rs
```diff
@@ -219,8 +227,35 @@ where
...
+    }
+    return false;
+}
+
+fn path_to_name() -> Result<String> {
+    let mut cwd = env::current_dir()?;
+    cwd = cwd.strip_prefix("/")?.to_path_buf();
+    let pathstr = cwd.to_str().expect("Path to be parseable");
+    return Ok(pathstr.replace("/", "--"));
+}
```
mwrock: I don't think this will work on Windows. You may want to replace backslashes with slashes first and then strip *up to* the first `/`.

elliott-davis: Would it also work to make this function a cfg target since all the path logic is isolated here?

PR #5433, opened 2018-08-09
Elliott/legacy studio

elliott-davis:
This has some merge commits from open PR's. I'll cleanup this PR after those have merged but they are required to test this change.

### DIFF
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -19,7 +19,6 @@ use std::process::{Command, Stdio};
 
 use command::studio::enter::ARTIFACT_PATH_ENVVAR;
 use common::ui::UI;
-use feat;
 use hcore::crypto::default_cache_key_path;
 use hcore::env as henv;
 use hcore::fs::{find_command, CACHE_ARTIFACT_PATH, CACHE_KEY_PATH};
@@ -30,7 +29,7 @@ use VERSION;
 
 const DOCKER_CMD: &'static str = "docker";
 const DOCKER_CMD_ENVVAR: &'static str = "HAB_DOCKER_BINARY";
-const DOCKER_IMAGE: &'static str = "habitat-docker-registry.bintray.io/studio";
+const DOCKER_IMAGE: &'static str = "habitat/default-studio";
 const DOCKER_WINDOWS_IMAGE: &'static str = "habitat-docker-registry.bintray.io/win-studio";
 const DOCKER_IMAGE_ENVVAR: &'static str = "HAB_DOCKER_STUDIO_IMAGE";
 const DOCKER_OPTS_ENVVAR: &'static str = "HAB_DOCKER_OPTS";
@@ -197,10 +196,6 @@ where
     for vol in volumes {
         cmd_args.push("--volume".into());
         cmd_args.push(vol.as_ref().into());
-    }
-
-    if !feat::is_enabled(feat::RootlessStudio) {
-        cmd_args.push("--privileged".into());
     }
     cmd_args.push(image_identifier(docker_cmd).into());
     cmd_args.push("-V".into());
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -38,7 +38,19 @@ const HAB_STUDIO_SECRET: &'static str = "HAB_STUDIO_SECRET_";
 
 pub fn start_docker_studio(_ui: &mut UI, mut args: Vec<OsString>) -> Result<()> {
     let docker_cmd = find_docker_cmd()?;
+    // We need to strip out the -D if it exists to avoid
+    // it getting passed to the sup on entering the studio
+    let to_cull = OsString::from("-D");
+    if let Some(index) = args.iter().position(|x| *x == to_cull) {
+        args.remove(index);
+    }
+    if args[0] == OsString::from("rm") {
+        return rm_container(&docker_cmd);
+    }
+    return enter(&docker_cmd, &args);
+}
 
+fn enter(docker_cmd: &Path, args: &Vec<OsString>) -> Result<()> {
     if is_image_present(&docker_cmd) {
         debug!("Found Studio Docker image locally.");
     } else {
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

fnichol:
I'm good following up to the point where we're swapping the default `chroot` impl Studio for the Docker impl Studio. Personally, I'm still not 100% convinced either way, however I would rather have users exercising the improvements and fixes with all of this work before cutting folks over to a scenario where they now require a running Docker engine.

I guess that's way of saying I'd rather have a way to use the Docker impl explicitly in the next release but not make it the default until a) we're convinced this is the way we want to go on Linux, b) we've addressed all the documentation and tutorial updates required for users to transition, c) account for these changes in the bootstrapping and core Plans refresh workflow, d) understand any differences and limitations when dealing with new Linux package targets that are for older Linux kernels (which may or may not have full support to run Docker engines).

This isn't in place of any code review and testing it out, but I wanted to raise this as a concern that only affects the last commit (I believe).

elliott-davis:
@fnichol I totally agree. I wanted to get this PR out there so we could talk about it and (assuming the other two PR's merge) test it out after the next release. I figured there would be opinions on if this should merge at all, and if it did merge, what the migration should look like.

To be clear, I'm :-1: on merging this before the other container PRs have at least one release cycle to soak, if not longer.

elliott-davis:
The upside is mostly consistency across osx, windows and linux. This also solves the isolation issue that the current chroot studio has for processes. Admittedly, this wasn't the biggest issue the project or the studio code is facing but it was a cheap change to get consistency across the way we currently launch studios.

adamhjk:
I think I worry about this on a number of different fronts. One is that we often build on platforms that may or may not work with docker. For example - if you were to start supporting OS X plans, you won't get a docker implementation that runs OS X in a container. So you'll have a playform that needs its own isolation. I think this is the same with windows builds?

That's without the platform issues, where docker gets fixed to a kernel thats newer than at least one of our supported architectures. 

I'm not sure our initial decision to support linux builds natively on alternate platforms was the right one, as our cross platform story gets more fleshed out. And it makes me nervous about switching to the docker approach by default, given that i'm not sure it's a sane default everywhere. What do y'all think?

adamhjk:
To be clear - I'm not opposed to having the docker isolation available. I don't think it should be the default - it means that every user needs multiple pieces of software, it wont' work across every platform, and it isn't broke. :)

adamhjk:
Over time, I think the default should be the native platform tooling if we support it. And the docker isolation can be available for cross platform building.

mwrock:
I agree that docker should not be the default. Windows is a good example. Docker only works on the latest releases of Win 10 and server 2016. Lots of enterprises do not support those. They can get around this by adding the `-w` flag to `hab studio enter` or `hab pkg build` but its a weird experience. Really the default should be the one that "always works."

However, if you DO have docker. Having that isolation level can be a great experience. It's just better to "opt in" to that.

adamhjk:
Also, if it's about process isolation with chroot, we can get that without docker if we need it. ie: if there is an isolation bug in the default studio, we can likely just do the work to get that isolation bug fixed (while also making docker available)

christophermaier:
Based on preceding conversation, I'm adding the `DO-NOT-MERGE` label here for now.

elliott-davis:
@mwrock I'm a bit confused. Today, Docker is the default in a Windows studio. Are you saying you'd prefer to see a change that removes that as the default? I don't really have a preference either way, I just want the experience to be consistent. 

We don't have native mac builds anywhere on the roadmap so that isn't as big of a concern for me at the moment. 

I agree that needing multiple pieces of software to run a studio build is a bummer. Maybe a solid compromise would be to check if Docker is available and use it, otherwise fall back to the chroot studio? Users on OSX and Windows already have this problem as well.

Currently the native studio on Linux provides filesystem isolation but doesn't isolate processes or network traffic. A couple of the issues this causes are:
* A Supervisor running on the host having port collisions with a Supervisor running in the studio
* Host caching processes affecting builds inside the studio (sccache, ccache)
* Not being able to run multiple studios concurrently on a build slave

Additionally, to make the studio functional, we mound in `dev` and `proc` which has caused a trove of umount issues when the studio fails to exit cleanly. I think @fnichol is making a pass at fixing some of this behavior in #5434. There have been several other attempts at cleaning this up also https://github.com/habitat-sh/habitat/blob/master/components/studio/bin/hab-studio.sh#L427-L431

With regards to kernel version support, that could be a legitimate problem. As of right now, Docker supports kernel version 3.1, which is the same version that our latest version of glibc supports. I think the compromise mentioned above would be a clean escape hatch here where we default to Docker, but fall back to the chroot if it's unavailable (it would be unavailable be if you're running on OS with an unsupported kernel)

We advertise the studio as a clean room and not providing the right level of isolation feels incorrect. I remember my first time running a build on a Linux workstation and being shocked when running `ps` inside the studio and seeing my chrome tab processes :joy:. If we think driving the current Windows and Linux studios toward providing process and network isolation without Docker that's totally fine with me. I just saw this as a quick win opportunity to get there with minimal effort.

mwrock:
Hey sorry to add to confusion @elliott-davis ! I think with the exception of MACs, the default studio should be non-docker even if it is not the most isolated. In fact on Windows, the isolation problem is worse than linux. A chrooted studio is way better that the "native/local" windows studio. However, `hab studio enter` will ALWAYS work locally and it feels overly "magical" to run in a docker studio if docker is installed even if it is a better studio.

I would suggest that when entering a local windows or chrooted linux studio as default, we add messaging that says something like "use the --docker argument to enter a docker studio".

adamhjk:
I agree with Matt.

adamhjk:
Longer form - the issue is the obstacle for the user. We are now reliant on the external environment to do the most foundational task in habitat. I think it’s got to work cleanly and seamlessly across multiple platforms without dependencies by default - it’s part of the magic :tm:. Obviously, we made compromises (Mac Linux builds, Linux by default on windows) - but I’m not sure in hindsight they were the rifht ones. 

When we say clean room, it was intentionally not about isolation in the process management sense - it was isolation from pollution from outside side effects. I think improving the studios isolation is great, and having docker studio options is great, but I think it’s the wrong default (Windows should switch).

adamhjk:
Isolated from the external build environment.

elliott-davis:
Closing since it seems like this is unwanted.

PR #5434, opened 2018-08-09
[studio] Fast-fail when unmounting on Studio cleanup.

fnichol:
This change alters the strategy used when a Studio program invocation is
cleaning up before shutting down. During a normal shutdown, a
pre-emptive aborting, and when calling `hab studio rm` directly (or
indirectly via `hab studio build`/`hab pkg build`), an internal function
called `unmount_filesystems` is called in order to unmount several
important filesystems, namely:

* `$root/src` (bind mounted)
* `$root/hab/cache/artifacts` (bind mounted, if artifact caching is
enabled)
* `$root/run` (mounted via tmpfs)
* `$root/sys` (normally bind mounted)
* `$root/proc` (mounted via procfs)
* `$root/dev/pts` (mounted via devpts)
* `$root/dev` (bind mounted)
* `$root/var/run/docker.sock` (bind mounted, if a local Docker socket is
found)

Where `$root` is the Studio root directory, typically something like
`/hab/studios/home--user--prj`.

Prior to this change during `unmount_filesystems` the code *attempted*
to unmount these filesystems, meaning that it continued to the next
unmount regardless of success or failure on each `umount` call. This
change takes a very different approach, that of "do no further harm" and
"fail fast". That is, if any filesystem umount fails, the program will
now error and abort with a well-known exit code rather than proceeding
to umount more filesystems. This means that a user can attempt to remedy
any filesystem or operating system issues before attempting to re-run
their last Studio command.

A final change is specifically around the `hab studio rm` behavior which
happens after unmounting known filesystems under the Studio root
directory (note again that the `hab studio rm` behavior is triggered
internally by `hab studio build`/`hab pkg build` before creating a new
Studio instance). Prior to this change, the code assumed that if all
filesystems were unmounted (and it couldn't assume 100% success which
was a faulty assumption), then we could safely call `rm -rf $root` to
remove all remaining files and directories. What can happen in practice
is that there might be additional filesystems unknown to that Studio
instance which are still actively mounted. This change performs a sanity
check to guard the `rm -rf` call to ensure that there are no mounted
filesystems. If it finds any mounts, it will error an abort similar to
the above behavior so as to avoid doing any potential harm to inner
projects, host filesystems such as `/dev`, `/proc`, etc. This new
behavior leaves it to the user to fix/remedy/repair any issues before
retrying the last Studio command. This change is intended to guard
against situations where an active inner studio might have the system's
`/proc`/`/sys`/`/dev` filesystems mounted, and the outer Studio tries to
`rm` these "files".

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -712,8 +712,23 @@ rm_studio() {
 
   cleanup_studio
 
-  # Remove remaining filesystem
-  $bb rm -rf $v "$HAB_STUDIO_ROOT"
+  if $bb mount | $bb grep -q "on ${HAB_STUDIO_ROOT}.* type"; then
+    # There are still mounted filesystems under the root. In the spirit of "do
+    # no further harm", we abort here with a message so the user can resolve
+    # and retry.
+    >&2 echo "After unmounting all known filesystems, there are still \
+remaining mounted filesystems under ${HAB_STUDIO_ROOT}:"
+    $bb mount \
+      | $bb grep "on ${HAB_STUDIO_ROOT}.* type" \
+      | $bb sed 's#^#    * #'
+    >&2 echo "Unmount these remaining filesystem using \`umount(8)'and retry \
+the last command."
+    exit_with "Remaining mounted filesystems found under $HAB_STUDIO_ROOT" \
+      "$ERR_REMAINING_MOUNTS"
+  else
+    # No remaining mounted filesystems, so remove remaining files and dirs
+    $bb rm -rf $v "$HAB_STUDIO_ROOT"
+  fi
 }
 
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/studio/bin/hab-studio.sh
```diff
@@ -712,8 +712,23 @@ rm_studio() {
 
   cleanup_studio
 
-  # Remove remaining filesystem
-  $bb rm -rf $v "$HAB_STUDIO_ROOT"
+  if $bb mount | $bb grep -q "on ${HAB_STUDIO_ROOT}.* type"; then
```
christophermaier: Would it be possible to augment`is_fs_mounted` to be able to use it here, too?

Just thinking of ways to consolidate the core logic in one place.

fnichol: I noticed that as well and thought the best I could do was to check for `grep -q "on ${HAB_STUDIO_ROOT}/.* type"` (notice the directory slash) which could allow for `if is_fs_mounted '.*'; then …; fi` although that felt a bit aggressive too. This current case additionally should be able to detect if the `$HAB_STUDIO_ROOT` path itself was mounted (why in the world would this be mounted you ask? I don't know but I figured in the pathological case, it was worth checking).

PR #5441, opened 2018-08-13
add launcher appveyor build and always publish to unstable

mwrock:
We currently manually build the windows launcher on a developer laptop. The laptop is pretty nice but we want the build to come from CI. This will ensure that all launcher merges will produce an unstable build on builder. It will never include it in our RC channel so it does not get promoted to stable in every release. We will just promote manually on an as needed basis.


thesentinels:

Thank you for contributing!

PR #5444, opened 2018-08-13
Fix some release issues

christophermaier:
- It turns out that "repository" has a "t" in it; who knew?
- Use the `is_fake_release` helper function in the DockerHub scripts
  These were failing because the check for the `FAKE_RELEASE_TAG`
  variable would fail if the variable was completely unset.

tenor-21441224


thesentinels:

Thank you for contributing!

Issue #5162, opened 2018-06-06
[feature] hab-sup controlled upgrade


stevendanna:
## Background

The auto-upgrade feature of hab-sup allows the supervisor to upgrade without killing the running services.  This is a great feature.

We have an application that makes heavy use of Habitat.  While we some day hope to move almost all coordination into the hands of habitat, right now, one part of that application orchestrates certain lifecycle events by interacting with hab-sup via the hab command line.  This application thus has command line flags and workarounds that are specific to a particular version of habitat.  As new versions of Habitat come out, we'd like to upgrade since it allows us to remove workarounds and pass more responsibilities to Habitat.  However, since we depend on particular command line switches, we cannot upgrade hab-sup at any time.  Rather, we need to upgrade only after our own software has been upgraded to a version that supports the new version of hab-sup.

## Feature

`hab-sup` should allow the user to initiate restart of the supervisor in a way that does not shutdown any of the services.  This would allow us to upgrade the hab-sup package on disk and then trigger the restart when we were ready for it.

Right now, `hab sup term` and killing the `hab-sup` process with a signal both result in all supervised processes being shutdown.

## Possible implementation

Either of these would work for us, but I imagine there are others y'all could dream up:

1. Define a signal we can send to hab-sup that will cause hab-sup to exit in the same way it exits before an auto-upgrade.

2. Define a new `hab sup BLAH` command that will cause hab-sup to exit cleanly OR which can upgrade the hab-sup to a given hab-sup version.

baumanj:
This seems like a good thing to have and which should be very easy to implement with the capabilities that the supervisor already possesses.

stevendanna:
@baumanj In thinking about this more.  I think some more thought about how this would interact with other behavior might work.

Namely, right now, if we want to be sure that a rouge package install can't break us, we set HAB_SUP_BINARY in our supervisor startup script.  While we remove that environment variable, that would leave us open to hab-sup being upgraded on us unexpectedly.  I think it might be good to take a look at how users are expected to manage hab-sup's lifecycle.

baumanj:
@stevendanna: take a look at https://github.com/habitat-sh/habitat/pull/5409

Basically it should let you `hab sup run` without `--auto-update`, but it will still update the supervisor (without taking down services) if you send `SIGHUP`, so you have control over exactly when that happens.

Issue #5300, opened 2018-07-09
Simplify launcher to make changes rare


baumanj:
Launcher exists to be the parent process of the supervisor and habitat service processes. This allows the supervisor to restart (for upgrade or in case of failure) without disturbing running services. Without this model, any supervisor upgrades would require any services running under habitat to experience downtime. This model works great, unless the launcher itself needs to change. In that case, the launcher needs to be manually installed and any running services need to be stopped.

In the last 6 releases (0.54–0.58), the there have been changes to the launcher in 3 of them. If half of our releases require users to take downtime in production, that will pose a serious problem to adoption. To avoid this, we should endeavor to make the launcher as simple as possible *now* so that future changes will indeed be rare.

Currently, the launcher is not only responsible for starting processes, but also killing them. This introduces significant, unnecessary complexity since sibling processes with correct permissions (such as the supervisor or the `hab` process) are perfectly capable of handling process termination on both Linux and Windows.

Dealing with process termination has already required [subtle and complex code to deal with orphaned zombies](https://github.com/habitat-sh/habitat/pull/4683), and there is at least one additional [known issue with the signal handling](https://github.com/habitat-sh/core/issues/11) that would require launcher changes to fix. Furthermore, [recent work to ensure PID files for service processes are removed when the supervisor shuts down](https://github.com/habitat-sh/habitat/pull/5236) made it apparent that the supervisor/launcher interaction is susceptible to deadlock since there is no defined order of resource acquisition.

To address all these issues and help reduce the risk of launcher updates in the future, we should reduce the functionality of launcher to its bare minimum: executing processes so that the supervisor process is not the parent process of the service processes.

baumanj:
https://github.com/habitat-sh/habitat/issues/5358 will have to be resolved as part of this work since the solution depends upon changes on the launcher.

baumanj:
[Unsafe, platform-specific Process struct is obviated by rust stdlib now](url
https://github.com/habitat-sh/habitat/issues/5361) should also be addressed as part of this work.

PR #5445, opened 2018-08-13
Shut down the supervisor (but not services) when SIGHUP is received

baumanj:
This allows for a controlled update to she supervisor binary.

Resolves https://github.com/habitat-sh/habitat/issues/5162

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -53,6 +53,7 @@ use hcore::crypto::SymKey;
 use hcore::env;
 use hcore::fs::FS_ROOT_PATH;
 use hcore::os::process::{self, Pid, Signal};
+use hcore::os::signals::{self, SignalEvent};
 use hcore::package::metadata::PackageType;
 use hcore::package::{Identifiable, PackageIdent, PackageInstall};
 use hcore::service::ServiceGroup;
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -692,6 +696,11 @@ impl Manager {
                 self.shutdown(ShutdownReason::Departed);
                 return Err(sup_error!(Error::Departed));
             }
+            if let Some(SignalEvent::Passthrough(Signal::HUP)) = signals::check_for_signal() {
+                outputln!("Supervisor shutting down for signal");
+                self.shutdown(ShutdownReason::Signal);
```
christophermaier: This `ShutdownReason` enum and its handling is one of the targets of my current refactoring... it's essentially being passed down ultimately to the individual service-shutdown logic. For this work, it would be good to trace this through and in any match / if/else statements where we have branching logic based on `ShutdownReason`, we should ensure that `ShutdownReason::Signal` behaves like `ShutdownReason::PkgUpdating` (since that's what gets passed through when the Supervisor is shutting down for an update).

My examination of the code thus far suggests that we've likely built up a bit too much complexity around `ShutdownReason`, and I think I'll be simplifying a lot of it (ultimately, I think we'll end up with a 2-member enum that just encapsulates the "should we shutdown the services, too?" question). For the purposes of getting this feature in, though, making the above change here should suffice.

baumanj: I agree wholeheartedly with the complexity around `ShutdownReason` and that's what inspired https://github.com/habitat-sh/habitat/issues/5300 in the first place.

However, I think this is fully plumbed. I followed all the code paths through and I believe it does the correct thing without any additional logic. I could've re-used `ShutdownReason::PkgUpdating`, but that seemed misleading. However, there's nothing that checks for `PkgUpdating`, specifically. Any `ShutdownReason` that *isn't* `LauncherStopping` or `SvcStopCmd` behaves identically.

Do you see the need for a code change somewhere?

christophermaier: I was thinking from the perspective of documenting intent more than anything; I suspect you are correct that functionally things are OK as is.

Since I'm going to be working on cleaning this up _now_, I'm not going to forget this context, though, so we can merge now if you like.

PR #5449, opened 2018-08-14
Fix exit code for publish when pushing containers to dockerhub

elliott-davis:
Ensure we set the correct channel for buildkite

### DIFF
--- a/components/rootless_studio/publish-to-dockerhub.sh
+++ b/components/rootless_studio/publish-to-dockerhub.sh
@@ -13,7 +13,7 @@ if [[ -z $DOCKER_LOGIN_PASSWORD ]]; then
 fi
 
 docker login -u="$DOCKER_LOGIN_USER" -p="$DOCKER_LOGIN_PASSWORD"
-trap 'rm -f $HOME/.docker/config.json; exit $?' INT TERM EXIT
+trap 'rm -f $HOME/.docker/config.json' INT TERM EXIT
 
 # Always push the version
 docker push "${IMAGE_NAME}:${VERSION}"
--- a/.buildkite/scripts/dockerhub_publish.sh
+++ b/.buildkite/scripts/dockerhub_publish.sh
@@ -4,6 +4,7 @@ set -euo pipefail
 
 source .buildkite/scripts/shared.sh
 version=$(buildkite-agent meta-data get "version")
+channel=$(buildkite-agent meta-data get release-channel)
 
 if [[ "${FAKE_RELEASE_TAG}" ]]; then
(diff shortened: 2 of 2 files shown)


components/rootless_studio/publish-to-dockerhub.sh
```diff
@@ -13,7 +13,7 @@ if [[ -z $DOCKER_LOGIN_PASSWORD ]]; then
 fi
 
 docker login -u="$DOCKER_LOGIN_USER" -p="$DOCKER_LOGIN_PASSWORD"
-trap 'rm -f $HOME/.docker/config.json; exit $?' INT TERM EXIT
+trap 'EXIT=$?; rm -f $HOME/.docker/config.json; exit EXIT' INT TERM EXIT
```
christophermaier: I don't think this is exit status handling is needed.

components/rootless_studio/publish-to-dockerhub.sh
```diff
@@ -13,7 +13,6 @@ if [[ -z $DOCKER_LOGIN_PASSWORD ]]; then
 fi
 
 docker login -u="$DOCKER_LOGIN_USER" -p="$DOCKER_LOGIN_PASSWORD"
-trap 'rm -f $HOME/.docker/config.json; exit $?' INT TERM EXIT
```
christophermaier: No, I mean I don't think you need to manually manage the exit status; I think the `trap` itself is still probably useful.

elliott-davis: If you trap, won't the program exit with whatever command trap ran?

christophermaier:
```sh
bash-4.4$ cat ./traptest.sh
#!/bin/bash

exit_with=${1}

trap 'rm foo' EXIT
exit $exit_with
bash-4.4$ rm foo
rm: foo: No such file or directory
bash-4.4$ echo $?
1
bash-4.4$ ./traptest.sh 3
rm: foo: No such file or directory
bash-4.4$ echo $?
3
bash-4.4$ ./traptest.sh 0
rm: foo: No such file or directory
bash-4.4$ echo $?
0
bash-4.4$ bash --version
GNU bash, version 4.4.23(1)-release (x86_64-apple-darwin16.7.0)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
bash-4.4$
```
hat-tip to @baumanj

PR #5451, opened 2018-08-14
Don't quote HAB_STUDIO_SUP

christophermaier:
If the variable isn't set, we end up passing "" as an argument, which
is interpreted by the Supervisor as a package identifier, which is not
legal.

This results in the Supervisor not being able to start properly when
entering a Studio. Manually running `sup-run` afterward starts things
up with no issues, though.

### DIFF
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -219,7 +219,8 @@ case "\${HAB_STUDIO_SUP:-}" in
     # If false, we don't run the Supervisor
     ;;
   *)
-    sup-run "\${HAB_STUDIO_SUP:-}"
+    # shellcheck disable=2086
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/studio/libexec/hab-studio-type-default.sh
```diff
@@ -219,7 +219,7 @@ case "\${HAB_STUDIO_SUP:-}" in
     # If false, we don't run the Supervisor
     ;;
   *)
-    sup-run "\${HAB_STUDIO_SUP:-}"
+    sup-run \${HAB_STUDIO_SUP:-}
```
elliott-davis: For these without an explicit shellcheck disable, I think a comment would be good so we don't repeat the past

christophermaier: good point

PR #5452, opened 2018-08-15
move build logic for rootless studio into buildkite

elliott-davis:


thesentinels:

Thank you for contributing!

.buildkite/scripts/dockerhub_publish.sh
```diff
@@ -4,15 +4,27 @@ set -euo pipefail
...
 if is_fake_release; then
   # This overrides IMAGE_NAME in docker-base.sh
   export IMAGE_NAME="habitat/fakey-mc-fake-face-studio"
+else
+  export IMAGE_NAME="habitat/default-studio"
 fi
 
+if [[ -z $DOCKER_LOGIN_USER ]]; then
+  echo "DockerHub user required" && exit 1
+fi
```
christophermaier: Strictly speaking, the `set -u` bit at the top will handle this.

.buildkite/scripts/dockerhub_publish.sh
```diff
@@ -4,15 +4,27 @@ set -euo pipefail
...
+
+if [[ -z $DOCKER_LOGIN_PASSWORD ]]; then
+  echo "DockerHub password required" && exit 1
+fi
+
+docker login -u="$DOCKER_LOGIN_USER" -p="$DOCKER_LOGIN_PASSWORD"
```
christophermaier: Minor, but it's generally good to use the long options in scripts for maximum readability, even though `u` and `p` are probably pretty idiomatic.

```sh
docker login \
       --username="$DOCKER_LOGIN_USER" \
       --password="$DOCKER_LOGIN_PASSWORD"
```

.buildkite/scripts/dockerhub_publish.sh
```diff
@@ -4,15 +4,27 @@ set -euo pipefail
 
 source .buildkite/scripts/shared.sh
 version=$(buildkite-agent meta-data get "version")
-channel=$(buildkite-agent meta-data get release-channel)
 
 if is_fake_release; then
   # This overrides IMAGE_NAME in docker-base.sh
   export IMAGE_NAME="habitat/fakey-mc-fake-face-studio"
+else
+  export IMAGE_NAME="habitat/default-studio"
```
christophermaier: Does this need to be exported now? It looks like you're using it pretty explicitly.

I think the general convention for things that are just local variables (as opposed to environment variables) is to have them be in lower-case, e.g., `image_name`.

It doesn't look like we're sourcing `docker-base.sh`, either, so we can take that comment out.

.buildkite/scripts/dockerhub_upload.sh
```diff
@@ -4,14 +4,30 @@ set -euo pipefail
 
 source .buildkite/scripts/shared.sh
 version=$(buildkite-agent meta-data get "version")
+channel=$(buildkite-agent meta-data get "release-channel")
 
 if is_fake_release; then
   # This overrides IMAGE_NAME in docker-base.sh
   export IMAGE_NAME="habitat/fakey-mc-fake-face-studio"
+else
+  export IMAGE_NAME="habitat/default-studio"
```
christophermaier: Same w/r/t exporting and upper/lower case, and the comment about `docker-base.sh`

.buildkite/scripts/dockerhub_upload.sh
```diff
@@ -4,14 +4,30 @@ set -euo pipefail
...
 fi
 
+if [[ -z $DOCKER_LOGIN_USER ]]; then
+  echo "DockerHub user required" && exit 1
+fi
+
+if [[ -z $DOCKER_LOGIN_PASSWORD ]]; then
+  echo "DockerHub password required" && exit 1
+fi
```
christophermaier: Same w/r/t `set -u`

.buildkite/scripts/dockerhub_upload.sh
```diff
@@ -4,14 +4,30 @@ set -euo pipefail
...
+
+docker login -u="$DOCKER_LOGIN_USER" -p="$DOCKER_LOGIN_PASSWORD"
```
christophermaier: Same w/r/t long options

.buildkite/scripts/dockerhub_upload.sh
```diff
@@ -4,14 +4,30 @@ set -euo pipefail
...
+trap 'rm -f $HOME/.docker/config.json' INT TERM EXIT
+
 pushd ./components/rootless_studio >/dev/null
-./build-docker-image.sh "${version}"
-./tag-docker-image.sh "${version}" "rc"
-./publish-to-dockerhub.sh "${version}" "rc"
+docker build -t habitat:hab-base .
+docker build --build-arg BLDR_CHANNEL="${channel}" --no-cache -t "${IMAGE_NAME}:${version}" "./default"
```
christophermaier: `-t` could be `--tag`.

Also, I don't think you need to quote `"./default"`

Splitting this onto several lines could help the readability, as well.

Issue #5460, opened 2018-08-16
0 byte dat file


stevendanna:
We recently had a user report that a Habitat supervisor was failing to start. The supervisor log showed:

```
Aug 15 18:30:17 redacted.host hab[1242]: hab-sup(ER)[components/sup/src/error.rs:449:9]: Butterfly error: Error reading or writing to DatFile, /hab/sup/default/data/cab66ba4cf12423db7b6cce2a1d76c54.rst, failed to fill whole buffer
```

Further investigation showed that `/hab/sup/default/data/cab66ba4cf12423db7b6cce2a1d76c54.rst` existed but was a 0-byte file.

While we haven't been able to investigate the problem further yet, the code that handles the writes:

https://github.com/habitat-sh/habitat/blob/52b8c4eafbc35bf3efaaab4dd702bb5fa740757e/components/butterfly/src/rumor/dat_file.rs#L234-L267

appears to be missing at least two calls to fsync if we want those writes to be durable.  We likely need an fsync on the temporary file before the rename to ensure that the content of the file has been written to disk and an fsync on the enclosing directory after the rename to ensure that the rename operation is flushed to disk.

mwrock:
Closed by #5461

PR #5461, opened 2018-08-16
[butterfly] Make DatFile write more durable

stevendanna:
This adds two calls to fsync in the DatFile write method:

1) On the temporary file before the rename
2) On the parent directory after the rename

The first fsync is to persist the content of the dat file to disk. I
think this might have prevented of the 0-byte DatFile seen in #5460.

The second fsync is to persist the rename operation to disk. I've
no-oped it on Windows since we found that calling sync_all on a
directory in windows produces an error in
b688d8ae25d9f38b4053fd29372e2733c9254327.

### DIFF
--- a/components/butterfly/src/rumor/dat_file.rs
+++ b/components/butterfly/src/rumor/dat_file.rs
@@ -243,7 +243,7 @@ impl DatFile {
                 .truncate(true)
                 .open(&tmp_path)
                 .map_err(|err| Error::DatFileIO(tmp_path.clone(), err))?;
-            let mut writer = BufWriter::new(file);
+            let mut writer = BufWriter::new(&file);
             self.init(&mut writer)?;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

stevendanna:
I haven't been able to confirm that this is the cause of #5460 yet it is a bit tricky to test since (if this is the cause of the failure) it requires us killing/crashing the OS before it has a chance to automatically flush to disk.  Perhaps we could use a FUSE filesystem that injects errors or something.

christophermaier:
@stevendanna Update: the Expeditor failure is a regression around dealing with forks, is currently being worked on, and isn't anything we need to worry about here (as expected).

stevendanna:
Fixed up rust formatting

christophermaier:
@stevendanna Excellent! Thanks again!

PR #5465, opened 2018-08-16
Improve error message when supervisor unavailable

### LINKED ISSUE #5367: Error message improvement
Hi!

I had a troubles with loading service into svc, (I believe because of hard rebooting my pc previously).
I've tried `sudo hab svc load core/postgresql` but got an error stated that `Connection refused (os error 111)`

I've run command `find /hab/svc -name PID -type f -exec cat {} \;`, then restarted hab via
`# systemctl restart hab`;
After that I've been able to load package.

In short chat with @eeyun I got explanation that this error is due to

> the sup is controlled through a remote gateway now* so when you see that connection refused it's because the sup isnt running.

\* a new error message since v0.56 or so

So since it is known issue, may be it worth to change error message for something more meaningful than just `blablabla os error 111`? At least for DEBUG level.

And thanks for such a great support guys!

---
```
$ hab --version
hab 0.59.0/20180712155441
$ cat /etc/issue
Ubuntu 17.10 \n \l
```

adamhjk:
This fixes #5367.

tenor-57499504

Previously, when doing a `hab svc (load|unload)`, or any command that
requires remote control of the supervisor, if there was a network issue,
we would simply return the upstream description of the problem (most
often, something like `Connection refused (os error 111)`.

This wasn't a very user friendly error message. Now, on any `SvcClient`
command that returns an `SvcClientError::Io` error, we will print the
following:

```
✗✗✗
✗✗✗ Unable to contact the Supervisor.
✗✗✗
✗✗✗ If the Supervisor you are contacting is local, this probably means it is not running. You can run a Supervisor in the foreground with:
✗✗✗
✗✗✗ hab sup run
✗✗✗
✗✗✗ Or try restarting the Supervisor through your oerating systems init process.
✗✗✗
✗✗✗ Original error is:
✗✗✗
✗✗✗ Connection refused (os error 111)
✗✗✗
```

Which is, I think, much more helpful.

### DIFF
--- a/components/sup-client/src/lib.rs
+++ b/components/sup-client/src/lib.rs
@@ -100,7 +100,15 @@ impl fmt::Display for SrvClientError {
                 path.display()
             ),
             SrvClientError::Decode(ref err) => format!("{}", err),
-            SrvClientError::Io(ref err) => format!("{}", err),
+            SrvClientError::Io(ref err) => format!(
+                "Unable to contact the Supervisor.\n\n\
+                If the Supervisor you are contacting is local, this probably means it is not running. You can run a Supervisor in the foreground with:\n\n\
+                   hab sup run\n\n\
+                Or try restarting the Supervisor through your oerating systems init process.\n\n\
+                Original error is:\n\n\
+                   {}",
+                err
+            ),
             SrvClientError::NetErr(ref err) => format!("{}", err),
         };
         write!(f, "{}", content)
--- a/components/sup-client/src/lib.rs
+++ b/components/sup-client/src/lib.rs
@@ -104,7 +104,7 @@ impl fmt::Display for SrvClientError {
                 "Unable to contact the Supervisor.\n\n\
                 If the Supervisor you are contacting is local, this probably means it is not running. You can run a Supervisor in the foreground with:\n\n\
                    hab sup run\n\n\
-                Or try restarting the Supervisor through your oerating systems init process.\n\n\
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/sup-client/src/lib.rs
```diff
@@ -100,7 +100,15 @@ impl fmt::Display for SrvClientError {
                 path.display()
             ),
             SrvClientError::Decode(ref err) => format!("{}", err),
-            SrvClientError::Io(ref err) => format!("{}", err),
+            SrvClientError::Io(ref err) => format!(
+                "Unable to contact the Supervisor.\n\n\
+                If the Supervisor you are contacting is local, this probably means it is not running. You can run a Supervisor in the foreground with:\n\n\
+                   hab sup run\n\n\
+                Or try restarting the Supervisor through your oerating systems init process.\n\n\
```
mwrock: `oerating` -> `operating `
`init process` -> `init process or Windows service`

Windows users will not likely be familiar with the term `init process`

christophermaier: `systems` -> `system's`, too

baumanj:
This is a nice improvement!

I think we could go even one better by checking whether there *is* a supervisor running locally and/or whether a `--remote-sup` argument was provided, but this is definitely a good start.

alex-bender:
Hey guys! It's been a while since I've suggested that improvement and I would say that it is really great to see that changes are accepted. Such a good community, thank you all!

PR #5485, opened 2018-08-20
explicitly error when attempting to remove a docker studio

### LINKED ISSUE #5155: hab studio rm of a local windows studio tries to delete c:\
`hab studio rm` is not correctly setting `$HAB_STUDIO_ROOT` and is defaulting to `c:\`

mwrock:
fixes #5155. The Windows Docker studio will try to delete `c:\` which fails but the fact is that removing an ephemeral studio is pointless and this just makes that more explicit.

### DIFF
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -37,6 +37,10 @@ const DOCKER_SOCKET: &'static str = "/var/run/docker.sock";
 const HAB_STUDIO_SECRET: &'static str = "HAB_STUDIO_SECRET_";
 
 pub fn start_docker_studio(_ui: &mut UI, mut args: Vec<OsString>) -> Result<()> {
+    if args.get(0) == Some(&OsString::from("rm")) {
+        return Err(Error::CannotRemoveDockerStudio);
+    }
+
     let docker_cmd = find_docker_cmd()?;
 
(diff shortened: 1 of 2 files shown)


components/hab/src/command/studio/docker.rs
```diff
@@ -37,6 +37,10 @@ const DOCKER_SOCKET: &'static str = "/var/run/docker.sock";
 const HAB_STUDIO_SECRET: &'static str = "HAB_STUDIO_SECRET_";
 
 pub fn start_docker_studio(_ui: &mut UI, mut args: Vec<OsString>) -> Result<()> {
+    if args[0] == OsString::from("rm") {
```
baumanj: You generally want to avoid the index `[…]` operator in rust, as [it can panic](http://doc.rust-lang.org/1.28.0/alloc/vec/struct.Vec.html#indexing).

A good alternative is the [`get` operation provided by the slice type](http://doc.rust-lang.org/1.28.0/std/primitive.slice.html#method.get) (which  a [Vec can deref to implicitly](http://doc.rust-lang.org/1.28.0/alloc/vec/struct.Vec.html#method.deref)), which returns an [`Option`](http://doc.rust-lang.org/1.28.0/core/option/enum.Option.html).

mwrock:
Great tip @baumanj ! Change made.

PR #5486, opened 2018-08-20
fix powershell pkg_svc_run when run outside of a studio

### LINKED ISSUE #5198: pkg_svc_run fails outside of a studio on windows
When writing out the `run` file, plan-build expects the presense of the `$FS_ROOT` which is not defined outside of a studio. This the script in the `run` file errors trying to pass `$null` to `join-path`. We should null check that variable in the script.

mwrock:
fixes #5198 


thesentinels:

Thank you for contributing!

Issue #5467, opened 2018-08-16
Update Cargo.lock


baumanj:
Currently [`Cargo.lock`](https://github.com/habitat-sh/habitat/blob/master/Cargo.lock) is badly out of sync with the dependency versions that would be generated from scratch. Add to this the fact that [`cargo update` for a single package is broken](https://github.com/rust-lang/cargo/issues/5702) and the only option left for accessing new code (which we frequently need to do across our own repositories) is error-prone hand edits to the `Cargo.lock` file.

In order to get `Cargo.lock` to a maintainable state, we must go through the painful process of closing the significant gap between the current dependent package versions and the versions that our various `Cargo.toml` files declare (usually `*`, which resolves to the newest version). Once we get `Cargo.lock` into a good state, it can be regularly updated as part of our post-release process and should be relatively painless.

To get to that point, we must first close the gap. Based on experiences with similar work for the [builder repo](https://github.com/habitat-sh/builder), there may be a significant amount of effort and frustration required. In order to avoid large, risky changes, we will attempt to first lock all out-of-sync dependency versions to their current values in `Cargo.lock`, and then gradually relax them while making the necessary code changes to ensure compatibility. This may result in a significant number of commits, so this issue will be used to track all the added version locking and the PRs which will relax them.

Dependencies that are version-locked by https://github.com/habitat-sh/habitat/pull/5468#issuecomment-413702883, that need to later be relaxed:

<span data-href="#aha_task_list_start" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/>

- [x] rand https://github.com/habitat-sh/habitat/pull/5487 https://github.com/habitat-sh/habitat/pull/5656

- [x] prost, prost-derive, prost-build, prost-types https://github.com/habitat-sh/habitat/pull/5501 https://github.com/habitat-sh/habitat/pull/5506

- [x] uuid https://github.com/habitat-sh/habitat/pull/5492

- [x] protobuf →  [prost](https://crates.io/crates/prost) https://github.com/habitat-sh/habitat/pull/6153, https://github.com/habitat-sh/habitat/pull/6105

- [x] rustoto_core, rusoto_credential, rusoto_ecr https://github.com/habitat-sh/habitat/pull/5493

- [x] pkg-config, tabwriter, serde and serde-json https://github.com/habitat-sh/habitat/pull/5729
[...]

PR #5487, opened 2018-08-20
Relax the version pin for the `rand` crate

baumanj:
This makes the very small necessary fix in components/sup-protocol/src/lib.rs
to support the new version. It also includes unrelated changes as the result
of running `cargo update` to refresh Cargo.lock to the new `rand` version.

See https://github.com/habitat-sh/habitat/issues/5467


thesentinels:

Thank you for contributing!

Issue #5481, opened 2018-08-20
Can't attach in docker studio


elliott-davis:
The error:

`WARN Non interactive studio, skipping 'attach'`

To repro:

put an `attach` in a plan, then try and build it in an interactive studio

PR #5491, opened 2018-08-21
Fix attach not working in docker studio

elliott-davis:

Closes: #5481

### DIFF
--- a/components/rootless_studio/default/profile
+++ b/components/rootless_studio/default/profile
@@ -8,7 +8,7 @@
 umask 022
 
 export PATH=$PATH:/bin
-STUDIO_TYPE="default"
+export STUDIO_TYPE="default"
 # Colorize ls by default
 if command -v dircolors > /dev/null; then
   eval "$(dircolors -b)"


thesentinels:

Thank you for contributing!

components/hab/plan.sh
```diff
@@ -23,6 +23,7 @@ pkg_bin_dirs=(bin)
 bin=$_pkg_distname
 
 _common_prepare() {
+  attach
```
baumanj: I'm not very familiar with this. Can you explain why it wasn't necessary before? Or was this always missing?

elliott-davis:
Oh snap. I put that in for testing and forgot to remove it. Apologies. I'll
get it removed when I get back to my desk.

components/hab/plan.sh
eeyun: I think this might be mistakenly added?

baumanj: Verily 
https://github.com/habitat-sh/habitat/pull/5491#issuecomment-414700519

PR #5492, opened 2018-08-21
Relax the version pin for the `uuid` crate

baumanj:
This removes the pinning in the `Cargo.toml` files, but the changes to the `Cargo.lock` don't concern `uuid` as it was already moved past 0.6.0 with 
https://github.com/habitat-sh/habitat/commit/cf3cc543716aa681a28154ba6c4e364d21bf8e93. Some of our dependencies are also locked to other versions of `uuid`, but that is outside our control and doesn't seem to cause any problems.

See https://github.com/habitat-sh/habitat/issues/5467


thesentinels:

Thank you for contributing!

PR #5493, opened 2018-08-21
Relax rusoto version pins

baumanj:
The newer version of the rusoto crates depends on futures and changes the API a bit, so there are some minor changes, but they seem fairly straightforward.

If anyone has good suggestions of tests for exercising `pkg-export-docker`, that would be helpful.

### DIFF
--- a/components/pkg-export-docker/src/lib.rs
+++ b/components/pkg-export-docker/src/lib.rs
@@ -35,7 +35,6 @@ extern crate tempdir;
 extern crate url;
 
 extern crate failure;
-extern crate futures;
 #[macro_use]
 extern crate failure_derive;
 
@@ -59,7 +58,6 @@ use hcore::{channel, PROGRAM_NAME};
 
 use aws_creds::StaticProvider;
 use clap::App;
-use futures::future::Future;
 use rusoto_core::request::*;
 use rusoto_core::Region;
 use rusoto_ecr::{Ecr, EcrClient, GetAuthorizationTokenRequest};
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

baumanj:
Added `T-DO-NOT-MERGE` since I want to do a bit of testing before actually merging this.

christophermaier:
Looks like rusoto is used to push things to AWS ECR... guess if you can do that, you're good.

Paging @elliott-davis for more...

baumanj:
Good thing I did some testing (h/t @elliott-davis). The first cut at this called `wait` at the end of the future chain, but that resulted in
```
[42][default:/src:0]# hab pkg exec $HAB_ORIGIN/hab-pkg-export-docker hab-pkg-export-docker --push-image --registry-type amazon --registry-url "799338195663.dkr.ecr.us-east-2.amazonaws.com/jbauman/sample-node-app" --username "$AWS_ACCESS_KEY" --password "$AWS_SECRET_ACCESS_KEY" /src/results/jbauman-sample-node-app-1.0.1-20180822152320-x86_64-linux.hart
…
✗✗✗
✗✗✗ executor failed to spawn task
✗✗✗
```
Switching to calling `sync` on the first `Future`-returning function allows for handling of future returns in a non-async context.

components/pkg-export-docker/src/lib.rs
```diff
@@ -177,6 +175,7 @@ impl Credentials {
                 let auth_token_req = GetAuthorizationTokenRequest { registry_ids: None };
                 let token = client
                     .get_authorization_token(auth_token_req)
+                    .sync()
```
christophermaier: For posterity, it should be noted that `sync` is a function [provided by Rusoto's Future implementation](https://rusoto.github.io/rusoto/rusoto_core/struct.RusotoFuture.html#method.sync), and not something that is generally available. But for not having to use the `futures` crate, it works like a charm.

Nice find!

baumanj: Good to know!

PR #5503, opened 2018-08-24
Miscellaneous Butterfly refactors

christophermaier:
A handful of clarifying refactors for Butterfly, as well as a few minor bug fixes... read the commits for further details.

This is based on https://github.com/habitat-sh/habitat/pull/5506, and should be rebased on top of master once that merges.

### DIFF
--- a/components/butterfly/src/member.rs
+++ b/components/butterfly/src/member.rs
@@ -343,7 +343,7 @@ impl MemberList {
 
     /// Inserts a member into the member list with the given health.
     pub fn insert(&self, incoming_member: Member, incoming_health: Health) -> bool {
-        let mut accept_and_propagate_rumor: bool = false;
+        let mut accept_and_propagate_rumor = false;
 
         match self
             .members
--- a/components/butterfly/src/member.rs
+++ b/components/butterfly/src/member.rs
@@ -339,87 +339,87 @@ impl MemberList {
         }
     }
 
+    /// Inserts a member into the member list with the given health,
+    /// but only if the criteria for insertion are met. Returns `true`
+    /// if the rumor information was actually accepted, and `false`
+    /// otherwise.
+    ///
+    /// There are a few rules governing how we choose to accept
+    /// Membership rumors.
+    ///
+    /// First, if we have absolutely no record of having seen
+    /// `incoming_member` before, we'll accept the rumor without
+    /// question.
+    ///
+    /// On the other hand, if we _have_ seen `incoming_member` we need
+    /// to compare the incoming information to what we currently have
+    /// before we decide whether to accept the new information.
+    ///
+    /// If the incarnation number of the `incoming_member` is lower
+    /// than that of the rumor we already have, then we reject
+    /// it. Incarnation numbers for Members are only ever incremented
+    /// by that Member itself, so the fact that we already have one
+    /// that is higher means that we have more up-to-date information.
+    ///
+    /// Similarly, if the incoming incarnation number is greater than
+    /// what we have, we'll accept it as more up-to-date information.
+    ///
+    /// If the incarnation numbers match, we need to look at the
+    /// health to determine if we accept the rumor.
+    ///
+    /// We only accept the incoming rumor if its health is strictly
+    /// "worse than" the health we currently have for the member.
+    ///
+    /// Alternatively, you can think of "worse than" as "greater
+    /// than", given this ordering of Health states (this is governed
+    /// by the `PartialOrd` and `Ord` trait implementations on `Health`):
+    ///
+    ///     Alive < Suspect < Confirmed < Departed
+    ///
+    /// For example, if we think that "Supervisor X (at incarnation 1)
+    /// is Alive", but the rumor is telling us that "Supervisor X (at
+    /// incarnation 1) is Suspect", that means that whoever we've
+    /// received this rumor from is having trouble contacting
+    /// Supervisor X. We should accept this rumor and propagate it to
+    /// ensure that a) information about degraded connectivity makes
+    /// it around the network, and b) the odds that Supervisor X will
+    /// receive this rumor increase, allowing it to refute this (if
+    /// indeed Supervisor X is still around.)
+    ///
+    /// If we were to just accept the rumor regardless of what the
+    /// health was, we could basically start "arguing" across the
+    /// network; one Supervisor thinks X is Alive, another thinks it's
+    /// Suspect, and we just keep flip-flopping between the two
+    /// without any sort of resolution.
+    ///
+    /// Below is the truth table that illustrates this. "Current Health"
+    /// is down the left side, while "Incoming Health" is across the
+    /// top. We only propagate when Incoming is "worse than" Current:
+    ///
+    ///     |           | Alive | Suspect   | Confirmed | Departed  |
+    ///     |-----------+-------+-----------+-----------+-----------|
+    ///     | Alive     |       | propagate | propagate | propagate |
+    ///     | Suspect   |       |           | propagate | propagate |
+    ///     | Confirmed |       |           |           | propagate |
+    ///     | Departed  |       |           |           |           |
+    ///
     // TODO (CM): why don't we just insert a membership record here?
-
-    /// Inserts a member into the member list with the given health.
     pub fn insert(&self, incoming_member: Member, incoming_health: Health) -> bool {
-        let mut accept_and_propagate_rumor = false;
-
-        match self
+        let accept_and_propagate_rumor = match self
             .members
             .read()
             .expect("Member List read lock poisoned")
             .get(&incoming_member.id)
         {
-            None => {
-                // We haven't seen this member before, so we always
-                // accept the rumor
-                accept_and_propagate_rumor = true;
-            }
+            None => true,
             Some(existing_member) => {
-                // We have seen this member before; we need to examine
-                // it a bit more before we decide whether we want to
-                // accept it.
                 match existing_member
                     .incarnation
                     .cmp(&incoming_member.incarnation)
                 {
-                    cmp::Ordering::Greater => {
-                        // If the incarnation number we have is
-                        // greater than what we're being given, it is
-                        // more recent, so we reject the incoming rumor.
-                        accept_and_propagate_rumor = false;
-                    }
-                    cmp::Ordering::Less => {
-                        // If the incarnation number we have is less
-                        // than that of the incoming rumor, then we
-                        // must accept the new rumor. That's how the
-                        // member itself is updating its status to the
-                        // rest of the network.
-                        accept_and_propagate_rumor = true;
-                    }
+                    cmp::Ordering::Greater => false,
+                    cmp::Ordering::Less => true,
                     cmp::Ordering::Equal => {
-                        // If the incarnation numbers match, we need
-                        // to look at the health to determine if we
-                        // accept the rumor.
-                        //
-                        // We only accept the incoming rumor if its
-                        // health is strictly "worse than" the health
-                        // we currently have for the member.
-                        //
-                        // For example, if we think that "Supervisor X
-                        // (at incarnation 1) is Alive", but the rumor
-                        // is telling us that "Supervisor X (at
-                        // incarnation 1) is Suspect", that means that
-                        // whoever we've received this rumor from is
-                        // having trouble contacting Supervisor X. We
-                        // should accept this rumor and propagate it
-                        // to ensure that a) information about
-                        // degraded connectivity makes it around the
-                        // network, and b) the odds that Supervisor X
-                        // will receive this rumor increase, allowing
-                        // it to refute this (if indeed Supervisor X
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

raskchanky:
There's a lot here, and I'm new to this part of the code.  While nothing stands out as bad, and everything seems reasonable when I read through it, I also don't necessarily consider myself qualified at this point to say it's good to merge.

So instead, how about a gif?

components/butterfly/src/member.rs
```diff
@@ -678,58 +731,67 @@ impl MemberList {
...
+    /// Query the list of aging Suspect members to find those which have now
+    /// timed out to Confirmed. Health is updated appropriately, and a
+    /// list of newly-Confirmed Member IDs is returned.
+    pub fn get_newly_expired_confirmed(&self, timeout: Duration) -> Vec<String> {
+        let now = SteadyTime::now();
+        let mut newly_confirmed: Vec<String> = Vec::new();
+
         {
-            with_closure((id, suspect));
+            let mut aging_suspects = self
```
baumanj: I don't think this needs to be in a separate block; it seems to build without it for me. Also, you can do without the `aging_suspects` variable entirely:
```rust
        self.aging_suspects
            .write()
            .expect("aging_suspects lock is poisoned")
            .retain(|ref member_id, ref suspect_time| {
                let timed_out = now >= **suspect_time + timeout;
                if timed_out {
                    newly_confirmed.push(member_id.to_string());
                }
                !timed_out
            });
```
Also, and this is more my personal style of what I find easier to grok, you could eliminate the `timed_out` variable and write the inner block as:
```rust
                if now >= **suspect_time + timeout {
                    newly_confirmed.push(member_id.to_string());
                    false
                } else {
                    true
                }
```
This is mostly to eliminate the `!timed_out` return value. For me interpreting a negated return to `retain` is somewhat trickier than just seeing the two branches with explicit `true` and `false` returns. However this is more of a style thing, so take or leave it.

components/butterfly/src/member.rs
```diff
@@ -678,58 +731,67 @@ impl MemberList {
...
-        let mut depart = self
-            .depart
-            .write()
-            .expect("Departure list lock is poisoned");
-        depart.insert(member_id.to_string(), SteadyTime::now());
-    }
+    /// Query the list of aging Confirmed members to find those which have now
+    /// timed out to Departed. Health is updated appropriately, and a
+    /// list of newly-Departed Member IDs is returned.
+    pub fn get_newly_expired_departed(&self, timeout: Duration) -> Vec<String> {
```
baumanj: This seems essentially identical to `get_newly_expired_confirmed`. Why not parameterize the `HashMap` and make this one function?

components/butterfly/src/protocol/swim.rs
```diff
@@ -15,3 +15,48 @@
...
 pub use self::{membership::Health, swim::Payload as SwimPayload, swim::Type as SwimType};
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // These tests assure that we can properly compare Health values
+    // along the spectrum of
+    //
+    //   Alive -> Suspect -> Confirmed -> Departed
```
baumanj: Since `Health` implements the `PartialOrd` trait, isn't is enough to assert

1. Alive < Suspect
2. Suspect < Confirmed
3. Confirmed < Departed

?

These tests are certainly thorough, but if we wanted to add another state or change something, it seems to require more effort than it needs to.

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -233,9 +234,12 @@ impl DatFile {
 
     pub fn write(&self, server: &Server) -> Result<usize> {
         let mut header = Header::default();
-        let tmp_path = self
-            .path
-            .with_extension(thread_rng().gen_ascii_chars().take(8).collect::<String>());
+        let tmp_path = self.path.with_extension(
+            thread_rng()
+                .sample_iter(&Alphanumeric)
+                .take(8)
```
baumanj: As long as we're making a change here, can we avoid the magic number? Or perhaps replace this with something like [`mktemp`](https://crates.io/crates/mktemp)? We're already using it elsewhere.

components/butterfly/src/server/expire.rs
```diff
@@ -12,92 +12,57 @@
...
 
-use time::SteadyTime;
-
-use member::Health;
 use rumor::{RumorKey, RumorType};
 use server::timing::Timing;
 use server::Server;
-use trace::TraceKind;
+
+const LOOP_PERIOD_MS: u64 = 500;
```
baumanj: I think it would be more accurate to call this `LOOP_DELAY_MS` since it always sleeps this amount irrespective of how long the rest of the loop takes to run rather than having the loop's runtime subtracted from the sleep.

components/butterfly/src/server/mod.rs
```diff
@@ -513,7 +492,24 @@ impl Server {
...
+            ));
+
             let check_list = self.member_list.check_list(&member.id);
+
+            // TODO (CM): Even though we marked the rumor as hot
+            // above, when we gossip, we send out the 5 "coolest but
+            // still warm" rumors. Sending to 10 members increases the
+            // chances that we'll get to this hot one now, but I don't
+            // think that we can strictly guarantee that this
+            // departure health update actually gets out in all cases.
```
baumanj: Perhaps change this magic number to a (ideally tunable) constant?

components/butterfly/src/server/mod.rs
```diff
@@ -1111,8 +1107,8 @@ mod tests {
             let gossip_port = GOSSIP_PORT.fetch_add(1, Ordering::Relaxed);
             let gossip_listen = format!("127.0.0.1:{}", gossip_port);
             let mut member = Member::default();
-            member.swim_port = (swim_port as i32);
-            member.gossip_port = (gossip_port as i32);
+            member.swim_port = swim_port as i32;
+            member.gossip_port = gossip_port as i32;
```
baumanj: I wonder why these ports are `i32` instead of `u16`

components/butterfly/src/member.rs
```diff
@@ -330,114 +339,197 @@ impl MemberList {
...
-    pub fn insert(&self, member: Member, health: Health) -> bool {
-        let share_rumor: bool;
-        let mut start_suspicion: bool = false;
-        let mut stop_suspicion: bool = false;
-        let mut stop_departure: bool = false;
-
-        // If we have an existing member record..
-        if let Some(current_member) = self
+    pub fn insert(&self, incoming_member: Member, incoming_health: Health) -> bool {
+        let mut accept_and_propagate_rumor: bool = false;
```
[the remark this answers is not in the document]
christophermaier: Good catches on both!

components/butterfly/src/member.rs
```diff
@@ -330,114 +339,197 @@ impl MemberList {
...
+                        // | Alive     |       | propagate | propagate | propagate |
+                        // | Suspect   |       |           | propagate | propagate |
+                        // | Confirmed |       |           |           | propagate |
+                        // | Departed  |       |           |           |           |
+
+                        // We know we have a health if we have a
+                        // record.
+                        //
+                        // (Doing this in two operations is necessary
+                        // due to the lock and lifetimes.)
```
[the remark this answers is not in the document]
christophermaier: Yup!

components/butterfly/src/member.rs
```diff
@@ -330,114 +339,197 @@ impl MemberList {
...
+                        let health_lock = self.health.read().expect("Health lock is poisoned");
+                        let existing_health = health_lock.get(&existing_member.id).expect(
+                            "No health for a membership record should be impossible; did you use insert?",
+                        );
+
+                        if incoming_health > *existing_health {
```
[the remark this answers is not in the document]
christophermaier: Yeah, a little bit... my brain rebelled at first, too. Absent some protocol wrangling, or more complicated logic around converting protobuf data into Rust types, I think that's what we're left with, though.

I suppose we could rename `Health` to make it negative, to reflect "rising levels of concern" or something similar. Long term, I actually think that may be useful to consider, particularly once we start making better use of _Service_ health. Part of our issue of conflating Supervisor health and Service health could stem from the fact that we're talking about them both in terms of "health", when it means subtly different things depending on context.

components/butterfly/src/member.rs
```diff
@@ -330,114 +339,197 @@ impl MemberList {
...
     /// Inserts a member into the member list with the given health.
-    pub fn insert(&self, member: Member, health: Health) -> bool {
-        let share_rumor: bool;
-        let mut start_suspicion: bool = false;
-        let mut stop_suspicion: bool = false;
-        let mut stop_departure: bool = false;
-
-        // If we have an existing member record..
-        if let Some(current_member) = self
+    pub fn insert(&self, incoming_member: Member, incoming_health: Health) -> bool {
```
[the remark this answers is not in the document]
christophermaier: I'll add a note in the docs about this.

I really like having Enums in place of booleans (I'd kinda like to get rid of the use of booleans throughout large chunks of our code in favor of Enums, actually, just like I'd like to remove most uses of "stringly-typed" APIs in favor of real domain types). There's a bit more refactoring I'd like to do around this insertion logic (insert and store `Membership` rumors directly, collapse the `insert` and `insert_health_by_id` functions, rework the internal storage to be more consistent, etc.) Converting the return type to an Enum then (and cascading it outward to the rest of the code) would be a great addition.

Alternatively, there are also a number of other places in this kind of Butterfly code that return booleans (with similar semantics), so we could also consider Enum-ifying everything at once.

components/butterfly/src/member.rs
```diff
@@ -330,114 +339,197 @@ impl MemberList {
...
-                .insert(member.id.clone(), health);
-            if start_suspicion == true {
-                self.suspect
-                    .write()
-                    .expect("Suspect lock is poisoned")
-                    .insert(member.id.clone(), SteadyTime::now());
+                .expect("Member list lock is poisoned")
+                .insert(incoming_member.id.clone(), incoming_member);
+            let updated = self.insert_health_by_id(&member_id, incoming_health);
+            if !updated {
```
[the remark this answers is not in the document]
christophermaier: I wrestled with that. Though I think incrementing twice is probably OK, I didn't want to change too much here.

Also, I think with some additional refactoring, we can collapse `insert` and `insert_health_by_id` into a single function (which has a few follow-on consistency gains), at which point, we'll definitely have only a single place incrementing the counter.

components/butterfly/src/member.rs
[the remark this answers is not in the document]
baumanj: Health → HitPoints
😂

components/butterfly/src/member.rs
christophermaier: Good calls; looks like I was unnecessarily retaining some structure that was needed in prior versions of this refactoring.

I'll tweak the booleans, too. For some reason, the `timed_out` formulation read better to me initially, but now they're both fine. I was probably using this as a way to more thoroughly figure out / document what the code was doing during the various refactorings that led here. Now it's clear enough without `timed_out`, I think.

components/butterfly/src/rumor/dat_file.rs
baumanj: For example: https://github.com/habitat-sh/habitat/pull/5511

components/butterfly/src/protocol/swim.rs
christophermaier: Derp... yes.

components/butterfly/src/rumor/dat_file.rs
christophermaier: I see you've already addressed this in #5511

components/butterfly/src/server/expire.rs
christophermaier: Yup, good point.

components/butterfly/src/rumor/dat_file.rs
baumanj: Same idea, different place in the code. I'm holding back on merging that until after the release since it's not urgent.

components/butterfly/src/server/mod.rs
christophermaier: That is a fantastic point 🤔 

Our protobuf definitions define this as `int32`. There _is_ a `uint32` [Protobuf scalar type](https://developers.google.com/protocol-buffers/docs/proto#scalar), but no `u16`. We could probably tweak this a bit in our protobuf-to-Rust conversion.

In practice, these types get populated in the first place after being parsed by networking code, so they should be legal port numbers, but we could probably do some more work to ensure that at the type level.

components/butterfly/src/server/mod.rs
christophermaier: I'll extract it to a constant, but leave it fixed for now. I'll also make a follow-up issue to make this tunable, along with all the other currently-hard-coded parameters that affect our gossip subsystem.

PR #5501, opened 2018-08-24
Relax the version pins for the various `prost` crates

baumanj:
Specifically `prost`, `prost-types`, `prost-build` and `prost-derive`.

This makes the minor fix in components/butterfly/build.rs to support the new
version and updates Cargo.lock.

See https://github.com/habitat-sh/habitat/issues/5467

PR #5506, opened 2018-08-25
Update sup-protocol protocol generation process

christophermaier:
Looks like these tweaks got left off #5501 accidentally.

(The new version of Prost automatically derives `Hash` for Enums, and
apparently none of the other types needed it.)

### DIFF
--- a/components/sup-protocol/build.rs
+++ b/components/sup-protocol/build.rs
@@ -26,7 +26,7 @@ fn main() {
 
 fn generate_protocols() {
     let mut config = prost_build::Config::new();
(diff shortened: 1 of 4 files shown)


christophermaier:
@baumanj Indeed... at first, I suspected we were missing something in `.bldr.toml` to link `sup-protocol` with `sup` (knowing "when to build something" in Travis builds is driven from this), but that appears to be wired up properly. 🤔

baumanj:
The issues raised in 
https://github.com/habitat-sh/habitat/pull/5506#pullrequestreview-149724233 will be addressed in a separate issue. The changes here are good.

PR #5513, opened 2018-08-28
implement binlink for windows

### LINKED ISSUE #2617: hab pkg binlink support on windows
It might be useful enough, because updating the PATH environment variable without rebooting is not a trivial task

mwrock:
fixes #2617 

### DIFF
--- a/components/hab/src/command/cli/setup.rs
+++ b/components/hab/src/command/cli/setup.rs
@@ -12,11 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
 use std::path::Path;
+
+#[cfg(windows)]
+use std::ptr;
+#[cfg(windows)]
+use user32;
+#[cfg(windows)]
+use widestring::WideCString;
+#[cfg(windows)]
+use winapi;
+#[cfg(windows)]
+use winreg::enums::{HKEY_LOCAL_MACHINE, KEY_ALL_ACCESS, KEY_READ};
+#[cfg(windows)]
+use winreg::RegKey;
 
 use common::ui::{UIReader, UIWriter, UI};
 use hcore::crypto::SigKeyPair;
-use hcore::env;
+use hcore::env as henv;
 use hcore::package::ident;
 use hcore::Error::InvalidOrigin;
 
@@ -26,7 +40,12 @@ use config;
 use error::Result;
 use {AUTH_TOKEN_ENVVAR, CTL_SECRET_ENVVAR, ORIGIN_ENVVAR};
 
-pub fn start(ui: &mut UI, cache_path: &Path, analytics_path: &Path) -> Result<()> {
+pub fn start(
+    ui: &mut UI,
+    cache_path: &Path,
+    analytics_path: &Path,
+    binlink_path: &Path,
+) -> Result<()> {
     let mut generated_origin = false;
 
     ui.br()?;
@@ -146,6 +165,27 @@ pub fn start(ui: &mut UI, cache_path: &Path, analytics_path: &Path) -> Result<()
     } else {
         ui.para("Okay, maybe another time.")?;
     }
+    if cfg!(windows) {
+        ui.heading("Habitat Binlink Path")?;
+        ui.para(
+            "The `hab` command-line tool can create binlinks for package binaries \
+             in the 'PATH' when using the 'pkg binlink' or 'pkg install --binlink' \
+             commands. By default, Habitat will create these binlinks in the '/hab/bin' \
+             directory. This directory will always be included in your 'PATH' when \
+             inside a Studio environment. However, you will want this directory on your \
+             machine's persistent 'PATH' in order to access binlinks outside of a Studio.",
+        )?;
+        if ui.prompt_yes_no("Add binlink directory to PATH?", Some(true))? {
+            set_binlink_path(&binlink_path)?;
+            ui.para(&format!(
+                "{} has been added to your path. You will need to open a \
+                 new console window for this added entry to take effect.",
+                binlink_path.display()
+            ))?;
+        } else {
+            ui.para("Okay, maybe another time.")?;
+        }
+    }
     ui.heading("Analytics")?;
     ui.para(
         "The `hab` command-line tool will optionally send anonymous usage data to \
@@ -237,7 +277,7 @@ fn prompt_origin(ui: &mut UI) -> Result<String> {
             ))?;
             Some(o)
         }
-        None => env::var(ORIGIN_ENVVAR).or(env::var("USER")).ok(),
+        None => henv::var(ORIGIN_ENVVAR).or(henv::var("USER")).ok(),
     };
     Ok(ui.prompt_ask("Default origin name", default.as_ref().map(|x| &**x))?)
 }
@@ -266,7 +306,7 @@ fn prompt_auth_token(ui: &mut UI) -> Result<String> {
             )?;
             Some(o)
         }
-        None => env::var(AUTH_TOKEN_ENVVAR).ok(),
+        None => henv::var(AUTH_TOKEN_ENVVAR).ok(),
     };
     Ok(ui.prompt_ask(
         "Habitat personal access token",
@@ -284,7 +324,7 @@ fn prompt_ctl_secret(ui: &mut UI) -> Result<String> {
             )?;
             Some(o)
         }
-        None => env::var(CTL_SECRET_ENVVAR).ok(),
+        None => henv::var(CTL_SECRET_ENVVAR).ok(),
     };
     Ok(ui.prompt_ask(
         "Habitat Supervisor CtlGateway secret",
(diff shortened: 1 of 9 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/cli/setup.rs
```diff
@@ -146,6 +167,27 @@ pub fn start(ui: &mut UI, cache_path: &Path, analytics_path: &Path) -> Result<()
...
     }
+    if cfg!(target_os = "windows") {
+        ui.heading("Habitat Binlink Path")?;
+        ui.para(
+            "The `hab` command-line tool can create binlinks for package binaries  \
+             in the 'PATH' when using the 'pkg binlink' or 'pkg install --binlink' \
+             commands. By default, Habitat will create these binlinks in the '/hab/bin' \
+             directory. This directory will always be included in your 'PATH' when \
+             inside a Studio environment. However, you will want this directory on your \
+             machine's persistent 'PATH' in order to access binlinks outside of a Studio.",
```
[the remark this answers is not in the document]
mwrock: the `ask_set_binlink_path` function below actually serves this purpose. The convention we have been following in this file is to explain the scenario in the `para` blocks and then ask directly in a `prompt_yes_no` question. So after this text the user is given a yes/no question: `Add binlink directory to PATH?`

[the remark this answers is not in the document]
apriofrost: I see. So it would look like the following, right?

- Habitat Binlink Path
  - [this paragraph]
- Add binlink directory to PATH? [Yes/no/quit] (if type 'yes')
  - Enter your Binlink directory.
- Binlink directory: _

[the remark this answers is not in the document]
mwrock: Almost. If `yes` is entered, we add the default binlink directory `c:\hab\bin` to the path and output:
```
c:\hab\bin has been added to your path. You will need to open a new console window for this added entry to take effect.
```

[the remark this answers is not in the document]
apriofrost: I see. Looks good! Thanks for explaining this to me 🤓

components/hab/src/command/pkg/binlink.rs
```diff
@@ -106,6 +118,9 @@ where
             if bin_file.file_type()?.is_dir() {
                 continue;
             }
+            if cfg!(target_os = "windows") && !is_executable(bin_file.path().extension()) {
+                continue;
+            }
```
baumanj: I found the `is_executable` function hard to understand and a little misleading since it's not marked `#[cfg(windows)]`, but it only *works* for windows. I understand it's necessary because it's a separate function, so otherwise you'd have to provide a non-windows implementation, but how about side-stepping the whole issue by doing the check inline:
```rust
            if cfg!(windows) {
                if let Some(executable_extensions) = env::var_os("PATHEXT") {
                    if env::split_paths(&executable_extensions)
                        .any(|e| extensions_match(Path::new(&e), &bin_file.path()))
                    {
                        continue;
                    }
                }
            }
```
Then the factored-out `extensions_match` is platform-agnostic:
```rust
fn extensions_match(p1: &Path, p2: &Path) -> bool {
    match (p1.extension(), p2.extension()) {
        (Some(ext1), Some(ext2)) => {
            ext1.to_string_lossy().to_uppercase() == ext2.to_string_lossy().to_uppercase()
        }
        _ => false,
    }
}
```

components/hab/src/command/cli/setup.rs
```diff
@@ -311,3 +353,66 @@ fn opt_out_analytics(ui: &mut UI, analytics_path: &Path) -> Result<()> {
...
+                ptr::null_mut(),
+            );
+        }
+    }
+    Ok(())
+}
+
+#[cfg(not(target_os = "windows"))]
+fn set_binlink_path(binlink_path: &Path) -> Result<()> {
+    Ok(())
```
baumanj: I think [`unreachable!()`](http://doc.rust-lang.org/1.28.0/core/macro.unreachable.html) may be more appropriate here, since I think this would be a logic error if we called this from a non-windows context.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -126,9 +141,105 @@ where
...
+}
+
+#[cfg(not(target_os = "windows"))]
+impl Binlink {
+    pub fn new<T: AsRef<Path>>(src: T, dest_dir: T) -> Self {
+        Binlink {
+            path: dest_dir.as_ref().join(
+                src.as_ref()
+                    .file_name()
+                    .expect("Binlink source path does not resolve to a file"),
```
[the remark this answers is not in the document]
mwrock: I figured the `expect` was ok here since `src` is coming from `find_command_in_pkg` which would never return a non-file. Do you think there is still merit in a `result`?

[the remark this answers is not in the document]
baumanj: I agree that I don't see a way that it could happen now, but there's no guarantee based on `Binlink`'s own handling of the source input that it won't end in `..`. A new caller or a change in `find_command_in_pkg` could cause a panic that I think would better be avoided.

components/hab/src/command/cli/setup.rs
```diff
@@ -311,3 +351,60 @@ fn opt_out_analytics(ui: &mut UI, analytics_path: &Path) -> Result<()> {
...
+
+/// this sets the permanent machine PATH and not
+/// the path of this process. These are maintained
+/// in the Windows registry
+#[cfg(windows)]
+fn set_binlink_path(binlink_path: &Path) -> Result<()> {
+    if !binlink_is_on_path(binlink_path) {
+        let hklm = RegKey::predef(HKEY_LOCAL_MACHINE);
+        let env = hklm.open_subkey_with_flags(
+            r"System\CurrentControlSet\Control\Session Manager\Environment",
```
baumanj: I didn't mean to say with 
https://github.com/habitat-sh/habitat/pull/5513#discussion_r213380316 that you had to change this, I just thought it was useful to know for a windows context.

However, whichever way you go, I would like it to be consistent with the  path on L358.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -126,9 +154,96 @@ where
...
+    pub fn from_file<T: AsRef<Path>>(path: T) -> Result<Self> {
+        use toml::Value::Table;
+
+        let file = File::open(&path)?;
+        for line in BufReader::new(file).lines() {
+            let ln = line?;
+            if ln.to_uppercase().starts_with(BAT_COMMENT_MARKER) {
+                let (_, rest) = ln.split_at(BAT_COMMENT_MARKER.len());
+                if let Ok(Table(toml_exp)) = rest.parse() {
+                    if let Some(src) = toml_exp["source"].as_str() {
```
baumanj: Do we want to panic here if the the LHS of the assignment expression isn't `source`?

If not, we could use something like
```rust
            if let Some(src) = toml_exp.get("source") {
```

components/hab/src/command/pkg/binlink.rs
```diff
@@ -126,9 +154,96 @@ where
...
+        }
+        Err(Error::CannotParseBinlinkSource(path.as_ref().to_path_buf()))
+    }
+
+    pub fn link(&self) -> Result<()> {
+        let template = format!(
+            "@echo off\nREM source='{0}'\n\"{0}\" %*",
+            self.src.display()
+        );
+        Ok(fs::write(&self.dest, template)?)
```
baumanj: If you prefer `Ok(expression?)` over `expression.map_err(Into::into)`, that's fine with me, but either way, I think the windows and not(windows) versions should be consistent.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -106,6 +118,22 @@ where
             if bin_file.file_type()?.is_dir() {
                 continue;
             }
+            if cfg!(target_os = "windows") {
+                match bin_file.path().extension() {
+                    Some(ext) => match env::var_os("PATHEXT") {
```
baumanj: I think it's hard to infer the meaning of this code without already knowing what `PATHEXT` signifies. Perhaps instead of `ext`, calling it `executable_extensions` or something like that would help.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -106,6 +118,22 @@ where
...
+                        Some(val) => {
+                            if !env::split_paths(&val).any(|e| {
+                                e.to_string_lossy()
```
baumanj: Is it a guarantee that the values from `PATHEXT` will be uppercase?

components/hab/src/command/pkg/binlink.rs
mwrock: oh i missed that the indexer could panic

components/hab/src/command/pkg/binlink.rs
mwrock: I personally prefer (not super strongly) the `?` over `map_err`.

components/hab/src/command/pkg/binlink.rs
mwrock: that is how windows will render it but there is no stopping someone from changing it. I'll force it to uppercase just in case.

PR #5515, opened 2018-08-28
Don't require `protoc` in package builds

christophermaier:
As things currently stand, we shouldn't actually need a `protoc`
binary in the `core/hab-sup` package build because we shouldn't
actually be generating protobuf code in those builds.

This was done in
https://github.com/habitat-sh/habitat/pull/5179/commits/b167369c5ac4a69ef1e0f7731076f28e6de9d0c0,
but it turns out that we actually had misconfigured our
`habitat_butterfly` dependency. We should not have built with any
default features (of which protocol generation is one), and we
shouldn't have duplicated `habitat_butterfly` in our
dev-dependencies.

(I'm unsure why, but both of these changes seem to be required to
successfully build without a `protoc` binary. In any event,
duplicating the dev-dependency doesn't make sense and should be
removed anyway.)

### DIFF
--- a/components/sup/Cargo.toml
+++ b/components/sup/Cargo.toml
@@ -24,7 +24,7 @@ features = "*"
 futures = "*"
 glob = "*"
 hab = { path = "../hab" }
-habitat_butterfly = { path = "../butterfly" }
+habitat_butterfly = { path = "../butterfly", default-features = false }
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

PR #5517, opened 2018-08-28
Fix composite bindings

### LINKED ISSUE #5448: Binds fail for composite package
We're currently running into this issue on hab v0.59.0, when we load the same package using v0.55.0 everything works as expected.

We're running linux containers on windows 10, so the studio we're entering and executing from is linux on windows.

```
pkg_services=(
  aem/aem-author
  aem/AEM-6_4_1_0-6_4_1
)

pkg_bind_map=(
  [aem/AEM-6_4_1_0-6_4_1]="aem-svr:aem/aem-author"
)
```

The supervisor output is as follows:

```
hab-sup(AG): The aem/AEM-6_4_1_0-6_4_1 service was successfully loaded
hab-sup(AG): The aem/aem-author service was successfully loaded
...
hab-sup(MR): Starting aem/AEM-6_4_1_0-6_4_1
hab-sup(MR): Unable to start aem/AEM-6_4_1_0-6_4_1, hab-sup(SS)[components/sup/src/manager/service/spec.rs:358:24]: Missing required bind(s), aem-svr
```

Looking in `/hab/sup/default/specs/AEM-6_4_1_0-6_4_1.spec` the binds array is completely empty.

```
ident = "aem/AEM-6_4_1_0-6_4_1"
group = "default"
channel = "stable"
topology = "standalone"
update_strategy = "none"
binds = []
binding_mode = "strict"
desired_state = "up"
composite = "aem-author-composite"
```

christophermaier:
Fixes #5448 

As is, this fixes the missing binds issue. Absent some additional work around the control flow, however, it will _not_ warn if you try to override any composite binds with a "two-part" bind, or if you try to specify a "three-part" bind for a standalone service. Incorrectly-formatted binds would not get applied erroneously (which is good); they would just be silently discarded (which is not so good).

Opening this for review now, while I investigate ways to fix the validation.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -1133,22 +1133,27 @@ impl Manager {
         opts: protocol::ctl::SvcUnload,
     ) -> NetResult<()> {
         let ident: PackageIdent = opts.ident.ok_or(err_update_client())?.into();
-        // Gather up the paths to all the spec files we care about. This
-        // includes all service specs as well as any composite spec.
-        let spec_paths = match Self::existing_specs_for_ident(&mgr.cfg, &ident)? {
-            Some(Spec::Service(spec)) => vec![Self::spec_path_for(&mgr.cfg, &spec)],
+        // Gather up the paths to all the spec files we care about,
+        // along with their corresponding idents (we do this to ensure
+        // we emit a proper "unloading X" message for each member of a
+        // composite).
+        //
+        // This includes all service specs as well as any composite
+        // spec.
+        let path_ident_pairs = match Self::existing_specs_for_ident(&mgr.cfg, &ident)? {
+            Some(Spec::Service(spec)) => vec![(Self::spec_path_for(&mgr.cfg, &spec), ident)],
             Some(Spec::Composite(composite_spec, specs)) => {
                 let mut paths = Vec::with_capacity(specs.len() + 1);
                 for spec in specs.iter() {
-                    paths.push(Self::spec_path_for(&mgr.cfg, spec));
+                    paths.push((Self::spec_path_for(&mgr.cfg, spec), spec.ident.clone()));
                 }
-                paths.push(Self::composite_path_for(&mgr.cfg, &composite_spec));
+                paths.push((Self::composite_path_for(&mgr.cfg, &composite_spec), ident));
                 paths
             }
             None => vec![],
         };
 
-        for file in spec_paths {
+        for (file, ident) in path_ident_pairs {
             if let Err(err) = std::fs::remove_file(&file) {
                 return Err(net::err(
                     ErrCode::Internal,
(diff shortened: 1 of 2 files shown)


components/sup/src/manager/mod.rs
```diff
@@ -1133,22 +1133,27 @@ impl Manager {
...
                 }
-                paths.push(Self::composite_path_for(&mgr.cfg, &composite_spec));
+                paths.push((Self::composite_path_for(&mgr.cfg, &composite_spec), ident));
                 paths
             }
             None => vec![],
         };
 
-        for file in spec_paths {
+        for (file, ident) in path_ident_pairs {
```
baumanj: It's kind of confusing that this works by shadowing the `ident` var from L1135. I know that shadowing is more idiomatic is rust than most other languages, but in this case the shadowed var is far enough away that it took me awhile to figure out.

It's not a huge deal, but perhaps renaming one or the other would make things a bit clearer.

components/sup/src/manager/service/spec.rs
```diff
@@ -192,21 +192,25 @@ impl IntoServiceSpec for protocol::ctl::SvcLoad {
...
 
-        let composite_binds = if let Some(ref list) = self.binds {
-            let binds: Vec<ServiceBind> = list.binds.clone().into_iter().map(Into::into).collect();
-            let (composite, _) = binds.into_iter().partition(|ref bind| bind.is_composite());
-            Some(composite)
-        } else {
-            None
+        // We permit the selective overriding of binds with CLI
+        // arguments; if any such overrides are present, this will
+        // pull them all out.
```
baumanj: This seems kind of contrary to the doc comment on L170:
```
     ///   (We do not set binds here, though, because that requires
     ///   specialized, service-specific handling.)
```

components/sup/src/manager/service/spec.rs
```diff
@@ -418,6 +422,7 @@ impl FromStr for ServiceSpec {
 pub struct ServiceBind {
     pub name: String,
     pub service_group: ServiceGroup,
+    /// Only set if this is a bind targeting a composite service
```
baumanj: Maybe call it `composite_service_name`, then?

baumanj: Or even better, change `ServiceBind` to make the difference explicit:
```rust
pub enum ServiceBind {
    Simple {
        name: String,
        service_group: ServiceGroup,
    },
    Composite {
        name: String,
        service_group: ServiceGroup,
        service_name: String,
    },
}
```
or
```rust
pub enum ServiceBindServiceName {
    Simple {},
    Composite { service_name: String },
}

pub struct ServiceBind {
    pub name: String,
    pub service_group: ServiceGroup,
    pub service_name: ServiceBindServiceName,
}
```
But that may be too disruptive at the current moment

christophermaier: This is more or less exactly what I'd like to do, but didn't want to tackle it at this point because this is getting rather close to the protobuf representations, and I think we need to take another look at how we're doing the protobuf-type-to-Habitat-internal-Rust-type conversion there. I think we ultimately want to do something like we're doing with Butterfly, but that's gonna be it's own dedicated change.

PR #5528, opened 2018-08-30
ensure pkg build and appveyor hard fails if any powershell error occurs

### LINKED ISSUE #5273: Set ErrorActionPreference to `Stop` in plan-build-ps1
We want ANY error to fail the build. I don't think this will have any ill effect but want to test it.

mwrock:
fixes #5273 

This makes sure than any powershell error in both studio logic and the appveyor script result in a "Terminating" exception. That means it will halt execution where some powershell errors are merely reported to the console but processing continues. In the context of setting up a studio or our appveyor build, any error whatsoever should be considered a failure. Note that plan-build already had this set.

Also we always want `hab pkg build` to emit a positive exit code if such a failure occurs. This will cause our appveyor script (or anyone else's habitat based CI) to fail when it checks the result of `hab pkg build`.

### DIFF
--- a/components/studio/bin/hab-studio.ps1
+++ b/components/studio/bin/hab-studio.ps1
@@ -383,6 +383,10 @@ function Invoke-StudioRun($cmd) {
 }
 
 function Invoke-StudioBuild($location, $reuse) {
+  # This trap will cause powershell to return an exit code of 1
+  hab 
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

PR #5531, opened 2018-08-31
fix linux resolution of FS_ROOT_ENVVAR in hab commands and the splitting and joining of PATH dirs in setup

mwrock:
This fixes a regression introduced in the Windows binlinking pr (#5513) by reverting the `FS_ROOT` static variable ommision to the `hab` `main.rs`.

The original implementatiion of `FS_ROOT` here at first lance seemed like a duplication of `FS_ROOT_PATH` in https://github.com/habitat-sh/core/blob/2cd4b0c03e2b2ce6238fded35d1738ae41d8d94f/components/core/src/fs.rs#L59. For Windows, the `core` implementation is superior because it properly roots the fs_root into the `systemdrive` - usually `c:\`. So to reduce repetition and to get the desired Windows behavior, I swaped in the `core` implementation.

The key problem here is that the `core` implementation for Linux does not ever draw from `FS_ROOT_ENVVAR` which causes havoc in a chrooted studio.

Now I believe that the truly correct way forward is to make the `core` implementation take `FS_ROOT_ENVVAR` into account for linux because as it stands we treat this foundational variable (FS_ROOT) in different ways in these 2 areas - hab and core. I am holding off on that change to discuss with @fnichol since there were some strong concerns he had to honoring `FS_ROOT_ENVVAR`.

So this PR at least gets us over the hump and returns the `fs_root` resolution in hab commands to what it was and adds root drive resolution for Windows.

### DIFF
--- a/components/hab/src/command/cli/setup.rs
+++ b/components/hab/src/command/cli/setup.rs
@@ -379,12 +379,10 @@ fn set_binlink_path(binlink_path: &Path) -> Result<()> {
             r"System\CurrentControlSet\Control\Session Manager\Environment",
             KEY_ALL_ACCESS,
         )?;
+        let mut paths = vec![binlink_path.to_path_buf()];
         let path: String = env.get_value("path")?;
-        let new_paths = [binlink_path, Path::new(&path)];
-        env.set_value(
-            "path",
-            &env::join_paths(new_paths.iter())?.to_str().unwrap(),
-        )?;
+        paths.append(&mut env::split_paths(&path).collect());
+        env.set_value("path", &env::join_paths(paths)?.to_str().unwrap())?;
 
         // After altering the machine environment, we must broadcast
         // a WM_SETTINGCHANGE message to all windows so the user
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -58,7 +58,7 @@ use hcore::crypto::dpapi::encrypt;
 use hcore::crypto::keys::PairType;
 use hcore::crypto::{default_cache_key_path, init, BoxKeyPair, SigKeyPair};
 use hcore::env as henv;
-use hcore::fs::{cache_analytics_path, cache_artifact_path, cache_key_path, FS_ROOT_PATH};
+use hcore::fs::{cache_analytics_path, cache_artifact_path, cache_key_path};
 use hcore::package::PackageIdent;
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
I found another bug introduced in the process of addressing comments which was breaking the formation of the PATH variable. This propperly splits the dirs, appends them after the new binlink dir and then rejoins them.

components/hab/src/command/cli/setup.rs
```diff
@@ -379,12 +379,10 @@ fn set_binlink_path(binlink_path: &Path) -> Result<()> {
...
         )?;
+        let mut paths = vec![binlink_path.to_path_buf()];
         let path: String = env.get_value("path")?;
-        let new_paths = [binlink_path, Path::new(&path)];
-        env.set_value(
-            "path",
-            &env::join_paths(new_paths.iter())?.to_str().unwrap(),
-        )?;
+        paths.append(&mut env::split_paths(&path).collect());
+        env.set_value("path", &env::join_paths(paths)?.to_str().unwrap())?;
```
baumanj: It took me a bit to get my head around this. Why not start with the vector containing the paths from the environment and add the binlink path to the front? Something like:
```rust
        let mut paths: Vec<PathBuf> = env::split_paths(&path).collect();
        paths.insert(0, Path::new("/bin/link").to_path_buf());
```

components/hab/src/main.rs
```diff
@@ -95,6 +96,30 @@ lazy_static! {
...
         ]
     };
+
+    /// The default filesystem root path to base all commands from. This is lazily generated on
+    /// first call and reflects on the presence and value of the environment variable keyed as
+    /// `FS_ROOT_ENVVAR`.
+    static ref FS_ROOT: PathBuf = {
+        use hcore::fs::FS_ROOT_ENVVAR;
+
+        if cfg!(target_os = "windows") {
```
baumanj: It feels like a minor thing, but I'm sure when people see `cfg!(windows)` in some places and `cfg!(target_os = "windows")` in others, it's going to cause some people confusion.

baumanj: Also, I know this is copied from core, so if we have a high degree of confidence it'll be removed soon, I'm not opposed to leaving it as is.

PR #5546, opened 2018-09-04
Spanning ring

### LINKED ISSUE #4211: Allow a ring that spans a network over NAT
- network abstraction
  - [x] replace direct use of UDP and zeromq sockets and socket addresses with abstract SWIM and gossip message senders and receivers
  - [x] implement an abstraction for real-world usage (with UDP and zeromq sockets using SocketAddr as addresses)
  - [x] use the real-world abstraction in supervisor and tests
- test infrastructure for spanning ring
  - [x] channel-based network abstraction for testing purposes
  - [x] NAT simulation
    - set up simulated private networks, expose addresses outside the private network, start Butterfly servers with a given address, route/block the messages sent between the servers
  - [ ] generating temporary addresses on the NAT boundary
    - mostly for testing if supervisors don't try to talk other supervisors without being exposed
  - [ ] customizing NAT features (ban hairpinning, port- or address-mapping stability)
    - rather low-priority, to test various types of NATs
- spanning ring
  - [x] zone concept
    - it's a supervisor's view of the private networks hierarchy
  - [x] send zone rumors
  - [x] zone merging, zone negotiation within a single private network, basic zone negotiation across private networks, tests
    - the three items are lumped into one as they all are quite related and touch the same code
    - tests are done
    - got the first two items to work, working on the third broke the first two
    - largest item in the work
  - [ ] add an "indirect" health status
    - can be done a bit later, not needed for the demo
  - [ ] zone negotiation across private networks with hidden supervisors
    - can be done a bit later, not needed for the demo
  - [ ] set up additional addresses and ports
    - for butterfly servers:
      - these are addresses of a supervisor that is exposed
      - the address is optional, it can be discovered, ports are not - they need to be specified
      - directly from command line
      - from a watched file
    - for services
      - a bit similar to additional addresses for butterfly servers
      - each port needs to be named like the variable in the exposed configuration it relates to
    - IN PROGRESS
  - [ ] update configuration template helpers
    - likely sys.ip should return an IP that the service in a specific zone can actually use to talk to the other service.
  - [ ] add sibling zones
    - to allow a scenario where we have two k8s clusters, each are running some services managed by supervisors and those are exposed and then bound to each other
  - [ ] add validation for incoming zone and member information
    - make sure that UUIDs sent as strings are actually valid
  - [ ] handle the case when the maintainer of the current zone disappears

Old old stuff below:

These notes are still work in progress. I hope to get to some concrete steps to be done soon.

Currently it is not possible to have a ring that contains supervisors talking to each other over NAT.

- supervisor behind NAT needs to know which address it got on the NAT boundary
  - so it can send a proper information in the Ping message.

- questions:
  - do we want the supervisors behind the NAT to only use the exposed IP for communication?
    - if yes, that would assume that all the supervisors behind NAT need to be exposed
      - also, if we have NAT behind another NAT, then all the supervisors needs to be exposed over the uppermost NAT and use that IP in the ring communication.
    - if not, we need a way of partitioning the ring into zones (where each zone only contains supervisors that can talk directly in both ways) and setting up some supervisors in the zone to act as a forwarder when supervisors are communicating over the NAT boundary.
      - this is quite involved and I was spending my time mostly on this scenario, as exposing all the supervisors may not scale.
      - this will certainly mean some changes in the swim/gossip messages, hopefully backward compatible.
      - what may not be possible is to extend an already existing ring with older supervisors so it starts spanning the NAT boundary.
        - might be a bit tricky to detect that…

krnowak:
This work has three areas:

- Zone negotiation.
  - Zone is supervisor's view of a network. So network inside the k8s cluster would be one zone and network outside it could be another zone.
  - Zones currently are organized in a tree-like hierarchy.
  - Normally a service in a child zone can initiate a communication with a service in parent zone, but the converse is false.
  - This might be a simplification, but I think it catches most of the networking cases.
    - This likely is not true in case of local k8s cluster created by minikube, though - the networking minikube sets up is… complicated. There are some hacks to make this scenario work, but I would like to drop them.
- Exposing supervisor (through extended member info in protocol messages) and its services (through extended service gossip messages). This basically is:
  - Watching a TOML file, where we describe how the supervisor and the services it manages are exposed.
  - Applying that information to our member information and to our service information.
    - Member information is usually sent as a part of either gossip messages or SWIM messages.
    - Service information is usually sent as generated service rumors.
- Consuming the extended member and gossip messages to properly bind services across zones.
  - This drops all the service information from census if the service is not reachable (like if it is running inside the k8s cluster and is not exposed outside).

Some things that could be done next:
- Drop address discovery.
  - While this is convenient to have, it may interfere with next point. Also, address discovery in some places looks like heuristics. Would still like to do it. This is something that probably could be done later.
- Add siblings to zones.
  - So zones are in parent-child relationships, but there could be a situation where we two k8s clusters with some supervisors exposed and there is no supervisor we can talk to outside the clusters.
- Add detection if two separate children of a parent zone are actually the same zone
  - That would likely be a thread where we send pings to supervisors in the sibling zone through their private IP to see if we get an ACK. If so, the zones are not siblings but aliases.
- Drop predecessors over time.
  - During zone negotiation we may end up with many aliases for a single zone. Would be nice to phase them out gradually to keep only the zone with the highest ID.
- Limit the number of zone aliases and zone children sent in pings (to avoid breaking the 4k bytes limit in some cases).
- Better debug story for zones code.
  - The debug was done the way it is, because I wanted to have a readable multiline debug messages from a server when running a test that has several servers running and printing messages at the same time.

Fixes #4211

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -783,7 +783,6 @@ dependencies = [
  "habitat_butterfly_test 0.1.0",
  "habitat_core 0.0.0 (git+https://github.com/habitat-sh/core.git)",
  "heck 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "lazy_static 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "pkg-config 0.3.14 (registry+https://github.com/rust-lang/crates.io-index)",
  "prost 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,6 +178,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "cast"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "cc"
 version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -772,6 +777,7 @@ version = "0.1.0"
 dependencies = [
  "byteorder 1.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "bytes 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cast 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "clippy 0.0.212 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.5.13 (registry+https://github.com/rust-lang/crates.io-index)",
  "habitat_butterfly_test 0.1.0",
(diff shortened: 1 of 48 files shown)


thesentinels:

Thank you for contributing!

baumanj:
Looks like there are some test failures. Let me know if you need any travis help there, otherwise I'll plan to take a look when those are resolved. Does that sound good, @krnowak ?

krnowak:
@baumanj: I updated the PR with four additional commits that should address a failure I was sometimes getting. But I still sometimes am getting failures on the `nat::nat::different_zones_get_different_ids_with_unexposed_servers` test - two servers don't see each other as healthy, but I'm not sure why, yet. Seems to be easier to trigger when other tests are running in parallel, so I don't know if I'm getting some starvation here…

Also, I'm sorry for the lousy pull request message, will update it tomorrow.

krnowak:
Added a bunch of commits to fix some intermittent failures I was getting. I also dropped the discovery feature from the code, so the supervisors needs to be told explicitly about its exposed address. Also rebased to current master branch.

krnowak:
Rebased to current master and refactored zone handling a bit. Added some unit tests for that.

baumanj:
I'm going to be out for a bit, so @christophermaier is going to take over reviewing this. Thanks for your patience.

baumanj:
This isn't under active review, so closing for now.

PR #5511, opened 2018-08-27
Replace deprecated rand usage with mktemp

baumanj:
A similar change can be made to address [the issue I noted here](https://github.com/habitat-sh/habitat/pull/5503#discussion_r213125144) as well as a couple other places.

PR #5562, opened 2018-09-05
Replace deprecated usage of rand crate with mktemp

baumanj:
This replaces https://github.com/habitat-sh/habitat/pull/5511 since that one got messy with the intervening changes to Cargo.lock.

This is more concise and has the benefit of cleaning up afterwards if there is
an error before the temporary file is renamed.

Also, since we implemented io:Error -> error:Error::IO, simplify some of the
lines that used map_err previously.


thesentinels:

Thank you for contributing!

PR #5573, opened 2018-09-05
Create root's $HOME, /root

### LINKED ISSUE #5569: New studio does not have a /root directory
The `$HOME` for the `root` user is `/root` (as it should be) but that directory doesn't exist in the studio. We observed this with habitat version 0.62.1:

```
% hab -V
hab 0.62.1/20180905002018
% hab studio enter

[1][default:/src:0]# echo $HOME
/root
[2][default:/src:0]# ls /root
ls: cannot access '/root': No such file or directory
```

It seems that this may be related to the "rootless studio" work that was done recently in version 0.58.0 (ish) of Habitat.

This manifests itself when using `scaffolding-ruby` for our Rails application build. The scaffolding relies on parsing the output of `bundle --version` in a couple places. It uses this output to display nice messages during build time. The scaffolding script also sets a `_bundler_version` variable that gets set by parsing the output, in the `_setup_vars` function. Excerpt:

```
  _bundler_version="$("$(pkg_path_for bundler)/bin/bundle" --version \
    | awk '{print $NF}')"
```

This is used to vendor the `bundler` gem itself with the following command (excerpt):

```
   gem install \
     --local "$(pkg_path_for bundler)/cache/bundler-${_bundler_version}.gem" \
     --install-dir "$GEM_HOME" \
```

However, when the `$HOME` (`/root`) directory does not exist, the output of the `bundle --version` command looks like this:

```
[1][default:/src:0]# bundle --version
`/root` is not a directory.
Bundler will use `/tmp/bundler/home/unknown' as your home directory temporarily.
Bundler version 1.16.2
```

And the value of the variable will be:

```
directory.
temporarily.
1.16.2
```

This doesn't cause an error when displaying the output (other than looking weird), but it does when we try to vendor a file literally named:

```
/hab/pkgs/core/bundler/1.16.2/20180702163315/cache/bundler-directory.
temporarily.
1.16.2.gem
```

That is, the scaffolding is running:

```
gem install --local "/hab/pkgs/core/bundler/1.16.2/20180702163315/cache/bundler-directory.
temporarily.
1.16.2.gem" --install-dir "$GEM_HOME"
```

The last issue is that bundler itself doesn't send warnings and errors to STDERR, so our naive parsing of the output results in mangled data that we expect to use in the scaffolding. This isn't something we can change in Habitat or the scaffolding, but the workaround for the scaffolding is to match the output on the line that starts with `Bundler version`, using `awk` or `grep` depending on where this is relevant for the scaffolding script.

For Habitat itself, we should ensure that the `$HOME` directory exists for the default user, because there may be other software out there that expects to use `$HOME` for something.

jtimberman:
Fixes #5569

This creates root's $HOME, /root, with permissions as it was done in
ages past.

### DIFF
--- a/components/rootless_studio/default/entrypoint.sh
+++ b/components/rootless_studio/default/entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 hab pkg exec core/hab-backline mkdir -m 1777 -p /tmp
+hab pkg exec core/hab-backline mkdir -m 0750 -p /root
 
 source /etc/habitat-studio/import_keys.sh
 source /etc/habitat-studio/environment.sh


thesentinels:

Thank you for contributing!

Issue #5390, opened 2018-07-27
`hab sup *` commands should be able to determine CTL Gateway IP:Port of local Supervisor


jeremymv2:
With Habitat 0.56.0 there is the concept of a remote control and subsequently, a shared secret that needs to be synced in `CTL_SECRET`. 

Typically, this is transparent with working with a local Supervisor since the client uses `127.0.01` plus the default CTL Gateway port: `9632` . When it's a remote Supervisor, there's a handy argument for targeting it: `--remote-sup IP:PORT`.

However, in a Dockerized environment (Host) with multiple Hab exported docker containers, each running a Supervisor managing a service, it can be problematic and non-intuitive to determine the `IP:Port` to use.

This is because, as a prerequisite, to that set up, you must assign each Container Supervisor a unique CTL Gateway Port. An example of this is passing `--listen-ctl 0.0.0.0:9633` option to `docker run ..` and incrementing the port for each subsequent Container Supervisor.

When that is the case, you have to then know the unique Port for the locally running Supervisor in each container when running any `hab sup *` commands.

One way to do that is:

```
~ $ export SUP_PORT=$($(hab pkg path core/curl)/bin/curl -s http://127.0.0.1:963
1/butterfly | $(hab pkg path core/jq-static)/bin/jq '.service.list."workflow-ser
ver.default"[].service.sys.ctl_gateway_port' -e -c -M -r)

~ $ hab sup status --remote-sup 127.0.0.1:$SUP_PORT
package                                             type        state  elapsed (s)  pid   group
jeremymv2/workflow-server/1.8.0-dev/20180726192935  standalone  up     857          3777  workflow-server.default

~ $
```

However, it is my opinion that the CTL Gateway Port could be stored on disk under `/hab/sup/` perhaps so that when not intentionally communicating with a remote Supervisor, the `hab sup *` commands can determine the correct Port to use. Currently they just assume the default Port (9632).

irvingpop:
https://github.com/habitat-sh/habitat/issues/3025 is related and relevant, it would be ideal if you could ask the supervisor to dynamically allocate a ctl port  - if you have to run multiple supervisors on the same box (which seems to happen to me all the time),  it gets super confusing with secrets mismatches.

Issue #5266, opened 2018-06-27
Remove support for multiple supervisors on a single host


baumanj:
The support for multiple supervisors on a single host is currently poor and error-prone. Since it's possible to cleanly do this with VMs or containers, there's little incentive to maintain a tricky feature which is little used and hard to get right.

It's still fine to be in multiple chroot studios at the same time. This is useful for building even without the supervisor running. But only on studio should have a running supervisor. It's simple to start and stop with the `sup-run` and `sup-term` helpers.

Additional issues will be filed to address usability issues around what happens when a supervisor is expected to be running but isn't and if starting a supervisor fails because there's already one running.

See [v0.56.0 [Err: 4] secret key mismatch on sup restart studio](https://github.com/habitat-sh/habitat/issues/5169)

baumanj:
When implementing this, it might be worth changing the standard `/hab/sup/default` path to supervisor files to just `/hab/sup`.

irvingpop:
We have at least one production customer situation where this is needed.   It's using docker containers, but with host-based networking.

baumanj:
Can you go into more detail, @irvingpop? Since the containers have their own process space/file system, that still sounds like one supervisor process per what we mean by "host" in this sense.

irvingpop:
Habitat in containers has already been pretty painful to deal with when the network namespace is shared (documented in #3025), and unfortunately the remote controlled supervisor feature made it even more painful. 

In Kubernetes land, sharing process namespaces appears to be quite commonplace.  Specifically the "side car" pattern for loading multiple containers into a single pod.  I'm concerned that we could make that impossible to load two different habitat-created containers into the same pod, and it would be a very difficult to troubleshoot gotcha. 

To answer your question, I don't believe it's possible to do `--override-name` in docker-launched supervisors.

baumanj:
If they're not using `--override-name`, then this issue shouldn't affect them. We're just trying to eliminate a feature that was never particularly well functioning or supported anyway.

The design of the supervisor isn't really amenable to running multiple instances on the same host. It's purpose is to coordinate *between* hosts. All services in a single process space/file system should be running under the same supervisor.

irvingpop:
Okay, I understand the rationale behind this and it makes sense.    Two things I would ask for:
1.  If at all possible, can you please ensure that a second supervisor fails gracefully and logs a useful error message?
[...]

PR #5576, opened 2018-09-06
store control gateway address in file

jeremymv2:
Resolves https://github.com/habitat-sh/habitat/issues/5390

## Description :nut_and_bolt:
This change creates a new `CTL_GATEWAY_ADDRESS` file under `sup_root`.
The purpose is to improve the UX when the Supervisor was started with
`--listen-ctl <LISTEN_CTL>`. Instead of having to specify `--remote-sup
<REMOTE_SUP>` on every `hab sup *` command, hab will now first attempt
to read the ctl gateway address from file on disk.

This scenario is especially common in Docker exported hab pkg
situations where users might have multiple hab services running in
docker containers, on one Docker Host. In order to do so, they must
customize the `--listen-ctl` address of each Supervisor. In each
container, users must then follow up with `--remote-sup` when talking to the
local supervisor, which is cumbersome and error prone (with users often
not specifying the correct address and talking to a different
supervisor!).


thesentinels:

Thank you for contributing!

baumanj:
I'm taking a look. One thing to be aware of is that we're [planning to remove support for running multiple supervisors on a single host](url
https://github.com/habitat-sh/habitat/issues/5266).

irvingpop:
We have at least one production customer situation where this is needed.   It's using docker containers, but with host-based networking.   That's what this PR is attempting to solve for.

baumanj:
Quick summary of the chat that I just had with @jeremymv2:

Following the pattern of the way we use environment variables to override CLI options (e.g., `HAB_BLDR_URL` for `--url`), we can provide overrides for `--listen-{ctl,gossip,http}`. That should be simpler than reading/writing files.

christophermaier:
An alternative to look at would be adding an address entry to the [cli.toml](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/config.rs). That may be more appropriate since this is a client-side concern; adding a new file to the Supervisor root seems like it may be overly-tailored to the specific use case you all are dealing with @irvingpop @jeremymv2 

Always happy to chat more about details of the situation, though 😄

christophermaier:
Environment variables are also cool 👍

jeremymv2:
@baumanj @christophermaier thank you for your feedback. After some fresh research into Docker host mode networking, I do not see a way we can achieve the appropriate Supervisor network port level isolation per container with this customer.

I'm going to go the route of the Environment variable pattern override to the command line arguments: `--listen-{ctl,gossip,http}` This should prove to be a less heavy handed approach and also provide a nice feature that adds value for differing runtime environments.

I can just as easily close this PR or add the new commits to this one,  likely renaming it.

baumanj:
Let's modify this one, it has useful context.

jeremymv2:
@baumanj @christophermaier @irvingpop I have added some new commits that are ready for review.

With these changes, the `--listen-{ctl,gossip,http}` arguments can be overridden via the following Environment variables `HAB_LISTEN_CTL`, `HAB_LISTEN_GOSSIP` and `HAB_LISTEN_HTTP` respectively, if set.

The order of precedence is:
1. value taken from `--listen-xxx`, if specified.
2. value taken from `HAB_LISTEN_XXX`, if set.
3. default value will be used.
 
Trying it out can be accomplished with any combination of:
```
sudo -E HAB_LISTEN_CTL=127.0.0.1:9635 HAB_LISTEN_GOSSIP=0.0.0.0:9645 HAB_LISTEN_HTTP=127.0.0.1:9644 ./target/debug/hab sup run &
```

I am much more satisfied with the Environment variable approach. It solves my specific Docker environment concern quite adequately and is sufficiently flexible and light weight to be useful in other scenarios.

components/sup/src/config.rs
```diff
@@ -55,12 +59,30 @@ impl GossipListenAddr {
     }
 }
 
+// Returns a GossipListen listener address value if set in the environment. Does *not*
+// return any default value if the value was not found in the environment!
+pub fn listen_gossip_from_env() -> Option<String> {
```
baumanj: Same comments as for `listen_ctl_from_env`.

What might be useful is a general-purpose function like `socket_addr_from_env(&str: env_var, default: SocketAddr) -> SocketAddr`, since we're using that pattern several places.

components/sup/src/http_gateway.rs
```diff
@@ -42,6 +43,18 @@ use feat;
...
 
+/// Default listening port for the HTTPGateway listener.
+pub const DEFAULT_PORT: u16 = 9631;
+
+/// Default environment variable override for HTTPGateway listener address.
+pub const DEFAULT_ADDRESS_ENVVAR: &'static str = "HAB_LISTEN_HTTP";
+
+// Returns a HTTPGateway listener address value if set in the environment. Does *not*
+// return any default value if the value was not found in the environment!
+pub fn listen_http_from_env() -> Option<String> {
```
baumanj: Same comments

jeremymv2:
@baumanj I am curious what your thoughts are on these most recent changes and thanks 🎉 😄

baumanj:
Hopefully I'll get to this tomorrow, @jeremymv2. Some relatively pressing stuff came up today.

PR #5588, opened 2018-09-07
Increase Network Stability across Supervisor Restarts

christophermaier:
Fixes two bugs that were contributing to general network instability when restarting a Supervisor (especially when restarting for Supervisor upgrades).

Further details are present in the individual commits, but the high-level summary is:

* Only send a Departed Membership rumor when the Supervisor is really shutting down (including all services).
* Properly serialize Membership rumors to ensure they are appropriately reconstituted on restart, providing a consistent view of the gossip network.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -1572,8 +1572,44 @@ impl Manager {
     }
 
     fn shutdown(&mut self, cause: ShutdownReason) {
-        outputln!("Gracefully departing from butterfly network.");
-        self.butterfly.set_departed();
+        match cause {
+            ShutdownReason::PkgUpdating | ShutdownReason::Signal => {
+                // Previously, we would unconditionally set our health
+                // to departed. However, given our current conflation
+                // of Supervisor reachability and Service health, I
+                // suspect this causes instability when simply
+                // shutting down the Supervisor for a Supervisor
+                // upgrade.
+                //
+                // In the future, we may want to notify the rest of
+                // the network more formally of our intent to update,
+                // but for now, NOT convincing everyone that our
+                // services are gone seems like a good intermediate
+                // step. (We should also, of course, stop conflating
+                // Supervisor reachability and Service health!) Our
+                // existing suspicion mechanism should serve us well
+                // here if it takes longer than expected for the new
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -1572,8 +1572,44 @@ impl Manager {
...
+                //
+                // Thus, for these given ShutdownReasons, we just
+                // won't send any Membership rumors out.
+            }
+            ShutdownReason::SvcStopCmd => {
+                // Just to call it out specifically, we shouldn't ever
+                // be called with this ShutdownReason.
+                //
+                // This is all being refactored elsewhere right now,
+                // for what it's worth.
```
baumanj: If this code path should really never execute, let's put either an `unreachable!()` or at least a strong `error!(…)` here.

components/sup/src/manager/mod.rs
```diff
@@ -1572,8 +1572,44 @@ impl Manager {
...
+                // but for now, NOT convincing everyone that our
+                // services are gone seems like a good intermediate
+                // step. (We should also, of course, stop conflating
+                // Supervisor reachability and Service health!) Our
+                // existing suspicion mechanism should serve us well
+                // here if it takes longer than expected for the new
+                // Supervisor to come back up.
+                //
+                // Thus, for these given ShutdownReasons, we just
+                // won't send any Membership rumors out.
```
baumanj: Can we move this and the other long comments to a doc comment? I'd say they're relevant to the contract of the function, not just implementation details.

christophermaier:
@baumanj As near as I can reconstruct, the fact that `Membership` didn't implement `Message` is most likely something that got missed in the large overall change to migrate to Prost (like, an initial pass at the refactoring, the cleanup of which got missed as the overall work wrapped up). Actually, whatever functions are calling these should have a `Message` trait bound. I'll see if adding that is straightforward; if so, I'll add it to this PR.

As for your other comments, now that I'm done with this, I'm moving back to my previous work on refactoring service shutdown, which actually removes `ShutdownReason` altogether and simplifies the "should we send a departure rumor" logic drastically. So all that code and comments will go away.

PR #5591, opened 2018-09-07
Remove/replace set_{permissions,owner} calls which aren't supported on windows

baumanj:
(NB: This will be easier to review with whitespace changes hidden)

These are misleading since the core implementation doesn't actually do what the
function names imply, so replace with more illustrative code on a per-platform
basis. Long-term, we should implement the appropriate windows FFI calls, but
the current situation is dangerously misleading.

See https://github.com/habitat-sh/core/blob/master/components/core/src/os/filesystem/windows.rs#L28-L34

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -32,6 +32,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use butterfly::rumor::service::Service as ServiceRumor;
+use hcore;
 use hcore::crypto::hash;
 use hcore::fs::FS_ROOT_PATH;
 use hcore::package::metadata::Bind;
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/mod.rs
```diff
@@ -957,7 +964,21 @@ impl Service {
         }
 
         if abilities::can_run_services_as_svc_user() {
```
christophermaier: I notice that you pulled this test into the `if cfg!(linux)` test above, but left it outside here; was that intentional?

baumanj: Good catch. They should be the same. I'll fix that up.

baumanj: Actually, no. They were intentionally different to accommodate the different error handling. I needed to set `result` to an appropriate `Result` type in both branches, so I think this is a good approach. Do you see any problems with it?

christophermaier: Well, it always returns `true` on Windows: https://github.com/habitat-sh/habitat/blob/7b590f394c7aa211af75823e9e96d8697eb485de/components/sup/src/sys/windows/abilities.rs#L18-L27

I think that's appropriate for what we're trying to do (it was certainly backward compatible with our Windows behavior when it was introduced), though @mwrock will have the final say on that. If it is indeed appropriate, then we should take it outside of the platform-dependent branching logic we have here at the call sites. It seems that the benefit of pulling the branching logic out here is that it makes it obvious that we haven't quite gotten the right abstraction over the different behaviors, and prematurely hiding that can give us a false sense of security.

I feel that putting the call inside the branching in one scenario and outside in another makes things less clear on that point.

baumanj: Depending on how `can_run_services_as_svc_user` is currently implemented breaks the abstraction. Ultimately the reason these two code paths are different is because they were different before: one does an `outputln!()` on `Err` return from `set_owner` and the other doesn't. I'm not trying to clean up the code so much as just remove the misleading nature of calls that don't do what they advertise on Windows.

It's definitely ugly, but it should all get refactored when we're done with https://github.com/habitat-sh/core/issues/58.

baumanj: Are you ok with this as-is, @christophermaier? If not, what code would you rather see?

christophermaier: @baumanj It's fine for now, pending the bigger refactor 👍

PR #5592, opened 2018-09-08
pkg export and hab-sup/hab-launcher commands should honor fs_root

mwrock:
This week @trevorghess noticed that `hab pkg export docker` inside of a local windows studio was failing with:

```
thread 'main' panicked at 'Could not find 'docker' in the 'core/docker' package!
```

After researching this it ends up that `hab pkg export` commands do not honor the `FS_ROOT` environment variable on Windows which will cause export to break in a local (non-docker) studio.

This fixes that scenario. It also fixes another case I have been aware of for a while but was not broken enough to research. Formerly `hab-launch` and `hab-sup` binary calls have not honored fs_root either. This means that in a local Windows studio, the launcher and supervisor binaries in `c:\hab\pkgs` are invoked. Functionally that is really not a big deal but is out of alignment with the self contained nature of a studio.


thesentinels:

Thank you for contributing!

components/hab/src/exec.rs
```diff
@@ -77,8 +77,8 @@ where
         return Err(Error::ExecCommandNotFound(command));
     }
 
-    let fs_root_path = Path::new("/");
-    match PackageInstall::load_at_least(ident, None) {
+    let fs_root_path = Path::new(&*FS_ROOT_PATH);
```
baumanj: How about:
```rust
    let fs_root_path = FS_ROOT_PATH.as_path();
```

I think doing this more explicitly is easier to understand for those who may not intuitively understand all of the rust magic going on here. Especially when the name `FS_ROOT_PATH` doesn't make it clear that its type is actually a reference to a `PathBuf`.

```rust
    let fs_root_path = &*FS_ROOT_PATH;
```
Works too, but I think that while it's shorter, it's more confusing.

components/launcher/src/server/mod.rs
```diff
@@ -605,8 +606,9 @@ fn supervisor_cmd() -> Result<PathBuf> {
         return Ok(PathBuf::from(command));
     }
     let ident = PackageIdent::from_str(SUP_PACKAGE_IDENT).unwrap();
-    match PackageInstall::load_at_least(&ident, None) {
-        Ok(install) => match core::fs::find_command_in_pkg(SUP_CMD, &install, "/") {
+    let fs_root_path = Path::new(&*FS_ROOT_PATH);
```
baumanj: Same comment

Issue #5329, opened 2018-07-12
Run lifecycle hooks asynchronously


christophermaier:
## Description
Lifecycle hooks currently run synchronously in the main Supervisor loop. This is Bad(TM), because it leaves open the possibility of a misbehaving hook of one service to block the loop for _everyone_. When this happens the supervisor will no longer respond to commands or be able to manage any other services until that hook completes.

It would be better to handle this asynchronously.

## Acceptance Criteria
* `post-run` hook no longer blocks
* `init` hook no longer blocks
* `file-updated` hook no longer blocks
* `health-check` hook no longer blocks
* `post-run` hook no longer blocks
* `reconfigure` hook no longer blocks
* ` suitability` hook no longer blocks

christophermaier:
Related to https://github.com/habitat-sh/habitat/issues/1973

PR #5603, opened 2018-09-10
supervisor should not init signals on windows

mwrock:
0.62 is currently breaking some customers that have long running `init` hooks. Killing a supervisor via ctrl+c or stopping the windows service simply hangs because 0.62.0 turns off the default ctrl+c handler for the supervisor.

This exposes some potential snags when we get to moving service termination from the launcher to the supervisor. We can work those out when we get there but this should unbreak windows supervisor consumers for now.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -669,7 +669,16 @@ impl Manager {
             Some(ref evg) => Some(events::EventsMgr::start(evg.clone())),
             None => None,
         };
-        signals::init();
+        // On Windows initializng the signal handler will create a ctrl+c handler for the
+        // process which will disable default windows ctrl+c behavior and allow us to
+        // handle via check_for_signal. However, if the supervsor is in a long running
+        // non-run hook, the below loop will not get to check_for_signal in a reasonable
+        // amount of time and the supervisor will not respond to ctrl+c. On Windows, we
+        // let the launcher catch ctrl+c and gracefully shut down services. ctrl+c should
+        // simply halt the supervisor
+        if !feat::is_enabled(feat::IgnoreSignals) {
+            signals::init();
+        }
         loop {
             if feat::is_enabled(feat::TestExit) {
                 if let Ok(exit_file_path) = env::var("HAB_FEAT_TEST_EXIT") {
@@ -696,10 +705,12 @@ impl Manager {
                 self.shutdown(ShutdownReason::Departed);
                 return Err(sup_error!(Error::Departed));
             }
-            if let Some(SignalEvent::Passthrough(Signal::HUP)) = signals::check_for_signal() {
-                outputln!("Supervisor shutting down for signal");
-                self.shutdown(ShutdownReason::Signal);
-                return Ok(());
+            if !feat::is_enabled(feat::IgnoreSignals) {
+                if let Some(SignalEvent::Passthrough(Signal::HUP)) = signals::check_for_signal() {
+                    outputln!("Supervisor shutting down for signal");
+                    self.shutdown(ShutdownReason::Signal);
+                    return Ok(());
+                }
             }
             if let Some(package) = self.check_for_updated_supervisor() {
                 outputln!(
--- a/components/sup/src/lib.rs
+++ b/components/sup/src/lib.rs
@@ -137,10 +137,11 @@ lazy_static! {
 /// Search for feat::is_enabled(feat::FeatureName) to learn more
 features! {
     pub mod feat {
-        const List         = 0b00000001,
-        const TestExit     = 0b00000010,
-        const TestBootFail = 0b00000100,
-        const RedactHTTP   = 0b00001000
+        const List          = 0b00000001,
+        const TestExit      = 0b00000010,
+        const TestBootFail  = 0b00000100,
+        const RedactHTTP    = 0b00001000,
+        const IgnoreSignals = 0b00010000
     }
 }
 
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -421,6 +421,7 @@ fn enable_features_from_env() {
         (feat::TestExit, "TEST_EXIT"),
         (feat::TestBootFail, "BOOT_FAIL"),
         (feat::RedactHTTP, "REDACT_HTTP"),
+        (feat::IgnoreSignals, "IGNORE_SIGNALS"),
     ];
 
     // If the environment variable for a flag is set to _anything_ but


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -669,7 +669,16 @@ impl Manager {
...
+        // On Windows initializng the signal handler will create a ctrl+c handler for the
+        // process which will disable default windows ctrl+c behavior and allow us to
+        // handle via check_for_signal. However, if the supervsor is in a long running
+        // non-run hook, the below loop will not get to check_for_signal in a reasonable
+        // amount of time and the supervisor will not respond to ctrl+c. On Windows, we
+        // let the launcher catch ctrl+c and gracefully shut down services. ctrl+c should
+        // simply halt the supervisor
+        if cfg!(not(windows)) {
+            signals::init();
+        }
```
christophermaier: @mwrock To make sure I've got the right understanding, once #5329 is done this could be reverted and still behave correctly, right?

I suppose this also means that there won't be any restarting of the Supervisor (with #5445) on Windows with this in place.

Just thinking out loud, could there be an alternative appropriate fix in `signals::init()` that could handle things in a cross-platform way?

baumanj: As far as I can tell, the only thing the `signals` implementation in our `core` crate does for windows is set up a ctrl-c handler, so https://github.com/habitat-sh/habitat/pull/5445 didn't add any functionality to that platform. I'll file an issue against the `core` repo to do more thorough handling on windows.

If we make this change (which I think is appropriate), we should also add a `not(windows)` conditions to the other references:

```rust
use hcore::os::signals::{self, SignalEvent};
```
on L56 and
```rust
            if let Some(SignalEvent::Passthrough(Signal::HUP)) = signals::check_for_signal() {
```
on L699

Currently we get lucky because [ATOMIC_BOOL_INIT](https://github.com/habitat-sh/core/blob/master/components/core/src/os/signals/windows.rs#L24) is `false`, but we shouldn't be calling `signals` functions if we haven't called `init`.

mwrock: I think ideally `signals::init` would be called on windows and we would need no platform branching. Then below we would check for both `SIGHUP`(signaled on Linux) and `Shutdown`(signaled on Windows) and do the same thing for both:`self::shutdown`.

We just can't do this until we get non-run hooks in their own threads. Actually Linux suffers from the same potential problem here that a long running init hook would delay the shutdown. Its just that Windows plans are much more likely to have long runnin init hooks and thus this is a more dire problem on that platform.

mwrock: but yeah, you are right that we should hide the above 2 lines for now on Windows. I'll add that.

baumanj: Filed https://github.com/habitat-sh/core/issues/59

baumanj: Maybe we should add an environment variable to toggle the signal catching behavior on or off if it can apply to either platform. Better to add that now (or just after this release if you think it's too much) than scramble later if it's affecting a production system.

mwrock: thats a great idea and I could rev a new windows-service package that sets that variable

mwrock:
I have a windows-service PR about to come in that will ensure `HAB_SUP_IGNORE_SIGNAL` is set by default but can be configured off.

components/sup/src/manager/mod.rs
```diff
@@ -696,10 +705,15 @@ impl Manager {
                 self.shutdown(ShutdownReason::Departed);
                 return Err(sup_error!(Error::Departed));
             }
-            if let Some(SignalEvent::Passthrough(Signal::HUP)) = signals::check_for_signal() {
-                outputln!("Supervisor shutting down for signal");
-                self.shutdown(ShutdownReason::Signal);
-                return Ok(());
+            if env::var("HAB_SUP_IGNORE_SIGNAL").is_err() {
+                match signals::check_for_signal() {
+                    Some(SignalEvent::Passthrough(Signal::HUP)) | Some(SignalEvent::Shutdown) => {
```
christophermaier: Shutting down with `ShutdownReason::Signal` will cause the supervisor to restart, which seems at odds with a `SignalEvent::Shutdown` signal.

mwrock: OH! That is not the desired result at all.  Fixing.

components/sup/src/manager/mod.rs
```diff
@@ -669,7 +669,16 @@ impl Manager {
...
+        // On Windows initializng the signal handler will create a ctrl+c handler for the
+        // process which will disable default windows ctrl+c behavior and allow us to
+        // handle via check_for_signal. However, if the supervsor is in a long running
+        // non-run hook, the below loop will not get to check_for_signal in a reasonable
+        // amount of time and the supervisor will not respond to ctrl+c. On Windows, we
+        // let the launcher catch ctrl+c and gracefully shut down services. ctrl+c should
+        // simply halt the supervisor
+        if env::var("HAB_SUP_IGNORE_SIGNAL").is_err() {
+            signals::init();
+        }
```
baumanj: I'd have a small preference for setting a constant over duplicating code. Something like
```rust
let handle_signals = env::var("HAB_SUP_IGNORE_SIGNAL").is_err()
```
But really, it's probably best to implement this as a feature and do:
```rust
if feat::is_enabled(feat::HandleSignals)
```
Either way, not a big deal, just a suggestion. I think this code should work.

components/sup/src/manager/mod.rs
```diff
@@ -696,10 +705,15 @@ impl Manager {
...
-                return Ok(());
+            if env::var("HAB_SUP_IGNORE_SIGNAL").is_err() {
+                match signals::check_for_signal() {
+                    Some(SignalEvent::Passthrough(Signal::HUP)) | Some(SignalEvent::Shutdown) => {
+                        outputln!("Supervisor shutting down for signal");
+                        self.shutdown(ShutdownReason::Signal);
+                        return Ok(());
+                    }
+                    _ => {}
+                }
```
baumanj: Per @christophermaier's comment, if we're not also matching on `Some(SignalEvent::Shutdown)` (which we shouldn't), I think leaving this as the `if let` form is preferable to changing it to `match`.

components/sup/src/manager/mod.rs
mwrock: how would you feel if I added a new `ShutdownReason` called `CancelProcess` that a `shutdown` signal would trigger and result in a butterfly departure?

christophermaier: I wouldn't make a new `ShutdownReason`, mainly because that's going away on the branch I'm working on for terminating services from the Supervisor 😄 

I think the `LauncherStopping` reason is what you want here, right?

mwrock:
@baumanj I made the feature `IgnoreSignals` since I think the default state should be to handle them because that is what the current behavior is.

components/sup/src/manager/mod.rs
```diff
@@ -696,10 +705,20 @@ impl Manager {
...
-                self.shutdown(ShutdownReason::Signal);
-                return Ok(());
+            if !feat::is_enabled(feat::IgnoreSignals) {
+                match signals::check_for_signal() {
+                    Some(SignalEvent::Passthrough(Signal::HUP)) => {
+                        outputln!("Supervisor shutting down for signal");
+                        self.shutdown(ShutdownReason::Signal);
+                        return Ok(());
+                    }
+                    Some(SignalEvent::Shutdown) => {
```
baumanj: Why are we making this change? The supervisor isn't supposed to receive these signals. Shutdown needs to be orchestrated through the launcher and [the launcher communicates that via the IPC channel here](https://github.com/habitat-sh/habitat/blob/master/components/launcher/src/server/mod.rs#L180).

Arguably the catching of the `HUP` signal here should've been done in the launcher and passed to the supervisor through IPC, but it was less code to just let the signal pass through directly. Adding additional signal handling the Supervisor is something worth discussing, but I don't think it should be part of this change.

PR #5637, opened 2018-09-17
do not install and upload launcher in release builds

mwrock:
I have narrowed down the launcher build failures on appveyor to release builds. It only fails on release builds and I HAVE NO IDEA WHY. This will turn off the install/upload whic fails and allow the build to succeed. This should allow me to download the built hart file that claims to have no target file.

Since this only seems to happen during release builds, I'll just make sure to examine this on the next release.

Note: that we don't actually need a uploaded launcher during release builds so this should have no adverse effect.


thesentinels:

Thank you for contributing!

PR #5651, opened 2018-09-18
Create /usr/bin when running the studio

### LINKED ISSUE #5646: npm 6.4.1 dependencies directory throws error
I'm running the latest version of Habitat 0.62 on OSX High Sierra version 10.13.6. I have a Node js application that I've written and I'm trying to Habitize it using core/scaffolding-node. When I go to build it locally, it throws an error when npm tries install the dependencies;

Installing dependencies using npm 6.4.1
'/usr/bin/env' -> '/hab/pkgs/core/coreutils/8.29/20180608092141/bin/env'
ln: /usr/bin/env: No such file or directory

I can successfully build it out in Builder using the source from Github.

smacfarlane:
Resolves #5646 
Ref: habitat-sh/core-plans#1857

This may need to move into the image creation if we accumulate any more of these mkdirs,  but I think for now it's ok to do it at runtime. 

### DIFF
--- a/components/rootless_studio/default/entrypoint.sh
+++ b/components/rootless_studio/default/entrypoint.sh
@@ -2,6 +2,7 @@
 
 hab pkg exec core/hab-backline mkdir -m 1777 -p /tmp
 hab pkg exec core/hab-backline mkdir -m 0750 -p /root
+hab pkg exec core/hab-backline mkdir -m 0755 -p /usr/bin
 
 source /etc/habitat-studio/import_keys.sh
 source /etc/habitat-studio/environment.sh


christophermaier:
@smacfarlane Actually, we've already got three such `mkdir` calls... that's my personal signal that it's time to consider refactoring. Now may be time to move these into the image creation stage.

Thoughts, @elliott-davis ?

elliott-davis:
I tried this initially and found that during image creation we don't actually have a shell program to create directories, so we end up just copying from our parent (alpine) before we destroy that layer. I'm not sure what the right answer is but I don't know if I'd block solving this issue for a really deep yak dive

PR #5652, opened 2018-09-18
windows defaults to local studio

### LINKED ISSUE #5464: Local studios should be the default on Windows
Currently the default `hab studio` behavior is to spawn a container which may not work especially on older OSs that cannot run docker. We should make the current local `hab studio -w` the default (no more `-w`) and make container studios explicit via an arg.

mwrock:
Fixes #5464

This makes the local studio the default on Windows. To run a dockerized container, `-D` must be passed. This applies to both `hab studio` and `hab pkg build` commands.

For now, we will still honor `-w` but state that it is deprecated and emit a warning when it is used. We will alo emit a warning when entering the studio without `-w` or `-D` that advises how to enter a dockerized studio (the former default).

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -686,8 +686,8 @@ fn sub_pkg_build() -> App<'static, 'static> {
         (@arg SRC_PATH: -s --src +takes_value
             "Sets the source path (default: $PWD)")
         (@arg PLAN_CONTEXT: +required +takes_value
-            "A directory containing a `plan.sh` file \
-            or a `habitat/` directory which contains the `plan.sh` file")
+            "A directory containing a plan file \
+            or a `habitat/` directory which contains the plan file")
     );
     // Only a truly native/local Studio can be reused--the Docker implementation will always be
     // ephemeral
@@ -697,22 +697,18 @@ fn sub_pkg_build() -> App<'static, 'static> {
                 .help("Reuses a previous Studio for the build (default: clean up before building)")
                 .short("R")
                 .long("reuse"),
+        ).arg(
+            Arg::with_name("DOCKER")
+                .help("Uses a Dockerized Studio for the build")
+                .short("D")
+                .long("docker"),
         );
     }
 
-    if cfg!(target_os = "linux") {
-        sub.arg(
-            Arg::with_name("DOCKER")
-                .help(
-                    "Uses a Dockerized Studio for the build (default: Studio uses a chroot on \
-                     linux)",
-                ).short("D")
-                .long("docker"),
-        )
-    } else if cfg!(target_os = "windows") {
+    if cfg!(target_os = "windows") {
         sub.arg(
             Arg::with_name("WINDOWS")
-                .help("Use a Windows Studio instead of a Docker Studio")
+                .help("Use a local Windows Studio instead of a Docker Studio")
                 .short("w")
                 .long("windows"),
         )
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -241,12 +241,20 @@ mod inner {
         }
 
         for arg in args.iter() {
+            let str_arg = arg.to_string_lossy();
+            if str_arg == String::from("-D") {
+                return false;
+            }
+        }
+
+        // -w/--windows is deprecated and should be removed in a post 0.64.0 release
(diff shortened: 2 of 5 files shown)


thesentinels:

Thank you for contributing!

components/studio/bin/hab-studio.sh
```diff
@@ -84,8 +84,8 @@ COMMON FLAGS:
     -q  Prints less output for better use in scripts
     -v  Prints more verbose output
     -V  Prints version information
-    -D  Use a Docker Studio instead of a chroot Studio (only available on Linux)
-    -w  Use a Windows Studio instead of a Docker Studio (only available on Windows)
+    -D  Use a Docker Studio instead of a chroot Studio
+    -w  Use a Windows Studio instead of a Docker Studio (DEPRECATED)
```
christophermaier: This is a Bash file... why are Windows-only options mentioned in here in the first place? 🤔

components/studio/bin/hab-studio.ps1
```diff
@@ -66,7 +67,8 @@ COMMON FLAGS:
     -n  Do not mount the source path into the Studio (default: mount the path)
     -q  Prints less output for better use in scripts
     -v  Prints more verbose output
-    -w  Use a Windows studio instead of a Docker Studio (only available on windows)
+    -D  Use a Docker Studio instead of a native Studio
+    -w  Use a Windows studio instead of a Docker Studio (DEPRECATED)
```
christophermaier: Might be nice to indicate that it's just the _flag_ that's deprecated, and not the use of a Windows studio itself.

components/hab/src/command/studio/enter.rs
```diff
@@ -240,13 +240,20 @@ mod inner {
...
         }
 
+        for arg in args.iter() {
+            let str_arg = arg.to_string_lossy();
+            if str_arg == String::from("-D") {
+                return false;
+            }
+        }
+
         for arg in args.iter() {
```
christophermaier: A comment to point out that this particular chunk of code is in support of a deprecated flag, and thus fair-game for removal soon would be helpful.

components/hab/src/command/studio/enter.rs
```diff
@@ -240,13 +240,20 @@ mod inner {
             return false;
         }
 
+        for arg in args.iter() {
+            let str_arg = arg.to_string_lossy();
+            if str_arg == String::from("-D") {
+                return false;
+            }
+        }
+
```
baumanj: Since this is nearly identical to the next block, I think it could be better simplified to:
```rust
        for arg in args.iter() {
            match arg.to_str().unwrap_or("") {
                "-D" => return false,
                "-w" | "--windows" => return true,
                _ => (),
            }
        }
```
This also avoids the need for as many objects to be created by using `&str`s instead of `String`s.

There is one small change here, the conversion to lower case is removed, but I think that's actually incorrect; `-W` (currently unused) shouldn't necessarily be treated the same as `-w`. Also, I don't see anywhere that `--windows` is a documented option, so I think we should either remove that or document it.

components/hab/src/cli.rs
```diff
@@ -687,8 +687,8 @@ fn sub_pkg_build() -> App<'static, 'static> {
         (@arg SRC_PATH: -s --src +takes_value
             "Sets the source path (default: $PWD)")
         (@arg PLAN_CONTEXT: +required +takes_value
-            "A directory containing a `plan.sh` file \
-            or a `habitat/` directory which contains the `plan.sh` file")
+            "A directory containing a `plan.sh` or `plan.ps1` file \
```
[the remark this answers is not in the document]
christophermaier: I really like this ☝️

components/studio/bin/hab-studio.sh
mwrock: Leftover from when there was no windows docker studio and this was the best way to tell people using the linux docker studio on a windows host how to open a windows studio. I'm just gonna remove this.

components/hab/src/command/studio/enter.rs
mwrock: `--windows` comes from `hab pkg build` which forwards to here. I had thought about doing what you suggest above but my concern was if someone included both flags. I know thats an edge case and technically erroneous but the `-D` flag should trump.

Issue #5462, opened 2018-08-16
Import studio environment variables explicitly rather than based on HAB_STUDIO_SECRET prefix


baumanj:
Currently any environment variable whose name begins with `HAB_STUDIO_SECRET` will be imported into the studio environment with the `HAB_STUDIO_SECRET` prefix stripped. For example:
```
➤ env HAB_STUDIO_SECRET_FOO=bar hab studio enter
…
   hab-studio: Entering Studio at /hab/studios/home--jbauman--habitat (default)
   hab-studio: Exported: HAB_AUTH_TOKEN=[redacted]
   hab-studio: Exported: HAB_ORIGIN=jbauman
   hab-studio: Exported: FOO=[redacted]
   hab-studio: Exported: GITHUB_TOKEN=[redacted]
…
[1][default:/src:0]# echo $FOO
bar
```

This is valuable functionality, but it lacks [affordances](https://en.wikipedia.org/wiki/Affordance) and operates in an implicit rather than explicit manner. To improve usability, moving to an interface where environment variables to export to the studio are specified via CLI options. For example:
```
➤ hab studio enter --env FOO=bar
```

This will require some changes to utilities that invoke `hab studio` and depend upon the implicit behavior such as the worker (h/t @elliott-davis), but in the long run, it should make the system more robust. It's also possible to implement the new behavior and gradually deprecate the old style to help minimize the risk of the transition.

See also https://github.com/habitat-sh/habitat/issues/5274

seth:
Having to specify the environment variables on the command line makes things explicit, which is nice. It might be helpful to have a mode where you only need to provide the list of env variable names to take across if those vars are set already. For transferring secrets, it would be better not to have to put var=val onto the command line. In many setups, you want to carry across the same set of env vars each time (our dev flow, e.g.). So this change as currently proposed doesn't seem like an improvement.

baumanj:
Totally agree on the ability to export an existing env var without putting it on the command line, @seth. It could work something like:
```
➤ hab studio enter --env FOO=bar HAB_AUTH_TOKEN
```
Where the existing value of `HAB_AUTH_TOKEN` in the environment would be exported inside the studio, but FOO could either not exist or be overridden. What should happen if `HAB_AUTH_TOKEN` isn't already set in the current environment is open to discussion. This is just an issue to say that the current approach isn't very user-friendly or robust (easy to pollute the studio env unintentionally as well), not to say exactly what the solution looks like.

[...]

PR #5657, opened 2018-09-19
Propagate DO_CHECK into studio if set

### LINKED ISSUE #3355: DO_CHECK is not passed to studio
Following up on conversation with @fnichol on slack, DO_CHECK is not white-listed to be passed to the studio.

Finding if I do a

```bash
DO_CHECK=true hab pkg build <path>
```

or

```bash
export DO_CHECK=true
hab pkg build <path>
```

tests are never run.  Currently, DO_CHECK has to be explicitly set in the studio to take effect, which means when builds are being run pragmatically, they have to be executed in a way similar to this:

```bash
hab studio run "DO_CHECK=true build <path>"
```

rsertelon:
This fixes #3355

### DIFF
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -82,6 +82,7 @@ pub fn start_docker_studio(_ui: &mut UI, mut args: Vec<OsString>) -> Result<()> 
 
     let mut env_vars = vec![
         String::from("DEBUG"),
+        String::from("DO_CHECK"),
         String::from("HAB_AUTH_TOKEN"),
         String::from("HAB_BLDR_URL"),
         String::from("HAB_BLDR_CHANNEL"),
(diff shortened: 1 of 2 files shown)


baumanj:
Does using `HAB_STUDIO_SECRET_DO_CHECK` as a workaround suffice? If so, I'd prefer to use that temporarily and make https://github.com/habitat-sh/habitat/issues/5462 the long term solution, as I think it makes for better UX.

rsertelon:
It might indeed.

I'd say it depends on whether this env variable should be considered as a built-in one, or not.
For the former, I think this PR is better, and for the latter, your proposal is really nice IMHO (I wasn't aware of it).

All `HAB_*` vars shouldn't be passed via the `--env` option I think, and I tend to think `DO_CHECK` should go along with `HAB_*` too. I'm surprised it's not named `HAB_PLAN_CHECK` btw ;)

rsertelon:
@elliott-davis sorry for the delay, I've added the declaration for the docker studio. Tell me if anything else seems missing :)

PR #5658, opened 2018-09-19
Copies /etc/inputrc in studio if available

### LINKED ISSUE #4060: Allow ctrl+<arrow> in studio shell
> hab 0.38.0/20171026215954

This is a nice to have, but frequently, I'm frustrated because I can't use CTRL + :arrow_left: nor CTRL + :arrow_right: in the studio.

Instead the studio shell prints things like `;5D` or `;5C` which I have to delete and use only arrow keys.

rsertelon:
This fixes #4060 by using host configuration for input in the studio.

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -566,6 +566,27 @@ users:x:999:
 EOF
   fi
 
+  # If `/etc/inputrc` is not present, create a minimal version to provide
+  # some standard key mappings in shell
+  if [ ! -f "$HAB_STUDIO_ROOT/etc/inputrc" ]; then
+    if [ -n "$VERBOSE" ]; then
+      echo "> Creating minimal /etc/inputrc"
+    fi
+    $bb cat > "$HAB_STUDIO_ROOT/etc/inputrc" << "EOF"
+# allow the use of the Home/End keys
+"\e[1~": beginning-of-line
+"\e[4~": end-of-line
+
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

HT154:
Does this also apply to the docker studio? Or just the chroot studio?

It would also be neat if it generated a sensible default inputrc if none was present on the host system (in /etc/inputrc or ~/.inputrc)

rsertelon:
@HT154 I think this might be just for the chroot studio. Not sure what is needed to make it work in docker.

christophermaier:
I think it might be good to define a minimal `/etc/inputrc` that we could use. That may make it easier for the Docker experience to be the same as the chroot studio.

rsertelon:
@christophermaier yep, that would be nice indeed. Thing is, I don't really know about `libreadline` and appart from proposing the default Ubuntu 18.04 `/etc/inputrc`... ;)

What should be have in your opinion at a minimum?

Of course, I'm thinking about support for ctrl+left/right arrow, anything else we could think of?

rsertelon:
Ping :)

rsertelon:
@christophermaier I've changed the implementation to create a default file in the studio.

Is there anything to do for the docker studio?

christophermaier:
Sorry for the delay @rsertelon ... This is on my list to get to very soon!

christophermaier:
Hrmm... I'm having trouble verifying that this works. I'm no readline guru, but I wonder if there's some kind of terminal shenanigans afoot. I'll continue investigating.

PR #5659, opened 2018-09-19
Honor secrets in `hab studio run` and `hab pkg build` in rootless studio image

elliott-davis:

### DIFF
--- a/components/rootless_studio/default/entrypoint.sh
+++ b/components/rootless_studio/default/entrypoint.sh
@@ -16,6 +16,7 @@ case "$1" in
     hab pkg exec core/hab-plan-build hab-plan-build "$@";;
   run)
     shift
-    hab pkg exec core/hab-backline bash --login -c "$@";;
+    # shellcheck disable=SC2046
+    hab pkg exec core/hab-backline env $(load_secrets) bash --login -c "$@";;
   *) echo "Unknown Studio Command" && exit 1;;
 esac
--- a/components/rootless_studio/default/entrypoint.sh
+++ b/components/rootless_studio/default/entrypoint.sh
@@ -15,7 +15,7 @@ case "$1" in
     hab pkg exec core/hab-backline env STUDIO_ENTER=true "${secrets[@]}" bash --login +h;;
   build)
     shift
-    hab pkg exec core/hab-plan-build hab-plan-build "$@";;
+    hab pkg exec core/hab-backline env "${secrets[@]}" /bin/build "$@";;
   run)
     shift
     hab pkg exec core/hab-backline env "${secrets[@]}" bash --login -c "$@";;
--- a/components/rootless_studio/default/entrypoint.sh
+++ b/components/rootless_studio/default/entrypoint.sh
@@ -7,16 +7,18 @@ hab pkg exec core/hab-backline mkdir -m 0755 -p /usr/bin
 source /etc/habitat-studio/import_keys.sh
 source /etc/habitat-studio/environment.sh
 
+declare -a secrets
+readarray -t secrets < <(load_secrets)
+
 case "$1" in
   enter)
-     # shellcheck disable=SC2046
-     hab pkg exec core/hab-backline env STUDIO_ENTER=true $(load_secrets) bash --login +h;;
+    hab pkg exec core/hab-backline env STUDIO_ENTER=true "${secrets[@]}" bash --login +h;;
   build)
     shift
     hab pkg exec core/hab-plan-build hab-plan-build "$@";;
   run)
     shift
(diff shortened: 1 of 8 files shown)


thesentinels:

Thank you for contributing!

components/rootless_studio/default/entrypoint.sh
```diff
@@ -15,6 +15,7 @@ case "$1" in
     hab pkg exec core/hab-plan-build hab-plan-build "$@";;
   run)
     shift
-    hab pkg exec core/hab-backline bash --login -c "$@";;
+    # shellcheck disable=SC2046
+    hab pkg exec core/hab-backline env $(load_secrets) bash --login -c "$@";;
```
baumanj: Considering that things like randomly generated keys are passed in through secrets, and may contain globbing characters, I think it's probably important to do this the safe way:
```bash
    secrets=($(load_secrets))
    hab pkg exec core/hab-backline env "${secrets[@]}" bash --login -c "$@";;
```

elliott-davis: @christophermaier tried this in a previous PR and it doesn't work. That's why we also disable the shellcheck above.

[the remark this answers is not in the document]
elliott-davis: Same answer as below

christophermaier:
I've sorted out how to allow the `SC2046` check to run successfully.

christophermaier:
@afiune I'm not 100% sure about Windows. This PR is to fix specific regressions in the Linux studio image.

components/rootless_studio/tests/bats/secrets.bats
```diff
@@ -0,0 +1,22 @@
...
+
+# TODO (CM): Better parameterize the HAB_DOCKER_STUDIO_IMAGE
+# environment variable.
+#
+# As it is, you can just execute these tests with that in the
+# environment, but it would be nice to make it explicit.
+#
+# HAB_DOCKER_STUDIO_IMAGE=my_image bats secrets.bats
+
+@test "hab pkg build (rootless Docker image): secrets are passed" {
```
elliott-davis: Since we are leveraging the hab cli to run these tests, we should probably pull them out of the rootless studio and into the general testing directory for the cli

components/rootless_studio/tests/fixtures/plans/test_build_with_secrets/default.toml
```diff
@@ -0,0 +1,3 @@
+# Use this file to templatize your application's native configuration files.
```
elliott-davis: This file can probably be removed

components/rootless_studio/tests/dejagnu/studio/secrets_are_passed.exp
```diff
@@ -0,0 +1,55 @@
...
+    # Be sure to always shut the Studio down; otherwise we'll leave
+    # the container running.
+    send "exit\r"
+    wait 2
+}
+
+# Define this as a variable for convenience and readability.
+set studio_prompt "default:/src:0\]# "
+
+# TODO (CM): This is currently a bit of a hack, in that it assumes
```
elliott-davis: I think this is because we're missing `HAB_NOCOLORING` here https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/studio/docker.rs#L83-L97

components/rootless_studio/tests/bats/secrets.bats
christophermaier: Yeah, once we get these plugged into automation, the organization will definitely change. This was just to get some tests in place for the regressions being fixed.

components/rootless_studio/tests/fixtures/plans/test_build_with_secrets/default.toml
christophermaier: Good point; gonna pull it now.

components/rootless_studio/tests/fixtures/plans/test_build_with_secrets/plan.sh
```diff
@@ -0,0 +1,16 @@
+pkg_name=test_build_with_secrets
+pkg_origin=core
+pkg_version="0.1.0"
+pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+pkg_license=("Apache-2.0")
+
+do_build() {
+    set -u
+    # The build will fail if the FOO environment variable is not set.
+    echo "The secret is $FOO"
```
afiune: tenor-118967254

mwrock:
@afiune it looks like this would work in a windows docker studio but not a local studio. I am going to open an issue for that.

PR #5660, opened 2018-09-20
fix local windows studio regressions

mwrock:
This fixes 2 local windows studio regressions.

1. This was released in 0.63.0 and fails to find `powershell` when invoking hooks because it is not looking in the `FS_ROOT`. This works in a docker studio, outside the studio or if powershell has been installed outside the studio root. However on a clean machine a local studio will fail to run any hook.
2. Recently commited but not yet released, `Get-SupervisorLog` in a local studio fails because it cannot find the `Test-InContainer` function which is out of scope.


thesentinels:

Thank you for contributing!

PR #5666, opened 2018-09-21
supervisor and svc pkg versions in sup log output

jeremymv2:
This change is targeted to humans reading Habitat Supervisor
logs whom wish to know the precise version of the Supervisor
and the Services that are managed.

```
17:48 $ sudo -E ./target/debug/hab sup run &
[1] 16618
hab-sup(MR): version core/hab-sup/0.64.0-dev <--
hab-sup(MR): Supervisor Member-ID 7807eb11c928433caffa0d04c8ae8a13
hab-sup(MR): Starting chef/automate-elasticsearch
hab-sup(MR): pkg version chef/automate-elasticsearch/6.2.2/20180527141927 <--
automate-elasticsearch.default(UCW): Watching user.toml
hab-sup(MR): Starting core/redis
hab-sup(MR): pkg version core/redis/3.2.4/20170514150022 <--
redis.default(UCW): Watching user.toml
...
```

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -603,14 +603,14 @@ impl Manager {
                         .pkg
                         .ident
                         .version
-                        .clone()
-                        .unwrap_or("UNKNOWN".to_string()),
+                        .as_ref()
+                        .unwrap_or(&"UNKNOWN".to_string()),
                     service
                         .pkg
                         .ident
                         .release
-                        .clone()
-                        .unwrap_or("UNKNOWN".to_string())
+                        .as_ref()
+                        .unwrap_or(&"UNKNOWN".to_string())
                 );
                 service
             }
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -335,7 +335,7 @@ impl Manager {
     fn new(cfg: ManagerConfig, fs_cfg: FsCfg, launcher: LauncherCli) -> Result<Manager> {
         debug!("new(cfg: {:?}, fs_cfg: {:?}", cfg, fs_cfg);
         let current = PackageIdent::from_str(&format!("{}/{}", SUP_PKG_IDENT, VERSION)).unwrap();
-        outputln!("version {}", current);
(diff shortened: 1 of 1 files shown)


components/sup/src/manager/mod.rs
```diff
@@ -594,7 +594,26 @@ impl Manager {
...
+                        .version
+                        .as_ref()
+                        .unwrap_or(&"UNKNOWN".to_string()),
+                    service
+                        .pkg
+                        .ident
+                        .release
+                        .as_ref()
+                        .unwrap_or(&"UNKNOWN".to_string())
+                );
```
christophermaier: I think you should be able to replace all this with
```rust
outputln!("pkg version {}", service.pkg.ident);
```
since `PkgIdent` implements `Display`.

jeremymv2:
I appears to me that the AppVeyor build failed with an unrelated error:
```
✗✗✗
✗✗✗ We tried 5 times but could not download core/hab-studio/0.63.0/20180914030343. Giving up.
✗✗✗
```

components/sup/src/manager/mod.rs
christophermaier: Thinking a bit more, In terms of unifying the log output more, it may be better to remove the `outputln!` call from line 585 and do it all down here:

```rust
outputln!("Starting {} ({})", spec.ident, service.pkg.ident);
```
That would give us a single line like this:
```
hab-sup(MR): Starting core/redis (core/redis/3.2.4/20170514150022)
```
rather than
```
hab-sup(MR): Starting core/redis
hab-sup(MR): pkg version core/redis/3.2.4/20170514150022
```

christophermaier:
@jeremymv2 Yeah, I think that error is unrelated to this work.

jeremymv2:
I like your idea @christophermaier 
I've tried to unify the output further, here is what it looks like now:

```
hab-sup(MR): core/hab-sup (core/hab-sup/0.64.0-dev)
hab-sup(MR): Supervisor Member-ID 7807eb11c928433caffa0d04c8ae8a13 
hab-sup(MR): Starting core/redis (core/redis/3.2.4/20170514150022)
redis.default(UCW): Watching user.toml
```

PR #5673, opened 2018-09-25
Refuting an incarnation immediately increments to one-greater

christophermaier:
Previously, whenever we refuted a rumor that we were anything but
`Alive`, there was an implicit assumption that the incarnation of the
rumor we were refuting was equal to our own. Thus, simply incrementing
our number by 1 would be enough to have a successful refutation.

In general, this *should* be true. However, reality is sometimes different.

In particular, given the Butterfly server's history of not persisting
incarnation numbers, as well as the Supervisor's practice of sending
out Departure health rumors when it shut down for upgrade (fixed
recently in #5588), it can be possible for the overall network to have
one conception of a member's incarnation number, while the member
itself thinks it has a lower value. This can result in several rounds
of refutation, where the member sends out new rumors with
incrementally larger incarnation numbers that are nevertheless smaller
than what the network's version of the incarnation is. This results in
a possibly long wait until the rumor is refuted enough times to
finally be recognized by the network.

This may also happen if a server cannot for some reason persist its
incarnation number to disk. The in-memory (and thus, in-network)
incarnation will continue to increase, while the on-disk version will
stagnate. Absent any additional intervention, upon a restart, a server
will think its incarnation is less than what the network does, leading
to the same scenario described above.

Now, instead of simply incrementing our incarnation by one to refute a
rumor, we'll set our incarnation to _one greater_ than the refuted
rumor. This eliminates the possibly-long wait while we repeatedly
refute the same rumor, and should also reduce overall network
traffic. Additionally, it should reduce "incarnation churn" greatly,
since now we will now send out a single refutation for the entire
network, rather than up to _n_ refutations for every member of the
network.

Implementaion-wise, we introduce a new `refute_incarnation` method
alongside the existing `increment_incarnation` method, to reflect two
subtly different scenarios. When we are shutting down, we send a
Departure health rumor out with an incremented incarnation
number. There is no other rumor we are refuting; we're basically
"refuting ourselves". When we are genuinely refuting some other rumor,
though, we call this out in the name (and signature) of the
`refute_incarnation` method.

To eliminate redundant code paths, `increment_incarnation` is now
implemented in terms of `refute_incarnation`.

We also replace the `AddAssign` implementation for Incarnation with
`Add` to reflect the new operations we are doing. There is no reason
to retain `AddAssign`, so we remove it.

### DIFF
--- a/components/butterfly/src/member.rs
+++ b/components/butterfly/src/member.rs
@@ -20,7 +20,7 @@ use std::fmt;
 use std::iter::IntoIterator;
 use std::net::SocketAddr;
 use std::num::ParseIntError;
-use std::ops::AddAssign;
+use std::ops::Add;
 use std::result;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicUsize, Ordering};
@@ -45,6 +45,9 @@ const PINGREQ_TARGETS: usize = 5;
 
 /// Wraps a `u64` to represent the "incarnation number" of a
 /// `Member`. Incarnation numbers can only ever be incremented.
+///
+/// Note: we're intentionally deriving `Copy` to be able to treat this
+/// like a "normal" numeric type.
 #[derive(Clone, Debug, Ord, PartialEq, PartialOrd, Eq, Copy)]
 pub struct Incarnation(u64);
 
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
Looks good but travis tests look unhappy. May have adversely effected other tests.

christophermaier:
Travis failure looked like a transient thing

christophermaier:
Don't merge yet; I think I found a buglet

PR #5676, opened 2018-09-26
Don't show ident for incomplete jobs

chefsalim:
Currently when getting job status, we show an ident even if the job is not successful.  This is an idiosyncrasy of the API that will be fixed in the API eventually, but for now we can patch it on the client side.

tenor-229905964


thesentinels:

Thank you for contributing!

PR #5679, opened 2018-09-27
remove recursive read lock on members from dat file writing

mwrock:
The code comment pretty much says it all. We had a customer predominantly on windows where the supervisor would hang shortly after starting. The supervisor loop was waiting on a read lock on the butterfly data that it was never going to get and was therefore not responding to commands or processing services.

Turns out we had a read lock acquired on the butterfly members and were recursively locking the same data as we iterated each member. Meanwhile, requesting a write lock in another thread blocked and also seemed to block subsequent read locks which lead to a deadlock caused by the outer read lock.

This is rather counter-intuitive given what one would expect from a `RWLock` but diving into some windows specific docs and commentary reveals that this recursive locking can lead to undefined behavior. I have validated with the customer that this change eliminates the hanging behavior.

### DIFF
--- a/components/butterfly/src/rumor/dat_file.rs
+++ b/components/butterfly/src/rumor/dat_file.rs
@@ -368,8 +368,20 @@ impl DatFile {
             .read()
             .expect("Member list lock poisoned");
         for member in members.values() {
-            if let Some(membership) = member_list.membership_for(&member.id) {
-                total += self.write_member(writer, &membership)?;
+            // we do not call membership_for here since that would recursively enter
+            // the members read locj that we are currently in now. One would assume that
+            // is acceptable but this has proven to potentially deadlock Windows threads
+            // in some scenarios. Windows uses a Slim Reader/Writer lock (SRW) and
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -368,8 +368,14 @@ impl DatFile {
             .read()
             .expect("Member list lock poisoned");
         for member in members.values() {
-            if let Some(membership) = member_list.membership_for(&member.id) {
-                total += self.write_member(writer, &membership)?;
+            if let Some(health) = member_list.health_of(member) {
```
christophermaier: I think some comments here stating why we're using this, rather than `member_list.membership_for` will be useful, so some zealous engineer doesn't refactor this fix away sometime in the future.

mwrock: totally! I'm reading docs on the underlying lock implementations on windows in hopes to make a coherent comment.

PR #5686, opened 2018-09-28
Fail if we have more release candidate packages than we should

christophermaier:
If for some reason there are more release candidate packages than can
fit on a single page of API results, we should call this out as
something to fix.

At some point we may legitimately have more than a single page of
results, and then we'll need to change this approach. Of course, at
that point, we'll likely have everything building in Buildkite, and
will have access to much more information.


thesentinels:

Thank you for contributing!

PR #5688, opened 2018-09-28
[studio] Fix errant path in sup-term command

stevendanna:
The extra space resulted in the pid file never being found.

Opening this to document the error.  I see there are some
cool new tests for the studio, I'll see if I can write one for this
this weekend. But, feel free to merge this if you want if I'm not
quick enough.

### DIFF
--- a/components/rootless_studio/default/profile
+++ b/components/rootless_studio/default/profile
@@ -73,7 +73,7 @@ sup-run() {
 }
 
 sup-term() {
-  local pid_file="/hab/sup/default/LOCK "
+  local pid_file="/hab/sup/default/LOCK"
   if [ -f "$pid_file" ]; then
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

stevendanna:
On a related note, I'm wondering if these functions can call `hab sup term` instead?  It seems to do more or less the same thing:

https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L291-L300

But without the lockfile cleanup, which I imagine hab-sup should be doing itself.

stevendanna:
It seems that the expect scripts don't quiet work correctyl on OS X at the moment (at least not with the version of expect I have installed)

Issue #4493, opened 2018-01-24
binlink needs to generate wrappers


themightychris:
When a bin gets binlinked, `PATH` and any arbitrary runtime environment variables defined through `plan.sh`'s new mechanism for setting runtime environment variables are not set. This leads to a fundamental difference between using `hab pkg exec ...` and using a binlink, where `hab pkg exec` is the only way to consistently honor the runtime specifications of `plan.sh`

In what seems like an attempt to work around the `PATH` limitation, `hab pkg install -b` will binlink not just every script in the target package, but every script in every dependency too. This pollutes the bin and only partially addresses the need for binlinked scripts to have complete environments. One major problem with this approach is that binlinking packages that depend on different versions of the same package will lead to a single version of that dependency being used by both.

In order to provide a PATH for every bin script that matches the versions it was built against and provides for arbitrary runtime variables defined in `plan.sh`, some sort of wrapper script will be needed. Perhaps rather than writing symlinks, a wrapper script with PATH and other env variables being sourced or copied from `RUNTIME_ENVIRONMENT` could be generated.

Or, if it is preferable to keep using symlinks that semantically identify what they're pointing at, maybe the scripts could be written to `/hab/bin/command` or `/hab/bin/origin/pkg/version/build/command` and then the bin target symlinked to that

themightychris:
This is related to #2376 but supersedes it in scope

themightychris:
Use case: `csvkit`

Here's a plan for a python application that provides a bunch of command-line tools for manipulating CSVs: https://github.com/JarvusInnovations/habitat-plans/tree/master/csvkit

It is published to `jarvus/csvkit`

`hab pkg exec` works:
```
[1][default:/src:0]# hab pkg exec jarvus/csvkit csvsort --help
usage: csvsort [-h] [-d DELIMITER] [-t] [-q QUOTECHAR] [-u {0,1,2,3}] [-b]
               [-p ESCAPECHAR] [-z FIELD_SIZE_LIMIT] [-e ENCODING] [-L LOCALE]
               [-S] [--blanks] [--date-format DATE_FORMAT]
               [--datetime-format DATETIME_FORMAT] [-H] [-K SKIP_LINES] [-v]
               [-l] [--zero] [-V] [-n] [-c COLUMNS] [-r] [-y SNIFF_LIMIT] [-I]
               [FILE]

Sort CSV files. Like the Unix "sort" command, but for tabular data.

...
```

binlink crashes:
```
[...]

Issue #2376, opened 2017-05-16
hab pkg binlink does not use PATH correctly


smith:
```
[1][default:/src:0]# hab --version
hab 0.23.0/20170511215221
[2][default:/src:0]# hab install chef/inspec
...
[3][default:/src:0]# hab pkg exec chef/inspec inspec version
1.24.0
[4][default:/src:0]# hab pkg binlink chef/inspec inspec --dest /hab/bin
» Symlinking inspec from chef/inspec into /hab/bin
★ Binary inspec from chef/inspec/1.24.0/20170511185749 symlinked to /hab/bin/inspec
[5][default:/src:0]# inspec version
/hab/pkgs/core/coreutils/8.25/20161208223423/bin/env: 'ruby': No such file or directory
```

Expected behavior: It shows the InSpec version.

Running a program with `hab pkg exec` sets the PATH to the path of the `pkg_bin_dirs` of all the transitive dependencies, so this works correctly. `hab pkg binlink` just makes a symlink, so you get the same results as if you would have run `$(hab pkg path chef/inspec)/bin/inspec`.

ShalokShalom:
Confirmed and it exists since at least 0.27: 

![screenshot_20170811_164819](https://user-images.githubusercontent.com/6344099/29218401-0673d7f0-7eb5-11e7-88ed-5e0ac4ec86ed.png)

smith:
@ShalokShalom this is a different issue. In a studio, /bin/ is not on the `$PATH`. If you want to have something symlinked in a studio, use `hab pkg binlink --dest /hab/bin core/pkg-config pkg-config`.

What you're describing is #1680. Hope that helps.

baumanj:
Subsumed by https://github.com/habitat-sh/habitat/issues/4493

PR #5696, opened 2018-10-03
Binlink stubs

jamessewell:
This PR creates a stub script in /hab/binlinks (can be moved) which is used instead of a direct
link when binlinking.

The scripts look like this:

(file: /hab/binlinks/jirotech/node_exporter/release-0.16/20180926010538/bin/node_exporter)

```
if ! [ -x "$(command -v hab)" ]; then
    echo 'ERROR: The core/hab package needs to be binlinked'
    echo
    exit 1
fi
hab pkg exec jirotech/node_exporter/release-0.16/20180926010538 node_exporter "$@"
```

This happens for all binlinks EXCEPT for `core/hab` (which works like before), which must be linked for other binlinks to work.

These changes mean that runtime environment variables which are provided by Habitat are honored for binlinks (exec is the same as running a binlink).

Issues:
- This PR doesn't yet attempt to make Windows work
- This PR will cause binlinks to fail if `core/hab` is not binlinked / in $PATH
- `/hab/binlinks` is probably not the right place, happy to take advice

Signed-off-by: James Sewell <james.sewell@gmail.com>


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/binlink.rs
```diff
@@ -178,20 +193,57 @@ struct Binlink {
...
+    echo
+    exit 1 
+fi 
+hab pkg exec {0} {1} \"$@\"",
+            pkg_ident,
+            bin_name.to_string_lossy()
+        );
+
+        // Need to treat core/hab differently otherwise we get into a link loop
+        match format!("{}/{}", pkg_ident.origin, pkg_ident.name).as_ref() {
```
christophermaier: You shouldn't need to allocate a new string to check if this is `core/hab`... I'd just check if the origin was "core" and the name was "hab", and go from there.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -178,20 +193,57 @@ struct Binlink {
...
 
-        Ok(Self {
-            dest: dest_dir.as_ref().join(bin_name),
-            src: src.as_ref().to_path_buf(),
-        })
+        let template = format!(
+            "\
+#!/bin/sh 
+if ! [ -x \"$(command -v hab)\" ]; then 
+    echo 'ERROR: The core/hab package needs to be binlinked'
```
christophermaier: Doesn't necessarily need to be binlinked, just on `$PATH`, right?

components/eventsrv-protocol/src/message/event.rs
```diff
@@ -2909,45 +2909,37 @@ impl ::protobuf::reflect::ProtobufValue for ServiceUpdate {
...
+    \x0c\x20\x01(\t\x12\x17\n\x0fupdate_strategy\x18\r\x20\x01(\t\x12\x13\n\
+    \x0bapplication\x18\x0e\x20\x01(\t\x12\x13\n\x0benvironment\x18\x0f\x20\
+    \x01(\t\x12\x0e\n\x06leader\x18\x10\x20\x01(\x08\x12\x10\n\x08follower\
+    \x18\x11\x20\x01(\x08\x12\x15\n\rupdate_leader\x18\x12\x20\x01(\x08\x12\
+    \x17\n\x0fupdate_follower\x18\x13\x20\x01(\x08\x12\x1b\n\x13election_is_\
+    running\x18\x14\x20\x01(\x08\x12\x1d\n\x15election_is_no_quorum\x18\x15\
+    \x20\x01(\x08\x12\x1c\n\x14election_is_finished\x18\x16\x20\x01(\x08\x12\
+    \"\n\x1aupdate_election_is_running\x18\x17\x20\x01(\x08\x12$\n\x1cupdate\
+    _election_is_no_quorum\x18\x18\x20\x01(\x08\x12#\n\x1bupdate_election_is\
+    _finished\x18\x19\x20\x01(\x08\
```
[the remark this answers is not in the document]
jamessewell: Odd! Will remove

mwrock:
Seems like we could bring the windows/linux implementations closer together now since this method comes much closer to the windows implementation.

jamessewell:
> Seems like we could bring the windows/linux implementations closer together now since this method comes much closer to the windows implementation.

Yup! The only difference would be the stub payload (and what happens with core/hab I suppose)

components/hab/src/command/pkg/binlink.rs
jamessewell: Fixed

components/hab/src/command/pkg/binlink.rs
jamessewell: Ha, fixed. No idea why I chose this way of doing it. ☕️ needed maybe,

jamessewell:
@fnichol Any comments before I start working on this some more?

themightychris:
Shouldn't it use `exec` at the end?

stevendanna:
Potentially:

Fixes #2376 
Fixes #4493

raskchanky:
@jamessewell I'm so sorry that this PR has sat for as long as it has.  We're making an effort to review things in a more timely manner now, and it's my sincere hope that this kind of situation doesn't happen again.

After looking through the code here and reading the associated issues, specifically https://github.com/habitat-sh/habitat/issues/2376 and https://github.com/habitat-sh/habitat/issues/4493, I think this sounds like a good way forward.  If you wish to continue working on this, feel free, and I'd be happy to review the code when you're done.

mwrock:
@jamessewell if you are tied up and not in a position where you can work on this soon, I'd like to carry the PR through if that is ok with you. I will of course preserve your commits.

jamessewell:
Hi Matt,

Go for it! I was mainly waiting for someone to chime in on the way that
Windows should be handled.

Happy for you to take over.

Cheers,

baumanj:
This might be the right time to address https://github.com/habitat-sh/habitat/issues/4493 as part of this.

jamessewell:
This PR was created to address that issue!

chefsalim:
A few comments - this direction/change looks promising, however, it also looks like it has the potential to be disruptive. Are there any backward compatibility implications?  If so, it would be good to have a plan for handling backward compatibility so that existing installations are not potentially inadvertently broken. It would be good to also have some test cases and/or a test plan + documentation. Finally, it might be worth considering putting this behind a feature flag or something similar so that it can be slowly eased in rather than all at once.

themightychris:
Thinking about backwards compatibility implications, I can only think of three:

- If someone was previous working around the missing environment elements, their workaround might conflict with the issue being fixed
- If someone was parsing the target of binlink symlinks to read package version information, their matching may get broken by the changed path
- If someone was accidentally relying on binlinked commands having a missing environment, its introduction might cause issues

baumanj:
@mwrock: can you let me know when you have this rebased against master? It's a bit tricky to review without being able to build, and I figured you'd need to rebase anyhow.

mwrock:
I am closing this PR in favor of #6670. It preserves the commits of @jamessewell but can't share this PR since it is no longer in his fork.

PR #5701, opened 2018-10-03
patching another recursive member RWLock access

mwrock:
This used to obtain a read lock and then issue a ping which also attempts a lock. Now we grab a cloned member from the initial read and then release the lock before the ping.


thesentinels:

Thank you for contributing!

PR #5702, opened 2018-10-03
Use busybox commands directly in studio setup

jonsmorrow:
* Adds simlinks for all busybox commands in libexec/bin
      so that we can remove the $bb leader from commands when
      setting up studios.
    * Removes Busybox from record function since this env var
      will never be set after executing chroot.
    * Uses functions in set_libexec_path to avoid needing $bb
    * Updates set_libexec_path canonicalization logic to use
      pwd -P which should be more portable.
    * Use hab from PATH (libexec_path) instead of hardcoded
      binary.

    This pr is starting work to make habitat work on mac or other
    platforms that don't have busybox.

    Signed-off-by: Jon Morrow <jmorrow@chef.io>

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -1,8 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 #
 # # Usage
 #
-# ```sh
+# ```bash
 # $ hab-studio [FLAGS] [OPTIONS] <SUBCOMMAND> [ARG ...]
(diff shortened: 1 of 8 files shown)


thesentinels:

Thank you for contributing!

mwrock:
After reviewing our open PRs as a team, we have decided that this may very well be something we take on in the future but are not ready to prioritize at this time. So for now we will close this but may indeed reopen in the future.

PR #5706, opened 2018-10-04
[hab-sup] Copy binding_mode when transforming Service to ServiceSpec

### LINKED ISSUE #5703: Erroneous service changes for services with relaxed binding mode
When updating a service, hab-sup restarts other services that have been loaded with a binding mode of "relaxed".  I believe I have tracked this down.  Here is the cause:

1) The to_spec function does not copy the binding mode into the generated spec: https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/mod.rs#L389-L408

2) The to_spec function is called when generating the "active_specs" to pass to the spec watcher: https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L1657

3) The spec watcher checks for differences between all active and desired specs in reaction to any filesystem change.  Since the binding mode field wasn't copied, it is erroneously detected as changed and is restarted.

stevendanna:
Failing to copy the binding mode leads the spec_watcher to erroneously
detect changes for binding_mode-relaxed services.

Fixes #5703

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -400,6 +400,7 @@ impl Service {
         spec.topology = self.topology;
         spec.update_strategy = self.update_strategy;
         spec.binds = self.binds.clone();
+        spec.binding_mode = self.binding_mode;
         spec.config_from = self.config_from.clone();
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #5713, opened 2018-10-05
Make logging around user.toml changes more specific

christophermaier:
Previously, no indication was given as to *why* a service was
restarting here. Additionally, the service doesn't necessarily have to
restart; that will only if the `user.toml` changes result in any
changes to templated files.


thesentinels:

Thank you for contributing!

PR #5716, opened 2018-10-05
Don't spam the Supervisor logs quite so much

christophermaier:
This brings a little bit of sanity to our logging.

Two large "spam" offenders are the "Hooks compiled" and "The group X satisfies the Y bind" lines. The former was emitted any time we processed hook files, and not just when a hook file actually changed. Similarly, the bind satisfaction line was emitted every time the census changed and bind satisfaction was re-evaluated, whether they changed or not.

Now we only log about these things when something actually changes.

Additionally, the content of the logging around these has been modified a bit for clarity. As always, see the commit messages for further details.

tenor-37136541


thesentinels:

Thank you for contributing!

PR #5718, opened 2018-10-05
Remove "maintenance window" step from release pipeline

christophermaier:
We had a bug that complicated the rollout of new Supervisors to
Builder. To compensate for that until we were able to track it down,
we would declare maintenance windows for the service, just in case we
had an issue with the release.

Thankfully, that bug is now fixed (see #5588, #5616, and #5673), so we
no longer need to do this!

We'll keep the blocking step between the updating of Builder and the
promotion of release artifacts to the stable channel, as one last
safety valve, just in case. Better to have it and not need it, than to
need it and not have it.


thesentinels:

Thank you for contributing!

PR #5720, opened 2018-10-06
Don't change svc directory permissions except when the directory is created

### LINKED ISSUE #4475: Have the Supervisor be less pedantic re: service directory permissions
The tale began with #4366 and continued with habitat-sh/core-plans#1145.

After some face-to-face discussion among the Habitat team, we determined that the underlying issue in these cases is that the Supervisor changes service directory permissions when it starts up. Normally this is fine, but in the case of Supervisor upgrades, it has the effect of potentially changing things under an already-running process. This likely became an issue when the Launcher was introduced, since before that, there was never a scenario in which a supervised process would already be running when the Supervisor started.

If the Supervisor only tries to change permissions on service directories when it creates them (but not when they already exist), this should allow users more leeway to customize things (e.g., in run hooks) without having to worry about the Supervisor undoing that work when it upgrades.

jaym:
When hab-sup is restarted, it used to change the directory permissions
to the default ones that it expected. This is problematic for things
like nginx because nginx itself runs as root, however the hooks set up
permissions so that the children don't have to run as root. If hab-sup
was restarted, these directory permissions would be reset and the hooks
themselves don't necessarily run.

Fixes #4475

### DIFF
--- a/components/sup/src/manager/service/dir.rs
+++ b/components/sup/src/manager/service/dir.rs
@@ -116,6 +116,13 @@ impl<'a> SvcDir<'a> {
     where
         P: AsRef<Path>,
     {
+        // We do not want to change the permissions of an already
+        // existing directory
+        // See https://github.com/habitat-sh/habitat/issues/4475
+        if path.as_ref().exists() {
+            return Ok(());
+        }
+
         Self::create_dir_all(&path)?;
         if cfg!(not(windows)) {
             if abilities::can_run_services_as_svc_user() {
--- a/test/integration/sup-hup.bats
+++ b/test/integration/sup-hup.bats
@@ -1,0 +1,54 @@
+#!/usr/bin/env bats
+
+load 'helpers'
+
+setup() {
+    reset_hab_root
+}
+
+teardown() {
+    stop_supervisor
+}
+
+sup_restarted() {
+    local old_pid="${1}"
+    local new_pid="$(pgrep 'hab-sup')"
+    if [[ "$old_pid" -eq "$new_pid" ]]; then
+        return 0
+    fi
+    return 1
+}
+
+@test "supervisor: restart does not chown directories" {
+    background ${hab} run
+    sleep 2 # give it a sec to come up
+
+    ${hab} pkg install core/runit --binlink
+
+    # start up nginx
+    ${hab} pkg install core/nginx
+    ${hab} svc load core/nginx
+    wait_for_service_to_run nginx
+
+    # create an index.html so there is a page to fetch
+    echo "test" > /hab/svc/nginx/data/index.html
+
+    # the nginx children (running as hab) should all have access 
+    # to the index.html at this point
+    run curl -s -o /dev/null -w "%{http_code}" http://localhost
+    [ "$output" = "200" ]
+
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/dir.rs
```diff
@@ -136,7 +147,11 @@ impl<'a> SvcDir<'a> {
         }
     }
 
-    fn create_dir_all<P: AsRef<Path>>(path: P) -> Result<()> {
+    fn create_dir_all<P: AsRef<Path>>(path: P) -> Result<CreateDirectoryStatus> {
```
baumanj: `create_dir_all` has quite a few callers, but if only this one case cares about this newly added return type, I'd rather leave `create_dir_all` alone and add the `exists()` check to the caller side. It should be less code and complexity and eliminates the need for the new `CreateDirectoryStatus` type (though I do really appreciate the use of the more readable `enum` over a bool).

I think the change could be as simple as
```rust
    fn create_svc_owned_dir<P>(&self, path: P) -> Result<()>
    where
        P: AsRef<Path>,
    {
        if path.as_ref().exists() {
            Ok(())
        } else {
            // Original body of create_svc_owned_dir
        }
    }
```

components/sup/src/manager/service/dir.rs
```diff
@@ -116,6 +116,12 @@ impl<'a> SvcDir<'a> {
     where
         P: AsRef<Path>,
     {
+        // We do not want to change the permissions of an already
+        // existing directory (on linux)
```
christophermaier: This comment implies that something else would be done on Windows, but the code doesn't reflect that. 

Is the code right, or is the comment?

components/sup/src/manager/service/dir.rs
christophermaier: I agree with @baumanj here, but particularly on the use "custom boolean enum" pattern; that's one of my favorite things, and I'd like to see a lot more of it!
tenor-82140374

components/sup/src/manager/service/dir.rs
jaym: window's doesn't do anything with permissions. I can get rid of it if it's confusing

christophermaier: Yeah, let's go ahead and remove the "(on linux)" bit. Thanks!

run-bats.sh
```diff
@@ -14,4 +14,4 @@ docker run -it --rm \
        --env HAB_BIN_DIR=/bin \
        --workdir=/test \
        hab-bats-cleanroom \
-       bats .
```
christophermaier: Let's leave it as `bats .` for now.

test/integration/sup-hup.bats
```diff
@@ -24,7 +24,7 @@ sup_restarted() {
     sleep 2 # give it a sec to come up
 
     # core/nginx seems to implicitly depend on chpst
-    ${hab} pkg install core/runit --binlink
+    ${hab} pkg install core/hab-sup
```
christophermaier: The comment no longer makes sense here.

PR #5722, opened 2018-10-08
Skip Appveyor if only the docs are updated

btm:
We've had some periods of builds queueing up on Appveyor lately.
This change will keep Appveyor from running if only docs are modified.
This will also avoid running against the Expeditor commits against the
`CHANGELOG.md` that happens with every merge.

Signed-off-by: Bryan McLellan <btm@loftninjas.org>

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,6 +19,12 @@ init:
 
 clone_folder: c:\projects\habitat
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #5728, opened 2018-10-09
Corrects api_client response on hab bldr job start

### LINKED ISSUE #5694: Output of successful `hab bldr job start` looks failure-esqe
When I submit a job with `hab bldr job start` and the request succeeds, the output looks like something went wrong:

image

It should be greener and happier-looking. 💚

eeyun:
Fixes #5694 

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -398,7 +398,7 @@ impl Client {
         };
         match result {
             Ok(response) => {
-                if response.status == StatusCode::Ok {
+                if response.status == StatusCode::Created || response.status == StatusCode::Ok {
                     let sr: SchedulerResponse = decoded_response(response)?;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-client/src/lib.rs
```diff
@@ -398,7 +398,7 @@ impl Client {
         };
         match result {
             Ok(response) => {
-                if response.status == StatusCode::Ok {
+                if response.status == StatusCode::Created {
```
chefsalim: Would be better to check for either Ok or Created in order to be backward compatible (on-prem instances running pre-Actix Builder)

eeyun: Ah I was confused by what you meant by that. I'll get that added.

PR #5730, opened 2018-10-09
Adds upload of public keys for deps as well as target package

eeyun:

### DIFF
--- a/components/hab/src/command/pkg/upload.rs
+++ b/components/hab/src/command/pkg/upload.rs
@@ -69,40 +69,10 @@ where
     U: AsRef<Path>,
 {
     let mut archive = PackageArchive::new(PathBuf::from(archive_path.as_ref()));
-    let hart_header = get_artifact_header(&archive_path.as_ref())?;
-    let public_keyfile_name = format!("{}.pub", &hart_header.key_name);
-    let public_keyfile = key_path.as_ref().join(&public_keyfile_name);
 
-    ui.status(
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/upload.rs
```diff
@@ -293,3 +269,43 @@ fn attempt_upload_dep(
...
+    let public_keyfile = key_path.as_ref().join(&public_keyfile_name);
+
+    let (name, rev) = parse_name_with_rev(&hart_header.key_name)?;
+
+    match api_client.put_origin_key(&name, &rev, &public_keyfile, token, ui.progress()) {
+        Ok(()) => {
+            ui.begin(format!(
+                "Uploading public origin key {}",
+                &public_keyfile_name
+            ))?;
```
baumanj: Previously, this message printed before the call to `api_client.put_origin_key`. Is that an intentional change? Either way seems reasonable to me.

eeyun: yes this was intentional. There might be a cleaner way of handling it, but since we're uploading each dep's key now, that message would get printed N+1 times even if we didn't end up uploading due to the key already existing. Now it only gets printed if we've actually uploaded the key. We could tweak this further but i know @fnichol has some plans for changing this behavior quite a bit more so I figured this is fine for now.

Issue #5361, opened 2018-07-19
Unsafe, platform-specific Process struct is obviated by rust stdlib now


baumanj:
At the time we added [`try_wait`](https://github.com/habitat-sh/habitat/blame/master/components/launcher/src/sys/unix/service.rs#L88-L101) to the launcher, the rust standard library lacked the functionality we needed and so we had to implement it ourselves. However, it requires both unsafe code and two platform-specific implementations as the process models for Linux and Windows is fairly different.

Since rust added [`try_wait`](https://doc.rust-lang.org/std/process/struct.Child.html#method.try_wait) with release [1.18.0](https://blog.rust-lang.org/2017/06/08/Rust-1.18.html) in June 2017, we now have all the primitives we need in the `std::process::Child` struct which is provided by the standard library.

We should remove our custom [`Process`](https://github.com/habitat-sh/habitat/blob/master/components/launcher/src/sys/unix/service.rs#L35) struct in favor of using standard library implementations.

mwrock:
I believe `kill` was another specialized implementation where we initially attempt to gracefully terminate the process before resorting to killing it. We also have some logic to determine if we need to send the signal to the process group.

fnichol:
In general, yeah!! This comes up as we often need an implementation before APIs stabilize in the Rust standard library. Can you imagine how great it'll be when [`TryConvert`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) finally stabilizes? Yes!

Having said that, we'll want to make sure we cover our basis to make sure that we have the functionality and features as @mwrock alluded to above.

fnichol:
Oh, I should say that in the past we've removed code in favor of newly stabilized Rust APIs and occasionally crate updates as well, so I'd be for looking at swapping out these implementations. Also, I want to keep this practice moving forward 😄

mwrock:
Yes! Love removing code!!

baumanj:
Good point, @mwrock. I'll definitely make sure to preserve any special logic we have. I think the kill one is the most special, but comparing the others (`wait`, `try_wait`, etc.) our code looks almost identical to the rust stdlib.

PR #5733, opened 2018-10-10
Replace launcher's Process struct's unsafe implementations with rust stdlib

baumanj:
Now that try and try_wait exist, Process can be a thin wrapper around
std::process::Child on unix. Unfortunately, the same isn't currently feasible
on Windows, since the process spawning depends on additional features not
available in the rust stdlib currently.

Also, fix some typos.

Resolves https://github.com/habitat-sh/habitat/issues/5361

### DIFF
--- a/components/launcher/src/server/handlers/restart.rs
+++ b/components/launcher/src/server/handlers/restart.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use core::os::process::Pid;
 use protocol;
 
 use super::{HandleResult, Handler};
@@ -25,7 +24,7 @@ impl Handler for RestartHandler {
     type Reply = protocol::SpawnOk;
 
     fn handle(msg: Self::Message, services: &mut ServiceTable) -> HandleResult<Self::Reply> {
-        let mut service = match services.remove(msg.get_pid() as Pid) {
+        let mut service = match services.remove(msg.get_pid() as u32) {
             Some(service) => service,
             None => {
                 let mut reply = protocol::NetErr::new();
@@ -44,7 +43,11 @@ impl Handler for RestartHandler {
                 }
                 Err(err) => Err(protocol::error(err)),
             },
-            Err(err) => Err(protocol::error(err)),
+            Err(_) => {
+                let mut reply = protocol::NetErr::new();
+                reply.set_code(protocol::ErrCode::ExecWait);
+                Err(reply)
+            }
         }
     }
 }
--- a/components/launcher/src/service.rs
+++ b/components/launcher/src/service.rs
@@ -20,16 +20,13 @@ use std::thread;
 
 #[cfg(windows)]
 use core::os::process::windows_child::{ChildStderr, ChildStdout, ExitStatus};
-use core::os::process::Pid;
 use protocol;
 
-use error::Result;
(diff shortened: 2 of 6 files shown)


thesentinels:

Thank you for contributing!

components/launcher/src/service.rs
```diff
@@ -20,16 +20,13 @@ use std::thread;
...
-use core::os::process::Pid;
 use protocol;
 
-use error::Result;
 pub use sys::service::*;
 
 pub struct Service {
     args: protocol::Spawn,
     process: Process,
-    status: Option<ExitStatus>,
```
baumanj: Note this field was never mutated, so removing it should be safe

components/launcher/src/server/handlers/restart.rs
```diff
@@ -25,7 +24,7 @@ impl Handler for RestartHandler {
     type Reply = protocol::SpawnOk;
 
     fn handle(msg: Self::Message, services: &mut ServiceTable) -> HandleResult<Self::Reply> {
-        let mut service = match services.remove(msg.get_pid() as Pid) {
+        let mut service = match services.remove(msg.get_pid() as u32) {
```
baumanj: I like descriptive types as much as the next person, but [`std::process::Child::id()`](http://doc.rust-lang.org/1.29.1/std/process/struct.Child.html#method.id) returns a `u32`, so I think it's preferable to align with the stdlib. This ripples out in a number of places, but also requires some casts since certain `libc` calls take `i32`.

components/launcher/src/sys/unix/service.rs
```diff
@@ -85,33 +75,12 @@ impl Process {
...
-        match unsafe { libc::waitpid(self.pid, &mut status, libc::WNOHANG) } {
-            0 => Ok(None),
-            -1 => Err(Error::ExecWait(io::Error::last_os_error())),
-            _ => {
-                self.status = Some(ExitStatus::from_raw(status));
-                Ok(Some(ExitStatus::from_raw(status)))
-            }
-        }
+    pub fn try_wait(&mut self) -> io::Result<Option<ExitStatus>> {
+        self.0.try_wait()
```
baumanj: As you can see [the stdlib implementation is nearly identical](https://github.com/rust-lang/rust/blob/b7c6e8f1805cd8a4b0a1c1f22f17a89e9e2cea23/src/libstd/sys/unix/process/process_unix.rs#L390).

components/launcher/src/sys/unix/service.rs
```diff
@@ -85,33 +75,12 @@ impl Process {
...
-        let mut status = 0 as c_int;
-        match unsafe { libc::waitpid(self.pid, &mut status, 0) } {
-            -1 => Err(Error::ExecWait(io::Error::last_os_error())),
-            _ => {
-                self.status = Some(ExitStatus::from_raw(status));
-                Ok(ExitStatus::from_raw(status))
-            }
-        }
+    pub fn wait(&mut self) -> io::Result<ExitStatus> {
+        self.0.wait()
```
baumanj: [stdlib for comparison](https://github.com/rust-lang/rust/blob/b7c6e8f1805cd8a4b0a1c1f22f17a89e9e2cea23/src/libstd/sys/unix/process/process_unix.rs#L379)

PR #5737, opened 2018-10-10
Add 'hab pkg uninstall' command

jamesc:
This is the inverse of `hab pkg uninstall`.  given a package ident it will safely delete the package and it's dependencies.  What does safely mean? - the command will not delete any package which is a dependency of another package currently on disk.

There is also a 'dry-run' mode (`-d`) which will output what packages would be deleted.

If no version is supplied it deletes the latest package.

Some example outputs:
* Dry run of `pkg uninstall`
``` bash
$ sudo hab pkg  uninstall core/mysql
» Uninstalling package core/mysql/5.7.21/20180609181124
☒ Deleting core/mysql/5.7.21/20180609181124
☒ Deleting core/sed/4.4/20180608091938
☒ Deleting core/procps-ng/3.3.12/20180608092056
∅ Skipping package core/perl/5.26.1/20180608094208. It is a dependency of 2 packages
∅ Skipping package core/pcre/8.41/20180608092740. It is a dependency of 2 packages
∅ Skipping package core/openssl/1.0.2n/20180608102213. It is a dependency of 2 packages
∅ Skipping package core/ncurses/6.1/20180608091810. It is a dependency of 4 packages
☒ Deleting core/inetutils/1.9.4/20180608094116
☒ Deleting core/grep/3.1/20180608092809
∅ Skipping package core/glibc/2.27/20180608041157. It is a dependency of 27 packages
∅ Skipping package core/gcc-libs/7.3.0/20180608091701. It is a dependency of 5 packages
☒ Deleting core/gawk/4.2.0/20180608093856
∅ Skipping package core/coreutils/8.29/20180608092141. It is a dependency of 1 packages
∅ Skipping package core/acl/2.2.52/20180608091922. It is a dependency of 1 packages
∅ Skipping package core/zlib/1.2.11/20180608050617. It is a dependency of 5 packages
∅ Skipping package core/less/530/20180608094202. It is a dependency of 1 packages
∅ Skipping package core/gdbm/1.14.1/20180608094002. It is a dependency of 1 packages
∅ Skipping package core/db/5.3.28/20180608094030. It is a dependency of 1 packages
∅ Skipping package core/bzip2/1.0.6/20180608091727. It is a dependency of 1 packages
∅ Skipping package core/cacerts/2018.03.07/20180608102212. It is a dependency of 1 packages
∅ Skipping package core/libcap/2.25/20180608091936. It is a dependency of 1 packages
∅ Skipping package core/linux-headers/4.15.9/20180608041107. It is a dependency of 1 packages
☒ Deleting core/mpfr/4.0.1/20180608051629
∅ Skipping package core/gmp/6.1.2/20180608051426. It is a dependency of 1 packages
∅ Skipping package core/attr/2.4.47/20180608091914. It is a dependency of 3 packages
★ Deleted package core/mysql/5.7.21/20180609181124 and 6 dependencies
```
* Uninstall a package
``` bash 
$ sudo hab pkg  uninstall core/mysql
» Uninstalling package core/mysql/5.7.21/20180609181124
☒ Deleting core/mysql/5.7.21/20180609181124
☒ Deleting core/sed/4.4/20180608091938
☒ Deleting core/procps-ng/3.3.12/20180608092056
∅ Skipping package core/perl/5.26.1/20180608094208. It is a dependency of 2 packages
∅ Skipping package core/pcre/8.41/20180608092740. It is a dependency of 2 packages
∅ Skipping package core/openssl/1.0.2n/20180608102213. It is a dependency of 2 packages
∅ Skipping package core/ncurses/6.1/20180608091810. It is a dependency of 4 packages
☒ Deleting core/inetutils/1.9.4/20180608094116
☒ Deleting core/grep/3.1/20180608092809
∅ Skipping package core/glibc/2.27/20180608041157. It is a dependency of 27 packages
∅ Skipping package core/gcc-libs/7.3.0/20180608091701. It is a dependency of 5 packages
☒ Deleting core/gawk/4.2.0/20180608093856
∅ Skipping package core/coreutils/8.29/20180608092141. It is a dependency of 1 packages
∅ Skipping package core/acl/2.2.52/20180608091922. It is a dependency of 1 packages
∅ Skipping package core/zlib/1.2.11/20180608050617. It is a dependency of 5 packages
∅ Skipping package core/less/530/20180608094202. It is a dependency of 1 packages
∅ Skipping package core/gdbm/1.14.1/20180608094002. It is a dependency of 1 packages
∅ Skipping package core/db/5.3.28/20180608094030. It is a dependency of 1 packages
∅ Skipping package core/bzip2/1.0.6/20180608091727. It is a dependency of 1 packages
∅ Skipping package core/cacerts/2018.03.07/20180608102212. It is a dependency of 1 packages
∅ Skipping package core/libcap/2.25/20180608091936. It is a dependency of 1 packages
∅ Skipping package core/linux-headers/4.15.9/20180608041107. It is a dependency of 1 packages
☒ Deleting core/mpfr/4.0.1/20180608051629
∅ Skipping package core/gmp/6.1.2/20180608051426. It is a dependency of 1 packages
∅ Skipping package core/attr/2.4.47/20180608091914. It is a dependency of 3 packages
★ Deleted package core/mysql/5.7.21/20180609181124 and 6 dependencies
```
* try to uninstall a package which is a dependency
``` bash
$ hab pkg  uninstall core/attr
» Uninstalling package core/attr/2.4.47/20180608091914
✗✗✗
✗✗✗ Can't remove package: core/attr/2.4.47/20180608091914. It is a dependency of 3 packages
✗✗✗
```

* remove a package which has no dependencies
``` bash
hab pkg  uninstall core/linux-headers/4.3/20160612063537
» Uninstalling package core/linux-headers/4.3/20160612063537
☒ Deleting core/linux-headers/4.3/20160612063537
★ Deleted package core/linux-headers/4.3/20160612063537 and 0 dependencies
```

* remove only the package, but not the dependencies 
``` bash
hab pkg uninstall core/postgresql96 --no-deps
» Uninstalling core/postgresql96/9.6.9/20180716152137
☒ Deleting core/postgresql96/9.6.9/20180716152137
… Skipping dependencies (--no-deps specified)
★ Uninstall of core/postgresql96/9.6.9/20180716152137 and 0 dependencies complete

```


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,131 @@
...
+fn maybe_delete(ui: &mut UI, install: &PackageInstall, dryrun: bool) -> Result<bool> {
+    let ident = install.ident();
+    if dryrun {
+        ui.status(Status::DryRunDeleting, &ident)?;
+        return Ok(false);
+    }
+
+    ui.status(Status::Deleting, &ident)?;
+    let pkg_dir = install.installed_path();
+    fs::remove_dir_all(&pkg_dir)?;
```
stevendanna: On linux at least, this isn't atomic so concurrent `hab` users might end up trying to execute a half-delete package, causing a confusing error.  While we obviously can't avoid such errors completely, we could rename and then delete for a little bit more safety here.

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,131 @@
...
+use error::{Error, Result};
+use hcore::fs as hfs;
+use hcore::package::{all_packages, PackageIdent, PackageInstall};
+
+//use std::iter::IntoIterator;
+
+///
+/// Delete a package and all dependencies which are not used by other packages
+///
+/// 1. We find the fully qualified package ident and all it's dependencies
```
stevendanna: [nit] s/it's/its/

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,131 @@
...
+use common::package_graph::PackageGraph;
+use common::ui::{Status, UIWriter, UI};
+use error::{Error, Result};
+use hcore::fs as hfs;
+use hcore::package::{all_packages, PackageIdent, PackageInstall};
+
+//use std::iter::IntoIterator;
+
+///
+/// Delete a package and all dependencies which are not used by other packages
```
stevendanna: 💭 Should we also be checking wehther this particular package is currently loaded as a service in the supervisor?

baumanj:
It looks like the travis tests are failing. I haven't looked at the code enough yet to say whether it's related to your changes, but can you successfully run the tests locally?

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    ui.status(Status::Deleting, &ident)?;
+    let pkg_dir = install.installed_path();
+    fs::remove_dir_all(&pkg_dir)?;
+    Ok(true)
+}
+
+fn load(fs_root_path: &Path, idents: &Vec<PackageIdent>) -> Result<Vec<PackageInstall>> {
+    let mut result = Vec::<PackageInstall>::new();
```
baumanj: Might as well use [`Vec::with_capacity`](http://doc.rust-lang.org/1.29.1/alloc/vec/struct.Vec.html#method.with_capacity) since we know the final size we expect.

Also, the `::<PackageInstall>` can be omitted; the compiler can infer the type of `Vec` you need from context in this case.

Also, since the return value of `all_packages` doesn't seem to be used other than as an argument to this function, why not call it internally and change this function to `load_all_packages` or something?

jamesc:
@baumanj Yep, tests pass locally. I've some more changes so let's see if it is transient

components/hab/src/command/pkg/uninstall.rs
jamesc: Good point.  I had been going back and forward on if I should change all_packages into a `HashMap` to make the lookup above faster.  Probably simplest to leave as `Vec` for now and see if it becomes a bottleneck

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,131 @@
...
+                    "Skipping package {}. It is a dependency of {} packages",
+                    p.clone(),
+                    c
+                ))?;
+            }
+        }
+    }
+
+    if dryrun {
+
```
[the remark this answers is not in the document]
jamesc: +1

components/hab/src/command/pkg/uninstall.rs
jamesc: I was going to add some allow/prohibit functionality next:
* don't delete `hab` :)
* don't delete running services
* don't delete anything specified on CLI or in a ignore file

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,131 @@
...
+    let mut graph = PackageGraph::new();
+    graph.build(&package_installs)?;
+
+    // 3.
+    let deps = graph.ordered_deps(&ident);
+
+    // 4.
+    match graph.count_rdeps(&ident) {
+        None => (), // package not in graph - m
```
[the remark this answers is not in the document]
jamesc: If this fails, it's related to the possible race issues @stevendanna  mentions above.  I'll go with a message right now for these cases and we can revisit if we come to decision around locks for `/hab/pkgs` which is a much bigger job

jamesc:
@baumanj Good points on the semantics of an uninstall.  I took #3499 as a starting point for requirements and want to add a few features to address different aspects. 

1) I see uninstall as the tool that could be used for a semi-manual CLI approach to deleting packages that is pretty generic. 

2) A lot of the same underlying code can be used in the supervisor to implement the 'keep the last X versions' logic that folks want.
3) Finally we need a `hab pkg purge` which is a bit more like what you want (I look to docker images purge as an example - delete everything not used by a running service).

I want `hab pkg uninstall` to be the reverse of `hab pkg install` as much as possible.  i.e doing an install and then an uninstall should leave you in the same place where possible.  Of course your example of a package you've installed that only has the new package as a dependant does break this model. But if it's in your artifact cache it should be quick to re-inflate.

you have a good usecase in terms of 'just delete this one package'.  I think adding a --no-deps option would meet the needs of that.

baumanj:
Good point about the package cache. In that case, I think I'm more ok with this approach. It's a tricky problem to have install/uninstall be perfect mirrors of each other, but I think removing the dependencies that had only the removed package as a dependent will be the expected behavior the vast majority of the time.

Providing a `--no-deps` option when one is merely trying to clean up a botched install sounds like a good middle ground.

components/hab/src/cli.rs
```diff
@@ -430,6 +430,14 @@ pub fn get() -> App<'static, 'static> {
                     "The destination path to the signed Habitat Artifact \
                     (ex: /home/acme-redis-3.0.7-21120102031201-x86_64-linux.hart)")
             )
+            (@subcommand uninstall =>
+                (about: "Uninstall a package and dependencies from the local filesystem")
+                (aliases: &["un", "unin"])
+                (@arg PKG_IDENT: +required +takes_value
+                    "A package identifier (ex: core/busybox-static/1.42.2/21120102031201)")
+                (@arg DRYRUN: -d --dryrun "Just show what would be uninstalled, don't actually do it")
+                (@arg NO_DEPS: --("no-deps") "Down uninstall dependencies")
```
christophermaier: "Down" -> "Don't"

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    // 3.
+    let deps = graph.ordered_deps(&ident);
+
+    // 4.
+    match graph.count_rdeps(&ident) {
+        None => {
+            // package not in graph - this shouldn't happen but could be a race condition with another hab uninstall. We can continue as
+            // what we wanted (package to be removed) has already happened
+            ui.warn(format!("Tried to find dependencies of {} but it wasn't in graph.  Maybe another uninstall command was run at the same time?", &ident))?;
```
christophermaier: Might be worth saying explicitly (even if just in comments) that we're going to continue because there may be dependencies (which we've already computed at this point) that we'll need to delete, too.

christophermaier: Also, wouldn't this be searching for "dependents on" rather than "dependencies of"?

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,140 @@
...
+    let mut count = 0;
+    if no_deps {
+        ui.status(Status::Skipping, "dependencies (--no-deps specified)")?;
+    } else {
+        for p in &deps {
+            match graph.count_rdeps(&p) {
+                None => {
+                    // package not in graph - this shouldn't happen but could be a race condition with another hab uninstall. We can continue as
+                    // what we wanted (package to be removed) has already happened
+                    ui.warn(format!("Tried to find dependencies of {} but it wasn't in graph.  Maybe another uninstall command was run at the same time?", &p))?;
```
christophermaier: Ditto

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,140 @@
...
+/// 5. For each dependency we check if there are any packages which depend on it
+///     5a. If there are, we skip it
+///     5b. If there are not, we delete it from disk and the graph
+///
+pub fn start(
+    ui: &mut UI,
+    ident: &PackageIdent,
+    fs_root_path: &Path,
+    dryrun: bool,
+    no_deps: bool,
```
christophermaier: One thing we've been doing more of lately is trying to use enums rather than bare booleans as arguments, with the idea being that they're more readable, documentable, etc.

So, for example, maybe instead of a boolean `no_deps` we could have something like:
```rust
enum DeletionStrategy {
  PackageOnly
  PackageAndDependencies
}
```
I admit that naming can sometimes be tricky here, but it's proven to be a useful approach in other areas of the codebase, particularly when taking more than one boolean as an argument to a function (then, you get type-safety benefits, too).

jamesc: It makes the code a lot nicer too in terms of reading the logic of the match clauses.  Thanks!

jamesc:
@baumanj @christophermaier @stevendanna All review comments now addressed. If this looks good, I'd sqaush it down and push

jaym:
Has there been any discussion about keeping track of what packages were explicitly installed so that we can know whether something was actually a dependency?

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -0,0 +1,199 @@
...
+                        // continue as what we wanted (package to be removed) has already happened. We're going to continue and try and delete down through the
+                        // dependency tree anyway
+                        ui.warn(format!("Tried to find dependant packages of {} but it wasn't in graph.  Maybe another uninstall command was run at the same time?", &p))?;
+                    }
+                    Some(0) => {
+                        let install = packages.iter().find(|&i| i.ident() == p).unwrap();
+                        if let Ok(true) =
+                            maybe_delete(ui, &fs_root_path, &install, &execution_strategy)
+                        {
+                        }
```
christophermaier: What are these empty braces supposed to do? Are the following two lines supposed to be inside them?

christophermaier:
@jaym I think this is going to be the first step toward supporting a variety of potential uninstallation scenarios. The functionality here is basic stuff they'd all likely share.

components/hab/src/command/pkg/uninstall.rs
jamesc: It was an unneeded `if let` and `rust fmt` added them.  Changed the code to not use the if let.

PR #5740, opened 2018-10-12
Shelving composites post

mgamini:
tenor-239046481

Needs:
- Version number

Signed-off-by: Garrett Amini <kadjar@gmail.com>


thesentinels:

Thank you for contributing!

nellshamrell:
Something that should be addressed - what happens to currently existing composite plans?  Will they stop working with the next version of Habitat?

christophermaier:
@nellshamrell The code is still in there, so things will continue to work. We just need to determine when we're going to take the code out and give that version in this blog post.

It will _not_ be the next version, though... I anticipate we'll give at least 1 or 2 releases-worth of a heads-up.

www/source/blog/2018-10-12-shelving-composites.html.md
```diff
@@ -0,0 +1,22 @@
...
+
+## What's Happening Now?
+
+Starting with version *X.XX.X*, we will be removing the ability to run composite plans from Habitat. We've learned enough over the past year to realize that we need a complete re-implementation of how that feature will work with the Supervisor, so we don't want to lead our users down a dead-end path. This is one of those cases where rebuilding the feature will be much faster to implement than refactoring what we have, so we are taking the hard step of removing the current implementation now.
+
+## The Path Forward
+
+As we near 1.0, we are seeking to make Habitat as robust and reliable as possible. Once we achieve that milestone, we're looking forward to once again adding the ability to create composite packages, developed with help and feedback from the community.
+
+If you have feedback that will help us drive this in the right direction, we'd love to hear from you [in the forums](https://forums.habitat.sh). Happy Habitizing!
```
christophermaier: It might be a good idea to start a forum post on this, and give the link to that post here (rather than a link to the forums in general), just as a way to centralize feedback.

www/source/blog/2018-10-12-shelving-composites.html.md
```diff
@@ -0,0 +1,22 @@
...
+
+This past year has taught us that while we were on the right track with the idea of composite packages, our implementation left much to be desired. Users kept trying to utilize composite packages, but would inevitably find themselves bumping up against limitations of the feature, and would thus end implementing the same sorts of creative solutions we were trying to simplify in the first place.
+
+## What's Happening Now?
+
+Starting with version *X.XX.X*, we will be removing the ability to run composite plans from Habitat. We've learned enough over the past year to realize that we need a complete re-implementation of how that feature will work with the Supervisor, so we don't want to lead our users down a dead-end path. This is one of those cases where rebuilding the feature will be much faster to implement than refactoring what we have, so we are taking the hard step of removing the current implementation now.
```
christophermaier: Rather than say it's easier to rebuild rather than refactor (which is often a rash decision), I'd frame this more as "we put out a prototype and have learned a lot from it; we're now going to make the real feature". 

With that knowledge, we're going to take out the existing code, which should simplify our current stability and push-to-1.0 work.

christophermaier: It would also be good to really emphasize that the code is _not being removed right now_, and there will be ample time for any users of composites to consider their next moves.

www/source/blog/2018-10-12-shelving-composites.html.md
```diff
@@ -0,0 +1,22 @@
...
+
+## The Path Forward
+
+As we near 1.0, we are seeking to make Habitat as robust and reliable as possible. Once we achieve that milestone, we're looking forward to once again adding the ability to create composite packages, developed with help and feedback from the community.
```
christophermaier: It may be that we don't actually have a different kind of package, but provide ways of limiting binds to the same supervisor, rather than going out over the network.

So (without saying that in the post), I'd wouldn't say that we're going to "re-implement composite packages", but rather that "we'll be looking at how to satisfy the usecases that composite packages were meant to serve" (or words to that effect).

And definitely with feedback from the community. 👍

www/source/blog/2018-10-12-shelving-composites.html.md
```diff
@@ -0,0 +1,22 @@
...
+
+This past year has taught us that while we were on the right track with the idea of composite packages, our implementation left much to be desired. Users kept trying to utilize composite packages, but would inevitably find themselves bumping up against limitations of the feature, and would thus end implementing the same sorts of creative solutions we were trying to simplify in the first place.
+
+## What's Happening Now?
+
+We will soon be removing the ability to run composite plans from Habitat. In the short term, this will simplify our roadmap to 1.0 and increase critical code stability. We have learned much over the past year from our initial implementation, and based upon that feedback and with an eye toward hearing more from the community, we're looking forward to being able to build new features to accomplish the same things that composite plans were designed to do. 
+
+## The Path Forward
+
+As we near 1.0, we are seeking to make Habitat as robust and reliable as possible. Once we achieve that milestone, we're looking forward to once again providing ways of efficiently limiting binds to the same supervisor.
```
christophermaier: This seems a bit redundant with the above paragraph. I'd also leave out mention of any potential concrete solutions like "limiting binds to the same supervisor"; keeping it to "meeting the usecases that composites were designed for" is good enough.

PR #5744, opened 2018-10-15
Purge "rumor heat" information for departed members

christophermaier:
This should allow us to do a degree of cleanup of old rumor
information. Absent this, large and long-running Habitat networks can
accumulate significant amounts of wasted memory on what is effectively
useless historical information.

The `RumorHeat` data structure effectively maps a rumor to a mapping
of members to the number of times the member has seen the rumor. This
allows us to know who's seen a rumor already, and who hasn't.

Now, when a member departs the ring, we'll remove any of the
"member-to-number-of-times-they've-seen-the-rumor" mappings for that
member.

Additionally, we'll remove _all_ such mappings for any rumors about
_services_ hosted on that member. After all, the member is going away,
so all of its services will too.

We aren't removing the membership rumors, though, because that's how
we pass the word that the member is departed.

Future work may add more cleanups and rearrangements.

### DIFF
--- a/components/butterfly/src/rumor/heat.rs
+++ b/components/butterfly/src/rumor/heat.rs
@@ -143,13 +143,24 @@ impl RumorHeat {
         // Remove any information about Service rumors for this
         // particular member... it's leaving, so none of its services
         // will be around either.
+        let count_before = heat_map.len();
         heat_map.retain(|k, _| !(k.kind == RumorType::Service && k.id == id));
+        let count_after = heat_map.len();
+        debug!(
+            "Purged {} service rumor mappings for {:?}",
+            count_before - count_after,
+            id
+        );
 
         // Remove any "cooling" information for this member, across
         // all types of rumors.
+        let mut count = 0;
         for heat in heat_map.values_mut() {
-            heat.remove(id);
+            if heat.remove(id).is_some() {
+                count += 1;
+            }
         }
+        debug!("Purged {} heat count entries for {:?}", count, id);
     }
 }
 
--- a/components/butterfly/src/rumor/heat.rs
+++ b/components/butterfly/src/rumor/heat.rs
@@ -27,7 +27,7 @@ use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 
 // Internal Modules
-use rumor::RumorKey;
+use rumor::{RumorKey, RumorType};
 
 // TODO (CM): Can we key by member instead? What do we do more frequently?
 // TODO (CM): Might want to type the member ID explicitly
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
I added the "do not merge" label while I run this through some further local testing. Pushing it out for early feedback, though.

components/butterfly/src/server/mod.rs
```diff
@@ -622,17 +622,22 @@ impl Server {
...
             trace_it!(
                 MEMBERSHIP: self,
                 TraceKind::MemberUpdate,
-                trace_member_id,
+                member_id,
                 trace_incarnation,
                 trace_health
             );
+
+            // Purge "heat" information for a member that's gone
```
baumanj: Again, I'm not sure this comment says anything the code doesn't convey quite clearly (ditto to the several other times this same comment is used).

What *is* confusing to me though is why we still `start_hot_rumor` even if we're purging. I'm not saying this is wrong (still trying to grok this code), but it seems strange enough that a comment about _that_ might be warranted.

components/butterfly/src/server/mod.rs
```diff
@@ -703,18 +711,23 @@ impl Server {
...
+            if member_id != self.member_id() && health == Health::Departed {
```
baumanj: This is _almost_ identical to the code in `insert_member`. I wonder if it's possible to factor out the commonality as this sort of duplication seems prone to bugs. I don't just mean for the added code, but for the two functions more generally.

components/butterfly/src/server/mod.rs
```diff
@@ -744,6 +757,9 @@ impl Server {
                 {
                     // TODO (CM): Why are we inferring departure from
                     // a service rumor?
+
+                    // Purge "heat" information for a member that's gone
+                    self.rumor_heat.purge(&service_rumor.member_id);
```
baumanj: I wonder if this and the duplication on L804 could be avoided by putting this logic inside `insert_health_by_id`.

components/butterfly/src/rumor/heat.rs
```diff
@@ -125,6 +125,32 @@ impl RumorHeat {
...
+        // Remove any information about Service rumors for this
+        // particular member... it's leaving, so none of its services
+        // will be around either.
+        heat_map.retain(|k, _| !(k.kind == RumorType::Service && k.id == id));
+
+        // Remove any "cooling" information for this member, across
+        // all types of rumors.
+        for heat in heat_map.values_mut() {
+            heat.remove(id);
+        }
```
jeremymv2: Perhaps either quantifying or summarizing the bytes purged might be of some use here?

```
+use std::mem;
...
         for heat in heat_map.values_mut() {
-            heat.remove(id);
+            if let Some(r) = heat.remove(id) {
+                debug!("Purging rumor heat of size {} bytes", mem::size_of_val(&r));
+            } else {
+                debug!("Member {} unexpectedly had no rumor heat to purge.", id);
+            }
         }
```

components/butterfly/src/server/mod.rs
christophermaier: Good point... I'm not clear on if there's actually a meaningful distinction between these functions; the names don't give it away, at least. This would probably be good to tackle in a follow-on PR.

components/butterfly/src/server/mod.rs
christophermaier: I'd initially wanted to do that, too, but the `MemberList` knows nothing of the `RumorHeat` data structure 😦

components/butterfly/src/rumor/heat.rs
christophermaier: `size_of_val` doesn't always give you what you think you want, since it's the size of the _value_, which may not be the same as the "total size" of the value, including any heap allocations. For example, you'll always get `24` if your value is a `String`, since a `String` is a 3-byte "fat pointer", regardless of how many characters are in the actual character string pointed to on the heap.

However, in this case, it's always going to be 64 bytes; 56 for the member ID (24 for the "fat pointer" and 32 for the ID string itself) and 8 for the `usize` counter.

PR #5746, opened 2018-10-16
Use sup-run instead of hab sup run in demo

### LINKED ISSUE #5538: Process supervisor tutorial gives command examples that don't work
On step 10, I'm asked to run the following two commands:
```
[1][default:/src:0]# hab sup run
[2][default:/src:0]# hab svc load yourorigin/sample-node-app --strategy at-once
```
Unfortunately, the terminal just hangs after `hap sup run` so you can't really ever execute the second command.

rsertelon:
> Closes #5538

### DIFF
--- a/www/source/demo/process-supervisor/steps/10.html.slim
+++ b/www/source/demo/process-supervisor/steps/10.html.slim
@@ -36,7 +36,7 @@ classes: demo_step
           | $ hab studio enter
         = code(:shell) do
           |
-            [1][default:/src:0]# hab sup run
+            [1][default:/src:0]# sup-run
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

rsertelon:
Should I go with @cnunciato comment then and remove mentions of starting the supervisor? (instead say it's already started in the studio?)

mwrock:
I'm going to merge this since the tutorial explicitly has the user enter a studio and there is no mention of using `HAB_STUDIO_SUP`. We could definitely go into some detail about starting a supervisor in different environments but this PR as feels like an easy win.

mwrock:
Doh! we are in a merge freeze. Will merge as soon as that is lifted.

Issue #5752, opened 2018-10-16
[docker studio] Set image name based on ActiveTarget


smacfarlane:
In order to support the docker studio for x86_64-linux-kernel2 PackageTarget,  the tag for the rootless_studio should be modified to contain the ActiveTarget of the hab cli contained in it.  

This will also require that the default docker image name be updated to include the ActiveTarget as part of the  tag and that the installation script download the correct hab cli. 

It's possible this will need to be done in a few phases, due to ordering of steps in the release process.  

https://github.com/habitat-sh/habitat/blob/master/components/rootless_studio/Dockerfile#L13
https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/studio/docker.rs#L32

PR #5753, opened 2018-10-16
Enable kernel2 builds in pipeline

smacfarlane:
This PR enables x86_64-linux-kernel2 builds in the release pipeline.   This will result in a new set of Habitat artifacts that need to be validated on a CentOS6-era Linux system.

Blocked by #5752

TODO:

- [x] Enable kernel2 rootless studio builds
- [x] Update install.sh script to take --target
- [x] Document x86_64-linux-kernel2 validation steps


thesentinels:

Thank you for contributing!

.buildkite/release_pipeline.yaml
```diff
@@ -19,6 +19,14 @@ steps:
     command: .buildkite/scripts/build_component.sh hab
     agents:
       queue: habitat-release
+  
+  - label: ":linux: :see_no_evil: :habicat: core/hab/kernel2"
+    command: .buildkite/scripts/build_component.sh hab
+    concurrency_group: "habitat/release/x86_64-linux-kernel2"
```
christophermaier: What's the intention behind the concurrency group for just the kernel2 builds? And also without a corresponding `concurrecy` key?

.buildkite/release_pipeline.yaml
```diff
@@ -48,6 +56,15 @@ steps:
     command: .buildkite/scripts/build_component.sh plan-build
     agents:
       queue: habitat-release
+  
+  - label: ":linux: :habicat: :hammer_and_wrench: core/hab-plan-build"
```
christophermaier: As you did with the `core/hab` build above, it would be good to either visually or textually differentiate the kernel2 builds from the "normal" builds to reduce confusion and ambiguity when looking at the overall pipeline.

.buildkite/release_pipeline.yaml
```diff
@@ -19,6 +19,14 @@ steps:
     command: .buildkite/scripts/build_component.sh hab
     agents:
       queue: habitat-release
+  
+  - label: ":linux: :see_no_evil: :habicat: core/hab/kernel2"
```
christophermaier: I'd separate the "kernel2" text from the package name, just so it's very clear that "kernel2" isn't some kind of weird version string 😄

christophermaier: Also, is kernel 2 "evil"? 🤣

.buildkite/release_pipeline.yaml
```diff
@@ -170,6 +227,14 @@ steps:
     command: .buildkite/scripts/bintray_upload.sh
     agents:
       queue: habitat-release
+  
+  - label: ":linux: Upload (but don't publish!) artifacts to Bintray"
```
christophermaier: Just to clarify: uploading to Bintray for kernel2 is just easier if we run the script again with a different `BUILD_PKG_TARGET`, right? Is there any benefit at all to modifying the overall script to try and upload _all_ Linux packages in a single operation?

.buildkite/scripts/build_component.sh
```diff
@@ -29,16 +44,28 @@ set_hab_binary() {
...
-        declare -g hab_binary="/hab/pkgs/${hab_ident}/bin/hab"
+        hab_binary="/hab/pkgs/${hab_ident}/bin/hab"
         declare -g new_studio=1
     else
-        echo "Buildkite metadata NOT found; using previously-installed hab binary"
-        hab_binary="$(which hab)"
-        declare -g hab_binary
+        echo "Buildkite metadata NOT found; using previously-installed hab binary: $hab_binary"
     fi
+    declare -g hab_binary
```
christophermaier: Ha, I guess it makes more sense to declare this outside the if/else 😄

.buildkite/scripts/build_component.sh
```diff
@@ -29,16 +44,28 @@ set_hab_binary() {
...
     echo "--- :habicat: Using $(${hab_binary} --version)"
 }
 
+install_hab_kernel2_binary() {
```
christophermaier: The comments above about this being a temporary step may be better attached to this function directly.

.buildkite/scripts/build_mac_release.sh
```diff
@@ -74,9 +74,7 @@ source results/last_build.env
 
 echo "--- :buildkite: Annotating build"
 
-# TODO (CM): Replace "macOS" below with ${pkg_target:?} once that's in
-# hab-plan-build (see https://github.com/habitat-sh/habitat/pull/5373)
-echo "<br>* ${pkg_ident:?} (macOS)" | buildkite-agent annotate --append --context "release-manifest"
+echo "<br>* ${pkg_ident:?} (${pkg_target:?})" | buildkite-agent annotate --append --context "release-manifest"
```
christophermaier: 🎉

.buildkite/scripts/build_component.sh
```diff
@@ -93,9 +120,7 @@ case "${component}" in
         ;;
 esac
 
-# TODO (CM): Replace "Linux" below with ${pkg_target:?} once that's in
-# hab-plan-build (see https://github.com/habitat-sh/habitat/pull/5373)
-echo "<br>* ${pkg_ident} (Linux)" | buildkite-agent annotate --append --context "release-manifest"
+echo "<br>* ${pkg_ident:?} (${pkg_target:?})" | buildkite-agent annotate --append --context "release-manifest"
```
christophermaier: :tada:

.buildkite/scripts/build_component.sh
```diff
@@ -29,16 +44,28 @@ set_hab_binary() {
...
     echo "--- :habicat: Using $(${hab_binary} --version)"
 }
 
+install_hab_kernel2_binary() {
+    local hab_src_url tempdir
+    hab_src_url="http://habitat-boostrap-artifacts.s3.amazonaws.com/x86_64-linux-kernel2/stage2/habitat-stage2-x86_64-linux-kernel2-latest"
+    tempdir=$(mktemp -d hab-kernel2-XXXX)
+
+    pushd $tmpdir >/dev/null
+    wget "$hab_src_url" hab-x86_64-linux-kernel2
```
christophermaier: We use `curl` everywhere else in our pipeline scripts. For consistency, it'd be good to pick either `curl` or `wget` and just use that everywhere.

I don't have particularly strong opinions either way; it's `curl` everywhere only because that's what I personally use the most, and I wrote the rest of the scripts 😄

.buildkite/release_pipeline.yaml
smacfarlane: That was in part a copy/paste from the bootstrap pipeline where we use it to enforce isolation.   I chatted with @scotthain about this after I initially published this branch and we decided we need to think on it a bit when we didn't have end-of-day pudding brain 😹   IIRC the question came down to: Do we need to ensure isolation between our two Linux builds for the release?

.buildkite/release_pipeline.yaml
smacfarlane: 👍   I couldn't find iconography I was happy with at the time so I punted on it for a bit.  💯 agree that it's needed.

.buildkite/release_pipeline.yaml
smacfarlane: Good point,  I'll fix that.  

A little bit 😂  That just happened to be in front of me when I stopped looking for a good representative icon for the kernel2 builds.  It was a small delighter for myself at the end of the day 😄   It should change when we decide on an icon to use.

.buildkite/release_pipeline.yaml
smacfarlane: Reading the source for bintray_upload, it uses `hab pkg install` to get the artifact on disk and then call bintray-publish.  Since that will only install for the ActiveTarget, I elected to split the steps out, similar to the Mac step. 

I think we'll want to update that to use curl instead, which would allow us to collapse all targets to a single "bintray upload" step.  I felt that that was a larger change that should be separately from this change though.

I'm not sure I understand the BUILD_PKG_TARGET question.  If I remember correctly, that only comes in to play when you're bootstrapping the hab ecosystem and are executing plan-build out of the source tree, rather than from a package.

.buildkite/scripts/build_component.sh
smacfarlane: Copy/paste 😄   Will update to curl

.buildkite/release_pipeline.yaml
christophermaier: I don't _think_ we do... the only kind of isolation I think we need is the sequencing, but that's taken care of using the various `wait` steps.

The only thing we're currently using concurrency on is the macOS builder, and then only because the builder is a single, long-lived machine, and we've had issues with our Travis-based release of having concurrent builds stepping on each other. It doesn't feel like there's a similar concern with the kernel 2 stuff, but I may be missing something; you guys have been more in that work than I have.

.buildkite/release_pipeline.yaml
christophermaier: Could maybe just use :two:?

.buildkite/release_pipeline.yaml
christophermaier: I don't necessarily object 😄 I'm a big fan of using "delighter" emoji in Buildkite.

.buildkite/release_pipeline.yaml
christophermaier: I think your approach is a good one; I was just wanting to confirm my suspicion.

I also think that we're probably at the point where we could likely replace the bintray package with a shell script in the pipeline and call it a day. But I agree, that's a bigger change for a later time.

I'm 👍 on what you're doing here.

.buildkite/release_pipeline.yaml
smacfarlane: ☕️ Kicking in now.  My concern was that *if* a kernel-current and kernel2 build were scheduled on the same host, you would end up with failed or corrupted builds because `hab pkg build` is going to attempt to destroy any existing studio at the current path, which would be the same for the two Targets.    

We could probably work around this by naming the studio (`-r`) with the Target as part of the studio name, but it might just be safer to enforce them executing on separate hosts.  

I'm not sure if it's possible for builds to be scheduled on the same host, I'm not familiar enough with how it's set up, so this may be a non-concern.

christophermaier: I think the path should be different, though, since they'd be different Buildkite jobs, and I think that's part of the path.

As for being scheduled on the same host, I think that depends on how the agent is configured, which is a question for @scotthain, methinks.

scotthain: After more thoughts, we shouldn't need the concurrency group in this case. (And if there's something we've manage to forget/nuke from our brains, we can add it back)

scotthain: Also as a side note, if we don't cut the concurrency down to 1 (as we do in some other pipelines as a forced limiting factor) there's no real point in this case :)

christophermaier:
@smacfarlane This continues to look really nice 👍 

How close do you think we are to having this not be "WIP"?

smacfarlane:
@christophermaier   Very soon.   The build portion of the pipeline looks good. I keep hitting what look like infrastructure issues so haven't had a successful end-to-end run without failures yet.  I'd like to do a bit more digging to confirm that the errors are unrelated to the changes in the scripts before removing the WIP.

smacfarlane:
@christophermaier  I've removed the WIP 🎉 🌮      I'm working on the docker changes noted in the initial PR.  While not a strict blocker to this one, I would like to make sure both changes make it in before our next release so they should be merged around the same time.

smacfarlane:
@christophermaier   One thing to note, I think we'll have to do a manual release of the x86_64-linux-kernel2 docker studio, with a follow on PR to enable that in the pipeline post-release.    This is because the image building process relies on a published hab release in bintray, which doesn't occur until after the docker images are built and uploaded.

.buildkite/release_pipeline.yaml
```diff
@@ -190,6 +247,12 @@ steps:
       Appveyor has generated the necessary artifacts and uploaded them
       to Builder and Bintray. Please ensure those artifacts have been
       uploaded before continuing with the release.
+  
+  - block: ":dango: :linux: :two: Build x86_64-linux-kernel2 Launcher"
+    prompt: |
+      x86_64-linux-kernel2 launcher will need to be built by hand.  If 
+      you are promoting launchers in the following step, please build 
+      and upload and x86_64-linux-kernel2 launcher before continuing.
```
christophermaier: Is this because of a lack of kernel2 workers in Builder (which would normally be building this)?

(Adding or linking to explicit directions for building a kernel2 launcher would be good here.)

Also, why :dango:? 😄

.buildkite/scripts/bintray_publish.sh
```diff
@@ -26,16 +26,22 @@ echo "--- :habicat: Publishing all Habitat CLI artifacts in Bintray"
...
+# Linux x86-64-linux Publish
+release=$(buildkite-agent meta-data get "hab-release-x86_64-linux")
+echo "--- :linux: Publishing x86-64-linux 'hab' ${version}-${release} on Bintray"
 publish "https://api.bintray.com/content/habitat/${bintray_repository}/hab-x86_64-linux/${version}-${release}/publish"
 
+########################################################################
+# Linux x86-64-linux-kernel2 Publish
+release=$(buildkite-agent meta-data get "hab-release-x86_64-linux-kernel2")
+echo "--- :linux: :two: Publishing x86-64-linux-kernel2 'hab' ${version}-${release} on Bintray"
+publish "https://api.bintray.com/content/habitat/${bintray_repository}/hab-x86_64-linux-kernel2/${version}-${release}/publish"
```
christophermaier: I think we're at the point where we want to refactor the `publish` function to take `repository`, `target`, `version`, and `release` arguments, and generate the complete URL internally. Otherwise, we've duplicated the URL creation logic four times.

.buildkite/scripts/bintray_upload.sh
```diff
@@ -40,9 +42,14 @@ echo "--- :habicat: Uploading core/hab to Bintray"
 # If we use Buildkite, we can potentially upload many different
 # platform artifacts to Bintray from a single platform (e.g., upload
 # Windows artifacts from Linux machines.)
-sudo hab pkg install core/hab --channel="${channel}"
+sudo "${hab_binary:?}" pkg install core/hab --channel="${channel}"
 
-hab_artifact=$(buildkite-agent meta-data get "hab-artifact")
+# TODO (SM): Should we specify Target for all of our platforms?
```
christophermaier: Probably so, particularly once we get the Windows builds working in Buildkite. Might as well take care of that now.

.buildkite/scripts/bintray_upload_macos.sh
```diff
@@ -24,7 +24,7 @@ sudo hab pkg install \
      core/hab-bintray-publish
 
 echo "--- :buildkite: Retrieving macOS core/hab artifact"
-hab_artifact=$(buildkite-agent meta-data get "hab-artifact-macos")
+hab_artifact=$(buildkite-agent meta-data get "hab-artifact-x86_64-darwin")
```
christophermaier: Wonder if we can get a bit more consistency, and maybe be able to abstract some of these metadata calls into functions, if we require a package target environment variable to be set everywhere, and then take advantage of that here, rather than hardcoding?

.buildkite/scripts/build_component.sh
```diff
@@ -75,27 +41,31 @@ else
 fi
 source results/last_build.env
 
+# TODO (SM): The 0.59.0 hab cli that we rely on for x86_64-linux builds 
+# doesn't emit pkg_target. Until we've sufficiently bootstrapped ourselves
+# we need to set it. This can be removed when studio-ci-common pulls 0.63.0 
+# or newer. This is safe to do because the x86_64-linux-kernel2 builds will
+# already have this value set.
```
christophermaier: Might be good to create an issue for this and add a link here.

.buildkite/release_pipeline.yaml
smacfarlane: Correct.  I will add those steps to the Release documentation and link to them.

I was thinking about how good a dango would be at the time I was working on this. I am happy to report that that particular issue was resolved 😄  Also, three little launchers all lined up?(I just made that up, but I kinda like it)

.buildkite/scripts/bintray_publish.sh
smacfarlane: 👍 Will refactor

.buildkite/scripts/bintray_upload_macos.sh
smacfarlane: I started to look at this because I think it's the right idea, but I've hit a bit of a snag.  We use the `meta-data get` for both the current $pkg_target, but in the case of bintray_publish for _all_ $pkg_targets. 

Creating a function that always requires the target be passed doesn't get us much (anything?) from requiring a Target environment variable. Having the target be optional I'd argue is worse because it would be pretty easy to miss a spot where it's necessary and hard to debug in the pipeline.  

We could split out bintray_publish to act only on a single target, and have a step for each target... I think this is what I'll try. Refactoring all scripts to act on a single target where appropriate. It means we'd need N workers to run each curl rather than 1 worker to run N curls. I think the failure case of one not working isn't any worse than current state either.

christophermaier: I may have been overly prescriptive... the function doesn't need to operate on the specific environment variable, but we could take advantage of the environment variable and pass it in as an argument.

One thing this would allow us to do is hide the nature of the _keys_ behind a function. For example, we do that here: https://github.com/habitat-sh/habitat/blob/58f19cf1c2d756f4e2c8e2495a40fef4fd3c3f0d/.buildkite/scripts/shared.sh#L47-L61

Every place that needs to operate on the metadata for a "fake release" doesn't need to know what the metadata key is (or that it's even stored in Buildkite metadata). It feels like we're getting to the point where we're making more and more use of the metadata feature, to the point where it could become confusing and error prone to be replicating all this logic across the various scripts.

Then here, we'd just have a list of targets we care about (not even environment variables, necessarily) and we'd just iterate through them, operating on each appropriate package in turn.

PR #5754, opened 2018-10-17
Add macro for cli testing

mpeck:
Adds a macro to help with verifying that cli commands are being parsed as expected.

### DIFF
--- a/components/sup/src/cli.rs
+++ b/components/sup/src/cli.rs
@@ -1,0 +1,77 @@
+use clap::App;
+
+use hab::cli::{
+    sub_sup_bash, sub_sup_depart, sub_sup_run, sub_sup_secret, sub_sup_sh, sub_sup_term,
+    sub_svc_status,
+};
+use VERSION;
+
+pub fn cli<'a, 'b>() -> App<'a, 'b> {
+    clap_app!(("hab-sup") =>
+        (about: "The Habitat Supervisor")
+        (version: VERSION)
+        (author: "\nAuthors: The Habitat Maintainers <humans@habitat.sh>\n")
+        // set custom usage string, otherwise the binary
+        // is displayed confusingly as `hab-sup`
+        // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
+        (usage: "hab sup <SUBCOMMAND>")
+        (@setting VersionlessSubcommands)
+        (@setting SubcommandRequiredElseHelp)
+        // this is the _full_ list of supervisor related cmds
+        // they are all enumerated here so that the entire help menu
+        // can be displayed from `hab sup --help`
+        (subcommand: sub_sup_bash().aliases(&["b", "ba", "bas"]))
+        (subcommand: sub_sup_depart().aliases(&["d", "de", "dep", "depa", "depart"]))
+        (subcommand: sub_sup_run().aliases(&["r", "ru"]))
+        (subcommand: sub_sup_secret().aliases(&["sec", "secr"]))
+        (subcommand: sub_sup_sh().aliases(&[]))
+        (subcommand: sub_svc_status().aliases(&["stat", "statu"]))
+        (subcommand: sub_sup_term().aliases(&["ter"]))
+    )
+}
+
+#[cfg(test)]
+mod test {
+    use super::cli;
+
+    macro_rules! assert_cli_cmd {
+        ($test:ident, $cmd:expr, $( $key:expr => $value:tt ),+) => {
+            #[test]
+            fn $test() {
+                assert_cmd!(cli(), $cmd, $( $key => $value ),+ );
+            }
+        }
+    }
+
+    mod sup_run {
+        use super::*;
+
+        assert_cli_cmd!(should_handle_multiple_peer_flags,
+                        "hab-sup run --peer 1.1.1.1 --peer 2.2.2.2",
+                        "PEER" => ["1.1.1.1", "2.2.2.2"]);
+
+        assert_cli_cmd!(should_handle_single_peer_flag_with_multiple_values,
+                        "hab-sup run --peer 1.1.1.1 2.2.2.2",
+                        "PEER" => ["1.1.1.1", "2.2.2.2"]);
+
+        assert_cli_cmd!(should_handle_peer_flag_with_arguments,
+                        "hab-sup run --peer 1.1.1.1 2.2.2.2 -- core/redis",
+                        "PEER" => ["1.1.1.1", "2.2.2.2"],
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/cli_test_helpers.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl fmt::Display for CliTestError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let content = match self {
+            CliTestError::ValueMismatch(flag, expected, actual) => format!(
+                r#"
+'{}' validation failed:
+  Expected: '{}'
+  Actual:   '{}'
+"#,
```
baumanj: This is fine, but just FYI, there's another option you can use to keep the indentation in the code nicer. Since a backslash can be used to escape leading whitespace:
```rust
        println!("'{}' validation failed:\n  \
                  Expected: '{}'\n  \
                  Actual:   '{}'",
                  "foo", "bar", "baz");
```
yields:
```
'foo' validation failed:
  Expected: 'bar'
  Actual:   'baz'
```

components/sup/src/cli_test_helpers.rs
```diff
@@ -0,0 +1,151 @@
...
+                err.to_string()
+            ),
+        };
+        write!(f, "{}", content)
+    }
+}
+
+pub fn assert_command(app: App, cmd: &str, assertions: Vec<(&str, Expectation)>) {
+    let cmd_vec = Vec::from_iter(cmd.split_whitespace());
+    let errors: Vec<CliTestError> = match app.get_matches_from_safe(cmd_vec) {
```
baumanj: Just a fun rust FYI, the compiler can infer the type here so 
```suggestion
    let errors = match app.get_matches_from_safe(cmd_vec) {
```
will work just as well.

components/sup/src/cli_test_helpers.rs
```diff
@@ -0,0 +1,151 @@
...
+            (_, None) => vec![CliTestError::MissingSubcmd()],
+        },
+        Err(err) => vec![CliTestError::ClapError(err.to_string())],
+    };
+
+    if !errors.is_empty() {
+        let error_string = errors
+            .into_iter()
+            .map(|error| format!("{}", error))
+            .collect::<Vec<String>>()
```
baumanj: More rust fun: the inner type of the `Vec` can be inferred if you do:
```suggestion
            .collect::<Vec<_>>()
```
Though you can't leave the `Vec` part out entirely.

components/sup/src/cli_test_helpers.rs
```diff
@@ -0,0 +1,151 @@
...
+"#,
+            cmd, error_string
+        );
+    }
+}
+
+fn assert_matches(matches: &ArgMatches, assertions: Vec<(&str, Expectation)>) -> Vec<CliTestError> {
+    assertions
+        .into_iter()
+        .fold(Vec::new(), |mut errs, (flag, expected_value)| {
```
baumanj: I'm a big fan of FP, but in this case I don't think it's adding to the legibility of this code. I'd try writing it out with a `for` loop over the assertions and see if that ends up shorter/simpler.

Mostly I'm not a fan of using `fold` when the computation in the closure doesn't really depend on the memoized object*. It's possible I would like this better if it were a `map` of assertion → error. But since assertions may map to no error, I think a [`filter_map`](http://doc.rust-lang.org/1.29.2/core/iter/trait.Iterator.html#method.filter_map) would be appropriate and your closure would return `Option<CliTestError>` and a `collect` would leave you with `Vec<CliTestError>`.

\* I know `errs` carries the state, but that's not really the same thing in my mind.

components/sup/src/cli.rs
```diff
@@ -0,0 +1,120 @@
...
+        (subcommand: sub_sup_run().aliases(&["r", "ru"]))
+        (subcommand: sub_sup_secret().aliases(&["sec", "secr"]))
+        (subcommand: sub_sup_sh().aliases(&[]))
+        (subcommand: sub_svc_status().aliases(&["stat", "statu"]))
+        (subcommand: sub_sup_term().aliases(&["ter"]))
+    )
+}
+
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
mpeck: I agree with your thinking here, but I think this might be worth it's own PR. The way the CLI is defined is kinda weird and I think it deserves a bit more thought to unravel it all.

components/sup/src/cli.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        #[test]
+        fn should_handle_single_peer_flag_with_multiple_values() {
+            assert_cmd!(cli(),
+                        "hab-sup run --peer 1.1.1.1 2.2.2.2",
+                        "PEER" => ["1.1.1.1", "2.2.2.2"]);
+        }
+
+        #[test]
+        fn should_handle_peer_flag_after_arguments() {
```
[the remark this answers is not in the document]
mpeck: That's reasonable. I added those when I was testing the functionality of the macro. But I agree, they aren't particularly useful.

components/sup/src/cli.rs
```diff
@@ -0,0 +1,223 @@
...
+            fn $test() {
+                assert_cmd!(cli(), $cmd, $( $key => $value ),+ );
+            }
+        }
+    }
+
+    mod sup_run {
+        use super::*;
+
+        assert_cli_cmd!(should_take_listen_gossip_flag,
```
baumanj: I think tests like this are a bit overkill. Considering the code that establishes this behavior is only
```rust
        (@arg LISTEN_GOSSIP: --("listen-gossip") +takes_value {valid_socket_addr}
            "The listen address for the Gossip System Gateway. If not specified, the value will \
            be taken from the HAB_LISTEN_GOSSIP environment variable if defined. [default: 0.0.0.0:9638]")
```
I don't know that it calls for having a separate test that only exercises the CLI parsing, not the behavior that the argument should result in.

I could see some tests to exercise that `valid_socket_addr` behaves as expected (and that would have more value since several options depend upon it), but I think all the `should_take…` tests are adding code without sufficient value.

components/sup/src/cli.rs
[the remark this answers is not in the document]
baumanj: One other benefit of this alternate approach is that you could leave this `cli` method where it is and avoid creating this new file.

components/sup/src/cli.rs
mpeck: Generally I would agree with you. But the thinking is that we will be moving away from the `clap_app!` macro for defining the cli as we bring everything into a single binary. Without tests for cli parsing it would be tough to know if we are getting the same cli behavior as before.

PR #5756, opened 2018-10-18
cleanup supervisor startup and termination in a local windows studio

### LINKED ISSUE #5708: Check if sup PID exists before killing it exiting the windows studio
When you exit a windows studio we check the supervisor `LOCK` file and kill that pid. If for whatever reason that pid does not exist, this results in a confusing error.

At the very least we should check if the pid exists. Bonus points for emitting a warning if a supervisor is already running or if we are unable to start the supervisor in the studio which is usually a precursor to this scenario.

mwrock:
fixes #5708 

This polishes up a few tings regarding supervisor startup and termination in a local Windows studio:

* If a supervisor is already running either in another studio or outside of the studio, It will warn the user and not attempt to start a new supervisor
* When terminating the supervisor, if the `LOCK` file points to a non existent process, it will not throw an error
* It will remove the lock file after termination so that a future termination will not reference it again.


thesentinels:

Thank you for contributing!

PR #5771, opened 2018-10-19
do not check for running supervisor in windows container studio

mwrock:
This fixes an issue where we are emitting a warning in a container windows studio because the docker windows studio starts the supervisor immediately. The warning states that we will not start a supervisor which is simply not true.


thesentinels:

Thank you for contributing!

PR #5773, opened 2018-10-20
remove -w parameter for invoking a local studio in internal calls

mwrock:
We don't need to pass `-w` now that it is default behavior and also emits a deprecation warning.


thesentinels:

Thank you for contributing!

components/hab/src/command/studio/docker.rs
```diff
@@ -270,7 +270,6 @@ where
     cmd_args.push(image_identifier(&docker_cmd).into());
     cmd_args.extend_from_slice(args.as_slice());
     if is_serving_windows_containers(&docker_cmd) {
-        cmd_args.push("-w".into());
```
christophermaier: Not necessarily relevant to the goal of this PR, but it would be nice for us to use long options whenever possible for readability.

Issue #5547, opened 2018-09-04
Move hab-win-deps to dedicated windows core-plans 


mwrock:
There are a few low level hab dependencies that we pull from an S3 bucket: https://s3-us-west-2.amazonaws.com/habitat-win-deps/hab-win-deps.zip. This includes openssl, zlib, libsodium and I think a couple others.

We set this up "back in the day" when we were trying to get the most basic Habitat build out the door and theer was no propper windows Habitat build capabilities. It's safe to say we have moved past that point and should now get proper plans for all of these.

mwrock:
closed by #5777

PR #5777, opened 2018-10-22
Add windows based native build plan

scotthain:

Requires https://github.com/habitat-sh/core-plans/pull/1924

### DIFF
--- a/components/hab/habitat/plan.ps1
+++ b/components/hab/habitat/plan.ps1
@@ -55,6 +55,17 @@ function Invoke-Build {
     finally { Pop-Location }
 }
 
+function Invoke-Check {
+    Push-Location "$PLAN_CONTEXT"
+    try {
+        cargo test --release --verbose
+        if($LASTEXITCODE -ne 0) {
+            Write-Error "Cargo test failed!"
+        }
+    }
+    finally { Pop-Location }
+}
+
 function Invoke-Install {
     Write-BuildLine "$HAB_CACHE_SRC_PATH/$pkg_dirname"
     Copy-Item "$env:CARGO_TARGET_DIR/release/hab.exe" "$pkg_prefix/bin/hab.exe"
--- a/components/hab/habitat/plan.ps1
+++ b/components/hab/habitat/plan.ps1
@@ -3,16 +3,14 @@ $pkg_origin = "core"
 $pkg_version = "$(Get-Content $PLAN_CONTEXT/../../../VERSION)"
 $pkg_maintainer = "The Habitat Maintainers <humans@habitat.sh>"
 $pkg_license = @("Apache-2.0")
-$pkg_source = "https://s3-us-west-2.amazonaws.com/habitat-win-deps/hab-win-deps.zip"
-$pkg_shasum="00b34fb983ebc43bfff9e8e2220d23db200cb45494a4971a5e2e733f1d73d04b"
 $pkg_deps=@(
     "core/openssl",
+    "core/zlib",
     "core/libarchive",
     "core/libsodium"
 )
 $pkg_bin_dirs = @("bin")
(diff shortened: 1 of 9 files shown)


components/hab/plan.ps1
```diff
@@ -0,0 +1,62 @@
+$pkg_name = "hab"
+$pkg_origin = "core"
+$pkg_version = "$(Get-Content $PLAN_CONTEXT/../../VERSION)"
+$pkg_maintainer = "The Habitat Maintainers <humans@habitat.sh>"
+$pkg_license = @("Apache-2.0")
+$pkg_source = "https://s3-us-west-2.amazonaws.com/habitat-win-deps/hab-win-deps.zip"
```
scotthain: this can be removed

thesentinels:

Thank you for contributing!

components/hab/habitat/plan.ps1
```diff
@@ -19,12 +29,12 @@ function Invoke-Prepare {
     $env:SSL_CERT_FILE              = "$(Get-HabPackagePath "cacerts")/ssl/certs/cacert.pem"
     $env:LIB                        += ";$HAB_CACHE_SRC_PATH/$pkg_dirname/lib"
```
mwrock: This currently points to the lib directory in the `hab-win-deps` aws bundle. I found that if that bundle is missing the build fails.

components/hab/habitat/plan.ps1
```diff
@@ -19,12 +29,12 @@ function Invoke-Prepare {
...
     $env:INCLUDE                    += ";$HAB_CACHE_SRC_PATH/$pkg_dirname/include"
```
mwrock: same as above

components/hab/habitat/plan.ps1
```diff
@@ -5,8 +5,18 @@ $pkg_maintainer = "The Habitat Maintainers <humans@habitat.sh>"
...
 $pkg_shasum="00b34fb983ebc43bfff9e8e2220d23db200cb45494a4971a5e2e733f1d73d04b"
+$pkg_deps=@(
+    "core/openssl",
+    "core/libarchive",
+    "core/libsodium"
+)
 $pkg_bin_dirs = @("bin")
-$pkg_build_deps = @("core/visual-cpp-redist-2013", "core/rust", "core/cacerts")
+$pkg_build_deps = @(
+    "core/visual-cpp-redist-2013",
```
mwrock: I **think** this 2013 dep can go since everything is compiled via the 2015 runtime.

components/hab/habitat/plan.ps1
mwrock: Actually, we can just delete lines 30 and 31. Oddly we need to rename libsodium.lib in tthe libsodium package to sodium.lib. Maybe there is some other variable to set to avoid that?

mwrock:
Everything but the Supervisor worked for me. It exits witH DLL_NOT_FOUND. And it only fails inside a docker studio. I ran the same binary locally while running procmon to see what deps it was using. I noticed the 2015 vc runtime DLLs so I copied those over from my local environment to my container but it was still failing. Tomorrow I might try this trick: https://stefanscherer.github.io/find-dependencies-in-windows-containers/ which looks like a worthwhile educational exercise.

mwrock:
ok after a night's sleep I quickly find that the supervisor works just fine in a container as long as you have the vcruntime140 (2005 vc) redist DLLs. So we need to package those and declare a runtime dep on them.

scotthain:
This will resolve https://github.com/habitat-sh/habitat/issues/5547

scotthain:
This is superceded by https://github.com/habitat-sh/habitat/pull/5777

PR #5780, opened 2018-10-23
adds in blog post around packaging inspec profiles with Habitat

nellshamrell:
Greetings!

In my time in both the Habitat and InSpec communities, I have encountered people who are very interested in using Habitat both to package up and distribute InSpec profiles.

We do have a Habitat/Inspec integration, but currently it creates a hart file for the InSpec profile, rather than creating a habitat directory with a plan.sh file in it in the same repo as the profile files.

This is certainly useful, but prevents us from taking advantage of Github integrations with Builder.

We should amend the Habitat/InSpec integration to provide this functionality - but with so much already on our roadmaps for both projects, I thought it might be prudent to document the work around in this post so that people who need this ability now can use it.

Should we add this functionality into the Habitat/Inspec integration I would be all too happy to amend this blog post :)


thesentinels:

Thank you for contributing!

eeyun:
I love the idea of having an updated blog post on packing up an inspec profile with habitat but I have a concern with the content of this for a specific reason - it feels like we're documenting a workaround to a bug, rather than documenting a first class experience. I would much rather we fix the issue with the inspec cli and publish a blog post that uses both tools fully than a post that could codify an unintentional behavior in one of our tools.

www/source/blog/2018-10-23-habitat-inspec.html.md
```diff
@@ -0,0 +1,236 @@
...
+pkg_version="0.1.0"
+```
+
+* Then add in this content (note that I've modified the pkg_version string to match the version of nginx-baseline):
+
+**plan.sh**
+```
+pkg_name=nginx-baseline
+pkg_origin=<YOUR ORIGIN>
+pkg_version="2.0.2"
```
james-stocks: This `pkg_version` may work as a bonus because the version string inside an Inspec profile metadata is often based on the compliance standard it implements (e.g. 'CIS CentOS 7 v1.1.0 Benchmark' results in Inspec profile version `1.1.0`) and is not semver for the version control changes made to the profile's files. 
We can maybe have `pkg_version` be semver for changes to the profile; and keep the Inspec profile version tied to the compliance standard

www/source/blog/2018-10-23-habitat-inspec.html.md
```diff
@@ -0,0 +1,236 @@
...
+pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+pkg_license=("Apache-2.0")
+pkg_source=https://github.com/<YOUR GITHUB NAME>/nginx-baseline/archive/2.0.2.tar.gz
+pkg_shasum="b546c3940ce03d72f4bbb77503ecf792fd54e608cd729f6511c03e11c5cd0727"
+pkg_deps=(chef/inspec)
+pkg_bin_dirs=(bin)
+
+do_build() {
+    return 0
```
james-stocks: The actual build action for an Inspec profile is `inspec archive [profile]` - that resolves the profile's dependencies and packs them into a tgz

Should that be happening at this point?

james-stocks: Oh, I now see at https://www.inspec.io/docs/reference/habitat/ that there is `inspec habitat profile create` which should resolve dependencies and generate a hart. I'm not sure if that hart yields the tgz version of a profile or just packs up all the source files for it

nellshamrell:
@eeyun - I could see delaying this if adding this functionality to the inspec/habitat integration is on our road map in the next few months.  It currently is not, and unless it is added, I'd rather give users a work around (as this has been requested) than give them no information at all.

nellshamrell:
Should this functionality be added to the current Habitat/Inspec integration, I would be most happy to amend this blog post to point toward it.

nellshamrell:
At the kind suggestion of @elliott-davis, I have amended the commit message to include more information about why I thought it necessary to create this post :)

cnunciato:
I'd rather see this as a wiki or forum post than a blog post, to @eeyun's point. Any objection to that @nellshamrell?

nellshamrell:
Closing, as an alternative was found :)

PR #5789, opened 2018-10-26
Add the possible values for --topology into the online help

baumanj:
CLI help before:

    -t, --topology <TOPOLOGY>                  Service topology; [default: none]

After:

    -t, --topology <TOPOLOGY>                  Service topology; [default: none] [possible values: standalone, leader]

With bad input such as `hab sup run --topology foobar` before:
```
error: Invalid value for '--topology <TOPOLOGY>': Service topology: 'foobar' is not valid
```
After:
```
error: 'foobar' isn't a valid value for '--topology <TOPOLOGY>'
	[possible values: leader, standalone]
```
Also, as long as clap is validating the topology argument, we can remove the
redundant validator function. This decouples the CLI from the code a bit, but
it's hard to do better unless we move away from the macro-based approach to
clap.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -856,7 +856,7 @@ pub fn sub_sup_run() -> App<'static, 'static> {
             "Environment name; [default: not set].")
         (@arg GROUP: --group +takes_value
             "The service group; shared config and topology [default: default].")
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #5790, opened 2018-10-29
Add `hab pkg list` command

jamesc:
Added a simple package listing command.  You can list:

* all packages
* an origin
* a specific package identifier.  If this has no release (or version)
   it will list all packages which match

Examples:
*  List all packages
```hab pkg list --all
azure-habitat-example/mysql/5.7.17/20180705230519
azure-habitat-example/osba-mysql-proxy/0.1.0/20180706000536
azure-habitat-example/osba-mysql-proxy/0.1.1/20180709202012
azure-habitat-example/wordpress/4.9.5/20180705230516
azure-habitat-example/wordpress/4.9.5/20180807195958
core/acl/2.2.52/20180608091922
core/attr/2.4.47/20161208223238
core/attr/2.4.47/20170513213059
core/attr/2.4.47/20180608091914
core/busybox-static/1.28.1/20180608102729
core/bzip2/1.0.6/20180608091727
core/cacerts/2018.03.07/20180608102212
core/coreutils/8.29/20180608092141
core/curl/7.54.1/20180608142121
core/db/5.3.28/20180608094030
core/docker/18.03.0/20180608150948
core/gawk/4.2.0/20180608093856
core/gcc-libs/5.2.0/20161208223920
core/gcc-libs/5.2.0/20170513212920
core/gcc-libs/7.3.0/20180608091701
core/gdbm/1.14.1/20180608094002
core/glibc/2.22/20160612063629
core/glibc/2.22/20170513201042
core/glibc/2.27/20180608041157
core/gmp/6.1.2/20180608051426
core/grep/3.1/20180608092809
core/hab/0.65.0/20181004215832
core/hab-pkg-export-docker/0.65.0/20181004222735
core/hab-studio/0.65.0/20181004221345
core/inetutils/1.9.4/20180608094116
core/jq-static/1.10/20160909011845
core/kubectl/1.9.6/20180609041406
core/less/530/20180608094202
core/libcap/2.25/20180608091936
core/libedit/3.1.20150325/20180608184640
core/libjpeg-turbo/1.5.0/20180608183037
core/libpng/1.6.21/20180608142039
core/libpng/1.6.35/20180717232103
core/libxml2/2.9.6/20180608141053
core/linux-headers/4.15.9/20180608041107
core/linux-headers/4.3/20160612063537
core/linux-headers/4.3/20170513200956
core/mpfr/4.0.1/20180608051629
core/mysql-client/5.7.21/20180608165812
core/ncurses/6.0/20161213233720
core/ncurses/6.0/20170513213009
core/ncurses/6.1/20180608091810
core/nghttp2/1.31.0/20180608142107
core/nginx/1.13.10/20180609041552
core/nginx/1.15.2/20180731172406
core/openssl/1.0.2n/20180608102213
core/pcre/8.41/20180608092740
core/perl/5.26.1/20180608094208
core/php/7.1.4/20180627180032
core/php/7.1.4/20180717232729
core/procps-ng/3.3.12/20180608092056
core/readline/6.3.8/20161213234107
core/readline/6.3.8/20170513213506
core/readline/7.0.3/20180608092900
core/sed/4.4/20180608091938
core/zlib/1.2.11/20180608050617
core/zlib/1.2.8/20161118033245
core/zlib/1.2.8/20170513201911
```

* List an origin
```
hab pkg list -o azure-habitat-example
azure-habitat-example/mysql/5.7.17/20180705230519
azure-habitat-example/osba-mysql-proxy/0.1.0/20180706000536
azure-habitat-example/osba-mysql-proxy/0.1.1/20180709202012
azure-habitat-example/wordpress/4.9.5/20180705230516
azure-habitat-example/wordpress/4.9.5/20180807195958
```

* list a particular Package Identifier
```
hab pkg list core/zlib
core/zlib/1.2.11/20180608050617
core/zlib/1.2.8/20161118033245
core/zlib/1.2.8/20170513201911
```

* Error case
```
hab pkg list core/zlib/1.2.8/fdd/dd
error: Invalid value for '<PKG_IDENT>': 'core/zlib/1.2.8/fdd/dd' is not valid. Package identifiers have the form origin/name[/version[/release]]
```

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -18,6 +18,7 @@ use std::result;
 use std::str::FromStr;
 
 use clap::{App, AppSettings, Arg};
+use hcore::package::ident;
 use hcore::package::{Identifiable, PackageIdent};
 use hcore::{crypto::keys::PairType, service::ServiceGroup};
 use protocol;
@@ -410,7 +411,7 @@ pub fn get() -> App<'static, 'static> {
                     (@attributes +required)
                     (@arg ALL: -a --all
                             "List all installed packages")
-                    (@arg ORIGIN: -o --origin +takes_value
+                    (@arg ORIGIN: -o --origin +takes_value {valid_origin}
                             "An origin to list")
                     (@arg PKG_IDENT: +takes_value {valid_ident}
                     "A package identifier (ex: core/redis, core/busybox-static/1.42.2).")
@@ -1118,6 +1119,14 @@ fn valid_fully_qualified_ident(val: String) -> result::Result<(), String> {
             "'{}' is not valid. Fully qualified package identifiers have the form origin/name/version/release",
             &val
         )),
+    }
+}
+
+fn valid_origin(val: String) -> result::Result<(), String> {
+    if ident::is_valid_origin_name(&val) {
+        Ok(())
+    } else {
+        Err(format!("'{}' is not valid. A valid origin contains a-z, 0-9, and _ or - after the first character", &val))
     }
 }
 
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -403,6 +403,21 @@ pub fn get() -> App<'static, 'static> {
                 (@arg PKG_IDENT: +required +takes_value {valid_ident}
                     "A package identifier (ex: core/redis, core/busybox-static/1.42.2)")
             )
+            (@subcommand list =>
+                (about: "List all versions of installed packages")
+                (aliases: &["li"])
+                (@group prefix =>
+                    (@attributes +required)
+                    (@arg ALL: -a --all
+                            "List all installed packages")
+                    (@arg ORIGIN: -o --origin +takes_value
+                            "An origin to list")
+                    (@arg PKG_IDENT: +takes_value {valid_ident}
+                    "A package identifier (ex: core/redis, core/busybox-static/1.42.2).")
+
+                )
+
+            )
             (@subcommand provides =>
                 (about: "Search installed Habitat packages for a given file")
                 (@arg FILE: +required +takes_value
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -135,7 +135,7 @@ pub fn get() -> App<'static, 'static> {
                         "The job id that was returned from \"hab bldr job start\" \
                         (ex: 771100000000000000)")
                     (@arg CHANNEL: +takes_value +required "The target channel name")
-                    (@arg ORIGIN: -o --origin +takes_value
+                    (@arg ORIGIN: -o --origin +takes_value {valid_origin}
                         "Limit the promotable packages to the specified origin")
                     (@arg INTERACTIVE: -i --interactive
                         "Allow editing the list of promotable packages")
@@ -152,7 +152,7 @@ pub fn get() -> App<'static, 'static> {
                         "The job id that was returned from \"hab bldr start\" \
                         (ex: 771100000000000000)")
                     (@arg CHANNEL: +takes_value +required "The target channel name")
-                    (@arg ORIGIN: -o --origin +takes_value
+                    (@arg ORIGIN: -o --origin +takes_value {valid_origin}
                         "Limit the demotable packages to the specified origin")
                     (@arg INTERACTIVE: -i --interactive
                         "Allow editing the list of demotable packages")
@@ -170,7 +170,7 @@ pub fn get() -> App<'static, 'static> {
                         (@arg GROUP_ID: +takes_value
                             "The group id that was returned from \"hab bldr job start\" \
                             (ex: 771100000000000000)")
-                        (@arg ORIGIN: -o --origin +takes_value
+                        (@arg ORIGIN: -o --origin +takes_value {valid_origin}
                             "Show the status of recent job groups created in this origin \
                             (default: 10 most recent)")
                     )
@@ -197,7 +197,7 @@ pub fn get() -> App<'static, 'static> {
                          be taken from the HAB_BLDR_URL environment variable if defined. (default: \
                          https://bldr.habitat.sh)")
                     (@arg CHANNEL: +required + takes_value "The channel name")
-                    (@arg ORIGIN: -o --origin +takes_value
+                    (@arg ORIGIN: -o --origin +takes_value {valid_origin}
                         "Sets the origin to which the channel will belong. Default is from \
                         'HAB_ORIGIN' or cli.toml")
                 )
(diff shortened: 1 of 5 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/list.rs
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+    }
+}
+
+/// Compare two `PackageIdent`s component by component:
+/// i.e. start with origin, then name, then version, then
+/// release. The first component to be not equal, then return
+/// the greater/lesser
+fn component_cmp(a: &PackageIdent, b: &PackageIdent) -> Ordering {
```
baumanj: Why not add this to the core implementation of `PackageIdent`? In fact, it might be possible to just add `#[derive(Ord, PartialOrd)]` to get the desired behavior with even less code. Then you can replace the `packages.sort_unstable_by(…)` above with just `packages.sort()`.

components/hab/src/main.rs
```diff
@@ -682,6 +683,13 @@ fn sub_pkg_path(m: &ArgMatches) -> Result<()> {
     command::pkg::path::start(&ident, &*FS_ROOT)
 }
 
+fn sub_pkg_list(ui: &mut UI, m: &ArgMatches) -> Result<()> {
+    let prefix = m.value_of("PKG_PREFIX").and_then(|u| Some(u.to_string()));
+    // .or_else(|| None);
```
baumanj: Left in by mistake?

components/hab/src/cli.rs
```diff
@@ -405,6 +405,14 @@ pub fn get() -> App<'static, 'static> {
                 (@arg PKG_IDENT: +required +takes_value
                     "A package identifier (ex: core/redis, core/busybox-static/1.42.2)")
             )
+            (@subcommand list =>
+                (about: "List all installed packages with a given prefix")
+                (aliases: &["li"])
+                (@arg PKG_PREFIX: +takes_value
```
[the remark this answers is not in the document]
jamesc: yeah, that looks like a useful change.  I originally made it more like `ls` but I can see how this approach fits with some of the existing usage.

One thing I see is that it probably conflicts with the previous "merge with search" suggestion as it takes the approach I did of accepting a string (but also does a LIKE '%PATTERN%' style match).

I'll sync with @mpeck and see how we oculd unify this a bit.

components/hab/src/command/pkg/list.rs
jamesc: `PackageIdent` already has `Ord` and `PartialOrd` but it's a different implementation which only compares the last 2 segments of a package ident - https://github.com/habitat-sh/core/blob/master/components/core/src/package/ident.rs#L226-L311

Happy to move it over if the fn is more generally useful but I guess I'd need a better name and it'd still be called via `sort_unstable_by` ?

components/hab/src/main.rs
jamesc: Oops :) fixed.

components/hab/src/command/pkg/list.rs
baumanj: Huh, that seems like a surprising default for `Ord` that comparing `my-origin/redis` and `core/redis` are equal. I'm guessing there's a need for that, but to my mind *that*'s the situation that needs a special comparison function and merely sorting a list of packages for display should use a default which considers all the fields. Maybe @fnichol can provide more insight, but the origin-ignoring comparisons seem like a pretty un-intuitive and surprising default to me.

components/hab/src/cli.rs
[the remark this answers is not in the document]
baumanj: I bet we can come up with something appropriate. And even if we don't merge `list` and `search`, I do think avoiding overloading the origin and package ident argument would improve usability.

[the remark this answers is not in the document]
jamesc: I talked to @mpeck on this and he thinks the `ls` style is better for scripting (pipe output of list into `uninstall` or `dependencies`)  I do like the --all option as listing everything when many packages installed should probably be more explicit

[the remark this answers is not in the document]
baumanj: I can get behind separate `ls` and `search` commands if the other usability stuff is addressed.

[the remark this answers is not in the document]
jamesc: Moved the the <--all | -o origin | PKG_IDENT> structure for input

components/hab/src/command/pkg/list.rs
jamesc: Yeah.  I modified the `Ord::cmp` for `PackageIdent` and no tests failed in Habitat, so I'm not touching the refactor of `Ord`/`PartialOrd` until I understand what it's supposed to be doing !

:+1: on moving the function into `PackageIdent` as first steps

jamesc: @baumanj Updated the code and it looks a lot cleaner.  Requires habitat-sh/core#84. Will update Cargo.lock when that is merged

components/hab/src/main.rs
```diff
@@ -1312,6 +1322,26 @@ fn org_param_or_env(m: &ArgMatches) -> Result<String> {
     }
 }
 
+/// Check to see if a package idenifier or
+///
```
christophermaier: Looks like you got distracted mid-documentation 😄

components/hab/src/main.rs
```diff
@@ -1312,6 +1322,26 @@ fn org_param_or_env(m: &ArgMatches) -> Result<String> {
...
+fn ident_or_origin(m: &ArgMatches) -> ListingType {
```
christophermaier: The name doesn't quite capture the functionality now.

Stepping back a bit, I wonder what it would be like to express this by implementing `From<&ArgMatches>` for `ListingType`, and if that might be a pattern we could use elsewhere in the CLI.

components/hab/src/cli.rs
```diff
@@ -403,6 +403,21 @@ pub fn get() -> App<'static, 'static> {
...
+                (aliases: &["li"])
+                (@group prefix =>
+                    (@attributes +required)
+                    (@arg ALL: -a --all
+                            "List all installed packages")
+                    (@arg ORIGIN: -o --origin +takes_value
+                            "An origin to list")
+                    (@arg PKG_IDENT: +takes_value {valid_ident}
+                    "A package identifier (ex: core/redis, core/busybox-static/1.42.2).")
+
```
christophermaier: What's here is completely fine, but is there any thought to having `--all` be the default, so `hab pkg list` would also work?

components/hab/src/main.rs
jamesc: That actually works pretty well.  I think the end logical conclusion for real grouping of implementation would be that the {valid_XXX} and the CLI `@arg` definition should be all together.  But that might be taking it a bit far

components/hab/src/cli.rs
jamesc: We had that before, but moved towards this more structured style during review.  I'd leave it as is and get feedback from the community

Cargo.toml
```diff
@@ -22,3 +22,6 @@ members = [
   "components/sup-client",
   "components/sup-protocol",
 ]
+
+[patch."https://github.com/habitat-sh/core.git"]
+habitat_core = { git = "https://github.com/jamesc/core.git", branch = "jc/ident-by_parts_cmp"}
```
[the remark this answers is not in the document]
jamesc: done

PR #5793, opened 2018-10-29
Refactor ManagerConfig generation from ArgMatches

mpeck:
Refactors generating a ManagerConfig from ArgMatches and adds some tests around it.

### DIFF
--- a/components/sup/src/main.rs
+++ b/components/sup/src/main.rs
@@ -30,6 +30,9 @@ extern crate time;
 extern crate tokio_core;
 extern crate url;
 
+#[cfg(test)]
+extern crate tempdir;
+
 use std::env;
 use std::io::{self, Write};
 use std::net::{SocketAddr, ToSocketAddrs};
@@ -44,13 +47,12 @@ use hcore::channel;
 use hcore::crypto::dpapi::encrypt;
 use hcore::crypto::{self, default_cache_key_path, SymKey};
 use hcore::env as henv;
+use hcore::service::ServiceGroup;
 use hcore::url::{bldr_url_from_env, default_bldr_url};
 use launcher_client::{LauncherCli, ERR_NO_RETRY_EXCODE};
 use protocol::{
     ctl::ServiceBindList,
-    types::{
-        ApplicationEnvironment, BindingMode, ServiceBind, ServiceGroup, Topology, UpdateStrategy,
-    },
+    types::{ApplicationEnvironment, BindingMode, ServiceBind, Topology, UpdateStrategy},
 };
 
 use sup::cli::cli;
@@ -202,27 +204,43 @@ fn sub_term(m: &ArgMatches) -> Result<()> {
 ////////////////////////////////////////////////////////////////////////
 
 fn mgrcfg_from_matches(m: &ArgMatches) -> Result<ManagerConfig> {
-    let mut cfg = ManagerConfig::default();
-    cfg.auto_update = m.is_present("AUTO_UPDATE");
-    cfg.update_url = bldr_url(m);
-    cfg.update_channel = channel(m);
+    let mut cfg = ManagerConfig {
+        auto_update: m.is_present("AUTO_UPDATE"),
+        update_url: bldr_url(m),
+        update_channel: channel(m),
+        http_disable: m.is_present("HTTP_DISABLE"),
+        organization: m.value_of("ORGANIZATION").map(|org| org.to_string()),
+        gossip_permanent: m.is_present("PERMANENT_PEER"),
+        ring_key: get_ring_key(m)?,
+        gossip_peers: get_peers(m)?,
+        ..Default::default()
+    };
     if let Some(addr_str) = m.value_of("LISTEN_GOSSIP") {
         cfg.gossip_listen = GossipListenAddr::from_str(addr_str)?;
     }
     if let Some(addr_str) = m.value_of("LISTEN_HTTP") {
         cfg.http_listen = http_gateway::ListenAddr::from_str(addr_str)?;
     }
-    cfg.http_disable = m.is_present("HTTP_DISABLE");
     if let Some(addr_str) = m.value_of("LISTEN_CTL") {
-        cfg.ctl_listen =
-            SocketAddr::from_str(addr_str).unwrap_or_else(|_err| protocol::ctl::default_addr());
+        cfg.ctl_listen = SocketAddr::from_str(addr_str)?;
     }
-    cfg.organization = m.value_of("ORGANIZATION").map(|org| org.to_string());
-    cfg.gossip_permanent = m.is_present("PERMANENT_PEER");
+    if let Some(watch_peer_file) = m.value_of("PEER_WATCH_FILE") {
+        cfg.watch_peer_file = Some(String::from(watch_peer_file));
+    }
+    if let Some(events) = m.value_of("EVENTS") {
+        cfg.eventsrv_group = ServiceGroup::from_str(events).ok();
+    }
+    Ok(cfg)
+}
+
+// Various CLI Parsing Functions
+////////////////////////////////////////////////////////////////////////
+
+fn get_peers(matches: &ArgMatches) -> Result<Vec<SocketAddr>> {
     // TODO fn: Clean this up--using a for loop doesn't feel good however an iterator was
     // causing a lot of developer/compiler type confusion
-    let mut gossip_peers: Vec<SocketAddr> = Vec::new();
-    if let Some(peers) = m.values_of("PEER") {
+    let mut gossip_peers = Vec::new();
+    if let Some(peers) = matches.values_of("PEER") {
         for peer in peers {
             let peer_addr = if peer.find(':').is_some() {
                 peer.to_string()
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/main.rs
```diff
@@ -464,3 +473,149 @@ fn update_svc_load_from_input(m: &ArgMatches, msg: &mut protocol::ctl::SvcLoad)
...
+        #[test]
+        fn gossip_listen_should_be_set() {
+            let config = config_from_cmd_str("hab-sup run --listen-gossip 1.1.1.1:1111");
+            let expected_addr = GossipListenAddr::from_str("1.1.1.1:1111")
+                .expect("Could not create GossipListenAddr");
+            assert_eq!(config.gossip_listen, expected_addr);
+
+            let config = config_from_cmd_str("hab-sup run");
+            let expected_addr = GossipListenAddr::from_str("0.0.0.0:9638")
+                .expect("Could not create GossipListenAddr");
```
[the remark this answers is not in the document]
mpeck: It does, we get `0.0.0.0:9638` as the default.

components/sup/src/main.rs
```diff
@@ -202,76 +202,85 @@ fn sub_term(m: &ArgMatches) -> Result<()> {
...
+    if let Some(events) = m.value_of("EVENTS") {
+        cfg.eventsrv_group = ServiceGroup::from_str(events).ok().map(Into::into);
+    }
+    Ok(cfg)
+}
+
+// Various CLI Parsing Functions
+////////////////////////////////////////////////////////////////////////
+
+fn get_peers(peers: clap::Values) -> Result<Vec<SocketAddr>> {
```
baumanj: I realize you're just moving existing code into a function, but at long as it's getting mutated, we might take this opportunity of fix some things up.

First, why not have `get_peers` take `m: &ArgMatches` similar to `get_ring_key`? Add an outer condition like so:
```rust
    if let Some(peers) = m.values_of("PEER") {
        for peer in peers {
            …
        }
    }
```

and you can get the same simplicity of passing `gossip_peers: get_peers(m)?` directly to the `ManagerConfig` initializer.

components/sup/src/main.rs
```diff
@@ -202,76 +202,85 @@ fn sub_term(m: &ArgMatches) -> Result<()> {
...
+            format!("{}:{}", peer, GOSSIP_DEFAULT_PORT)
+        };
+        let addrs: Vec<SocketAddr> = match peer_addr.to_socket_addrs() {
+            Ok(addrs) => addrs.collect(),
+            Err(e) => {
+                outputln!("Failed to resolve peer: {}", peer_addr);
+                return Err(sup_error!(Error::NameLookup(e)));
+            }
+        };
+        let addr: SocketAddr = addrs[0];
```
baumanj: This is unsafe since according to the [`to_socket_addrs` docs](http://doc.rust-lang.org/1.29.2/std/net/trait.ToSocketAddrs.html#tymethod.to_socket_addrs):
> The returned iterator may not actually yield any values depending on the outcome of any resolution performed.

In general, it's best to avoid the [`index` operation](http://doc.rust-lang.org/1.29.2/std/ops/trait.Index.html#tymethod.index) (which `[…]` is sugar for) because it will panic if the index is invalid.

In this case, it's not too hard to do something a little safer:
```rust
            if let Some(addr) = addrs.get(0) {
                gossip_peers.push(*addr);
            }
```
Or you could avoid the unnecessary `collect` and creation of the `addrs` `Vec` with:
```rust
            match peer_addr.to_socket_addrs() {
                Ok(mut addrs) => if let Some(addr) = addrs.next() {
                    gossip_peers.push(addr);
                },
```

It also might be worth thinking a bit about in what cases `to_socket_addrs` *could* return either 0 or more than 1 results and what we want to do in those cases.

components/sup/src/main.rs
```diff
@@ -202,76 +202,85 @@ fn sub_term(m: &ArgMatches) -> Result<()> {
...
+    let mut cfg = ManagerConfig {
+        auto_update: m.is_present("AUTO_UPDATE"),
+        update_url: bldr_url(m),
+        update_channel: channel(m),
+        http_disable: m.is_present("HTTP_DISABLE"),
+        organization: m.value_of("ORGANIZATION").map(|org| org.to_string()),
+        gossip_permanent: m.is_present("PERMANENT_PEER"),
+        ring_key: get_ring_key(m)?,
+        ..Default::default()
+    };
```
baumanj: I like what you're doing here. I'd like to see it taken even further, because I much prefer to have objects initialized all at once rather than in several steps. If you did something like:
```rust
    let gossip_listen = if let Some(addr_str) = m.value_of("LISTEN_GOSSIP") {
        GossipListenAddr::from_str(addr_str)?
    } else {
        GossipListenAddr::default()
    };

    let eventsrv_group = if let Some(events) = m.value_of("EVENTS") {
        ServiceGroup::from_str(events).ok()
    } else {
        None
    };

    let ctl_listen = if let Some(addr_str) = m.value_of("LISTEN_CTL") {
        SocketAddr::from_str(addr_str).ok()
    } else {
        None
    };

    let http_listen = if let Some(addr_str) = m.value_of("LISTEN_HTTP") {
        http_gateway::ListenAddr::from_str(addr_str)?
    } else {
        http_gateway::ListenAddr::default()
    };

    Ok(ManagerConfig {
        auto_update: m.is_present("AUTO_UPDATE"),
        custom_state_path: None,
        eventsrv_group,
        update_url: bldr_url(m),
        update_channel: channel(m),
        gossip_listen,
        ctl_listen: ctl_listen.unwrap_or(protocol::ctl::default_addr()),
        http_listen,
        http_disable: m.is_present("HTTP_DISABLE"),
        gossip_peers: get_peers(m)?,
        gossip_permanent: m.is_present("PERMANENT_PEER"),
        ring_key: get_ring_key(m)?,
        organization: m.value_of("ORGANIZATION").map(|org| org.to_string()),
        watch_peer_file: m.value_of("PEER_WATCH_FILE").map(Into::into),
    })
```

You could do away with the `Default` trait for `ManagerConfig` entirely. You'd have to [remove these tests](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L2018-L2054), (or migrate them to the place you're adding), but I think that would be appropriate. In fact,  I think you can eliminate the [`ManagerConfig::custom_state_path`](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L131) field entirely.

components/sup/src/main.rs
```diff
@@ -464,3 +473,149 @@ fn update_svc_load_from_input(m: &ArgMatches, msg: &mut protocol::ctl::SvcLoad)
...
+            let config = config_from_cmd_str("hab-sup run --peer-watch-file foobar");
+            assert_eq!(config.watch_peer_file, Some("foobar".to_string()));
+        }
+
+        #[test]
+        fn events_is_set() {
+            let config = config_from_cmd_str("hab-sup run --events event.service");
+            assert!(config.eventsrv_group.is_some());
+        }
+
```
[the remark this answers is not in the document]
mpeck: There are some issues unit testing functions that utilize env vars. Namely, it's not possible to effectively test them in isolation. We've already talked about this and I think we have a path forward, I'm just recording it here to help alleviate some potential confusion.

components/sup/src/main.rs
mpeck: This is actually just the first step in a slightly grander refactor that I think leads more to what you're suggesting here. My plan is to move this code out of main and then leverage default values from clap to eliminate a bunch of this code. In the end we should end up with just a `ManagerConfig {...}` block without any of the `if` bits.

I thought that that would warrant it's own PR though, so I didn't include it here.

components/sup/src/main.rs
```diff
@@ -464,3 +473,149 @@ fn update_svc_load_from_input(m: &ArgMatches, msg: &mut protocol::ctl::SvcLoad)
...
+                .get_matches_from_safe(cmd_vec)
+                .expect("Error while getting matches");
+            let (_, sub_matches) = matches.subcommand();
+            let sub_matches = sub_matches.expect("Error getting sub command matches");
+
+            mgrcfg_from_matches(&sub_matches).expect("Could not get config")
+        }
+
+        #[test]
+        fn auto_update_should_be_set() {
```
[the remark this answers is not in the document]
mpeck: I get what you're saying, but my next step is to move the manger config generation bit out of here and refactor how we're dealing with default args. The testing might be a bit granular, but I didn't want to miss anything.

components/sup/src/main.rs
```diff
@@ -30,6 +30,9 @@ extern crate time;
 extern crate tokio_core;
 extern crate url;
 
+#[cfg(test)]
+extern crate tempdir;
```
christophermaier: In light of https://github.com/habitat-sh/habitat/issues/5806, we should just use `tempfile` from the beginning.

(Sorry for leading you astray, I just realized it was deprecated yesterday 😢)

components/sup/src/main.rs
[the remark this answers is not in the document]
christophermaier: I'd just use `let expected_addr = GossipListenAdd::default();` then. It's less code and more intention-revealing; you _know_ that's the default value just by looking at it.

components/sup/src/main.rs
```diff
@@ -464,3 +480,173 @@ fn update_svc_load_from_input(m: &ArgMatches, msg: &mut protocol::ctl::SvcLoad)
...
+            let config = config_from_cmd_str("hab-sup run --listen-http 2.2.2.2:2222");
+            let expected_addr = http_gateway::ListenAddr::from_str("2.2.2.2:2222")
+                .expect("Could not create http listen addr");
+            assert_eq!(config.http_listen, expected_addr);
+        }
+
+        #[test]
+        fn http_listen_is_set_default_when_not_specified() {
+            let config = config_from_cmd_str("hab-sup run");
+            let expected_addr = http_gateway::ListenAddr::from_str("0.0.0.0:9631")
```
christophermaier: Same here, but with `ListenAddr::default()`

PR #5348, opened 2018-07-17
Make sup-run launch the supervisor in its own session, detached from the TTY

baumanj:
1. Use setsid to start the supervisor in its own session and process group.
   It's the process group that's important, though since that prevents it from
   being killed by a stray ^c on the CLI. That SIGINT goes to the entire
   process group, and when sup-run is executed as part of the profile startup,
   the hab-launch process will share the same process group.

2. Use disown to remove the supervisor jobspec from the list of active jobs.
   This prevents it from showing up in `jobs` output and from being brought
   to the foreground. There's no need to do that since, `sup-log` and
   `sup-term` still work to view logs and stop the supervisor.

Revolves https://github.com/habitat-sh/habitat/issues/5312
Resolves https://github.com/habitat-sh/habitat/issues/4397
Resolves https://github.com/habitat-sh/habitat/issues/5271

Signed-off-by: Jon Bauman <5906042+baumanj@users.noreply.github.com>

PR #5800, opened 2018-10-31
Deprecate the confusing sup-log command in favor of directly accessing logs

baumanj:
This command seems to have caused more confusion about accessing logs than it has created convenience.

This adds a `$SUP_LOG_FILE` environment variable and the command prints out
information necessary for accessing the log directly.

New behavior:
```
[1][HAB_STUDIO_BINARY][default:/src:0]# sup-log
This command is deprecated. View Supervisor logs directly at $SUP_LOG_FILE (/hab/sup/default/sup.log)

[2][HAB_STUDIO_BINARY][default:/src:0]# tail $SUP_LOG_FILE
→ Using core/openssl/1.0.2l/20171014213633
→ Using core/xz/5.2.2/20170513214327
→ Using core/zlib/1.2.8/20170513201911
✓ Installed core/hab-launcher/8744/20181016214636
★ Install of core/hab-launcher/8744/20181016214636 complete with 1 new packages installed.
hab-sup(MR): core/hab-sup (core/hab-sup/0.65.0/20181004221503)
hab-sup(MR): Supervisor Member-ID 842df9dabdc74b67b86846e508d601ec
hab-sup(MR): Starting ctl-gateway on 127.0.0.1:9632

[3][HAB_STUDIO_BINARY][default:/src:0]# tail /hab/sup/default/sup.log
… [same as above]
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -115,7 +115,7 @@ Then you can actually exercise the software as follows:
 
 1. It pulls down the correct studio image
 1. That studio's `hab` is at the correct version
-1. A `sup-log` shows a running supervisor (if `sup-log` does not show a supervisor running, run `hab install core/hab-sup --channel release_channel` then `hab sup run`)
+1. A `hab sup status` shows a running supervisor (if it does not show a supervisor running, run `hab install core/hab-sup --channel release_channel` then `hab sup run`)
 1. Verify that the supervisor is the correct version by running `ls /hab/pkgs/core/hab-sup`
 
 When testing the linux studio, you will need to `export CI_OVERRIDE_CHANNEL` to the rc channel of the release. So if you are releasing 0.75.2, the channel would be `rc-0.75.2`.
--- a/components/rootless_studio/default/profile
+++ b/components/rootless_studio/default/profile
@@ -55,11 +55,15 @@ if [[ -n "${HAB_STUDIO_SUP}" ]]; then
   HAB_STUDIO_SUP="$(echo "$HAB_STUDIO_SUP" | sed 's/__sp__/ /g')"
 fi
 
+export SUP_LOG_FILE=/hab/sup/default/sup.log
+
 sup-run() {
   mkdir -p /hab/sup/default
+  touch $SUP_LOG_FILE
   echo "--> Launching the Habitat Supervisor in the background..."
   echo "    Running: hab sup run $*"
-  setsid hab sup run "$@" > /hab/sup/default/sup.log 2>&1 &
+  echo "    Logging to $SUP_LOG_FILE"
+  setsid hab sup run "$@" > $SUP_LOG_FILE 2>&1 &
   disown $!
   echo "    * Use 'hab svc start' & 'hab svc stop' to start and stop services"
   echo "    * Use 'sup-log' to tail the Supervisor's output (Ctrl+c to stop)"
@@ -85,10 +89,7 @@ sup-term() {
 }
 
 sup-log() {
-  mkdir -p /hab/sup/default
-  touch /hab/sup/default/sup.log
-  echo "--> Tailing the Habitat Supervisor's output (use 'Ctrl+c' to stop)"
-  tail -f /hab/sup/default/sup.log
+  echo "This command is deprecated. View Supervisor logs directly at \$SUP_LOG_FILE ($SUP_LOG_FILE)"
 }
 
 alias sr='sup-run'
--- a/components/rootless_studio/default/profile
+++ b/components/rootless_studio/default/profile
@@ -66,7 +66,6 @@ sup-run() {
   setsid hab sup run "$@" > $SUP_LOG_FILE 2>&1 &
   disown $!
   echo "    * Use 'hab svc start' & 'hab svc stop' to start and stop services"
-  echo "    * Use 'sup-log' to tail the Supervisor's output (Ctrl+c to stop)"
   echo "    * Use 'sup-term' to terminate the Supervisor"
   if [[ -z "${HAB_STUDIO_SUP:-}" ]]; then
     echo "    * To pass custom arguments to run the Supervisor, export"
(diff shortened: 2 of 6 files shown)


thesentinels:

Thank you for contributing!

mwrock:
Hmm. Has it? I use it all the time and find it super convenient. Just curious where the confusion lies.

baumanj:
The confusion is that people don't realize the supervisor log is just a regular file and they're uncertain about how to exit `sup-log` since it's unclear that it's just `tail -f` under the hood.

Some of the problem definitely owes to accidental killing of the supervisor itself before https://github.com/habitat-sh/habitat/pull/5348 was implemented, but as it stands I think the command doesn't provide much utility in exchange for the useful details it obfuscates.

christophermaier:
@mwrock When you boil it down, all `sup-log` is doing is `tail`ing a file. The interaction between it and `sup-run` (or rather, with the running Supervisor process) has caused confusion for people using it, where they're not sure whether it's "safe" to `^C` to get out of it, whether it should be in the foreground or background, and so on. 

... and now I see that @baumanj has said everything else I was going to say 😄

mwrock:
I'm sure `sup-log` as worked its way into the muscle memory of several users so if we are gonna boot it we should over clarify the docs and with internal/external community what we are doing.

baumanj:
I think I've fixed all the docs that make reference to `sup-log`. Except for some blog posts, but I think we're generally not in the practice of changing all those when features change.

For anyone with the command in muscle-memory (or who reads old docs), the output itself should make the situation clear:
```
[1][HAB_STUDIO_BINARY][default:/src:0]# sup-log
This command is deprecated. View Supervisor logs directly at $SUP_LOG_FILE (/hab/sup/default/sup.log)
```

And of course, people can always add an alias for themselves if they like.

mwrock:
Thanks for the doc changes! I think I'd like to see a little more socialization of this idea before merge and maybe some additional approvals. `hab-sup` is used in just about every tutorial, talk and workshop on hab so my concern is that we might be taking something away that others find useful. This change does not directly impact the scenarios where I tend to play (windows), so I think my "vote" should be weighted appropriately.  I am all for eliminating confusion but I just don't really grasp why this would confuse most people. Then again there is alot in this world I do not grasp.

cnunciato:
My confusion hasn't been with not knowing the supervisor log is just a regular file, but with not not knowing when i can and can't use it; it's weird to me, for instance, that i can `sl` in the studio, but there's no `hab sup log` available to me outside the studio. Of course I understand the technical difference between the two situations, but the asymmetry here is what I've personally found most unsatisfying.

So my vote would actually be to go a step further here with a `hab sup log` command that knows how to show me the logs, wherever they are.

christophermaier:
@cnunciato That asymmetry is a good point. However, I think it would be challenging to do `hab sup log` correctly, as that depends on exactly how someone is deploying the Supervisor. We'd have to deal with `systemd`, `runit`, and any other kind of process supervisor someone might use, for instance.

baumanj:
I like that idea, @cnunciato! Unfortunately, I don't think it's possible without a significant arch change because the supervisor doesn't _know_ where its logs are going; it just writes to stdout. You can see that if you run `hab sup run` directly instead of using the `sup-run` helper, `sup-log` won't work (unless you happen to redirect the supervisor output to the `/hab/sup/default/sup.log`, where `sup-log` expects it).

If we made things different and the log was an argument to `hab sup run`, we could do something better, but this is really just a small patch to help with user confusion that I've seen several times from users who don't have the familiarity that the team itself does.

cnunciato:
@christophermaier @baumanj Yep, I suspected that might be the case, but figured I'd throw it out there as a suggestion anyway. 😄  So in the absence of a command that does something useful, I'm all right with this change, since it does seem to simplify and clarify.

baumanj:
@mwrock, @elliott-davis: anything I can do to make this change more acceptable for y'all? If not, that's fine, we don't have to be in perfect harmony on everything.

mwrock:
I'd be most interested in hearing commentary from @fnichol @smacfarlane and @nellshamrell who have probably spent the most intimate time with that command.

Don't get me wrong, I am totally for clarification and REALLY appreciate the intent of this PR but just want to be super careful about this.

bdangit:
My vote is don’t get rid of this functionality until we have another way to get the logs from the supervisor that you deemed production worthy.  Sure we can ask people to tail the log, but doesn’t that get away from the essence of habitat that it has everything included? (Well may be not those exact words).  You can express a warning that sup-log only works inside studio and that it really is tailing a certain location.

baumanj:
@bdangit: as a script helper that only exists inside the studio, `sup-log` is never used in production, so I'm not sure what you mean by "production worthy".

I like @cnunciato's suggestion of moving towards a world where one can have something like `hab log`, but in the meantime, I'd like people to know where their logs are and be freed up to use all their normal tools to interrogate them.

bdangit:
@baumanj fair point!  Perhaps “production worthy” is not the right phrase to use.  While it is not useable outside of the studio, `sup-log` has been become very useful for debugging services, even more useful than `hab svc status`.  Taking it away just seems very crippling. 

Definitely would love to see `hab log` or some form of it to come about.

baumanj:
I hear you @bdangit. And that's why I'm not just eliminating the command. If you're used to using it, it'll provide this helpful message:
```
[2][default:/src:0]# sup-log
This command is deprecated. View sup logs directly at $SUP_LOG_FILE (/hab/sup/default/sup.log)
```

I think that should suffice to smooth any transition bumps. I'm pretty confident the people who are using habitat should be comfortable with viewing text files on the CLI. If you have any suggestions as to how that could be a better experience (short of a `hab log` command), I'm all ears.

baumanj:
Sounds like the consensus is against this particular change, but I think this was a useful discussion that has surfaced a lot of useful ways we can address the current situation. I'll file some issues and link them back here.

baumanj:
Filed https://github.com/habitat-sh/habitat/issues/5818

PR #5804, opened 2018-11-01
Reduce excessive disk I/O when persisting rumors

christophermaier:
The Butterfly server writes all rumors to disk periodically for
durability. It also writes rumors to disk in the `Drop` implementation
of the `Server` struct. While this could be used to implement a "save
on shutdown" behavior, it also ended up causing a worst-case behavior
of writing the rumors out five times _per second_. This is because a
copy of the Butterfly server is passed into each of the five
short-lived worker threads that are spawned every second for pushing
rumors out to network members. When these threads exit, they drop
their copy of the Butterfly server struct, triggering a write.

These worker threads are started when there are new rumors to send
out, so the more active the network, the more frequently these writes
would happen. With larger networks, there are both more members to
send rumors to, as well as a larger number of rumors to persist. Thus,
large and active networks would generate scenarios of excessive disk
load.

Here, we simply remove the offending `Drop` implementation, thereby
eliminating the 5x/s worst case scenario. For persistence, we instead
rely on the existing periodic write of once every 30s. Additionally,
we add an explicit call to persist the data at Supervisor shutdown
time.

Finally, the logging message for a successful disk write is now
`info`, rather than `debug`, since this is an important event to
notify about.

### DIFF
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -1174,15 +1174,12 @@ impl Server {
         message::unwrap_wire(payload, (*self.ring_key).as_ref())
     }
 
-    fn persist_data(&self) {
+    pub fn persist_data(&self) {
         if let Some(ref dat_file) = *self.dat_file.write().expect("DatFile lock poisoned") {
             if let Some(err) = dat_file.write(self).err() {
                 error!("Error persisting rumors to disk, {}", err);
             } else {
-                debug!(
-                    "Successfully persisted rumors to disk: {}",
-                    dat_file.path().display()
-                );
+                info!("Rumors persisted to disk: {}", dat_file.path().display());
             }
         }
     }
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

stevendanna:
Wowzers.

adamhjk:
Is this info because we want it to stand out from the noise in debugging? Because it's going to make `info` pretty noisy too.

christophermaier:
I'd like to let this play out at `info` for now. This allows "normal" activities to rise above the noise of the `debug` level, which is particularly noisy with Tokio output. While it's true that `env_logger` can be very finely tuned depending on the value provided for `RUST_LOG`, being able to easily surface this information with a simple `RUST_LOG=info` is really nice from a debugging perspective, particularly as a customer without intimate knowledge of the Rust module structure of the Supervisor code (which you need to know to tune the logger as described).

Had we had something like this in place before this fix, for instance, it would have been immediately obvious that we were writing to this file far too often.

PR #5807, opened 2018-11-01
Add validation for all CLI args which accept package identifiers

baumanj:
Since we added `valid_ident` in https://github.com/habitat-sh/habitat/pull/5801, we might as well use it everywhere to provide a more consistent UX. This also allows downstream consumers to simplify their logic, but this change doesn't modify any of that downstream logic.


thesentinels:

Thank you for contributing!

jamesc:
nice :+1:

PR #5808, opened 2018-11-02
Add a --render-only flag to svc load that only renders templates

mwrock:
This allows one to `hab svc load` and pass a `--render-only` flag which will simply render its `config` and `hook` templates and not actually start and run the service.

## Why would you want such functionality
There are scenarios where one wants a habitat package to install software (involving hook execution) or perform one time system configuration operations but these types of packages do not act like a "service" because they are essentially "fire and forget" operations. However users run them as a service because they want to leverage health checks, update strategies and need to execute bahavior that can interact with configuration data and templates.

In order to eliminate the need to load these kinds of packages as individual services, we will create a custom Habitat Service that will wrap the installation, updates and health checking of a collection of packages that should be much lighter weight than running them as separate services.

This service wil need to be able to render the hooks and config templates of such packages and they will be able to leverage the functionality in this PR to do so.

## Refactorings this includes

* The `RenderContext`'s `Svc` is now an `Option<Svc>`. The `Svc` basically encapsulates a service's census status and these packages will not be census members so the `CensusRing` will not include them.

* The `CensusRing` is added to the Supervisor `Manager`'s state. This allow us to pass the active `CensusRing` into the `RenderContext`. Although the package itself is not in the ring, it may need to bind to services that are in the ring.

## Alternative approaches
I recognize that this flag is symantically awkward on the `load` command because nothing is actually being loaded. I originally wrote this as a separate command `hab svc render` that took nearly identical arguments to `load`. However including this in `load` eliminates alot of boiler plate code and also makes this functionality more of a "low profile" change. Chances are this command will only be used by the above mentioned service so not giving it the glory of its own command may cause less confusion.

### DIFF
--- a/components/sup-protocol/protocols/ctl.proto
+++ b/components/sup-protocol/protocols/ctl.proto
@@ -104,31 +104,8 @@ message SvcLoad {
   optional sup.types.Topology topology = 12;
   // Update strategy for the service.
   optional sup.types.UpdateStrategy update_strategy = 13;
-}
-
-// Request to render templates of a service.
-message SvcRender {
-  // Package identifier for the service to load. Using a more qualified identifier will load a
-  // more specific package.
-  optional sup.types.PackageIdent ident = 1;
-  // An application environment to place the service in.
-  optional sup.types.ApplicationEnvironment application_environment = 2;
-  // List of service binds to use when configuring the service.
-  optional ServiceBindList binds = 3;
-  // Set to true if any binds were set by the caller. This field is needed because a blank list
-  // and map will be present in a case when no binds were set *and* when binds are removed.
-  optional bool specified_binds = 4;
-  // Indicate how bind availability affects service start-up
-  optional sup.types.BindingMode binding_mode = 5;
-  // Remote http URL for the Builder service to receive package updates from.
-  optional string bldr_url = 6;
-  // Remote channel on the Builder service to receive package updates from.
-  optional string bldr_channel = 7;
-  // A filepath on disk which can be specified to override the package's configuration and hooks.
-  // This is useful when testing services on a local Supervisor before packaging them.
-  optional string config_from = 8;
-  // Service group name for the service.
-  optional string group = 9 [default = "default"];
+  // only render package templates
+  optional bool render_only = 15 [default = false];
 }
 
 // Request to unload a loaded service.
--- a/components/sup-protocol/protocols/ctl.proto
+++ b/components/sup-protocol/protocols/ctl.proto
@@ -106,6 +106,31 @@ message SvcLoad {
   optional sup.types.UpdateStrategy update_strategy = 13;
 }
 
+// Request to render templates of a service.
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -611,18 +611,7 @@ impl Manager {
...
-                "Can't create directory {}: {}",
-                service.pkg.svc_path.display(),
-                e
-            );
-            outputln!(
-                "If this service is running as non-root, you'll need to create \
-                 {} and give the current user write access to it",
-                service.pkg.svc_path.display()
-            );
-            outputln!("{} failed to start", &spec.ident);
```
christophermaier: Why was all this removed?

components/sup/src/manager/mod.rs
```diff
@@ -1106,11 +1122,41 @@ impl Manager {
...
-                    ))?;
+                    match opts.render_only {
+                        None => {
+                            Self::save_spec_for(&mgr.cfg, spec)?;
+                            req.info(format!(
+                                "The {} service was successfully loaded",
+                                spec.ident
+                            ))?;
+                        }
+                        Some(_) => {
```
christophermaier: By using `Some(_)` it seems like we'd be open to potentially confusing scenarios where `Some(false)` behaves as though it were `true`.

baumanj:
I agree with @christophermaier. This strains the abstraction, so let's sit down and figure out if our abstraction needs to be reconsidered or if there's a more intuitive way to address the use case. @mpeck will probably want in on this discussion as well.

components/sup/src/manager/mod.rs
mwrock: It was moved to a `create_svc_path` function

mwrock:
Yes I'll arrange for us to meet next week. I totally agree there is some confusion this can introduce and would love to explore this more deeply to come out with the right approach.

components/sup-protocol/protocols/ctl.proto
```diff
@@ -104,6 +104,8 @@ message SvcLoad {
   optional sup.types.Topology topology = 12;
   // Update strategy for the service.
   optional sup.types.UpdateStrategy update_strategy = 13;
+  // only render package templates
+  optional bool render_only = 15 [default = false];
```
baumanj: Why not 14?

christophermaier: `14` is already used by `binding_mode` above.

baumanj: 🤦‍♂️  Maybe we should put these in order.

christophermaier:
Since we talked offline about this approach and figured out a new and dramatically different one to pursue (namely, look at extracting the templating logic and introducing a kind of "post-install" hook that runs at `hab pkg install` time), I'm going to close out this PR.

PR #5809, opened 2018-11-02
Add commented out patch code for testing habitat_core changes locally

baumanj:
I'm always having to look up the syntax, so I figure this will save me (and hopefully others) some time.


jamesc:
:+1: I have a version stashed away that I copy into place so this is much better :)

PR #5810, opened 2018-11-02
Simplify elections code by replacing ServiceGroup structs with strings

baumanj:
This should have no effect on behavior, it's a simplification that came out of
auditing the elections code.

The change is fairly seamless because ServiceGroup is String wrapper, but for
the purpose of elections, it's only used as an opaque identifier. Using a
string directly allows for simplification of code where the complexity is
appropriate for validating user input that's unnecessary for data that either
comes from protocol messages or existing ServiceGroup instances.

See https://github.com/habitat-sh/core/blob/master/components/core/src/service.rs#L85

### DIFF
--- a/components/butterfly-test/src/lib.rs
+++ b/components/butterfly-test/src/lib.rs
@@ -51,7 +51,7 @@ lazy_static! {
 #[derive(Debug)]
 struct NSuitability(u64);
 impl Suitability for NSuitability {
-    fn get(&self, _service_group: &ServiceGroup) -> u64 {
+    fn get(&self, _service_group: &str) -> u64 {
         self.0
     }
 }
@@ -518,7 +518,7 @@ impl SwimNet {
     }
 
     pub fn add_election(&mut self, member: usize, service: &str) {
-        self[member].start_election(ServiceGroup::new(None, service, "prod", None).unwrap(), 0);
+        self[member].start_election(&ServiceGroup::new(None, service, "prod", None).unwrap(), 0);
     }
 }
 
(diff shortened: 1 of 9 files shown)


thesentinels:

Thank you for contributing!

components/butterfly-test/src/lib.rs
```diff
@@ -518,7 +518,7 @@ impl SwimNet {
     }
 
     pub fn add_election(&mut self, member: usize, service: &str) {
-        self[member].start_election(ServiceGroup::new(None, service, "prod", None).unwrap(), 0);
+        self[member].start_election(&ServiceGroup::new(None, service, "prod", None).unwrap(), 0);
```
baumanj: Would this be preferable?
```rust
        self[member].start_election(&format!("{}.prod", service), 0);
```

components/butterfly/src/rumor/election.rs
```diff
@@ -303,15 +299,15 @@ mod tests {
     fn create_election(member_id: &str, suitability: u64) -> Election {
         Election::new(
             member_id,
-            ServiceGroup::new(None, "tdep", "prod", None).unwrap(),
+            &ServiceGroup::new(None, "tdep", "prod", None).unwrap(),
```
baumanj: Would this be preferable?
```suggestion
            "tdep.prod",
```

components/butterfly/src/server/mod.rs
```diff
@@ -974,37 +972,17 @@ impl Server {
...
-            let sg = match ServiceGroup::from_str(&service_group) {
-                Ok(sg) => sg,
-                Err(e) => {
-                    error!(
-                        "Failed to process service group from string '{}': {}",
-                        service_group, e
-                    );
-                    return;
-                }
-            };
```
baumanj: These blocks were the the original inspiration for the change.

christophermaier:
Had some Slack discussion with @baumanj ... we'll move forward with this approach since the rest of the lower-level code is explicitly based around `String`s.

PR #5812, opened 2018-11-02
Reworks sup-term to use hab sup term

### LINKED ISSUE #5690: Rework `sup-term` to use `hab sup term`
It seems like the `sup-term` helper function in the Studio could be reworked to use `hab sup term`, rather than essentially rewriting it in Bash.

https://github.com/habitat-sh/habitat/blob/d68309486b02dca9491ed4d9933e40e835631a20/components/studio/libexec/hab-studio-type-default.sh#L176-L186

https://github.com/habitat-sh/habitat/blob/2f49dd066f082ca72ef0a0d37d6526f25736efe5/components/rootless_studio/default/profile#L75-L85

Mentioned by @stevendanna in #5688 

cc: @fnichol

rsertelon:
> Closes #5690

This is a first attempt. I think it might become better, but would need some advice for that. (Except you think it's ok as is ;)).

### DIFF
--- a/components/rootless_studio/default/profile
+++ b/components/rootless_studio/default/profile
@@ -73,14 +73,10 @@ sup-run() {
 }
 
 sup-term() {
-  local pid_file="/hab/sup/default/LOCK"
-  if [ -f "$pid_file" ]; then
-    echo "--> Killing Habitat Supervisor running in the background..."
-    kill "$(cat "$pid_file")" \
-      && (echo "    Supervisor killed." && rm -f "$pid_file") \
-      || echo "--> Error killing Supervisor."
+  if hab sup term ; then
+    echo "--> Killed Habitat Supervisor running in the background."
   else
-    echo "--> No Launcher pid file found, Supervisor may not be running."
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/rootless_studio/default/profile
```diff
@@ -73,14 +73,10 @@ sup-run() {
...
-  if [ -f "$pid_file" ]; then
-    echo "--> Killing Habitat Supervisor running in the background..."
-    kill "$(cat "$pid_file")" \
-      && (echo "    Supervisor killed." && rm -f "$pid_file") \
-      || echo "--> Error killing Supervisor."
+  if hab sup term ; then
+    echo "--> Killed Habitat Supervisor running in the background."
   else
-    echo "--> No Launcher pid file found, Supervisor may not be running."
+    echo "--> Error killing Supervisor, may not be running"
```
baumanj: A minor language tweak (same for the other file)
```suggestion
    echo "--> Error killing Supervisor; it may not be running"
```

rsertelon:
@baumanj changes made :)

PR #5813, opened 2018-11-02
Add Windows-friendly behavior to `hab pkg env`

mwrock:
This PR depends on https://github.com/habitat-sh/core/pull/86

This makes `hab pkg env` windows friendly:

```
[HAB-STUDIO] Habitat:\src> hab pkg env core/powershell
$env:PATH="C:\hab\studios\dev--blank\hab\pkgs\core\powershell\6.0.0\20180124123643\bin"
```

This will be necessary for the upcoming "package updater" service.

### DIFF
--- a/components/hab/src/command/pkg/env.rs
+++ b/components/hab/src/command/pkg/env.rs
@@ -12,18 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::path::Path;
 
 use hcore::package::{PackageIdent, PackageInstall};
 
 use error::Result;
 
-// TODO: This needs a windows compatible version
 pub fn start(ident: &PackageIdent, fs_root_path: &Path) -> Result<()> {
     let pkg_install = PackageInstall::load(ident, Some(fs_root_path))?;
     let env = pkg_install.environment_for_command()?;
+    render_environment(env);
+    Ok(())
+}
+
+#[cfg(unix)]
+fn render_environment(env: HashMap<String, String>) {
     for (key, value) in env.into_iter() {
         println!("export {}=\"{}\"", key, value);
     }
-    Ok(())
+}
+
+#[cfg(windows)]
+fn render_environment(env: HashMap<String, String>) {
+    for (key, value) in env.into_iter() {
+        println!("$env:{}=\"{}\"", key, value);
+    }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1028,7 +1028,7 @@ dependencies = [
 [[package]]
 name = "habitat_core"
 version = "0.0.0"
-source = "git+https://github.com/habitat-sh/core.git#0f02f3a87bc6cf70c80a7086b9d8480518b14317"
+source = "git+https://github.com/habitat-sh/core.git#c94625e54ead18821e2a0a14b69ee68867e93406"
 dependencies = [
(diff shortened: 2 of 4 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/env.rs
```diff
@@ -12,18 +12,56 @@
...
+
+    let mut paths: Vec<PathBuf> = match env.get("PATH") {
+        Some(path) => env::split_paths(&path).collect(),
+        None => vec![],
+    };
+    for i in 0..paths.len() {
+        if paths[i].starts_with("/") {
+            paths[i] = Path::new(&*FS_ROOT_PATH).join(paths[i].strip_prefix("/").unwrap());
+        }
+    }
```
christophermaier: Does this logic exist anywhere else in the code? It seems like something that would be generally useful as some kind of library function.

components/hab/src/command/pkg/env.rs
```diff
@@ -12,18 +12,56 @@
...
+    let joined = env::join_paths(paths)?;
+    let mut mut_env = env.clone();
```
christophermaier: Instead of cloning the HashMap, it might be better to just do a `println!` of the modified `PATH`, and then print out the rest of the unmodified map, filtering out the unmodified `PATH` entry.

That way we don't do any unnecessary allocations, and we also more clearly drive home what the key difference is between the two implementations.

components/hab/src/command/pkg/env.rs
```diff
@@ -12,18 +12,56 @@
...
+
+    mut_env.insert(
+        String::from("PATH"),
+        joined
+            .into_string()
+            .expect("Unable to convert OsStr path to string!"),
+    );
+
+    for (key, value) in mut_env.into_iter() {
```
baumanj: Since this is just to print the values, rather than cloning the `env` argument and inserting, you could:
```rust
    let joined = &env::join_paths(paths)?
        .into_string()
        .expect("Unable to convert OsStr path to string!");

    for (key, value) in env.iter() {
        println!(
            "$env:{}=\"{}\"",
            key,
            if key == "PATH" { joined } else { value }
        );
    }
```

components/hab/src/command/pkg/env.rs
baumanj: [*jinx*](url
https://github.com/habitat-sh/habitat/pull/5813#discussion_r230834269)

components/hab/src/command/pkg/env.rs
mwrock: It does. I just added https://github.com/habitat-sh/core/pull/86 to move this prepending there.

mwrock:
@cm I think this addresses your comments now

components/hab/src/command/pkg/env.rs
```diff
@@ -12,18 +12,30 @@
...
 
 use hcore::package::{PackageIdent, PackageInstall};
 
 use error::Result;
 
-// TODO: This needs a windows compatible version
 pub fn start(ident: &PackageIdent, fs_root_path: &Path) -> Result<()> {
     let pkg_install = PackageInstall::load(ident, Some(fs_root_path))?;
     let env = pkg_install.environment_for_command()?;
+    render_environemt(env);
```
christophermaier: Spelling! 😱

PR #5825, opened 2018-11-07
Fix bug where having no excludes for uninstall would trigger a panic

jamesc:

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -1365,7 +1365,7 @@ fn install_sources_from_matches(matches: &ArgMatches) -> Result<Vec<InstallSourc
 fn excludes_from_matches(matches: &ArgMatches) -> Vec<PackageIdent> {
     matches
         .values_of("EXCLUDE")
-        .unwrap() // Required via clap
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/main.rs
```diff
@@ -1363,11 +1363,13 @@ fn install_sources_from_matches(matches: &ArgMatches) -> Result<Vec<InstallSourc
 }
 
 fn excludes_from_matches(matches: &ArgMatches) -> Vec<PackageIdent> {
-    matches
-        .values_of("EXCLUDE")
-        .unwrap() // Required via clap
```
baumanj: I think a simpler fix would be changing the `unwrap` to `unwrap_or_default`.

jamesc: You're a master of the clap API :)  Done.

baumanj: FYI, that's not clap-specific, rather a feature of [`Option`](http://doc.rust-lang.org/1.30.0/core/option/enum.Option.html#method.unwrap_or_default).

PR #5835, opened 2018-11-08
Jc/uninstall exclude services

jamesc:
If a supervisor is running, we now additionally check if any
package we're trying to delete is currently running and skip if it
is


thesentinels:

Thank you for contributing!

components/hab/src/main.rs
```diff
@@ -1490,6 +1501,51 @@ where
...
+    let sup_addr = protocol::ctl::default_addr();
+    let msg = protocol::ctl::SvcStatus::default();
+
+    let mut out: Vec<PackageIdent> = vec![];
+    SrvClient::connect(&sup_addr, &secret_key)
+        .and_then(|conn| {
+            conn.call(msg).for_each(|reply| {
+                match reply.message_id() {
+                    "ServiceStatus" => {
+                        let m = reply.parse::<protocol::types::ServiceStatus>().unwrap();
```
baumanj: Are these `unwraps` logically guaranteed to be `Ok`? If not, we should probably handle them rather than panic.

components/hab/src/main.rs
```diff
@@ -1490,6 +1501,51 @@ where
...
+                    }
+                    "NetErr" => {
+                        let err = reply.parse::<protocol::net::NetErr>().unwrap();
+                        return Err(SrvClientError::from(err));
+                    }
+                    _ => {
+                        warn!("Unexpected status message, {:?}", reply);
+                    }
+                };
+                Ok(())
```
baumanj: I think it's a little easier to follow to have each `match` arm return the appropriate `Result` value than fall through here.

Cargo.toml
```diff
@@ -29,3 +29,5 @@ members = [
 # [patch."https://github.com/habitat-sh/core.git"]
 # habitat_core = {path = "../core/components/core"}
 # habitat_win_users = {path = "../core/components/win-users"}
+[patch."https://github.com/habitat-sh/core.git"]
+habitat_core = {git ="https://github.com/jamesc/core.git", branch = "jc/launcher_root_path"}
```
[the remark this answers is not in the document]
jamesc: Yep, waiting on habitat-sh/core#88

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -50,13 +50,27 @@ pub fn start(
     execution_strategy: ExecutionStrategy,
     scope: Scope,
     excludes: Vec<PackageIdent>,
+    maybe_services: Option<Vec<PackageIdent>>,
```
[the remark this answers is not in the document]
jamesc: I want to be able to distinguish between no supervisor running (therefore `None`) and a supervisor running but with no loaded services (with vec![]).  This is a by-product of doing the connect to the supervisor in the cli main - I need to pass the info across so I can output to the console after the `Uninstalling ...` line

 I should add a comment.

components/hab/src/main.rs
jamesc: Lemme check.  I copied from the doc example so it might not be bulletproof.

components/hab/src/command/pkg/uninstall.rs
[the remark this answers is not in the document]
baumanj: What's the value of being able to distinguish between no supervisor running and a supervisor running with no services?

components/hab/src/main.rs
jamesc: Looks like they are guaranteed to be Ok.  All our other sup communication code in `main.rs` follows the same pattern of using `unwrap()`.

baumanj: We've been wrong before https://github.com/habitat-sh/habitat/issues/5148 https://github.com/habitat-sh/habitat/issues/5151, so I'm not sure the fact that existing code uses that pattern is totally convincing. If you can point to an earlier point in the code that would handle `Err` values before they get here, that's different.

jamesc: Indeed.  Diving through the marshalling code it looks like it's not handled properly in any of these calls.  Will fix.

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -162,6 +181,15 @@ fn maybe_delete(
         return Ok(false);
     }
 
+    let is_running = services.iter().any(|i| i.satisfies(ident));
+    if is_running {
+        ui.status(
+            Status::Skipping,
+            format!("{}. It is currently running in the supervisor", &ident),
+        )?;
```
christophermaier: A lot of the output around services feels a bit misleading, at the code level, given that we're not actually determining that they're running here.

It feels like the `services` list is a special variant or subset of the `excludes` list, since they're also excluded identifiers... seems like the main difference is how we log when we come across one of them. Is there another difference that is perhaps implicit in the code that I'm not seeing (e.g., the `services` idents are fully-qualified, but the `excludes` can be fuzzy)?

Along those lines, if the service idents _can_ be fuzzy, this would seem to keep any satisfying package, rather than the specific one that is running (or I guess the "latest" package, which should work out to be the same thing).

If service idents are meant to always be fully-qualified, we could tighten this up by verifying that's the case in this code.

It might be worth adjusting the documentation comment to reflect that these service idents are just being treated as a special kind of `excludes`, and that no actual determination of "live-ness" or "service-ness" is being made here.

(Paging @fnichol on this, too, since he was doing some exploratory work a while back on expressing fully-qualified vs. fuzzy identifiers at the type level, which could come in handy here.)

components/hab/src/main.rs
```diff
@@ -1439,13 +1468,17 @@ where
...
+            .map_err(SrvClientError::Decode)?,
         "NetOk" => {
             println!("No services loaded.");
             return Ok(());
         }
         "NetErr" => {
-            let err = reply.parse::<protocol::net::NetErr>()?;
+            let err = reply
+                .parse::<protocol::net::NetErr>()
+                .map_err(SrvClientError::Decode)?;
```
christophermaier: Good call on adding `map_err` to all these. 👍

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -162,6 +181,15 @@ fn maybe_delete(
         return Ok(false);
```
christophermaier: As an aside, we should probably add some logic so can't blow away the Launcher or Supervisor, either.

Perhaps in another PR, though 😄

components/hab/src/command/pkg/uninstall.rs
jamesc: Yeah, this is a little bit about trying to separate concerns between `main.rs` and `uninstall.rs`.  I actually check for running services in `main.rs` as that is where all the sup API code is already.  But I want to output these messages *after* the `"Uninstalling {}"` line. This is the only way I had to do that.

I had an alternative design where `services` was a `Result<Option<Vec<PackageIdent>>>` but as @baumanj pointed out, this code is much cleaner.

jamesc: Yes, services are fully qualified, and we only exclude the running version.  This is so we can use uninstall to clean up old versions of running services. It doesn't have to be the latest even, just the version you've decided to run right now.

I'll update the doc comment to note that the running services are passed in.  My general assumption was that you shouldn't treat the commands as something run differently than the way they're run in `main.rs` right now.

components/hab/src/command/pkg/uninstall.rs
jamesc: If it's not running, they get auto-downloaded on the next `hap sup run` so I was less worried about them as compare to `hab` itself.

Even if we delete it and you then do a restart of hab sup then things should just work but it's probably cleaner if we add them in to the services exclude list.  I'll need to work out how to get the running version of the sup which I don't think I can get right now.

baumanj: Accurately getting the version that's *running* would require a bit more work, but if you want to get the version of whatever `hab sup` executes, that shouldn't be too hard. You may want to crib some of the code for where [the launcher checks the version before launching](https://github.com/habitat-sh/habitat/blob/master/components/launcher/src/server/mod.rs#L534-L562).

Also, generally the version of the supervisor is the same as the `hab` binary.

PR #5836, opened 2018-11-08
Adds LD_LIBRARY_PATH in well known aggregate vars

### LINKED ISSUE #4419: `LD_LIBRARY_PATH` should default to `aggregate` type
It seems that this variable defaults to the `primitive` type, but it seems to make more sense to assume the final value will be aggregated from several paths.

rsertelon:
> Closes #4419

This answers the need of mentioned issue, although @christophermaier wasn't sure if it was a good idea (AFAICT).

Opening this PR with the fix, so we can discuss if needed and merge/close when decided :)

### DIFF
--- a/components/plan-build/bin/environment.bash
+++ b/components/plan-build/bin/environment.bash
@@ -18,6 +18,7 @@ declare -A __buildtime_environment_provenance
 # PRs welcome!)
 declare -A -g __well_known_aggregate_env_vars=(
     # Shell
+    [LD_LIBRARY_PATH]=":"
     [PATH]=":"
 
     # Go


thesentinels:

Thank you for contributing!

christophermaier:
Sorry for the delay @rsertelon ... This is on my list to get to very soon!

smacfarlane:
I think this is behavior we need to be very careful how we support, if at all.  Given a package that contains a `push_runtime_env` or `set_runtime_env` that sets `LD_LIBRARY_PATH`:

```
pkg_name="one"
pkg_origin="$HAB_ORIGIN"
pkg_version="0.0.1"

do_before() {
 push_runtime_env LD_LIBRARY_PATH "/bar"
}
do_build() { :; }
do_install() { :; }
```

and a second package that depends on the first:

```
pkg_name="two"
pkg_origin="$HAB_ORIGIN"
pkg_version="0.0.1"
pkg_deps=($HAB_ORIGIN/one core/coreutils)

do_build() { :; }
do_install() { :; }
```

You will see that LD_LIBRARY_PATH is contained in the `RUNTIME_ENVIRONMENT` of `two`, and is present in the environment of `two` when invoked via `hab pkg exec`. 

```
> build one
> build two
> cat /hab/pkgs/$HAB_ORIGIN/0.0.1/20190304225618/RUNTIME_ENVIRONMENT
LD_LIBRARY_PATH=/bar
> hab pkg exec $HAB_ORIGIN/two env
...
LD_LIBRARY_PATH=/bar
```

All this is expected behavior for `push_runtime_env` and friends.  Where is gets interesting and potentially dangerous is if look at the library load order described in the [ld.so man page](http://man7.org/linux/man-pages/man8/ld.so.8.html)

Specifically, LD_LIBRARY_PATH comes first, after a deprecated DT_RPATH, and before the DT_RUNPATH that we use as part of our isolation guarantees.  This would allow for a package you depend on to create a scenario where your application is linking against libraries you didn't expect it to at runtime, leading to incorrect behavior or segfaults.    

We've seen similar behavior with the [core-plans perl](https://github.com/habitat-sh/core-plans/blob/master/perl/plan.sh#L94) plan.sh. Setting LD_LIBRARY_PATH was causing segfaults during the builds because it was executing a hard-coded `/bin/sh`, which at the time had been built against an older glibc than was being referenced in LD_LIBRARY_PATH.  This was fortunately confined to the build of that software,  but if we had been using `push_runtime_env` instead it could have affected another application at run time. 

All that being said, we currently don't do anything to prevent setting `LD_LIBRARY_PATH` through `push_runtime_env` and related functions, so this change is correct in setting the type of variable that `LD_LIBARARY_PATH` is (an aggregate).

I do think that rather than, or perhaps in addition to, supporting this it would be _safer_ to treat LD_LIBRARY_PATH like we do PATH and add LD_LIBRARY_PATH to `protected` defined [here](https://github.com/habitat-sh/habitat/blob/master/components/plan-build/bin/environment.bash#L53).  What I'm not sure if the usability trade offs are worth the extra safety. I think in many cases they would be, but my experience with core-plans is likely different than our users developing applications.

christophermaier:
Based on @smacfarlane's feedback, we're going to close this out and do some investigation around how to do this generally and safely, in #6263.

Thanks @rsertelon !

PR #5840, opened 2018-11-09
Add abstractions for better dealing with environment variables

christophermaier:
Adds some new testing capabilities and abstractions for more effectively using environment variables throughout the code, as well as making the testing of such code easier.


thesentinels:

Thank you for contributing!

components/sup/src/manager/service_updater.rs
```diff
@@ -288,49 +317,16 @@ struct Worker {
...
-                        DEFAULT_FREQUENCY
-                    );
-                    DEFAULT_FREQUENCY
-                }
-            },
-            Err(_) => DEFAULT_FREQUENCY,
+    fn update_period(&self) -> u64 {
+        let val = ServiceUpdatePeriod::configured_value().0;
+        if val < MIN_ALLOWED_PERIOD {
+            if env::var(PERIOD_BYPASS_CHECK_ENVVAR).is_ok() {
```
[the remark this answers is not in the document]
baumanj: Also, I think this is equivalent, but a bit simpler/more concise to me:
```rust
        if val >= MIN_ALLOWED_PERIOD || env::var(PERIOD_BYPASS_CHECK_ENVVAR).is_ok() {
            val
        } else {
            MIN_ALLOWED_PERIOD
        }
```
or, with the `u64` → `Duration` change, the `MIN_ALLOWED_PERIOD`s would need to be dereferenced.

components/sup/src/manager/periodic.rs
```diff
@@ -25,7 +25,7 @@ pub trait Periodic {
     /// When is the next time we should start a new task, given that
     /// we're going to start one right now?
     fn next_period_start(&self) -> SteadyTime {
-        SteadyTime::now() + TimeDuration::milliseconds(self.update_period())
+        SteadyTime::now() + TimeDuration::milliseconds(self.update_period() as i64)
```
baumanj: I'd suggest making `update_period()` return `Duration`, making this
```suggestion
        SteadyTime::now() + TimeDuration::from_std(self.update_period()).expect("out of range")
```

components/sup/src/manager/service_updater.rs
```diff
@@ -278,6 +282,31 @@ impl ServiceUpdater {
...
 }
 
+/// Represents how far apart checks for updates to individual services
+/// are, in milliseconds.
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Eq)]
+struct ServiceUpdatePeriod(u64);
+
+impl Default for ServiceUpdatePeriod {
+    fn default() -> Self {
+        ServiceUpdatePeriod(MIN_ALLOWED_PERIOD)
```
baumanj:
```suggestion
        ServiceUpdatePeriod(*MIN_ALLOWED_PERIOD)
```
If you make the `u64` → `Duration` change

components/sup/src/manager/service_updater.rs
```diff
@@ -278,6 +282,31 @@ impl ServiceUpdater {
...
+    }
+}
+
+impl FromStr for ServiceUpdatePeriod {
+    type Err = ParseIntError;
+    fn from_str(s: &str) -> result::Result<Self, Self::Err> {
+        let raw = s.parse::<u64>()?;
+        Ok(ServiceUpdatePeriod(raw))
```
baumanj:
```suggestion
        Ok(ServiceUpdatePeriod(Duration::from_millis(raw)))
```

components/sup/src/manager/service_updater.rs
```diff
@@ -288,49 +317,16 @@ struct Worker {
...
-                         to default {} MS frequency.",
-                        val,
-                        FREQUENCY_ENVVAR,
-                        DEFAULT_FREQUENCY
-                    );
-                    DEFAULT_FREQUENCY
-                }
-            },
-            Err(_) => DEFAULT_FREQUENCY,
+    fn update_period(&self) -> u64 {
```
baumanj:
```suggestion
    fn update_period(&self) -> Duration {
```

christophermaier:
@baumanj Great suggestions; I'll look into adding some more type-safety to the service updater code.

Ultimately I think this stuff could live in the core crate. However, I have a few other ideas I'd like to kick around for making type-level distinctions between documented and undocumented environment variables, handling environment variable deprecations (the "frequency" part of those variable names haunts my dreams), and a few other things. Ironing them out here in the code that'll use them seemed like the most efficient way to settle on the right APIs. At that point, I'd be fine moving them to a more general home.

components/sup/src/config.rs
```diff
@@ -39,6 +41,68 @@ pub const DEFAULT_ADDRESS_ENVVAR: &'static str = "HAB_LISTEN_GOSSIP";
...
+            }
+        }
+    }
+
+    /// Overridable function for logging when an environment variable
+    /// value was found and was successfully parsed as a `Self`.
+    ///
+    /// By default, we log a message at the `warn` level.
+    fn log_parsable(env_value: &String) {
+        warn!(
```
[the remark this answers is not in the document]
christophermaier: My thinking was to have it be a more severe message so it would stand out if the variable actually was set. In that case, you'd definitely want to be aware that you're changing something.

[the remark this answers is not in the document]
baumanj: Good point. If the env var is not set, nothing is printed.

components/sup/src/manager/periodic.rs
baumanj: Smart to use the `time` crate's `Duration` rather than stdlib. I like it.

baumanj:
Kicked off a new AppVeyor build since the last on timed out

PR #5842, opened 2018-11-09
fix a compile warning on windows

mwrock:
This just does away with a compiler warning on windows complaining that `real_install_base` was not used because the line using it only compiles on unix. So this switches from the `ifdef` style check to a runtime check.


thesentinels:

Thank you for contributing!

PR #5844, opened 2018-11-13
Update release process

baumanj:
View the two commit separately to see the minor tweaks vs significant reorganization


thesentinels:

Thank you for contributing!

RELEASE.md
```diff
@@ -114,27 +134,18 @@ See https://github.com/habitat-sh/habitat/issues/4656 for further context and id
...
 1. It pulls down the correct studio image
-1. That studio's `hab` is at the correct version
+1. That studio's `hab` is at the correct version (`hab --version`)
 1. A `sup-log` shows a running supervisor (if `sup-log` does not show a supervisor running, run `hab install core/hab-sup --channel release_channel` then `hab sup run`)
-1. Verify that the supervisor is the correct version by running `ls /hab/pkgs/core/hab-sup`
+1. Verify that the supervisor is the correct version (`hab sup --version`)
 
 When testing the linux studio, you will need to `export CI_OVERRIDE_CHANNEL` to the rc channel of the release. So if you are releasing 0.75.2, the channel would be `rc-0.75.2`.
 
-### Addressing issues with a Release
```
mwrock: I still think it merits having something here about what to do when validation fails

baumanj: All that info is still there, just earlier. See https://github.com/habitat-sh/habitat/pull/5844/files#diff-781a53e648f3df8d16a08ec083b04bf4R62

mwrock: yeah I saw that. But its easy to breeze through and forget that when you get to here and actually need it.

baumanj: Funny, I had the opposite problem. I needed to re-tag the last release, but I didn't think to look all the way down at the bottom of the document. I'll add the original section back in and reference it from the above.

mwrock: heh. yeah I guess it all depends on where you get hung up. could probably be either. I like havin something in both places even if one just points to the other. Thanks!

PR #5846, opened 2018-11-13
Use env var at the clap level for ring and ring_key

mpeck:
This change utilizes env var support in clap to simplify the logic a tiny bit when setting `HAB_RING` and `HAB_RING_KEY`. My plan is to move all the env var bits that I can into clap; I'm thinking that that will simplify a lot.

Note that this does change behavior a bit. It adds a new option `--ring-key` to the cli which shouldn't mean much. It also simplifies some of the logic around when to use which value when getting a key pair. Before the option `--ring` took precedence followed by the `HAB_RING_KEY` var and then the `HAB_RING` var, which should technically be the same value as the `--ring` option. Now both `--ring` and `HAB_RING` have the same precedence and are applied before `HAB_RING_KEY`.


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -858,7 +861,8 @@ pub fn sub_sup_run() -> App<'static, 'static> {
         (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with[peer]
             "Watch this file for connecting to the ring"
         )
-        (@arg RING: --ring -r +takes_value "Ring key name")
+        (@arg RING: --ring -r env(RING_ENVVAR) "Ring key name")
+        (@arg RING_KEY: --("ring-key") env(RING_KEY_ENVVAR) "Contents of the ring key")
```
mpeck: Should `RING` and `RING_KEY` be mutually exclusive?

baumanj: Yeah, I think they should. Also, I think it should be made more clear with examples what kind of values `RING` and `RING_KEY` would take on.

christophermaier: The resolution logic implies that they are.

components/hab/src/cli.rs
```diff
@@ -27,6 +27,9 @@ use url::Url;
 use command::studio;
 use feat;
 
+static RING_ENVVAR: &'static str = "HAB_RING";
+static RING_KEY_ENVVAR: &'static str = "HAB_RING_KEY";
```
christophermaier: Looks like the documentation should also be tweaked [here](https://github.com/habitat-sh/habitat/blob/46fc6ad232bad6a20e2a834251e2e3ac224645b8/www/source/partials/docs/_reference-environment-vars.html.md.erb#L24).

components/hab/src/cli.rs
christophermaier: One concern I have is that making an explicit option for this behavior encourages users to put key material directly into a command-line invocation, which may expose secrets.

(I have similar reservations our use of Builder auth token options, FWIW).

I think we could leave this fallback environment variable logic in the underlying resolution logic, without exposing it as a formal CLI option. Testing could be done using the abstractions introduced in #5840 when it merges.

components/hab/src/cli.rs
```diff
@@ -858,7 +861,12 @@ pub fn sub_sup_run() -> App<'static, 'static> {
         (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with[peer]
             "Watch this file for connecting to the ring"
         )
-        (@arg RING: --ring -r +takes_value "Ring key name")
+        (@arg RING: --ring -r env(RING_ENVVAR) conflicts_with("RING_KEY")
+            "The name of the ring used by the Supervisor when running with wire encryption. \
+             (ex: hab sup run --ring myring)")
+        (@arg RING_KEY: --("ring-key") env(RING_KEY_ENVVAR)
+            "The contents of the ring key when running with wire encryption. \
+             (ex: hab sup run --ring-key $(cat /hab/cache/keys/ring-key-file))")
```
baumanj: Have you tested this out? I'm almost certain you'd need quotes around the `$(cat …)` or else the contents would be interpreted by the shell in a way you don't want.

Also, since that example is `bash`-specific, I wonder if @mwrock could supply a Windows-equivalent.

mpeck: Good catch, I missed that. Honestly I'm not really jazzed about the example at all. I think it makes too many assumptions, but I'm not sure what else to show that would be helpful.

baumanj: I think it serves the purpose of making clear that the argument value should be the *contents* of the ring key. Another way to go would be fully explicit:
```
hab sup run --ring-key "SYM-SEC-1
foo-20181113185935

GCrBOW6CCN75LMl0j2V5QqQ6nNzWm6and9hkKBSUFPI="
```

This would also be a good opportunity to put in a validator.

mpeck: I'm gonna make this a hidden option. @christophermaier has some concerns over it's use that I agree with, but I would still like to keep the functionality provided by clap. I think hiding it alleviates both of those issues.

mpeck: It's hidden, but I think it's still useful to add your example if only to help us.

baumanj: I'm at a loss for why we have this option at all. The `--ring` option takes just the name, and we know where to find that file, but `--ring-key` seems like it could just as easily take a path to a file that contains the contents rather than the contents themselves. I imagine that would allay @christophermaier's concerns. And if for whatever reason someone was generating that data on the fly, they could still use something like [process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html):
```bash
$ hab sup run --ring-key <(command-that-generates-key-contents)
```

mpeck: I agree, I was modelling after the existing behavior. It would make more sense to take a path instead of the file contents. I can change the behavior, but that would be a breaking change. While I'm at it the option/variable names seem a bit confusing too. Maybe we should change those to something more descriptive?

Maybe something like: 
`HAB_RING` / `--ring`   to   `HAB_RING_KEY` / `--ring-key`
`HAB_RING_KEY` / `--ring-key`   to   `HAB_RING_KEY_PATH` / `--ring-key-path`

christophermaier:
I'm not very clear on the intended usecase for `--ring-key`. Seems like if you need to do that, you'd run `hab ring key import` beforehand, and then use `hab sup run --ring=$NAME` afterward.

I'd be in favor of deprecating `--ring-key` now, providing that as a deprecation message, and then yanking the option for a subsequent release.

@fnichol @adamhjk Any thoughts on that?

eeyun:
This is a bit of a generalization but I do have _some_ context on these envvars and their purpose. Historically we've maintained these mostly for folks that are building and running containers or folks that are hardcore on the 12-factor pattern. Personally I'd be :-1: on removing that functionality just because it could isolate users in the container/scheduler space which I think is a big :red_circle: :flags:

baumanj:
People are putting actual key contents in env vars for containers, @eeyun? Like:
```
$ export HAB_RING_KEY="SYM-SEC-1
foo-20181113185935

GCrBOW6CCN75LMl0j2V5QqQ6nNzWm6and9hkKBSUFPI="
```
If so, we should keep it, but I just wanted to be sure we're on the same page

mpeck:
Even still, couldn't the same functionality be duplicated by using `hab ring key import` and then setting `HAB_RING`? I'm pretty sure that would be equivalent to what's happening now.

eeyun:
I mean I don't know if we have active users doing it @baumanj but that is the idea - at runtime pull your key out of a secrets management tool and  inject it into a container with an envvar so you dont have to do any extra juggling. With scheduled workloads that are ephemeral (which I think is a user group we are 100% targeting with our kit despite not fully supporting those workflows today) you want folks to have as much flexibility in delivering configuration without obligating a user so ssh into a container etc. It all effectively needs to be able to be injected in without ssh.

christophermaier:
If container usage is the target audience (which sounds legit), let's keep the environment variable and the clap option (since that cleans up the code internally) and let's very clearly document this intended usage (both for our users as well as for us developers in the future).

Unit tests for the win, too 😄

christophermaier:
Also, thanks for that background @eeyun; super helpful 👍

christophermaier:
@mpeck This looks good, but adding some documentation around the intended container use-case for the ring key environment variable would be useful, both for users and developers.

Some of these commits could also be squashed.

After that, we can get this merged! Thanks!

PR #5847, opened 2018-11-13
Makes hooks file names consistent regarding _ or -

### LINKED ISSUE #4986: Hooks should be named consistently, with dashes
Currently our hook names use a mixture of underscores and dashes:

* health_check
* file_updated
* post-run
* post-stop

... which is maybe technically fine, but aesthetically irksome. We should name them all consistently, and according to @reset:

> the dash is correct and the underscore is not

So we should probably make them all use dashes (and probably honor the underscores for backward compatibility, too).

Adding a documentation label to ensure the docs also get updated when we address this.

rsertelon:
> Closes #4986

This is a first implementation. It's not been tested yet. Thus the WIP flag on the PR.

I'm creating the PR sooner rather than later to get feedback :) Still quite new to rust, even if I start to get used to it.

There's a thing that bothers me, about having to recompute this: 

```rust
// this:
let deprecated_template =
            Self::deprecated_file_name().map(|file_name| template_path.as_ref().join(file_name));

        let has_template = std::fs::metadata(&template).is_ok();
// deprecated_template is moved here with `and_then` call.
        let has_deprecated_template = deprecated_template
            .and_then(|t| std::fs::metadata(&t).ok())
            .is_some();

        let template_to_use = if has_deprecated_template && !has_template {
            warn!(
                "Deprecated hook file detected: {}. You should use {} instead.",
                Self::deprecated_file_name().unwrap(),
                Self::file_name()
            );
// HERE is the duplicate:
            Self::deprecated_file_name().map(|file_name| template_path.as_ref().join(file_name))
        } else if has_template {
```

I think next step would be to add tests!

### DIFF
--- a/components/sup/src/manager/service/hooks.rs
+++ b/components/sup/src/manager/service/hooks.rs
@@ -66,30 +66,60 @@ pub trait Hook: fmt::Debug + Sized {
 
     fn file_name() -> &'static str;
 
+    /// Tries to load a hook if a (deprecated) hook file exists.
+    ///
+    /// Returns the hook if template file (deprecated or not) is found
     fn load<C, T>(service_group: &ServiceGroup, concrete_path: C, template_path: T) -> Option<Self>
     where
         C: AsRef<Path>,
         T: AsRef<Path>,
     {
-        let concrete = concrete_path.as_ref().join(Self::file_name());
-        let template = template_path.as_ref().join(Self::file_name());
-        match std::fs::metadata(&template) {
-            Ok(_) => {
-                let pair = match RenderPair::new(concrete, &template) {
-                    Ok(pair) => pair,
-                    Err(err) => {
-                        outputln!(preamble service_group, "Failed to load hook: {}", err);
-                        return None;
-                    }
-                };
-                Some(Self::new(service_group, pair))
-            }
-            Err(_) => {
-                debug!(
-                    "{} not found at {}, not loading",
-                    Self::file_name(),
-                    template.display()
+        let file_name = Self::file_name();
+        let deprecated_file_name = if Self::file_name().contains("-") {
+            Some(Self::file_name().replace("-", "_"))
+        } else {
+            None
+        };
+
+        let concrete = concrete_path.as_ref().join(&file_name);
+        let template = template_path.as_ref().join(&file_name);
+        let deprecated_template = deprecated_file_name
+            .as_ref()
+            .map(|n| template_path.as_ref().join(n));
+
+        let has_template = template.exists();
+        let has_deprecated_template = deprecated_template.as_ref().map_or(false, |t| t.exists());
+
+        let template_to_use = if has_template {
+            if has_deprecated_template {
+                outputln!(preamble service_group,
+                    "Deprecated hook file detected along with expected one. \
+                     You should remove {} and keep only {}.",
+                    deprecated_file_name.unwrap(),
+                    &file_name
                 );
+            }
+            template
+        } else if has_deprecated_template {
+            outputln!(preamble service_group,
+                "Deprecated hook file detected: {}. You should use {} instead.",
+                deprecated_file_name.unwrap(),
+                &file_name
+            );
+            deprecated_template.unwrap()
+        } else {
+            debug!(
+                "{} not found at {}, not loading",
+                &file_name,
+                template.display()
+            );
+            return None;
+        };
+
+        match RenderPair::new(concrete, &template_to_use, Self::file_name()) {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/hooks.rs
```diff
@@ -66,33 +66,64 @@ pub trait Hook: fmt::Debug + Sized {
...
+        template_to_use.and_then(|t| {
+            let pair = match RenderPair::new(concrete, &t) {
+                Ok(pair) => pair,
+                Err(err) => {
+                    outputln!(preamble service_group, "Failed to load hook: {}", err);
+                    return None;
+                }
+            };
+            Some(Self::new(service_group, pair))
+        })
```
baumanj: I think this could be made a bit simpler if instead of making `template_to_use` an `Option<PathBuf>`, it was a `PathBuf`. That can be achieved by changing the code above to:
```rust
        let template_to_use = if has_template {
            if has_deprecated_template {
                …
            }
            template
        } else if has_deprecated_template {
            …
            deprecated_template.unwrap()
        } else {
            …
            return None;
        };
```
In this case the `unwrap` is safe because `has_deprecated_template` implies that `deprecated_template` is a `Some` variant. With that change, this last block becomes simpler:
```rust
        match RenderPair::new(concrete, &template_to_use) {
            Ok(pair) => Some(Self::new(service_group, pair)),
            Err(err) => {
                outputln!(preamble service_group, "Failed to load hook: {}", err);
                return None;
            }
        }
```

components/sup/src/manager/service/hooks.rs
```diff
@@ -66,33 +66,64 @@ pub trait Hook: fmt::Debug + Sized {
 
     fn file_name() -> &'static str;
 
+    fn deprecated_file_name() -> Option<&'static str> {
+        None
+    }
```
baumanj: This approach is good if we want to support underscores only for `file_updated`, `health_check` and `smoke_test`. But I wonder if it would be better UX to temporarily support (and deprecate) all hook names with underscores. That would lead to slightly simpler code because we could just calculate the deprecated name inline in `load` rather than add `deprecated_file_name` to the trait with something like:
```rust
        let deprecated_file_name = if Self::file_name().contains("_") {
            Some(Self::file_name().replace("_", "-"))
        } else {
            None
        };

```

rsertelon: That looks simpler indeed!

rsertelon:
@baumanj thanks a lot for the pointers to more idomatic rust! looks clearer now :) I'll try to add tests when possible.

rsertelon:
So this now works. It is tested indirectly in the tests that I fixed. Fixtures contain deprecated `health_check` hook file, and the tests explicitly use `health-check` to check for compiled content.

I'm now wondering what could be the approach to test these changes directly, if needed.

components/sup/src/manager/service/hooks.rs
```diff
@@ -66,30 +66,60 @@ pub trait Hook: fmt::Debug + Sized {
...
+        let deprecated_template = deprecated_file_name
+            .as_ref()
+            .map(|n| template_path.as_ref().join(n));
+
+        let has_template = template.exists();
+        let has_deprecated_template = deprecated_template.as_ref().map_or(false, |t| t.exists());
+
+        let template_to_use = if has_template {
+            if has_deprecated_template {
+                outputln!(preamble service_group,
```
baumanj: Looks like there's a `rustfmt` issue here that [travis caught](https://travis-ci.org/habitat-sh/habitat/jobs/455237768):
```
internal error: left behind trailing whitespace
  --> stdin:95
   |
95 |                 outputln!(preamble service_group, 
   |                                                  ^
```

components/sup/src/manager/service/hooks.rs
```diff
@@ -66,30 +66,60 @@ pub trait Hook: fmt::Debug + Sized {
...
+                outputln!(preamble service_group, 
+                    "Deprecated hook file detected along with expected one. \
+                     You should remove {} and keep only {}.",
+                    deprecated_file_name.unwrap(),
+                    &file_name
                 );
+            }
+            template
+        } else if has_deprecated_template {
+            outputln!(preamble service_group,
```
baumanj: Same thing here

rsertelon:
Hey @baumanj

My previous implementation was buggy, now it should be fine. I've fixed the `compile()` method that was not implemented properly.

As for the `deprecated_file_name()` method, it's removed. I think it just boils down to code style ;)

PR #5849, opened 2018-11-14
Enable UserConfigWatcher to be shared across threads

christophermaier:
We need to wrap the `states` `HashMap` in an `Arc<Mutex<...>>` in
order to share a `UserConfigWatcher` across threads. This will allow
us to remove watchers from futures running on a threadpool.

It must be a `Mutex` because the `Sender` and `Receiver` members of
`WorkerState` are not themselves shareable across threads, so we need
to ensure exclusive access.

### DIFF
--- a/components/sup/src/manager/user_config_watcher.rs
+++ b/components/sup/src/manager/user_config_watcher.rs
@@ -15,8 +15,11 @@
 use std::collections::HashMap;
 use std::io;
 use std::path::{Path, PathBuf};
-use std::sync::mpsc::{
-    channel, sync_channel, Receiver, SendError, Sender, SyncSender, TryRecvError, TrySendError,
+use std::sync::{
+    mpsc::{
+        channel, sync_channel, Receiver, SendError, Sender, SyncSender, TryRecvError, TrySendError,
+    },
+    Arc, Mutex,
 };
 use std::thread::Builder as ThreadBuilder;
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/user_config_watcher.rs
```diff
@@ -76,21 +79,26 @@ struct WorkerState {
 
 type ServiceName = String;
 pub struct UserConfigWatcher {
-    states: HashMap<ServiceName, WorkerState>,
+    // We use Arc/Mutex here, because this needs to be shareable
+    // across threads so we can remove watchers from futures; the
+    // Sender and Receiver members of WorkerState aren't themselves
+    // shareable.
+    states: Arc<Mutex<HashMap<ServiceName, WorkerState>>>,
```
baumanj: Why not a doc comment?

components/sup/src/manager/user_config_watcher.rs
```diff
@@ -76,21 +79,26 @@ struct WorkerState {
...
+            states: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
     /// Adds a service to the User Config Watcher, thereby starting a watcher thread.
     pub fn add<T: Serviceable>(&mut self, service: &T) -> io::Result<()> {
         // It isn't possible to use the `or_insert_with` function here because it can't have a
         // return value, which we need to return the error from `Worker::run`.
-        if self.states.get(service.name()).is_none() {
+        let mut states = self.states.lock().unwrap();
```
baumanj: Any reason not to use `expect` in case of mutex poisoning here as we do elsewhere?

components/sup/src/manager/user_config_watcher.rs
christophermaier: Only because it's more of an implementation detail for developers, rather than anything describing the fundamental intention of the data itself.

components/sup/src/manager/user_config_watcher.rs
christophermaier: I'll change it to be more consistent with our other usage.

It _does_ appear to be somewhat idiomatic usage in other Rust code I've been reading, though, where it's basically saying "A panic from thread X should propagate to the current thread Y".

`expect` gives better error messages, though.

christophermaier: Force-pushed an amended commit using `expect` instead of `unwrap`.

components/sup/src/manager/user_config_watcher.rs
baumanj: Fair. I don't know if anyone actually reads out documentation compiled from doc comments, but if they did, I figured they might wonder why that type since that's part of the info they'd be looking at.

PR #5851, opened 2018-11-15
Update to modern Tokio usage in Supervisor

christophermaier:
This removes the usage of the deprecated `tokio-core` in the Supervisor Manager code, paving the way for a more full-featured usage of Tokio internally.

`tokio-core` is still used in the `ctl-gateway` portion of the code, but that is self-contained and "off to the side", and can be refactored separately.

Read individual commit messages for further details.


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -838,8 +847,11 @@ impl Manager {
                     self.gossip_latest_service_rumor(&service);
                 }
             }
+
+            // This is really only needed until everything is running
+            // in futures.
             let time_to_wait = ((next_check - time::get_time()).num_milliseconds()).max(100);
-            core.turn(Some(Duration::from_millis(time_to_wait as u64)));
+            thread::sleep(Duration::from_millis(time_to_wait as u64));
```
baumanj: I think this code would be simpler and safer like this:
```rust
            let now = time::get_time();
            if now < next_check {
                let time_to_wait = next_check - now;
                thread::sleep(time_to_wait.to_std().unwrap());
            }
```

As it is, if we're past the time to wait, we still end up sleeping 100 ms. I'm not sure if that's what we want, but the code certainly isn't obvious. If that *is* what we want, I think this conveys it more clearly:

```rust
            let time_to_wait = if now < next_check {
                next_check - now
            } else {
                TimeDuration::milliseconds(100)
            };
            thread::sleep(time_to_wait.to_std().unwrap());
```

christophermaier: Ah, good catch... I think we can probably simplify this to something like your first suggestion. 

It looks like the `max(100)` bit was to allow some minimal amount of time to wait on events. Since our futures are off on their own threadpool now, we can likely dispense with that.

https://docs.rs/tokio-core/0.1.17/tokio_core/reactor/struct.Core.html#method.turn

PR #5857, opened 2018-11-15
Miscellaneous code cleanup

baumanj:
A few small things that make the code simpler/safer. I recommend hiding whitespace changes when viewing this diff.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -1402,21 +1402,20 @@ impl Manager {
 
     // Creates a rumor for the specified service.
     fn gossip_latest_service_rumor(&self, service: &Service) {
-        let mut incarnation = 1;
+        let incarnation = if let Some(rumor) = self
+            .butterfly
+            .service_store
+            .list
+            .read()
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -1402,21 +1402,20 @@ impl Manager {
 
     // Creates a rumor for the specified service.
     fn gossip_latest_service_rumor(&self, service: &Service) {
-        let mut incarnation = 1;
+        let incarnation = if let Some(rumor) = self
```
baumanj: It's a small thing, but making `incarnation` not mutable makes the intention of the code clearer and if there were a change where it wasn't set, that would trigger a compilation error.

The additional scope is unnecessary as the `LockResult` returned by [`read`](http://doc.rust-lang.org/1.30.0/std/sync/struct.RwLock.html#method.read) will be dropped before the call to `insert_service`. See an example in this [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=f58f2b74a929c528e638e4c009e41bb1).

PR #5859, opened 2018-11-16
Restart the election if the leader is dead

baumanj:
We weren't handling the case where the leader health was `Departed` or totally unknown. There will be follow-on work to make sure there aren't similar issues with other places in the code that use `check_health_of_by_id`, as I think the API there is error-prone.

There's a decent bit of refactoring here, so viewing the commits separately (and with whitespace changes hidden) will probably make it easier to understand.

### DIFF
--- a/components/butterfly/src/rumor/election.rs
+++ b/components/butterfly/src/rumor/election.rs
@@ -30,6 +30,24 @@ pub use protocol::newscast::{election::Status as ElectionStatus, Election as Pro
 use protocol::{self, newscast, FromProto};
 use rumor::{Rumor, RumorPayload, RumorType};
 
+// The default implementations leverage ElectionUpdate's Deref -> Election behavior, but this
+// generates a warning. In practice, it's fine:
+// See https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=162d64d74390df6a3a8bb815cd3a2c73
+#[allow(unconditional_recursion)]
+pub trait ElectionRumor {
+    fn member_id(&self) -> &str {
+        self.member_id()
+    }
+
+    fn is_finished(&self) -> bool {
+        self.is_finished()
+    }
+
+    fn term(&self) -> u64 {
+        self.term()
+    }
+}
+
 #[derive(Debug, Clone, Serialize)]
 pub struct Election {
     pub from_id: String,
@@ -88,10 +106,19 @@ impl Election {
     pub fn no_quorum(&mut self) {
         self.status = ElectionStatus::NoQuorum;
     }
+}
 
-    /// Returns true if the election is finished.
-    pub fn is_finished(&self) -> bool {
+impl ElectionRumor for Election {
+    fn member_id(&self) -> &str {
+        &self.member_id
+    }
+
+    fn is_finished(&self) -> bool {
         self.status == ElectionStatus::Finished
+    }
+
+    fn term(&self) -> u64 {
+        self.term
     }
 }
 
@@ -214,6 +241,8 @@ impl ElectionUpdate {
         ElectionUpdate(election)
     }
 }
+
+impl ElectionRumor for ElectionUpdate {}
 
 impl Deref for ElectionUpdate {
     type Target = Election;
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -948,12 +948,16 @@ impl Server {
                             .push((String::from(&service_group[..]), election.term()));
                     }
                 } else if election.is_finished() {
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

bixu:
For a little background, this issue was discovered partly because of stuck Elections when rolling infra at smartB. We periodically exercise failover/disaster-recovery code by tainting and re-provisioning Terraform resources, in this case, hosts running the `core/postgresql` service.

In our initial setup, we simply tainted and re-provisioned. When we noticed the bad behavior we tried adding an `on_destroy` `remote-exec` hook to forcibly depart the Supervisor on the tainted host, but continued to see the stuck Election behavior.

components/butterfly/src/server/mod.rs
```diff
@@ -1231,6 +1252,96 @@ impl<'a> Serialize for ServerProxy<'a> {
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+    use habitat_core::service::ServiceGroup;
+
+    impl Default for Service {
```
christophermaier: Rather than implementing `Default` in a test module, could this just be a helper function that takes a `member_id` and `service_group`?

components/butterfly/src/server/mod.rs
```diff
@@ -1231,6 +1252,96 @@ impl<'a> Serialize for ServerProxy<'a> {
...
+                initialized: Default::default(),
+                pkg: Default::default(),
+                cfg: Default::default(),
+                sys: Default::default(),
+            }
+        }
+    }
+
+    #[test]
+    fn elections_are_restarted_when_leader_health_is_unknown() {
```
christophermaier: Would it be possible / practical to extract the setup for this test and the following one, in order to clarify exactly which parts are changing between them, and to also perhaps facilitate the addition of tests around other leader health states?

components/butterfly/src/server/mod.rs
```diff
@@ -948,12 +948,16 @@ impl Server {
...
+                    let leader_health = member_list
+                        .health_of_by_id(election.member_id())
+                        .unwrap_or_else(|| {
+                            debug!(
+                                "No health information for {}; treating as Departed",
+                                election.member_id()
+                            );
+                            Health::Departed
+                        });
+                    if leader_health >= Health::Confirmed {
```
christophermaier: So nice! 👍

components/butterfly/src/rumor/election.rs
```diff
@@ -30,6 +30,24 @@ pub use protocol::newscast::{election::Status as ElectionStatus, Election as Pro
 use protocol::{self, newscast, FromProto};
 use rumor::{Rumor, RumorPayload, RumorType};
 
+// The default implementations leverage ElectionUpdate's Deref -> Election behavior, but this
+// generates a warning. In practice, it's fine:
+// See https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=162d64d74390df6a3a8bb815cd3a2c73
+#[allow(unconditional_recursion)]
+pub trait ElectionRumor {
```
christophermaier: Since there are only two "subtypes" of election rumors, and you have to override this default implementation for `Election` anyway, how about removing these default implementations altogether in favor of concretely implementing them on each election type?

That eliminates this odd recursion warning, and also diminishes reliance on the `Deref` implementation for `ElectionUpdate` (Given that the docs say that `Deref` should only be implemented for "smart pointers", it feels like it's wrong to have it for `ElectionUpdate`, and by extension, wrong to build further upon it).

It's not really more verbose (you'll still have two implementations, after all), but it will be much more concrete and  leaves less open to potential weirdness in the future.

baumanj: Fair point. I want to get rid of the separate types at some point and have `Election` be an `enum` with `Leader` and `Update` variants, but until then, your suggestion is cleaner and simpler.

christophermaier: That would be a nice future to get to, for sure 👍

components/butterfly/src/server/mod.rs
baumanj: Good call; that's a bit simpler and considerably more evocative

components/butterfly/src/server/mod.rs
baumanj: Yes, I just hadn't gotten there yet. I'll plan to do this is a follow-up PR where I flesh out the testing more. I wanted to get this one out to fix the bug ASAP. I think when I add more tests for `elections_to_restart_impl` the proper refactoring will become more apparent.

bixu:
🎉

PR #5863, opened 2018-11-19
Make the integration tests work with current supervisor

jamesc:
These tests were from before the split between supervisor and launcher so needed a lot of re-plumbing to make work. 

It's probably best to read commit by commit.  The final commit has the changes to the actual tests that were needed to make them work - the rest is for API changes in the supervisor.

Note 
* I deleted the `svc start` tests as they were identical to the `svc load` once they were modified to do a `svc install`
* All the composite tests were skipped and no changes made to them as they should be deleted soon


thesentinels:

Thank you for contributing!

run-bats.sh
```diff
@@ -4,6 +4,12 @@
 # executes the tests in that container, mounting the tests and Habitat
 # binaries as needed.
 
+if [ $# -eq 0 ] ; then
+    TESTS="."
+else
+    TESTS="$*"
+fi
+
```
christophermaier: Nice!

PR #5864, opened 2018-11-20
Election code cleanup

baumanj:
Some general cleanup and simplification motivated by the close reading of the elections code. This is probably best reviewed as the separate commits.

### DIFF
--- a/components/butterfly/src/rumor/election.rs
+++ b/components/butterfly/src/rumor/election.rs
@@ -53,7 +53,13 @@ pub struct Election {
 impl Election {
     /// Create a new election, voting for the given member id, for the given service group, and
     /// with the given suitability.
-    pub fn new<S1>(member_id: S1, service_group: &str, term: u64, suitability: u64) -> Election
+    pub fn new<S1>(
+        member_id: S1,
+        service_group: &str,
+        term: u64,
+        suitability: u64,
+        has_quorum: bool,
+    ) -> Election
     where
         S1: Into<String>,
     {
@@ -63,7 +69,11 @@ impl Election {
             service_group: service_group.into(),
             term: term,
             suitability: suitability,
-            status: ElectionStatus::Running,
+            status: if has_quorum {
+                ElectionStatus::Running
+            } else {
+                ElectionStatus::NoQuorum
+            },
             votes: vec![from_id],
         }
     }
@@ -227,11 +237,12 @@ impl ElectionUpdate {
         service_group: &str,
         term: u64,
         suitability: u64,
+        has_quorum: bool,
     ) -> ElectionUpdate
     where
         S1: Into<String>,
     {
-        let election = Election::new(member_id, service_group, term, suitability);
+        let election = Election::new(member_id, service_group, term, suitability, has_quorum);
         ElectionUpdate(election)
     }
 }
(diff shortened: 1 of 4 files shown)


components/butterfly/src/rumor/election.rs
```diff
@@ -53,7 +53,13 @@ pub struct Election {
 impl Election {
     /// Create a new election, voting for the given member id, for the given service group, and
     /// with the given suitability.
-    pub fn new<S1>(member_id: S1, service_group: &str, term: u64, suitability: u64) -> Election
+    pub fn new<S1>(
+        member_id: S1,
+        service_group: &str,
+        term: u64,
+        suitability: u64,
+        has_quorum: bool,
```
christophermaier: Initially, I was thinking it might be good to implement our "Enum as named boolean" pattern, but I wonder if it might be more readable and evocative to create `Election::with_quorum(....) -> Election` and `Election::without_quorum(...) -> Election` functions instead.

baumanj: Yeah, I thought about the enum trick, but the places where that really enhances readability are when you have calls with literals or several parameters of the same type that could otherwise get mixed up. Since there's only one `bool` parameter here and the only places that call `Election::new` with a literal are in test code, it didn't seem as valuable. The `start_[update_]election` calls need a separate variable to facilitate the logging, so I felt the `has_quorum` name was sufficiently evocative. If we went with an enum, it seems to be a bit *less* ergonomic:

```rust
        if let CheckQuorum::NoQuorum = check_quorum {
            warn!("start_election check_quorum failed: {:?}", e);
        }
```
compared to:
```rust
        if !has_quorum {
            warn!("start_election check_quorum failed: {:?}", e);
        }
```

The `Election::with_quorum` idea is interesting, but I feel like
```rust
    pub fn with_quorum<S1>(
        member_id: S1,
        service_group: &str,
        term: u64,
        suitability: u64,
        has_quorum: bool,
    ) -> Election
    where
        S1: Into<String>,
    {
        let from_id = member_id.into();
        Election {
            member_id: from_id.clone(),
            service_group: service_group.into(),
            term: term,
            suitability: suitability,
            status: ElectionStatus::Running,
            votes: vec![from_id],
        }
    }

    pub fn without_quorum<S1>(
        member_id: S1,
        service_group: &str,
        term: u64,
        suitability: u64,
        has_quorum: bool,
    ) -> Election
    where
        S1: Into<String>,
    {
        let from_id = member_id.into();
        Election {
            member_id: from_id.clone(),
            service_group: service_group.into(),
            term: term,
            suitability: suitability,
            status: ElectionStatus::NoQuorum,
            votes: vec![from_id],
        }
    }
```
seems like a lot of duplicated code relative to:
```rust
    pub fn new<S1>(
        member_id: S1,
        service_group: &str,
        term: u64,
        suitability: u64,
        has_quorum: bool,
    ) -> Election
    where
        S1: Into<String>,
    {
        let from_id = member_id.into();
        Election {
            member_id: from_id.clone(),
            service_group: service_group.into(),
            term: term,
            suitability: suitability,
            status: if has_quorum {
                ElectionStatus::Running
            } else {
                ElectionStatus::NoQuorum
            },
            votes: vec![from_id],
        }
    }
```
And ultimately I think the duplicated code makes it harder to see the important bit which is different.

baumanj:
Will merge this on Monday since I don't believe in making changes right before being away from the office for several days.

PR #5884, opened 2018-11-28
Move env var and default value for listen gossip to the clap layer

mpeck:
This change moves the env var and default value bits for the listen gossip option into the clap layer. This is just the first step, ultimately the goal is to move all of these bits to the edges of the app to simplify as much of the internals as possible.


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -860,9 +860,8 @@ pub fn sub_sup_run() -> App<'static, 'static> {
         // is displayed confusingly as `hab-sup`
         // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
         (usage: "hab sup run [FLAGS] [OPTIONS] [--] [PKG_IDENT_OR_ARTIFACT]")
-        (@arg LISTEN_GOSSIP: --("listen-gossip") +takes_value {valid_socket_addr}
-            "The listen address for the Gossip System Gateway. If not specified, the value will \
-            be taken from the HAB_LISTEN_GOSSIP environment variable if defined. [default: 0.0.0.0:9638]")
+              (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
```
christophermaier: Does this no longer need `+takes_value` if there's a `default_value`?

Also, does the generated help text reflect the fact that this can be set by an environment variable?

components/hab/src/default_values.rs
```diff
@@ -0,0 +1,17 @@
+// This file contains the string defaults values as well as environment variable strings
```
christophermaier: We should have the standard copyright boilerplate at the top.

These comments should also use `//!` instead of `//`, since they're actual module-level documentation.

components/sup/src/main.rs
```diff
@@ -284,6 +284,13 @@ fn get_ring_key(m: &ArgMatches) -> Result<Option<SymKey>> {
     }
 }
 
+fn get_gossip_listen_addr(m: &ArgMatches) -> Result<GossipListenAddr> {
+    match m.value_of("LISTEN_GOSSIP") {
+        Some(addr_str) => GossipListenAddr::from_str(addr_str),
+        None => Err(sup_error!(Error::IPFailed)),
```
christophermaier: Since you've got a default value for `LISTEN_GOSSIP`, it should be impossible to get `None` here, right? If so, we could simplify this to something like 

```rust
GossipListenAddr::from_str(m.value_of("LISTEN_GOSSIP").expect("LISTEN_GOSSIP should always have a value"))
```

components/sup/src/config.rs
```diff
@@ -127,7 +123,7 @@ impl GossipListenAddr {
 impl Default for GossipListenAddr {
     fn default() -> GossipListenAddr {
         GossipListenAddr(socket_addr_env_or_default(
-            DEFAULT_ADDRESS_ENVVAR,
+            GOSSIP_LISTEN_ADDRESS_ENVVAR,
```
christophermaier: Even though this particular `Default` implementation is a bit odd by relying on environment variables (which I suspect we'll want to change in a future refactoring, in order to more clearly delineate where in the code that environment variables' effects can be felt), it would be good to take advantage of the string default that you have created and harmonize that with this. That way we don't have two potentially divergent defaults in the codebase.

(It may be worth investigating transitioning this to `EnvConfig` to see if that helps.)

components/hab/src/cli.rs
mpeck: Yessir `default_value` implies `takes_value(true)` https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.default_value

and yessir, clap handles docs for both the env and default value :)
```
...
--listen-gossip <LISTEN_GOSSIP>        The listen address for the Gossip System Gateway. [env:
                                       HAB_LISTEN_GOSSIP=]  [default: 0.0.0.0:9638]
...
```

components/sup/src/main.rs
mpeck: Yeah, I like that better. I've been trying to stay away from using `expect` and `unwrap` outside of tests, but that does feel better than the match I'm using.

christophermaier: Seeing `unwrap` in production code is often a red flag because it will cause a panic if the thing isn't actually unwrappable. It is OK in some cases, though, when you are certain (because of other code around) that it can always be unwrapped (e.g., you ascertained that `thing.is_some()` returns `true` before you run `thing.unwrap()`. Usually we'll call that out with a comment in our code.

`expect` does the same thing as `unwrap`, but allows you to add a customized message string. More importantly, though, it also gives you the location in code where `expect` was called; due to internal code shenanigans, `unwrap` _will not_ do this, instead giving you the location of `unwrap` itself in the standard library code, which is useless.

Here, you could legitimately use either one because of the guarantees that clap provides, given how we've configured things. I opted for `expect`, though, because it would provide a useful message if at some point we ever removed the default value from the clap configuration, and then started getting unexpected panics.

components/hab/src/cli.rs
```diff
@@ -25,11 +25,11 @@ use protocol;
 use url::Url;
 
 use command::studio;
+use default_values::{
+    GOSSIP_DEFAULT_ADDR, GOSSIP_LISTEN_ADDRESS_ENVVAR, RING_ENVVAR, RING_KEY_ENVVAR,
+};
```
baumanj: I think these values can just live in `cli.rs` and be declared `pub`. Eventually I don't think any other code should need to reference them directly at which point `pub` can be removed.

components/hab/src/default_values.rs
```diff
@@ -0,0 +1,31 @@
...
+//! Eventually this will likely reside in it's own crate and be composed of fully typed
+//! default values. But as a first step we need a spot to consolidate those values and help simplify
+//! some of the logic around them.
+
+// Supervisor
+pub const GOSSIP_DEFAULT_IP: &str = "0.0.0.0";
+pub const GOSSIP_DEFAULT_PORT: u16 = 9638;
+lazy_static! {
+    pub static ref GOSSIP_DEFAULT_ADDR: String =
+        { [GOSSIP_DEFAULT_IP, &GOSSIP_DEFAULT_PORT.to_string()].join(":") };
```
baumanj: I think this is a bit simpler/more idiomatic for rust:
```suggestion
        { format!("{}:{}", GOSSIP_DEFAULT_IP, GOSSIP_DEFAULT_PORT) };
```

baumanj: If you want to go even bigger, I think it would be a good idea to move the entire parsing here instead of doing this _ad hoc_:

```rust
    pub static ref GOSSIP_DEFAULT_ADDR: SocketAddrV4 = {
        SocketAddrV4::new(
            GOSSIP_DEFAULT_IP
                .parse()
                .expect("GOSSIP_DEFAULT_IP can not be parsed."),
            GOSSIP_DEFAULT_PORT,
        )
    };
    pub static ref GOSSIP_DEFAULT_ADDR_STRING: String = { GOSSIP_DEFAULT_ADDR.to_string() };
```

components/hab/src/default_values.rs
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This file contains the string defaults values as well as environment variable strings
+//! for use in the clap layer of the application. This is not the final form for defaults.
+//! Eventually this will likely reside in it's own crate and be composed of fully typed
+//! default values. But as a first step we need a spot to consolidate those values and help simplify
+//! some of the logic around them.
+
+// Supervisor
```
baumanj: I don't think this or the `// Supervisor ENV vars` comments add much. If we get to a point where there are tons of things here and we want to break them into sections, it may be worth reconsidering, but right now I see them as not helpful, and I avoiding unhelpful comments leaves more room for seeing code.

components/sup/src/manager/mod.rs
```diff
@@ -160,7 +160,7 @@ impl Default for ManagerConfig {
             eventsrv_group: None,
             update_url: "".to_string(),
             update_channel: "".to_string(),
-            gossip_listen: GossipListenAddr::default(),
+            gossip_listen: GossipListenAddr::configured_value(),
```
christophermaier: Since the only place in real code that `ManagerConfig::default()` seems to get used is up in the CLI layer, where we're going to inject any configured `GossipListenAddr` value anyway, I think using `default()` here, rather than `configured_value()` is safe and more intention-revealing.

components/sup/src/config.rs
```diff
@@ -190,7 +192,7 @@ mod tests {
         use super::*;
         #[test]
         fn local_addr_for_gossip_listen_addr_works_for_unspecified_address() {
-            let listen_addr = GossipListenAddr::default();
+            let listen_addr = GossipListenAddr::configured_value();
```
christophermaier: Using `default` in tests is OK, and the safer thing to do. I'd only use `configured_value()` if you're explicitly testing that behavior (e.g., that it responds to the right environment variable, etc.) 

By using it here _without_ explicit need to, it creates uncertainty as to what is actually being tested ("Did this test fail because I have a certain environment variable set to something the test wasn't expecting?"). It also makes the test subject to interference from other tests that _do_ rely on the configured environment variable that might be running at the same time (gotta use the `locked_env_var!` macro for that!)

components/sup/src/config.rs
```diff
@@ -199,7 +201,7 @@ mod tests {
 
         #[test]
         fn local_addr_for_gossip_listen_addr_returns_same_ip_for_a_specified_address() {
-            let mut listen_addr = GossipListenAddr::default();
+            let mut listen_addr = GossipListenAddr::configured_value();
```
christophermaier: Ditto.

components/sup/src/manager/service/hooks.rs
```diff
@@ -1327,7 +1328,7 @@ echo "The message is Hola Mundo"
 
         let sys = Sys::new(
             true,
-            GossipListenAddr::default(),
+            GossipListenAddr::configured_value(),
```
christophermaier: Ditto, re: using `default()` in tests.

components/sup/src/manager/service/hooks.rs
```diff
@@ -1441,7 +1442,7 @@ echo "The message is Hello"
...
```
christophermaier: Ditto

components/sup/src/manager/service/mod.rs
```diff
@@ -1134,7 +1135,12 @@ mod tests {
         let socket_addr =
             SocketAddr::from_str("127.0.0.1:1234").expect("Can't parse IP into SocketAddr");
         let http_addr = http_gateway::ListenAddr::default();
-        let sys = Sys::new(false, GossipListenAddr::default(), socket_addr, http_addr);
+        let sys = Sys::new(
+            false,
+            GossipListenAddr::configured_value(),
```
christophermaier: Ditto, re: using `default()` in tests.

components/sup/src/manager/service/mod.rs
```diff
@@ -1125,6 +1125,7 @@ mod tests {
         manager::{sys::Sys, FsCfg},
         ServiceSpec,
     };
+    use config::EnvConfig;
```
christophermaier: Using `default()` instead of `configured_value()` means you can remove this.

components/sup/src/manager/service/hooks.rs
```diff
@@ -1087,6 +1087,7 @@ mod tests {
     use super::fs as supfs;
     use super::*;
     use census::CensusRing;
+    use config::EnvConfig;
```
christophermaier: Using `default()` instead of `configured_value()` means you can remove this.

components/sup/src/config.rs
```diff
@@ -158,7 +155,12 @@ impl FromStr for GossipListenAddr {
             Ok(addr) => Ok(GossipListenAddr(addr)),
             Err(_) => match IpAddr::from_str(val) {
                 Ok(ip) => {
-                    let mut addr = GossipListenAddr::default();
+                    // We shouldn't need to fallback to default here. The value
+                    // should always be set at the clap layer. I'm leaving this
+                    // here for now so as to not break anything unexpected, but
+                    // we should revisit this once we have the default bits all
+                    // moved to the clap layer.
+                    let mut addr = GossipListenAddr::configured_value();
```
christophermaier: Whoa, this is weird... I didn't realize we were using default down here in `FromStr`. Now that we've got this hooked up to `EnvConfig`, it creates a very confusing cyclic dependency (`EnvConfig` is itself implemented in terms of `FromStr`, but _this_ `FromStr` is now implemented in terms of `EnvConfig`!)

I think it's worth tracing this through to see where all this is being called. It's also odd that even though the contract of the `from_str` method is to return a `Result`, because of the `Default` fallback behavior, we're never going to return an error.

It seems like the original implementation wanted to be something like `EnvConfig`, but without being declared as  formally. I think we have the ability to clean that up now.

PR #5887, opened 2018-11-28
Add x86_64-linux-kernel2 validation to release checklist

smacfarlane:
Now that the x86_64-linux-kernel2 packages are built and promoted as part of the release, we also need to validate them as we do with linux,darwin and windows packages. 


thesentinels:

Thank you for contributing!

PR #5902, opened 2018-11-30
Unpin dependencies from all habitat components

smacfarlane:
With the introduction of the `x86_64-linux-kernel2` PackageTarget and release as of 0.69.0, these pins introduced in #5238 should no longer be necessary.     Since the default Plans no longer have pins,  the `x86_64-linux-kernel2` variant plans introduced in #5698 and #5751 are also no longer necessary. 

I have successfully built all plans affected for both `x86_64-linux` and `x86_64-linux-kernel2` to validate, with the exception of eventsrv as we aren't currently shipping that component.

tenor-199091573


thesentinels:

Thank you for contributing!

PR #5906, opened 2018-12-01
disable actix signal handling when the IgnoreSignals feature is enabled

### LINKED ISSUE #5904: Habitat supervisor on windows is not terminating while long running init hooks are running
This was discovered while diving into #5903 and is responsible for why customers get themselves into the #5903 scenario however this issue stands on its own and #5903 covers another problem that could possibly be encountered in other scenarios.

When the Habitat Supervisor receives a ctrl+c signal on Windows, either via a user or programatically in the windows habitat service harness, the expected behavior is that the Supervisor should immediately terminate under the default windows ctrl+c behavior and the launcher which waits for the supervisor to exit, will then properly clean things up.

It appears that the supervisor is not responding to ctrl+c. This would be fine if all services are in their `run` hooks. In that case the service loop is alerted by the launcher that things are shutting down and all comes down clean. However if an init hook is in progress, the service loop is blocked and the supervisor will remain alive and the launcher will keep waiting. The windows habitat service will wait up to a minute for the launcher to exit and if it fails to do so, it will forcefully terminate the process which leaves the services PID files around and results in the unhappy #5903 scenario.

This appears to be a regression between 0.65.0 and 0.66.0.

mwrock:
fixes #5904 

copied from the code comment:

```
            // On Windows the default actix signal handler will create a ctrl+c handler for the
            // process which will disable default windows ctrl+c behavior and allow us to
            // handle via check_for_signal in the supervisor service loop. However, if the 
            // supervsor is in a long running non-run hook, that loop will not get to 
            // check_for_signal in a reasonable amount of time and the supervisor will not 
            // respond to ctrl+c. On Windows, we let the launcher catch ctrl+c and gracefully
            // shut down services. ctrl+c should simply halt the supervisor. The IgnoreSignals
            // feature is always enabled in the Habitat Windows Service which relies on ctrl+c
            // signals to stop the supervisor.
```

### DIFF
--- a/components/sup/src/http_gateway.rs
+++ b/components/sup/src/http_gateway.rs
@@ -168,11 +168,25 @@ impl Server {
                 Err(_) => HTTP_THREAD_COUNT,
             };
 
-            let bind = server::new(move || {
+            let mut workers = server::new(move || {
                 let app_state = AppState::new(gateway_state.clone());
                 App::with_state(app_state).configure(routes)
-            }).workers(thread_count)
-            .bind(listen_addr.to_string());
+            }).workers(thread_count);
+
+            // On Windows the default actix signal handler will create a ctrl+c handler for the
+            // process which will disable default windows ctrl+c behavior and allow us to
+            // handle via check_for_signal in the supervisor service loop. However, if the
+            // supervisor is in a long running non-run hook, that loop will not get to
+            // check_for_signal in a reasonable amount of time and the supervisor will not
+            // respond to ctrl+c. On Windows, we let the launcher catch ctrl+c and gracefully
+            // shut down services. ctrl+c should simply halt the supervisor. The IgnoreSignals
+            // feature is always enabled in the Habitat Windows Service which relies on ctrl+c
+            // signals to stop the supervisor.
+            if feat::is_enabled(feat::IgnoreSignals) {
+                workers = workers.disable_signals();
+            }
+
+            let bind = workers.bind(listen_addr.to_string());
 
             // We need to create this scope on purpose here because if we don't, the lock never
             // releases, and the supervisor will wait forever on cvar. Creating this artifical


thesentinels:

Thank you for contributing!

mwrock:
Note there is related behavior in https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/mod.rs#L744

components/sup/src/http_gateway.rs
```diff
@@ -168,11 +168,25 @@ impl Server {
...
                 let app_state = AppState::new(gateway_state.clone());
                 App::with_state(app_state).configure(routes)
-            }).workers(thread_count)
-            .bind(listen_addr.to_string());
+            }).workers(thread_count);
+
+            // On Windows the default actix signal handler will create a ctrl+c handler for the
+            // process which will disable default windows ctrl+c behavior and allow us to
+            // handle via check_for_signal in the supervisor service loop. However, if the 
+            // supervsor is in a long running non-run hook, that loop will not get to
```
stevendanna: [nit] spelling: supervisor

components/sup/src/http_gateway.rs
```diff
@@ -168,11 +168,25 @@ impl Server {
...
+            // handle via check_for_signal in the supervisor service loop. However, if the 
+            // supervsor is in a long running non-run hook, that loop will not get to 
+            // check_for_signal in a reasonable amount of time and the supervisor will not 
+            // respond to ctrl+c. On Windows, we let the launcher catch ctrl+c and gracefully
+            // shut down services. ctrl+c should simply halt the supervisor. The IgnoreSignals
+            // feature is always enabled in the Habitat Windows Service which relies on ctrl+c
+            // signals to stop the supervisor.
+            if feat::is_enabled(feat::IgnoreSignals) {
+                workers = workers.disable_signals();
+            }
```
christophermaier: Is there ever a time where we would _not_ want to disable this signal handling behavior on Windows?

I'm just wondering if this kind of logic might better live in an `if cfg!(windows) {...}` block, instead.

Also, is this kind of conditional logic necessary only because long-running hooks can still block the main Supervisor thread? If so, we should file an issue to remove it once all the hooks are running on their own threads.

mwrock: Yes. Ideally we would allow ourselves to handle the signals. That would allow us to more safely clean up after ourselves. And under many scenarios, handling the signals ourdelves works well. It mainly comes down to the long running `init` hook scenarios where things can go wrong. I will file an issue to address their removal on the happy days when we handle the other hooks in their own threads.

PR #5913, opened 2018-12-03
Add testing exclusions for Appveyor

tas50:
Habitat builds are taking up all the available Appveyor slots for Chef
Inc. This adds a similar change to Habitat that we've added to our Chef
projects to conserve available slots. Otherwise we're going to need to
move Habitat to its own account.

Signed-off-by: Tim Smith <tsmith@chef.io>

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,20 @@ cache:
   - c:\cargo\git
   - c:\projects\habitat\target
 
+skip_commits:
+  # version bumps by Expeditor happen as a separate commit after the merge, we can skip
+  author: Chef Expeditor
+  # if ONLY the files listed below are changed in a commit, skip
+  files:
+    - BUILDING.MD
+    - MAINTAINERS.md
+    - CHANGELOG.md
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -6,6 +6,20 @@ cache:
   - c:\cargo\git
   - c:\projects\habitat\target
 
+skip_commits:
+  # version bumps by Expeditor happen as a separate commit after the merge, we can skip
+  author: Chef Expeditor
```
christophermaier: Unfortunately, releases are currently triggered on commits from Expeditor.

We tried this earlier, but it ended up killing our ability to do a release 😱 Guess we didn't get the proper fix in place :frown:

On the upside, I think we've only got one more release to do from Appveyor before we've got everything moved over to Buildkite.

christophermaier:
Actually, we'll get this merged after the currently scheduled release (either today or tomorrow) goes out. That should work out well for everyone involved.

@tas50 Sorry for the hassle!

PR #5921, opened 2018-12-05
fork term crate to get windows memory leak fix

mwrock:
The `term` crate creates handles to stdout on windows but the latest released crate does not close them when they go out of scope. There is a merged PR in its master branch that implements drops and plugs this leak.

Unless things change, we should probably migrate our UI code to use the `termcolor` crate instead. It looks like the `term` crate is largely unmaintained right now. See https://github.com/Stebalien/term/issues/93:

```
So, I thought I'd have time to maintain this but then I got a job (where a large portion of said job is maintaining a bunch of OSS projects...). Unfortunately, that leaves me little time (or motivation) to maintain this project.
```


thesentinels:

Thank you for contributing!

christophermaier:
Looks like https://github.com/Stebalien/term/issues/63 is the relevant issue.

Izian4321:
hi

> #

PR #5928, opened 2018-12-06
Extract unnecessary code from manager module

christophermaier:
The `manager` module, as well as the `Manager` struct itself, has
grown bloated, making it difficult to work with. A closer reading of
the code revealed lots of code that was not directly related to
managing the Supervisor's "main loop"; this code was moved to more
appropriate places.

In particular:

- All functions that implement the various `hab sup` commands were
  moved into a new `commands` module. All of these were static methods
  on the `Manager` struct, making it easy to completely remove them.

- A few accessory structs (`ProcessStatus` and `ServiceStatus`) were
  moved into this `commands` module, since they are only needed for
  the command functions.

- A number of accessory functions for dealing with specs that were
  only used in the commands were also extracted into `commands`.

- A few trait implementations for `ServiceBind` and `DesiredState`
  that existed in the `manager` module, instead of the `spec` module
  where the structs themselves are defined, were moved.

- The `clean_dirty_state` and `clean_state_path_dirs` functions were
  slightly reworked to pull information from `FsCfg`, which provided the
  same information as some static path functions, allowing the deletion
  of the latter.

  It should be noted that there remains a bit of duplication around
  the creation of these filesystem paths (see `FsCfg` and the
  `commands` module). Further exploration reveals that our filesystem
  path logic is actually split up across several modules across the
  project. Consolidation of all that should be a follow-on PR.

### DIFF
--- a/components/sup/src/ctl_gateway/server.rs
+++ b/components/sup/src/ctl_gateway/server.rs
@@ -44,7 +44,7 @@ use tokio_codec::Framed;
 use tokio_core::reactor;
 
 use super::{CtlRequest, REQ_TIMEOUT};
-use manager::{Manager, ManagerState};
+use manager::{commands, ManagerState};
 
 /// Sending half of an mpsc unbounded channel used for sending replies for a transactional message
 /// from the main thread back to the CtlGateway. This half is stored in a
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

PR #5940, opened 2018-12-07
Add --transitive option to hab pkg dependencies

jamesc:
After UX review it was suggested that we should either show direct
dependencies or all recursive dependencies with a -t/--transitive command

NOTE: There is a breaking change in how --transitive works. The default (with no options) used to show transitive dependencies, now it will only show direct dependencies. This was done after walking through the UX with @mpeck and considering this would be the most understandable behaviour for the user.

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -469,9 +469,10 @@ fn sub_pkg_binds(m: &ArgMatches) -> Result<()> {
 
 fn sub_pkg_dependencies(m: &ArgMatches) -> Result<()> {
     let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?;
-    let scope = match m.is_present("TRANSITIVE") {
-        true => command::pkg::Scope::PackageAndDependencies,
-        false => command::pkg::Scope::Package,
+    let scope = if m.is_present("TRANSITIVE") {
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/main.rs
```diff
@@ -469,7 +469,12 @@ fn sub_pkg_binds(m: &ArgMatches) -> Result<()> {
 
 fn sub_pkg_dependencies(m: &ArgMatches) -> Result<()> {
     let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?;
-    command::pkg::dependencies::start(&ident, &*FS_ROOT)
+    let scope = match m.is_present("TRANSITIVE") {
+        true => command::pkg::Scope::PackageAndDependencies,
+        false => command::pkg::Scope::Package,
+    };
```
baumanj: Why not
```rust
    let scope = if m.is_present("TRANSITIVE") {
        command::pkg::Scope::PackageAndDependencies
    } else {
        command::pkg::Scope::Package,
    };
```

baumanj: I go back and forth as well. Also about where to use `if let` vs `match`. My rule of thumb so far has been `match` whenever the exhaustiveness checking is likely to catch a logic error or will result in simpler code.

PR #5942, opened 2018-12-07
Update CLI docs for 0.70.0 release

eeyun:

### DIFF
--- a/www/source/partials/docs/_reference-template-data.html.md.erb
+++ b/www/source/partials/docs/_reference-template-data.html.md.erb
@@ -6,21 +6,7 @@ These configuration settings are referenced using the [Handlebars.js](https://gi
 
 ## sys
 
-Describes the details of how this specific Supervisor was started
 
-| Property | Type | Description |
-| -------- | ---- | ----------- |
-| version | string | Version of the Habitat Supervisor, e.g., `0.54.0/20180221023448` |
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

www/source/partials/docs/_reference-template-data.html.md.erb
```diff
@@ -6,21 +6,7 @@ These configuration settings are referenced using the [Handlebars.js](https://gi
 
 ## sys
 
-Describes the details of how this specific Supervisor was started
```
mwrock: seems strange that this block is removed givving us an empty heading

eeyun: That... is very strange. This was generated with the tooling so i wonder if that heading should be removed completely now or something?

christophermaier: Ugh... this has happened for the past few releases... CN was going to look into it, but I guess he wasn't able to before he left.

The issue is that this data is defined in a JSON schema file that is included in several places. Our tooling to automatically generate this documentation from the JSON schema was written back before this was an included file, and thus it does not understand how to deal with included files.

I'll file an issue for this. In the meantime, this whole deletion can just be dropped from this PR (none of this data has changed for this release). @eeyun Could you also make a note of this in the release documentation,  pending the proper solution of this issue?

christophermaier: Issue filed: #5948

PR #5950, opened 2018-12-10
ListenCtlAddr type

mpeck:
Creates a new type `ListenCtlAddr` in the common crate and uses it to define defaults and env var strings in the clap layer. This also moves the `EnvConfig` trait to the common crate.


thesentinels:

Thank you for contributing!

baumanj:
Looks like there are some build errors. Let me know if you need any help sorting them out, but I'll hold off on reviewing until they're resolved.

mpeck:
@baumanj yeah, I missed a couple bits. It should be cleared up now though.

christophermaier:
Appveyor timed out, so I restarted it.

components/hab/src/main.rs
```diff
@@ -869,7 +870,7 @@ fn sub_svc_set(m: &ArgMatches) -> Result<()> {
             .unwrap_or("UNKNOWN".to_string()),
     ))?;
     ui.status(Status::Creating, format!("service configuration"))?;
-    SrvClient::connect(&sup_addr, &secret_key)
+    SrvClient::connect(&sup_addr.into(), &secret_key)
```
christophermaier: This makes me want to update `SrvClient::connect` to take an `AsRef<SocketAddr>` instead of an `&SocketAddr`, and then implement that trait for `ListenCtlAddr`. 

That would better than `Into<SocketAddr>` given [this API guideline](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#c-conv-traits)

components/hab/src/main.rs
```diff
@@ -1651,23 +1652,23 @@ fn get_strategy_from_input(m: &ArgMatches) -> Option<UpdateStrategy> {
         .and_then(|f| UpdateStrategy::from_str(f).ok())
 }
 
-fn sup_addr_from_input(m: &ArgMatches) -> Result<SocketAddr> {
+fn sup_addr_from_input(m: &ArgMatches) -> Result<ListenCtlAddr> {
```
christophermaier: Might be good to change this to `listen_addr_from_input`

components/hab/src/main.rs
```diff
@@ -1651,23 +1652,23 @@ fn get_strategy_from_input(m: &ArgMatches) -> Option<UpdateStrategy> {
...
     match m.value_of("REMOTE_SUP") {
         Some(rs) => {
             let sup_addr = if rs.find(':').is_some() {
                 rs.to_string()
             } else {
-                format!("{}:{}", rs, protocol::ctl::DEFAULT_PORT)
+                format!("{}:{}", rs, ListenCtlAddr::default().port())
```
christophermaier: I think we should pull all this logic into the `FromStr` implementation for `ListenCtlAddr`, since that is essentially what it is doing.

Going back to my previous suggestion of embedding the default port into the type itself... I think we can use that for this usecase to avoid creating a new value, just to grab the port from it.

components/sup/src/main.rs
```diff
@@ -222,33 +218,32 @@ fn mgrcfg_from_matches(m: &ArgMatches) -> Result<ManagerConfig> {
...
+        eventsrv_group: m.value_of("EVENTS").and_then(|events| events.parse().ok()),
+        tls_files: m.value_of("KEY_FILE").and_then(|kf| {
+            Some((
+                PathBuf::from(kf),
+                PathBuf::from(
+                    m.value_of("CERT_FILE")
+                        .expect("CERT_FILE should always have a value if KEY_FILE has a value."),
+                ),
+            ))
+        }),
```
christophermaier: As an aside, I think the only thing we still need the `..Default::default()` bit for is for `custom_state_path`, which is something I'd eventually like to remove... it appears to only exist for manipulating testing data.

Might be good to add a note to that effect here.

components/common/src/listen_ctl_addr.rs
```diff
@@ -53,20 +51,24 @@ impl FromStr for ListenCtlAddr {
...
+        ListenCtlAddr(socket_addr)
     }
 }
 
-impl fmt::Display for ListenCtlAddr {
-    fn fmt(&self, f: &mut fmt::Formatter) -> result::Result<(), fmt::Error> {
-        write!(f, "{}", self.0)
+impl Into<SocketAddr> for ListenCtlAddr {
+    fn into(self) -> SocketAddr {
+        self.0
```
christophermaier: Based on [this API guideline](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#c-conv-traits) we probably shouldn't implement this. 

(sorry, I know that's the opposite of what I suggested yesterday... I am full of fail :frowning:)

Reading [further](https://rust-lang-nursery.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), if we really need this functionality, we should probably have an `into_inner()` method that returns `self.0` instead.

components/common/src/listen_ctl_addr.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use std::fmt;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::result;
+use std::str::FromStr;
+
+use error::{Error, Result};
+use EnvConfig;
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
baumanj: I don't think it's a good idea to make `ListenCtlAddr` implement `Copy`. In general, copy types should be very small and not contain heap-allocated data which may have more significant performance implications. Since `SocketAddr` doesn't implement `Copy`, that's probably a good hint that a type wrapping it shouldn't either. This has the benefit of forcing the code that uses it to either explicitly call `clone`, which makes the performance issues more apparent and easier to track down later, or figuring out how to make appropriate use of references.

I'm assuming you derived copy here, because without it, you get errors like this:
```
error[E0382]: use of moved value: `sup_addr`                                                                                                                                                                
   --> components/hab/src/main.rs:895:56                                                                                                                                                                    
    |                                                                                                                                                                                                       
873 |     SrvClient::connect(&sup_addr.into(), &secret_key)                                                                                                                                                 
    |                         -------- value moved here                                                                                                                                                     
...                                                                                                                                                                                                         
895 |     ui.status(Status::Applying, format!("via peer {}", sup_addr))?;                                                                                                                                   
    |                                                        ^^^^^^^^ value used here after move                                                                                                            
    |                                                                                                                                                                                                       
    = note: move occurs because `sup_addr` has type `common::types::ListenCtlAddr`, which does not implement the `Copy` trait                                                                               
```
What's happening is that previously we could use the same `sup_addr` **reference** in two places because  `SrvClient::connect` takes a `&SocketAddr` parameter. However, now that we have wrapped `SocketAddr` in a `ListenCtlAddr`, we must call `into` to convert to a `SocketAddr` (and then take a reference), but because `Into<SocketAddr> for ListenCtlAddr`'s `into` method takes ownership of `self`, you get the above error.

Also note that, it's generally not necessary to implement `Into` since implementing `From` gives `Into` automatically:
> Library authors should not directly implement this trait, but should prefer implementing the From trait, which offers greater flexibility and provides an equivalent Into implementation for free, thanks to a blanket implementation in the standard library.

See [the docs](http://doc.rust-lang.org/1.31.0/std/convert/trait.Into.html#method.into) for more.

However, in this case, what we want is something else: we want to be able to use a `ListenCtlAddr` where we could previously provide a `&SocketAddr`. Fortunately, there's a type for just that purpose: [`AsRef`](http://doc.rust-lang.org/1.31.0/core/convert/trait.AsRef.html). If you add this implementation:
```rust
impl AsRef<SocketAddr> for ListenCtlAddr {
    fn as_ref(&self) -> &SocketAddr {
        &self.0
    }
}
```
Then you can resolve the above error with:
```rust
    SrvClient::connect(sup_addr.as_ref(), &secret_key)
```
However, you may instead want to change the interface of `SrvClient::connect` to accept a `&ListenCtlAddr` instead of a `&SocketAddr`. After all, one of the points of this wrapping type is enforcing correctness through the compiler, right?

In that case, the call to `SrvClient::connect` stays as it was:
```rust
    SrvClient::connect(&sup_addr, &secret_key)
```
but the code *in* `connect` that needs a `&SocketAddr` specifically (because it's a library external to our project) makes use of our `AsRef` implementation:
```rust
        let conn = TcpStream::connect(addr.as_ref())
```

components/common/src/listen_ctl_addr.rs
baumanj: I think `AsRef<SockAddr>` is what we want. See https://github.com/habitat-sh/habitat/pull/5950/files#r240687807.

components/hab/src/main.rs
```diff
@@ -1651,23 +1652,23 @@ fn get_strategy_from_input(m: &ArgMatches) -> Option<UpdateStrategy> {
...
             };
             let addrs: Vec<SocketAddr> = match sup_addr.to_socket_addrs() {
                 Ok(addrs) => addrs.collect(),
                 Err(e) => {
                     return Err(Error::RemoteSupResolutionError(sup_addr, e));
                 }
             };
-            Ok(addrs[0])
+            Ok(ListenCtlAddr::from(addrs[0]))
```
baumanj: I realize the `addrs[0]` was already here, but as long as we're changing code in this area, I think it's worth fixing up. In general, [`Index`](http://doc.rust-lang.org/1.31.0/std/ops/trait.Index.html) operations are unsafe. In this case, it's kind of OK because the CLI validation has already run, but we can structure the code better and make this more explicit:
```rust
            match sup_addr.to_socket_addrs() {
                Ok(addrs) => match addrs.collect::<Vec<_>>().as_slice() {
                    [addr] => return Ok(ListenCtlAddr::from(*addr)),
                    _ => panic!(
                        "CLI validation failure: REMOTE_SUP should contain single SocketAddr"
                    ),
                },
                Err(e) => {
                    return Err(Error::RemoteSupResolutionError(sup_addr, e));
                }
            }
```
Also, the overall structure of the function could be improved. I'd suggest:
```rust
    m.value_of("REMOTE_SUP")
        .map_or(Ok(ListenCtlAddr::default()), |rs| {
            // parse string → ListenCtlAddr
        })
```

components/sup/src/main.rs
```diff
@@ -221,33 +218,32 @@ fn mgrcfg_from_matches(m: &ArgMatches) -> Result<ManagerConfig> {
...
+            .parse()?,
+        http_listen: m
+            .value_of("LISTEN_HTTP")
+            .expect("LISTEN_HTTP should always have a value.")
+            .parse()?,
+        ctl_listen: m
+            .value_of("LISTEN_CTL")
+            .expect("LISTEN_CTL should always have a value.")
+            .parse()?,
+        watch_peer_file: m.value_of("PEER_WATCH_FILE").map(|file| file.to_string()),
```
baumanj:
```suggestion
        watch_peer_file: m.value_of("PEER_WATCH_FILE").map(str::to_string),
```

components/common/src/listen_ctl_addr.rs
christophermaier: Agree with not deriving `Copy` here... that snuck by me initially 👍 

(Also agree with all the rest, too)

components/hab/src/main.rs
christophermaier: tenor-82425110

components/sup/src/manager/mod.rs
```diff
@@ -47,6 +47,9 @@ use butterfly;
 use butterfly::member::Member;
 use butterfly::server::{timing::Timing, ServerProxy, Suitability};
 use butterfly::trace::Trace;
+use common::command::package::install::InstallSource;
```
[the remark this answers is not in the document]
mpeck: I think this was a merge issue.

components/common/src/lib.rs
```diff
@@ -38,9 +38,14 @@ extern crate winapi;
...
+pub mod cli_defaults;
 pub mod command;
-pub mod defaults;
 pub mod error;
 pub mod locked_env_var;
 pub mod package_graph;
 pub mod ui;
+
+mod listen_ctl_addr;
+pub use listen_ctl_addr::ListenCtlAddr;
```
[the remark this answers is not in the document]
christophermaier: I'm actually interested in making better and more principled use of `pub use` in the code base, particularly in parts of the manager code. And while, yes, several different types can all go into a single module, I often find this to be more troublesome to use, particularly when you get a lot of types, or the types have a lot of associated code and trait implementations ("where in this 1000 line module is the one type I care about?"). In those cases, it can actually be more useful (from an author's perspective, and in my opinion) to have a single module dedicated to all the code (and tests!) for that type (or constellation of highly related types). Then, through judicious use of `use`, `pub use`, `mod`, and `pub mod` you can create an API that's easier from a user's perspective (because having a bunch of `use listen_ctl_addr::ListenCtlAddr;` lines is admittedly pretty rough).

In this particular case, I agree that gathering things together under a higher-level `types` namespace could provide some better indication of sourcing and logical grouping; `use common::types::ListenCtlAddr;` seems very readable and obvious to me. However, that doesn't necessarily mean that it'd be better to place everything into a single `types.rs`. 

As with all things, moderation is key here, but I think it could be a pretty powerful tool, and could drive some more principled visibility decisions.

components/sup/src/main.rs
```diff
@@ -48,7 +48,9 @@ use std::{
 
 use clap::ArgMatches;
 use common::command::package::install::InstallSource;
+use common::defaults::GOSSIP_DEFAULT_PORT;
```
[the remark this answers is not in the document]
mpeck: I agree. I'm gonna tweak `GossipListenAddr` to work more like `ListenCtlAddr` in a follow up PR; I'll move the const bit when I do that.

components/common/src/lib.rs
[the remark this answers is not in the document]
mpeck: Is there some standard practice around this? I agree with @christophermaier here, but I'm happy to go with whatever pattern we decide to use.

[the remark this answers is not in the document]
baumanj: My priorities are:
1. Don't fall prey to analysis paralysis and spend too much time on this; there's no perfect solution, but time, experience and refactoring will get us closer
1. Prefer putting things that are logically related in the same module over keeping files small
1. Avoid redundancy like `use foo::Foo;` (see "[Name Things Once](https://www.youtube.com/watch?v=1__lNTlj1_w)")
1. Avoid confusing statements like `use Foo;` instead of `use bar::Foo` caused by `pub use`
1. Use `pub use` so consumers can write `use cratename::Type;` instead of `use cratename::internal::implementation::specific::details::Type;`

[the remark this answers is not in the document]
christophermaier: We've not really had any principled discussions around this, so this is the beginning. For sure don't spend a lot of time sweating over this; time will show us what works best.

@baumanj's guidelines are a great start, though.

components/hab/src/main.rs
mpeck: I can't pattern match on addrs here because the vector can contain more than one item and still be valid. I tweaked things a bit to implement your other suggestions though.

mpeck: Sorry @baumanj, I should have left more explanation. We only allow for a single value for `--remote-sup` but we don't require that value to be an IP and thus aren't guaranteed that `to_socket_addrs` will return only a single value. If the user passes a name then `to_socket_addrs` can resolve it to both an IPv4 and IPv6 returning a vector with 2 values.

mpeck: Also, I did add a check to make sure that the vector isn't empty. I think there is still more refactoring that could be done to this function, but that should at least make this operation safer.

baumanj: Ah, I follow you now. Thanks for the clarification.

If we were further in the future, you could use this [advanced slice pattern](https://doc.rust-lang.org/1.9.0/book/slice-patterns.html) syntax:
```rust
    match addrs_slice {
        [primary, _secondaries..] => primary,
        _ => …,
    }
```
But it's not available yet, so I'd suggest this approach:
```rust
    match listen_ctl_addr.to_socket_addrs().and_then(|mut addrs| {
        addrs.next().ok_or(err))
    }) {
        Ok(addr) => addr,
        Err(e) => …,
    }
```

I've left out the details of the error handling, but you'll probably want some sort of [`io::Error`](http://doc.rust-lang.org/1.31.0/std/io/struct.Error.html#method.new) (probably [`ErrorKind::AddrNotAvailable`](http://doc.rust-lang.org/1.31.0/std/io/enum.ErrorKind.html#variant.AddrNotAvailable)) in the case that the iterator returns no values.

Additionally, this definitely merits a comment about why `to_socket_addrs` could return multiple values and why we're only paying attention to the first one.

components/hab/src/main.rs
```diff
@@ -1562,11 +1563,11 @@ fn supervisor_services() -> Result<Vec<PackageIdent>> {
 
     let cfg = config::load()?;
     let secret_key = ctl_secret_key(&cfg)?;
-    let sup_addr = protocol::ctl::default_addr();
```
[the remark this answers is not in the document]
mpeck: Is this what you mean?

https://github.com/habitat-sh/habitat/pull/5950/files/ba52b88b73365f1c6b85f780e14df8d6dc3c5455#diff-1c0b53ba7e532ece4ea2b109af50188bL34

components/sup/src/main.rs
```diff
@@ -223,6 +223,40 @@ fn mgrcfg_from_sup_run_matches(m: &ArgMatches) -> Result<ManagerConfig> {
...
+            || {
+                let default = sup::http_gateway::ListenAddr::default();
+                println!(
+                    "Value for LISTEN_HTTP has not been set. Using default: {}",
+                    default
+                );
+                Ok(default)
+            },
+            |v| v.parse(),
+        )?,
```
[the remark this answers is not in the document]
mpeck: I changed them to `error!`. I'm happy to remove them if preferred though.

Issue #4759, opened 2018-03-15
After stopping the sup leader I don't see a new election taking place


epierotto:
Hi, I'm doing some testing on how leader election will trigger hooks and reconfigure my app.

I'm using a docker compose file that runs 3 containers of my app.
- On the first start I can see the leader election happening on the logs and the app is configured and started properly
- Once the app is running I stop the leader container and I can see that there new election never happens
- Even if I start the stopped container there is no new leader election, since that my config is relying on `{{#if svc.me.leader ~}}` to configure some stuff that never happens.

I realise that this also happens running this [example of posgresql](https://github.com/habitat-sh/core-plans/tree/master/postgresql), a new leader election is never triggered.

```yml
version: '3'

services:
  pg1:
    image: habitat/postgresql
    command: --group cluster
      --topology leader
    volumes:
      - pg1-data:/hab/svc/postgresql/data
  pg2:
    image: habitat/postgresql
    command: --group cluster
      --topology leader
      --peer pg1
    volumes:
      - pg2-data:/hab/svc/postgresql/data
  pg3:
    image: habitat/postgresql
    command: --group cluster
      --topology leader
      --peer pg1
    volumes:
      - pg3-data:/hab/svc/postgresql/data

volumes:
  pg1-data:
  pg2-data:
  pg3-data:
```
So the steps to reproduce are:
- Start the docker-compose example
- Stop the leader
- Restart the stopped container to re-join

Tested on both OS with same result

<span data-href="#aha_task_list_start" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/>

- [x] The OS (including version) where you are running any of the Habitat commands.
  - CentOS Linux release 7.3.1611 (Core) 
    - Docker version 17.12.1-ce, build 7390fc6
    - docker-compose version 1.19.0, build 9e633ef
    - hab 0.54.0/20180221022026 (My package is running this version)
  - Mac OS X 10.13.3 BuildVersion: 17D47
    - Docker version 17.12.0-ce, build c97c6d6
    - docker-compose version 1.18.0, build 8dd22a9
<span data-href="#aha_task_list_end" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/>

baumanj:
From the [Leader Election](https://www.habitat.sh/docs/internals/#leader-election) section in the docs:
> When a service group starts in a leader topology, it will wait until there are sufficient members to form a quorum (at least three). At this point, an election cycle can happen

[...]

PR #5965, opened 2018-12-11
Remove minimum 3 member requirement from quorum check

baumanj:
Resolves https://github.com/habitat-sh/habitat/issues/4759

The purpose of quorum is to prevent multiple elections occurring on different sides of a network partition leading to split-brain. This doesn't require a minimum of 3 members to work. Also, there's no risk of a tie in the election process, since the merge operation breaks ties with the member
ID, which is guaranteed to be unique (or else the consequences would be much worse than split brain).


thesentinels:

Thank you for contributing!

baumanj:
This is ready for review now

PR #5971, opened 2018-12-12
Identify generated files so GitHub can default to collapsing their diffs

baumanj:
See https://github.com/github/linguist for more

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+components/sup-protocol/src/generated/*.rs linguist-generated
+components/butterfly/src/generated/*.rs linguist-generated


thesentinels:

Thank you for contributing!

.gitattributes
```diff
@@ -0,0 +1,2 @@
+components/sup-protocol/src/generated/*.rs linguist-generated
```
raskchanky: The example at https://github.com/github/linguist#generated-code has `linguist-generated=true`.  Is the `=true` required?

baumanj: Probably not; I took my inspiration from https://robots.thoughtbot.com/github-diff-supression, but I guess we'll see.

christophermaier:
tenor-173784934

PR #5980, opened 2018-12-13
Split appveyor package jobs into 2 smaller jobs

scotthain:

### DIFF
--- a/support/ci/appveyor.ps1
+++ b/support/ci/appveyor.ps1
@@ -179,11 +179,14 @@ if((Test-DocsOnlyChanged -eq $true) -and (Test-ReleaseBuild -eq $false)) {
                     Copy-Item "/hab/pkgs/core/hab/*/*/bin/*" (Split-Path $habExe -Parent) -Force
                 }
             }
-            if(!(Test-PullRequest) -and $env:hab_components -ne "launcher") {
-                $env:HAB_BLDR_CHANNEL = $channel
-                & $habExe pkg exec core/hab-bintray-publish publish-studio
-                $env:HAB_BLDR_CHANNEL = $null
-                if ($LASTEXITCODE -ne 0) {exit $LASTEXITCODE}
+            # only publish the studio component
+            if($env:hab_components -eq "hab-bintray-publish") {
+                if(!(Test-PullRequest)) {
+                    $env:HAB_BLDR_CHANNEL = $channel
+                    & $habExe pkg exec core/hab-bintray-publish publish-studio
+                    $env:HAB_BLDR_CHANNEL = $null
+                    if ($LASTEXITCODE -ne 0) {exit $LASTEXITCODE}
+                }
             }
         }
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
Separating the exporters seems right. One thing that you will need to add is to ensure that the docker studio container is only built in the build that includes `hab-bintray-publish`. Which also probably means moving that to the same build that builds the studio component. As the code stands now, it builds the studio image if it is not building the launcher which means both of these builds will try to build the container.

support/ci/appveyor.ps1
```diff
@@ -179,7 +179,7 @@ if((Test-DocsOnlyChanged -eq $true) -and (Test-ReleaseBuild -eq $false)) {
                     Copy-Item "/hab/pkgs/core/hab/*/*/bin/*" (Split-Path $habExe -Parent) -Force
                 }
             }
-            if(!(Test-PullRequest) -and $env:hab_components -ne "launcher") {
+            if(!(Test-PullRequest) -and $env:hab_components -eq "studio") {
```
mwrock: `$env:hab_components` will never equal `studio` rather it would be equal to `hab;plan-build-ps1;studio;sup;bintray-publish`. However what might be cleaner is to move the entire `if` block up just 2 lines to be inside of the component loop. then change to `if(!(Test-PullRequest) -and $component -eq "bintray-publish") {` this would make sure the container only gets built in a merge commit just after building `bintray-publish` which is really what you want.

support/ci/appveyor.ps1
```diff
@@ -179,11 +179,14 @@ if((Test-DocsOnlyChanged -eq $true) -and (Test-ReleaseBuild -eq $false)) {
                     Copy-Item "/hab/pkgs/core/hab/*/*/bin/*" (Split-Path $habExe -Parent) -Force
                 }
             }
-            if(!(Test-PullRequest) -and $env:hab_components -ne "launcher") {
-                $env:HAB_BLDR_CHANNEL = $channel
-                & $habExe pkg exec core/hab-bintray-publish publish-studio
-                $env:HAB_BLDR_CHANNEL = $null
-                if ($LASTEXITCODE -ne 0) {exit $LASTEXITCODE}
+            if(!(Test-PullRequest)) {
```
mwrock: this whole block should move to line 181 above the closing brace that is on that line now and change `$env:hab_components -eq "studio"` to `$component -eq "bintray-publish"`

christophermaier:
@scotthain @mwrock What's the status of this PR? Are we close enough to the completion of the Buildkite work that this is no longer needed, or is there still utility in it?

scotthain:
@christophermaier there may still be value but I'd say let's revisit this in a day or two after we've talked about the migration.

scotthain:
Closing this PR, as current work will subsume the reason for this change.

Issue #5854, opened 2018-11-15
hab-sup panics when pkg_export tries to export struct


jaym:
This is order dependent:
```toml
# default.toml
[service]
http-port=1111

[foo]
foo="foo"
bar="bar"

```

```bash
pkg_exports=(
[http-port]=service.port
[foo]=foo
)
```
does not work works

```bash
pkg_exports=(
[foo]=foo
[http-port]=service.port
)
```
does not work works
--------------------

The following works:
```toml
# default.toml
[service]
http-port=1111

[moo]
foo="foo"
bar="bar"
```

```bash
pkg_exports=(
[http-port]=service.port
[moo]=moo
)
```

```bash
pkg_exports=(
[moo]=moo
[http-port]=service.port
)
```
does not work works

jamesc:
Full stacktrace on panic
```
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:477
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:391
   6: rust_begin_unwind
             at libstd/panicking.rs:326
   7: core::panicking::panic_fmt
             at libcore/panicking.rs:77
   8: core::result::unwrap_failed
             at libcore/macros.rs:26
   9: <core::result::Result<T, E>>::expect
             at libcore/result.rs:835
  10: habitat_butterfly::rumor::service::Service::new::{{closure}}
             at ./components/butterfly/src/rumor/service.rs:116
  11: <core::option::Option<T>>::map
             at libcore/option.rs:414
  12: habitat_butterfly::rumor::service::Service::new
             at ./components/butterfly/src/rumor/service.rs:115
  13: habitat_sup::manager::service::Service::to_rumor
             at components/sup/src/manager/service/mod.rs:671
  14: habitat_sup::manager::Manager::gossip_latest_service_rumor
             at components/sup/src/manager/mod.rs:1473
  15: habitat_sup::manager::Manager::add_service
[...]

PR #5987, opened 2018-12-15
Correctly serialize exports in service rumors

jamesc:
Solves #5854 by making sure that we always serialize the export configuration as valid TOML.

This PR moves the logic into `Cfg::to_exported` from the `Service::new` constructor so we can handle the error more sanely and reuse some of the serialization logic that is already there.

It feels like there might be a more natural way to do this via `Serde` by constructing some more objects to represent this service config but not sure if it's worth the extra code and abstraction. Would love some input on this.


thesentinels:

Thank you for contributing!

baumanj:
This is code in an area I'm not at all familiar with, so it's going to take me some time to give you good feedback. What's the urgency level of this change?

jamesc:
The bug has been there for a while already and there are workarounds.  Doesn't need to be squeezed into this release.

baumanj:
Apologies for the delay; I'll start digging into this tomorrow

baumanj:
I've mostly got my head wrapped around this code now (it's kind of a doozy). Hopefully I'll be able to finish this review tomorrow.

components/butterfly/src/rumor/service.rs
```diff
@@ -87,7 +87,7 @@ impl Service {
         package: &T,
         service_group: ServiceGroup,
         sys: SysInfo,
-        cfg: Option<&toml::value::Table>,
+        cfg: Option<Vec<u8>>,
```
christophermaier: We might be able to simplify things by using`toml::Value` instead. That's what's recommended in https://github.com/alexcrichton/toml-rs/issues/142, and clicking through to the code in question (current code is [here](https://github.com/alexcrichton/toml-rs/blob/56f9afb4d2abdf38ed3caf9f819ac8bacc290940/src/value.rs#L421-L451) actually shows that we have copied that code into our module (presumably from before that code was actually released).

We might be able to simplify the representation, fix the bug, _and_ ditch some complex code that we can now get for free.

baumanj: Now that we're dealing with the weird TOML table ordering issue, is there any reason we expect this should fail? I feel like this existing bug would've been a lot harder to find and fix it we handled the error before instead of panicking.

In other words, I think panicking here is appropriate unless we expect there to be an error case we want to handle more gracefully.

jamesc: I don't think it should now fail given we're using the right way from the crate instead of doing it in our code.

baumanj: So, do we agree that panicking is appropriate, then?

jamesc: Yep.

PR #5990, opened 2018-12-17
Don't panic if CLI parameters for "sup run" are missing from "sup term"

baumanj:
Both "sup run" and "sup term" use mgrcfg_from_matches, but only the former
guarantees a value (possibly default) for LISTEN_GOSSIP and LISTEN_HTTP.


thesentinels:

Thank you for contributing!

PR #6018, opened 2018-12-21
Don't panic in the unusual event persisting butterfly data takes over 30 s

baumanj:
The [`Duration`](http://doc.rust-lang.org/1.31.0/core/time/struct.Duration.html#method.sub) type is unsigned, so [subtracting a later instant from an earlier one](http://doc.rust-lang.org/1.31.0/std/time/struct.Instant.html#method.duration_since) causes this panic, observed in CI:

```
thread 'persist-e38384ae82144423ac70ec342e865997' panicked at 'specified instant was later than self', libstd/sys/unix/time.rs:292:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Change the calculation of our persist loop's wait time to use `checked_sub`, and if persist_data takes longer than our desired persist loop period, print a warning.

Thanks to to @stevendanna for surfacing this.

### DIFF
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -1212,11 +1212,22 @@ impl fmt::Display for Server {
 }
 
 fn persist_loop(server: Server) {
+    // TODO: Make this configurable with EnvConfig. That trait needs to move
+    // to common or core first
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

Issue #5292, opened 2018-07-03
End of line character DOESN'T get ignored when reading /hab/sup/default/CTL_SECRET 


jonasv:
Hi,

- Linux/OSX

if you open  /hab/sup/default/CTL_SECRET with your favourite editor (like vi for example) and if you just save that file - the secret gets changed by adding end of the line character which is usually invisible for the humans. This is a problem because if you look at the file - it looks without issues, so if you try to copy and use that secret on another end - it wont work because end of the line character does not get ignored which imho is a bug/problem.

PR #6022, opened 2018-12-27
fixes problem when newlines exist in CTL_SECRET file

jeremymv2:
Fixes https://github.com/habitat-sh/habitat/issues/5292
 - End of line character doesn't get ignored when reading /hab/sup/default/CTL_SECRET

This implementation uses [lines](https://doc.rust-lang.org/std/primitive.str.html#method.lines) to strip off any `\n` or `\r\n` from the secret.

Happy Holidays Habitat!

fireplace


thesentinels:

Thank you for contributing!

baumanj:
Restarted AppVeyor test that timed out

components/sup-protocol/src/lib.rs
```diff
@@ -168,3 +181,52 @@ pub fn socket_addr_env_or_default(env_var: &str, default: SocketAddr) -> SocketA
         .parse()
         .unwrap_or(default)
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::fs::File;
```
baumanj: With the above suggestions, you only need to add the one import here to cover the use of the `write!` macro.
```suggestion
    use std::fs::File;
    use std::io::Write
```

components/sup-protocol/Cargo.toml
```diff
@@ -17,6 +17,7 @@ prost-derive = "*"
 rand = "*"
 serde = "*"
 serde_derive = "*"
+tempfile = "*"
```
baumanj: Moving this to `dev-dependencies` avoids adding it when building for a non-dev, non-test environment. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

components/sup-protocol/src/lib.rs
```diff
@@ -69,7 +71,8 @@ use core::env as henv;
 use net::{ErrCode, NetResult};
 use rand::RngCore;
 use std::fs::File;
-use std::io::Read;
+use std::io::prelude::*;
+use std::io::BufReader;
```
baumanj: I think the fix can be made with a simpler change that obviates these imports. See below.

components/sup-protocol/src/lib.rs
```diff
@@ -119,19 +122,29 @@ where
...
+                _ => Err(net::err(
                     ErrCode::Io,
                     format!(
-                        "IoError while reading or writing ctl secret, {}, {}",
+                        "Could not find secret in ctl secret file, {}",
                         secret_key_path.display(),
-                        e
                     ),
-                )
-            })?;
```
baumanj: I think this change is sufficient to fix the error. I'd prefer to keep the change as small as possible.
```rust
        *out = out.trim_end().into();
```

components/sup-protocol/Cargo.toml
jeremymv2: Good call. I have moved this under `dev-dependencies`.

components/sup-protocol/src/lib.rs
jeremymv2: Nice recommendation. When I first read the `trim_end()` docs I didn't realize it would also handle CRLF that is typical on Windows.  That's great!

components/sup-protocol/src/lib.rs
jeremymv2: agreed!

components/sup-protocol/src/lib.rs
jeremymv2: completely agreed.

jeremymv2:
Thanks for the feedback @baumanj. I finally got some time to circle back on this. I have functionally tested the changes and everything seems good.

jeremymv2:
It looks like one of the AppVeyor jobs timed out after an hour.

baumanj:
This looks good to me, though there's one little conflict that needs to be resolved due to some intervening changes.

jeremymv2:
🚀 Thank you @baumanj! 
I've rebased, resolved the minor conflict and pushed up.

jeremymv2:
This travis-ci error does not look related to the changes in the commit:
```
error: process didn't exit successfully: `/home/travis/build/habitat-sh/habitat/target/debug/deps/compilation-5b01b594d4eb48df` (signal: 11, SIGSEGV: invalid memory reference)
make: *** [unit-sup] Error 101
The command "./support/ci/rust_tests.sh" exited with 2.
```

baumanj:
Restarted the travis job

baumanj:
Merged! Thanks again for a great contribution.

PR #6026, opened 2019-01-02
Refactor some misuse of RwLock fields on the butterfly Server struct

baumanj:
- Remove unnecessary `Arc<RwLock<_>`> around `SocketAddr` fields that were never mutated
- Change read-only `block_list` access to use `RwLock::read`, not `::write`
- Return `SocketAddr` references rather than owned values to avoid cloning
- Make `swim_port()` and `gossip_port()` private
- Change `dat_file` RwLock to Mutex since there is no read/write distinction and move the `Option` to the outside so we don't incur the unlock cost to check if the option is some

### DIFF
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -37,7 +37,7 @@ use std::path::PathBuf;
 use std::result;
 use std::sync::atomic::{AtomicBool, AtomicIsize, Ordering};
 use std::sync::mpsc::{self, channel};
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::thread;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/butterfly/src/server/mod.rs
```diff
@@ -321,11 +321,11 @@ impl Server {
...
                     update_store: RumorStore::default(),
                     departure_store: RumorStore::default(),
-                    swim_addr: Arc::new(RwLock::new(swim_socket_addr)),
-                    gossip_addr: Arc::new(RwLock::new(gossip_socket_addr)),
+                    swim_addr: swim_socket_addr,
+                    gossip_addr: gossip_socket_addr,
                     suitability_lookup: Arc::new(suitability_lookup),
                     data_path: Arc::new(data_path.as_ref().map(|p| p.into())),
-                    dat_file: Arc::new(RwLock::new(None)),
+                    dat_file: None,
```
christophermaier: tenor-173784934

PR #6032, opened 2019-01-04
Parameterize the rumor sharing limit

christophermaier:
Further reading of the "Epidemic Algorithms for Replicated Database
Maintenance" paper reveals that we may not be managing our "rumor
heat" correctly. The paper describes many variations of "cooling" a
hot rumor, all of which speak in terms of sending out a given rumor to
the entire network as a whole, rather than tracking the number of
times each rumor is sent to every individual in the network.

To begin to investigate the impact of bringing our approach more in
line with the documented algorithm, we are making the number of times
we send a rumor to any individual Supervisor tunable using the
environment variable `HAB_RUMOR_SHARE_LIMIT`. This should be
considered an "undocumented" variable, and is only being exposed for
the purposes of internal testing; general Habitat users shouldn't be
using it.

Right now, testers would only ever want to set this variable to `1`,
meaning that each rumor is sent to every other Supervisor once, rather
than the current default of twice. Doing so ought to cut overall
network traffic in half. (It should be noted that each Supervisor
still stands to receive any given rumor as many times as there are
Supervisors in the network, so there is still a considerable amount of
redundancy.)


components/butterfly/src/rumor/heat.rs
```diff
@@ -23,23 +23,62 @@
...
+    type Err = Error;
+    fn from_str(s: &str) -> result::Result<Self, Self::Err> {
+        let raw = s
+            .parse::<usize>()
+            .map_err(|_| Error::InvalidRumorShareLimit)?;
+        if raw > 0 {
+            Ok(RumorShareLimit(raw))
+        } else {
+            Err(Error::InvalidRumorShareLimit)
+        }
```
baumanj: I'm fond of the slightly more FP approach, but either one is fine by me:
```rust
        s.parse::<usize>()
            .map_err(|_| Error::InvalidRumorShareLimit)
            .and_then(|raw| {
                if raw > 0 {
                    Ok(RumorShareLimit(raw))
                } else {
                    Err(Error::InvalidRumorShareLimit)
                }
            })
```

christophermaier:
@baumanj Using a `RumorHeat` type is an interesting idea... I think I'd like to go ahead and get this merged as-is, but I'll play around with it. Thanks!

PR #6037, opened 2019-01-04
split sections by custom token when generating cli docs

### LINKED ISSUE #6035: make cli_docs is omitting several args of sup commands
As an example see https://github.com/habitat-sh/habitat/pull/6034

This seems to have begun with the 0.72.0 release and is caused by an extra line feed after the `--config-from` arg. I'm not sure why the extra linefeed is appearing.

mpeck:
We were assuming that sections were separated by two newlines, but that isn't always the case. When commands have long argument/option names the help output wraps and adds multiple newlines. This adds a custom token before each section heading and then splits on that token. There are still some issues with spacing on commands with long argument/option names, but at least everything shows up.

Fixes #6035 

### DIFF
--- a/www/scripts/generate-cli-docs.js
+++ b/www/scripts/generate-cli-docs.js
@@ -21,8 +21,10 @@ function runCommand(command, args) {
 }
 
 function parseOutput(command, output) {
+  const sectionToken = '\^/--IMATOKEN--\^/';
+
   const lines = output.split('\n');
-  const sections = output.split('\n\n');
+  const sections = output.replace(/\n\n(.+):\n/g, sectionToken + '$1:').split(sectionToken);
 
   let result = {
     name: lines[0].trim(),
--- a/www/scripts/generate-cli-docs.js
+++ b/www/scripts/generate-cli-docs.js
@@ -2,7 +2,7 @@ const { spawnSync } = require('child_process');
 const { platform } = require('os');
 
 function getHelp(command, sub) {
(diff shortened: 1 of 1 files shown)


mwrock:
Great! This worked well. However I got an error for `unknown command --` included for `hab studio` subcommands `run`, `rm`, and `new`. Passing `-h` instead of `--help` got rid of those.

www/scripts/generate-cli-docs.js
```diff
@@ -21,8 +21,10 @@ function runCommand(command, args) {
 }
 
 function parseOutput(command, output) {
+  const sectionToken = '\^/--IMATOKEN--\^/';
+
   const lines = output.split('\n');
-  const sections = output.split('\n\n');
+  const sections = output.replace(/\n\n(.+):\n/g, sectionToken + '$1:').split(sectionToken);
```
christophermaier: Do we need to do this dance with a custom `sectionToken` if we already know enough to place that custom token correctly in the first place?

mpeck: The regex matches on two empty spaces followed by some text and then a colon. The text is the section heading. I can split on that and along with the capture group for the heading text to get an array with the heading text followed by the section text, but then I would have to refactor how we break out and markup sections later on. I went with a token because it seemed like a simpler fix in this instance and I think the pattern leads to a bit more flexibility in the future, namely with fixing some of the spacing issues that we still have.

Issue #5935, opened 2018-12-06
Remove Composites Code


christophermaier:
## Description 
Since we've [announced the deprecation of composites](https://www.habitat.sh/blog/2018/10/shelving-composites/), we can now remove composite code from the Supervisor (including anything around the "three-part binds"), `hab-plan-build`, tests, etc.

We've had a sufficient number of releases since the announcement that we're able to yank the code without further ceremony.

## Acceptance Criteria
- [x] Composites code removed from Habitat
- [x] Remove `must_be_runnable ` parameter from [`satisfy_or_install`](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/util/pkg.rs#L74)

PR #6043, opened 2019-01-08
Only start services that are runnable

### LINKED ISSUE #3468: The Supervisor repeatedly tries to run "services" that are not actually services
Before loading or starting a service, we should actually check that the package is actually a service (i.e., something that has a `run` script). Otherwise, the supervisor will endlessly try to execute a non-existent file, as seen below when I try to start `core/hab-butterfly` (a library, not a service).

```
hab-sup(MR): Supervisor Member-ID b3f9ee13770f4cffa5f0f04d624790e7
hab-sup(MR): Starting core/hab-butterfly
hab-butterfly.default(HK): Hooks compiled
hab-butterfly.default(SR): Error finding run file: No such file or directory (os error 2)
hab-butterfly.default(SR): Initializing
hab-butterfly.default(SV): Starting service as user=hab, group=hab
hab-butterfly.default(SR): Service start failed: hab-sup(ER)[src/error.rs:479:8]: Unknown: Unable to spawn process, No such file or directory (os error 2)
hab-butterfly.default(SV): Starting service as user=hab, group=hab
hab-butterfly.default(SR): Service restart failed: hab-sup(ER)[src/error.rs:479:8]: Unknown: Unable to spawn process, No such file or directory (os error 2)
hab-butterfly.default(SV): Starting service as user=hab, group=hab
hab-butterfly.default(SR): Service restart failed: hab-sup(ER)[src/error.rs:479:8]: Unknown: Unable to spawn process, No such file or directory (os error 2)
hab-butterfly.default(SV): Starting service as user=hab, group=hab
hab-butterfly.default(SR): Service restart failed: hab-sup(ER)[src/error.rs:479:8]: Unknown: Unable to spawn process, No such file or directory (os error 2)
... etc, until the heat death of the universe
```

`PackageArchive` already has an [is_a_service](https://github.com/habitat-sh/habitat/blob/ab7d932ac4ece39a922b2961742f5531f84299ad/components/core/src/package/archive.rs#L113-L122) function, which can surface this information. We may want to add a similar function to `PackageInstall`, because that is what we'll have access to in the `hab` CLI when we're starting / loading services.

(We could add some checking in the supervisor itself to prevent it from trying to re-run if there's no `run` hook; that may be a more intricate change, though, and may end up falling out of the upcoming rewrite of the supervisor's internals. In any event, alerting the user as soon as possible that they can't run a given package is something we should do anyway.)

Incidentally, it may also be worth looking at factoring out these metadata-lookup functions from `PackageArchive` and `PackageInstall`; they're both returning the same information, but one is from a `.hart` file, and the other from an expanded directory on disk.

jamesc:
Fixes #3468 

Before writing out new service spec files check that the service is
actually runnable. We do the checks in `manager::command` so we
never actually trigger an attempt to start the service (which gives
unhelpful user messages).

We still need to install the package and it's dependencies in order
to check if it's runnable.  We don't clean up those packages

NOTE:  This code doesn't cover the code paths covering the composite
use cases as they're going to be removed.

### DIFF
--- a/components/sup/src/manager/commands.rs
+++ b/components/sup/src/manager/commands.rs
@@ -244,7 +244,8 @@ pub fn service_load(
             // desired package identifier, it will be used;
             // otherwise, we'll install the latest suitable
             // version from the specified Builder channel.
-            let installed = util::pkg::satisfy_or_install(req, &source, &bldr_url, &bldr_channel)?;
+            let installed =
+                util::pkg::satisfy_or_install(req, &source, &bldr_url, &bldr_channel, true)?;
 
             let mut specs = generate_new_specs_from_package(&installed, &opts)?;
 
@@ -298,6 +299,7 @@ pub fn service_load(
                         &source,
                         &service_spec.bldr_url,
                         &service_spec.channel,
+                        true,
                     )?;
 
                     save_spec_for(&mgr.cfg, &service_spec)?;
@@ -349,6 +351,7 @@ pub fn service_load(
                             // like services can.
                             &bldr_url,
                             &bldr_channel,
+                            false,
                         )?;
 
                         // Generate new specs from the new composite package and
--- a/components/sup/src/error.rs
+++ b/components/sup/src/error.rs
@@ -155,6 +155,7 @@ pub enum Error {
     NotifyError(notify::Error),
     NulError(ffi::NulError),
     PackageNotFound(package::PackageIdent),
+    PackageNotRunnable(package::PackageIdent),
     Permissions(String),
     PidFileCorrupt(PathBuf),
     PidFileIO(PathBuf, io::Error),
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/commands.rs
```diff
@@ -715,6 +717,17 @@ fn existing_specs_for_ident(cfg: &ManagerConfig, ident: &PackageIdent) -> Result
     }
 }
 
+fn assert_runnable_package(package: &PackageInstall) -> Result<()> {
```
baumanj: I think the naming is a bit misleading since "assert" is a formal construct in rust and it implies a panic.

jamesc:
> Since there are only 3 callers for `satisfy_or_install`, I think I'd prefer a slightly different approach where we add a parameter

Once the composite code is removed then the two remaining `satisfy_or_install` commands should always be checking `is_runnable()`.  So I'm happy to do this but it should only be a short term fix and the parameter would be remove when we remove composites.  I'll add a note the that effect

components/sup/src/manager/commands.rs
jamesc: Good point.  This issue disappears when I move the implementation into satisfy_or_install

baumanj:
I added a note to https://github.com/habitat-sh/habitat/issues/5935 as well

PR #6044, opened 2019-01-09
Simplify insert_service

baumanj:
This simplifies the code without any functional change.

A potential follow-up may avoid unnecessarily departing members when doing so isn't necessary to preserve quorum.


thesentinels:

Thank you for contributing!

PR #6053, opened 2019-01-14
Improve butterfly Server::insert_service

baumanj:
- Don't depart `Confirmed` members when adding new members to a service group unless required to maintain quorum
- Improve test coverage
- Remove flaky test which is now redundant


thesentinels:

Thank you for contributing!

components/butterfly/src/server/mod.rs
```diff
@@ -1335,6 +1347,123 @@ mod tests {
...
+            incarnation: 0,
+            initialized: true,
+            pkg: "foo/bar".into(),
+            cfg: vec![],
+            sys: SysInfo::default(),
+        }
+    }
+
+    impl RumorStore<Service> {
+        fn contains(&self, service: &Service) -> bool {
```
christophermaier: Would it potentially be worth adding this to the "real" `RumorStore<Service>`, rather than having it be just a test-only construct?

components/butterfly/src/server/mod.rs
```diff
@@ -1255,14 +1259,18 @@ impl<'a> Serialize for ServerProxy<'a> {
...
     use super::*;
-    use crate::rumor::{election::Term, service::SysInfo};
+    use crate::rumor::election::Term;
     use habitat_core::service::ServiceGroup;
     use std::str::FromStr;
 
-    fn get_mock_service(member_id: String, service_group: ServiceGroup) -> Service {
+    fn check_quorum_returns(val: bool) -> impl Fn(&str) -> bool {
+        move |_: &str| val
+    }
```
christophermaier: This is very nice.

components/butterfly/src/server/mod.rs
```diff
@@ -1377,8 +1369,15 @@ mod tests {
         let service_store = RumorStore::default();
         let member_list = MemberList::new();
         let rumor_heat = RumorHeat::default();
+        // let check_quorum = |_: &str| true;
```
christophermaier: Might want to delete this.

components/butterfly/src/server/mod.rs
baumanj: Currently, I don't see any other places in the non-test code where it would be used, so I'd rather leave it in the test scope. That makes it clearer that it can be removed if the tests that use it no longer need it in the future.

PR #6065, opened 2019-01-16
Fix "uncontroversial" clippy lints and start making them errors in TravisCI

baumanj:
Which lints are clearly ones we should fix (and fail on in the future) is subjective, but this is a first cut at making things better, so that we can be more idiomatic, consistent and performant going forward and clippy can help let us know when we stray off the path.

Deciding what to do about the remaining lints in the `UNEXAMINED_LINTS` category in the [Makefile](https://github.com/habitat-sh/habitat/blob/master/Makefile) will be an exercise for another PR.

This is a longer PR than ideal, but since so many of the changes are totally mechanical, I hope it's not totally unworkable. I've tried to call out any changes that are not totally automatic so they can be appropriately scrutinized.

I **highly** recommend viewing this [with whitespace changes hidden](https://github.com/habitat-sh/habitat/pull/6065/files?utf8=✓&diff=split&w=1).

Some changes that occur a lot that might benefit from some explanation:

## Avoiding unnecessary execution
```rust
.expect(&format!(…))
```
→
```rust
.unwrap_or_else(|_| panic!(…))
```
avoids unconditionally evaluating the contents of the `format` macro, and allocating a `String`. The alternative only runs in the case we're panicking. The same goes for constructs like changing `unwrap_or(function_call())` → `unwrap_or_else(|| function_call())` or `unwrap_or(Foo::default())` → `unwrap_or_default()`.

## Avoiding passing unit values
```rust
Ok(result_or_result_returning_func?)
```
→
```rust
result_or_result_returning_func?;
Ok()
```
This isn't terrible on it's own, though it is a bit weird. The real reason it fix it is that it trips the [unit_arg](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#unit_arg) lint, which is something fairly useful. Plus, I think the meaning of:
```rust
Ok(process::become_command(cmd, args)?)
```
is less obvious than
```rust
process::become_command(cmd, args)?;
Ok(())
```

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -53,7 +53,7 @@ use url::percent_encoding::{percent_encode, PATH_SEGMENT_ENCODE_SET};
 header! { (XFileName, "X-Filename") => [String] }
 header! { (ETag, "ETag") => [String] }
 
-const DEFAULT_API_PATH: &'static str = "/v1";
+const DEFAULT_API_PATH: &str = "/v1";
 
 #[derive(Clone, Deserialize)]
 #[serde(rename = "error")]
@@ -127,9 +127,9 @@ impl fmt::Display for SchedulerResponse {
             output.push(format!("Created at: {}", c.to_string()));
         }
 
-        if self.projects.len() > 0 {
+        if !self.projects.is_empty() {
             output.push("".to_string());
-            output.push(format!("Reverse dependencies:"));
+            output.push("Reverse dependencies:".to_string());
             let mut projects = self.projects.clone();
             projects.sort_by(|a, b| a.ident.cmp(&b.ident));
 
@@ -163,6 +163,7 @@ pub struct JobGroupPromoteResponse {
 mod json_u64 {
     use serde::{self, Deserialize, Deserializer, Serializer};
 
+    #[allow(clippy::trivially_copy_pass_by_ref)]
     pub fn serialize<S>(num: &u64, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
@@ -640,7 +641,6 @@ impl Client {
         debug!("Response body: {:?}", encoded);
         let channels: Vec<String> = serde_json::from_str::<Vec<String>>(&encoded)?
             .into_iter()
-            .map(|m| m.into())
             .collect();
         Ok(channels)
     }
@@ -805,7 +805,7 @@ impl Client {
         });
 
         match self.download(req_builder, dst_path.as_ref(), token, progress) {
-            Ok(file) => Ok(PackageArchive::new(PathBuf::from(file))),
+            Ok(file) => Ok(PackageArchive::new(file)),
             Err(e) => Err(e),
         }
     }
@@ -1042,16 +1042,13 @@ impl Client {
         include_sandbox_channels: bool,
     ) -> Result<Vec<String>> {
         let path = format!("depot/channels/{}", origin);
-        let mut res;
-
-        if include_sandbox_channels {
-            res = self
-                .0
+        let mut res = if include_sandbox_channels {
+            self.0
                 .get_with_custom_url(&path, |url| url.set_query(Some("sandbox=true")))
-                .send()?;
+                .send()?
         } else {
-            res = self.0.get(&path).send()?;
-        }
+            self.0.get(&path).send()?
+        };
 
         match res.status {
             StatusCode::Ok | StatusCode::PartialContent => {
@@ -1309,8 +1306,8 @@ mod tests {
     #[test]
     fn json_round_trip_u64_fields() {
         let pre = OriginPrivateSigningKey {
-            id: 705705315793903646,
-            origin_id: 705705305031319582,
+            id: 705_705_315_793_903_646,
+            origin_id: 705_705_305_031_319_582,
             name: "core".to_string(),
             revision: "20160810182414".to_string(),
             body: vec![1, 2, 3],
--- a/components/butterfly-test/src/lib.rs
+++ b/components/butterfly-test/src/lib.rs
@@ -122,12 +122,13 @@ impl DerefMut for SwimNet {
 
 impl SwimNet {
     pub fn new_with_suitability(suitabilities: Vec<u64>) -> SwimNet {
-        let count = suitabilities.len();
-        let mut members = Vec::with_capacity(count);
-        for x in 0..count {
-            members.push(start_server(&format!("{}", x), None, suitabilities[x]));
+        SwimNet {
+            members: suitabilities
+                .into_iter()
+                .enumerate()
+                .map(|(x, suitability)| start_server(&format!("{}", x), None, suitability))
+                .collect(),
         }
-        SwimNet { members: members }
     }
 
     pub fn new(count: usize) -> SwimNet {
@@ -165,12 +166,10 @@ impl SwimNet {
                 if pos == x_pos {
                     continue;
                 }
-                let server_b = self.members.get(x_pos).unwrap();
-                to_mesh.push(member_from_server(server_b))
+                to_mesh.push(member_from_server(&self.members[x_pos]))
             }
-            let server_a = self.members.get(pos).unwrap();
             for server_b in to_mesh.into_iter() {
-                server_a.insert_member(server_b, Health::Alive);
+                self.members[pos].insert_member(server_b, Health::Alive);
             }
         }
     }
@@ -250,24 +249,22 @@ impl SwimNet {
         self.gossip_rounds().iter().map(|r| r + count).collect()
     }
 
-    pub fn check_rounds(&self, rounds_in: &Vec<isize>) -> bool {
-        let mut finished = Vec::with_capacity(rounds_in.len());
-        for (i, round) in rounds_in.into_iter().enumerate() {
-            if self.members[i].paused() {
-                finished.push(true);
-            } else {
-                if self.members[i].swim_rounds() > *round {
-                    finished.push(true);
-                } else {
-                    finished.push(false);
-                }
+    fn check_rounds_impl(
+        &self,
+        rounds_in: &[isize],
+        get_rounds: impl Fn(&Server) -> isize,
+    ) -> bool {
+        for (member, round) in self.members.iter().zip(rounds_in) {
+            if !member.paused() && (get_rounds)(member) <= *round {
+                return false;
             }
         }
-        if finished.iter().all(|m| m == &true) {
-            return true;
-        } else {
-            return false;
-        }
+
+        true
+    }
+
+    pub fn check_rounds(&self, rounds_in: &[isize]) -> bool {
+        self.check_rounds_impl(rounds_in, Server::swim_rounds)
     }
 
     pub fn wait_for_rounds(&self, rounds: isize) {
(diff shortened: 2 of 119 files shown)


thesentinels:

Thank you for contributing!

components/butterfly-test/src/lib.rs
```diff
@@ -122,12 +122,13 @@ impl DerefMut for SwimNet {
...
-        let count = suitabilities.len();
-        let mut members = Vec::with_capacity(count);
-        for x in 0..count {
-            members.push(start_server(&format!("{}", x), None, suitabilities[x]));
+        SwimNet {
+            members: suitabilities
+                .into_iter()
+                .enumerate()
+                .map(|(x, suitability)| start_server(&format!("{}", x), None, suitability))
+                .collect(),
```
baumanj: This is a bit of a refactor, but I'm pretty sure it's correct and more idiomatic

components/butterfly-test/src/lib.rs
```diff
@@ -165,12 +166,10 @@ impl SwimNet {
...
                     continue;
                 }
-                let server_b = self.members.get(x_pos).unwrap();
-                to_mesh.push(member_from_server(server_b))
+                to_mesh.push(member_from_server(&self.members[x_pos]))
             }
-            let server_a = self.members.get(pos).unwrap();
             for server_b in to_mesh.into_iter() {
-                server_a.insert_member(server_b, Health::Alive);
+                self.members[pos].insert_member(server_b, Health::Alive);
```
baumanj: Changing `members.get(foo).unwrap()` → `members[foo]` is totally mechanical, but once I did that, the need for the `let` statements seemed to go away, so I further simplified.

components/butterfly-test/src/lib.rs
```diff
@@ -250,24 +249,22 @@ impl SwimNet {
...
             }
         }
-        if finished.iter().all(|m| m == &true) {
-            return true;
-        } else {
-            return false;
-        }
+
+        true
+    }
```
baumanj: `check_rounds ` and `check_gossip_rounds` were identical aside from one identifier, so I couldn't resist.

Additionally, I simplified the logic; there's no need to go through the whole iteration if at the end you're returning that *all* the values were `true`.

components/butterfly-test/src/lib.rs
```diff
@@ -428,13 +409,7 @@ impl SwimNet {
...
         loop {
             let network_health = self.network_health_of(to_check);
-            if network_health.iter().all(|x| {
-                if let &Some(ref h) = x {
-                    *h == health
-                } else {
-                    false
-                }
-            }) {
+            if network_health.iter().all(|&x| x == Some(health)) {
```
baumanj: Much more concise, but functionally equivalent

components/common/src/ui.rs
```diff
@@ -432,46 +428,32 @@ impl UIReader for UI {
...
-                        true => format!(
-                            "{}{}{}",
-                            Colour::White.paint("[yes/"),
-                            Colour::White.bold().paint("No"),
-                            Colour::White.paint("/quit]")
-                        ),
-                        false => format!("[yes/No/quit]"),
-                    }
-                }
+        let stream = &mut self.shell.out;
```
baumanj: The code below is a refactor. The existing version was awfully redundant and make the similarities hard to see across the colored/non-colored versions.

components/hab/src/command/plan/init.rs
```diff
@@ -225,33 +221,25 @@ fn render_ignorefile(ui: &mut UI, root: &str) -> Result<()> {
...
-
-    if let Some(parent) = path.parent() {
-        return is_git_managed(&parent);
-    }
-
-    return false;
+    path.join(".git").is_dir()
+        || path
+            .parent()
+            .map_or(false, |parent| is_git_managed(&parent))
```
baumanj: This is a refactor, but hopefully a clear improvement

components/hab/src/main.rs
```diff
@@ -499,8 +498,8 @@ fn sub_pkg_export(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     let url = bldr_url_from_matches(&m)?;
     let channel = m
         .value_of("CHANNEL")
-        .and_then(|c| Some(c.to_string()))
-        .unwrap_or(channel::default());
+        .map(str::to_string)
+        .unwrap_or_else(channel::default);
```
baumanj: Note that we *don't* want `unwrap_or_default` here, as `String::default()` != `channel::default()`

components/hab/src/main.rs
```diff
@@ -1440,18 +1438,18 @@ fn enable_features_from_env(ui: &mut UI) {
...
         }
     }
 }
 
 fn handle_ctl_reply(reply: SrvMessage) -> result::Result<(), SrvClientError> {
-    let mut bar = pbr::ProgressBar::<io::Stdout>::new(0);
-    bar.set_units(pbr::Units::Bytes);
-    bar.show_tick = true;
-    bar.message("    ");
+    let mut progress_bar = pbr::ProgressBar::<io::Stdout>::new(0);
```
baumanj: Clippy complains about the name `bar`.

components/pkg-export-docker/src/build.rs
```diff
@@ -543,11 +544,14 @@ impl BuildRootContext {
         let gid = DEFAULT_USER_AND_GROUP_ID;
 
         let pkg = self.primary_svc()?;
-        let user_name = pkg.svc_user().unwrap_or(Some(String::from("hab"))).unwrap();
+        let user_name = pkg
+            .svc_user()
+            .unwrap_or_default()
+            .unwrap_or_else(|| String::from("hab"));
```
baumanj: [`svc_user`](https://github.com/habitat-sh/core/blob/master/components/core/src/package/install.rs#L579) (and similarly `group` below) return `Result<Option<String>>`, so the first `unwrap_or_default` will generate an `Option::None` and trigger the `or_else` behavior in the case of `Result::Err`.

components/sup/src/census.rs
```diff
@@ -144,10 +145,11 @@ impl CensusRing {
         // is an indeterminate health anywhere.
         service_rumors.with_keys(|(service_group, rumors)| {
             if let Ok(sg) = service_group_from_str(service_group) {
+                let local_member_id = Cow::from(&self.local_member_id);
```
baumanj: This is a bit tricky, with the original `or_insert` code, we allocated a new `CensusGroup` unconditionally. However, we can't just change to `or_insert_with` because the mutable borrow of self required to call [`entry`](http://doc.rust-lang.org/1.31.1/std/collections/hash_map/struct.HashMap.html#method.entry) conflicts with the immutable borrow to access `&self.local_member_id`. We could've unconditionally cloned `&self.local_member_id`, but using [`Cow`](http://doc.rust-lang.org/1.31.1/std/borrow/enum.Cow.html) gives us the ability to do the right thing at runtime.

components/sup/src/census.rs
```diff
@@ -370,25 +373,24 @@ impl CensusGroup {
...
+    pub fn members(&self) -> impl Iterator<Item = &CensusMember> {
+        self.population.values()
     }
 
     /// Same as `members`, but only returns members that are either
     /// alive or suspect, i.e., nothing that is confirmed dead or
     /// departed. These are the members that we'll reasonably be
     /// interacting with at runtime.
-    pub fn active_members(&self) -> Vec<&CensusMember> {
+    pub fn active_members(&self) -> impl Iterator<Item = &CensusMember> {
```
baumanj: Most of the consumers don't need to `collect`, so we can avoid some allocations

components/sup/src/manager/service_updater.rs
```diff
@@ -245,10 +245,10 @@ impl ServiceUpdater {
...
                     {
                         Some(census_group) => {
-                            if census_group.members().iter().any(|cm| {
-                                cm.pkg.as_ref().unwrap()
-                                    != census_group.me().unwrap().pkg.as_ref().unwrap()
-                            }) {
+                            if census_group
+                                .members()
+                                .any(|cm| cm.pkg != census_group.me().unwrap().pkg)
+                            {
```
baumanj: This should be equivalent since we can compare `Option`s

components/sup/tests/utils/test_butterfly.rs
```diff
@@ -64,20 +64,19 @@ impl Client {
         T: ToString,
     {
         let config = config.to_string();
+        let config = config.as_bytes();
```
baumanj: This avoids calling `config.as_bytes()` twice

components/builder-api-client/src/lib.rs
```diff
@@ -1309,8 +1306,8 @@ mod tests {
     #[test]
     fn json_round_trip_u64_fields() {
         let pre = OriginPrivateSigningKey {
-            id: 705705315793903646,
-            origin_id: 705705305031319582,
+            id: 705_705_315_793_903_646,
+            origin_id: 705_705_305_031_319_582,
```
christophermaier: tenor-158320510

components/butterfly-test/src/lib.rs
christophermaier: Nice one!

components/butterfly-test/src/lib.rs
```diff
@@ -250,24 +249,22 @@ impl SwimNet {
...
-                } else {
-                    finished.push(false);
-                }
+    fn check_rounds_impl(
+        &self,
+        rounds_in: &[isize],
+        get_rounds: impl Fn(&Server) -> isize,
+    ) -> bool {
+        for (member, round) in self.members.iter().zip(rounds_in) {
+            if !member.paused() && (get_rounds)(member) <= *round {
```
christophermaier: Why is `get_rounds` parenthesized here?

components/butterfly-test/src/lib.rs
christophermaier: tenor-128602034

components/hab/src/error.rs
```diff
@@ -55,7 +55,8 @@ pub enum Error {
     FileNotFound(String),
     HabitatCommon(common::Error),
     HabitatCore(hcore::Error),
-    HandlebarsRenderError(handlebars::TemplateRenderError),
+    // Boxed because https://rust-lang-nursery.github.io/rust-clippy/master/index.html#large_enum_variant
+    HandlebarsRenderError(Box<handlebars::TemplateRenderError>),
```
christophermaier: Neat trick

components/hab/src/lib.rs
```diff
@@ -51,18 +51,18 @@ pub mod error;
...
 pub use crate::hcore::AUTH_TOKEN_ENVVAR;
 
 features! {
     pub mod feat {
-        const List = 0b00000001,
-        const OfflineInstall = 0b00000010,
-        const IgnoreLocal = 0b00000100
+        const List           = 0b0000_0001,
+        const OfflineInstall = 0b0000_0010,
+        const IgnoreLocal    = 0b0000_0100
```
christophermaier: _Sooooooooo_ good

components/hab/src/main.rs
christophermaier: Ugh... my kingdom for a proper `Channel` type!

https://github.com/habitat-sh/core/blob/22f09fc7ca4e8dd87a655cd2186eff1c20d6eea9/components/core/src/channel.rs

components/hab/src/main.rs
christophermaier: I guess to prevent you from adding lots of meaningless variables named `foo`, `bar`, `baz`, etc.?

components/pkg-export-docker/src/build.rs
christophermaier: Slick.

components/sup/src/census.rs
christophermaier: I friggin' love `Cow`s... they're so useful.

baumanj: 🐮

components/butterfly-test/src/lib.rs
baumanj: because it's a function pointer. It may not be strictly necessary to add parens in this case, but I know rust sometimes requires it. I'll see if they can be removed and if so make that change later.

baumanj: Yeah, I guess it's only when you're accessing a closure as a field of some other member. I'll get rid of those.

baumanj: Fixed in https://github.com/habitat-sh/habitat/pull/6068

PR #6066, opened 2019-01-16
Support for recursive subdirectories for config.

### LINKED ISSUE #5173: Support for subdirectories in config template directory
I'm using Habitat to start up a service that expects a large number of configuration files in a very specific directory structure. Currently, subdirectories in /hab/pkg/«ident»/config are ignored when config is templated out to /hab/svc/«svc»/config. Habitat should instead walk the directory tree and template out any files in all subdirectories as well (recursively).

Currently, I'm working around this by including all templated config files in the top level of the config directory and then manually copying those files into the necessary structure, but this is error-prone and a major pain.

jamesc:
Fixes #5173 

Allows for the `config` directory in a package to have sub-directories.  we now will generate `svc` config files for all files in the sub-directories.

### DIFF
--- a/components/common/src/templating/config.rs
+++ b/components/common/src/templating/config.rs
@@ -38,9 +38,9 @@ static ENV_VAR_PREFIX: &'static str = "HAB";
 /// is deeper than this value crosses into overly complex territory when describing configuration
 /// for a single service.
 static TOML_MAX_MERGE_DEPTH: u16 = 30;
-#[cfg(not(windows))]
+#[cfg(unix)]
 pub const CONFIG_PERMISSIONS: u32 = 0o740;
-#[cfg(not(windows))]
+#[cfg(unix)]
 pub const CONFIG_DIR_PERMISSIONS: u32 = 0o770;
 
 /// Describes the path to user configuration that is used by the
@@ -427,12 +427,22 @@ impl Serialize for Cfg {
 pub struct CfgRenderer(TemplateRenderer);
 
 impl CfgRenderer {
+    /// Create a new `CfgRenderer` and load template files from a
+    /// configuration directory, if it exists.
     pub fn new<T>(templates_path: T) -> Result<Self>
     where
         T: AsRef<Path>,
     {
-        let templates = load_templates(templates_path)?;
-        Ok(CfgRenderer(templates))
+        if templates_path.as_ref().is_dir() {
+            load_templates(
+                templates_path.as_ref(),
+                &PathBuf::new(),
+                TemplateRenderer::new(),
+            )
+            .map(CfgRenderer)
+        } else {
+            Ok(CfgRenderer(TemplateRenderer::new()))
+        }
     }
 
     /// Compile and write all configuration files to the configuration directory.
@@ -466,47 +476,46 @@ impl CfgRenderer {
                     String::new()
                 }
             };
-            if file_hash.is_empty() {
+            changed |= if file_hash.is_empty() {
                 debug!(
                     "Configuration {} does not exist; restarting",
                     cfg_dest.display()
                 );
 
-                create_directory_structure(
+                ensure_directory_structure(
                     render_path.as_ref(),
                     &cfg_dest,
                     &pkg.svc_user,
                     &pkg.svc_group,
                 )?;
-                write_templated_file(&cfg_dest, compiled, &pkg.svc_user, &pkg.svc_group)?;
+                write_templated_file(&cfg_dest, &compiled, &pkg.svc_user, &pkg.svc_group)?;
                 outputln!(
                     preamble service_group_name,
                     "Created configuration file {}",
                     cfg_dest.display()
                 );
 
-                changed = true
+                true
             } else if file_hash == compiled_hash {
                 debug!(
                     "Configuration {} {} has not changed; not restarting.",
                     cfg_dest.display(),
                     file_hash
                 );
-                continue;
+                false
             } else {
                 debug!(
                     "Configuration {} has changed; restarting",
                     cfg_dest.display()
                 );
-                write_templated_file(&cfg_dest, compiled, &pkg.svc_user, &pkg.svc_group)?;
+                write_templated_file(&cfg_dest, &compiled, &pkg.svc_user, &pkg.svc_group)?;
                 outputln!(
                     preamble service_group_name,
                     "Modified configuration file {}",
                     cfg_dest.display()
                 );
-
-                changed = true;
-            }
+                true
+            };
         }
         Ok(changed)
     }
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/service/config.rs
```diff
@@ -597,23 +591,40 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
 }
 
 #[cfg(not(windows))]
-fn set_permissions<T: AsRef<Path>>(path: T, pkg: &Pkg) -> hcore::error::Result<()> {
+fn set_permissions<T: AsRef<Path>>(
+    path: T,
+    user: &String,
+    group: &String,
+) -> hcore::error::Result<()> {
```
christophermaier: This new signature is a good improvement 👍

components/sup/src/manager/service/config.rs
```diff
@@ -1085,6 +1109,38 @@ mod test {
...
+        let tmp = TempDir::new().expect("create temp dir");
+        let template_dir = tmp.path().join("output");
+        fs::create_dir_all(&template_dir).expect("create output dir");
+
+        let file = template_dir.join("foo/config.cfg");
+        let contents = "foo\nbar\n";
+
+        assert_eq!(file.exists(), false);
+        create_directory_structure(&template_dir, &file).expect("create output dir structure");
+        create_directory_structure(&template_dir, &file).expect("create output dir structure");
```
christophermaier: Is this duplicated line a typo?

components/sup/src/manager/service/config.rs
```diff
@@ -616,6 +617,29 @@ fn set_permissions<T: AsRef<Path>>(
...
+
+    let dir = file.as_ref().parent().unwrap();
+    if dir.exists() {
+        Ok(())
+    } else {
+        std::fs::create_dir_all(&dir)?;
+
+        // TODO: set permissions
+        Ok(())
+    }
```
christophermaier: Maybe this instead?

```rust
if !dir.exists() {
    // do stuff
}
Ok(())
```

jamesc:
@christophermaier This is ready for review now

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,156 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Helper function to construct a new PathBuf for each level of subdir we
+/// iterate into.
+fn subdir_path(context: &PathBuf, entry: &std::fs::DirEntry) -> PathBuf {
+    let mut subdir = context.clone();
```
mwrock: perhaps the context should be mutable to avoid the clones

jamesc: @mwrock I need to do the `clone`s as the context is different for each subdirectory I recurse into.  I could rewrite to not use a `PathBuf` and pass e.g. a vec! of components, which would mean I don't have to `clone()` 

My feeling was that  there shouldn't be that many config files and it only happens on svc load so not a recurring hit.

mwrock: oh right i see. yeah this is pretty micro and not worth a refactor IMO

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,156 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
     use crate::hcore::util::posix_perm;
 
     if users::can_run_services_as_svc_user() {
-        posix_perm::set_owner(path.as_ref(), &pkg.svc_user, &pkg.svc_group)?;
+        posix_perm::set_owner(path.as_ref(), &user, &group)?;
+    }
+    if path.as_ref().is_dir() {
+        posix_perm::set_permissions(path.as_ref(), CONFIG_DIR_PERMISSIONS)
+    } else {
+        posix_perm::set_permissions(path.as_ref(), CONFIG_PERMISSIONS)
```
christophermaier: Since the activity here is the same, but the argument is different, could we have the `if`/`else` block determine the permissions to use, and then we pass that into a single `set_permissions` call?

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,156 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+    T: AsRef<Path>,
+{
+    let mut template = TemplateRenderer::new();
+    if dir.as_ref().is_dir() {
+        load_templates_recurse(&dir, &PathBuf::new(), &mut template)?;
+    } else {
+        debug!(
+            "Configuration directory is not a directory: {:?}",
+            dir.as_ref()
+        );
```
christophermaier: Is there any danger / downside to having this _not_ be an error? The documentation string implies that being a directory is a necessary precondition.

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,156 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+    Ok(template)
+}
+
+/// Helper function for `load_templates` to recurse into the directory structure.
+///
+/// `dir` should be a directory
+fn load_templates_recurse<T>(
+    dir: T,
+    context: &PathBuf,
+    mut template: &mut TemplateRenderer,
```
christophermaier: I wonder if it'd be easier to test / reason about if this function were to return just the paths for all the templates, and then up in `load_templates`, we take that list and add them all to the `TemplateRenderer` (perhaps `TemplateRenderer` could even have a constructor that takes the templates.)

components/common/src/templating/config.rs
christophermaier: Can should be able to pass a `&Path` instead. When you `join` to a `Path`, you'll get a new, owned`PathBuf`. When you join to a `PathBuf`, you get a totally brand-new `PathBuf`, so by using a `&Path` here, you cut the allocations in half.

By taking advantage of that, I think you can get rid of this function altogether. When you're recurring on a directory, you just pass a `Path` reference to your `PathBuf`.

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,156 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+///
+fn create_directory_structure<T>(root: T, file: T, user: &str, group: &str) -> Result<()>
+where
+    T: AsRef<Path>,
+{
+    // We check that `file` is below `root` in the directory structure and
+    // that `root` exists, so that we don't create arbitrary directory
+    // structures on disk
+    assert!(root.as_ref().is_dir());
+    assert!(file.as_ref().starts_with(root.as_ref()));
```
christophermaier: What's the overall effect of a panic being thrown here? Should this be an error instead.

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,156 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+    let mut subdir = context.clone();
+    subdir.push(entry.file_name());
+    subdir
+}
+
+/// Create the appropriate directories between a `root` directory
+/// and a file within that directory structure that we're about
+/// to create.
+///
+fn create_directory_structure<T>(root: T, file: T, user: &str, group: &str) -> Result<()>
```
christophermaier: Naming this something like `ensure_directory_structure` might be a little more reflective of the intention of this function.

components/common/src/templating/config.rs
jamesc: +1.  That's what I was looking for. Thanks!

components/common/src/templating/config.rs
jamesc: +1.  That's what I was looking for. Thanks!

baumanj:
Is this ready for review? I see the "WIP" is still in the title.

components/common/src/templating/config.rs
jamesc: The panic should never happen as we control all access into this function.  It was pretty helpful while refactoring this code and actually caught a bug. I can probably remove it once the code is complete.

components/common/src/templating/config.rs
jamesc: No downside - I'll throw an error

jamesc:
> Is this ready for review? I see the "WIP" is still in the title.

Fixed :)  Thanks - missed that. Yep, it's ready for review.  I've one outstanding change from @christophermaier review to do still.

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,150 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+    // that `root` exists, so that we don't create arbitrary directory
+    // structures on disk
+    assert!(root.as_ref().is_dir());
+    assert!(file.as_ref().starts_with(root.as_ref()));
+
+    let dir = file.as_ref().parent().unwrap();
+
+    if !dir.exists() {
+        std::fs::create_dir_all(&dir)?;
+        for anc in dir.ancestors() {
```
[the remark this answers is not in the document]
jamesc: nice!  I like this.

components/common/src/templating/config.rs
jamesc: Added docs comments. Left in the asserts

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,150 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+) -> Result<()>
+where
+    T: AsRef<Path>,
+{
+    if let Ok(entries) = std::fs::read_dir(dir) {
+        for entry in entries {
+            if let Ok(entry) = entry {
+                // Skip any entries in the template directory which aren't files or directories
+                match entry.file_type() {
+                    Ok(file_type) => {
```
baumanj: And even one more (plus this obviates the comment in L625 with a useful log message):
```rust
        match entry.file_type() {
            Ok(file_type) if file_type.is_file() => {
                ...
            }
            Ok(file_type) if file_type.is_dir() => {
                ...
            }
            Ok(file_type) => trace!("Skipping non file/directory entry: {:?}", file_type),
            Err(e) => debug!("Failed to get file metadata for {:?} : {}", entry, e),
        }
```

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,150 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+                        if file_type.is_file() {
+                            // We're storing the pathname relative to the input config directory
+                            // as the identifier for the template
+                            let name = entry.file_name().to_string_lossy().into_owned();
+                            let relative_path = context.join(&name);
```
baumanj: The previous line is unnecessary if you make this simplification
```suggestion
                            let relative_path = context.join(entry.file_name());
```

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,150 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+
+                            // JW TODO: This error needs improvement. TemplateFileError is too generic.
+                            template
+                                .register_template_file(
+                                    &relative_path.to_string_lossy(),
+                                    &entry.path(),
+                                )
+                                .map_err(Error::TemplateFileError)?;
```
baumanj: If we were ever going a lot of template processing, it might behoove us to avoid allocating all these new strings:
```rust
                if let Some(relative_path) = relative_path.to_str() {
                    template
                        .register_template_file(relative_path, &entry.path())
                        .map_err(Error::TemplateFileError)?;
                } else {
                    // Not sure whether we'd want to throw an error or just log something; should be very rare
                }
```

components/common/src/templating/config.rs
```diff
@@ -588,34 +565,150 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+                        } else if file_type.is_dir() {
+                            load_templates_recurse(
+                                &entry.path(),
+                                &context.join(&entry.file_name()),
```
baumanj: Probably better to move the duplication before the match on L626
```rust
        let relative_path = context.join(entry.file_name());
        match entry.file_type() {

```
Which simplifies both the `is_file` case and lets this be
```suggestion
                                &relative_path,
```

components/common/src/templating/config.rs
christophermaier: even better

jamesc: Yep, removed the `if let` that was dropping that error and this looks better.

components/common/src/templating/config.rs
```diff
@@ -495,16 +472,19 @@ impl CfgRenderer {
...
+                write_templated_file(&cfg_dest, compiled, &pkg.svc_user, &pkg.svc_group)?;
                 outputln!(
                     preamble service_group_name,
                     "Created configuration file {}",
                     cfg_dest.display()
                 );
 
-                set_permissions(&cfg_dest, pkg)?;
-
                 changed = true
```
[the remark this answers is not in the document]
jamesc: I like this.  I had wanted to come back and refactor this anyway

components/common/src/templating/config.rs
jamesc: So it looks like we need to keep an `is_dir()` check somewhere as it's not required for a package to have a `config` directory.  I've moved it up to the the `CfgRenderer::new()` so it's more apparent when we're doing and keeps functions down the call chain simpler

components/common/src/templating/config.rs
```diff
@@ -587,35 +573,109 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+                    .register_template_file(&relative_path.to_string_lossy(), &entry.path())
+                    .map_err(Error::TemplateFileError)?;
+            }
+            Ok(file_type) if file_type.is_dir() => {
+                template = load_templates(&entry.path(), &relative_path, template)?
+            }
+            Ok(file_type) => trace!("Skipping non file/directory entry: {:?}", file_type),
+            Err(e) => {
+                debug!("Failed to get file metadata for {:?} : {}", entry, e);
+                continue;
```
baumanj: I think this `continue` is superfluous. Not a big deal, but eliminating it might obviate the braces for the `match` arm and save a couple lines.

components/common/src/templating/config.rs
```diff
@@ -587,35 +573,109 @@ fn is_toml_value_a_table(key: &str, table: &toml::value::Table) -> bool {
...
+    posix_perm::set_permissions(&path, permissions)
 }
 
 #[cfg(windows)]
-fn set_permissions<T: AsRef<Path>>(path: T, _pkg: &Pkg) -> hcore::error::Result<()> {
+fn set_permissions(path: &Path, _user: &str, _group: &str) -> hcore::error::Result<()> {
     use crate::hcore::util::win_perm;
 
-    win_perm::harden_path(path.as_ref())
+    win_perm::harden_path(&path)
```
baumanj: This is already a reference
```suggestion
    win_perm::harden_path(path)
```

components/common/src/templating/config.rs
jamesc: `to_string_lossy()` returns a `Cow<str>`.  I thought that in the normal case where `relative_path` is a UTF-8 string then there would be no allocation ?  Did I misunderstand how `Cow` works ?

baumanj: After digging through the stdlib sources, you're totally right. I was getting thrown off by the convention that functions with `to` tend to allocate, whereas `as` don't.

That said, I'm not really certain if the path not being valid UTF-8 would work correctly in our system. Can you think of a situation where this would occur? If so, it's worth a test to see that this implementation would handle it, otherwise, I think throwing an error (or even just panicking) would be appropriate.

That said, this is a pretty edge case, and I don't think we have a thoughtful or consistent approach to handling the UTF-8 conversions errors generally, so if you left this as is, I think that's understandable.

jamesc: Reading around, I'm think this is a non-issue.  All major linux distros default to UTF-8 and it should be a problem on Windows. The only case I can think of is that someone creates a package on a non UTF-8 locale system with config filenames encoded using something other than UTF-8.  Then the user of the package runs it on a system with UTF-8 encoding. We'd end up with templates filenames which are different than the original file names.  But the probability of this all seems vanishingly small and I'm not sure some other part of the supervisor would not have broken first.

I'm for leaving this as-is as it was the behavior before this change, so we're not breaking anyone who might have relied on that behavior.  Also this doesn't seem to be the place to enforce UTF-8 filenames first in our code base.  If we made that decision then we should enforce in studio first and then it'd make sense to do a panic here as well.

baumanj:
For the travis failure, I'd suggest cherry-picking [my fix](https://github.com/habitat-sh/habitat/pull/6101/commits/0d152056d4a91c99af8a293a19909f4ef2ea755e). We've seen this on a number of branches, but it makes zero sense to me why.

PR #6068, opened 2019-01-17
Fix some clippy lints that were missed due to `make lint` missing some files

baumanj:
Notably launcher and pkg-export-helm weren't covered, but now that's fixed. Additionally, the linting should run a bit faster and less redundantly now that it's a single command.

Also fix an over-correction from the previous PR
See https://github.com/habitat-sh/habitat/pull/6065#discussion_r248709400


thesentinels:

Thank you for contributing!

Makefile
```diff
@@ -274,11 +274,11 @@ DENIED_LINTS = clippy::assign_op_pattern \
 
 define LINT
 lint-$1: image ## executes the $1 component's linter checks
-	$(run) sh -c 'cd components/$1 && cargo clippy --all-targets --tests $(CARGO_FLAGS) -- \
-	                                               $(addprefix -A ,$(UNEXAMINED_LINTS)) \
-	                                               $(addprefix -A ,$(ALLOWED_LINTS)) \
-	                                               $(addprefix -W ,$(LINTS_TO_FIX)) \
-	                                               $(addprefix -D ,$(DENIED_LINTS))'
+	$(run) cargo clippy --all-targets --tests $(CARGO_FLAGS) -- \
```
jamesc: Now that you only run one clippy command over all the code, why do we still have the `lint-$1` targets.  It looks like this now runs the same `cargo clippy` command multiple times for each `lint-*` target.

baumanj: Good call.

I'll fix this up. The different targets should pass `--"$COMPONENTS"` ([from .travis.yml](https://github.com/habitat-sh/habitat/blob/master/.travis.yml#L37)) such that they run over just the binary or library code.

The benefit in removing the `sh -c 'cd components/$1` is that we apply this to everything in the workspace in one fell swoop rather that doing a bunch of redundant work and skipping things like the launcher.

baumanj: Addressed this feedback with this commit https://github.com/habitat-sh/habitat/pull/6068/commits/487b9419b664af73484c418785a5e569b50175cf

PR #6070, opened 2019-01-17
Put integration tests behind cargo feature flag

scotthain:

⚠️ This will effectively disable these tests from running on every PR, which we may not want to do at this point in time. The tests can be run with `cargo test --feature integration_tests` which should be run prior to release but not necessarily during verification. (This also may be a good time to discuss the value/risk and the line between verification/unit and integration tests)

### DIFF
--- a/components/butterfly/Cargo.toml
+++ b/components/butterfly/Cargo.toml
@@ -46,3 +46,4 @@ tempfile = "*"
 [features]
 default = ["protocols"]
 protocols = []
+integration_tests = []
--- a/components/butterfly/tests/encryption/mod.rs
+++ b/components/butterfly/tests/encryption/mod.rs
@@ -18,6 +18,7 @@ use habitat_core::crypto::keys::sym_key::SymKey;
 use crate::btest;
 
 #[test]
+#[cfg_attr(not(feature = "integration_tests"), ignore)]
 fn symmetric_encryption_of_wire_payloads() {
     let ring_key = SymKey::generate_pair_for_ring("wolverine")
         .expect("Failed to generate an in memory symkey");
--- a/components/butterfly/tests/integration.rs
+++ b/components/butterfly/tests/integration.rs
@@ -23,6 +23,7 @@ mod rumor;
 use habitat_butterfly::member::Health;
 
 #[test]
+#[cfg_attr(not(feature = "integration_tests"), ignore)]
 fn two_members_meshed_confirm_one_member() {
     let mut net = btest::SwimNet::new(2);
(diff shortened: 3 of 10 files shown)


thesentinels:

Thank you for contributing!

baumanj:
I need to ponder this a bit. I think we *do* want all tests running for all PRs because feedback on breakage at the time the change is being made (rather than just before release) makes fixes much easier.

If these tests are flaky or need special parameters to run reliably, I think we should address that, but I'm not comfortable with just turning them off.

scotthain:
@baumanj that is exactly why I wanted to get some discussion going on! I'm also not comfortable turning these off right now but I want to make sure we're all on the same page regardless of what we decide. (and also what is/isn't important to fix right now vs later)

christophermaier:
Per external discussion, it appears that some tests are making assumptions about system resources that our testing infrastructure doesn't satisfy (because test nodes are intentionally locked down).

A lot of things that currently exist as "integration" tests would be better formulated as unit tests (and vice versa), so there's likely some work that can be done to reduce the number and kinds of system resources (like ports) that are required for these tests.

Using attributes to flag such offending tests seems like a good way to help us triage things and figure out which tests we'd be well-served to refactor first, or to figure out what kind of "special" testing node we might want to set up to run those tests on for the near term.

To that end, it might be better to name the feature to reflect this... that is, instead of calling it "integration_tests", call it something like "does_not_run_in_buildkite" (I'm not good at naming). There could be integration test that we have today that _do_ run just fine in our test infrastructure, and if that's the case, we should keep running them. It might also be good name it in the negative, so we could call it like `cargo test --features=without_buildkite_problems`, and have the attribute declarations set to `ignore` like you have them currently.

Also, I think you can use the `!` syntax to apply an attribute to all the tests in a module, rather than one-by-one (i.e. use `#![cfg_attr(...)]` at the top-level inside a module), but if some tests in a module work and others don't, that might be more bother than it's worth.

christophermaier:
I do agree with @baumanj that we want these particular tests to run for each PR. We can set up bigger, more involved tests that run at a later stage, after we've built packages, but these are good to run as often as we can. With the parallelism we've got for test runners, they won't be adding any appreciable time to the overall test run (and even without the parallelism, they're not excessively time-consuming).

baumanj:
Can we close this? It looks like the conversation has shifted over to https://github.com/habitat-sh/habitat/pull/6085 unless I'm misunderstanding.

scotthain:
Yes we can @baumanj sorry for the confusion

PR #6074, opened 2019-01-18
Replace Strings and &strs with type-safe habitat_core::ChannelIdent type

baumanj:
See https://github.com/habitat-sh/core/pull/107

### DIFF
--- a/components/hab/src/command/pkg/demote.rs
+++ b/components/hab/src/command/pkg/demote.rs
@@ -28,7 +28,7 @@
 
 use crate::api_client::Client;
 use crate::common::ui::{Status, UIWriter, UI};
-use crate::hcore::package::PackageIdent;
+use crate::hcore::{package::PackageIdent, ChannelIdent};
 
 use crate::error::{Error, Result};
 use crate::{PRODUCT, VERSION};
@@ -44,17 +44,17 @@ pub fn start(
     ui: &mut UI,
     bldr_url: &str,
     ident: &PackageIdent,
-    channel: &str,
+    channel: &ChannelIdent,
     token: &str,
 ) -> Result<()> {
     let api_client = Client::new(bldr_url, PRODUCT, VERSION, None)?;
 
     ui.begin(format!("Demoting {} from {}", ident, channel))?;
 
-    if channel == "unstable" {
+    if channel == &ChannelIdent::unstable() {
         return Err(Error::CannotRemoveFromChannel((
             format!("{}", ident),
-            "unstable".to_string(),
+            channel.to_string(),
         )));
     }
 
--- a/components/hab/src/command/pkg/demote.rs
+++ b/components/hab/src/command/pkg/demote.rs
@@ -53,7 +53,7 @@ pub fn start(
 
     if channel == &ChannelIdent::unstable() {
         return Err(Error::CannotRemoveFromChannel((
-            format!("{}", ident),
+            ident.to_string(),
             channel.to_string(),
         )));
     }
--- a/components/hab/src/exec.rs
+++ b/components/hab/src/exec.rs
@@ -17,11 +17,11 @@ use std::path::{Path, PathBuf};
 use crate::common;
 use crate::common::command::package::install::{InstallHookMode, InstallMode, LocalPackageUsage};
 use crate::common::ui::{Status, UIWriter, UI};
-use crate::hcore::env as henv;
+use crate::hcore;
 use crate::hcore::fs::{self, cache_artifact_path, FS_ROOT_PATH};
 use crate::hcore::package::{PackageIdent, PackageInstall, PackageTarget};
 use crate::hcore::url::default_bldr_url;
-use crate::hcore::{self, channel};
+use crate::hcore::ChannelIdent;
 
 use crate::error::{Error, Result};
 use crate::{PRODUCT, VERSION};
@@ -111,9 +111,9 @@ where
 
 /// Determine the channel from which to install Habitat-specific
 /// packages.
-fn internal_tooling_channel() -> String {
-    match henv::var(INTERNAL_TOOLING_CHANNEL_ENVVAR) {
+fn internal_tooling_channel() -> ChannelIdent {
+    match ChannelIdent::from_env_var(INTERNAL_TOOLING_CHANNEL_ENVVAR) {
         Ok(channel) => channel,
-        Err(_) => channel::STABLE_CHANNEL.to_string(),
+        Err(_) => ChannelIdent::stable(),
     }
 }
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -1623,8 +1623,8 @@ fn bldr_url_from_input(m: &ArgMatches<'_>) -> Option<String> {
 }
 
 /// A channel name, but *only* if the user specified via CLI args.
-fn channel_from_input(m: &ArgMatches<'_>) -> Option<String> {
-    m.value_of("CHANNEL").and_then(|c| Some(c.to_string()))
+fn channel_from_input(m: &ArgMatches<'_>) -> Option<ChannelIdent> {
+    m.value_of("CHANNEL").map(ChannelIdent::from)
 }
 
 /// If the user provides both --application and --environment options,
@@ -1765,7 +1765,7 @@ fn ui() -> UI {
 /// populates all *shared* options between `run` and `load`.
 fn update_svc_load_from_input(m: &ArgMatches<'_>, msg: &mut protocol::ctl::SvcLoad) -> Result<()> {
     msg.bldr_url = bldr_url_from_input(m);
-    msg.bldr_channel = channel_from_input(m);
+    msg.bldr_channel = channel_from_input(m).map(ChannelIdent::into);
     msg.application_environment = get_app_env_from_input(m)?;
     msg.binds = get_binds_from_input(m)?;
     if m.is_present("FORCE") {
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -1622,11 +1622,6 @@ fn bldr_url_from_input(m: &ArgMatches<'_>) -> Option<String> {
         .or_else(bldr_url_from_env)
 }
 
-/// A channel name, but *only* if the user specified via CLI args.
-fn channel_from_input(m: &ArgMatches<'_>) -> Option<ChannelIdent> {
-    m.value_of("CHANNEL").map(ChannelIdent::from)
-}
-
 /// If the user provides both --application and --environment options,
 /// parse and set the value on the spec.
 fn get_app_env_from_input(m: &ArgMatches<'_>) -> Result<Option<ApplicationEnvironment>> {
@@ -1765,7 +1760,7 @@ fn ui() -> UI {
 /// populates all *shared* options between `run` and `load`.
 fn update_svc_load_from_input(m: &ArgMatches<'_>, msg: &mut protocol::ctl::SvcLoad) -> Result<()> {
     msg.bldr_url = bldr_url_from_input(m);
-    msg.bldr_channel = channel_from_input(m).map(ChannelIdent::into);
+    msg.bldr_channel = m.value_of("CHANNEL").map(str::to_string);
     msg.application_environment = get_app_env_from_input(m)?;
     msg.binds = get_binds_from_input(m)?;
     if m.is_present("FORCE") {
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -47,7 +47,6 @@ use crate::common::command::package::install::{
 use crate::common::types::ListenCtlAddr;
 use crate::common::ui::{Coloring, Status, UIWriter, NONINTERACTIVE_ENVVAR, UI};
 use crate::hcore::binlink::default_binlink_dir;
-use crate::hcore::channel;
 #[cfg(windows)]
 use crate::hcore::crypto::dpapi::encrypt;
 use crate::hcore::crypto::keys::PairType;
@@ -57,6 +56,7 @@ use crate::hcore::fs::{
     cache_analytics_path, cache_artifact_path, cache_key_path, launcher_root_path,
 };
 use crate::hcore::package::{PackageIdent, PackageTarget};
+use crate::hcore::ChannelIdent;
 use clap::{ArgMatches, Shell};
 use futures::prelude::*;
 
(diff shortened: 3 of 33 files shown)


components/pkg-export-docker/src/lib.rs
```diff
@@ -237,7 +237,7 @@ pub fn export_for_cli_matches(
     ui: &mut UI,
     matches: &clap::ArgMatches<'_>,
 ) -> Result<Option<DockerImage>> {
-    let default_channel = channel::default();
+    let default_channel = hcore::ChannelIdent::default().to_string();
```
christophermaier: Similarly, can we push whatever is consuming `default_channel` to take a `ChannelIdent`, rather than a `String`?

components/pkg-export-tar/src/lib.rs
```diff
@@ -35,7 +37,7 @@ pub const VERSION: &str = include_str!(concat!(env!("OUT_DIR"), "/VERSION"));
 const BUSYBOX_IDENT: &str = "core/busybox-static";
 
 pub fn export_for_cli_matches(ui: &mut UI, matches: &clap::ArgMatches<'_>) -> Result<()> {
-    let default_channel = channel::default();
+    let default_channel = ChannelIdent::default().to_string();
```
christophermaier: Ditto.

components/sup/src/main.rs
```diff
@@ -170,9 +170,10 @@ fn sub_run(m: &ArgMatches, launcher: LauncherCli) -> Result<()> {
                         .as_ref()
                         .unwrap_or(&*protocol::DEFAULT_BLDR_URL),
                     &source,
-                    msg.bldr_channel
+                    &msg.bldr_channel
                         .as_ref()
-                        .unwrap_or(&*protocol::DEFAULT_BLDR_CHANNEL),
+                        .map(|s| ChannelIdent::from(s))
+                        .unwrap_or_default(),
```
christophermaier: This is another instance where I don't think we want to let an environment variable value sneak in.

components/sup/src/main.rs
```diff
@@ -341,13 +342,13 @@ fn bldr_url_from_input(m: &ArgMatches) -> Option<String> {
 
 /// Resolve a channel. Taken from CLI args, or (failing that), a
 /// default value.
-fn channel(matches: &ArgMatches) -> String {
-    channel_from_input(matches).unwrap_or_else(channel::default)
+fn channel(matches: &ArgMatches) -> ChannelIdent {
+    channel_from_input(matches).unwrap_or_default()
```
christophermaier: This feels like it wants to use `EnvConfig`

components/sup/src/main.rs
```diff
@@ -497,7 +498,7 @@ fn ui() -> UI {
 /// populates all *shared* options between `run` and `load`.
 fn update_svc_load_from_input(m: &ArgMatches, msg: &mut protocol::ctl::SvcLoad) -> Result<()> {
     msg.bldr_url = Some(bldr_url(m));
-    msg.bldr_channel = Some(channel(m));
+    msg.bldr_channel = Some(channel(m).to_string());
```
christophermaier: Can `SvcLoad` handle converting a `ChannelIdent` into a string when it gets converted into a Protobuf? That'd help to keep the "string"ness of channels at the boundaries of the system, while keeping only `ChannelIdent` in the interior.

components/sup/src/manager/commands.rs
```diff
@@ -226,14 +227,12 @@ pub fn service_load(
     opts: protocol::ctl::SvcLoad,
 ) -> NetResult<()> {
     let ident: PackageIdent = opts.ident.clone().ok_or_else(err_update_client)?.into();
-    let bldr_url = opts
-        .bldr_url
-        .clone()
-        .unwrap_or_else(|| protocol::DEFAULT_BLDR_URL.to_string());
+    let bldr_url = opts.bldr_url.clone().unwrap_or_default();
```
christophermaier: Was this a mistake? I don't think we need to be modifying url stuff when making this channel change.

components/sup/src/manager/commands.rs
```diff
@@ -298,7 +297,7 @@ pub fn service_load(
                         req,
                         &source,
                         &service_spec.bldr_url,
-                        &service_spec.channel,
+                        &ChannelIdent::from(&service_spec.channel),
```
christophermaier: For consistency, `ServiceSpec` should return a `ChannelIdent`, rather than a String.

components/sup/src/manager/mod.rs
```diff
@@ -164,7 +164,7 @@ impl Default for ManagerConfig {
             custom_state_path: None,
             eventsrv_group: None,
             update_url: "".to_string(),
-            update_channel: "".to_string(),
+            update_channel: ChannelIdent::default(),
```
christophermaier: Somewhat tangentially, I think we can just derive `Default` for `ManagerConfig` now.

(Though perhaps getting a default update URL of something other than `""` should come before that 😬 )

components/sup/src/manager/mod.rs
```diff
@@ -221,7 +221,6 @@ pub struct Manager {
     pub state: Arc<ManagerState>,
     butterfly: butterfly::Server,
     census_ring: CensusRing,
-    events_group: Option<ServiceGroup>,
```
christophermaier: We'll want to rebase before merging, so we get these eventsrv-related changes naturally.

components/sup/src/manager/service/mod.rs
```diff
@@ -92,7 +92,7 @@ enum BindStatus<'a> {
 pub struct Service {
     pub service_group: ServiceGroup,
     pub bldr_url: String,
-    pub channel: String,
+    pub channel: String, // TODO: change to hcore::ChannelIdent
```
christophermaier: Can we do this now, or was there some kind of gnarlier refactoring it entails?

components/sup/src/manager/service/spec.rs
```diff
@@ -417,7 +417,7 @@ impl Default for ServiceSpec {
             group: DEFAULT_GROUP.to_string(),
             application_environment: None,
             bldr_url: DEFAULT_BLDR_URL.to_string(),
-            channel: STABLE_CHANNEL.to_string(),
+            channel: ChannelIdent::stable().to_string(),
```
christophermaier: `String` -> `ChannelIdent`

christophermaier: Also, just look at all those other default strings just begging to be refactored to types 😄

components/sup/src/util/pkg.rs
```diff
@@ -52,7 +53,7 @@ where
         // function, though, should always have a channel. We should
         // push this "Option-ness" as far down the stack as we can,
         // with the ultimate goal of eliminating it altogether.
-        Some(channel),
+        Some(&channel),
```
christophermaier: 🤔 How far out are we from getting rid of this `Option` and having everything take a channel?

cc: @chefsalim

components/sup/src/manager/service/mod.rs
baumanj: It seemed like a rabbit hole for questionable benefit. I'd like to get the initial change in and then try running down some of the TODOs so we can make the important progress (replacing all the misleading defaults) first.

baumanj:
Rebased

components/hab/src/exec.rs
```diff
@@ -110,9 +110,9 @@ where
 
 /// Determine the channel from which to install Habitat-specific
 /// packages.
-fn internal_tooling_channel() -> String {
-    match henv::var(INTERNAL_TOOLING_CHANNEL_ENVVAR) {
+fn internal_tooling_channel() -> ChannelIdent {
+    match ChannelIdent::from_env_var(INTERNAL_TOOLING_CHANNEL_ENVVAR) {
```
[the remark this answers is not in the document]
baumanj: I want to do this in a controlled way to minimize the confusion and keep the scope under control. In step 1, I want to just introduce the new `ChannelIdent` type with the minimal changes to support it. I'll save moving things around between repos for a follow up.

components/hab/src/main.rs
```diff
@@ -496,10 +496,10 @@ fn sub_pkg_export(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?;
     let format = &m.value_of("FORMAT").unwrap();
     let url = bldr_url_from_matches(&m)?;
-    let channel = m
+    let channel: ChannelIdent = m
         .value_of("CHANNEL")
-        .map(str::to_string)
-        .unwrap_or_else(channel::default);
+        .map(ChannelIdent::from)
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
baumanj: See https://github.com/habitat-sh/habitat/pull/6074#discussion_r251517957

components/pkg-export-docker/src/build.rs
```diff
@@ -370,7 +373,7 @@ impl<'a> BuildSpec<'a> {
         let package_install = common::command::package::install::start(
             ui,
             url,
-            Some(channel),
+            Some(&ChannelIdent::from(channel)),
```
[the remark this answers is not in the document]
baumanj: Yeah, I was trying to figure out which places to change were a good return on investment and not waste too much time where the type wasn't valuable. I think I initially held off here because the internal types had `&str` fields and they can't be converted to `&ChannelIdents` because it requires an owned `String`. We could get fancy and change `ChannelIdent` to use `Cow` internally, but I think that's making things more complicated than necessary, so [I've just changed the `&str` fields to owned `ChannelIdent`s](https://github.com/habitat-sh/habitat/pull/6074/commits/4c12c3c93a0fc8a63cfb85ae966627ef65b0afdf).

components/pkg-export-docker/src/lib.rs
baumanj: This is obviated by removing the `default_channel` argument from `BuildSpec::new_from_cli_matches`. It was only ever used to return the `ChannelIdent::default()`.

components/pkg-export-tar/src/lib.rs
baumanj: Fixed now

components/sup/src/main.rs
baumanj: Will add to the list for follow up

components/sup/src/main.rs
baumanj: Will add to the list for follow up

components/sup/src/main.rs
baumanj: Good idea, added in https://github.com/habitat-sh/habitat/pull/6074/commits/cbebb204f2ee3b0b2cafe139cbf99ed1baf73d94

components/sup/src/manager/commands.rs
baumanj: Yeah, I'm not sure how that got in there. Reverted in https://github.com/habitat-sh/habitat/pull/6074/commits/8eee5cee3788d6f029ec591ec7394e98ef008ae8

components/sup/src/manager/mod.rs
baumanj: Done

components/sup/src/manager/service/mod.rs
baumanj: I sucked it up and just did it: https://github.com/habitat-sh/habitat/pull/6074/commits/cbebb204f2ee3b0b2cafe139cbf99ed1baf73d94#diff-320293bb5f68f709add66ac43f5ea52eR103

components/sup/src/manager/service/spec.rs
baumanj: [Done](https://github.com/habitat-sh/habitat/pull/6074/commits/cbebb204f2ee3b0b2cafe139cbf99ed1baf73d94#diff-67161b6446fbddc099ba24f430c21d4bR420)

components/sup/src/util/pkg.rs
baumanj: I'll leave that one to @chefsalim

components/sup/src/manager/commands.rs
baumanj: Changed the `channel` type in `ServiceSpec` in https://github.com/habitat-sh/habitat/pull/6074/commits/cbebb204f2ee3b0b2cafe139cbf99ed1baf73d94#diff-67161b6446fbddc099ba24f430c21d4bR283, allowing this [to be simplified](https://github.com/habitat-sh/habitat/pull/6074/commits/0a5cdd959beebcade60238a09d13bf21ea9a5c2e) to `&service_spec.channel`

components/sup/src/manager/mod.rs
baumanj: I'll leave that one to you as my list of follow on tasks is significant 😂

components/sup-protocol/protocols/ctl.proto
```diff
@@ -82,7 +82,7 @@ message SvcLoad {
   // Remote http URL for the Builder service to receive package updates from.
   optional string bldr_url = 6;
   // Remote channel on the Builder service to receive package updates from.
-  optional string bldr_channel = 7;
+  optional sup.types.ChannelIdent bldr_channel = 7;
```
christophermaier: Hrmm... I'm not sure that we can it this way. The [docs](https://developers.google.com/protocol-buffers/docs/proto#updating) imply that you can change strings to messages if the string contains an encoding of the message, which I don't think is the case here.

I think you might run into issues if your `hab-sup` and `hab` binaries aren't in lockstep.

You might be able to have a new field, called simply `channel`, and then handle the fallback conversion logic when the protobuf is consumed.

If this gets too hairy, this might be better as a separate PR.

baumanj: That sounds like a rather large amount of effort for negligible returns. The big thing motivating the change initially is wrapping the `String` so that we can have a channel-specific `Default`. However, the necessity of another protobuf-specific type to also wrap the `String`, but which we *can't` make have a sensible default seems potentially more confusing. Maybe it would be different if we had `ChannelIdent` when the .proto was created, but as is, I think we should leave it as `string bldr_channel` and do the conversion when getting data in and out of `SvcLoad`. That seems to me just as valid a frontier.

christophermaier: Sounds good... as long as we don't break the protocol, and the logic for going from Protobuf into our internal types (and vice versa) is centralized, I'm 👍

baumanj: https://github.com/habitat-sh/habitat/pull/6074/commits/5e6d7914111baa0b7b4a9917b1a10d280b8b8343 should remove this

baumanj:
I've fixed the protobuf stuff and [added ChannelIdent to even more places](url
https://github.com/habitat-sh/habitat/pull/6074/commits/c613a235dfd84f975f9f82b5cb807d608a49d001). When you get a chance, let me know if there's anything else you'd like me to change, @christophermaier.

components/hab/src/command/pkg/demote.rs
```diff
@@ -44,17 +44,17 @@ pub fn start(
...
     token: &str,
 ) -> Result<()> {
     let api_client = Client::new(bldr_url, PRODUCT, VERSION, None)?;
 
     ui.begin(format!("Demoting {} from {}", ident, channel))?;
 
-    if channel == "unstable" {
+    if channel == &ChannelIdent::unstable() {
         return Err(Error::CannotRemoveFromChannel((
             format!("{}", ident),
```
christophermaier: Might want to change this to `ident.to_string()`... odd that Clippy isn't flagging that 🤔

baumanj: My guess is that it has to do with crossing the crate boundary, but good catch, will fix.

PR #6077, opened 2019-01-23
Some code cleanups after #5866

jamesc:
These are some misc cleanups after #5866 was merged

* Delete unneeded file
* remove some unused exports
* Update Makefile so tests have `TESTING_FS_ROOT` set.  This stops the tests writing to `/hab/svc/test` during a unit test run (or failing if there is no permission to write there)

### DIFF
--- a/support/ci/rust_tests.sh
+++ b/support/ci/rust_tests.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-sudo mkdir -p /hab
-sudo chmod o+w /hab
-
 git log HEAD~1..HEAD | grep -q '!!! Temporary Commit !!!'
 is_tmp_commit=$?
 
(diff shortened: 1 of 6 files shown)


support/ci/rust_tests.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-sudo mkdir -p /hab
-sudo chmod o+w /hab
```
mwrock: Now I remember this was the code I had to add for travis to pass otherwise the dir was created by the non-root travis user and things fell down. We'll see how travis does here.

mwrock: Hmm this may "pass" though because nothing in `common` changed and thus those tests do not run.

jamesc: Looks like that test ran (and now passed) - https://travis-ci.org/habitat-sh/habitat/jobs/483675988#L1123

mwrock: Oh I bet its the FS_ROOT changes that makes this work which is a better way to go

PR #6080, opened 2019-01-24
Simplify hooks test to look for appropriate change

scotthain:
...rather than full text. This allows us to avoid any issues we may run into with line endings on windows vs unix

### DIFF
--- a/components/common/src/templating/hooks.rs
+++ b/components/common/src/templating/hooks.rs
@@ -756,11 +756,7 @@ echo "The message is Hola Mundo"
         assert_eq!(hook.compile(&service_group, &ctx).unwrap(), true);
 
         let post_change_content = file_content(&hook);
-        let expected = r#"#!/bin/bash
-
-echo "The message is Hello"
-"#;
-        assert_eq!(post_change_content, expected);
+        assert!(post_change_content.contains("The message is Hello"));
 
         // Compiling again should result in no changes
         assert_eq!(hook.compile(&service_group, &ctx).unwrap(), false);


thesentinels:

Thank you for contributing!

scotthain:
I actually found a few other places where this happens, let me fix those up too!

components/sup/src/manager/service/hooks.rs
```diff
@@ -845,23 +845,21 @@ mod tests {
...
         assert_eq!(hook_table.compile(&service_group, &ctx), false);
 
         // Re-Verify init hook
         let init_hook_content = file_content(&hook_table.init.as_ref().expect("no init hook??"));
-        assert_eq!(init_hook_content, expected_init_hook);
+        assert!(init_hook_content.contains("The message is Hello"));
 
         // Re-Verify run hook
         let run_hook_content = file_content(&hook_table.run.as_ref().expect("no run hook??"));
-        assert_eq!(run_hook_content, expected_run_hook);
```
christophermaier: 🤔 

So, I think what I was doing here originally was just asserting that compiling the entire hook table would work for files that had templating placeholders (like the init hook), as well as those that didn't have any placeholders (like this run hook). 

This still tests that, but it might be a little obscured. Would you mind adding some comments alluding to this, as well as the reason that we're just doing a partial match (i.e., the line-ending issue)?

@mwrock Any ideas why we haven't run into this line-ending issue on Appveyor yet?

As aside, my urge to rewrite this test entirely is rising 😄

mwrock: I think this can all go away if we set autocrlf to `false` before we clone. The default on Windows is `true` which will replace `LF`s with `CRLF`s in the checked out code. Appveyor sets it to `input` which replaces `CRLF`s with `LF`s when writing the file to git. For us I think `false` is ideal, it basically just leaves everything alone.

PR #6081, opened 2019-01-25
Add target param to bldr job start command

chefsalim:
This change adds a PKG_TARGET param to the 'hab bldr job start' command. It will allow us to request builds for new build targets such as Windows.

tenor-46709735

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -329,21 +329,22 @@ impl Client {
     pub fn schedule_job(
         &self,
         ident: &PackageIdent,
+        target: &PackageTarget,
         package_only: bool,
         token: &str,
     ) -> Result<(String)> {
-        // TODO (SA): This API needs to be extended to support a target param.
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-client/src/lib.rs
```diff
@@ -329,21 +329,22 @@ impl Client {
...
-        // TODO (SA): This API needs to be extended to support a target param.
         let path = format!("depot/pkgs/schedule/{}/{}", ident.origin(), ident.name());
-        let result = if package_only {
-            let custom = |url: &mut Url| {
-                url.query_pairs_mut().append_pair("package_only", "true");
-            };
-            self.add_authz(self.0.post_with_custom_url(&path, custom), token)
-                .send()
-        } else {
-            self.add_authz(self.0.post(&path), token).send()
```
mwrock: is this else condition no longer necessary?

chefsalim: No, we don't need it as we now always send the custom url query params

PR #6087, opened 2019-01-26
set results folder in PWD not PLAN_CONTEXT

### LINKED ISSUE #6084: Inconsistent results location between Windows and Linux studio builds
Currently, if you build from the root of the studio, and your plan path is not in the root, Linux and Windows builds will generate results in different locations if you issue a build from the root.  Linux will generate results in the root, whereas Windows will generate results where the plan path is.  We should make these consistent.

mwrock:
fixes #6084 


thesentinels:

Thank you for contributing!

PR #6093, opened 2019-01-29
install interpreter if not already installed

mwrock:
This installs the interpreter if it is not installed and addresses a bug found when executing an install hook on a fresh hab install where powershell or busybox is not installed.

### DIFF
--- a/components/common/src/util/path.rs
+++ b/components/common/src/util/path.rs
@@ -19,9 +19,16 @@ use std::io::BufReader;
 use std::path::PathBuf;
 use std::str::FromStr;
 
+use crate::command::package::install::{self, InstallHookMode, InstallMode, LocalPackageUsage};
 use crate::error::{Error, Result};
-use crate::hcore::fs::{find_command, FS_ROOT_PATH};
-use crate::hcore::package::{PackageIdent, PackageInstall};
+use crate::ui;
+use habitat_core::{
+    channel,
+    fs::{cache_artifact_path, find_command, FS_ROOT_PATH},
+    package::{PackageIdent, PackageInstall, PackageTarget},
+    url::default_bldr_url,
+    PROGRAM_NAME,
+};
 
 /// The package identifier for the OS specific interpreter which the Supervisor is built with,
 /// or which may be independently installed
@@ -33,7 +40,9 @@ const INTERPRETER_COMMAND: &str = "busybox";
 #[cfg(target_os = "windows")]
 const INTERPRETER_IDENT: &str = "core/powershell";
 #[cfg(target_os = "windows")]
-const INTERPRETER_COMMAND: &str = "powershell";
+const INTERPRETER_COMMAND: &str = "pwsh";
+
+const VERSION: &str = include_str!(concat!(env!("OUT_DIR"), "/VERSION"));
 
 /// Returns a list of path entries, one of which should contain the interpreter binary.
 ///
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/common/src/util/path.rs
```diff
@@ -113,17 +119,24 @@ pub fn interpreter_paths() -> Result<Vec<PathBuf>> {
...
+                                &(ident.clone(), *PackageTarget::active_target()).into(),
+                                &*PROGRAM_NAME,
+                                VERSION,
+                                FS_ROOT_PATH.as_path(),
+                                &cache_artifact_path(None::<String>),
+                                None,
+                                &InstallMode::default(),
+                                &LocalPackageUsage::default(),
+                                InstallHookMode::default(),
+                            )?;
```
christophermaier: Not a change for this PR, but just reading this function call makes my eyes bleed a little... it'd be nice to see what this would look like implemented using the [builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).

components/common/src/util/path.rs
```diff
@@ -19,9 +19,13 @@ use std::io::BufReader;
...
 use std::str::FromStr;
 
+use crate::command::package::install::{self, InstallHookMode, InstallMode, LocalPackageUsage};
 use crate::error::{Error, Result};
-use crate::hcore::fs::{find_command, FS_ROOT_PATH};
-use crate::hcore::package::{PackageIdent, PackageInstall};
+use crate::hcore::fs::{cache_artifact_path, find_command, FS_ROOT_PATH};
+use crate::hcore::package::{PackageIdent, PackageInstall, PackageTarget};
+use crate::hcore::url::default_bldr_url;
+use crate::hcore::{channel, PROGRAM_NAME};
```
christophermaier: One pattern that I've found useful lately is to write these imports like this:

```rust
use habitat_core::{
    fs::{cache_artifact_path, find_command, FS_ROOT_PATH},
    package::{PackageIdent, PackageInstall, PackageTarget},
    url::default_bldr_url,
    channel, PROGRAM_NAME,
};
```
I find the `crate::hcore` bit a little misleading, because `hcore` is really just an alias we have in the root of this crate to point to `habitat_core`. With the changes in Rust 2018, we can change things up to what I've got above.

The nice thing about this is that you can clearly distinguish between what's from the current crate (it starts with `crate::`) and what isn't.

Similarly, you can group everything from this crate in a `use crate::{ ... };`  statement. And if you put all the `use` statements together with no intervening newlines, `rustfmt` will keep everything nicely organized in a very sensible way.

components/common/src/util/path.rs
fnichol: Ah, that could be a good idea!

fnichol:
To confirm, we'll get this into the 0.74.0 release to help make the net-new feature even more supported!

components/common/src/util/path.rs
```diff
@@ -113,17 +122,24 @@ pub fn interpreter_paths() -> Result<Vec<PathBuf>> {
...
-                                 interpreter's command on your $PATH. Aborting...",
-                                INTERPRETER_IDENT
-                            );
-                            return Err(Error::PackageNotFound(ident.to_string()));
+                            install::start(
+                                &mut ui::UI::default_with_env(),
+                                &default_bldr_url(),
+                                Some(&channel::STABLE_CHANNEL.to_string()),
+                                &(ident.clone(), *PackageTarget::active_target()).into(),
+                                &*PROGRAM_NAME,
```
baumanj:
```suggestion
                                &PROGRAM_NAME,
```
No need to dereference this

components/common/src/util/path.rs
```diff
@@ -113,17 +122,24 @@ pub fn interpreter_paths() -> Result<Vec<PathBuf>> {
...
+                                VERSION,
+                                FS_ROOT_PATH.as_path(),
+                                &cache_artifact_path(None::<String>),
```
baumanj:
```suggestion
                                cache_artifact_path(None::<String>),
```
[`cache_artifact_path`](https://github.com/habitat-sh/core/blob/master/components/core/src/fs.rs#L147) returns `PathBuf`, which satisfies the `AsRef<Path>` requirement on the parameter.

PR #6098, opened 2019-01-30
generate api docs for windows http gateway

### LINKED ISSUE #5925: HTTP Gateway on windows returns 'No API docs provided at build'
Hi,

I am running the latest habitat...
```
PS C:\Users\hab> hab --version
hab 0.69.0/20181127182027
```

If I hit the Web Gateway at `http://localhost:9631` it returns `No API docs provided at build` rather than returning the docs for the Supervisor API.

mwrock:
fixes #5925 

Depends on promotion of https://github.com/habitat-sh/core-plans/pull/2233

### DIFF
--- a/components/sup/build.rs
+++ b/components/sup/build.rs
@@ -3,9 +3,9 @@ include!("../libbuild.rs");
 
 use std::env;
 use std::fs::File;
-use std::io::Write;
-use std::path::Path;
-use std::process::Command;
+use std::io::{self, Write};
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitStatus};
 
 fn main() {
     habitat::common();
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/sup/build.rs
```diff
@@ -33,3 +27,28 @@ fn generate_apidocs() {
...
+        .arg("raml2html.bat")
+        .arg("-i")
+        .arg(src)
+        .arg("-o")
+        .arg(dst)
+        .status()
+}
+
+#[cfg(unix)]
+fn raml2html_cmd() -> io::Result<ExitStatus> {
```
baumanj: Where are `src` and `dst` coming from?

In the spirit of making it clear what parts are different per-platform and catching bugs at compile time, I think we can have one implementation:
```rust
fn raml2html_cmd(dst: PathBuf, src: PathBuf) -> io::Result<ExitStatus> {
    let mut cmd = Command::new(if cfg!(windows) { "cmd" } else { "raml2html" });

    let cmd = if cfg!(windows) {
        cmd.arg("/c").arg("raml2html.bat")
    } else {
        &mut cmd
    };

    cmd.arg("-i").arg(src).arg("-o").arg(dst).status()
}
```

The need for two separate `if cfg!` blocks owes to the fact that the builder pattern returns different types for `new` and `arg`. I think this is the cleanest way to do it.

christophermaier: Seems like a good simplification 👍

mwrock:
cleaned up with your suggestion @baumanj Thanks!

PR #6100, opened 2019-01-30
Classify the remaining unfixed clippy lints as ones to fix versus ignore

baumanj:
No code changes here, this is just to confirm our consensus about what our goals and expectations are. For each lint I'm proposing to classify, I've provided an argument.

For the ones in `ALLOWED_LINTS`, I'm explaining why I don't think they're worth fixing. If you agree, give it a 👍. If instead you think we *should* fix the lint, give it a 👎 and give some reasoning.

For the ones in `LINTS_TO_FIX`, it's the opposite. If you agree we should fix them, give it a 👍. If you think it's not worth the effort, please 👎 and say why. The intent is that once all the instances of a given lint in `LINTS_TO_FIX` and fixed (or suppressed with a case-by-case `#[allow(…)]`, we can move them to `DENIED_LINTS`, which will cause them to fail CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,18 +194,18 @@ UNEXAMINED_LINTS =
 
 # Lints we disagree with and choose to keep in our code with no warning
 ALLOWED_LINTS = clippy::module_inception \
-                clippy::needless_return \
                 clippy::new_ret_no_self \
                 clippy::new_without_default \
-                clippy::new_without_default_derive \
-                clippy::redundant_field_names
+                clippy::new_without_default_derive
 
 # Known failing lints we want to receive warnings for, but not fail the build
 LINTS_TO_FIX = clippy::cyclomatic_complexity \
                clippy::large_enum_variant \
                clippy::len_without_is_empty \
                clippy::needless_pass_by_value \
+               clippy::needless_return \
                clippy::question_mark \
+               clippy::redundant_field_names \
                clippy::too_many_arguments \
                clippy::trivially_copy_pass_by_ref \
                clippy::wrong_self_convention
--- a/Makefile
+++ b/Makefile
@@ -190,26 +190,25 @@ unit-sup: build-launcher-for-supervisor-tests
 .PHONY: build-launcher-for-supervisor-tests
 
 # Lints we need to work through and decide as a team whether to allow or fix
-UNEXAMINED_LINTS = clippy::cyclomatic_complexity \
-                   clippy::large_enum_variant \
-                   clippy::len_without_is_empty \
-                   clippy::module_inception \
-                   clippy::needless_pass_by_value \
-                   clippy::needless_return \
-                   clippy::new_ret_no_self \
-                   clippy::new_without_default \
-                   clippy::new_without_default_derive \
-                   clippy::question_mark \
-                   clippy::redundant_field_names \
-                   clippy::too_many_arguments \
-                   clippy::trivially_copy_pass_by_ref \
-                   clippy::wrong_self_convention
+UNEXAMINED_LINTS =
 
 # Lints we disagree with and choose to keep in our code with no warning
-ALLOWED_LINTS =
+ALLOWED_LINTS = clippy::module_inception \
+                clippy::needless_return \
+                clippy::new_ret_no_self \
+                clippy::new_without_default \
+                clippy::new_without_default_derive \
+                clippy::redundant_field_names
 
 # Known failing lints we want to receive warnings for, but not fail the build
-LINTS_TO_FIX =
+LINTS_TO_FIX = clippy::cyclomatic_complexity \
+               clippy::large_enum_variant \
+               clippy::len_without_is_empty \
+               clippy::needless_pass_by_value \
+               clippy::question_mark \
+               clippy::too_many_arguments \
+               clippy::trivially_copy_pass_by_ref \
+               clippy::wrong_self_convention
 
 # Lints we don't expect to have in our code at all and want to avoid adding
 # even at the cost of failing the build


thesentinels:

Thank you for contributing!

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
-                   clippy::wrong_self_convention \
-                   renamed_and_removed_lints
+UNEXAMINED_LINTS =
 
 # Lints we disagree with and choose to keep in our code with no warning
-ALLOWED_LINTS =
+ALLOWED_LINTS = clippy::module_inception \
+                clippy::needless_return \
+                clippy::new_ret_no_self \
+                clippy::new_without_default \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_without_default

I think implementing `Default` should be an affirmative choice, otherwise it's adding complexity unnecessarily and could undermine type-safety by allowing a type to be used where it's not expected merely because it has a default.

[new_without_default.txt](https://github.com/habitat-sh/habitat/files/2813939/new_without_default.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+                clippy::new_without_default_derive \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_without_default_derive

Basically the same as above, but with types that could have their `Default` implementation derived.

[new_without_default_derive.txt](https://github.com/habitat-sh/habitat/files/2813946/new_without_default_derive.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+                clippy::redundant_field_names
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#redundant_field_names

Personally, I'd rather change these as I think the shorthand version is more idiomatic, but I don't feel very strongly and I know that @raskchanky prefers the more explicit version, so I'll defer to what other people think.

[redundant_field_names.txt](https://github.com/habitat-sh/habitat/files/2813955/redundant_field_names.txt)

raskchanky: Since we last spoke on this topic, I've become more comfortable with the shorthand version, so I'm happy to go that route.

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
 
 # Known failing lints we want to receive warnings for, but not fail the build
-LINTS_TO_FIX =
+LINTS_TO_FIX = clippy::cyclomatic_complexity \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#cyclomatic_complexity

I think the best argument that we should fix these is that it may prevent the compiler from optimizing our code more effectively. It doesn't mean we need to fix these particular instances (we can use #[allow(…)] on a case-by-case) or do it right away, but I would definitely like to be alerted of new code that is overly complex.

[cyclomatic_complexity.txt](https://github.com/habitat-sh/habitat/files/2813963/cyclomatic_complexity.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::large_enum_variant \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#large_enum_variant

This seems like a pretty easy win to save on memory.

[large_enum_variant.txt](https://github.com/habitat-sh/habitat/files/2814044/large_enum_variant.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::len_without_is_empty \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#len_without_is_empty

There are very few instances in our codebase and this helps encourage us to both write easier to understand code and take advantages of opportunities for simple optimization.

[len_without_is_empty.txt](https://github.com/habitat-sh/habitat/files/2814048/len_without_is_empty.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::needless_pass_by_value \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#needless_pass_by_value

The argument in favor of fixing is that we stand to eliminate unnecessary overhead. The argument against is that we do this in *a lot* of places. Possibly we should try fixing some see if a stronger feeling develops.

[needless_pass_by_value.txt](https://github.com/habitat-sh/habitat/files/2814059/needless_pass_by_value.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::question_mark \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#question_mark

This only triggered in [one place in the `habitat` repo](https://github.com/habitat-sh/habitat/blob/master/components/butterfly/src/server/push.rs#L398-L400), but further investigation led to the realization that this:
```rust
        let mut member = None;
        self.server.member_list.with_member(&rumor_key.key(), |m| {
            if let Some(m) = m {
                member = Some(m.clone());
            }
        });
        if member.is_none() {
            return None;
        }
        let payload = Membership {
            member: member.unwrap(),
```
could be refactored into:
```rust
        let member = self
            .server
            .member_list
            .with_member(&rumor_key.key(), |m| m.cloned())?;

        let payload = Membership {
            member: member,
```
(plus a trivial change to[`with_member`](https://github.com/habitat-sh/habitat/blob/master/components/butterfly/src/member.rs#L689) to support a non-`()` return type)

So I think this lint is a good hint that we can likely do better. Let's fix this instance and avoid others in the future.

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::too_many_arguments \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#too_many_arguments

I'm not totally opposed to `#[allow(…)]`'ing some cases, but I think this is generally good practice since this makes things pretty hard to understand at the call sites.

[too_many_arguments.txt](https://github.com/habitat-sh/habitat/files/2814087/too_many_arguments.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::trivially_copy_pass_by_ref \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

This lets us both eliminate a lot of unnecessary `&`s, improving legibility as well as potentially improve performance by better using registers!

[trivially_copy_pass_by_ref.txt](https://github.com/habitat-sh/habitat/files/2814092/trivially_copy_pass_by_ref.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               clippy::wrong_self_convention \
```
baumanj: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#wrong_self_convention

The more rust we write, the more adhering to these kinds of conventions will reduce the cognitive overhead of dealing with our codebase. Especially since we don't have a ton of instances to fix, I think this one's worth it.

[wrong_self_convention.txt](https://github.com/habitat-sh/habitat/files/2814099/wrong_self_convention.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+               renamed_and_removed_lints
```
baumanj: This one should go away once we finally purge the `protobuf` crate in favor of `prost`, but keeping this here serves as a good reminder.

[renamed_and_removed_lints.txt](https://github.com/habitat-sh/habitat/files/2814103/renamed_and_removed_lints.txt)

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
-                   clippy::redundant_field_names \
-                   clippy::too_many_arguments \
-                   clippy::trivially_copy_pass_by_ref \
-                   clippy::wrong_self_convention \
-                   renamed_and_removed_lints
+UNEXAMINED_LINTS =
 
 # Lints we disagree with and choose to keep in our code with no warning
-ALLOWED_LINTS =
+ALLOWED_LINTS = clippy::module_inception \
```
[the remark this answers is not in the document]
christophermaier: We can hold off on this for now, but we might be able to add it if we stop committing generated code to the repo 😄

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+                clippy::needless_return \
```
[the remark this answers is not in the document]
christophermaier: I like having the option to have an explicit `return` in those large code blocks, but it would be nice to catch them on smaller ones, since that does strike me as noisy.

It might be worth enabling globally, but specifically excepting it on thorny functions where it is useful.

Further down the road, we might be able to use code that runs afoul of this as a signal that it could be refactored to something that's easier to read.

Makefile
```diff
@@ -190,27 +190,26 @@ unit-sup: build-launcher-for-supervisor-tests
...
+                clippy::new_ret_no_self \
```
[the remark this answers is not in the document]
christophermaier: I don't know that there's an established naming convention for things that return `Result<Self>`, but I'm open to suggestions.

Makefile
christophermaier: I don't know... I think a `new() -> Self` is signaling the same thing as `Default`... I think this lint is basically acknowledging that, and ensuring that your types work properly in the standard Rust API.

If we were to adopt this, it probably should go along with adopting `new_ret_no_self` and `new_without_default_derive`.

Makefile
baumanj: In light of @raskchanky's change of heart, I'm voting against this lint's current place in `ALLOWED_LINTS`. We should change these.

christophermaier: I'm currently in the "explicit" camp, but I could be persuaded otherwise. I'm happy to try it out, particularly as it looks like it's becoming more idiomatic.

Makefile
christophermaier: Yup, I like the warning flag.

Makefile
christophermaier: I like this one. There might be a lot of warnings initially, but I feel like it'll be good in the long run.

Makefile
christophermaier: Agreed.

Makefile
christophermaier: Wow, some of those argument lists are bonkers.  Let's do this.

Makefile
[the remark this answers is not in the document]
baumanj: One holy war at a time, please

Makefile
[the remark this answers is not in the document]
baumanj: Currently we have ~50 instances of this. If you're down to evaluate them on a case-by-case basis, I'm cool with moving this to `LINTS_TO_FIX`.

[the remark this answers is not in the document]
mpeck: I agree with enabling it and then just adding exceptions for those large code blocks. Those huge blocks are generally pretty hard to follow anyway, so it makes sense to me to use this as a signal to do some future refactoring.

Makefile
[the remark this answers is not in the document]
baumanj: The stdlib seems to use [`try_new`](https://github.com/rust-lang/rust/blob/master/src/librustc_data_structures/owning_ref/mod.rs#L889) or [`from_*`](http://doc.rust-lang.org/1.32.0/alloc/str/trait.FromStr.html#tymethod.from_str) for these.

Makefile
baumanj: The main difference is that having a `new` isn't implementing a trait and so doesn't opt-in to special handling by the compiler.

Makefile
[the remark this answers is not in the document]
mpeck: Is it idiomatic to only return `Self` from `new`? I haven't found any examples showing otherwise. I lean towards *not* returning `Result<Self>` and just returning `Self` if that be the case. It seems like this particular lint is tied closely with `new_without_default`.

Makefile
mpeck: Seems like `new_ret_no_self` and this should go hand in hand.

Makefile
mpeck: These three(`new_ret_no_self`, `new_without_default` and `new_without_default_derive`) seem like they should go together. If we use one then we should use them all.

Makefile
mpeck: I'm in the shorthand camp here. I think it's generally pretty obvious what's going on once you know the syntax.

Makefile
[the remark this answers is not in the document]
baumanj: > Is it idiomatic to only return `Self` from `new`?

Can you find an example in the stdlib that returns something other than `Self` for `new`?

Makefile
baumanj: Here's a [slightly different fix](https://github.com/habitat-sh/habitat/pull/6101)

Makefile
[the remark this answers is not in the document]
mpeck: I can't, though that's just from me clicking through. I don't how to search specifically for that.

Makefile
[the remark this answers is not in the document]
christophermaier: "Please implement all holy wars incrementally."

Makefile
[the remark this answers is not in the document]
christophermaier: Hrmm... `try_new` might be interesting to try on for size. I think I'd like to boil off the other lints before tackling that, though, so I'll be 👎 on this for now.

Makefile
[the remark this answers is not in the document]
christophermaier: @baumanj Yeah, I'll be happy to evaluate our existing instances of this lint.

Makefile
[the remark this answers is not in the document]
baumanj: > @baumanj are you saying that we _should_ be returning `Self` from `new()` and _not_ `Result<Self>`? If you are, then I think we are agreeing with each other.

The lint [`clippy::new_ret_no_self`](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_ret_no_self) asserts that any associated function called `new` should be returning `Self`. I agree that is the rust convention, but I also think that `Result<Self>` is an acceptable variant since it's sill a *kind* of `Self`. Given the static typing, it's not like you're going to forget to handle the error case, so I don't know if there's a ton of value in `try_new`, so I'm suggesting we keep this in `ALLOWED_LINTS`. That said, I don't feel terribly strongly either way.

Makefile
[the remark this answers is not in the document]
baumanj: Recursive holy war has led to stack overflow

Makefile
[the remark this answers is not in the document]
raskchanky: I don't have a strong opinion on this one.

Makefile
raskchanky: No strong opinions here.

Makefile
raskchanky: This is definitely something we should fix, and I'm guessing it will be fairly instructive as well.  I know when I first started with Rust, I passed everything by value by default, just because that's more common in other languages.

Makefile
raskchanky: My prost PR moves this to the right place.

Makefile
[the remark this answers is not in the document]
jamesc: +1 on this one.  Most situations where I've run into this recently have led to cleaner code on early return removal

Makefile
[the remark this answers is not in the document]
jamesc: Where the code is now I'd agree with @baumanj and say this is ok, in that it's a kind of `Self`.  

One pattern I've noticed is where we build up a struct in a `new`, by calling other `new` -> `Result<Self>` functions which ends up spreading out through the code. There seems to be [cases](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/mod.rs#L187-L188) where all we're doing is creating directories or some FS ops which can return e.g. permission denied.  There could be cases where we do this later in the flow on-demand rather than up front.

Makefile
jamesc: No strong opinion but I do like the look of the short version.

Makefile
jamesc: +1.  We should do this.

chefsalim:
If I'm reading this correctly, we're considering globally turning on/off some lints. I'm not a fan of that strategy except maybe for a very small list (like 2 or 3).  My preference would be to enable everything and create exceptions on a one-off basis where needed.  My concern with doing it otherwise is that we're going to continue writing non-Clippy code that we could be in compliance now, and not have to do even more work on the day that we turn on those lints.

baumanj:
> If I'm reading this correctly, we're considering globally turning on/off some lints. I'm not a fan of that strategy except maybe for a very small list (like 2 or 3).

The list of lints we're unconditionally allowing is indeed small:
1. [`clippy::module_inception`](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#module_inception)
1. [`clippy::new_ret_no_self`](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_ret_no_self)
1. [`clippy::new_without_default`/`new_without_default_derive`](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_without_default)

These are turned off because we concluded they're not worth fixing or explicitly disagree that they *should* be. Since (1) comes from generated code only and isn't easy to one-off disable and (2) has a lot of instances, I don't think it's really worth the effort to put in config exceptions. (3) only has a couple, but I think we agree that it's just not a lint that we agree with, so disabling it globally seems cleaner.

If you feel strongly, feel free to open a PR and take a different approach.

chefsalim:
To be clear @baumanj I'm not advocating to stop progress on this. The ALLOWED_LINTS list is small enough.  

On the new_ret_no_self lint, it did make the code more consistent across the codebase in Builder, so it felt like a win, and then one never has to think about it again.  If its too onerous for now for the hab codebase, fine. 

Good stuff overall...

PR #6102, opened 2019-01-30
if the studio is rooted immediately under the system drive, do not try to delete it

mwrock:
This address a specific scenario where `hab pkg build . -D` breaks on windows containers because the docker studio roots the studio right in `c:\` which you cannot delete. To work around this today, you must pass `-R` to "persist" the studio.

Of course you would not want to delete `c:\` in any scenario whatsoever so this avoids such a removal and emits a message stating that it is not going to remove the system drive.

### DIFF
--- a/components/studio/bin/hab-studio.ps1
+++ b/components/studio/bin/hab-studio.ps1
@@ -419,9 +419,14 @@ function Remove-Studio {
     Write-RmHelp
     return
   }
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
"Seems wise"
&nbsp;&nbsp;&nbsp;&nbsp;@baumanj, 2019 International Understatement Grand Champion

PR #6109, opened 2019-01-31
Run health check in the first tick after service load

jaym:
This brings back the behavior prior to hab 0.72 where the
first health check would run at the first tick after the
service is loaded. The behavior when the service gets changed
either through upgrading or reconfiguring also now does a health
check at the next tick

### DIFF
--- a/components/sup/src/manager/service/mod.rs
+++ b/components/sup/src/manager/service/mod.rs
@@ -26,7 +26,7 @@ use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::result;
 use std::sync::{Arc, RwLock};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use crate::butterfly::rumor::service::Service as ServiceRumor;
 use crate::common::templating::config::CfgRenderer;
@@ -210,7 +210,7 @@ impl Service {
             topology: spec.topology,
             update_strategy: spec.update_strategy,
             config_from: spec.config_from,
-            scheduled_health_check: None,
+            scheduled_health_check: Some(Instant::now()),
             svc_encrypted_password: spec.svc_encrypted_password,
             health_check_interval: spec.health_check_interval,
             defaults_updated: false,
@@ -397,7 +397,7 @@ impl Service {
             }
         }
         if svc_updated {
-            self.schedule_special_health_check();
+            self.schedule_health_check_at_next_tick();
         }
 
         svc_updated
@@ -643,8 +643,6 @@ impl Service {
                 outputln!(preamble self.service_group,
                             "Updating service {} to {}", self.pkg.ident, pkg.ident);
 
-                self.schedule_special_health_check();
-
                 match CfgRenderer::new(&Self::config_root(&pkg, self.config_from.as_ref())) {
                     Ok(renderer) => self.config_renderer = renderer,
                     Err(e) => {
@@ -682,6 +680,7 @@ impl Service {
         }
 
         self.initialized = false;
+        self.schedule_health_check_at_next_tick();
     }
 
     pub fn to_rumor(&self, incarnation: u64) -> ServiceRumor {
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

jeremymv2:
@jaym having had some ☕️ I see now where my concerns above are moot. 

https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/mod.rs#L1002-L1007

The `else` branch will ensure that if the first check fails, the next check will be at `HealthCheckInterval::default()` interval. I haven't tested that to ensure I'm 100% correct, but I feel better now with the direction of the changes here. Thanks! 👍

components/sup/src/manager/service/mod.rs
```diff
@@ -27,7 +27,7 @@ use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::result;
 use std::sync::{Arc, RwLock};
-use std::time::Instant;
+use std::time::{Instant, Duration};
```
baumanj: Either run `cargo fmt` (or set up rustfmt to automatically run in your editor) or fix this one by hand to make travis pass
```suggestion
use std::time::{Duration, Instant};
```

baumanj:
The DCO sign-off needs to be on every commit. Can you amend that last one or squash?

jaym:
thoughts on where such a test would go? I'm thinking a simple package that always return success for the health check and test that the health check endpoint returns success in < 30s.

baumanj:
> thoughts on where such a test would go?

I'm more a fan of unit tests for these kinds of regressions, so I would first try a test that just created a new `Service` and then asserted that the `scheduled_health_check` was `Some` and within a couple seconds of `now`.

components/sup/src/manager/service/mod.rs
```diff
@@ -213,7 +213,7 @@ impl Service {
             topology: spec.topology,
             update_strategy: spec.update_strategy,
             config_from: spec.config_from,
-            scheduled_health_check: None,
+            scheduled_health_check: Some(Instant::now()),
```
christophermaier: With this change, I don't think `scheduled_health_check` ever really needs to be an `Option`, which can simplify some of the rest of this code (it would require a bit of tweaking to `run_health_check_hook`, though).

It also feels like this needs to be harmonized with `schedule_health_check_at_next_tick`, though, since it is the same intent as that method (though not explicitly stated), but different code.

baumanj: I looked at making it not an option and it breaks the other logic. Maybe there's a way to refactor it more, but we do need something to indicate the difference between a health check which is scheduled but is waiting to be run and no health check being scheduled.

christophermaier: Alternatively, we make an operational distinction between the scenarios when we want to "schedule something if nothing is already scheduled earlier" and when we want to "schedule the next invocation, because I just finished running this one". The only place where `None` is used is immediately after we execute the health check hook, and we don't leave that function without scheduling something.

I've got a couple ideas of how we might approach this, but I need to play with them a bit. Also, I can already see how some of this is going to have to change when #6107 lands, so I'm happy to try and address this there instead.

baumanj: > The only place where `None` is used is immediately after we execute the health check hook, and we don't leave that function without scheduling something.

Yeah, you're right. I didn't dig into the code enough, and it's not obvious from this level:
```rust
        self.scheduled_health_check = None;

        if check_result == HealthCheck::Ok {
            self.schedule_routine_health_check();
            debug!(
                "Service ({}) health check is: {}",
                self.spec_ident, check_result
            );
        } else {
            debug!(
                "Service ({}) health check is: {}; scheduling special health check",
                self.spec_ident, check_result
            );
            self.schedule_special_health_check();
        }
```
That `self.scheduled_health_check` is `Some` by the end. Perhaps restructuring that a bit would make it clearer. We should add some unit tests while we're at it, though.

components/sup/src/manager/service/mod.rs
```diff
@@ -1280,8 +1280,27 @@ mod tests {
...
+    fn health_check_is_due_at_creation() {
+        let service = initialize_test_service();
+        assert!(
+            service.scheduled_health_check.is_some(),
+            "Expected a scheduled health check"
+        );
+        assert!(
+            service.scheduled_health_check.unwrap() < Instant::now(),
+            "Expected health check due at creation"
+        );
```
baumanj: This works fine, but just a rust testing FYI, you could structure it like this for a bit less redundancy
```rust
if let Some(scheduled_health_check) = service.scheduled_health_check {
    assert!(scheduled_health_check < Instant::now(), "Expected health check due at creation");
} else {
    panic("Expected a scheduled health check");
}
```

PR #6113, opened 2019-02-03
Fix error runining install.sh/Dockerfile

qubitrenegade:
For some reason, even though the shbang line is set to `/bin/bash` `sh` executes `/tmp/install.sh` which fails at line [202](https://github.com/habitat-sh/habitat/blob/master/components/hab/install.sh#L202)

Since we expect `bash` anyway, I changed the invocation of the install script to `bash /tmp/install.sh` instead, (although, why isn't this just executed from `components/hab/install.sh`?) which allows the install script and Dockerfile to complete successfully.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -21,7 +21,6 @@ cd habitat
 Then, run the system preparation scripts
 
 ```
-cp components/hab/install.sh /tmp/
 sh support/linux/install_dev_0_ubuntu_latest.sh
 sh support/linux/install_dev_9_linux.sh
 ```
@@ -58,6 +57,13 @@ The binaries will be in `habitat/target/debug` and can be run directly, or you c
 cd components/sup
 cargo run -- --help
 cargo run -- status
+```
+
+Or from the habitat directory:
+
+```
+cargo run -p hab plan --help
+cargo run -p hab sup --help
 ```
 
 ## Compiling launcher
@@ -248,7 +254,6 @@ cd habitat
 Then, run the system preparation scripts and try to compile the project:
 
 ```
-cp components/hab/install.sh /tmp/
 sh support/linux/install_dev_0_ubuntu_14.04.sh
 sh support/linux/install_dev_9_linux.sh
 . ~/.profile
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

baumanj:
@qubitrenegade: are you still interested in fixing this? If not, I'm going to close it in the interest of keeping the open PRs limited to active work, but we can always reopen it later.

qubitrenegade:
Yep, totally!  Real/work life has been hectic and actually had another idea...

How about executing the script relative to our repository?

```
sudo "$(dirname -- "$0")/../../components/hab/install.sh"
```

This does a few things:

* ensures we're picking the appropriate interpreter according to the shbang line for our `install.sh` script.
* executes the install script relative to the `support/linux` directory, ensuring we always run the "latest" bundled with the source 
* since we're executing relative to our linux install script (`install_dev_9_linux.sh`) we eliminate a step copying to `/tmp` (and solve the problem of having to remount temp without `noexec` :smile: )

I debated... since the habitat install script is a bash script, we could make the `9_linux.sh` script bash, and then use `$BASH_SOURCE`.   But I figure we're safe using `$0` since the odds of someone linking to the `9_linux.sh` script and then trying to execute it are slim, and we're not sourcing it.  So no real need for Bash features.

The alternative would be:

```
sudo ${BASH_SOURCE%/*}/../../components/hab/install.sh"
```

But again, that would require Bash, and I didn't know if there was a specific reason that the `9_linux.sh` script was trying to not use Bash.

Thoughts?

Thanks!

baumanj:
The safest bash way I know to get the directory the script is in is [like this](https://github.com/habitat-sh/habitat/blob/0359ba8665f724d1be1802093f50d7ecc2345da0/components/plan-build/bin/hab-plan-build.sh#L328):
```bash
source_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
But since I don't think this script is getting sourced, I think the simpler approach works, and leaving this as POSIX sh, should be fine. I need to run a release, but after that's done (hopefully tomorrow or the day after), I'll dig into this a bit more.

PR #6121, opened 2019-02-04
Adds origin deletion tool to hab binary

eeyun:

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -522,7 +522,27 @@ impl Client {
 
         let res = self.add_authz(self.0.delete(&path), token).send()?;
 
-        if res.status != StatusCode::Ok {
+        // We expect NoContent, because the origin must be empty to delete
+        if res.status == StatusCode::NoContent {
+            Ok(())
+        } else {
+            Err(err_from_response(res))
+        }
+    }
+
+    /// Delete an origin
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin is populated with > 0 packages
+    ///  * Submitted Origin is not found
+    pub fn delete_origin(&self, origin: &str, token: &str) -> Result<()> {
+        let path = format!("depot/origins/{}", origin);
+
+        let res = self.add_authz(self.0.delete(&path), token).send()?;
+
+        if res.status != StatusCode::NoContent {
             return Err(err_from_response(res));
         }
 
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -237,6 +237,16 @@ pub fn get() -> App<'static, 'static> {
             (about: "Commands relating to Habitat origin keys")
             (aliases: &["o", "or", "ori", "orig", "origi"])
             (@setting ArgRequiredElseHelp)
+            (@subcommand delete =>
+                (about: "Removes an unused/empty origin")
+                (aliases: &["del", "dele"])
+                (@arg ORIGIN: +required {valid_origin} "The origin name")
+                (@arg BLDR_URL: -u --url +takes_value {valid_url}
+                     "Specify an alternate Builder endpoint. If not specified, the value will \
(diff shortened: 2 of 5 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-client/src/lib.rs
```diff
@@ -517,6 +517,25 @@ impl Client {
...
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin is populated with > 0 packages
+    ///  * Submitted Origin is not found
+    pub fn delete_origin(&self, origin: &str, token: &str) -> Result <()> {
+        let path = format!("depot/origins/{}", origin);
+
+        let res = self.add_authz(self.0.delete(&path), token).send()?;
+
+        if res.status != StatusCode::NoContent {
```
baumanj: If this is correct, it might warrant a comment. We *want* to see `StatusCode::NoContent` because we're trying to delete an empty origin, correct?

Alternatively, it might be better to structure the code like this:
```rust
        // We expect NoContent, because the origin must be empty to delete
        if res.status == StatusCode::NoContent {
            Ok(())
        } else {
            Err(err_from_response(res))
        }
```

components/hab/src/cli.rs
```diff
@@ -237,6 +237,16 @@ pub fn get() -> App<'static, 'static> {
             (about: "Commands relating to Habitat origin keys")
             (aliases: &["o", "or", "ori", "orig", "origi"])
             (@setting ArgRequiredElseHelp)
+            (@subcommand delete =>
+                (about: "Removes an unused/empty origin")
+                (aliases: &["del", "dele"])
+                (@arg ORIGIN: +required {valid_origin} "The origin name")
+                (@arg BLDR_URL: -u --url +takes_value {valid_url}
```
baumanj: CLAP groks default values, and then you don't have to include them in the string yourself
```suggestion
                (@arg BLDR_URL: -u --url +takes_value {valid_url} default_value[https://bldr.habitat.sh]
```
We don't use this elsewhere, but we really should. I don't think there's a downside to gradually changing it, but you can take this or leave it.

components/hab/src/command/origin/delete.rs
```diff
@@ -0,0 +1,33 @@
...
+pub fn start(ui: &mut UI, bldr_url: &str, token: &str, origin: &str) -> Result<()> {
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
+
+    ui.status(Status::Deleting, format!("origin {}.", origin))?;
+
+    api_client
+        .delete_origin(origin, token)
+        .map_err(Error::APIClient)?;
+
+    ui.status(Status::Deleted, format!("origin {}.", origin))?;
```
baumanj: Make this return the value of the expression and you can eliminate the `Ok(())` on L32
```suggestion
    ui.status(Status::Deleted, format!("origin {}.", origin))
```

eeyun: Removing the try here gives us a mismatch in error types because of the error type that ui.status returns. So i went with `Ok(ui.status(Status::Deleted, format!("origin {}.", origin))?)`

baumanj: My bad. I thought I had successfully built that suggestion since I'm aware of that kind of failure that the `?` operator avoids.

In this case, I think what you had is the best and I retract my suggestion. I'm not really a fan of the `Ok(…?)` form since I think it's kind of confusing. Also, when the return type is `Result<()>`, it triggers [this lint](https://rust-lang.github.io/rust-clippy/master/#unit_arg).

If you like the look of omitting the `Ok(())`, another alternative is
```rust
    ui.status(Status::Deleted, format!("origin {}.", origin))
        .map_err(Into::into)
```
That's basically doing the same thing as `?`, but without the need to re-wrap in `Ok(…)`. I kind of like the explicitness of it, but whichever form you like best is fine with me.

eeyun: :+1:

PR #6123, opened 2019-02-05
Add flag to skip startup of the supervisors gossip layer

mpeck:
Adds a new flag to the cli `--no-listen-gossip` that causes the supervisor to skip startup of the gossip layer. 

To test just start up the supervisor with the `--no-listen-gossip` flag and then try to connect to it with either `telnet` or `netcat` in another terminal.

`hab sup run --no-listen-gossip`

In another terminal
`telnet 127.0.0.1 9638` or `nc 127.0.0.1 9638`

If it's doing what it's supposed to you should **NOT** be successful in connecting.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -946,6 +946,7 @@ pub fn sub_sup_run() -> App<'static, 'static> {
         Implies NO_COLOR")
     (@arg HEALTH_CHECK_INTERVAL: --("health-check-interval") -i +takes_value {valid_health_check_interval}
         "The interval (seconds) on which to run health checks [default: 30]")
+    (@arg NO_LISTEN_GOSSIP: --("no-listen-gossip") "Don't startup the gossip listener.")
     )
 }
 
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -182,6 +182,7 @@ pub struct ManagerConfig {
     pub organization: Option<String>,
     pub watch_peer_file: Option<String>,
     pub tls_files: Option<(PathBuf, PathBuf)>,
+    pub no_listen_gossip: bool,
 }
 
 impl ManagerConfig {
@@ -207,6 +208,7 @@ impl Default for ManagerConfig {
             organization: None,
             watch_peer_file: None,
             tls_files: None,
+            no_listen_gossip: false,
         }
     }
 }
@@ -265,6 +267,7 @@ pub struct Manager {
     service_states: HashMap<PackageIdent, Timespec>,
     sys: Arc<Sys>,
     http_disable: bool,
+    no_listen_gossip: bool,
 }
 
 impl Manager {
@@ -378,6 +381,7 @@ impl Manager {
             service_states: HashMap::new(),
             sys: Arc::new(sys),
             http_disable: cfg.http_disable,
+            no_listen_gossip: cfg.no_listen_gossip,
         })
     }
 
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -182,6 +182,7 @@ pub struct ManagerConfig {
     pub organization: Option<String>,
     pub watch_peer_file: Option<String>,
     pub tls_files: Option<(PathBuf, PathBuf)>,
+    pub no_listen_gossip: bool,
```
baumanj: Instead of adding this `bool`, why not make the `gossip_listen` field into an `Option`?

components/hab/src/cli.rs
```diff
@@ -946,6 +946,7 @@ pub fn sub_sup_run() -> App<'static, 'static> {
         Implies NO_COLOR")
     (@arg HEALTH_CHECK_INTERVAL: --("health-check-interval") -i +takes_value {valid_health_check_interval}
         "The interval (seconds) on which to run health checks [default: 30]")
+    (@arg NO_LISTEN_GOSSIP: --("no-listen-gossip") "Don't startup the gossip listener.")
```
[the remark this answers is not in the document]
baumanj: Lets also think about how this should interact with the `--peer` argument.

components/sup/src/manager/mod.rs
eeyun: Out of curiosity would that just be for the sake of standardization or is there a utilization difference between `bool` vs `Some()` or `None`?

christophermaier: I think it's more to prevent having a bunch of struct fields than anything. If you're not going to be listening to gossip, it makes sense to fold that into the `gossip_listen` field as a `None` then having another field to keep track of. 

The more Rust I write, the more I think booleans are a kind of code smell that indicates that a more expressive solution is hiding somewhere.

(No, booleans aren't always a bad thing, but they do cause me to think a bit deeper.)

baumanj: In this case, it's the difference between having two fields:
```rust
   gossip_listen: GossipListenAddr,
   no_listen_gossip: bool,
```
and one:
```rust
   gossip_listen: Option<GossipListenAddr>,
```
The biggest reason to prefer the latter is that the former allows for the existence of an inconsistent state. Any time `no_listen_gossip` was `false`, there would still have to be an addr value in `gossip_listen`, but wouldn't make sense. If there were a value in there, code could access it and try to listen when we don't want it to. Combining them into one type prevents that situation by ensuring that in the case we don't want to listen, there is no addr value we could possibly use to do so. Code that tried would fail to compile and we'd catch the logic error early.

mpeck: What are your thoughts on just using the `--listen-gossip` option and allowing the user to pass `false`?

baumanj: I'm not a fan because that would imply that `--listen-gossip` accepts a `bool`, and it doesn't. If `--listen-gossip false` is valid, the fact that `--listen-gossip true` is non-sensical is surprising.

Is there something you don't like about `<--listen-gossip <LISTEN_GOSSIP>|--no-listen-gossip>`? We have other instances where an option and a flag are mutually exclusive.

mpeck: Eh, it wasn't a great suggestion. It's not that I don't like having the `--no-listen-gossip` flag, I'm just having difficulties getting it to work properly with clap.

PR #6137, opened 2019-02-08
Remove the hard-coded hab user in the tests

jamesc:
The username/groupname used doesn't effect the test, so just use current user/group instead of `hab`/`hab`.

This fixes some tests running in BuildKite.

Signed-off-by: James Casey [james@chef.io](mailto:james@chef.io)

### DIFF
--- a/components/common/src/templating/config.rs
+++ b/components/common/src/templating/config.rs
@@ -661,6 +661,7 @@ fn ensure_directory_structure(root: &Path, file: &Path, user: &str, group: &str)
 
 fn write_templated_file(path: &Path, compiled: &str, user: &str, group: &str) -> Result<()> {
     File::create(path).and_then(|mut file| file.write_all(compiled.as_bytes()))?;
+
     set_permissions(&path, &user, &group)?;
     Ok(())
 }
@@ -670,7 +671,10 @@ mod test {
     use super::*;
     use crate::{
         error::Error,
-        hcore::package::{PackageIdent, PackageInstall},
+        hcore::{
+            os::users,
+            package::{PackageIdent, PackageInstall},
+        },
         templating::{context::RenderContext, test_helpers::*},
     };
     use std::{
@@ -680,8 +684,13 @@ mod test {
     use tempfile::TempDir;
     use toml;
 
-    const USER: &str = "hab";
-    const GROUP: &str = "hab";
+    fn curr_username() -> String {
+        users::get_current_username().expect("Can get current username")
+    }
+
+    fn curr_groupname() -> String {
+        users::get_current_groupname().expect("Can get current groupname")
+    }
 
     fn toml_from_str(content: &str) -> toml::value::Table {
         toml::from_str(content)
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/common/src/templating/config.rs
```diff
@@ -644,24 +644,27 @@ fn ensure_directory_structure(root: &Path, file: &Path, user: &str, group: &str)
     let dir = file.parent().unwrap();
 
     if !dir.exists() {
-        std::fs::create_dir_all(&dir)?;
+        std::fs::create_dir_all(&dir).map_err(Error::IO)?;
         for anc in dir.ancestors().take_while(|&d| d != root) {
-            set_permissions(&anc, &user, &group)?;
+            set_permissions(&anc, &user, &group).map_err(Error::HabitatCore)?;
```
baumanj: It's not clear to me why these `map_err`s are added. The code builds without them. Is there somewhere else we handle those specifically mapped errors?

components/common/src/templating/config.rs
```diff
@@ -644,24 +644,27 @@ fn ensure_directory_structure(root: &Path, file: &Path, user: &str, group: &str)
...
 
 fn write_templated_file(path: &Path, compiled: &str, user: &str, group: &str) -> Result<()> {
-    File::create(path).and_then(|mut file| file.write_all(compiled.as_bytes()))?;
-    set_permissions(&path, &user, &group)?;
-    Ok(())
+    match File::create(path) {
+        Ok(mut file) => file.write_all(compiled.as_bytes()).map_err(Error::IO),
+        Err(_) => Err(Error::PermissionFailed(path.to_string_lossy().into_owned())),
+    }?;
+    set_permissions(&path, &user, &group).map_err(Error::HabitatCore)
```
baumanj: Similar question here. The existing code seems to work.

components/common/src/templating/config.rs
```diff
@@ -1150,7 +1159,13 @@ mod test {
...
 
         assert_eq!(file.exists(), false);
-        write_templated_file(&file, &contents, &USER, &GROUP).expect("should fail on permissions");
+        match write_templated_file(&file, &contents, &curr_username(), &curr_groupname()) {
+            Ok(_) => panic!("should fail on permissions"),
+            Err(e) => match e {
+                Error::PermissionFailed(e) => assert_eq!(file.to_string_lossy(), e.to_string()),
+                wrong => panic!("Unexpected error returned: {:?}", wrong),
+            },
+        };
```
baumanj: Unless there are places in the code where we specifically handle `PermissionFailed` differently, I think this test (and the code above to generate the `PermissionFailed` error) is overly-specific. I'm not sure why we should be concerned about the details beyond the fact that `write_templated_file` should fail. That leads to a test which is less likely to break in the future if we change some specifics that don't invalidate the overall behavior. Plus, it makes the code simpler:
```rust
        assert!(
            write_templated_file(&file, &contents, &curr_username(), &curr_groupname()).is_err()
        );
```

jamesc: Yeah, I was trying to get better error messages during the test failure.  the `io::Error` we get back doesn't give much info, but I think it's too much for this one specific case. Happy to roll back the error mapping code

PR #5738, opened 2018-10-11
Launcher creates its own PID file at /hab/launcher/PID

baumanj:
And removes it at exit

Removing the supervisor code that manages `/hab/sup/default/LOCK` will be somewhat more involved, but at least it won't require any additional launcher changes, so this can usefully go in now.

Resolves https://github.com/habitat-sh/habitat/issues/5382

PR #6139, opened 2019-02-11
Update docs for running as non-root in a container

christophermaier:
With the merging of #5738, the Launcher began to manage its own
state (previously, the Supervisor was actually doing this). This
resulted in the need to write to `/hab/launcher` which is fine, but
ran into trouble in the very specific circumstance of a user trying to
run Habitat in a container, but as a non-root user that was also not
in the `root` group. Our documentation shows how to do this by
mounting in properly writable volumes, but was not updated to take
into account the new `/hab/launcher` directory.

Without this mount being present, nothing could start, because the
Launcher was not able to write its own PID file.

This is just a documentation change; no code needs to change at this
time.


thesentinels:

Thank you for contributing!

PR #6141, opened 2019-02-11
Move ServiceBind to core

jamesc:
* Remove a bunch of duplicated logic between sup and sup-protocol
* Remove duplicate ServiceBind from k8s exporter
* Update multiple call sites for newly private ServiceBind struct members


thesentinels:

Thank you for contributing!

components/sup-protocol/src/types.rs
```diff
@@ -159,23 +159,19 @@ impl FromStr for DesiredState {
...
+        Ok(sb.into())
+    }
+}
+
+impl From<core::service::ServiceBind> for ServiceBind {
+    fn from(bind: core::service::ServiceBind) -> Self {
+        let mut proto = ServiceBind::default();
+        proto.name = bind.name().to_string();
+        proto.service_group = ServiceGroup::from(bind.service_group().clone());
+        proto
```
baumanj: I'm not sure why the code in this file seems to all follow this really clunky pattern of starting with a default and assigning each field with a separate statement. We can do the safer and more elegant thing:
```rust
        Self {
            name: bind.name().to_string(),
            service_group: ServiceGroup::from(bind.service_group().clone()),
        }
```
Perhaps this wasn't possible before we transitioned to Prost.

components/sup/src/manager/service/spec.rs
```diff
@@ -124,7 +124,12 @@ impl IntoServiceSpec for protocol::ctl::SvcLoad {
             spec.update_strategy = UpdateStrategy::from_i32(update_strategy).unwrap_or_default();
         }
         if let Some(ref list) = self.binds {
-            spec.binds = list.binds.clone().into_iter().map(Into::into).collect();
+            spec.binds = list
+                .binds
+                .clone()
+                .into_iter()
```
baumanj: Both the `clone` and `into_iter` can be omitted in favor of a simple `iter`. We only need to clone the `service_group` field, and we can do it like this:
```rust
                .iter()
                .map(|b| ServiceBind::new(&b.name, b.service_group.clone().into()))
```
Note, I also removed the superfluous `hcore::service::` since there's already a
```rust
use crate::hcore::service::{…, ServiceBind};
```
in this file. If it was intentional to show the type conversion, I think it's more illustrative to fully qualify *both* types:
```rust
                .map(|pb: &habitat_sup_protocol::types::ServiceBind| {
                    hcore::service::ServiceBind::new(&pb.name, pb.service_group.clone().into())
                })
```

components/sup/src/manager/service/spec.rs
```diff
@@ -249,7 +254,7 @@ impl ServiceSpec {
     /// * If any given service binds are in neither required nor optional package binds
     fn validate_binds(&self, package: &PackageInstall) -> Result<()> {
         let mut svc_binds: HashSet<String> =
-            HashSet::from_iter(self.binds.iter().cloned().map(|b| b.name));
+            HashSet::from_iter(self.binds.iter().map(|b| b.name().to_string()));
```
baumanj: While it is a bit longer in this case, I prefer this form and it obviates the need for creation of closures:
```suggestion
            HashSet::from_iter(self.binds.iter().map(ServiceBind::name).map(str::to_string));
```
Perhaps the compiler can optimize this away, but I alway like to hand it pre-existing functions when they exist to do the job.

baumanj: Additionally, you might be able to get away with:
```suggestion
            HashSet::from_iter(self.binds.iter().map(ServiceBind::name));
```
Making svc_binds a `HashSet<&str>`. It requires other changes, but avoids unnecessary cloning. I haven't worked it out completely, but I think it should be possible.

components/sup/src/manager/service/spec.rs
jamesc: Yeah, it was to show the type conversion.  Good idea on showing both types

components/sup-protocol/src/types.rs
jamesc: Also fixed for a couple of the other types at the same time

components/pkg-export-kubernetes/src/cli.rs
```diff
@@ -206,6 +208,16 @@ fn valid_natural_number(val: String) -> result::Result<(), String> {
...
 
+fn valid_bind(val: String) -> result::Result<(), String> {
+    match ServiceBind::from_str(&val) {
+        Ok(_) => Ok(()),
+        Err(_) => Err(format!(
+            "'{}' is not valid. Service binds have the format name:service.group",
+            &val
+        )),
+    }
+}
```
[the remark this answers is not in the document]
jamesc: I'd argue for leaving this here for now.  Looking at the k8s exporter code it could probably do with a cleanup generally.  I'll add a comment to say we could move it to core.

components/sup/src/manager/service/spec.rs
jamesc: Yeah, this made the code better.  I also added some tests around this as the logic was all untested.

PR #6142, opened 2019-02-12
ensure install hooks do not run from cli if feature is not enabled

mwrock:
This was something that was in my original PR and removed after some code review comments. I just noticed looking at the windows worker builds that the install hooks are running in their studios even though the feature is not enabled. I now recall why this conditional is important.

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -668,11 +668,12 @@ fn sub_pkg_install(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
         LocalPackageUsage::default()
     };
 
-    let install_hook_mode = if m.is_present("IGNORE_INSTALL_HOOK") {
-        InstallHookMode::Ignore
-    } else {
-        InstallHookMode::default()
-    };
+    let install_hook_mode =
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/main.rs
```diff
@@ -668,7 +668,7 @@ fn sub_pkg_install(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
         LocalPackageUsage::default()
     };
 
-    let install_hook_mode = if m.is_present("IGNORE_INSTALL_HOOK") {
+    let install_hook_mode = if !feat::is_enabled(feat::InstallHook) || m.is_present("IGNORE_INSTALL_HOOK") {
```
baumanj: Since `default` seems more like the happy path, I find it somewhat simpler to think about it in these terms:
```rust
    let install_hook_mode =
        if feat::is_enabled(feat::InstallHook) && !m.is_present("IGNORE_INSTALL_HOOK") {
            InstallHookMode::default()
        } else {
            InstallHookMode::Ignore
        };
```
But the two are equivalent, so either way is fine.

PR #6144, opened 2019-02-12
Prevent a crash in Supervisor protocol processing

christophermaier:
The Supervisor protocol, which mediates the communication of the `hab` binary with a running instance of the Supervisor, implicitly assumed that all message content was valid UTF-8 by calling `unwrap()` unconditionally on the output of [std::str::from_utf8][1].

If non-UTF8 bytes are present, however, this will ultimately cause the Supervisor's control gateway to crash and become non-responsive.

This change converts a failure to parse valid UTF-8 into a custom `std::io::Error`, which allows the error to be recognized while also allowing processing to continue. Also see the documentation for [tokio_codec::Decoder::decode][2] for additional detail.

This commit also rearranges our `use` statements to take advantage of Rust 2018 features, and also adds a few more `trace` logging statements for added debuggability. These can be activated by adding `habitat_sup_protocol::codec=trace` to the `RUST_LOG` environment variable.

Thanks to @stevendanna for discovering this bug!

[1]: http://doc.rust-lang.org/1.32.0/std/string/struct.String.html#method.from_utf8
[2]: https://docs.rs/tokio-codec/0.1.1/tokio_codec/trait.Decoder.html#tymethod.decode

### DIFF
--- a/components/sup-protocol/src/codec.rs
+++ b/components/sup-protocol/src/codec.rs
@@ -54,11 +54,9 @@
 //! Contains the actual payload of the message encoded using Google
 //! [Protobuf 2](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec).
 
-use std::{fmt,
-          io::{self,
-               Cursor},
-          str};
-
+use crate::{message::MessageStatic,
+            net::{NetErr,
+                  NetResult}};
 use bytes::{Buf,
             BufMut,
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

christophermaier:
@baumanj I agree with you 100% on both counts.

PR #6151, opened 2019-02-12
Adds printing of target in job status output

eeyun:
Should not get merged until https://github.com/habitat-sh/builder/pull/933 merges

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -85,6 +85,7 @@ pub struct Project {
     pub ident: String,
     pub state: String,
     pub job_id: String,
+    pub target: String,
 }
 
 #[derive(Clone, Deserialize)]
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-client/src/lib.rs
```diff
@@ -84,6 +84,7 @@ pub struct Project {
     pub ident: String,
     pub state: String,
     pub job_id: String,
+    pub target: String,
```
smacfarlane: Would there be any benefit to using a PackageTarget type here instead of a String?

eeyun: In this case I don't believe so. I _think_ we could make a change like that, but not as a part of this PR. That would be a significant refactor of the way we're handling all of these pieces of data in the cli/api client.

PR #6152, opened 2019-02-13
Fix verification pipeline

scotthain:


thesentinels:

Thank you for contributing!

components/butterfly/tests/integration.rs
```diff
@@ -143,11 +143,16 @@ fn six_members_unmeshed_allows_graceful_departure() {
...
+// error.
+//
+// #[test]
+// #[cfg_attr(feature = "ignore_inconsistent_tests", ignore)]
+// fn fifty_members_meshed_confirm_one_member() {
+//     let mut net = btest::SwimNet::new(50);
+//     net.mesh();
+//     net[0].pause();
+//     assert_wait_for_health_of!(net, 0, Health::Confirmed);
+// }
```
baumanj: I don't see any real value of this test. I'm guessing that it's a question of ZMQ resources that are causing issues. If it's reliable with 50 changed to 10, or even 5, I think it would be just as helpful.

PR #6163, opened 2019-02-15
Add Launcher build stages to the release pipeline

christophermaier:
The pipeline will now ask if we need to build new Launcher artifacts
for this release. If we do, build steps will automatically be added;
if we don't, we'll just add the currently stable Launcher artifacts to
the release channel.

This adds automation that previously did not exist, and also *removes*
the manual step where we had to supply fully-qualified identifiers for
all Launcher packages we wanted to promote into the release
channel. Wins all around!


thesentinels:

Thank you for contributing!

christophermaier:
Added the `DO-NOT-MERGE` label because there's still some testing that needs to be done, as well as addition of the Windows Service build.

.buildkite/launcher_build_steps.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - docker#v2.0.0:
+          image: "chefes/windows-base:latest"
+          shell: [ "powershell", "-Command" ]
+          environment:
+            - HAB_BLDR_CHANNEL
+            - HAB_AUTH_TOKEN
+            - BUILDKITE_JOB_ID
+            - BUILDKITE_AGENT_ACCESS_TOKEN
+
+  - wait
```
mwrock: Assuming each build is serialized? We should build the windows-service after the windows launcher.

.buildkite/launcher_build_steps.yaml
```diff
@@ -0,0 +1,33 @@
...
+    env:
+      HAB_BLDR_CHANNEL: unstable
+    plugins:
+      - docker#v2.0.0:
+          image: "chefes/windows-base:latest"
+          shell: [ "powershell", "-Command" ]
+          environment:
+            - HAB_BLDR_CHANNEL
+            - HAB_AUTH_TOKEN
+            - BUILDKITE_JOB_ID
```
mwrock: please add `-HAB_CRYPTO_KEY`

.buildkite/launcher_build_steps.yaml
christophermaier: Yup, there's a space immediately after that line where we'll add it :) 

Need to coordinate with @smacfarlane for that, so it'll probably come after this PR.

PR #6172, opened 2019-02-18
[sup] Enable client-authentication with TLS

stevendanna:
TLS allows clients to also provide a certificate when initiating a
connection. This certifcate can be used to authenticate clients as an
alternative to other HTTP authentication schems such as basic auth.

I've factored out the TLS configuration into a struct because passing
around `Option<(PathBuf, PathBuf, Option<PathBuf>)>` felt like it
would become a bit error prone and hard to extend.

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -90,7 +90,9 @@ use prometheus::{HistogramVec,
                  IntGauge,
                  IntGaugeVec};
 use rustls::{internal::pemfile,
+             AllowAnyAuthenticatedClient,
              NoClientAuth,
+             RootCertStore,
              ServerConfig};
 use serde_json;
 use std::{self,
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -1239,29 +1248,42 @@ impl Manager {
...
+        Some(ref path) => {
+            let mut root_store = RootCertStore::empty();
+            let ca_file = &mut BufReader::new(File::open(&path)?);
+            let (added, _) = root_store
+                .add_pem_file(ca_file)
+                .map_err(|_| sup_error!(Error::InvalidCertFile(path.to_path_buf())))?;
+            if added < 1 {
+                return Err(sup_error!(Error::InvalidCertFile(path.to_path_buf())));
+            }
+            AllowAnyAuthenticatedClient::new(root_store)
```
baumanj: We can reduce the duplication here with `and_then`:
```rust
            root_store
                .add_pem_file(ca_file)
                .and_then(|(added, _)| {
                    if added < 1 {
                        Err(())
                    } else {
                        Ok(AllowAnyAuthenticatedClient::new(root_store))
                    }
                })
                .map_err(|_| sup_error!(Error::InvalidCertFile(path.clone())))?
```

stevendanna:
Updated the style issues and changed the flag requirements so that you get an error if you try to turn on client-auth but haven't turned on the other tls flags.  To be honest, it will be a while before we have time to figure out how to transition to turning this on so no rush.

baumanj:
Our merge freeze is lifted now. Let me know when you've rebased to fix the conflict, and I can get this merged, @stevendanna

stevendanna:
Rebased and re-rustfmt'd

baumanj:
Thanks again!

PR #6176, opened 2019-02-19
Add --no-listen-gossip flag to prevent gossip layer from starting up

mpeck:

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -886,8 +886,11 @@ pub fn sub_sup_run() -> App<'static, 'static> {
     (about: "Run the Habitat Supervisor")
     (@group GOSSIP =>
      (@arg NO_LISTEN_GOSSIP: --("no-listen-gossip") "Don't startup the gossip listener.")
+     // We have to hardcode the env and default bits in the usage string here because setting them with them
+     // env and default_value functions causes the exclusivity rule between NO_LISTEN_GOSSIP and LISTEN_GOSSIP
+     // to fail. https://github.com/clap-rs/clap/issues/1205
      (@arg LISTEN_GOSSIP: --("listen-gossip") +takes_value {valid_socket_addr}
-      "The listen address for the Gossip System Gateway.")
+      "The listen address for the Gossip System Gateway. [env: HAB_LISTEN_GOSSIP] [default: 0.0.0.0:9638]")
     )
     (@arg LISTEN_HTTP: --("listen-http") env(LISTEN_HTTP_ADDRESS_ENVVAR) default_value(&LISTEN_HTTP_DEFAULT_ADDR) {valid_socket_addr}
         "The listen address for the HTTP Gateway.")
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -707,7 +707,7 @@ pub fn sup_commands() -> App<'static, 'static> {
     // set custom usage string, otherwise the binary
     // is displayed as the clap_app name, which may or may not be different.
     // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
-    (usage: "hab sup <SUBCOMMAND>")
+    (bin_name: "hab sup")
     (@setting VersionlessSubcommands)
     (@setting SubcommandRequiredElseHelp)
     (subcommand: sub_sup_bash().aliases(&["b", "ba", "bas"]))
@@ -884,9 +884,6 @@ pub fn sub_sup_bash() -> App<'static, 'static> {
 pub fn sub_sup_run() -> App<'static, 'static> {
     clap_app!(@subcommand run =>
     (about: "Run the Habitat Supervisor")
-    // set custom usage string, otherwise the binary
-    // is displayed confusingly as `hab-sup`
-    // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
     (@group GOSSIP =>
      (@arg NO_LISTEN_GOSSIP: --("no-listen-gossip") "Don't startup the gossip listener.")
      (@arg LISTEN_GOSSIP: --("listen-gossip") +takes_value {valid_socket_addr}
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -29,8 +29,8 @@ use crate::{
     command::studio,
     common::{
         cli_defaults::{
-            GOSSIP_DEFAULT_ADDR, GOSSIP_LISTEN_ADDRESS_ENVVAR, LISTEN_CTL_DEFAULT_ADDR_STRING,
-            LISTEN_HTTP_ADDRESS_ENVVAR, LISTEN_HTTP_DEFAULT_ADDR, RING_ENVVAR, RING_KEY_ENVVAR,
+            LISTEN_CTL_DEFAULT_ADDR_STRING, LISTEN_HTTP_ADDRESS_ENVVAR, LISTEN_HTTP_DEFAULT_ADDR,
+            RING_ENVVAR, RING_KEY_ENVVAR,
         },
         types::{EnvConfig, ListenCtlAddr},
     },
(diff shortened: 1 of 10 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -720,18 +728,20 @@ impl Manager {
...
-            self.census_ring.update_from_rumors(
-                &self.butterfly.service_store,
-                &self.butterfly.election_store,
-                &self.butterfly.update_store,
-                &self.butterfly.member_list,
-                &self.butterfly.service_config_store,
-                &self.butterfly.service_file_store,
-            );
+            if let Some(ref butterfly) = self.butterfly {
+                Self::update_peers_from_watch_file(butterfly, &self.peer_watcher)?;
```
raskchanky: I haven't dug in enough to fully grok the reasons why, but off the cuff, it strikes me as odd that this went from a struct level function/method to a type level function.  What was the reasoning behind that?

mpeck: Ah, good catch. There was a borrow error because I wasn't doing the pattern matching properly on line 733. Changing the function was the only way I could figure out to make it work. I fixed the bit on 733 though, so this shouldn't be necessary.

components/sup/src/manager/mod.rs
```diff
@@ -523,9 +528,11 @@ impl Manager {
             return;
         }
 
-        self.gossip_latest_service_rumor(&service);
-        if service.topology == Topology::Leader {
-            self.butterfly.start_election(&service.service_group, 0);
+        if let Some(ref butterfly) = self.butterfly {
+            self.gossip_latest_service_rumor(&service);
```
baumanj: There's an `if let Some(…) = self.butterfly` conditional both here *and* inside `gossip_latest_service_rumor`. For simplicity and testability, I'm a fan of having code be either:

1. Logic with loops and branches that encode behavior
1. Lists of mostly unconditional commands that plumb together different chunks of (1)

Try to minimize mixing (1) and (2). The (1)-type code is where the testing should be focused, so making that code as purely functional as possible, with a minimum of side effects or superfluous state carried around will make the testing easiest. The (2)-type code won't really have any behavior to test and with a language like rust, errors in plumbing will nearly all be caught by the compiler.

Despite it having a conditional in the form of `if let`, I would call this more plumbing (2) than behavioral logic (1), so keeping it at the higher level and making `gossip_latest_service_rumor` more pure by passing it the specific data it needs to operate on rather than all of `self` should eliminate the redundancy.

components/sup/src/manager/mod.rs
baumanj: Now that we're changing `self.butterfly` to be an `Option`, I think this can be a good pattern to avoid having `if let Some(…) = self.butterfly` at all levels of the code and passing only the specific data functions *need* into associated functions (i.e., `Self::f()`) rather than using methods  (i.e., `self.f()`) which carry around superfluous state and require more places where we encounter the `Option`. See https://github.com/habitat-sh/habitat/pull/6176#discussion_r258524765.

components/hab/src/cli.rs
```diff
@@ -887,9 +887,11 @@ pub fn sub_sup_run() -> App<'static, 'static> {
     // set custom usage string, otherwise the binary
     // is displayed confusingly as `hab-sup`
     // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
-    (usage: "hab sup run [FLAGS] [OPTIONS] [--] [PKG_IDENT_OR_ARTIFACT]")
```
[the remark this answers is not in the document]
mpeck: It was actually accidental. I removed it when I was testing. But I did not know about `bin_name`. Perhaps it will tern out to be a happy accident.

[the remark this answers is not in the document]
baumanj: image

components/hab/src/cli.rs
```diff
@@ -887,9 +887,11 @@ pub fn sub_sup_run() -> App<'static, 'static> {
...
-          (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
-        "The listen address for the Gossip System Gateway.")
+    (@group GOSSIP =>
+     (@arg NO_LISTEN_GOSSIP: --("no-listen-gossip") "Don't startup the gossip listener.")
+     (@arg LISTEN_GOSSIP: --("listen-gossip") +takes_value {valid_socket_addr}
+      "The listen address for the Gossip System Gateway.")
```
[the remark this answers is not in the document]
mpeck: Oops I forgot to add that. This still irritates me a bit; we have to hardcode it in the string. I can't even use the constants for the env var string or the default value. Maybe switching over to the builder syntax will help with some of that?

components/sup/src/main.rs
```diff
@@ -239,17 +251,7 @@ fn mgrcfg_from_sup_run_matches(m: &ArgMatches) -> Result<ManagerConfig> {
...
-                let default = sup::config::GossipListenAddr::default();
-                error!(
-                    "Value for LISTEN_GOSSIP has not been set. Using default: {}",
-                    default
-                );
-                Ok(default)
-            },
-            |v| v.parse(),
-        )?,
+        gossip_listen,
```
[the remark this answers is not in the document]
mpeck: No real reason. I just thought it was easier to read the other way. I'm not married to it.

components/sup/src/manager/mod.rs
baumanj: Here's the general pattern I think we should try to avoid:
```rust
        if let Some(ref butterfly) = self.butterfly {
            self.some_function_that_needs_butterfly(…);
```
This is how we end up exploding the number of redundant `if let` conditionals with `self.butterfly`. Inside any of those conditionals, if we're calling functions that need butterfly, we should provide it as a parameter. This gives is a purer, more testable function and avoids the need for redundant checks.

components/sup/src/manager/commands.rs
```diff
@@ -195,22 +182,27 @@ pub fn service_file_put(
...
-        }
-    };
-    match client.send_service_file(service_group, filename, version, &content, is_encrypted) {
-        Ok(()) => {
-            req.reply_complete(net::ok());
-            return Ok(());
+    create_butterfly_client(mgr)?
+        .send_service_file(service_group, filename, version, &content, is_encrypted)
+        .map(|_| req.reply_complete(net::ok()))
+        .map_err(|e| net::err(ErrCode::Internal, e.to_string()))
```
baumanj: I like the way you refactored this to reduce the code duplication! I think you could go even further, since 3 of these 4 lines are repeated 3 times. In order to put the emphasis on what's *different*, you could further factor out the rest of the error handling. I poked at it a bit and was able to get this to compile:
```rust
    with_butterfly_client(mgr, req, |mut client| {
        client.send_service_file(
            service_group.clone(),
            filename.clone(),
            version,
            &content,
            is_encrypted,
        )
    })
```
where `with_butterfly_client` is like:
```rust
fn with_butterfly_client(
    mgr: &ManagerState,
    req: &mut CtlRequest,
    f: impl Fn(butterfly::client::Client) -> butterfly::error::Result<()>,
) -> NetResult<()> {
    … // basically what you had already inside
```
With a bit more work, I imagine you could get rid of the clones as well, though for such small pieces of data attached to user-initiated commands, it may not be worthwhile.

However, there's an alternative to consider here: is this the right level to put the conditional handling of `mgr.cfg.gossip_listen`? Conceptually, a single check for whether we have a gossip server running at the point commands are dispatched seems to reduce the complexity, though if you looked at https://github.com/habitat-sh/habitat/blob/master/components/sup/src/ctl_gateway/server.rs#L303, I can certainly understand why you didn't go that route: it's far from obvious to see how to get at `ManagerState` at that level. Given the state of that code (which @christophermaier is in the midst of refactoring), I think your current approach is pretty reasonable, but I'd like to see an issue filed to revisit this once the control gateway code has been simplified because I think that's the more appropriate level to handle this.

mpeck:
Closing for now in favor of #6209

PR #6196, opened 2019-02-21
Tests for hab sup run cli behavior

mpeck:
Here are some tests and tweaks for the `hab sup run` cli command. I'm trying to capture the existing behavior here in preparation of switching away from the macro when defining the cli. I tried to not change much of the existing behavior, but I did make a couple tweaks around default values and I fixed one bug.

The reason for making this change is to get some functionality that does not seem to work when using the macro syntax; `default_value_if` specifically (https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.default_value_if). Also, I'm not a fan of the macro syntax :|

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -943,14 +943,14 @@ pub fn sub_sup_run() -> App<'static, 'static> {
     (@arg GROUP: --group default_value("default")
         "The service group; shared config and topology.")
     (@arg TOPOLOGY: --topology -t +takes_value possible_value[standalone leader]
-        "Service topology; [default: none]")
-    (@arg STRATEGY: --strategy -s default_value("none") {valid_update_strategy}
-        "The update strategy; [values: none, at-once, rolling]")
+        "Service topology.")
+    (@arg STRATEGY: --strategy -s possible_value("none") possible_value("at-once") possible_value("rolling") default_value("none")
+        "The update strategy.")
     (@arg BIND: --bind +takes_value +multiple
         "One or more service groups to bind to a configuration")
-    (@arg BINDING_MODE: --("binding-mode") default_value("strict") {valid_binding_mode}
+    (@arg BINDING_MODE: --("binding-mode") default_value("strict") possible_value[relaxed strict]
         "Governs how the presence or absence of binds affects service startup. `strict` blocks \
-         startup until all binds are present. [values: relaxed, strict]")
+         startup until all binds are present.")
     (@arg VERBOSE: -v "Verbose output; shows file and line/column numbers")
     (@arg NO_COLOR: --("no-color") "Turn ANSI color off")
     (@arg JSON: --("json-logging") "Use structured JSON logging for the Supervisor. \
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -888,21 +888,20 @@ pub fn sub_sup_run() -> App<'static, 'static> {
     // is displayed confusingly as `hab-sup`
     // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
     (usage: "hab sup run [FLAGS] [OPTIONS] [--] [PKG_IDENT_OR_ARTIFACT]")
-          (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
+    (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
         "The listen address for the Gossip System Gateway.")
     (@arg LISTEN_HTTP: --("listen-http") env(LISTEN_HTTP_ADDRESS_ENVVAR) default_value(&LISTEN_HTTP_DEFAULT_ADDR) {valid_socket_addr}
         "The listen address for the HTTP Gateway.")
     (@arg HTTP_DISABLE: --("http-disable") -D
         "Disable the HTTP Gateway completely [default: false]")
     (@arg LISTEN_CTL: --("listen-ctl") env(ListenCtlAddr::ENVVAR) default_value(&LISTEN_CTL_DEFAULT_ADDR_STRING) {valid_socket_addr}
-        "The listen address for the Control Gateway. If not specified, the value will \
-        be taken from the HAB_LISTEN_CTL environment variable if defined. [default: 127.0.0.1:9632]")
+        "The listen address for the Control Gateway.")
     (@arg ORGANIZATION: --org +takes_value
         "The organization that the Supervisor and its subsequent services are part of.")
     (@arg PEER: --peer +takes_value +multiple
         "The listen address of one or more initial peers (IP[:PORT])")
     (@arg PERMANENT_PEER: --("permanent-peer") -I "If this Supervisor is a permanent peer")
-    (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with[peer]
+    (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with("PEER")
         "Watch this file for connecting to the ring"
     )
     (@arg RING: --ring -r env(RING_ENVVAR) conflicts_with("RING_KEY")
@@ -917,10 +916,9 @@ pub fn sub_sup_run() -> App<'static, 'static> {
                   GCrBOW6CCN75LMl0j2V5QqQ6nNzWm6and9hkKBSUFPI=')")
     (@arg CHANNEL: --channel +takes_value
         "Receive Supervisor updates from the specified release channel [default: stable]")
-    (@arg BLDR_URL: -u --url +takes_value {valid_url}
+    (@arg BLDR_URL: -u --url default_value("https://bldr.habitat.sh") {valid_url}
         "Specify an alternate Builder endpoint. If not specified, the value will \
-         be taken from the HAB_BLDR_URL environment variable if defined. (default: \
-         https://bldr.habitat.sh)")
+         be taken from the HAB_BLDR_URL environment variable if defined.")
 
     (@arg CONFIG_DIR: --("config-from") +takes_value {dir_exists}
         "Use package config from this path, rather than the package itself")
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -888,21 +888,20 @@ pub fn sub_sup_run() -> App<'static, 'static> {
...
-        "The listen address for the Control Gateway. If not specified, the value will \
-        be taken from the HAB_LISTEN_CTL environment variable if defined. [default: 127.0.0.1:9632]")
+        "The listen address for the Control Gateway.")
     (@arg ORGANIZATION: --org +takes_value
         "The organization that the Supervisor and its subsequent services are part of.")
     (@arg PEER: --peer +takes_value +multiple
         "The listen address of one or more initial peers (IP[:PORT])")
     (@arg PERMANENT_PEER: --("permanent-peer") -I "If this Supervisor is a permanent peer")
-    (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with[peer]
+    (@arg PEER_WATCH_FILE: --("peer-watch-file") +takes_value conflicts_with("PEER")
```
mpeck: The conflict wasn't actually working with the previous syntax. I wrote a test to confirm that this fixes the problem.

components/sup/src/cli.rs
```diff
@@ -44,6 +104,163 @@ mod test {
...
+
+        assert_cli_cmd!(auto_update_should_have_a_short_flag,
+                        "hab-sup run -A",
+                        "AUTO_UPDATE" => true);
+
+        // TODO: key file cert file test
+
+        // TODO: key file requires test
+
+        // TODO: cert file requires test
```
mpeck: The TODO's here all involve options that require interacting with the file system. I wasn't sure the best way to handle that. Suggestions would be appreciated.

components/sup/src/cli.rs
```diff
@@ -44,6 +105,224 @@ mod test {
...
+                        "hab-sup run -u http://not.a.real.com",
+                        "BLDR_URL" => "http://not.a.real.com");
+
+        assert_cli_cmd!(bldr_url_should_have_a_default_value,
+                        "hab-sup run",
+                        "BLDR_URL" => "https://bldr.habitat.sh");
+
+        #[test]
+        fn invalid_bldr_url_should_fail() {
+            let cmd_vec = Vec::from_iter("hab-sup run -u bad-url".split_whitespace());
```
baumanj: Similar to the addr args, a more plausibly bad URL would be a more valuable test:
```suggestion
            let cmd_vec = Vec::from_iter("hab-sup run -u https:/bldr.habitat.sh".split_whitespace());
```
I believe this example actually *does* pass our validation indicating we may want to do something a bit stricter than `Url::parse`, depending on how we're consuming that value.

components/sup/src/cli.rs
```diff
@@ -44,6 +105,224 @@ mod test {
...
+        assert_cli_cmd!(should_handle_auto_update,
+                        "hab-sup run --auto-update",
+                        "AUTO_UPDATE" => true);
+
+        assert_cli_cmd!(auto_update_should_have_a_short_flag,
+                        "hab-sup run -A",
+                        "AUTO_UPDATE" => true);
+
+        #[test]
+        fn should_take_cert_file_and_key_file() {
```
baumanj: This seems like too much logic to be worthwhile. The clap declarations `{file_exists} requires[CERT_FILE]` and `{file_exists} requires[KEY_FILE]` are pretty straightforward. Similarly for the next two tests.

components/sup/src/cli.rs
```diff
@@ -44,6 +105,224 @@ mod test {
...
+                        "hab-sup run --topology leader",
+                        "TOPOLOGY" => "leader");
+
+        assert_cli_cmd!(topology_should_have_a_short_flag,
+                        "hab-sup run -t standalone",
+                        "TOPOLOGY" => "standalone");
+
+        #[test]
+        fn topology_should_fail_when_a_bad_value_is_passed() {
+            let cmd_vec = Vec::from_iter("hab-sup run --topology foobar".split_whitespace());
```
baumanj: A more likely wrong value would be if someone got `--topology` confused with strategy, so I'd test `at-once` and  `rolling`. Similarly (but reversed) for `--strategy`.

components/hab/src/cli.rs
```diff
@@ -917,10 +916,9 @@ pub fn sub_sup_run() -> App<'static, 'static> {
                   GCrBOW6CCN75LMl0j2V5QqQ6nNzWm6and9hkKBSUFPI=')")
     (@arg CHANNEL: --channel +takes_value
         "Receive Supervisor updates from the specified release channel [default: stable]")
-    (@arg BLDR_URL: -u --url +takes_value {valid_url}
+    (@arg BLDR_URL: -u --url default_value("https://bldr.habitat.sh") {valid_url}
```
baumanj: These are good improvements, but if we're making them to an argument that gets used many places we should update *all* the instances at once.

Similarly for `--group`, `--strategy`, `--binding-mode` and `--health-check-interval`

components/hab/src/cli.rs
```diff
@@ -942,23 +940,23 @@ pub fn sub_sup_run() -> App<'static, 'static> {
...
         "Environment name; [default: not set].")
-    (@arg GROUP: --group +takes_value
-        "The service group; shared config and topology [default: default].")
+    (@arg GROUP: --group default_value("default")
+        "The service group; shared config and topology.")
     (@arg TOPOLOGY: --topology -t +takes_value possible_value[standalone leader]
         "Service topology; [default: none]")
-    (@arg STRATEGY: --strategy -s +takes_value {valid_update_strategy}
-        "The update strategy; [default: none] [values: none, at-once, rolling]")
+    (@arg STRATEGY: --strategy -s default_value("none") {valid_update_strategy}
```
baumanj: I think we can get rid of the explicit values listing on the next line if we do it this way:
```suggestion
    (@arg STRATEGY: --strategy -s default_value("none") {valid_update_strategy} possible_value("at-once") possible_value("rolling")
```
I'm not sure if that means we should eliminate `valid_update_strategy` or not. It's just a question of more redundant code versus having an error caught a bit earlier.

Similarly for `BINDING_MODE`.

components/sup/src/cli.rs
mpeck: I agree that it seems like `Url::parse` is a bit too lenient, but I don't want to tackle that in this PR. Would prefer to keep this one relatively focused on capturing the existing behavior.

baumanj: That's fair; please file an issue for the follow-up.

components/sup/src/cli.rs
```diff
@@ -30,6 +30,67 @@ mod test {
...
+
+        #[test]
+        fn invalid_listen_ctl_address_should_fail() {
+            let cmd_vec = Vec::from_iter("hab-sup run --listen-ctl bad-address".split_whitespace());
+            assert!(cli().get_matches_from_safe(cmd_vec).is_err());
+        }
+
+        assert_cli_cmd!(should_handle_org,
+                        "hab-sup run --org monkeypants",
+                        "ORGANIZATION" => "monkeypants");
```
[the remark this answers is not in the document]
mpeck: Generally I would agree with you, but the primary purpose for these tests are in preparation to refactor `sub_sup_run` and remove the macro. I would prefer to have something for each option just so I can make sure that I'm getting the proper functionality and not leaving anything out when I do the refactor.

components/sup/src/cli.rs
```diff
@@ -44,6 +105,224 @@ mod test {
...
+            let cmd_str = format!("hab-sup run --config-from {}", path_str);
+            let cmd_vec = Vec::from_iter(cmd_str.split_whitespace());
+            let matches = cli()
+                .get_matches_from_safe(cmd_vec)
+                .expect("Could not get ArgMatches");
+            let subcommand = matches
+                .subcommand_matches("run")
+                .expect("Failed to get matches for sup run");
+            assert_eq!(subcommand.value_of("CONFIG_DIR"), Some(path_str));
+        }
```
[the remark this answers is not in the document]
mpeck: Yeah I agree that it's a bit overkill, but I want to make sure I've captured as much functionality as possible. I'm not opposed to revisiting these tests after we're sure that everything is working properly and pruning them a bit.

components/sup/src/cli.rs
mpeck: Yeah it is a good bit of logic, but I want to make sure that I've captured everything. I don't want to refactor `hab_sup_run` and accidentally leave something out.

components/hab/src/cli.rs
mpeck: Wow, you're right, we do use this option in a lot of spots. I foresee more work in testing and refactoring the cli bits to not use the macro. I would prefer to keep this one as focused as I can and tackle the other occurrences as I come to them. My hope is that using the builder style syntax will allow us to reuse more code and maybe eliminate a lot of the copy/pase stuff necessary when using the macro.

components/hab/src/cli.rs
mpeck: I thought about that, but the way the validation function was working made me pause. My thinking was that switching to the builder style pattern for the cli would make it easier to leverage `protocol::types::UpdateStrategy`. The more I look at it though, the more that I think that you may be right.

components/sup/src/cli.rs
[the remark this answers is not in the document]
baumanj: Unfortunately, testing is no guarantee here. If you miss a thing to test or make an error in the test code, you can let defects through as well. Beware of adding tests giving you more confidence than they should: test code is code, and at the end of the day the only way we can truly have confidence in test code's correctness is through human inspection and experience running both the test code and the system under test.

I do think the general pattern of adding good tests before a refactor is a good one, but I *don't* believe in the idea of adding tests that you wouldn't otherwise want to just have around, especially since we don't have any special knowledge of the kinds of defects this sort of refactoring is likely to introduce. Perhaps as we start the macro → builder refactor, we will learn some things which inspire more targeted testing, but for now, let's add tests that bring value to the code as it is and refrain from adding tests whose correctness is harder to verify than the code they are testing.

components/sup/src/cli.rs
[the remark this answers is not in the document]
baumanj: See 
https://github.com/habitat-sh/habitat/pull/6196#discussion_r259510217

components/hab/src/cli.rs
baumanj: If you want to make these improvements in a later PR, that's fine. But if that's the case, don't improve *any* of them. I think inconsistency is worse because when people later search to try to fix things up, it'll be harder to find them all because they'll look different.

components/sup/src/cli.rs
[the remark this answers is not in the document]
mpeck: I agree in principle, but I disagree in this specific case. That said, I'm fine with conceding here if you feel it would be better.

components/hab/src/cli.rs
mpeck: So after this PR merges the next thing is to refactor this function. I agree that inconsistency is a bad thing, but I don't think it's unreasonable to take refactors in small bits.

components/sup/src/cli.rs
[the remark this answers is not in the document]
baumanj: I do. We have a lot of code, and every bit that we add increases the weight of what we have to maintain. I'm not convinced these changes pay for themselves, so let's leave them out unless something changes that calculus.

components/hab/src/cli.rs
baumanj: Nothing wrong with small pieces, but the nature of those pieces is important since work often gets stopped not exactly where we were planning it to. Every change should be conceptually atomic. So, it's fine to make that unit "fix all of `hab sup run`" or "fix all the `--url` args", but when there's an intersection like this, I don't like the resulting situation of introducing inconsistency among pieces of code that that are semantically identical. Breaking it down like:

1. Fix all the `hab sup run`-specific bits
1. Fix all the `--url` args
1. Fix all the `--group` args
1. etc…

Seems cleaner to me

components/sup/src/cli.rs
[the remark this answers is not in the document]
mpeck: I removed the presence and short flag tests. I'm gonna leave in the conflicts_with tests for now though; those tests already helped me find a bug.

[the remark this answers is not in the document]
baumanj: I understand the impulse to keep tests that helped you find a bug, but I'm not certain that's the most effective way to evaluate whether it makes sense to keep a particular test in the codebase. It's great that we found this bug (and good also that it's the only instance of this particular class of bug), but for tests that we _leave in_ the codebase, I think they should satisfy one of the following criteria:

1. Validating behavior to catch a likely regression from a future change
1. Documenting desired behavior that is made clearer through example

There's certainly an argument to be made that this test satisfies (1):
```rust
        #[test]
        fn peer_watch_file_conflicts_with_peer() {
            let cmd_vec = Vec::from_iter(
                "hab-sup run --peer-watch-file foobar --peer 1.1.1.1:1111".split_whitespace(),
            );
            assert!(cli().get_matches_from_safe(cmd_vec).is_err());
        }
```
But `conflicts_with` isn't our logic, it's a part of clap. And our bug was that we weren't using it correctly. If we add a 3-line test for every time we use `conflicts_with`, we end up with a lot of test code to manage. And any time we want to change any of the places in the code that use `conflicts_with`, we also have to change the test code. But that probably won't be recognized until the test fails in CI, causing delay. There are lots of ways we can have bugs on account of misusing a library we depend on, and I don't think that adding tests for all of them works well to help address that issue. There's alway a balance between testing thoroughly to avoid the introduction of bugs, and not testing so specifically as to hamper development on a codebase we expect to change.

I think it's great we found this bug, and in that sense, I think going through writing that test had value. But I think the tests we keep should be the ones we expect to have value in the future that exceeds the cost of maintaining them. I think this test's value was already realized and now that you've fixed the bug, I think future value is maximized by keeping the set of tests a bit leaner.

mpeck:
We can revisit this when we refactor the cli

Issue #5922, opened 2018-12-05
common::ui should use `termcolor` crate


mwrock:
Unless things change, we should migrate our UI code to use the termcolor crate. It looks like the term crate is largely unmaintained right now. See https://github.com/Stebalien/term/issues/93:

```
So, I thought I'd have time to maintain this but then I got a job (where a large portion of said job is maintaining a bunch of OSS projects...). Unfortunately, that leaves me little time (or motivation) to maintain this project.
```

Issue #2122, opened 2017-04-18
windows console sometimes display garbled supervisor output


mwrock:
For the most part, the console output on windows from the habitat supervisor and other commands is quite acceptable but there are places where it may appear mangled. This may differ slightly depending on the terminal being used: conemu, stock cmd.exe, default powershell console, etc.

Here are some examples:

![image](https://cloud.githubusercontent.com/assets/655165/25153784/b70c4fa2-2442-11e7-9989-b24eae20b316.png)

![image](https://cloud.githubusercontent.com/assets/655165/25153861/03ad6620-2443-11e7-9db6-3e24267882c1.png)

![image](https://cloud.githubusercontent.com/assets/655165/25153908/33741372-2443-11e7-9c2d-4cb4ef4f2e3c.png)

mwrock:
Adding some observations:

OS 
< 2016/win 10: No color or icons and ANSI color codes are rendered raw and unsightly
2016/win 10: Color renders correctly but some icons are just boxes
ConEMU: Terminal paradise. Text is rendered as god made it.

mwrock:
I'm pretty sure our output handling is supposed to detect if color is supported and suppress it when it is not. My guess is this detection is not correct for windows and we should ensure that for non-ansi compliant terminals on windows, the color is suppressed.

mwrock:
The following PRs address supervisor and launcher coloring:
https://github.com/habitat-sh/core/pull/126
#6256

mwrock:
also related:
#6312 
#6306 
#6290

PR #6203, opened 2019-02-23
use console api instead of ansi codes on windows

mwrock:
Versions of Windows prior to Windows 10 and server 2016 do not support ANSI color control codes and render them as "gibberish" test. This means that on many "not so legacy" windows environments, Habitat output looks very unsightly.

Server 2012 R2:
image

image

The ubiquitous way to do color in a windows terminal has been to leverage the `SetConsoleMode` API.

The `termcolor` crate leverages that API on Windows providing cross platform terminal color support for all of the versions of Windows that Habitat supports.

This PR refactors our `common::UI` struct to use this crate and discards the `term` and `ansi_term` crates. The `term` crate is no longer supported and so this has the added benefit of moving us off of that crate. This also replaces `Blue` with `Magenta` in our `UIWriter.end` function because `blue` is virtually invisible with the default windows console backgrounds.

Here is the same output above with the changes in this PR:

image

image

### DIFF
--- a/components/common/src/ui.rs
+++ b/components/common/src/ui.rs
@@ -27,11 +27,13 @@ use std::{env,
 use uuid::Uuid;
 
 use crate::api_client::DisplayProgress;
-use ansi_term::Colour;
 use pbr;
-use term::{terminfo::TermInfo,
-           Terminal,
-           TerminfoTerminal};
+use termcolor::{self,
+                Color,
+                ColorChoice,
+                ColorSpec,
+                StandardStream,
+                WriteColor};
 
 use self::tty::StdStream;
 use crate::error::{Error,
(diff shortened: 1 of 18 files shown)


thesentinels:

Thank you for contributing!

mwrock:
fixes #5922 and #2122

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -185,6 +169,17 @@ impl io::Write for CtlRequest {
         // be nice to find a cleaner way to model this, but the fact
         // that CtlRequest is sending output to two destinations with
         // different formatting requirements complicates things a bit.
+        //
+        // TODO (MW): For `han sup run` scenarios, it would be nice to
```
stevendanna: s/han/hab/

baumanj:
Looks like you need to locally update to nightly rusfmt based on [the CI failure](https://buildkite.com/chef/habitat-sh-habitat-master-verify/builds/475#83358b7c-983d-40c5-9b03-654f7f0438df).

components/common/src/ui.rs
```diff
@@ -130,18 +132,18 @@ pub trait UIReader {
     fn prompt_yes_no(&mut self, question: &str, default: Option<bool>) -> Result<bool>;
 }
 
+pub trait ColorPrinter {
```
baumanj: Why define our own trait when we could use [`termcolor::WriteColor`](https://docs.rs/termcolor/1.0.4/termcolor/trait.WriteColor.html)?

Since our streams will implement `Write` as well, we can use the standard [`write`](http://doc.rust-lang.org/1.32.0/std/macro.write.html) and [`writeln`](http://doc.rust-lang.org/1.32.0/std/macro.writeln.html) macros.

components/common/src/ui.rs
```diff
@@ -79,42 +81,42 @@ pub enum Status {
...
-            Status::Signed => ('✓', "Signed".into(), Colour::Cyan),
-            Status::Signing => ('☛', "Signing".into(), Colour::Cyan),
-            Status::Skipping => ('…', "Skipping".into(), Colour::Green),
-            Status::Uploaded => ('✓', "Uploaded".into(), Colour::Green),
-            Status::Uploading => ('↑', "Uploading".into(), Colour::Green),
-            Status::Using => ('→', "Using".into(), Colour::Green),
-            Status::Verified => ('✓', "Verified".into(), Colour::Green),
-            Status::Verifying => ('☛', "Verifying".into(), Colour::Green),
-            Status::Custom(c, ref s) => (c, s.to_string(), Colour::Green),
+            Status::Applying => ('↑', "Applying".into(), Color::Green),
```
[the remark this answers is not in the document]
mwrock: I'd prefer to leave that for another PR.

components/common/src/ui.rs
```diff
@@ -130,18 +132,18 @@ pub trait UIReader {
     fn prompt_yes_no(&mut self, question: &str, default: Option<bool>) -> Result<bool>;
 }
 
+pub trait ColorPrinter {
+    fn print(&mut self, buf: &[u8], color: Option<Color>, bold: bool) -> io::Result<()>;
```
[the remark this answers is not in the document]
mwrock: Do you mean `termcolor::NoColor` vs an Option<Color>?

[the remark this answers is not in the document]
mwrock: Yeah a weight enum seems like it would be more clear. I had thought about just passing a `ColorSpec` but it felt like that made alot of the calling code unnecesarily verbose.

components/common/src/ui.rs
```diff
@@ -130,18 +132,18 @@ pub trait UIReader {
...
+}
```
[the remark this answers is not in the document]
mwrock: yeah that makes total sense.

components/common/src/ui.rs
mwrock: Having a one liner color printing call seemed much simpler. I ended up adding this trait for the sake of [ctrl-gateway/src/mod.rs](https://github.com/habitat-sh/habitat/pull/6203/files#diff-25e976e0941144b82133d34113753ac4). Not sure how you would handle its needs and implement `WriteColor`.

components/common/src/ui.rs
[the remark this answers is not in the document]
baumanj: I'm ok with that, but can we file an issue and take it on as a direct follow-up? I'd like to take advantage of this code being in our headspace now.

components/common/src/ui.rs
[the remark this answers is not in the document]
baumanj: I think making the core interface totally general (i.e., `ColorSpec`) and building convenience functions into the trait on top of that to make our common usage ergonomic would be good. Most importantly, I don't want the callers to pass parameters values like `false` and `None`, when we can include more evocative text like `bold` and `yellow` (or, down the road a semantic identifier like `info` or `warning`)

components/common/src/ui.rs
baumanj: I'll defer to your judgement for what gives better ergonomics in the end. The nice thing about building upon `termcolor::WriteColor` is that we could keep a single code path for all our printing and just enable/disable colors with [`ColorChoice`](https://docs.rs/termcolor/1.0.4/termcolor/enum.ColorChoice.html).

If we find code like this
```rust
buffer.set_color(ColorSpec::new().set_fg(Some(Color::Green)))?;
writeln!(&mut buffer, "green text!")?;
bufwtr.print(&buffer)?;
```
to be too verbose, there's no reason we can't build helpers on top of it like:
```rust
our_helper.print("green text!", Color::Green, Weight::Normal)?;
```
or
```rust
our_helper.set_color(Green).print("green text!")?;
```
while still retaining the logic-unifying goodness of `termcolor::WriteColor` in the underlying implementation.

components/common/src/ui.rs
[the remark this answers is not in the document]
mwrock: yeah we could add functions like `fn create_no_color_spec() -> ColorSpec` and `fn create_color_spec(color: Color, weight: Weight) -> ColorSpec`

components/common/src/ui.rs
mwrock: Yeah my first draft used `WriteColor` but that breaks down in the `ctrl-gateway` when sending colorized messages through the `ConsoleLine` protocol where the actual coloring and underlying console API calls need to happen on the gateway client.

baumanj: Can you show me a little more specifically where your approach ran into trouble?

mwrock: I was hoping that the `termcolor::Buffer` would be serializable but its not :( Maybe you know of a way to send a `Buffer` through prost but I think it has to implement `Serializable`

mwrock: I'm thinking of this more and I think I see how to do this with WriteColor. I'll play with it.

baumanj: Sounds great. Let me know if you want input on anything.

components/common/src/ui.rs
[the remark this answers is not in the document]
christophermaier: I would _love_ to see this as a feature.

PR #6208, opened 2019-02-24
Fix concurrent install of local harts

stevendanna:
Previously, concurrent `hab pkg install`s of local packages could
result in one or both of the installs failing. This was because both
hab processes would directly copy the hart into the cache. If one
process unpacked the hart when the other process was still writing it,
it would fail.

We fix it here by using AtomicWriter to ensure that we never put an
incomplete file into the cache.

This also prevents other kinds of cache corruption, such as if you
were to kill a hab process in the middle of copying. An alternative
approach might be some sort of lock if we would prefer to avoid the
extra IO completely.

It might be a bit tricky to write an automated test for this cases, but
I've tested this locally with the following test script:

```
HAB=$1

$HAB pkg install ./core-erlang-20.0-20190115223057-x86_64-linux.hart &
sleep 1
$HAB pkg install ./core-erlang-20.0-20190115223057-x86_64-linux.hart

rm -rf /hab/pkgs/core/erlang/20.0/20190115223057
rm -rf /hab/cache/artifacts/core-erlang-20.0-20190115223057-x86_64-linux.hart
```

Using hab 0.74.0 one process fails, one succeeds:

```
> sudo ./test.sh hab
[1] 29243
» Installing ./core-erlang-20.0-20190115223057-x86_64-linux.hart
» Installing ./core-erlang-20.0-20190115223057-x86_64-linux.hart
✗✗✗
✗✗✗ Couldn't read MetaFile: IDENT, not found
✗✗✗
[ ... SNIP ... ]
★ Install of core/erlang/20.0/20190115223057 complete with 1 new packages installed.
```

Using hab from this branch

```
sudo LD_LIBRARY_PATH=/hab/pkgs/core/zeromq/4.2.5/20190117202641/lib:/hab/pkgs/core/openssl/1.0.2q/20190115014220/lib/ ./test.sh target/debug/hab
» Installing ./core-erlang-20.0-20190115223057-x86_64-linux.hart
» Installing ./core-erlang-20.0-20190115223057-x86_64-linux.hart
[... SNIP ...]
✓ Installed core/erlang/20.0/20190115223057
★ Install of core/erlang/20.0/20190115223057 complete with 1 new packages installed.
✓ Installed core/erlang/20.0/20190115223057
★ Install of core/erlang/20.0/20190115223057 complete with 1 new packages installed.
```

Fixes #5616

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -39,7 +39,8 @@ use std::{borrow::Cow,
           fmt,
           fs::{self,
                File},
-          io::{BufRead,
+          io::{self,
+               BufRead,
                BufReader},
           path::{Path,
                  PathBuf},
@@ -55,7 +56,8 @@ use crate::{api_client::{self,
                              SigKeyPair},
                     fs::{cache_key_path,
                          pkg_install_path,
-                         svc_hooks_path},
+                         svc_hooks_path,
+                         AtomicWriter},
                     package::{list::temp_package_directory,
                               Identifiable,
                               PackageArchive,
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #6209, opened 2019-02-25
Add option to run the supervisor in "local mode"

mpeck:
Adds an option to run the supervisor in local mode.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -902,7 +902,7 @@ pub fn sub_sup_run() -> App<'static, 'static> {
     (usage: "hab sup run [FLAGS] [OPTIONS] [--] [PKG_IDENT_OR_ARTIFACT]")
     (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
         "The listen address for the Gossip System Gateway.")
-    (@arg LOCAL_MODE: --("local-mode") "Start the supervisor in local mode. This sets the gossip listen address to 127.0.0.2 effectively scoping it to only the local node. This is equivalent to using 127.0.0.2 as the ip for --listen-gossip. Note: Setting local mode will override values for --listen-gossip, including values set via the env variable.")
+    (@arg LOCAL_MODE: --("local-mode") conflicts_with("LISTEN_GOSSIP") conflicts_with("PEER") "Start the supervisor in local mode.")
     (@arg LISTEN_HTTP: --("listen-http") env(LISTEN_HTTP_ADDRESS_ENVVAR) default_value(&LISTEN_HTTP_DEFAULT_ADDR) {valid_socket_addr}
         "The listen address for the HTTP Gateway.")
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -893,8 +888,9 @@ pub fn sub_sup_run() -> App<'static, 'static> {
     // is displayed confusingly as `hab-sup`
     // see: https://github.com/kbknapp/clap-rs/blob/2724ec5399c500b12a1a24d356f4090f4816f5e2/src/app/mod.rs#L373-L394
     (usage: "hab sup run [FLAGS] [OPTIONS] [--] [PKG_IDENT_OR_ARTIFACT]")
-          (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
+    (@arg LISTEN_GOSSIP: --("listen-gossip") env(GOSSIP_LISTEN_ADDRESS_ENVVAR) default_value(&GOSSIP_DEFAULT_ADDR) {valid_socket_addr}
         "The listen address for the Gossip System Gateway.")
+    (@arg LOCAL_MODE: --("local-mode") "Start the supervisor in local mode. This sets the gossip listen address to 127.0.0.2 effectively scoping it to only the local node. This is equivalent to using 127.0.0.2 as the ip for --listen-gossip. Note: Setting local mode will override values for --listen-gossip, including values set via the env variable.")
```
mpeck: I decided to not use an arg group here so we could continue to leverage `default_value` and `env` for `LISTEN_GOSSIP`; I thought that was more important.

components/hab/src/cli.rs
```diff
@@ -893,8 +893,9 @@ pub fn sub_sup_run() -> App<'static, 'static> {
...
```
baumanj: This should probably be named something like `local-gossip-mode` since there are other things the supervisor listens for and this only applies to the gossip subsystem.

components/hab/src/cli.rs
baumanj: Also, don't forget a comment with a link to the clap issue about why we can't have both `default` and `conflicts_with`.

baumanj: This should probably also conflict with `--peer-watch-file` and `--permanent-peer`

mpeck: I debated not including the implementation details; then I thought it might be useful information for a user to know that gossip is still running, but scoped to the local node. If you think it's more confusing than it's worth though I can drop it.

baumanj: I tend to err on the side of not bothering the user with any details they don't *need* to know because it helps them see the things that **are** important and frees our hand for changing those details later.

I can't think of a reason why anyone would need to know that the gossip server is actually running and only listening on 127.0.0.2, but if one occurs to you, let's discuss.

mpeck: Woot! It looks like `conflicts_with` doesn't have the same issue as using an ArgGroup does. We can specify `conflicts_with` on `local-gossip-mode` and still have `env` and `default`.

PR #6210, opened 2019-02-25
Fix clippy::redundant_field_names needless_update and add to DENIED_LINTS

baumanj:
All the instances of [`redundant_field_names`](https://rust-lang.github.io/rust-clippy/master/#redundant_field_names) had already been fixed (thanks @raskchanky), but this moves that lint to the category that makes new instances an error. Additionally, this fixes the one instance of [`needless_update`](https://rust-lang.github.io/rust-clippy/master/#needless_update) and adds it to the error-causing category as well.

PR #6202, opened 2019-02-22
Fix clippy::question_mark and add to DENIED_LINTS

baumanj:
Ironically we didn't have any code that failed [`question_mark`](https://rust-lang.github.io/rust-clippy/master/#question_mark) at this point, but this moves it to DENIED_LINTS and fixes some other lints.

PR #6198, opened 2019-02-22
Fix clippy::len_without_is_empty lint and add to DENIED_LINTS

baumanj:
See https://rust-lang.github.io/rust-clippy/master/#len_without_is_empty

Also updated callers to use is_empty when appropriate

[1 down, 9 to go](https://github.com/habitat-sh/habitat/blob/master/Makefile#L202)…

PR #6214, opened 2019-02-25
Update the list of clippy lints in run_clippy.sh

baumanj:
Previously [these were tracked in the Makefile](https://github.com/habitat-sh/habitat/blob/d0b63fdc6cc6942bbe308e05c368c1045315f16d/Makefile#L192-L263), but since the script is now the source of truth, we need to make an update to account for the recent changes in the Makefile.

See:
https://github.com/habitat-sh/habitat/issues/6213
https://github.com/habitat-sh/habitat/pull/6198
https://github.com/habitat-sh/habitat/pull/6202
https://github.com/habitat-sh/habitat/pull/6210


thesentinels:

Thank you for contributing!

PR #6219, opened 2019-02-26
Fix a buglet in stable launcher promotion during a release

christophermaier:
Turns out there's a very important difference between `$"{from_channel}"` and`"${from_channel}"`.

tenor-181284286

Normally, Shellcheck would catch this as [SC2034](https://github.com/koalaman/shellcheck/wiki/SC2034), but we don't currently check for this due to excessive false positive rates.


thesentinels:

Thank you for contributing!

PR #6228, opened 2019-02-27
fix array handling on windows scaffolding

mwrock:
This fixes a potential issue when scaffolding is declared as a string (which it likely would be if declared at all) and pulling the 0th element would resolve to the first character in the string.

Note that if `$pkg_scaffolding` is not declared then it will be `$null` which will habe a `count` of `0`. So we first check if it is a multi element array. 

Then `$pkg_scaffolding = @($pkg_scaffolding)[0]` gets the 0th element of an array or the entire string if it is just a string.


thesentinels:

Thank you for contributing!

PR #6256, opened 2019-03-02
refactor ctl_gateway and launcher to leverage windows color friendly StructuredOutput

mwrock:
This leverages https://github.com/habitat-sh/core/pull/126 and should not be merged until that is merged.

This focuses on 2 areas of output:

1. The launcher emitting the output of spawned processes
2. Output T'd from the ctrl_gateway

The ctrl_gateway implementation is a bit whacky because instead or receiving whole "chunks" of output (think of the output emitted when loading a new service that must be downloaded and installed), it now receives output piecemeal, separated when styling changes and so we can't blissfully send the entire chunk as a line of StructuredOutput. We only want the structured output header if this is either the beginning of a line or if we are emitting json.

### DIFF
--- a/components/common/src/lib.rs
+++ b/components/common/src/lib.rs
@@ -19,14 +19,11 @@ extern crate json;
 extern crate log;
 #[macro_use]
 extern crate lazy_static;
-
 extern crate serde;
 #[macro_use]
 extern crate serde_derive;
-
 #[cfg_attr(test, macro_use)]
 extern crate serde_json;
-
 #[cfg(windows)]
 extern crate winapi;
 
@@ -37,8 +34,21 @@ pub mod cli_defaults;
 pub mod command;
 pub mod error;
 pub mod locked_env_var;
+pub mod output;
 pub mod package_graph;
 pub mod templating;
 pub mod types;
 pub mod ui;
 pub mod util;
+
+lazy_static::lazy_static! {
+    pub static ref PROGRAM_NAME: String = {
+        match std::env::current_exe() {
+            Ok(path) => path.file_stem().and_then(|p| p.to_str()).unwrap().to_string(),
+            Err(e) => {
+                error!("Error getting path of current_exe: {}", e);
+                String::from("hab-?")
+            }
+        }
+    };
+}
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -16,36 +16,10 @@
 
 #[macro_use]
 extern crate clap;
-use env_logger;
-
-use habitat_common as common;
-use habitat_core as hcore;
-use habitat_sup_client as sup_client;
-use habitat_sup_protocol as protocol;
-
 #[macro_use]
 extern crate lazy_static;
 #[macro_use]
 extern crate log;
-use pbr;
-
-use std::{env,
-          ffi::OsString,
-          fs::File,
-          io::{self,
-               prelude::*,
-               Read},
-          net::ToSocketAddrs,
-          path::{Path,
-                 PathBuf},
-          process,
-          result,
-          str::FromStr,
-          thread};
-use termcolor::{self,
-                Color,
-                ColorChoice,
-                ColorSpec};
 
 #[cfg(windows)]
 use crate::hcore::crypto::dpapi::encrypt;
@@ -53,6 +27,8 @@ use crate::{common::{command::package::install::{InstallHookMode,
                                                  InstallMode,
                                                  InstallSource,
                                                  LocalPackageUsage},
+                     output::{self,
+                              OutputFormat},
                      types::ListenCtlAddr,
                      ui::{Status,
                           UIWriter,
@@ -72,23 +48,21 @@ use crate::{common::{command::package::install::{InstallHookMode,
                          launcher_root_path},
                     package::{PackageIdent,
                               PackageTarget},
-                    ChannelIdent}};
-use clap::{ArgMatches,
-           Shell};
-use futures::prelude::*;
-
-use crate::{hcore::{service::{HealthCheckInterval,
+                    service::{HealthCheckInterval,
                               ServiceGroup},
                     url::{bldr_url_from_env,
-                          default_bldr_url}},
+                          default_bldr_url},
+                    ChannelIdent},
             protocol::{codec::*,
                        ctl::ServiceBindList,
                        net::ErrCode,
                        types::*},
             sup_client::{SrvClient,
                          SrvClientError}};
-use tabwriter::TabWriter;
-
+use clap::{ArgMatches,
+           Shell};
+use env_logger;
+use futures::prelude::*;
 use hab::{analytics,
           cli,
           command::{self,
@@ -105,6 +79,29 @@ use hab::{analytics,
           ORIGIN_ENVVAR,
           PRODUCT,
           VERSION};
+use habitat_common as common;
+use habitat_core as hcore;
+use habitat_sup_client as sup_client;
+use habitat_sup_protocol as protocol;
+use pbr;
+use std::{env,
+          ffi::OsString,
+          fs::File,
+          io::{self,
+               prelude::*,
+               Read},
+          net::ToSocketAddrs,
+          path::{Path,
+                 PathBuf},
+          process,
+          result,
+          str::FromStr,
+          thread};
+use tabwriter::TabWriter;
+use termcolor::{self,
+                Color,
+                ColorChoice,
+                ColorSpec};
 
 /// Makes the --org CLI param optional when this env var is set
 const HABITAT_ORG_ENVVAR: &str = "HAB_ORG";
@@ -1775,10 +1772,9 @@ fn user_param_or_env(m: &ArgMatches<'_>) -> Option<String> {
 // function wouldn't be necessary. In the meantime, though, it'll keep
 // the scope of change contained.
 fn ui() -> UI {
-    let coloring = if hcore::output::is_color() {
-        ColorChoice::Auto
-    } else {
-        ColorChoice::Never
+    let coloring = match output::get_format() {
+        OutputFormat::Color => ColorChoice::Auto,
+        OutputFormat::NoColor | OutputFormat::JSON => ColorChoice::Never,
     };
     let isatty = if env::var(NONINTERACTIVE_ENVVAR).map(|val| val == "1" || val == "true")
                                                    .unwrap_or(false)
--- a/components/launcher/src/lib.rs
+++ b/components/launcher/src/lib.rs
@@ -12,12 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[macro_use]
 extern crate habitat_core as core;
 use habitat_launcher_protocol as protocol;
 #[macro_use]
 extern crate log;
-
 #[cfg(windows)]
 extern crate winapi;
 
--- a/components/launcher/src/service.rs
+++ b/components/launcher/src/service.rs
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::protocol;
+#[cfg(windows)]
+use core::os::process::windows_child::{ChildStderr,
+                                       ChildStdout,
+                                       ExitStatus};
+use habitat_common::output::{self,
+                             StructuredOutput};
 #[cfg(unix)]
 use std::process::{ChildStderr,
                    ChildStdout,
@@ -20,15 +27,8 @@ use std::{fmt,
           io::{self,
                BufRead,
                BufReader,
-               Read,
-               Write},
+               Read},
           thread};
-
-use crate::protocol;
-#[cfg(windows)]
-use core::os::process::windows_child::{ChildStderr,
-                                       ChildStdout,
-                                       ExitStatus};
 
 pub use crate::sys::service::*;
 
@@ -89,8 +89,12 @@ fn pipe_stdout<T>(out: T, id: &str)
     let mut reader = BufReader::new(out);
     let mut buffer = String::new();
     while reader.read_line(&mut buffer).unwrap() > 0 {
-        let line = output_format!(preamble &id, logkey "O", buffer);
-        writeln!(&mut io::stdout(), "{}", line).expect("unable to write to stdout");
+        let content = &buffer.trim_right_matches('\n');
+        let so = StructuredOutput::succinct(&id, "O", output::get_format(), content);
+        if let Err(e) = so.println() {
+            println!("printing output: '{}' to stdout resulted in error: {}",
+                     content, e);
+        }
         buffer.clear();
     }
 }
@@ -102,8 +106,12 @@ fn pipe_stderr<T>(err: T, id: &str)
     let mut reader = BufReader::new(err);
     let mut buffer = String::new();
     while reader.read_line(&mut buffer).unwrap() > 0 {
-        let line = output_format!(preamble &id, logkey "E", buffer);
-        writeln!(&mut io::stderr(), "{}", line).expect("unable to write to stderr");
+        let content = &buffer.trim_right_matches('\n');
+        let so = StructuredOutput::succinct(&id, "E", output::get_format(), content);
+        if let Err(e) = so.eprintln() {
+            eprintln!("printing output: '{}' to stderr resulted in error: {}",
+                      content, e);
+        }
         buffer.clear();
     }
 }
--- a/components/sup/src/ctl_gateway/mod.rs
+++ b/components/sup/src/ctl_gateway/mod.rs
@@ -22,36 +22,28 @@
 
 pub mod server;
 
-use std::{borrow::Cow,
-          fmt,
+use crate::error::{Error,
+                   Result};
+use futures::prelude::*;
+use habitat_api_client::DisplayProgress;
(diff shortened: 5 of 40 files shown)


thesentinels:

Thank you for contributing!

mwrock:
Before Change:
image

After Change:
image

mwrock:
related to https://github.com/habitat-sh/habitat/issues/2122

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -190,37 +188,54 @@ impl Write for CtlRequest {
...
+            } else {
+                StructuredOutput::new(
+                    PROGRAM_NAME.as_str(),
+                    LOGKEY,
+                    line!(),
+                    file!(),
+                    column!(),
+                    &line.trim_right_matches('\n'),
+                )
+            };
```
baumanj: We can eliminate the duplication here (also `.as_str()` can be replaced with just a borrow):
```rust
            let so = StructuredOutput::colored(
                &PROGRAM_NAME,
                LOGKEY,
                line!(),
                file!(),
                column!(),
                line.trim_right_matches('\n'),
                self.current_color_spec.clone().unwrap_or_default(),
            );
```

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -79,6 +73,7 @@ pub struct CtlRequest {
     /// Transaction for the given request.
     transaction: Option<protocol::codec::SrvTxn>,
     current_color_spec: Option<ColorSpec>,
+    is_new_line: bool,
```
baumanj: As long as we're changing this struct, I think we could eliminate the `Option` around `current_color_spec` and replace the `None` cases with `ColorSpec::default`. That should simplify a bunch of the logic, I think.

components/launcher/src/service.rs
```diff
@@ -95,8 +95,15 @@ where
     let mut reader = BufReader::new(out);
     let mut buffer = String::new();
     while reader.read_line(&mut buffer).unwrap() > 0 {
-        let line = output_format!(preamble &id, logkey "O", buffer);
-        writeln!(&mut io::stdout(), "{}", line).expect("unable to write to stdout");
+        let so = StructuredOutput::new(
```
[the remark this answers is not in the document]
mwrock: The panic should surface to the console or log and would likely be the last entry since it would crash the supervisor.

mwrock:
I think I have covered all of your great comments @baumanj . This PR is much smaller than it looks since there are many `use` changes due to the relocation of `outputln` and `PROGRAM_NAME`.

components/launcher/src/lib.rs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #[macro_use]
+extern crate habitat_common as common;
```
christophermaier: In general, we'd like to move away from this pattern since these `extern crate` statements are generally no longer needed with Rust 2018 Edition.

Instead, each module just directly does a `use habitat_common`.

This also means that any `use crate::common::{...};`  statements now become `use habitat_common::{...};`. This is nice because it will mean that anything in `use crate::{...};` will actually be things in the current crate; everything else is going to then be treated like the external crate it is.

components/launcher/src/service.rs
```diff
@@ -89,8 +90,15 @@ fn pipe_stdout<T>(out: T, id: &str)
...
-        let line = output_format!(preamble &id, logkey "O", buffer);
-        writeln!(&mut io::stdout(), "{}", line).expect("unable to write to stdout");
+        let so = StructuredOutput::new(&id,
+                                       "O",
+                                       line!(),
+                                       file!(),
+                                       column!(),
+                                       output::get_format(),
+                                       output::get_verbosity(),
+                                       &buffer.trim_right_matches('\n'));
```
christophermaier: I thought we had macros that wrapped all the verbosity of creating these `StructuredOutput` structs?

components/launcher/src/service.rs
```diff
@@ -89,8 +90,15 @@ fn pipe_stdout<T>(out: T, id: &str)
     let mut reader = BufReader::new(out);
     let mut buffer = String::new();
     while reader.read_line(&mut buffer).unwrap() > 0 {
-        let line = output_format!(preamble &id, logkey "O", buffer);
-        writeln!(&mut io::stdout(), "{}", line).expect("unable to write to stdout");
+        let so = StructuredOutput::new(&id,
+                                       "O",
+                                       line!(),
+                                       file!(),
+                                       column!(),
```
christophermaier: Are all these line/file/column numbers going to pertain to _this_ code (i.e., always going to be the same)?

Is that expected / OK?

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -22,37 +22,32 @@
...
           fs::{self,
                File},
           io::{self,
                Write},
           path::Path};
 
-use regex::Regex;
-use termcolor::{Buffer,
-                Color,
+use termcolor::{Color,
```
christophermaier: It would be good to remove the newline that separates `std` from everything else for consistency, and to let `rustfmt` handle everything for us.

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -22,37 +22,32 @@
...
 use futures::prelude::*;
 
 use crate::error::{Error,
                    Result};
 
-lazy_static! {
-    /// Shamelessly stolen from https://github.com/chalk/ansi-regex/blob/master/index.js
-    static ref STRIP_ANSI_CODES: Regex = Regex::new(
-        r"[\x1b\x9b][\[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-PRZcf-nqry=><]")
-        .unwrap();
```
christophermaier: We don't need this anymore?

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -163,56 +160,65 @@ impl Write for CtlRequest {
...
-            ansi_buffer.set_color(&spec)?;
-        }
-        ansi_buffer.write_all(buf)?;
-        ansi_buffer.flush()?;
-        let line = String::from_utf8_lossy(ansi_buffer.as_slice()).to_string();
 
-        let maybe_stripped = if output::is_json() || !output::is_color() {
-            STRIP_ANSI_CODES.replace_all(&line, "")
+        let is_line_ending = line.ends_with('\n');
+        if self.is_new_line || output::get_format() == OutputFormat::JSON {
```
christophermaier: This is a somewhat surprising / confusing if condition... why is "newline + JSON" the criterion?

components/sup/src/ctl_gateway/mod.rs
```diff
@@ -163,56 +160,65 @@ impl Write for CtlRequest {
...
+        msg.color = color_to_string(self.current_color_spec.fg());
+        msg.bold = self.current_color_spec.bold();
         self.reply_partial(msg);
 
         // Down here, however, we're doing double-duty by *also*
-        // sending the output to the Supervisor's output stream. In
-        // _this_ case, we want to honor whatever global logging flags
-        // have been set; in particular, if we're emitting
-        // JSON-formatted logs or if we've been instructed to not
-        // output ANSI color codes, we need to strip those codes out.
```
christophermaier: Does this still work?

components/launcher/src/service.rs
mwrock: these 2 calls in the launcher used a macro `output_format!` which wrapped the `StructuredOutput`. They were the only callers. That particular macro was problematic in that it just formats the output and returns a string that the launcher renders here to either `stdout` or `stderr`. That works great in an `ANSI` terminal world but it breaks down when you need to make system API calls when ever you need to change color or formatting.

We sould probably have the `stdout` call leverage `outputln!` but I thought it seemed strange to have one call a macro and the other deal directly with `StructuredOutput` and it seemed overboard to create a set of corresponding `outputerrln!` macros for only this one caller.

Hope that makes sense but totally open to suggestions.

components/sup/src/ctl_gateway/mod.rs
christophermaier: Actually, it looks like this breaks. You can see this by running `hab sup run --json-logging` with current stable Habitat vs. this PR and then doing something like:

```hab svc load <some package that isn't currently installed>```

In the supervisor terminal you should see only JSON, and no color, while in the `hab svc load` terminal, you should see "normal" colored output.

With this PR, you still get colored output in the server output.

File: components/sup/src/ctl_gateway/mod.rs (hunk 8 above)
Comment (mwrock) [reply components/sup/src/ctl_gateway/mod.rs]: nope. because now we don't have a scenario where we have injected ansi strings and later need to remove them. Using `termcolor` they are injected at the time the output is streamed based on the color preferences of the environment.

File: components/sup/src/ctl_gateway/mod.rs (hunk 10 above)
Comment (mwrock) [reply components/sup/src/ctl_gateway/mod.rs]: that is definitely not good. I'll have a look

File: components/sup/src/ctl_gateway/mod.rs (hunk 9 above)
Comment (mwrock) [reply components/sup/src/ctl_gateway/mod.rs]: yeah this could use some comments that I will add. Previous to 0.77, `write` on the `CtlRequest` would be called for each line of output. Now it can be called for line fragments when styling changes mid-line. Unless we are rendering JSON, we only want a `StructuredOutput` line header when we are beginning a new line, otherwise we just want the message content.

File: components/common/src/lib.rs [hunk 11]
```diff
@@ -37,8 +37,22 @@ pub mod cli_defaults;
...
 pub mod templating;
 pub mod types;
 pub mod ui;
 pub mod util;
+
+use std::path::PathBuf;
+
+lazy_static::lazy_static! {
+    pub static ref PROGRAM_NAME: String = {
+        let arg0 = std::env::args().next().map(PathBuf::from);
```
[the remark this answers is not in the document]
mwrock: I totally just copy/pasted this block from core but `Iterator` doesn't have a `first` member. I could do `nth(0)` but not so sure that is an improvement over `next` but I'm not very opinionated here.

components/launcher/src/service.rs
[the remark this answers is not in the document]
baumanj: It would log, but I'm pretty sure it *wouldn't* crash the supervisor. Just because a thread panics, doesn't mean the application crashes. That would only happen if the main thread were to try to join the panicked thread. Here's a toy example:
```rust
    let handle = std::thread::Builder::new()
        .spawn(move || panic!("I ded"))
        .unwrap();

    for _ in 1..5 {
        println!("I'm alive");
        std::thread::sleep(std::time::Duration::from_secs(5));
    }

    println!("calling join");
    handle.join().expect("join other thread");
    println!("join success");
```
When I run this locally, I only get a crash once I get to the `join` call:
```
➤ cargo run --release
I'm alive
thread '<unnamed>' panicked at 'I ded', src/main.rs:3:24
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
    …
I'm alive
I'm alive
I'm alive
calling join
thread 'main' panicked at 'join other thread: Any', src/libcore/result.rs:997:5
stack backtrace:
    …
```
I think an `error!` log would be more appropriate here. Panicking should be reserved for things that we think represent logical impossibilities.

components/common/src/lib.rs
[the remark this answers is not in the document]
baumanj: How about we use [`std::env::current_exe`](http://doc.rust-lang.org/1.33.0/std/env/fn.current_exe.html) instead?

components/launcher/src/service.rs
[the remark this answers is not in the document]
mwrock: oh man I forgot that these are run in separate threads

components/launcher/src/lib.rs
mwrock: I thought we still needed to do this if the crate had macros we are using. no?

mwrock: After a bit of googling its not clear how we would get away from the `extern crate` when needing to use macros and it seems we have maintained this pattern for the core crate for this reason if I am interpreting things correctly.

mwrock: never mind. I got it.

components/launcher/src/service.rs
mwrock: They are and this PR does not change this but maybe we should change that here because it does seem rather useless and noisy. This just emits output from the process. What would you think of forcing this launcher output to always be non verbose which would remove these details?

components/sup/src/ctl_gateway/mod.rs
mwrock: Simple logic bug that I fixed.

mwrock:
ok @christophermaier and @baumanj I think I got everything here. There is one question above I have about forcing `Normal` verbosity for launcher output.

I may have gone a little OCD on cleaning up our habitat `extern crate` usages in the supervisor. They are all gone now.

baumanj:
> I may have gone a little OCD on cleaning up our habitat `extern crate` usages in the supervisor. They are all gone now.

That's not OCD (which is a pretty awful disorder that someone close to me has), that's just healthy pedantry 😄. Thanks for doing that.

components/common/src/lib.rs
```diff
@@ -37,8 +37,15 @@ pub mod cli_defaults;
...
+pub mod output;
 pub mod package_graph;
 pub mod templating;
 pub mod types;
 pub mod ui;
 pub mod util;
+
+lazy_static::lazy_static! {
+    pub static ref PROGRAM_NAME: String = {
+        std::env::current_exe().unwrap().file_stem().and_then(|p| p.to_str()).unwrap().to_string()
```
baumanj: [`current_exe` is not infallible](http://doc.rust-lang.org/1.33.0/std/env/fn.current_exe.html#errors). Are we sure we want to abort the program if it fails? Maybe it's better to log an `error!` and return some default like `"hab-?"`

components/hab/src/main.rs
```diff
@@ -1775,10 +1777,9 @@ fn user_param_or_env(m: &ArgMatches<'_>) -> Option<String> {
 // function wouldn't be necessary. In the meantime, though, it'll keep
 // the scope of change contained.
 fn ui() -> UI {
-    let coloring = if hcore::output::is_color() {
-        ColorChoice::Auto
-    } else {
-        ColorChoice::Never
+    let coloring = match output::get_format() {
+        OutputFormat::Color => ColorChoice::Auto,
+        _ => ColorChoice::Never,
```
baumanj: Are we 100% certain that we'll never add any additional output formats? If not, and there's any chance that new format might want to use color, we should be exhaustive instead of using the catch-all:
```suggestion
        OutputFormat::NoColor | OutputFormat::JSON => ColorChoice::Never,
```
That way, the compiler will make sure we update this.

components/launcher/src/service.rs
baumanj: This is what happened before when we used the [`output_format!` macro](https://github.com/habitat-sh/core/blob/bba280adf5c46e57569caa1e15034747970ec245/components/core/src/output.rs#L264), right?

That said, I think we *should* fix this, and I think the most appropriate way is by making the `line`, `file` and `column` data associated specifically with the `Verbose` variant of the `Verbosity` enum.

Since we don't want to expose that to callers, we'll need a separate internal version
```rust
enum OutputVerbosityInternal {
    Normal,
    Verbose {
        line:   u32,
        file:   &'static str,
        column: u32,
    },
}
```
Then we simplify the struct:
```diff
 pub struct StructuredOutput<'a> {
     preamble: &'a str,
     logkey: &'static str,
-    line: u32,
-    file: &'static str,
-    column: u32,
     content: &'a str,
     /// The verbosity level of rendered content
-    verbosity: OutputVerbosity,
+    verbosity: OutputVerbosityInternal,
```
…add a bit of logic to `new`/`colored`:
```rust
        let verbosity = match verbosity {
            OutputVerbosity::Normal => OutputVerbosityInternal::Normal,
            OutputVerbosity::Verbose => OutputVerbosityInternal::Verbose { line, file, column },
        };
```
…change the verbosity `if let`s to:
```rust
                if let OutputVerbosityInternal::Verbose { line, file, column } = self.verbosity {
```
…and add a `succinct` associated function which creates the struct like:
```rust
        StructuredOutput { preamble,
                           logkey,
                           content,
                           verbosity: OutputVerbosityInternal::Normal,
                           format,
                           color_spec: ColorSpec::new() }
```
and the launcher can call:
```rust
        let so = StructuredOutput::succinct(&id, "O", output::get_format(), content);
```

components/sup/src/ctl_gateway/mod.rs
christophermaier: That's fun! 😄

components/sup/src/ctl_gateway/mod.rs
christophermaier: Confirmed: it works now 👍

components/hab/src/main.rs
christophermaier: I like this suggestion 👍

mwrock: yeah good thinking. This logic is in a couple spots and I'll get it in all of them.

mwrock:
fixes https://github.com/habitat-sh/core/issues/128

PR #6240, opened 2019-03-01
[plan-build] Avoid reading a newline delimited value to array

ryancragun:
Despite my sincerest efforts, some combination of the build environment
is making reading newline delimited output to array not workable. I
tried several combinations of `read` with delimiters and IFS variables
and could not get it to work properly.

This resolves and issue where wrapper scaffolding dependencies were not
resolving because we only read the first dependency.

Signed-off-by: Ryan Cragun <ryan@chef.io>

PR #6160, opened 2019-02-14
Enable clippy and shellcheck linting 

scotthain:

PR #6257, opened 2019-03-04
Fix plan build regression in 0.76.0 release

baumanj:
**tl;dr** see the fix in https://github.com/habitat-sh/habitat/pull/6257/commits/05d362304504770424ed076805dca570d938a669

# Full Details

Reverts habitat-sh/habitat#6248
Resolves https://github.com/habitat-sh/habitat/issues/6250

Now that we understand the nature of this error, we can re-implement that shellcheck fixes that we previously reverted and add the necessary fix to the problem they uncovered.

**The Bugs**

1. In `_resolve_scaffolding_dependencies` (and other similar usages), changing the code that collects the transitive dependencies from
    ```bash
    sdeps=($(_get_deps_for "$resolved"))
    ```
    to
    ```bash
    read -r -a sdeps <<< "$(_get_deps_for "$resolved")"
    ```
    as part of https://github.com/habitat-sh/habitat/pull/6160 to address `shellcheck` lint [SC2207](https://github.com/koalaman/shellcheck/wiki/SC2207#prefer-mapfile-or-read--a-to-split-command-output-or-quote-to-avoid-splitting). This resulted in a defect where because only the first dependency from `_get_deps_for` was added to `sdeps`, scaffolding-based builds failed. https://github.com/habitat-sh/habitat/pull/6240 was submitted to fix the issue.

1. The fix in https://github.com/habitat-sh/habitat/pull/6240 was correct. It changed
    ```bash
    read -r -a sdeps <<< "$(_get_deps_for "$resolved")"
    ```
    to
    ```bash
    mapfile -t sdeps < <(_get_deps_for "$resolved")
    ```
    The first command is for populating an array from a single line of input, with entries separated by whitespace. The second command is for populating an array from multiple lines of input, with entries separate by newlines (which is what `_get_deps_for` does).

    However, a different bug was discovered: unexpected newlines were appearing at the beginning of the `TDEPS` files for some packages, causing breakages. Eventually, the difference in behavior between
    ```bash
    sdeps=($(_get_deps_for "$resolved"))
    ```
    and
    ```bash
    mapfile -t sdeps < <(_get_deps_for "$resolved")
    ```
    was traced back to `_get_deps_for`:
    ```bash
    # **Internal** Returns (on stdout) the `TDEPS` file contents of another locally
    # installed package which contain the set of all direct and transitive run
    # dependencies. An empty set could be returned as whitespace and/or newlines.
    # The lack of a `TDEPS` file in the desired package will be considered an
    # unset, or empty set.
    #
    # ```
    # _get_tdeps_for /hab/pkgs/acme/a/4.2.2/20160113044458
    # # /hab/pkgs/acme/dep-b/1.2.3/20160113033619
    # # /hab/pkgs/acme/dep-c/5.0.1/20160113033507
    # # /hab/pkgs/acme/dep-d/2.0.0/20160113033539
    # # /hab/pkgs/acme/dep-e/10.0.1/20160113033453
    # # /hab/pkgs/acme/dep-f/4.2.2/20160113033338
    # # /hab/pkgs/acme/dep-g/4.2.2/20160113033319
    # ```
    #
    # Will return 0 in any case and the contents of `TDEPS` if the file exists.
    _get_tdeps_for() {
      local pkg_path="$1"
      if [[ -f "$pkg_path/TDEPS" ]]; then
        cat "$pkg_path"/TDEPS
      else
        # No file, meaning an empty set
        echo
      fi
      return 0
    }
    ```
    Because `_get_tdeps_for` returned a newline (rather than nothing) in the case of an absent `TDEPS` file (which is normal, some packages have no transitive dependencies), the `mapfile` version of the code interpreted this as a single element array, containing an empty string. This was eventually rendered into the subsequent `TDEPS` file as an empty line.

    Essentially, the previous code depended upon ["the shell's sloppy word splitting and glob expansion" that `shellcheck` warns against](https://github.com/koalaman/shellcheck/wiki/SC2207#prefer-mapfile-or-read--a-to-split-command-output-or-quote-to-avoid-splitting). Updating `_get_tdeps_for` (and similarly `_get_deps_for`) to output nothing in the case of an empty set (and omitting the superfluous `return`):
    ```bash
    _get_tdeps_for() {
      local pkg_path="$1"
      if [[ -f "$pkg_path/TDEPS" ]]; then
        cat "$pkg_path"/TDEPS
      fi
    }
    ```
    fixes the issue.

This PR makes some additional improvements to `_get_[t]deps_for`; see https://github.com/habitat-sh/habitat/pull/6257/commits/05d362304504770424ed076805dca570d938a669.

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -758,9 +758,9 @@ _install_dependency() {
 
 # **Internal** Returns (on stdout) the `TDEPS` file contents of another locally
 # installed package which contain the set of all direct and transitive run
-# dependencies. An empty set could be returned as whitespace and/or newlines.
-# The lack of a `TDEPS` file in the desired package will be considered an
-# unset, or empty set.
+# dependencies. An empty set generates no output. The lack of a `TDEPS` file or
+# a TDEPS file of zero bytes in the desired package will be considered an unset,
+# or empty set.
 #
 # ```
 # _get_tdeps_for /hab/pkgs/acme/a/4.2.2/20160113044458
@@ -774,20 +774,17 @@ _install_dependency() {
 #
 # Will return 0 in any case and the contents of `TDEPS` if the file exists.
 _get_tdeps_for() {
-  local pkg_path="$1"
-  if [[ -f "$pkg_path/TDEPS" ]]; then
+  local pkg_path="${1?_get_tdeps_for requires a pkg_path argument}"
+  if [[ -s "$pkg_path/TDEPS" ]]; then
     cat "$pkg_path"/TDEPS
-  else
-    # No file, meaning an empty set
-    echo
   fi
-  return 0
 }
 
 # **Internal** Returns (on stdout) the `DEPS` file contents of another locally
 # installed package which contain the set of all direct run dependencies. An
-# empty set could be returned as whitespace and/or newlines.  The lack of a
-# `DEPS` file in the desired package will be considered an unset, or empty set.
+# empty set could be returned as whitespace and/or newlines. An empty set
+# generates no output. The lack of a `DEPS` file or a DEPS file of zero bytes in
+# the desired package will be considered an unset, or empty set.
 #
 # ```
 # _get_deps_for /hab/pkgs/acme/a/4.2.2/20160113044458
@@ -801,14 +798,10 @@ _get_tdeps_for() {
 #
 # Will return 0 in any case and the contents of `DEPS` if the file exists.
 _get_deps_for() {
-  local pkg_path="$1"
-  if [[ -f "$pkg_path/DEPS" ]]; then
+  local pkg_path="${1?_get_deps_for requires a pkg_path argument}"
+  if [[ -s "$pkg_path/DEPS" ]]; then
     cat "$pkg_path"/DEPS
-  else
-    # No file, meaning an empty set
-    echo
   fi
-  return 0
 }
 
 # **Internal** Appends an entry to the given array only if the entry is not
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

eeyun:
Discussion on whether to merge this now or wait until we have some scripted tests is taking place in slack. As a note on the issue, the cases we were testing on Friday that were failing appear to be fine in this PR branch but as @smacfarlane pointed out there's still some concern around moving forward with reverting the reversion without some tests that can tell us that there aren't any other unforseen ripples from the change.

eeyun:
The decision here is to merge this, thaw the merge-freeze and backfill tests while this is in acceptance before the next release! Great work on these fixes @baumanj.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -953,30 +946,27 @@ _resolve_scaffolding_dependencies() {
...
-  done
+  _install_dependency "$pkg_scaffolding"
+  # Add scaffolding package to the list of scaffolding build deps
+  scaff_build_deps+=("$pkg_scaffolding")
+  if resolved="$(_resolve_dependency "$pkg_scaffolding")"; then
+    build_line "Resolved scaffolding dependency '$pkg_scaffolding' to $resolved"
+    scaff_build_deps_resolved+=("$resolved")
+    # Add each (fully qualified) direct run dependency of the scaffolding
+    # package.
+    mapfile -t sdeps < <(_get_deps_for "$resolved")
```
smacfarlane: @baumanj Could you provide a brief comment inline describing what the `< <()` syntax means? I'm having a little trouble understanding it.  I see in your pr you link to the shellcheck page that describes that this is the preferred style, but didn't see an explanation of the syntax.  Is it a particular bash idiom we could link to?

baumanj: Added https://github.com/habitat-sh/habitat/pull/6257/commits/d60e5212754f6c870429eb0786baa700a7ddce8e

PR #6272, opened 2019-03-06
Fix clippy::large_enum_variant and add to DENIED_LINTS

baumanj:
See https://rust-lang.github.io/rust-clippy/master/#large_enum_variant

Some related fixes as well, view by commit for full explanation.

Clippy tracking issue: https://github.com/habitat-sh/habitat/issues/6213

### DIFF
--- a/components/butterfly/build.rs
+++ b/components/butterfly/build.rs
@@ -14,6 +14,8 @@ fn main() {
 
 fn generate_protocols() {
     let mut config = prost_build::Config::new();
+    config.type_attribute(".butterfly.newscast.Rumor.payload",
+                          "#[allow(clippy::large_enum_variant)]");
     config.type_attribute(".", "#[derive(Serialize, Deserialize)]");
     config.compile_protos(&protocol_files(), &protocol_includes())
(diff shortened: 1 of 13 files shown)


components/butterfly/build.rs
```diff
@@ -14,6 +14,8 @@ fn main() {
 
 fn generate_protocols() {
     let mut config = prost_build::Config::new();
+    config.type_attribute(".butterfly.newscast.Rumor.payload",
+                          "#[allow(clippy::large_enum_variant)]");
```
christophermaier: Nice one.

components/sup/src/cli.rs
```diff
@@ -29,6 +28,7 @@ mod test {
 
     mod sup_run {
         use super::*;
+        use std::iter::FromIterator as _;
```
christophermaier: That's an interesting new trick!

baumanj:
The CI is going to fail until we get the build kite runner upgraded to rust 1.33.0

baumanj:
Rebased now that both linux and windows CI images are on rust 1.33.0

PR #6278, opened 2019-03-08
Make CLI channel args use CLAP env and default_values

baumanj:
Also rename a channel variable which confusingly referred to a different kind of channel in the same general area.

Resolves https://github.com/habitat-sh/habitat/issues/6168


thesentinels:

Thank you for contributing!

PR #6281, opened 2019-03-11
Update `type_` to `r#type` for prost 2018 edition update

baumanj:
Previously prost would generate code for a field named `type` in the .proto file as `type_`. With the updates to 2018 edition and raw identifier support, it is now `r#type`. This change updates the
generated code and the code that references it. 

See https://github.com/danburkert/prost/pull/159


thesentinels:

Thank you for contributing!

PR #6290, opened 2019-03-13
implement platform specific UI symbols

mwrock:
Our symbols in the UI module like `☁` and `☑` do not render in most Windows use cases. They should render fine in a console like ConEmu or cygwin but not in a standard Windows console (`cmd.exe` or `powershell.exe`). They also do not render in a Docker TTY (which is what a Docker studio uses) even if inside of a "friendly" console like ConEmu.

The standard Windows consoles can only render a subset of our symbols depending on the font that the console is set to. More modern console's have the concept of a fallback font so that they can find a font that supports a unicode character even if the primary font has no representation of the character to be rendered. It is possible to install a font on Windows that provides characters for all the symbols we use, but we do not want to be in the business of distributing fonts. So instead we will support multiple `UISymbolStyle`s: `full`, `limited` and `ascii`.

By default on non-Windows systems, `full` will be the default and it provides all the symbols we currently use today. On Windows, `limited` is the default which supports most of these symbols and substitutes others for the few that the default console fonts do not support.

The Docker TTY uses an ANSI code parser that filters out any non standard ascii character and replaces them with `?`. Thus, an `ascii` style will only render standard ascii characters instead of unicode glyphs. This style will be used in the Docker Windows studio.

The `UISymbolStyle` can be set via the `HAB_SYMBOL_STYLE` environment variable.

We also no longer force `ascii` encoding when redirecting the supervisor log in a local windows studio. This used to be necessary in beta versions of Powershell code but now Powershell uses a bomless UTF-8 default encoder and this will not strip out our symbols.

When running headless in a Windows service, we will always set the style to `ascii` because output is redirected through the default system locale which will distort the glyph encoding.

Finally, this PR also abstracts the colors to a `UIColor` enum as requested in the review of https://github.com/habitat-sh/habitat/pull/6256. This will allow us to more easily adjust colors used based on platform or user preference.

### DIFF
--- a/components/common/src/error.rs
+++ b/components/common/src/error.rs
@@ -35,6 +35,7 @@ pub enum Error {
     ArtifactIdentMismatch((String, String, String)),
     /// Occurs when there is no valid toml of json in the environment variable
     BadEnvConfig(String),
+    BadGlyphStyle(String),
     CantUploadGossipToml,
     ChannelNotFound,
     CryptoKeyError(String),
@@ -95,6 +96,7 @@ impl fmt::Display for Error {
             Error::BadEnvConfig(ref varname) => {
                 format!("Unable to find valid TOML or JSON in {} ENVVAR", varname)
             }
+            Error::BadGlyphStyle(ref style) => format!("Unknown symbol style '{}'", style),
             Error::CantUploadGossipToml => {
                 "Can't upload gossip.toml, it's a reserved file name".to_string()
             }
@@ -162,6 +164,7 @@ impl error::Error for Error {
             Error::APIClient(ref err) => err.description(),
             Error::ArtifactIdentMismatch((..)) => "Artifact ident does not match expected ident",
             Error::BadEnvConfig(_) => "Unknown syntax in Env Configuration",
+            Error::BadGlyphStyle(_) => "Unknown symbol style",
             Error::CantUploadGossipToml => "Can't upload gossip.toml, it's a reserved filename",
             Error::ChannelNotFound => "Channel not found",
             Error::CryptoKeyError(_) => "Missing or invalid key",
--- a/components/common/src/ui.rs
+++ b/components/common/src/ui.rs
@@ -23,13 +23,13 @@ use std::{env,
                Stdout,
                Write},
           process::{self,
-                    Command}};
+                    Command},
+          str::FromStr};
 use uuid::Uuid;
 
 use crate::api_client::DisplayProgress;
 use pbr;
 use termcolor::{self,
-                Color,
                 ColorChoice,
                 ColorSpec,
                 StandardStream,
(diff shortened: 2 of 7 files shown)


thesentinels:

Thank you for contributing!

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::UpArrow => "↑",
+                    UISymbol::FingerPoint => "☛",
+                    UISymbol::CheckMark => "√",
+                    UISymbol::BoxedCheckMark => "☑",
+                    UISymbol::Omega => "Ω",
+                    UISymbol::BoxedX => "☒",
+                    UISymbol::RightArrow => "→",
+                    UISymbol::Cloud => "☁",
+                    UISymbol::DownArrow => "↓",
+                    UISymbol::Elipses => "…",
```
baumanj: Technically Ellipsis. Elipses would be multiple: `……`

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::DottedTriangle => "∵",
```
baumanj: Based on the way we use this, I think it's unicode name is more appropriate:
∵
BECAUSE
Unicode: U+2235, UTF-8: E2 88 B5

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::RightShift => "»",
+                    UISymbol::Star => "★",
+                    UISymbol::SlashedZero => "Ø",
```
baumanj: I wonder if we want a slightly different character (and correspondingly name) here:
∅
EMPTY SET
Unicode: U+2205, UTF-8: E2 88 85

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::RightShift => "»",
+                    UISymbol::Star => "≡",
+                    UISymbol::SlashedZero => "Ø",
+                    UISymbol::ErrorX => "XXX",
+                }
+            }
+            UISymbolStyle::Full => {
+                match *self {
+                    UISymbol::UpArrow => "↑",
+                    UISymbol::FingerPoint => "☛",
```
baumanj: Based on the way this is *used*, I think a better name would be `::InProgress`. Since the rendered glyphs are going to vary, the name should express the common concept, right?

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::CheckMark => "√",
```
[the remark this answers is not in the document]
baumanj: Also, based on how it's used, I think `::Completed` would be a more appropriate name

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::BoxedCheckMark => "☑",
```
baumanj: Based on use: `::Cached`

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::Omega => "Ω",
```
baumanj: I think we should actually get rid of this one. The only place it's used is in a "Creating" message, and I think that would be more consistent to use `::FingerPoint` (or `::InProgress`) since that's consistently used for all the "-ing" statuses.

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::BoxedX => "☒",
```
baumanj: This is only used for "Deleting`, so I think a name like `::Deleting` would be better, and perhaps the "⌦" glyph instead. Or, we could eliminate this and use `::FingerPoint` (or `::InProgress`) to be consistent with the other "in-progress" statuses.

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::RightArrow => "→",
```
baumanj: This is used somewhat inconsistently for all the following statuses
```
Demoting
Found
Generated
Promoting
Using
Writing
```
I think it would be more useful UX to make the "-ing" statuses use the same glyph we use elsewhere for in-progress actions and the complete actions can use `::CheckMark` (or `::Completed`). Also, "Demoted" could use something like

↩︎
LEFTWARDS ARROW WITH HOOK
Unicode: U+21A9 U+FE0E, UTF-8: E2 86 A9 EF B8 8E

or

⎌
UNDO SYMBOL
Unicode: U+238C, UTF-8: E2 8E 8C

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::Cloud => "☁",
```
baumanj: This is only used for "Determining", perhaps the in-progress glyph would be better

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::DownArrow => "↓",
+                    UISymbol::Elipses => "…",
+                    UISymbol::DottedTriangle => "∵",
+                    UISymbol::RightShift => "»",
```
baumanj: Since this isn't used for a conceptual right shift, perhaps a better name would align with it's only usage: `::Begin`

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
...
+                    UISymbol::Star => "★",
```
baumanj: Similarly, I think this could be called `::End`

components/common/src/ui.rs
baumanj: Though, based on use, I think a more appropriate name would be `::Warning` and the more appropriate glyph may be
⚠︎
WARNING SIGN
Unicode: U+26A0 U+FE0E, UTF-8: E2 9A A0 EF B8 8E

components/common/src/ui.rs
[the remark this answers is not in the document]
mwrock: We actually used check mark before which was not available in the default windows fonts. square root was EXTREMELY close and looks like a check mark in all the fonts I checked out. What I'll do is keep it a check mark in `full` and go with the square root glyph in `limited`

components/common/src/ui.rs
mwrock: yeah I admit it seemed odd that we were using this character. I don't get the connection with warning. I just didn't want this PR to be about changing the underlying symbols where they already render.

components/common/src/ui.rs
mwrock: I had an initial draft where instead of using a literal term describing the glyph, I used a term describing what it is intended to indicate (like "progress"). What I personally found was that several of the glyphs could be called "progress" and in the end it just seemed simpler to call the symbol what it was for now and maybe move to a different approach later that trims down our glyphs to a smaller set with labels that indicates what action they represent.

components/common/src/ui.rs
baumanj: I mean, we're already doing that to a point, right? If you want to have another PR that directly follows up on this one, I'm happy with that, but otherwise I worry we won't ever come back to fix this situation.

components/common/src/ui.rs
baumanj: Again, I'm fine with moving this work to a follow-up PR if it's going to happen immediately after this. I think we we have a number of different glyphs that all mean "progress", we should consider whether using different ones improves the user experience, or if it would be more illustrative to consolidate them.

components/common/src/ui.rs
```diff
@@ -43,6 +44,169 @@ pub const NONINTERACTIVE_ENVVAR: &str = "HAB_NONINTERACTIVE";
 
 pub const NOCOLORING_ENVVAR: &str = "HAB_NOCOLORING";
 
+pub const SYMBOL_STYLE_ENVVAR: &str = "HAB_SYMBOL_STYLE";
```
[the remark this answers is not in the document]
apriofrost: Thanks for reaching out! @baumanj I'm not an expert on this topic. But I'll repost this to our UX team channel to see if there is anything they could help or provide any feedback.

mwrock:
ok @baumanj I think I covered everything I want to tackle in this PR. Definitely looking better!

components/common/src/error.rs
```diff
@@ -35,6 +35,7 @@ pub enum Error {
     ArtifactIdentMismatch((String, String, String)),
     /// Occurs when there is no valid toml of json in the environment variable
     BadEnvConfig(String),
+    BadSymbolStyle(String),
```
baumanj: Should this be `BadGlyphStyle` now for consistency?

mwrock: ahh! good catch.

PR #6296, opened 2019-03-14
Use subshells and unset to actually unset environment variables

christophermaier:
https://github.com/habitat-sh/habitat/pull/6278 made a change to
environment variable handling around the channel that `hab` uses to
install packages. As part of this, we no longer treated an empty
string as being synonymous with "unset".

This, in turn, revealed places in our shell code where the intention
was to unset environment variables, but in practice, they were set to
empty strings.

This PR changes our `_hab` wrapper function in our various studio
implementations to run in a subshell, using POSIX `unset` to remove
variables from the environment completely. Busybox's `env` also
supports the `-u` syntax, which we could have used, but we've opted to
keep with POSIX whenever possible.

Big thanks to @baumanj for this solution.

### DIFF
--- a/components/studio/libexec/hab-studio-type-bare.sh
+++ b/components/studio/libexec/hab-studio-type-bare.sh
@@ -114,10 +114,10 @@ EOT
   studio_env_command="$busybox_path/bin/env"
 }
 
-_hab() {
-  # shellcheck disable=2154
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

chefsalim:
tenor-33756550

christophermaier:
Rebased, with added comments per [feedback](https://github.com/habitat-sh/habitat/pull/6296#pullrequestreview-214283956)

PR #6297, opened 2019-03-14
Make Linux Kernel 2 validation just a little bit easier

christophermaier:
To validate `x86_64-linux-kernel2` target releases, we need a Linux
system with an old enough kernel. We have a `Vagrantfile` that does a
basic setup of a CentOS 6 machine, but there are extra steps you need
to take to run services, build plans, and so on.

This PR adds a bit more automation to help with some of that. There
are still manual steps (namely, getting a release candidate `hab`
binary on the box), but this change makes things a lot easier.

The changes are:

* Add a `hab` user and group

  If you want to run a service to test things out, you're probably
  going to need to have those anyway.

* Check out the core-plans repository

  If you want to build a package, you're going to need some plans!

* Install secret keys

  When you want to build a package, you'll need signing keys. When
  provisioning the VM, we'll grab the secret key that corresponds to
  the values of `HAB_ORIGIN` and `HAB_AUTH_TOKEN` in your
  workstation's environment. When you build in the VM, just set
  `HAB_ORIGIN` in the environment of the `hab pkg build` call.

  To do this, we also install the current stable `hab` binary. When
  validating a release candidate, you'll still need to manually get
  _that_ binary on disk. Having the stable binary around helps make
  this a more self-contained system, and also makes this VM a bit more
  useful outside of pure release candidate validation purposes.

* Added some additional documentation around all this.

  Because who doesn't love documentation?

### DIFF
--- a/support/validation/x86_64-linux-kernel2/Vagrantfile
+++ b/support/validation/x86_64-linux-kernel2/Vagrantfile
@@ -9,6 +9,21 @@ Vagrant.configure("2") do |config|
   end
 
   config.vm.provision "shell", inline: <<-SHELL
-    yum install -y wget curl 
+    yum install -y wget curl git
+
+    # You'll need this for running most services
+    adduser hab
+
+    # You'll want this if you want to build a plan
+    git clone https://github.com/habitat-sh/core-plans.git
+
+    # You'll also want the keys for building for whatever origin you
+    # choose.
+    #
+    # It expects you to have HAB_ORIGIN and HAB_AUTH_TOKEN set in your
+    # environment when you provision the VM.
+    curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh -o install.sh
+    bash install.sh -t x86_64-linux-kernel2
+    sudo -u vagrant hab origin key download #{ENV['HAB_ORIGIN']} --secret --auth=#{ENV['HAB_AUTH_TOKEN']}
   SHELL
 end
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

support/validation/x86_64-linux-kernel2/Vagrantfile
```diff
@@ -9,6 +9,21 @@ Vagrant.configure("2") do |config|
...
+    git clone https://github.com/habitat-sh/core-plans.git
+
+    # You'll also want the keys for building for whatever origin you
+    # choose.
+    #
+    # It expects you to have HAB_ORIGIN and HAB_AUTH_TOKEN set in your
+    # environment when you provision the VM.
+    curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh -o install.sh
+    bash install.sh -t x86_64-linux-kernel2
+    sudo -u vagrant hab origin key download #{ENV['HAB_ORIGIN']} --secret --auth=#{ENV['HAB_AUTH_TOKEN']}
```
smacfarlane: Rather than use a real key, would it make sense to set `HAB_ORIGIN` to something obviously intended for validation, and then generate a key for that origin?   In the context of validating a release, would we ever expect to upload the package we just built? 

Now that I've written that, I think yes we would want upload the package (to acceptance probably) as part of validating a release.   Going to go ahead and hit enter and see what you think.

christophermaier: I really like the idea of having a special origin for packages that are just for testing and validation... I know I've written a few of those, and I'm sure you have, too. It'd be nice to formalize that, share the packages, and also incorporate them into our broader testing initiatives and practices.

I was looking for an easy way to get up and running with something, and this seemed the most straightforward. If / when we have a validation origin, we could tweak this, or run the provisioning with a helper script / makefile target that set `HAB_ORIGIN` to that validation origin.

(I don't think that would need to block this, though.)

christophermaier:
See habitat-sh/habitat#6284

PR #6304, opened 2019-03-15
Cargo update following release 0.78.0

baumanj:
Some fixups were necessary as the toml crate upgrade from 0.4.10 → 0.5.0 change the type alias of `toml::value::Table` from the stdlib `BTreeMap<String, Value>` to toml's own `Map<String, Value>` type.

See
https://docs.rs/toml/0.4.10/toml/value/type.Table.html
https://docs.rs/toml/0.5.0/toml/value/type.Table.html

See https://github.com/habitat-sh/habitat/issues/6284


thesentinels:

Thank you for contributing!

PR #6306, opened 2019-03-16
turn off supervisor log coloring in local windows studio if ANSI is not supported

mwrock:
On Windows versions prior to 10.0.10586(TH2), the standard windows console did not understand ANSI code sequences. While we have adopted the `termcolor` crate to use console APIs instead of ANSI sequences on Windows, `termcolor` will fall back to ansi if it cannot find a console. This will happen if habitat is running in a headless environment like a windows service or in a local studio where we launch it in the background.

We already specify `--no-color` as a default parameter in the `windows-service` but the local studio should turn off color if runing in a older version of windows like 2012R2 or even an earlier 2016 build.

This sniffs the OS version and adds `--no-color` and also sets `HAB_NOCOLORING` on non ANSI supporting versions:

BEFORE:
image

AFTER:
image


thesentinels:

Thank you for contributing!

PR #6312, opened 2019-03-18
windows docker studio should output sup log in color and without timestamp

mwrock:
The comment basically says it all:

```
# Install the habitat windows service and then allow color and strip the log timestamps.
# Because this is a studio we are not worried about ANSI codes in the log. Users will 
# view them by tailing the log to the console. Because we are in a Docker studio, it is safe to
# assume that the terminal supports ANSI sequence codes.
```

I tested this locally and the studio is built as desired.

### DIFF
--- a/.buildkite/scripts/build_docker_image.ps1
+++ b/.buildkite/scripts/build_docker_image.ps1
@@ -46,11 +46,19 @@ try {
     $version = "$($pathParts[-2])-$($pathParts[-1])"
     
 @"
+# escape=``
 FROM microsoft/windowsservercore
 MAINTAINER The Habitat Maintainers <humans@habitat.sh>
 ADD rootfs /
 WORKDIR /src
-RUN /hab/pkgs/$ident/bin/hab/hab.exe pkg exec core/windows-service install
+SHELL ["powershell", "-Command", "`$ErrorActionPreference = 'Stop'; `$ProgressPreference = 'SilentlyContinue';"]
+# Install the habitat windows service and then allow color and strip the log timestamps.
+# Because this is a studio we are not worried about ANSI codes in the log. Users will 
+# view them by tailing the log to the console. Because we are in a Docker studio, it is safe to
+# assume that the terminal supports ANSI sequence codes.
+RUN &/hab/pkgs/$ident/bin/hab/hab.exe pkg exec core/windows-service install; ``
+    (Get-Content /hab/svc/windows-service/HabService.dll.config).replace('--no-color', '') | Set-Content /hab/svc/windows-service/HabService.dll.config; ``
+    (Get-Content /hab/svc/windows-service/log4net.xml).replace('%date - ', '') | Set-Content /hab/svc/windows-service/log4net.xml
 ENTRYPOINT ["/hab/pkgs/$ident/bin/powershell/pwsh.exe", "-ExecutionPolicy", "bypass", "-NoLogo", "-file", "/hab/pkgs/$ident/bin/hab-studio.ps1"]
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

mwrock:
closes #2122

.buildkite/scripts/build_docker_image.ps1
```diff
@@ -46,11 +46,19 @@ try {
     $version = "$($pathParts[-2])-$($pathParts[-1])"
     
 @"
+# escape=``
 FROM microsoft/windowsservercore
 MAINTAINER The Habitat Maintainers <humans@habitat.sh>
 ADD rootfs /
 WORKDIR /src
-RUN /hab/pkgs/$ident/bin/hab/hab.exe pkg exec core/windows-service install
+SHELL ["powershell", "-Command", "`$ErrorActionPreference = 'Stop'; `$ProgressPreference = 'SilentlyContinue';"]
```
christophermaier: @mwrock I yield to your Windows chops, but can we extract this to a stand-alone `Dockerfile` rather than baking it into a script?

I was noticing this about our Linux studio container creation just this morning... the only way it appears that we have to build such containers is embedded into our Buildkite code. If that logic was externalized, it'd be easier for developers and external contributors to run.

.buildkite/scripts/build_docker_image.ps1
```diff
@@ -46,11 +46,19 @@ try {
...
+# Install the habitat windows service and then allow color and strip the log timestamps.
+# Because this is a studio we are not worried about ANSI codes in the log. Users will 
+# view them by tailing the log to the console. Because we are in a Docker studio, it is safe to
+# assume that the terminal supports ANSI sequence codes.
+RUN &/hab/pkgs/$ident/bin/hab/hab.exe pkg exec core/windows-service install; ``
+    (Get-Content /hab/svc/windows-service/HabService.dll.config).replace('--no-color', '') | Set-Content /hab/svc/windows-service/HabService.dll.config; ``
```
christophermaier: Is this the only mechanism we have to affect the configuration? Is this expected to be the only scenario in which we'd want to do this?

.buildkite/scripts/build_docker_image.ps1
```diff
@@ -46,11 +46,19 @@ try {
...
+    (Get-Content /hab/svc/windows-service/log4net.xml).replace('%date - ', '') | Set-Content /hab/svc/windows-service/log4net.xml
```
christophermaier: Similarly here.

Why do we need to strip the timestamps?

.buildkite/scripts/build_docker_image.ps1
mwrock: I think so. In any "normal" (ie non-studio) windows service install, we want the logs to be consumable in a non-console context. So it makes sense in all cases to make sure they have no ANSI codes and stripping the timestamps is nice to make the studio sup output "look" like normal supervisor output identical to a local studio or a linux studio.

The `windows-service` readme has instructions on editing the config and log4net.xml to satisfy a user's unique requirements but the defaults we have I think are good for most cases.

.buildkite/scripts/build_docker_image.ps1
mwrock: it's actually pretty easy to create a windows docker studio using this script. I'm going to create a separate issue for removing the old non-buildkite bintray-publish script and add to the studio readme about using this one.

mwrock: see https://github.com/habitat-sh/habitat/issues/6317

PR #6313, opened 2019-03-19
eliminate false error noise from windows launcher

### LINKED ISSUE #6148: windows launcher shutdown emits errors regarding pipe removal
When shutting down the supervisor on windows, this error is emitted when the launcher shuts down:

```
2019-02-12 11:31:59,686 - [2019-02-12T19:31:59Z ERROR habitat_launcher::server] Could not remove old pipe to supervisor 3ff12965-ffe4-4dc4-9f1d-5001704c7cb4
2019-02-12 11:31:59,686 - [2019-02-12T19:31:59Z ERROR hab_launch] Launcher exiting with code 0
```

This can be ignored with no adverse behavior but the error is confusing and alarming and we should fix this.

mwrock:
This fixes #6148. When the launcher terminates, it attempts to remove the IPC pipe files. This is absolutely the right thing to do on *nix but on Windows the file is not user deletable and is saved to the NPFS (Named Pipe File System). It is automatically deleted when all handles to the named pipe are freed. The ipc-channel crate wraps the named pipe in a `WinHandle` struct which calls `CloseHandle` on its handle in its `drop` implementation.

So the attempt to delete the named pipe file on windows results in an error that is sent to the supervisor output and can give the perception that bad things are happening at worst and is just noisy at best.

This pr will only remove the file on `unix` it also changes the `error!` log in the launcher's `main.rs` to an `info!` if the launcher server exits `0`.

### DIFF
--- a/components/launcher-client/src/client.rs
+++ b/components/launcher-client/src/client.rs
@@ -21,7 +21,6 @@ use ipc_channel::ipc::{IpcOneShotServer,
                        IpcReceiver,
                        IpcSender};
 use std::{collections::HashMap,
-          fs,
           io,
           path::Path};
 
@@ -29,14 +28,20 @@ type Env = HashMap<String, String>;
 type IpcServer = IpcOneShotServer<Vec<u8>>;
 
 pub struct LauncherCli {
-    tx:   IpcSender<Vec<u8>>,
-    rx:   IpcReceiver<Vec<u8>>,
+    tx: IpcSender<Vec<u8>>,
+    rx: IpcReceiver<Vec<u8>>,
+    // We persist the pipe identifier so we can delete the file on drop.
+    // This is not necessary on Windows because named pipes are removed
+    // upon releasing the last handle to the pipe. The ipc-channel crate
+    // wraps the pipe in a WinHandle whose drop impl calls CloseHandle.
+    #[cfg(not(windows))]
     pipe: String,
 }
 
+#[cfg(not(windows))]
 impl Drop for LauncherCli {
     fn drop(&mut self) {
-        if fs::remove_file(&self.pipe).is_err() {
+        if std::fs::remove_file(&self.pipe).is_err() {
             error!("Could not remove old pipe to launcher {}", self.pipe);
         } else {
             debug!("Removed old pipe to launcher {}", self.pipe);
@@ -56,6 +61,7 @@ impl LauncherCli {
         Self::read::<protocol::NetOk>(&raw)?;
         Ok(LauncherCli { tx,
                          rx,
+                         #[cfg(not(windows))]
                          pipe: pipe_to_sup })
     }
 
--- a/components/launcher/src/main.rs
+++ b/components/launcher/src/main.rs
@@ -17,7 +17,9 @@ use habitat_common::output::{self,
                              OutputFormat,
                              OutputVerbosity};
 use habitat_launcher::server;
-use log::error;
+use log::{error,
+          log,
+          Level};
 use std::{env,
           process};
 
@@ -33,7 +35,8 @@ fn main() {
             process::exit(1);
         }
         Ok(code) => {
-            error!("Launcher exiting with code {}", code);
+            let level = if code == 0 { Level::Info } else { Level::Error };
+            log!(level, "Launcher exiting with code {}", code);
             process::exit(code);
         }
     }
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/launcher-client/src/client.rs
```diff
@@ -36,7 +36,7 @@ pub struct LauncherCli {
 
 impl Drop for LauncherCli {
     fn drop(&mut self) {
-        if fs::remove_file(&self.pipe).is_err() {
+        if cfg!(unix) && fs::remove_file(&self.pipe).is_err() {
```
baumanj: If Windows doesn't need to do anything here, why not make the whole `impl Drop` `#[cfg(not(windows))]`? As is, it will misleadingly print "Removed old pipe to launcher" on Windows, when this code has done no such thing.

Also, I suggest `(not(windows))` that rather than `(unix)` since I think we should err on the side of doing the cleanup, and experience will tell us when the cleanup we're doing is wrong. If we err on the side of doing nothing, we won't get any signals that we should be doing something.

components/launcher/src/main.rs
```diff
@@ -32,6 +33,10 @@ fn main() {
             error!("Launcher exiting with 1 due to err: {}", err);
             process::exit(1);
         }
+        Ok(0) => {
+            info!("Launcher exiting with code 0");
+            process::exit(0);
+        }
```
baumanj: I like the change to make a zero exit status an info-level log. I think we can make use of [the `log!` macro](https://docs.rs/log/0.4.6/log/macro.log.html) to emphasize the fact that only the *logging* is different in these two cases:
```rust
        Ok(code) => {
            log!(if code == 0 { Level::Info } else { Level::Error },
                 "Launcher exiting with code {}",
                 code);
            process::exit(code);
        }
```
or, if you like less nesting:
```rust
        Ok(code) => {
            let level = if code == 0 { Level::Info } else { Level::Error };
            log!(level, "Launcher exiting with code {}", code);
            process::exit(code);
        }
```

components/launcher/src/server/mod.rs
```diff
@@ -126,7 +126,7 @@ impl Server {
     }
 
     fn remove_pipe(&self) {
-        if fs::remove_file(&self.pipe).is_err() {
+        if cfg!(unix) && fs::remove_file(&self.pipe).is_err() {
```
baumanj: As with the other call, I think we should make the condition a level higher, or else get incorrect logging in the Windows case. Since we still need to remove the PID file, perhaps something like:
```rust
    fn drop(&mut self) {
        fs::remove_file(&self.pid_file_path).ok();
        #[cfg(not(windows))]
        self.remove_pipe();
    }
```
And similarly for the other `remove_pipe` call we want to omit on windows. Additionally, we can silence the windows-only warning about `remove_pipe` being unused:
```rust
    #[cfg_attr(windows, allow(dead_code))]
    fn remove_pipe(&self) {
```

mwrock: Thanks! would it be better to use `#[cfg(not(windows))]` on the `remove_pipe` function since you would never want to call that function on windows. Just allowing the dead code seems to imply that not calling it is ok but you could call it if you wanted to.

baumanj: Yes, using  `#[cfg(not(windows))]` on the `remove_pipe` would make more sense. I'm not sure what I was thinking there.

components/launcher-client/src/client.rs
```diff
@@ -21,21 +21,24 @@ use ipc_channel::ipc::{IpcOneShotServer,
...
 
 type Env = HashMap<String, String>;
 type IpcServer = IpcOneShotServer<Vec<u8>>;
 
 pub struct LauncherCli {
-    tx:   IpcSender<Vec<u8>>,
-    rx:   IpcReceiver<Vec<u8>>,
+    tx: IpcSender<Vec<u8>>,
+    rx: IpcReceiver<Vec<u8>>,
+    #[cfg_attr(windows, allow(dead_code))]
```
baumanj: In this case the field is still necessary on windows, it's just not *assigned* to, so it triggers the `dead_code` warning, is that right? A comment to explain that subtlety may be worthwhile.

components/launcher-client/src/client.rs
```diff
@@ -21,21 +21,24 @@ use ipc_channel::ipc::{IpcOneShotServer,
...
     pipe: String,
 }
 
+#[cfg(not(windows))]
 impl Drop for LauncherCli {
     fn drop(&mut self) {
+        use std::fs;
```
baumanj: You could save a few lines and omit the `use` statement with `std::fs::remove_file`

Issue #6314, opened 2019-03-19
Remove core's nonstandard env::var functions


baumanj:
The `core` crate has [`var`](https://github.com/habitat-sh/core/blob/master/components/core/src/env.rs#L42) and [`var_os`](https://github.com/habitat-sh/core/blob/master/components/core/src/env.rs#L76) functions which are identical to the [stdlib versions](http://doc.rust-lang.org/std/env/index.html) in interface, but have a subtle, yet important difference in behavior. In the event that a environment variable is set to the empty string, `core`'s `var` and `var_os` treat it as though the variable were not set at all.

This, combined with assumptions in shell code that set certain environment variables to empty strings rather than properly unsetting them (see the fix: https://github.com/habitat-sh/habitat/pull/6296), combined with the move to [standardize on CLAP's environment variable override handling](https://github.com/habitat-sh/habitat/pull/6278) led to an escape defect in the 0.78.0 release: https://forums.habitat.sh/t/we-fixed-an-escaped-defect-with-linux-0-78-0-studio/1048.

In order to avoid subtle surprises like this in the future, this issue will track the auditing of our use of `core::env::{var, var_os}` and the setting of environment variables to the empty string. Hopefully we can eliminate this confusing practice entirely, but if not, we can at least make it less confusing.

Links are to instances that need fixing, or once in the process of fixing, PRs that make the fix. **Bold items** are PRs currently in review.

### Uses of `core::env::{var, var_os}`

Many of these can probably be harmlessly changed to using the `std::env` versions, but we should give careful consideration to whether the behavior in the case of the empty string is what we want.

#### Core
- [ ] [habitat_core::binlink::default_binlink_dir](https://github.com/habitat-sh/core/blob/master/components/core/src/binlink.rs#L29) (remove `binlink` module)
  - [x] [habitat](https://github.com/habitat-sh/habitat/pull/6315)
  - [x] builder (does not use)
- [ ] [habitat_core::crypto::default_cache_key_path](https://github.com/habitat-sh/core/blob/bba280adf5c46e57569caa1e15034747970ec245/components/core/src/crypto/mod.rs#L272)
  - [ ] habitat ([part 1](https://github.com/habitat-sh/habitat/pull/6326)) ([part 2](https://github.com/habitat-sh/habitat/pull/6329)) ([part 3](https://github.com/habitat-sh/habitat/pull/6336)) (part 4, should be the last)
  - [x] builder (does not use)
[...]

PR #6315, opened 2019-03-19
Remove usage of habitat_core::binlink::default_binlink_dir

baumanj:
See https://github.com/habitat-sh/habitat/issues/6314

Since only habitat used it, we can move the necessary logic of `habitat_core::binlink::default_binlink_dir` into `habitat` itself. While doing so, we switch to using the standard [clap](https://clap.rs) approach to environment variable overrides, and add validation that the binlink dir is not set to the empty string.

After this merges, we should remove the `binlink` module from `core`.


thesentinels:

Thank you for contributing!

PR #6322, opened 2019-03-20
uninstall should not rename to existing directory on windows

### LINKED ISSUE #6277: pkg uninstall broken on windows
On 0.77.0 and likely all versions since `hab pkg uninstall` was introduced, uninstalling a package on Windows results in:

```
C:\dev\habitat [sup_color]> hab pkg uninstall mwrock/blank1
» Uninstalling mwrock/blank1/0.1.0/20181212111157
☒ Deleting mwrock/blank1/0.1.0/20181212111157
✗✗✗
✗✗✗ Access is denied. (os error 5)
✗✗✗
```

mwrock:
fixes #6277. 

Uninstalling a package creates a temporary directory, renames the existing package directory to that temp directory and then deletes the temp directory. This unfortunately and rather unexpectedly is a classic case where unix and windows underlying IO operations differ.

The details are expressed in https://doc.rust-lang.org/std/fs/fn.rename.html but essentially given rust's rename implementation, you cannot rename a directory to an existing empty directory on Windows. On *nix you must rename to an existing empty directory.

The simplest fix is to just delete the newly created temp directory on windows. It just needs a path.

### DIFF
--- a/components/hab/src/command/pkg/uninstall.rs
+++ b/components/hab/src/command/pkg/uninstall.rs
@@ -12,25 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::{ExecutionStrategy,
+            Scope};
+use crate::error::{Error,
+                   Result};
+use habitat_common::{package_graph::PackageGraph,
+                     ui::{Status,
+                          UIWriter,
+                          UI}};
+use habitat_core::{error as herror,
+                   fs as hfs,
+                   package::{list::temp_package_directory,
+                             Identifiable,
+                             PackageIdent,
+                             PackageInstall}};
 use std::{fs,
           path::Path,
           str::FromStr};
-
-use super::{ExecutionStrategy,
-            Scope};
-use crate::{common::{package_graph::PackageGraph,
-                     ui::{Status,
-                          UIWriter,
-                          UI}},
-            error::{Error,
-                    Result},
-            hcore::{error as herror,
-                    fs as hfs,
-                    package::{Identifiable,
-                              PackageIdent,
-                              PackageInstall}}};
-
-use crate::hcore::package::list::temp_package_directory;
 
 /// Delete a package and all dependencies which are not used by other packages.
 /// We do an ordered traverse of the dependencies, updating the graph as we delete a
@@ -217,7 +215,8 @@ fn do_clean_delete(pkg_root_path: &Path, real_install_path: &Path) -> Result<boo
     // below the pkg_root_path
     match real_install_path.parent() {
         Some(real_install_base) => {
-            let temp_install_path = temp_package_directory(real_install_path)?;
+            let temp_install_path = temp_package_directory(real_install_path)?.path()
+                                                                              .to_path_buf();
             fs::rename(&real_install_path, &temp_install_path)?;
             fs::remove_dir_all(&temp_install_path)?;
 


thesentinels:

Thank you for contributing!

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -217,7 +217,11 @@ fn do_clean_delete(pkg_root_path: &Path, real_install_path: &Path) -> Result<boo
     // below the pkg_root_path
     match real_install_path.parent() {
         Some(real_install_base) => {
+            // create empty temp directory
             let temp_install_path = temp_package_directory(real_install_path)?;
```
christophermaier: When trying to find the `temp_package_directory` function, I noticed that the `use` statements at the top of this module are in separate "chunks" and also mix up a lot of external crate usage with internal usage.

While you're in here, it'd be a great cleanup opportunity!

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -217,7 +217,11 @@ fn do_clean_delete(pkg_root_path: &Path, real_install_path: &Path) -> Result<boo
...
+            // on windows fs::rename cannot rename to an existing directory even if empty
+            #[cfg(windows)]
+            fs::remove_dir(&temp_install_path)?;
```
christophermaier: So, what we're really saying here is that we need `temp_package_directory` to provide us with a suitable _path_, rather than an actual directory, correct?

Since `temp_package_directory` returns a [TempDir](https://docs.rs/tempfile/3.0.7/tempfile/struct.TempDir.html), we may be able to take advantage of it's `Drop` behavior here. I'm thinking something like this:

```rust
let temp_install_path = {
    let td = temp_package_directory(real_install_path)?;
    td.into_path()
};
```

or, more succinctly:

```rust
 let temp_install_path = temp_package_directory(real_install_path)?.into_path();
```

That should end up removing the directory, and giving you the path you need, and it should be the same on both platforms.

mwrock: but we only want to do that on windows. We actually do need the directory on linux. In that case I'm wondering if just the single platform constrained delete line is better.

christophermaier: Unless I'm missing something, you don't actually need the empty directory on Linux in order to rename it.

```sh
maier@ubuntu:~/testing$ ls
test.rs
maier@ubuntu:~/testing$ cat test.rs 
use std::fs;

fn main() {
    fs::create_dir_all("foo").expect("couldn't create foo");
    fs::rename("foo", "bar").expect("couldn't rename to bar");
}
maier@ubuntu:~/testing$ rustc test.rs 
maier@ubuntu:~/testing$ ls
test  test.rs
maier@ubuntu:~/testing$ ./test
maier@ubuntu:~/testing$ ls
bar  test  test.rs
maier@ubuntu:~/testing$ ls -alh bar
total 8.0K
drwxr-xr-x 2 maier maier 4.0K Mar 20 16:04 .
drwxr-xr-x 3 maier maier 4.0K Mar 20 16:04 ..
maier@ubuntu:~/testing$ touch bar/blah
maier@ubuntu:~/testing$ ./test
thread 'main' panicked at 'couldn't rename to bar: Os { code: 39, kind: Other, message: "Directory not empty" }', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
maier@ubuntu:~/testing$ 

```

I think the docs are saying that if the `to` directory exists on Linux, it has to be empty.

mwrock: oh ok. cool. I will try that!

mwrock: Ended up that `into_path` "persists" the `TempDir` struct in which case its `drop` does not remove the file. However `path().to_path_buf()` accomplishes the same thing and does allow the `drop` to remove the directory. I validated this on both windows and linux.

christophermaier: Oh, TIL 😄

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -217,7 +215,10 @@ fn do_clean_delete(pkg_root_path: &Path, real_install_path: &Path) -> Result<boo
     // below the pkg_root_path
     match real_install_path.parent() {
         Some(real_install_base) => {
-            let temp_install_path = temp_package_directory(real_install_path)?;
+            let temp_install_path = {
+                let td = temp_package_directory(real_install_path)?;
+                td.path().to_path_buf()
```
christophermaier: We can make this a little simpler with:

```rust
let temp_install_path = temp_package_directory(real_install_path)?.path()
                                                                  .to_path_buf();

```

PR #6326, opened 2019-03-21
Remove uses of core's default_cache_key_path

baumanj:
See https://github.com/habitat-sh/habitat/issues/6314 for why.

This removes basically all the calls in the core hab CLI processing. All of which are of the form:

```rust
    default_cache_key_path(Some(&*FS_ROOT))
```

and replaces them with `cache_key_path_from_matches`, which instead of calling core's nonstandard `env::var` implementation, accesses values which have been validated by clap. To facilitate this, and improve discoverability, all the CLI commands which previously relied on `default_cache_key_path` now have a `--cache-key-path`, which is documented and overridable with the `HAB_CACHE_KEY_PATH` environment variable, but in a safe way, where an empty-string value generates an informative error.

Previously, empty-string values would be treated as specifying the current working directory as the search root. This is no longer legal, as it is too likely to be the result of an erroneous attempt to unset
the override environment variable. If this behavior is desired, the use of the more explicit `.` with either the `HAB_CACHE_KEY_PATH` env var or `--cache-key-path` argument will achieve the same effect more clearly.

Additional uses of default_cache_key_path exist and will be addressed in a subsequent commit.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -27,8 +27,10 @@ use habitat_common::{cli_defaults::{BINLINK_DIR_ENVVAR,
                                     RING_ENVVAR,
                                     RING_KEY_ENVVAR},
                      types::ListenCtlAddr};
-use habitat_core::{crypto::keys::PairType,
+use habitat_core::{crypto::{keys::PairType,
+                            CACHE_KEY_PATH_ENV_VAR},
                    env::Config,
+                   fs::CACHE_KEY_PATH,
                    package::{ident,
                              Identifiable,
                              PackageIdent,
@@ -102,6 +104,9 @@ pub fn get() -> App<'static, 'static> {
                 (@arg USER: -u --user +takes_value "Name of the user key")
                 (@arg REMOTE_SUP: --("remote-sup") -r +takes_value
                     "Address to a remote Supervisor's Control Gateway [default: 127.0.0.1:9632]")
+                (arg: arg_cache_key_path("Path to search for encryption keys. \
+                    Default value is hab/cache/keys if root and .hab/cache/keys under the home \
+                    directory otherwise."))
             )
         )
         (@subcommand bldr =>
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/hab/src/cli.rs
```diff
@@ -102,6 +104,9 @@ pub fn get() -> App<'static, 'static> {
                 (@arg USER: -u --user +takes_value "Name of the user key")
                 (@arg REMOTE_SUP: --("remote-sup") -r +takes_value
                     "Address to a remote Supervisor's Control Gateway [default: 127.0.0.1:9632]")
+                (arg: arg_cache_key_path("Path to search for encryption keys. \
+                    Default value is hab/cache/keys if root and .hab/cache/keys under the home \
+                    directory otherwise."))
```
christophermaier: Since we're using a function to set up the details of the option, and since the help text from "Default value is..." onward is always the same, could we not put that into the function as well? The context specifcity of the first sentence is really nice, though, and I wouldn't want to lose that.

I started thinking down this path because the fact that we're giving relative paths here instead of absolute struck me as odd. Knowing the details of the implementation, I understand why it's that way, but for user-facing documentation it seems like it presumes too much. I'd expect to see `/hab/cache/keys` and `~/.hab/cache/keys` if I were on Linux, for example.

Are users expected to input a relative path?

This may be something for a future PR, but I imagine it would also be nice to show Windows-specific information when using this on Windows, rather than requiring Windows users to mentally translate from Unix-ese.

baumanj: > Since we're using a function to set up the details of the option, and since the help text from "Default value is..." onward is always the same, could we not put that into the function as well? The context specifcity of the first sentence is really nice, though, and I wouldn't want to lose that.

That's what I tried at first. And the answer is that we can't (or I didn't see an obvious way) because the lifetime of the `&str` passed to `help` needs to be `'static`.

> Are users expected to input a relative path?

Both absolute and relative paths are supported. Some of the complication comes from the fact that `hab/cache/keys` isn't necessarily under `/` because `FS_ROOT` is dynamic. I think our whole approach to `hab` vs `.hab` is worth of re-thinking, but I'm trying to keep this particular change narrowly scoped.

christophermaier: Ah, that's right, they've got to be static strings!

👍

PR #6327, opened 2019-03-21
Simplify CLI definition for `hab svc load` command

christophermaier:
Previously, we had two entirely separate implementations depending on
the compilation platform (Unix vs Windows). However, the Windows
implementation differed by only a single command-line option. This was
very difficult to see among all the other duplicated code.

Since duplicated code is often a Bad Thing (TM), this PR simplifies
the definition of the load command by only adding the extra option
when compiling for Windows.

We'll also use the `cfg!(windows)` formulation, rather than
`cfg!(target_os = "windows")` since it's arguably simpler and (as far
as I can tell) equivalent.

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -787,7 +787,7 @@ fn sub_pkg_build() -> App<'static, 'static> {
                                               .long("docker"));
     }
 
-    if cfg!(target_os = "windows") {
+    if cfg!(windows) {
         sub.arg(Arg::with_name("WINDOWS").help("Use a local Windows Studio instead of a Docker \
                                                 Studio")
                                          .short("w")
@@ -1016,9 +1016,8 @@ fn sub_svc_stop() -> App<'static, 'static> {
     )
 }
 
-#[cfg(any(target_os = "linux", target_os = "macos"))]
 fn sub_svc_load() -> App<'static, 'static> {
-    clap_app!(@subcommand load =>
+    let mut sub = clap_app!(@subcommand load =>
         (about: "Load a service to be started and supervised by Habitat from a package \
             identifier. If an installed package doesn't satisfy the given package \
             identifier, a suitable package will be installed from Builder.")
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

baumanj:
> We'll also use the `cfg!(windows)` formulation, rather than
> `cfg!(target_os = "windows")` since it's arguably simpler and (as far
> as I can tell) equivalent.

`cfg!(windows)` is a shorthand for `cfg!(target_family = "windows")`, which I think is identical to `cfg!(target_os = "windows")` in the windows case, but there's no `cfg!(linux)`, since there's no `cfg!(target_family = "linux")`. Instead `cfg!(unix)` seems to cover a variety of `target_os` values.

See https://doc.rust-lang.org/reference/conditional-compilation.html#unix-and-windows

components/hab/src/cli.rs
```diff
@@ -787,7 +787,7 @@ fn sub_pkg_build() -> App<'static, 'static> {
                                               .long("docker"));
     }
 
-    if cfg!(target_os = "windows") {
+    if cfg!(windows) {
```
baumanj: Not exactly your change, but as long as we're touching this: I think this pattern is more suceptible to future defects than it could be. Plus, it's different than the `native_studio_support` code just above it. I suggest we change this from
```rust
    if cfg!(windows) {
        sub.arg(…)
    } else {
        sub
    }
```
to
```rust
    if cfg!(windows) {
        sub = sub.arg(…)
    }

    sub
```

components/hab/src/cli.rs
```diff
@@ -1016,9 +1016,8 @@ fn sub_svc_stop() -> App<'static, 'static> {
     )
 }
 
-#[cfg(any(target_os = "linux", target_os = "macos"))]
 fn sub_svc_load() -> App<'static, 'static> {
-    clap_app!(@subcommand load =>
+    let mut sub = clap_app!(@subcommand load =>
```
baumanj: This is a great simplification! I wonder if for the sake of future extensions it's worth leaving configuration:
```rust
#[cfg(any(unix, windows))]
fn sub_svc_load() -> App<'static, 'static> {
```
It's not strictly necessary now, but in the event we ever do extend to other platforms, it's a nice way to have the compiler point out areas of platform-specific consideration.

components/hab/src/cli.rs
christophermaier: Good call; will fix.

components/hab/src/cli.rs
christophermaier: Given how things appear to be defined in Rust right now, I think this would end up being noise right now. It looks like only the most esoteric of platforms fall outside of `unix` and `windows`, and I don't foresee us going there anytime soon.

Good thought, though 👍

components/hab/src/cli.rs
christophermaier: Fixed in https://github.com/habitat-sh/habitat/pull/6327/commits/52246278888fbba23e5ede6750a0d685f7000722

PR #6328, opened 2019-03-21
Platform-specific UserInfo proof of concept

baumanj:
See https://github.com/habitat-sh/habitat/pull/6305#discussion_r267052477 for more context.

I don't think this is worth going forward with because adding `required` types to launcher.proto shackles the launcher and supervisor together in an undesirable way. But if we keep our protobuf-generated structs with optional types, we are stuck with the decision between `unwrap` during the infallible conversion from `generated::Spawn` -> `Spawn`, or we can't really eliminate the Option<_>`s in UserInfo.


thesentinels:

Thank you for contributing!

baumanj:
This just doesn't seem like a good tradeoff right now.

PR #6332, opened 2019-03-22
Clarify signal behavior

christophermaier:
This is mainly to pick up changes in https://github.com/habitat-sh/core/pull/142


thesentinels:

Thank you for contributing!

components/launcher/src/server/mod.rs
```diff
@@ -263,26 +265,31 @@ impl Server {
...
-            Some(SignalEvent::WaitForChild) => {
-                // We only return Some if we ended up reaping our
-                // Supervisor; otherwise, we don't need to do anything
-                // special. If the supervisor exits but reap_zombie_orphans()
-                // doesn't catch the signal (such as on Windows), we will still
-                // handle that properly in handle_message().
-                if let Some(result) = self.reap_zombie_orphans() {
-                    return result;
+
+        if let Some(event) = signals::check_for_signal() {
```
baumanj: Why not completely disentangle the concepts of "signal as shutdown trigger" and "signal for other uses", so we don't need to sprinkle `#[cfg(unix)]` on individual match arms? Maybe a signals interface that can check only for a specific signal so that all platforms can do that `signals::received_shutdown()` and then put one `#[cfg(unix)]` around a subsequent `signals::check_for_signal()` call.

christophermaier: I'll look into that... My main goal (at least in this draft stage 😄 ) was to get things disentangled enough that we could actually see what was really going on.

This appears to be a useful distinction that has popped up 👍

PR #6336, opened 2019-03-22
Remove uses of default_cache_key_path

baumanj:
See https://github.com/habitat-sh/habitat/issues/6314 for why

There are some associated changes and cleanup, but most of this is just plumbing what were previous environment variable overrides and making them proper CLI argument (with the option for override).

### DIFF
--- a/components/sup/src/manager/mod.rs
+++ b/components/sup/src/manager/mod.rs
@@ -134,7 +134,7 @@ use winapi::{shared::minwindef::PDWORD,
              um::processthreadsapi};
 
 const MEMBER_ID_FILE: &str = "MEMBER_ID";
-const PROC_LOCK_FILE: &str = "LOCK";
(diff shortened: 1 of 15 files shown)


thesentinels:

Thank you for contributing!

components/sup/src/manager/mod.rs
```diff
@@ -1785,6 +1769,29 @@ mod test {
...
+                            http_listen:       http_gateway::ListenAddr::default(),
+                            http_disable:      false,
+                            gossip_peers:      vec![],
+                            gossip_permanent:  false,
+                            ring_key:          None,
+                            organization:      None,
+                            watch_peer_file:   None,
+                            tls_config:        None, }
+        }
+    }
```
christophermaier: tenor-23410050

PR #6339, opened 2019-03-23
Improved searchability of reference docs

echohack:

In real-world use, I've found these reference docs a bit cumbersome to use because they aren't very searchable with ctrl+f or `grep`. I've heard similar things from others.

This small change adds the object to the front of the property in the reference table to make them uniquely searchable, adds a few more practical examples.

This also fixes some minor problems I ran across recently in the binary wrapper doc.

cat_computer

### DIFF
--- a/www/source/partials/docs/_bp-binary-wrapper-packages.html.md.erb
+++ b/www/source/partials/docs/_bp-binary-wrapper-packages.html.md.erb
@@ -36,17 +36,16 @@ Most binaries compiled in a full Linux environment have a hard dependency on `/l
 2. Invoke `patchelf` on any binaries with this problem during the `do_install()` phase. For example:
 
 ```bash
-patchelf --interpreter "$(pkg_path_for glibc)/lib/ld-linux-x86-64.so.2" \
-         ${pkg_prefix}/bin/somebinary
+patchelf --interpreter "$(pkg_path_for core/glibc)/lib/ld-linux-x86-64.so.2" \\
+  "${pkg_prefix}/bin/somebinary"
 ```
 
 3. The binary may have other hardcoded dependencies on its own libraries that you may need to relocate using other flags to `patchelf` like `--rpath`. For example, Oracle Java provides additional libraries in `lib/amd64/jli` that you will need to relocate to the Habitat location:
 
 ```bash
 export LD_RUN_PATH=$LD_RUN_PATH:$pkg_prefix/lib/amd64/jli
-patchelf --interpreter "$(pkg_path_for glibc)/lib/ld-linux-x86-64.so.2" \
-        --set-rpath ${LD_RUN_PATH} \
-        ${pkg_prefix}/bin/java
+patchelf --interpreter "$(pkg_path_for core/glibc)/lib/ld-linux-x86-64.so.2" \\
+  --set-rpath ${LD_RUN_PATH} "${pkg_prefix}/bin/java"
 ```
 
 4. For more information, please see the [patchelf](https://nixos.org/patchelf.html) documentation.
--- a/www/source/partials/docs/_reference-template-data.html.md.erb
+++ b/www/source/partials/docs/_reference-template-data.html.md.erb
@@ -10,17 +10,17 @@ Describes the details of how this specific Supervisor was started
 
 | Property | Type | Description |
 | -------- | ---- | ----------- |
-| version | string | Version of the Habitat Supervisor, e.g., `0.54.0/20180221023448` |
-| member_id | string | The member's Supervisor ID, e.g., `3d1e73ff19464a27aea3cdc5c2243f74` |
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

www/source/partials/docs/_reference-template-data.html.md.erb
```diff
@@ -59,49 +59,49 @@ Information about the current service's service group
...
-| name | string | The name of the Habitat package |
+| origin | string | The origin of the Habitat package. |
+| name | string | The name of the Habitat package. |
 | version | string | The version of the Habitat package |
 | release | string | The release of the Habitat package |
 
 ### svc_member
 
-Represents a member of a service group
+Represents a member of a service group. Expressed by referencing a svc_member object. E.g. `svc.me.alive`
```
kagarmoe: Should the commands in this table be `svc.me.thing`? If that's the correct pattern, then let's apply it consistently.

echohack: Unfortunately no, you'd have to get an arbitrary `svc_member` object. But now that you mention it, it occurs to me that we're also doing this elsewhere, so I'll use an arbitrary designation for it.

echohack: @kagarmoe Alright, I've updated the commit. I don't think there's a good way really of expressing it, but maybe this will be better than nothing.

echohack:
@kagarmoe Solved all those issues for you. :)

kagarmoe:
Thanks! I'm PTO today. I will merge it on Monday.

:-)

echohack:
@kagarmoe gentle ping

echohack:
No rush on this, just hoping we can merge this before it gets too stale. 
image

kagarmoe:
tenor-93520047

www/source/partials/docs/_bp-binary-wrapper-packages.html.md.erb
```diff
@@ -36,17 +36,16 @@ Most binaries compiled in a full Linux environment have a hard dependency on `/l
 2. Invoke `patchelf` on any binaries with this problem during the `do_install()` phase. For example:
 
 ```bash
-patchelf --interpreter "$(pkg_path_for glibc)/lib/ld-linux-x86-64.so.2" \
-         ${pkg_prefix}/bin/somebinary
```
raskchanky: Since this is all one command, you'd still need the `\` at the end to split the line, no?  I'm not sure why the `\` was removed.

www/source/partials/docs/_bp-binary-wrapper-packages.html.md.erb
```diff
@@ -36,17 +36,16 @@ Most binaries compiled in a full Linux environment have a hard dependency on `/l
...
+  "${pkg_prefix}/bin/somebinary"
 ```
 
 3. The binary may have other hardcoded dependencies on its own libraries that you may need to relocate using other flags to `patchelf` like `--rpath`. For example, Oracle Java provides additional libraries in `lib/amd64/jli` that you will need to relocate to the Habitat location:
 
 ```bash
 export LD_RUN_PATH=$LD_RUN_PATH:$pkg_prefix/lib/amd64/jli
-patchelf --interpreter "$(pkg_path_for glibc)/lib/ld-linux-x86-64.so.2" \
-        --set-rpath ${LD_RUN_PATH} \
-        ${pkg_prefix}/bin/java
```
raskchanky: Same comment here as above re: `\`

www/source/partials/docs/_bp-binary-wrapper-packages.html.md.erb
echohack: Screen Shot 2019-04-05 at 15 11 22

I guess this is another one I missed, but yeah, it breaks the site.

echohack: you also don't want to just put it all on one line because then you get a gross horizontal scrollbar

raskchanky: I apologize if my comment wasn't clear.  I agree with you that putting it all on one line is not the answer.

What I meant to say was, when you're displaying an example of shell code, you need the `\` at the end of the line to show that the next line down is a continuation of the same command, and not a new command.  In this particular case, you can put a double backslash `\\` and it will display a single backslash correctly in the rendered code example.  I think that's what we want here, and below.

echohack:
Fixed @raskchanky 

Looks like this span issue is in other places in our docs... I'm going to scour the docs for this in another PR

raskchanky:
@echohack Sounds good.  Thanks for your diligence on that!

raskchanky:
CI on this is green, despite GH showing no status.

PR #6343, opened 2019-03-26
Remove offensively-named command in favor of updated version

baumanj:
See https://redis.io/commands/slaveof
See https://github.com/habitat-sh/habitat/blob/master/CODE_OF_CONDUCT.md#our-standards


thesentinels:

Thank you for contributing!

kagarmoe:
Just, wow. But I'm not sure 'replicaof' is clear to me. Should it be 'followerof' if we are keeping to leader/follower?

stevendanna:
@kagarmoe I think we are limited here with what redis, the example in use, supports as valid configuration.

Issue #5025, opened 2018-05-03
Error about missing /hab/cache/keys directory when entering a studio on a clean hab install


sandratiffin:
I saw the issue on Windows Server 2016 using Windows Containers on two different VMs -- one local using Virtualbox, and one on EC2 -- when first running 'hab studio enter'.

```
PS C:\Users\stiffin\habitat> hab studio enter
0.55.0: Pulling from win-studio

3889bb8d808b: Pull complete
449343c9d7e2: Pull complete
4f5a6afcccf3: Pull complete
183f01b81ada: Pull complete
ae40e62e4eb7: Pull complete
1c2087427f37: Pull complete
b9af171479e3: Pull complete
Digest: sha256:8b821d56b71e75e8279fb97b94e70db506de7fbd6d232387d1bb77de6222314e
Status: Downloaded newer image for habitat-docker-registry.bintray.io/win-studio:0.55.0
C:\Program Files\Docker\docker.EXE: Error response from daemon: invalid bind mount spec "C:\\hab/cache/keys:c:/hab/cache
/keys": invalid volume specification: 'C:\hab/cache/keys:c:/hab/cache/keys': invalid mount config for type "bind": bind
source path does not exist.
See 'C:\Program Files\Docker\docker.EXE run --help'.
```
After creating a c:\hab\cache\keys directory (c:\hab\cache existed, but not c:\hab\cache\keys), and re-running 'hab studio enter', all seems well. Should this directory not be created automatically if it does not exist?

mwrock:
This is great info! We should definitely validate that this directory exists and create it if it does not.

mwrock:
Ends up this is not specific to Windows. Linux container exhibits the same behavior.

ericcalabretta:
This issue still looks to be active. I ran into it today on 0.78.0. I received the same error even after I did hab setup. I declined to generate a new key during setup so maybe that had something to do with it. I could enter the studio after I manually created the keys directory.

PR #6345, opened 2019-03-26
Check for missing cache key path when entering dockerized studio

danielhertenstein:
Resolves https://github.com/habitat-sh/habitat/issues/5025

I ran into issue https://github.com/habitat-sh/habitat/issues/5025 when setting up Habitat on my Windows machine, so I figured I would submit a pull request along the lines of the solution suggested by @mwrock.

`Error::FileNotFound()` felt the closest to this error, but I am happy to switch it to something else or add a new enum element.

I know issue https://github.com/habitat-sh/habitat/issues/6314 also exists which intends to touch this code, but from what I can tell, a check would still be good here even if the way the path arrives in this method changes.

Lastly, I had started writing a test for this, but noticed the verification pipeline does not run any tests for studio and only checks to ensure it builds. It looks like I'd have to change the pipeline in order for any tests I write to be run, and I don't know what you all think about that.

Edit: I forgot to note that I needed to make the change to `build_component.ps1` in order to run the build test locally. The comment above the `pkg build` line made it seem like this change should have been made a little bit ago.

### DIFF
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -57,13 +57,11 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
         Err(_) => cache_key_path(None::<PathBuf>),
     };
     if !local_cache_key_path.exists() {
-        return Err(Error::FileNotFound(
-            format!(
-                "{}\nRun `hab setup` to create an origin or \
-                use `hab origin key` to configure your keys.",
-                local_cache_key_path.display()
-            )
-        ));
+        return Err(Error::FileNotFound(format!("{}\nRun `hab setup` to \
+                                                create an origin or use \
+                                                `hab origin key` to \
+                                                configure your keys.",
+                                               local_cache_key_path.display())));
     }
 
     let mut volumes = vec![format!("{}:{}{}",
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -61,7 +61,7 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
             format!(
                 "{}\nRun `hab setup` to create an origin or \
                 use `hab origin key` to configure your keys.",
-                local_cache_key_path.to_string_lossy().into_owned()
+                local_cache_key_path.display()
             )
         ));
     }
@@ -71,7 +71,7 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
                                    mnt_prefix,
                                    "/src"),
                            format!("{}:{}/{}",
-                                   local_cache_key_path.to_string_lossy(),
+                                   local_cache_key_path.display(),
                                    mnt_prefix,
                                    CACHE_KEY_PATH),];
     if let Ok(cache_artifact_path) = henv::var(ARTIFACT_PATH_ENVVAR) {
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -50,12 +50,24 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
     } else {
         ""
     };
+
+    let local_cache_key_path = default_cache_key_path(None);
+    if !local_cache_key_path.exists() {
+        return Err(Error::FileNotFound(
+            format!(
+                "{}\nRun `hab setup` to create an origin or \
+                use `hab origin key` to configure your keys.",
+                local_cache_key_path.to_string_lossy().into_owned()
+            )
+        ));
+    }
+
     let mut volumes = vec![format!("{}:{}{}",
                                    env::current_dir().unwrap().to_string_lossy(),
                                    mnt_prefix,
                                    "/src"),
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

baumanj:
Thanks for the PR!

I have a change in flight as part of https://github.com/habitat-sh/habitat/issues/6314 that will add a `cache_key_path: &Path` argument to [`start_docker_studio`](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/studio/docker.rs#L47). I believe that will supersede this change since we're trying to get away from using `default_cache_key_path` altogether. If you see any reason that's not the case, let's discuss, otherwise I apologize for any wasted effort on your part, I've only just recently started working on this.

baumanj:
After looking a bit more, I think we may want to do something like what you have here, but just at a different level. Let me get a PR up for what I'm doing and we can re-assess.

danielhertenstein:
I saw you've been tackling https://github.com/habitat-sh/habitat/issues/6314, and was also thinking this sort of check would wind up being moved to a higher level after this section of code was addressed, so that sounds good from my perspective.

baumanj:
I haven’t forgotten about this, it’s just gotten a bit delayed behind some other work. Will update when I’m able to finish up 6314

danielhertenstein:
No worries. I understand having a lot of work to do at the same time. I'm happy to look back into this whenever #6314 is done.

baumanj:
Just got back from a week out, so hopefully will have a meaningful update here by sometime *next* week.

danielhertenstein:
Sounds good. Welcome back, and good luck.

components/hab/src/command/studio/docker.rs
```diff
@@ -64,12 +64,24 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
     } else {
         ""
     };
+
+    let local_cache_key_path = default_cache_key_path(None);
```
baumanj: Since this is the only remaining caller of `default_cache_key_path`, and this is a pattern we'd like to eliminate, I think this is a good opportunity to move its implementation inline here.
```suggestion
    let local_cache_key_path = match henv::var(CACHE_KEY_PATH_ENV_VAR) {
        Ok(val) => PathBuf::from(val),
        Err(_) => cache_key_path(None::<PathBuf>),
    };
```

components/hab/src/command/studio/docker.rs
```diff
@@ -64,12 +64,24 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
...
+    if !local_cache_key_path.exists() {
+        return Err(Error::FileNotFound(
```
baumanj: Re-using this error variant is fine with me, but I think we should include *where* we looked for the path to exist. This error could be confusing in the event of an env var override.

components/hab/src/command/studio/docker.rs
```diff
@@ -64,12 +64,24 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
...
+            format!(
+                "{}\nRun `hab setup` to create an origin or \
+                use `hab origin key` to configure your keys.",
+                local_cache_key_path.to_string_lossy().into_owned()
```
baumanj: Both here and on L84, I think [`PathBuf::display`](http://doc.rust-lang.org/1.34.1/std/path/struct.PathBuf.html#method.display) would be preferable to `to_string_lossy`. Also, think we can omit `into_owned`.

support/ci/build_component.ps1
```diff
@@ -17,8 +17,7 @@ $env:HAB_CACHE_KEY_PATH="$job_temp_root/keys"
 Write-Host "--- :key: Generating fake origin key"
 hab origin key generate
 Write-Host "--- :hab: Running hab pkg build for $Component"
-# REMOVE -R AFTER 0.75.0
-hab pkg build -D -R components/$Component
+hab pkg build -D components/$Component
```
baumanj: It looks like this happened in the interim, so we can leave this change out after rebasing.

danielhertenstein:
@baumanj Thanks for looking back into this. I'll have plenty of time to fix this up next week, if that's an okay wait for you.

baumanj:
Hey @danielhertenstein. Is this still something you're interested in continuing with?

components/hab/src/command/studio/docker.rs
danielhertenstein: Isn't `local_cache_key_path` where we looked?

baumanj: Fair point. I'm not sure if I just missed that or was thinking about something else. This looks good to me.

baumanj:
No worries; just wanted to check.

danielhertenstein:
@baumanj, I've committed the requested changes, though I'm not sure how to move the pull request forward via the interface here. Do I hit "resolve conversation" on everything?

danielhertenstein:
@baumanj, neither of those links worked for me. The first one required I log into Okta. The second one 404'd. But the formatting is done :)

baumanj:
Thanks so much for the contribution, @danielhertenstein !

danielhertenstein:
@baumanj, I was happy to do so. Thanks for helping me through it!

PR #6347, opened 2019-03-26
Work in progress: studio CLI parsing in rust

baumanj:
See https://github.com/habitat-sh/habitat/issues/6314

To fully eliminate `default_cache_key_path`, we need to have proper CLI arguments for each command that needs a key cache path. That allows the env var override to be checked appropriately.

Unfortunately, the studio (in the docker case) needs this value, but the CLI parsing for the `hab studio …` commands is mostly deferred to `hab-studio.{sh,ps1}`, meaning either we need to add the validation in the scripts, do it ad-hoc in rust (not clap) in `studio::enter::start`, or pull the CLI parsing into our standard clap-based approach. The last option seems the best, so this is a start down that road.

Still to do are implementing the rest of the `hab studio …` subcommands in clap and possibly further refactoring of `studio::enter::start` to completely remove the `args: &[OsString]` argument. Finally, it likely makes sense to completely remove the argument parsing code from the `hab-studio` scripts once this is in place, but that approach needs more thought first.

Refactoring out the rest of the ad-hoc arg parsing in `exec_subcommand_if_called` is probably a good idea, but that can be in a follow-up PR.

For comparison, see how `hab studio --help` renders with the [new, clap-based approach](https://github.com/habitat-sh/habitat/files/3010744/clap.txt) versus [the old shell-based one](https://github.com/habitat-sh/habitat/files/3010749/clap.txt).

### DIFF
--- a/components/common/src/cli.rs
+++ b/components/common/src/cli.rs
@@ -30,9 +30,6 @@ lazy_static! {
     pub static ref GOSSIP_DEFAULT_ADDR: String =
         { format!("{}:{}", GOSSIP_DEFAULT_IP, GOSSIP_DEFAULT_PORT) };
 }
-pub const GOSSIP_LISTEN_ADDRESS_ENVVAR: &str = "HAB_LISTEN_GOSSIP";
-pub const RING_ENVVAR: &str = "HAB_RING";
-pub const RING_KEY_ENVVAR: &str = "HAB_RING_KEY";
 
 pub const LISTEN_HTTP_DEFAULT_PORT: u16 = 9631;
 pub const LISTEN_HTTP_DEFAULT_IP: &str = "0.0.0.0";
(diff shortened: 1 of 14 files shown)


thesentinels:

Thank you for contributing!

components/common/src/cli.rs
```diff
@@ -71,7 +67,36 @@ lazy_static! {
...
+    pub const HTTPS_PROXY: &str = "https_proxy";
+    pub const HTTP_PROXY: &str = "http_proxy";
+    pub const NO_ARTIFACT_PATH: &str = "NO_ARTIFACT_PATH";
+    pub const NO_PROXY: &str = "no_proxy";
+    pub const NO_SRC_PATH: &str = "NO_SRC_PATH";
+    pub const QUIET: &str = "QUIET";
+    pub const STUDIO_TYPE: &str = "STUDIO_TYPE";
+    pub const VERBOSE: &str = "VERBOSE";
+    // End to deprecate
+}
```
stevendanna: Thanks for this! I think changes like this are really valuable, while grepping the code base is obviously possible, this makes these variable much more discoverable.

christophermaier:
Going to close this for now, until we find more time :cry:

PR #6338, opened 2019-03-22
Don't use TESTING_FS_ROOT as an environment variable.

scotthain:
Rather than run all the tests in a seperate FS_ROOT, which can cause pathing and other issues,
run tests normally, except for the one test that needs to lay down a template file and verify
it. This seems like a reasonable use of TESTING_FS_ROOT.

Signed-off-by: Scott Hain <shain@chef.io>

PR #6349, opened 2019-03-27
Migrate to habitat rust rather than system rust

scotthain:

This is step 1 and will replace https://github.com/habitat-sh/habitat/pull/6338

After digging through with @baumanj - it looks like there may be a feature gap that was causing the issues I was seeing when using `TESTING_FS_ROOT` that caused the rabbithole in 6338.

Note - we will still be using the system rust until I remove it from the base build image, which will come shortly.


thesentinels:

Thank you for contributing!

PR #6350, opened 2019-03-28
Add package deletion command

chefsalim:
This PR adds a new command to remove packages from Builder to the hab cli. This is the corresponding change to https://github.com/habitat-sh/builder/pull/988

tenor-177628962

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -931,6 +931,27 @@ impl Client {
         }
     }
 
+    /// Delete a package from Builder
+    ///
+    /// # Failures
+    ///
+    /// * Remote Builder is not available
+    /// * If package does not exist in Builder
+    /// * If the package does not qualify for deletion
+    /// * Authorization token was not set on client
+    pub fn delete_package(&self, ident: &PackageIdent, token: &str) -> Result<()> {
+        let path = package_path(ident);
+        debug!("Deleting package {}", ident);
+
+        let res = self.add_authz(self.0.delete(&path), token).send()?;
+
+        if res.status != StatusCode::NoContent {
+            return Err(err_from_response(res));
+        };
+
+        Ok(())
+    }
+
     /// Promote a package to a given channel
     ///
     /// # Failures
--- a/components/hab/src/command/pkg/delete.rs
+++ b/components/hab/src/command/pkg/delete.rs
@@ -1,0 +1,43 @@
+//! Delete a package from Builder.
+//!
+//! # Examples
+//!
+//! ```bash
+//! $ hab pkg delete acme/redis/2.0.7/2112010203120101
+//! ```
+//! //! This will delete the acme package specified from Builder
+//! //! if certain conditions apply - for example, the package is not
+//! //! in the stable channel, and does not have any other packages that
+//! //! depend on it.
+//!
+//! //! Note: This command does not remove the package from disk
+
+use crate::{api_client::Client,
+            common::ui::{Status,
+                         UIWriter,
+                         UI},
+            error::{Error,
+                    Result},
+            hcore::package::PackageIdent,
+            PRODUCT,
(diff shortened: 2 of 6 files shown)


thesentinels:

Thank you for contributing!

components/builder-api-client/src/lib.rs
```diff
@@ -931,6 +931,30 @@ impl Client {
...
+    /// # Failures
+    ///
+    /// * Remote Builder is not available
+    /// * If package does not exist in Builder
+    /// * If the package does not qualify for deletion
+    /// * Authorization token was not set on client
+    pub fn delete_package(&self, ident: &PackageIdent, token: &str) -> Result<()> {
+        if !ident.fully_qualified() {
+            return Err(Error::IdentNotFullyQualified);
+        }
```
baumanj: This check seems superfluous. You can't get to this point unless the package is fully-qualified due to [the `valid_fully_qualified_ident` validator specified in the clap config](https://github.com/habitat-sh/habitat/pull/6350/files#diff-ee10e75b553edc813cf07f6a5036374cR536), and even if it did, builder would reject it.

components/hab/src/command/pkg/delete.rs
```diff
@@ -0,0 +1,47 @@
...
+//! //! depend on it.
+//!
+//! //! Note: This command does not remove the package from disk
+
+use crate::{api_client::Client,
+            common::ui::{Status,
+                         UIWriter,
+                         UI},
+            hcore::package::PackageIdent};
+
```
baumanj: We're trying to standardize on one block for all `use` statements with blank lines so `rustfmt` sorts them for us.

components/hab/src/command/pkg/delete.rs
```diff
@@ -0,0 +1,47 @@
+//! Delete a package from Builder.
+//!
+//! # Examples
+//!
+//! ```bash
+//! $ hab pkg delete acme/redis/2.0.7/2112010203120101 stable
```
baumanj: Should this be
```suggestion
//! $ hab pkg delete acme/redis/2.0.7/2112010203120101
```
?

The current command doesn't pass clap parsing:
```
error: Found argument 'stable' which wasn't expected, or isn't valid in this context

USAGE:
    hab pkg delete [OPTIONS] <PKG_IDENT>
```
Since we only are allowing deletion of packages that are not on the stable channel, it seems like we shouldn't be taking a channel arg at all.

components/hab/src/command/pkg/delete.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    ui.begin(format!("Deleting {} from Builder", ident))?;
+
+    match api_client.delete_package(ident, token) {
+        Ok(_) => (),
+        Err(e) => {
+            println!("Failed to delete '{}': {:?}", ident, e);
+            return Err(Error::from(e));
+        }
+    }
```
baumanj: Why not
```rust
    if let Err(e) = api_client.delete_package(ident, token) {
        println!("Failed to delete '{}': {:?}", ident, e);
        return Err(Error::from(e));
    }
```
?

components/builder-api-client/src/lib.rs
chefsalim: Good pt - removing

components/hab/src/command/pkg/delete.rs
chefsalim: OK

components/hab/src/command/pkg/delete.rs
chefsalim: Yep thats a typo - will fix

components/hab/src/command/pkg/delete.rs
chefsalim: Surprised clippy didn't call that out... will update

PR #6352, opened 2019-03-28
Migrate OSX build system

scotthain:
See commit comments for specific changes. This migrates from our long lived OSX instance to ephemeral OSX images.


.buildkite/scripts/build_mac_release.sh
```diff
@@ -18,16 +21,13 @@ sudo installer \
...
+export PATH="/opt/mac-bootstrapper/embedded/bin:$PATH"
 
-# TODO (CM): consider getting Bash updated on builders (likely via Homebrew)
-# TODO (CM): have a cleanup function
-#     Clear out /hab
-#     Uninstall Homebrew stuff
-#     Uninstall Rustup, Rust?
-
-echo "--- :beer: Updating Homebrew dependencies"
-brew bundle install --verbose --file=.buildkite/Brewfile
```
smacfarlane: Are there any remaining reference to the Brewfile?  Should we remove that as part of this PR? Does the update here contain the entirety of what the Brewfile was installing (wget)?

scotthain: Good question! I thought about that - I was going to ask @christophermaier when he's back.

christophermaier: @scotthain Sorry for not catching this comment earlier... yeah, if we're no longer using the Brewfile for this, we can get rid of it. I added it solely to make this work on our old macOS builders.

scotthain:
@christophermaier @tduffield fixed!

baumanj:
Any update here, @scotthain ?

scotthain:
This POC Proved out the point, we'll revisit this once we have the expeditor enabled executors setup

PR #6353, opened 2019-03-28
start supervisor outside of interactive powershell session in local windows studio

### LINKED ISSUE #4945: Properly stop supervisor when exiting a local windows studio
Currently when exiting a local windows studio, we simply kill the supervisor process which may not actually terminate its child processes. Now that we have better IPC behavior, it may be simple to cleanly terminate the supervisor and have it send a signal to all services to stop.

Note, We should also whack the studio's `\hab\sup\default`

mwrock:
This fixes #4945.

We currently enter an interactive Powershell session when entering a windows studio and we start a supervisor inside of that session. This presents a problem when the user types `exit` and closes the shell. When this happens, the `hab`, `hab-sup` and `hab-launch` processes are immediately terminated and the launcher/supervisor do not have the opportunity to properly stop and clean up services. The running `run` hooks are orphaned along with any processes that they spawned.

This PR makes 2 fundamental changes:

1. It moves the `hab sup run` just before entering the interactive shell. This means that the hab processes are still running after `exit` and we can then perform our cleanup.
2. To make this work smoothly, we move the supervisor startup and logging into C# instead of Powershell since it will respond to output from the supervisor process in a separate thread. Powershell's event subscription mechanism would block on the interactive shell and there would be no logged supervisor output until the studio exits. And that would be cray cray.


components/studio/bin/hab-studio.ps1
```diff
@@ -298,8 +298,96 @@ function Enter-Studio {
...
+      # Set console encoding to UTF-8 so that any redirected glyphs
+      # from the supervisor log are propperly encoded
+      [System.Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+
+      # We start the Supervisor and handle its output logging in C# which will handle
+      # the process's OutputDataReceived and ErrorDataReceived events in a separate thread.
+      # While we could use PowerShell's Register-ObjectEvent instead. That uses a PSJob
+      # which will be blocked while the interactive studio is open
+      $c = @"
+      using System;
```
baumanj: I'm generally opposed to embedding other languages inside shell strings like this unless it's absolutely necessary. Could this be a standalone C# file? I don't see any interpolation, but I'm pretty ignorant with respect to pwsh.

mwrock: fair point. Moved to its own file @baumanj

PR #6359, opened 2019-04-01
Add a blog post for the 0.79.0 release announcement

baumanj:


dmccown:
Do we have a good section to call out the Handlebars change that will be coming down the road?

baumanj:
> Do we have a good section to call out the Handlebars change that will be coming down the road?

Not really, but it's easy enough to add. However, I'd say it's premature to mention at least until https://github.com/habitat-sh/habitat/pull/6360 is merged.

dmccown:
I should have checked if that was merged yet! Carry on :)

Issue #6323, opened 2019-03-20
Potentially breaking upcoming change to template processing: PLEASE READ


mpeck:
TLDR;
The library being used for handlebars supports a nonstandard syntax for accessing array values. This syntax will be going away soon. If you are using the "bad" syntax, `{{ foo[bar] }}`, change it to the "good" syntax `{{ foo.[bar] }}` and let us know that you were using the "bad" syntax.

Back in version 0.52 a bug was discovered when accessing array items in handlebars templates, [4615](https://github.com/habitat-sh/habitat/issues/4615). Basically accessing an item in an array using syntax like, `{{ foo[bar]  }}` resulted in an error. The bug was tracked back to a version bump in the library we use to process handlebars, [handlebars-rust](https://github.com/sunng87/handlebars-rust). The solution was to just revert the version bump and pin the dependency to the previous version. That was a fine solution at the time, but upon further research we found that the `{{ foo[bar] }}` was actually nonstandard syntax for handlebars. The proper syntax should be `{{ foo.[bar] }}`. The bug discovered was due to a bugfix in the handlebars library. We would like to upgrade that library to a more recent version. Unfortunately the library does not return a good error in this case, [246](https://github.com/sunng87/handlebars-rust/issues/246); this makes it difficult for us to provide a deprecation notice or proper error messaging. We will be upgrading the library to a newer version in the future, but we would like to prevent as much turbulence as possible.

So what does this all mean to me you might be asking? Well fortunately the current version of the library we're using supports both syntaxes. So if you are already using the "good" syntax then you have nothing further to do. If, however, you are using the "bad" syntax then we ask you do two things: 1) Change to the good syntax; 2) Let us know that you were using the "bad" syntax. We want to give folks time to migrate, but we would also like to get a feeling for how many people are using the "bad" syntax in order to help us determine when to actually make the change.

If you are using the "bad" syntax just leave a comment on this issue.

PR #6360, opened 2019-04-01
Add deprecation message to templates using the foo[bar] syntax

mpeck:
This just adds a simple deprecation message to templates that are using the bad array access syntax. See https://github.com/habitat-sh/habitat/issues/6323 for more info.

### DIFF
--- a/components/common/src/templating/mod.rs
+++ b/components/common/src/templating/mod.rs
@@ -117,23 +117,34 @@ impl TemplateRenderer {
                                              TemplateFileError::IOError(e, name.to_owned())
                                          })?;
 
-        template_string.split("\n")
+        fn transform(text: &str) -> String {
+            let text = text.to_owned();
+            if RE.is_match(&text) {
+                transform(&RE.replace_all(&text, "$1.$2"))
+            } else {
+                text
+            }
+        }
+
+        template_string.lines()
             .enumerate()
-            .filter(|(_i, v)| RE.is_match(&v))
-            .for_each(|(i, v)| {
+            .filter(|(_i, line)| RE.is_match(&line))
+            .map(|(i, line)| (i, line, transform(&line)))
+            .for_each(|(i, old_line, new_line)| {
                 println!("\n\n***************************************************\n\
-                          Deprecated object access syntax in handlebars template\n\n\
-                          TEMPLATE: {}\n\
-                          LINE: {}: '{}'\n\n\
+                          Deprecated object access syntax in handlebars template\n\
                           Use 'object.[index]' syntax instead of 'object[index]'\n\
-                          See https://github.com/habitat-sh/habitat/issues/6323 for more information\n\
+                          See https://github.com/habitat-sh/habitat/issues/6323 for more information\n\n\
+                          TEMPLATE: {}\n\
+                          LINE: {}: '{}'\n\
+                          Update to '{}'\n\n\
                           *******************************************************\n\n",
-                         path.display(), i, v)
+                         path.display(), i + 1, old_line, new_line)
             });
 
         // Replace the deprecated syntax with the good syntax. This should make it easier to upgrade
         // the handlebars crate without breaking folks.
-        let updated_template_string = RE.replace_all(&template_string, "$1.$2");
+        let updated_template_string = transform(&template_string);
         self.0
             .register_template_string(name, updated_template_string)?;
         Ok(())
--- a/components/common/src/templating/mod.rs
+++ b/components/common/src/templating/mod.rs
@@ -20,12 +20,16 @@ pub mod package;
 pub mod test_helpers;
 
 use std::{fmt,
+          fs::read_to_string,
           ops::{Deref,
                 DerefMut},
           result};
 
+use regex::Regex;
+
 use handlebars::{Handlebars,
-                 RenderError};
+                 RenderError,
+                 TemplateFileError};
 use serde::Serialize;
 use serde_json;
 
@@ -90,6 +94,42 @@ impl TemplateRenderer {
         self.0
             .render(template, &raw)
             .map_err(|e| Error::TemplateRenderError(format!("{}", e)))
+    }
+
+    // This method is only implemented so we can intercept the call to Handlebars and display
+    // a deprecation message to users. More information here https://github.com/habitat-sh/habitat/issues/6323.
+    // When Handlebars is upgraded this can be safely removed.
+    pub fn register_template_file<P>(&mut self,
+                                     name: &str,
+                                     path: P)
+                                     -> result::Result<(), TemplateFileError>
+        where P: AsRef<std::path::Path>
+    {
+        let path = path.as_ref();
+
+        lazy_static! {
+            static ref RE: Regex =
+                Regex::new(r"\{\{[^}]+[^.]\[[^}]*\}\}").expect("Failed to compile template \
+                                                                deprecation regex");
+        }
+
+        if let Ok(template_string) = read_to_string(path) {
+            template_string.split("\n")
+                           .enumerate()
+                           .filter(|(_i, v)| RE.is_match(&v))
+                           .for_each(|(i, v)| {
+                               println!("\n\n***************************************************\n\
+                                         Deprecated object access syntax in handlebars template\n\n\
+                                         TEMPLATE: {}\n\
+                                         LINE: {}: '{}'\n\n\
+                                         Use 'object.[index]' syntax instead of 'object[index]'\n\
+                                         See https://github.com/habitat-sh/habitat/issues/6323 for more information\n\
+                                         *******************************************************\n\n",
+                                        path.display(), i, v)
+                           });
+        }
+
+        self.0.register_template_file(name, path)
     }
 }
 
--- a/components/common/src/templating/mod.rs
+++ b/components/common/src/templating/mod.rs
@@ -20,7 +20,6 @@ pub mod package;
 pub mod test_helpers;
 
 use std::{fmt,
-          fs::read_to_string,
           ops::{Deref,
                 DerefMut},
           result};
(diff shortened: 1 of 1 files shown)


components/common/src/templating/mod.rs
```diff
@@ -91,6 +95,35 @@ impl TemplateRenderer {
             .render(template, &raw)
             .map_err(|e| Error::TemplateRenderError(format!("{}", e)))
     }
+
+    // This method is only implemented so we can intercept the call to Handlebars and display
+    // a deprecation message to users. More information here https://github.com/habitat-sh/habitat/issues/6323.
+    // When Handlebars is upgraded this can be safely removed.
+    pub fn register_template_file(&mut self,
```
baumanj: In general, I'm a fan of the simpler approach like you employed with the `path` argument (`&Path` rather than `P: AsRef<Path>`). But in this case, since we're essentially overriding [a method of the `Handlebars` struct we're wrapping](https://docs.rs/handlebars/0.29.1/handlebars/struct.Handlebars.html#method.register_template_file), I think it makes sense to keep the signature the same. Especially since the change as is breaks existing tests which expect the other signature.

components/common/src/templating/mod.rs
```diff
@@ -91,6 +95,35 @@ impl TemplateRenderer {
...
+            if re.is_match(&template_string) {
+                println!("\n\n********** Deprecation Notice!!! **********\n\nThe template at \
+                          {:?} is using non-standard handlebars syntax for array access. \
+                          Discussed here https://github.com/habitat-sh/habitat/issues/6323. \
+                          Support for this syntax will be removed in an upcoming \
+                          release.\n\nNon-standard syntax found in the following statements:\n",
+                         path);
+
+                for mat in re.find_iter(&template_string) {
+                    println!("{}", mat.as_str());
```
baumanj: I would be more user-friendly to provide a line/column location of the deprecated syntax as well as a suggestion for how to correct it. The former might require splitting `template_string` into lines first since [the `Match` struct](https://docs.rs/regex/1.1.5/regex/struct.Match.html) only gives byte offsets.

baumanj: We *might* even be able to do the replacement with good syntax on the fly, allowing us to do the upgrade to the new version of handlebars now.

mpeck: I like the idea of updating the syntax on the fly; that feels like something we should do when we actually do the upgrade though.

baumanj: Sure, there's really no point in doing so *until* we upgrade our handlebars library. What I'm saying is that instead of:
1. Display a deprecation warning for some number of releases, hoping that people fix their templates 
1. Update to the new handlebars library and break any un-fixed templates

we can:
1. Display a deprecation warning for some number of releases, hoping that people fix their templates 
1. Update to the new handlebars library, fixing instances of the old syntax on the fly (with a warning)
1. Optionally at some point in the future remove the fixup code altogether and break any un-fixed templates

Previously we thought we were *unable* to update our very old version of handlebars because we didn't want to break people's templates that use the old syntax. The point is, given how simply we could fix up the old syntax on the fly, I don't think we need to wait on updating the handlebars library.

I do think starting with just a deprecation warning is a good idea, since there could be bugs with our on-the-fly fixes, but I also think there's a lot of value in updating to the new handlebars library version sooner rather than later.

mpeck: Yup, I think we are agreeing with each other.

components/common/src/templating/mod.rs
```diff
@@ -91,6 +94,50 @@ impl TemplateRenderer {
...
+                          LINE: {}: '{}'\n\n\
+                          Use 'object.[index]' syntax instead of 'object[index]'\n\
+                          See https://github.com/habitat-sh/habitat/issues/6323 for more information\n\
+                          *******************************************************\n\n",
+                         path.display(), i, v)
+            });
+
+        // Replace the deprecated syntax with the good syntax. This should make it easier to upgrade
+        // the handlebars crate without breaking folks.
+        let updated_template_string = RE.replace_all(&template_string, "$1.$2");
```
baumanj: If we change things on the fly, we should output the changes we're making as well as provide a way to disable this behavior in case our changes are wrong.

The current code won't handle something like `{{ foo[bar][baz] }}`. I think we may be better off not using regular expressions and preferring a more direct approach.

components/common/src/templating/mod.rs
```diff
@@ -91,6 +94,50 @@ impl TemplateRenderer {
...
+        }
+
+        let template_string =
+            std::fs::read_to_string(path).map_err(|e| {
+                                             TemplateFileError::IOError(e, name.to_owned())
+                                         })?;
+
+        template_string.split("\n")
+            .enumerate()
+            .filter(|(_i, v)| RE.is_match(&v))
```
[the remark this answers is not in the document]
mpeck: I'm using `enumerate` to get the line number. Moving it after the filter would give me the index of the match within the list of matches, which isn't as useful.

components/common/src/templating/mod.rs
```diff
@@ -91,6 +94,50 @@ impl TemplateRenderer {
...
+                Regex::new(r"(\{\{[^}]+[^.])(\[[^}]*\}\})").expect("Failed to compile template \
+                                                                    deprecation regex");
+        }
+
+        let template_string =
+            std::fs::read_to_string(path).map_err(|e| {
+                                             TemplateFileError::IOError(e, name.to_owned())
+                                         })?;
+
+        template_string.split("\n")
```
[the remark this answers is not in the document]
mpeck: Thanks, I didn't know about `lines()`

components/common/src/templating/mod.rs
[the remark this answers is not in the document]
baumanj: Good point! My mistake.

components/common/src/templating/mod.rs
mpeck: I'm not sure what you mean by "a more direct approach" but this does make me lean towards us not updating the templates on the fly. It seems like a lot of trouble for something that will shortly be deleted.

mpeck: It also adds another vector for bugs to creep in.

components/common/src/templating/mod.rs
```diff
@@ -91,6 +94,61 @@ impl TemplateRenderer {
...
+                          TEMPLATE: {}\n\
+                          LINE: {}: '{}'\n\
+                          Update to '{}'\n\n\
+                          *******************************************************\n\n",
+                         path.display(), i + 1, old_line, new_line)
+            });
+
+        // Replace the deprecated syntax with the good syntax. This should make it easier to upgrade
+        // the handlebars crate without breaking folks.
+        let updated_template_string = transform(&template_string);
```
mpeck: After our conversation I'm a bit reluctant to update the template string on the fly. But here it is now, for further discussion.

components/common/src/templating/mod.rs
baumanj: > I'm not sure what you mean by "a more direct approach"

Using methods on `str`/`String` to find `[` not preceded by `.` in a handlebars context and replace them with `.[`. It's just a suggestion. If you think regular expressions would work better (or a hybrid approach), you can pursue that. I'm just pointing out an instance that the current code wouldn't handle. I typically find it easier to follow code that doesn't use regular expressions, so that's why I suggested it.

As for not updating the templates on the fly, I could go either way. The benefit is that we can delay (possibly indefinitely) breaking old templates without delaying updating our `handlebars` crate. But you're right that it does add some complexity. I don't know if we have a commitment to the timeline for when we'd flip the switch from deprecation to unsupported, but if it has significant impact on users, I think it's worth a little complexity.

components/common/src/templating/mod.rs
baumanj: Nice! This seems to work, but I'd suggest putting in some tests. Looking through `core-plans` may be a good source of test case ideas. You'll likely want to pull `transform` (and `RE`) out of `register_template_file` so they can be tested.

components/common/src/templating/mod.rs
```diff
@@ -91,6 +94,50 @@ impl TemplateRenderer {
...
+                                         })?;
+
+        template_string.split("\n")
+            .enumerate()
+            .filter(|(_i, v)| RE.is_match(&v))
+            .for_each(|(i, v)| {
+                println!("\n\n***************************************************\n\
+                          Deprecated object access syntax in handlebars template\n\n\
+                          TEMPLATE: {}\n\
+                          LINE: {}: '{}'\n\n\
```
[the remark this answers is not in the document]
mpeck: I'm not showing the column because that would involve another loop, but I'm showing the complete corrected line. I think that's enough info for the user to fix the problem.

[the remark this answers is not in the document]
baumanj: Leaving out the column is fine. The important bit is putting the line number directly after the path, separated by a colon. For example, rustc gives warnings like:
```
warning: unused import: `validation`
  --> components/hab/src/cli.rs:23:28
```
This syntax allows editors to jump to the right line in the right file in a way that 
```
TEMPLATE: components/hab/src/cli.rs
LINE: 23
```
wouldn't. Does that make sense?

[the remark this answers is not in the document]
mpeck: Ah, I see. I can make that change.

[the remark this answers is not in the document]
mpeck:
```
***************************************************
warning: Deprecated object access syntax in handlebars template
Use 'object.[index]' syntax instead of 'object[index]'
See https://github.com/habitat-sh/habitat/issues/6323 for more information

--> /hab/pkgs/core/kibana/6.1.0/20190402203539/config/kibana.yml:12

change 'elasticsearch.url: "http://{{bind.elasticsearch.members[0].sys.ip}}:{{bind.elasticsearch.members[0].cfg.http-port}}"'
to     'elasticsearch.url: "http://{{bind.elasticsearch.members.[0].sys.ip}}:{{bind.elasticsearch.members.[0].cfg.http-port}}"'

*******************************************************
```
How does this look to you?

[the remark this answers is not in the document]
baumanj: 💯

PR #6367, opened 2019-04-02
Consolidate feature flag operations.

christophermaier:
Previously, feature flags were defined in both the `hab` and
`habitat_sup` crates. This was problematic because it required far too
much setup in order to create a feature flag.

Additionally, if you wanted to create a feature flag that is mainly
used in the Supervisor, but conditionally adds additional CLI options,
you were out of luck. Because of underlying code, as well as the fact
`hab` and `hab-sup` are entirely separate binaries, the feature flags
enabled in `hab` wouldn't necessarily be enabled in `hab-sup`.

This refactoring moves *all* feature flags into the common
crate. Additionally, we dispense with the use of the `features` crate,
in favor of using the underlying `bitflags` abstraction directly.

This means that we no longer rely on recording which feature flags are
currently enabled in global state, but in immutable data that we
explicitly thread through function calls.

This also means that functionality that differs based on feature flag
state can be more easily unit tested (otherwise, we have to contend
with the sharing of global state, which is always tricky).

### DIFF
--- a/components/common/src/lib.rs
+++ b/components/common/src/lib.rs
@@ -12,8 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::ui::{UIWriter,
+                UI};
 use habitat_api_client as api_client;
 use habitat_core as hcore;
+use lazy_static::lazy_static;
+use std::{collections::HashMap,
+          env,
+          iter::FromIterator};
+
 extern crate json;
 #[macro_use]
 extern crate log;
@@ -43,7 +50,7 @@ pub mod util;
 
 lazy_static::lazy_static! {
     pub static ref PROGRAM_NAME: String = {
-        match std::env::current_exe() {
+        match env::current_exe() {
             Ok(path) => path.file_stem().and_then(|p| p.to_str()).unwrap().to_string(),
             Err(e) => {
                 error!("Error getting path of current_exe: {}", e);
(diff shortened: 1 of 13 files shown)


christophermaier:
One downside of this approach is that, when you're dealing with the Supervisor, you'll see output for which flags are enabled _twice_; once for the `hab` binary, and once for the `hab-sup` binary. 😦

components/common/src/feature_flags.rs
```diff
@@ -0,0 +1,97 @@
...
+                           (feat::EventStream, "HAB_FEAT_EVENT_STREAM")];
+        HashMap::from_iter(mapping)
+    };
+}
+
+/// If the environment variable for a flag is set to _anything_ but
+/// the empty string, it is activated.
+pub fn enable_features_from_env(ui: &mut UI) {
+    for (feature, env_var) in ENV_VARS.iter() {
+        if henv::var(env_var).is_ok() {
```
baumanj: Since I think [we should move away from this interface entirely](https://github.com/habitat-sh/habitat/issues/6314), what would you say to handling this with `std::env`? Something like:
```rust
    match env::var_os("FOO") {
        Some(ref val) if val.is_empty() => println!("FOO is empty"),
        Some(_) => println!("FOO is on"),
        None => println!("FOO is off"),
    }
```

And given the trouble we've run into lately with empty env vars, I'd fully support changing to a system where for boolean feature flags any non-empty value is considered "on", not setting the var at all is "off" and an empty value is an error. What would you say to that?

components/common/src/feature_flags.rs
```diff
@@ -0,0 +1,97 @@
...
+            ui.warn(&format!("Enabling feature: {:?}", feature))
+              .unwrap();
+        }
+    }
+
+    // TODO (CM): Once the other TODOs above are done (especially the
+    // documentation bits), it would be nice to extract this logic
+    // into an actual discoverable CLI subcommand; it's a little weird
+    // that you have to know how to enable a feature flag before you
+    // can even find out that there *are* feature flags to enable.
```
baumanj: This is really a question about what the intended use of feature flags is. I've been treating them as guards around pre-release code that isn't documented and may be totally broken. If that's what they are, I think they're non-discoverable *by design*. Once a feature is documented and of standard quality, I don't see why it wouldn't just be a regular CLI argument (with optional `env`-based default) and therefore totally discoverable.

christophermaier: That's kind of what I'm getting at... `HAB_FEAT_LIST` isn't "broken pre-release code"; it's clearly intended as a way to show what features are currently in-flight. It seems like a useful bit of functionality, but it _itself_ isn't a "feature" like other things are.

The few "test" features that you added are also useful, but also aren't quite "features" as we're discussing here, either.

(Also, just as an aside, just because something is currently behind a feature flag doesn't necessarily mean that once it's fully baked it will become anything that's exposed on the CLI; it might just be a new way of doing something internally)

components/common/src/feature_flags.rs
christophermaier: Yeah, I can do that.

components/common/src/feature_flags.rs
baumanj: After thinking about it more, I think the test features I've added *shouldn't* be part of `HAB_FEAT_LIST`, they should be proper CLI options, but [hidden from the clap help](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.hidden), because they're not user-facing.

Maybe it's acceptable to have feature flag code be potentially broken and untested, but once we have confidence in the long-term utility, they should be moved into the proper CLI (hidden or not) and integrated into our interfaces in a properly testable way.

christophermaier: I guess I don't think `HAB_FEAT_LIST` should really be a thing at all. It's a little weird that that output could show up, say, as you start up `hab sup run`. It feels like the only time you'd want to enable that feature is if you were doing something like `HAB_FEAT_LIST=1 hab --help`... which implies something like `hab features` would be the clearer thing. That seems more in line with the intention of the existing code.

I think there are probably other wrinkles to this, centering around the fact that `hab` and `hab-sup` are two separate binaries, and I'm fine with punting this for later (which is why I just mentioned this in a comment).

christophermaier: Having the test features be hidden CLI options is nice, though. Since the `features` crate also lets you create multiple sets of flags, we could also move them out into a separate set, apart from the "real" feature flags.

I think either way is fine.

components/common/src/lib.rs
```diff
@@ -52,3 +59,99 @@ lazy_static::lazy_static! {
...
+// TODO (CM): It'd be great to have a built-in way to document them,
+// too.
+
+// TODO (CM): Part of that documentation might be *when* a flag was
+// added. In general, long-lived flags are a code-smell.
+
+// TODO (CM): It may also be useful to break out features by area of
+// concern. We can have any number of bitflags-generated structs.
+
+bitflags::bitflags! {
```
baumanj: Not to go into bike shedding territory, but are we gaining anything by using bit flags? Could we just use a plain map and not have to worry about coming up with a unique bit so specify each feature flag?

components/hab/src/main.rs
```diff
@@ -122,15 +122,15 @@ lazy_static! {
 fn main() {
     env_logger::init();
     let mut ui = UI::default_with_env();
-    enable_features_from_env(&mut ui);
+    let flags = FeatureFlag::from_env(&mut ui);
     thread::spawn(analytics::instrument_subcommand);
-    if let Err(e) = start(&mut ui) {
+    if let Err(e) = start(&mut ui, flags) {
```
baumanj: 😚👌

components/common/src/lib.rs
christophermaier: I haven't really dug into getting off `bitflags`. However, given all the `TODO`s I dropped in here, I think there's plenty of room for future refactoring; with that, we very well may leave `bitflags` behind, as well.

components/sup/src/manager/mod.rs
```diff
@@ -820,7 +825,7 @@ impl Manager {
             if self.check_for_departure() {
                 break ShutdownMode::Departed;
             }
-            if !feat::is_enabled(feat::IgnoreSignals) {
+            if self.feature_flags.contains(FeatureFlag::IGNORE_SIGNALS) {
```
christophermaier: Oops, this was an accident; will fix.

PR #6368, opened 2019-04-02
Add target param to pkg related APIs

chefsalim:
This adds a package target param to hab cli commands that were missing them : promote, demote, delete and channels.  If the param is not specified, the target defaults to ~~x86_64-linux~~ a sane system appropriate default.

NOTE: This change will require a corresponding Builder backend change. Until the Builder change is rolled out, this change will be a no-op.

tenor-11376684

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -24,6 +24,7 @@ use habitat_common::{cli::{BINLINK_DIR_ENVVAR,
                            LISTEN_CTL_DEFAULT_ADDR_STRING,
                            LISTEN_HTTP_ADDRESS_ENVVAR,
                            LISTEN_HTTP_DEFAULT_ADDR,
+                           PACKAGE_TARGET_ENVVAR,
                            RING_ENVVAR,
                            RING_KEY_ENVVAR},
                      types::ListenCtlAddr};
@@ -122,8 +123,7 @@ pub fn get() -> App<'static, 'static> {
                     (aliases: &["s", "st", "sta", "star"])
                     (@arg PKG_IDENT: +required +takes_value {valid_ident}
                         "The origin and name of the package to schedule a job for (eg: core/redis)")
-                    (@arg PKG_TARGET: +takes_value {valid_target}
-                        "A package target (ex: x86_64-windows) (default: x86_64-linux)")
+                    (arg: arg_target())
                     (@arg BLDR_URL: -u --url +takes_value {valid_url}
                         "Specify an alternate Builder endpoint. If not specified, the value will \
                          be taken from the cli.toml or HAB_BLDR_URL environment variable if defined. \
@@ -535,6 +535,7 @@ pub fn get() -> App<'static, 'static> {
                     environment variable if defined. (default: https://bldr.habitat.sh)")
                 (@arg PKG_IDENT: +required +takes_value {valid_fully_qualified_ident} "A fully qualified package identifier \
                     (ex: core/busybox-static/1.42.2/20170513215502)")
+                (arg: arg_target())
                 (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
             )
             (@subcommand promote =>
@@ -545,6 +546,7 @@ pub fn get() -> App<'static, 'static> {
                     environment variable if defined. (default: https://bldr.habitat.sh)")
                 (@arg PKG_IDENT: +required +takes_value {valid_fully_qualified_ident} "A fully qualified package identifier \
                     (ex: core/busybox-static/1.42.2/20170513215502)")
+                (arg: arg_target())
                 (@arg CHANNEL: +required +takes_value "Promote to the specified release channel")
                 (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
             )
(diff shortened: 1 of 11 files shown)


components/hab/src/cli.rs
```diff
@@ -535,6 +535,8 @@ pub fn get() -> App<'static, 'static> {
                     environment variable if defined. (default: https://bldr.habitat.sh)")
                 (@arg PKG_IDENT: +required +takes_value {valid_fully_qualified_ident} "A fully qualified package identifier \
                     (ex: core/busybox-static/1.42.2/20170513215502)")
+                (@arg PKG_TARGET: +takes_value {valid_target}
+                    "A package target (ex: x86_64-windows) (default: x86_64-linux)")
```
christophermaier: Shouldn't the default be whatever platform the `hab` executable you're using to run this command is compiled for?

I can imagine that with more people using Habitat on Windows, it'll quickly get old having to constantly specify `x86_64-windows` on everything.

(I know that's not part of this change, strictly speaking, but it feels like a good time to address it.)

components/hab/src/command/pkg/channels.rs
```diff
@@ -41,11 +41,16 @@ use crate::{error::Result,
...
+             bldr_url: &str,
+             ident: &PackageIdent,
+             target: PackageTarget,
+             token: Option<&str>)
+             -> Result<()> {
     let api_client = Client::new(bldr_url, PRODUCT, VERSION, None)?;
 
     ui.begin(format!("Retrieving channels for {}", ident))?;
-    let channels = api_client.package_channels(ident, token)?;
+    let channels = api_client.package_channels(ident, target, token)?;
```
christophermaier: Since "target" hasn't ever been a particularly obvious part of our API, we should add it to the `ui.begin(...)` line above.

(This applies to all modified commands... having clear feedback for _exactly_ which artifacts were operated on, particularly since we can have identical package identifiers apart from the target, is going to be very important.)

components/hab/src/cli.rs
chefsalim: There's pros and cons to doing that .. eg, it will make the usage on Mac more annoying - since 99% of the time we are operating on non-Mac packages.  There doesn't seem to be any great solution - even when we get the target to be part of the ident they basically will still have to type in the target.

components/hab/src/command/pkg/channels.rs
chefsalim: Good point - will fix

components/hab/src/cli.rs
christophermaier: 🤔 Hrmm, that's true....

I wonder if it would be worthwhile to allow users to define an environment variable for their commonly-used platform target so it could "do what they mean" in this case.

components/hab/src/command/pkg/channels.rs
```diff
@@ -41,11 +41,16 @@ use crate::{error::Result,
...
+             bldr_url: &str,
+             ident: &PackageIdent,
+             target: PackageTarget,
+             token: Option<&str>)
+             -> Result<()> {
     let api_client = Client::new(bldr_url, PRODUCT, VERSION, None)?;
 
-    ui.begin(format!("Retrieving channels for {}", ident))?;
-    let channels = api_client.package_channels(ident, token)?;
+    ui.begin(format!("Retrieving channels for {} ({})", ident, target))?;
```
christophermaier: I think I prefer the parenthesized target that you have here over something like 

```rust
debug!("Demoting package {}, target {}", ident, target);
```
elsewhere.

chefsalim: Yeah I tried it out and changed it from (target: {}) to just ({})

Leaving the debug statements as is for now seems ok

christophermaier: I just realized that this is a `ui.begin` call, whereas everything else was a `debug` call, and thus not visible by default.

Given that we've never emphasized target like this before, I'm a little concerned about the lack of feedback that's visible by default. 

In particular, I'm concerned about a user, say, demoting a linux2-targeted package that happens to have the same identifier as a linux-target package (or any such combination of targets with identical identifiers), but forgetting that they're defaulted to a linux target and demoting the wrong package, with no obvious feedback that that's what happened.

components/hab/src/cli.rs
baumanj: Also, instead of duplicating the same `@arg PKG_TARGET` 4 times, let's add a single function, like for [`arg_cache_key_path`](https://github.com/habitat-sh/habitat/blob/68b1ee1defb9f8b27b54d4e8233f1c707f158fa0/components/hab/src/cli.rs#L837).

components/hab/src/command/pkg/channels.rs
chefsalim: Are you proposing additional output lines or something else @christophermaier ?

components/hab/src/cli.rs
chefsalim: Thats an interesting proposal @baumanj .. having non-consistent defaults might be the pragmatic option. 

Alternatively, we could also consider enforcing target as a required param.  That way, there would be no confusion or misunderstanding of the user intent, and it would align to what we eventually want to do anyways (make the target mandatory).  The main downside is it would be a breaking change.

components/hab/src/command/pkg/channels.rs
christophermaier: Additional output lines.

components/hab/src/cli.rs
christophermaier: I like the non-consistent defaults idea, along with an environment variable.

christophermaier: Adding the package target option with a function call is also a good call.

components/hab/src/command/pkg/channels.rs
chefsalim: 👍 will add

components/hab/src/cli.rs
chefsalim: cool sounds good - will go with the non-consistent option

PR #6389, opened 2019-04-08
Switch to parking_lot's RwLock and add CI deadlock detection

baumanj:
Switch to the parking_lot implementation of RwLock since it's necessary to get the deadlock detection we want in CI.

Add feature-based deadlock detection for butterfly tests

Pass `--features=deadlock_detection` to `cargo test` to check for deadlocks in `habitat_butterfly` tests that rely on `wait_for_health_of` in `habitat_butterfly_test`. This currently requires a nightly toolchain. Running on stable is still possible without the feature.

See https://github.com/Amanieu/parking_lot/blob/master/src/deadlock.rs


components/butterfly-test/src/lib.rs
```diff
@@ -191,14 +217,29 @@ impl SwimNet {
     }
 
     pub fn health_of(&self, from_entry: usize, to_entry: usize) -> Option<Health> {
+        assert!(cfg!(feature = "deadlock_detection"),
```
mwrock: shouldn't this be `cfg!(not(feature = "deadlock_detection"))`? I'm getting `This test should be run with --features=deadlock_detection` even if I specify it in cargo test.

components/butterfly/src/member.rs
```diff
@@ -562,6 +555,28 @@ impl MemberList {
             .map(|member_list::Entry { health, .. }| *health)
     }
 
+    /// Returns the health of the member, blocking for a limited timeout
+    ///
+    /// Errors:
+    /// * `Error::Timeout` if the health data can't be accessed within `timeout`
+    /// * `Error::UnknownMember` if the member does not exist
+    pub fn health_of_by_id_with_timeout(&self,
```
christophermaier: Might be worth adding a note here that this is currently only intended for use from `butterfly-test`, and not "real" code.

test/run_cargo_test.sh
```diff
@@ -2,8 +2,14 @@
 
 set -eou pipefail
 
+source ./support/ci/shared.sh
+
+: "${toolchain:=stable}"
+
 while [[ $# -gt 1 ]]; do
   case $1 in
+    --nightly )             toolchain=$(get_current_toolchain)
```
christophermaier: It's probably time to change the name of `get_current_toolchain` to something like `get_nightly_toolchain` or `get_current_nightly_toolchain`.

mwrock:
Just to clarify: this does not fix the underlying deadlock issue but allows an "escape valve" in tests for when we hit the deadlock right? I can't seem to replicate the deadlock locally anymore but I'm guessing that's not because this "fixes" it.

baumanj:
> Just to clarify: this does not fix the underlying deadlock issue but allows an "escape valve" in tests for when we hit the deadlock right?

Correct, this isn't a fix for the fact that we have code paths that sometimes result in recursive lock acquisition. This change does a few things:

1. Change to test code to avoid deadlock by making it call the new [`health_of_by_id_with_timeout` ](https://github.com/habitat-sh/habitat/blob/df7e50ffdad133722e85fe8354b2bd585b31a1a7/components/butterfly/src/member.rs#L563)instead of [`health_of_by_id`](https://github.com/habitat-sh/habitat/blob/df7e50ffdad133722e85fe8354b2bd585b31a1a7/components/butterfly/src/member.rs#L552). This should turn tests that deadlock and timeout into tests that fail and show us where the problems are occurring.
1. 
I can't seem to replicate the deadlock locally anymore but I'm guessing that's not because this "fixes" it.

.expeditor/verify.pipeline.yml
```diff
@@ -221,22 +221,6 @@ steps:
         limit: 1
 
   - label: "[unit] :linux: sup"
```
scotthain: Did we intend to remove the inconsistent sup tests? I assume yes.

test/run_cargo_test.ps1
```diff
@@ -8,22 +8,29 @@ param (
     # Features to pass to cargo
     [string]$Features,
     # Options to pass to the cargo test command
-    [string]$TestOptions
+    [string]$TestOptions,
+    [string]$toolchain = "stable"
```
scotthain: In powershell this would generally be `$Toolchain`

.expeditor/verify.pipeline.yml
baumanj: Yes. It's possible we could add them back later if they prove to be unrelated to the locking issue, but hopefully this will change them from being inconsistent passing tests to consistently passing or failing, which we can deal with more directly.

test/run_cargo_test.ps1
baumanj: Honestly, this is a hacky solution that I was hoping to get changed to match the linux approach, but because I don't have the `pwsh` skills, I left it like this. If we're *happy* with linux doing
```
run_cargo_test.sh --nightly
```
while windows does
```
run_cargo_test.ps1 butterfly -toolchain "nightly" …
```
I can change it to
```
run_cargo_test.ps1 butterfly -Toolchain "nightly" …
```
But I think the better thing would be
```
run_cargo_test.ps1 butterfly -Nightly …
```
along with adding an equivalent for `get_nightly_toolchain` (before this PR: `get_current_toolchain`) for windows. Both versions should probably read that from a file rather than having it hard-coded in [`support/ci/shared.sh`](https://github.com/habitat-sh/habitat/blob/master/support/ci/shared.sh#L16). But because I didn't want to burn cycles doing `pwsh` that @mwrock could probably do much faster/better, I didn't.

If it's a simple thing, could you add a commit that makes that fix, @mwrock?

mwrock: done @baumanj

PR #6391, opened 2019-04-08
Be explicit when dealing with feature flags

christophermaier:
Here, we dispense with the use of the `features` crate, in favor of
using the underlying `bitflags` abstraction directly.

This means that we no longer rely on recording which feature flags are
currently enabled in global state, but in immutable data that we
explicitly thread through function calls.

This also means that functionality that differs based on feature flag
state can be more easily unit tested (otherwise, we have to contend
with the sharing of global state, which is always tricky).

### DIFF
--- a/components/common/src/lib.rs
+++ b/components/common/src/lib.rs
@@ -16,9 +16,9 @@ use crate::ui::{UIWriter,
                 UI};
 use habitat_api_client as api_client;
 use habitat_core as hcore;
-use habitat_core::env as henv;
 use lazy_static::lazy_static;
 use std::{collections::HashMap,
+          env,
           iter::FromIterator};
 
 extern crate json;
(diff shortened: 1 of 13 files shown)


components/common/src/lib.rs
```diff
@@ -52,3 +59,97 @@ lazy_static::lazy_static! {
...
+}
+
+impl FeatureFlag {
+    /// If the environment variable for a flag is set to _anything_ but
+    /// the empty string, it is activated.
+    pub fn from_env(ui: &mut UI) -> Self {
+        let mut flags = FeatureFlag::empty();
+
+        for (feature, env_var) in ENV_VARS.iter() {
+            if henv::var(env_var).is_ok() {
```
baumanj: Can we use use `std::env::var_os` and `is_some` instead?

christophermaier: Yup, I just realized I forgot about that... adding it now.

christophermaier:
Actually, I'm going to rebase this onto #6367 and merge it with that.

PR #6416, opened 2019-04-11
Change windows agents from single-use-windows-privileged -> default-windows-privileged

baumanj:
Per conversation with @tduffield, this should allow us to reduce the number of agents we require and make the overall CI suite faster.


baumanj:
[The clippy step](https://buildkite.com/chef/habitat-sh-habitat-master-verify/builds/1408#912f5e4e-ff9c-44e7-bb7d-ea00b947bf07) is failing, but that's expected. I have fixes queued up in a branch, but I wanted to merge this first to reduce the CI load/wait.

PR #6417, opened 2019-04-11
Fix clippy lints

baumanj:
Not sure how these got in. Perhaps they're newly detected?


baumanj:
Not sure why I keep getting more lints on CI that I don't see locally (even after a `cargo clean`), but I'm down to make improvements.

@raskchanky: https://github.com/habitat-sh/habitat/pull/6417/commits/8d32bf710f7889998d760cd716941db892ebdf7e may be of interest

PR #6421, opened 2019-04-12
Bump to Rust 1.34.0

christophermaier:


christophermaier:
We are using the current stable rust package from builder. Once that's updated, we'll get it in CI.

Issue #6418, opened 2019-04-11
`hab pkg install --binlink` does not use `HAB_BINLINK_DIR` environment variable


mwrock:
It seems like a recent change in 0.79.0 (maybe) has introduced a bug where binlinked binaries are not on the path in a local linux studio. Note that docker studios and windows studios are not affected. Also this is only affected by `hab pkg install --binlink` and not `hab pkg binlink`. studio.

`hab pkg install --binlink` links to `/bin` and not `HAB_PKG_BINLINK`

baumanj:
The change that's causing the issue seems to be https://github.com/habitat-sh/habitat/pull/6315/commits/6232c6728f29f825b4324b81ff7c77b7eec95d79. Specifically the handling of the `-b|--binlink` argument to `hab pkg install`; `hab pkg binlink` itself seems to still work correctly.

PR #6423, opened 2019-04-12
Make --binlink take an value (or env var) for `hab pkg install`

baumanj:
Resolves https://github.com/habitat-sh/habitat/issues/6418

Commit 6232c6728f29f825b4324b81ff7c77b7eec95d79 removed `binlink_dest_dir_from_matches`, which applied the value in the HAB_BINLINK_DIR env var to choose the destination when the `--binlink` flag was provided. This resulted in a bug where the binlinks weren't in the expected location in a new studio.

This commit changes `--binlink` from a flag to a option that takes a value. That value has platform-specific defaults, allowing the existing use case of `--binlink` without a value to work with either the default, or with a value supplied by the `HAB_BINLINK_DIR` env var.

One notable change is that since `--binlink` now takes a value, it must be placed *after* the `<PKG_IDENT_OR_ARTIFACT>` positional argument if the default value is desired. Otherwise, something like `hab pkg install --binlink core/redis` would interpret `core/redis` as
the binlink path. Fortunately, this would be caught by clap since the command appears to lack the required `<PKG_IDENT_OR_ARTIFACT>` positional argument, yielding the error:
```
error: The following required arguments were not provided:
    <PKG_IDENT_OR_ARTIFACT>...

USAGE:
    hab pkg install <PKG_IDENT_OR_ARTIFACT>... --binlink <BINLINK> --channel <CHANNEL>
```

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -24,7 +24,6 @@ extern crate log;
 #[cfg(windows)]
 use crate::hcore::crypto::dpapi::encrypt;
 use crate::{common::{cli::{cache_key_path_from_matches,
-                           DEFAULT_BINLINK_DIR,
                            FS_ROOT},
                      command::package::install::{InstallHookMode,
                                                  InstallMode,
@@ -92,7 +91,8 @@ use std::{env,
                prelude::*,
                Read},
           net::ToSocketAddrs,
-          path::Path,
+          path::{Path,
+                 PathBuf},
           process,
           result,
           str::FromStr,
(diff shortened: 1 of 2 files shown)


components/hab/src/main.rs
```diff
@@ -1797,6 +1808,123 @@ fn update_svc_load_from_input(m: &ArgMatches<'_>, msg: &mut protocol::ctl::SvcLo
...
+        fn matches_for_pkg_install<'a>(pkg_install_args: &'a [&'a str]) -> ArgMatches<'a> {
+            let pre_pkg_install_args = &["hab", "pkg", "install"];
+            let app_matches = cli::get(FeatureFlag::empty())
+                .get_matches_from_safe(pre_pkg_install_args.iter().chain(pkg_install_args.iter()))
+                .unwrap(); // Force panics on CLAP errors, so we can use #[should_panic]
+            match app_matches.subcommand() {
+                ("pkg", Some(matches)) => {
+                    match matches.subcommand() {
+                        ("install", Some(m)) => {
+                            println!("{:#?}", m);
```
mwrock: this looks like debug cruft that got unintentionally left?

baumanj: No, that's intentional. All test output is hidden unless the `--nocapture` flag is passed to the test binary. In the case things aren't working as expected, this output is very helpful to see what's going on.

PR #6427, opened 2019-04-15
Follow $refs in generate-template-reference

### LINKED ISSUE #5948: Template data documentation renderer does not honor $refs that point to other files
Recently, the JSON schema file that defines our template data was refactored to refer to the `sys` object defined in a separate file.

https://github.com/habitat-sh/habitat/blob/873aab85240cd7ad1867ddb25abd4bcf1e1c5201/components/sup/doc/render_context_schema.json#L220

The [generate-template-reference.js](https://github.com/habitat-sh/habitat/blob/master/www/scripts/generate-template-reference.js) script, which generates our HTML documentation from this JSON schema, cannot deal with `$ref`s that point to separate files. This causes the rendered documentation to be incorrect, necessitating manual fixes (it ends up deleting the contents of the top-level `sys` object).

We've caught this when generating the documentation after a release, so the documentation that users see should be OK, but it's one more thing to manually check up on for each release.

mpeck:
Uses the json schema $ref parser https://apidevtools.org/json-schema-ref-parser/ to follow refs when generating data documentation with generate-template-reference.js. resolves #5948

### DIFF
--- a/www/scripts/generate-template-reference.js
+++ b/www/scripts/generate-template-reference.js
@@ -1,6 +1,5 @@
-const stdin = process.stdin;
+let $RefParser = require('json-schema-ref-parser');
 const stdout = process.stdout;
-let input = [];
 let lines = [];
 
 function writeHeader() {
@@ -92,14 +91,11 @@ function getType(prop) {
   return '--';
 }
 
-stdin.on('data', d => {
-  input.push(d);
-});
-
-stdin.on('end', () => {
-  schema = JSON.parse(input.join());
-  writeHeader();
-  writeProperties();
-  writeDefinitions();
-  stdout.write(lines.join('\n'));
+$RefParser.bundle(process.argv[2]).then(function(deref_schema) {
+    console.log("ARGV", process.argv[2]);
+    schema = deref_schema;
+    writeHeader();
+    writeProperties();
+    writeDefinitions();
+    stdout.write(lines.join('\n'));
 });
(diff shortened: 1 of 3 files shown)


www/scripts/generate-template-reference.js
```diff
@@ -92,14 +91,11 @@ function getType(prop) {
...
-  input.push(d);
-});
-
-stdin.on('end', () => {
-  schema = JSON.parse(input.join());
-  writeHeader();
-  writeProperties();
-  writeDefinitions();
-  stdout.write(lines.join('\n'));
+$RefParser.bundle(process.argv[2]).then(function(deref_schema) {
```
raskchanky: This isn't anything that you need to change, but just for my own edification, I'm curious about the switch from processing stdin to taking a positional file argument.  Is that just personal preference or is there a different reason for it?

mpeck: If we pipe the file in with stdin we lose the context of the file. References to external files are relative so the library could have a hard time finding them dependent on the directory that the script is run in. This way it doesn't actually mater where you run the script from; it should be able to find its external dependencies regardless... At least that was my theory, I didn't actually test that. Also it just seems cleaner for the script to take an argument instead of reading on stdin :stuck_out_tongue_winking_eye:

raskchanky: No, that totally makes sense.  Thanks for the explanation!

christophermaier:
tenor-23528994

PR #6428, opened 2019-04-15
Add test habitat pipeline

scotthain:

### DIFF
--- a/.expeditor/build.habitat.yml
+++ b/.expeditor/build.habitat.yml
@@ -1,0 +1,3 @@
+---
+origin: scotthain
+smart_build: true
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -44,3 +44,4 @@ pipelines:
       description: Deploy the website
       definition: .expeditor/website.pipeline.yml
       trigger: pull_request
+  - habitat/build


.expeditor/build.habitat.yml
```diff
@@ -0,0 +1,3 @@
+---
+origin: scotthain
```
scotthain: This is intentionally set to my origin while testing so I don't pollute core with artifacts.

.expeditor/build.habitat.yml
```diff
@@ -0,0 +1,3 @@
...
+smart_build: true
```
christophermaier: What makes it "smart"?

.expeditor/config.yml
```diff
@@ -44,3 +44,4 @@ pipelines:
       description: Deploy the website
       definition: .expeditor/website.pipeline.yml
       trigger: pull_request
+  - habitat/build
```
christophermaier: I assume this maps to `build.habitat.yml`, yeah? If so, why the inversion in the name?

.expeditor/build.habitat.yml
scotthain: Expeditor only rebuilds packages that have been modified if smart build is true. (This actually will have no effect currently because all this PR does is *create* the pipeline

.expeditor/config.yml
scotthain: Yep - that's the default translation: https://expeditor.chef.io/docs/integrations/buildkite/#default-definition-values

tduffield: More context: `habitat` is the _type_ of pipeline and `build` is the name.

PR #6437, opened 2019-04-16
Add temporary hab release pipeline

scotthain:

Getting pipelines set up for future use.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -45,3 +45,5 @@ pipelines:
       definition: .expeditor/website.pipeline.yml
       trigger: pull_request
   - habitat/build
+  - habitat_release:
+      description: Habitat release process
--- a/.expeditor/habitat_release.pipeline.yml
+++ b/.expeditor/habitat_release.pipeline.yml
@@ -1,0 +1,25 @@
+expeditor:
+  defaults:
+    buildkite:
+      timeout_in_minutes: 30
+
+steps:
+#######################################################################
+# Release!
+#######################################################################
+
+  - label: "[lint] :linux: :bash: Shellcheck"
+    command:
+      - echo "HI"
+    agents:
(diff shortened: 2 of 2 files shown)


.expeditor/habitat_release.pipeline.yml
```diff
@@ -0,0 +1,25 @@
...
+    plugins:
+      docker#v3.0.1:
+        always-pull: true
+        user: "buildkite-agent"
+        group: "buildkite-agent"
+        image: "chefes/buildkite"
+    timeout_in_minutes: 10
+    retry:
+      automatic:
+        limit: 1
```
christophermaier: Any reason to retry Shellcheck, which should be deterministic?

.expeditor/config.yml
```diff
@@ -45,3 +45,5 @@ pipelines:
       definition: .expeditor/website.pipeline.yml
       trigger: pull_request
   - habitat/build
+  - habitat_release:
```
christophermaier: This is more for my edification than anything, but what's the distinction between `habitat/build` (with a `/`) and `habitat_release` (with a `_`)?

.expeditor/habitat_release.pipeline.yml
scotthain: Oh I this is a placeholder (check the command). That being said any of the steps has the possibility of "infrastructure instability" so having a single retry in case of a blip is a good thing.

.expeditor/config.yml
scotthain: Good eye - `habitat/build` defines a expeditor "habitat" type pipeline called "build". In this instance `habitat_release` is the name of a "generic" pipeline. More info here https://expeditor.chef.io/docs/integrations/buildkite/#supported-pipeline-trigger-types

PR #6449, opened 2019-04-17
Rebuild pipeline with the correct name

scotthain:


scotthain:
Because I read the docs wrong - it created the pipeline correctly but not the way I wanted it. This has the correct name (i want a default pipeline, not a hab pipeline)

Issue #2447, opened 2017-05-20
Configurable kill signal and timeout


sethvargo:
Right now, Habitat sends TERM to any process it's supervising on stop. Not all applications respond to TERM, some respond to INT or even USR1 for a custom, clean shutdown.

I talked with @fnichol about this, and I think we agreed that it makes sense for plans to describe their termination schedule, perhaps like this:

```sh
pkg_kill_signal=(INT)
```

Additionally, it would be nice if packages could customize the amount of time they want to wait for a clean shutdown like:

```sh
pkg_kill_timeout=(30)
```

smith:
How about `pkg_svc_kill_signal` to indicate that it concerns the service provided by the package? More consistent with variables like `pkg_svc_user`.

sethvargo:
I'm indifferent on the naming tbh - more concerned with the feature 😄

bodymindarts:
@sethvargo what is the use case you need this for?
I agree that the shutdown process needs to be configurable in a more granular way. I was thinking of a `shutdown` hook that could be called pre `kill`ing. How would that fit in to your needs?

bodymindarts:
https://github.com/habitat-sh/habitat/issues/1994 is related to supervisor shutdown behaviour.

sethvargo:
> what is the use case you need this for?

I have an app that stops gracefully on sigint (it's a webserver). Given INT, it properly drains connections and shuts down gracefully. Given TERM, it dies and closes all connections, potentially killing a download or active request.

christophermaier:
This is blocked pending the refactorings involved with #1973

stevendanna:
Somehow I didn't see this when filing https://github.com/habitat-sh/habitat/issues/5135 which is a aimed at solving the same basic issue

dmccown:
Is this still blocked on #1973?

PR #6450, opened 2019-04-17
Configurable Shutdown, Part 1

christophermaier:
Please read the commit messages for further details.

More will be coming, but this is most of the plumbing required for #2447 .

### DIFF
--- a/components/core/src/os/process/mod.rs
+++ b/components/core/src/os/process/mod.rs
@@ -29,17 +29,17 @@ pub use self::unix::{become_command,
                      current_pid,
                      is_alive,
                      signal,
-                     Pid,
-                     ShutdownSignal,
-                     Signal};
+                     Pid};
 #[cfg(windows)]
 pub use self::windows::{become_command,
                         current_pid,
                         handle_from_pid,
                         is_alive,
                         Pid};
+
 use crate::error::Error;
 use std::{fmt,
+          result,
           str::FromStr};
 use time::Duration;
 
@@ -81,4 +81,138 @@ impl From<ShutdownTimeout> for u32 {
 
 impl From<ShutdownTimeout> for Duration {
     fn from(timeout: ShutdownTimeout) -> Self { Duration::seconds(timeout.0.into()) }
+}
+
+// This defines a handful of Unix signals that we want to deal with,
+// but we are making it available on Windows as well for situations
+// where a Windows CLI is communicating with a Linux Supervisor.
+#[allow(non_snake_case)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Signal {
+    INT,
+    ILL,
+    ABRT,
+    FPE,
+    KILL,
+    SEGV,
+    TERM,
+    HUP,
+    QUIT,
+    ALRM,
+    USR1,
+    USR2,
+    CHLD,
+}
+
+impl FromStr for Signal {
+    type Err = Error;
+
+    fn from_str(s: &str) -> result::Result<Self, Self::Err> {
+        let signal = match s {
+            "HUP" => Signal::HUP,
+            "INT" => Signal::INT,
+            "QUIT" => Signal::QUIT,
+            "ILL" => Signal::ILL,
+            "ABRT" => Signal::ABRT,
+            "FPE" => Signal::FPE,
+            "KILL" => Signal::KILL,
+            "USR1" => Signal::USR1,
+            "SEGV" => Signal::SEGV,
+            "USR2" => Signal::USR2,
+            "ALRM" => Signal::ALRM,
+            "TERM" => Signal::TERM,
+            "CHLD" => Signal::CHLD,
+            _ => return Err(Error::ParseSignalError(s.to_string())),
+        };
+        Ok(signal)
+    }
+}
+
+impl fmt::Display for Signal {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let s = match self {
+            Signal::HUP => "HUP",
+            Signal::INT => "INT",
+            Signal::QUIT => "QUIT",
+            Signal::ILL => "ILL",
+            Signal::ABRT => "ABRT",
+            Signal::FPE => "FPE",
+            Signal::KILL => "KILL",
+            Signal::USR1 => "USR1",
+            Signal::SEGV => "SEGV",
+            Signal::USR2 => "USR2",
+            Signal::ALRM => "ALRM",
+            Signal::TERM => "TERM",
+            Signal::CHLD => "CHLD",
+        };
+        write!(f, "{}", s)
+    }
+}
+
+/// Encapsulates logic for defining the default shutdown signal we
+/// send services, and handles translation from external types at the
+/// edges of our system.
+#[derive(Debug, Clone)]
+pub struct ShutdownSignal(Signal);
+
+impl Default for ShutdownSignal {
+    /// Unless otherwise specified, the Supervisor will shut down
+    /// services by sending the `TERM` signal.
+    fn default() -> Self { Signal::TERM.into() }
+}
+
+impl FromStr for ShutdownSignal {
+    type Err = Error;
+
+    fn from_str(s: &str) -> result::Result<Self, Self::Err> { Ok(ShutdownSignal(s.parse()?)) }
+}
+
+impl fmt::Display for ShutdownSignal {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { write!(f, "{}", self.0) }
+}
+
+impl From<Signal> for ShutdownSignal {
+    fn from(signal: Signal) -> Self { ShutdownSignal(signal) }
+}
+
+impl From<ShutdownSignal> for Signal {
+    fn from(shutdown_signal: ShutdownSignal) -> Self { shutdown_signal.0 }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn signal_names_are_only_accepted_as_uppercase() {
+        assert_eq!(Signal::HUP, "HUP".parse().unwrap());
+        assert!("hup".parse::<Signal>().is_err());
+    }
+
+    #[test]
+    fn signals_can_render_as_strings() {
+        assert_eq!("HUP", Signal::HUP.to_string());
+    }
+
+    #[test]
+    fn signals_can_round_trip_through_parsing() {
+        for signal in &[Signal::HUP,
+                        Signal::INT,
+                        Signal::QUIT,
+                        Signal::ABRT,
+                        Signal::FPE,
+                        Signal::KILL,
+                        Signal::USR1,
+                        Signal::SEGV,
+                        Signal::USR2,
+                        Signal::ALRM,
+                        Signal::TERM,
+                        Signal::CHLD]
+        {
+            assert_eq!(*signal,
+                       signal.to_string()
+                             .parse::<Signal>()
+                             .expect("Couldn't parse back into a Signal!"));
+        }
+    }
 }
--- a/components/core/src/os/process/mod.rs
+++ b/components/core/src/os/process/mod.rs
@@ -22,13 +22,6 @@ mod windows;
 #[cfg(unix)]
 mod unix;
 
-#[cfg(windows)]
-pub use self::windows::{become_command,
-                        current_pid,
-                        handle_from_pid,
-                        is_alive,
-                        Pid};
-
 #[cfg(unix)]
 pub(crate) use self::unix::SignalCode;
 #[cfg(unix)]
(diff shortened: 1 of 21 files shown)


christophermaier:
Ping @stevendanna

components/sup-protocol/protocols/ctl.proto
```diff
@@ -104,6 +104,11 @@ message SvcLoad {
 // Request to unload a loaded service.
 message SvcUnload {
   optional sup.types.PackageIdent ident = 1;
+  // Name of the signal to send the service to shut it down (e.g.,
+  // "TERM" and not "SIGTERM"). Only applies to Unix platforms.
+  optional string signal = 2;
+  // Timeout in seconds before killing the service
+  optional uint32 timeout = 3;
```
baumanj: I'd prefer
```suggestion
  optional uint32 timeout_in_secs = 3;
```
and ditch the comment. Similarly elsewhere.

components/sup/src/ctl_gateway/acceptor.rs
```diff
@@ -0,0 +1,78 @@
...
+    type Item = CtlHandler;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        match self.shutdown_trigger.poll() {
+            Ok(Async::Ready(())) => {
+                info!("Signal received; stopping CtlAcceptor");
+                Ok(Async::Ready(None))
+            }
+            Err(e) => {
+                error!("Error polling CtlAcceptor shutdown trigger: {:?}", e);
```
baumanj:
```suggestion
                error!("Error polling CtlAcceptor shutdown trigger: {}", e);
```
This is user-facing

components/sup/src/ctl_gateway/server.rs
```diff
@@ -277,127 +392,20 @@ impl Future for SrvHandler {
...
-                                }
-                                _ => {
-                                    warn!("Unhandled message, {}", msg.message_id());
-                                    break;
-                                }
-                            };
-                            match self.mgr_tx.start_send(cmd) {
+
+                            let cmd =
+                                match Self::command_from_message(&msg, self.ctl_sender.clone()) {
```
baumanj: This is 3 nested matches inside a loop. Maybe worth breaking into some more functions?

components/hab/src/main.rs
```diff
@@ -1038,13 +1038,18 @@ fn sub_svc_load(m: &ArgMatches<'_>) -> Result<()> {
...
-fn sub_svc_unload(m: &ArgMatches<'_>) -> Result<()> {
+fn sub_svc_unload(m: &ArgMatches<'_>, feature_flags: FeatureFlag) -> Result<()> {
     let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?;
     let cfg = config::load()?;
     let listen_ctl_addr = listen_ctl_addr_from_input(m)?;
     let secret_key = ctl_secret_key(&cfg)?;
-    let mut msg = protocol::ctl::SvcUnload::default();
-    msg.ident = Some(ident.into());
+
+    let timeout = maybe_get_shutdown_timeout(m, feature_flags)?.map(Into::into);
```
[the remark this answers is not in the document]
christophermaier: 🤔 That _is_ nicer.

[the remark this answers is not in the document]
christophermaier: Well, hrmm... on second thought, there's a balancing act here. This is essentially adding type annotations, which we generally don't do unless we have to for the compiler's sake.

Do you have an idea on a general rule / guiding principle by which to determine whether we use `into` or `from`? Or why not something equivalent like 

```rust
let timeout: Option<u32> = maybe_get_shutdown_timeout(m, feature_flags)?.map(Into::into);
```

components/hab/src/main.rs
```diff
@@ -1053,12 +1058,49 @@ fn sub_svc_unload(m: &ArgMatches<'_>) -> Result<()> {
...
+    if feature_flags.contains(FeatureFlag::CONFIGURE_SHUTDOWN) {
+        // We convert the parsed ShutdownSignal to a String here
+        // because:
+        //
+        // a) We ultimately need it as a String to create a
+        //    protobuf message anyway, and
+        // b) ShutdownSignal doesn't exist on Windows, and we still
+        //    need to compile this function somehow.
+        Ok(Some(m.value_of("SHUTDOWN_SIGNAL")
+                 .unwrap() // safe because of Clap
```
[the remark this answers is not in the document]
christophermaier: Yup, you're right.

components/hab/src/main.rs
```diff
@@ -1053,12 +1058,49 @@ fn sub_svc_unload(m: &ArgMatches<'_>) -> Result<()> {
...
+        Ok(None)
+    }
+}
+
+fn maybe_get_shutdown_timeout(m: &ArgMatches<'_>,
+                              feature_flags: FeatureFlag)
+                              -> Result<Option<ShutdownTimeout>> {
+    if feature_flags.contains(FeatureFlag::CONFIGURE_SHUTDOWN) {
+        // unwrap is safe because we define a default with clap
+        Ok(Some(m.value_of("SHUTDOWN_TIMEOUT").unwrap().parse()?))
```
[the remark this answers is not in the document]
christophermaier: I like the utility function idea 👍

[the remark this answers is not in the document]
christophermaier: Might be worth doing that in a separate PR, and ensuring that we convert _everything_ over at once.

components/hab/src/main.rs
```diff
@@ -79,10 +41,47 @@ use hab::{analytics,
...
+                             PackageIdent,
+                             PackageTarget},
+                   service::{HealthCheckInterval,
+                             ServiceGroup},
+                   url::{bldr_url_from_env,
+                         default_bldr_url},
+                   ChannelIdent};
+use habitat_sup_client::{SrvClient,
+                         SrvClientError};
+use habitat_sup_protocol::{codec::*,
```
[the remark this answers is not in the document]
christophermaier: Yup!

components/sup-protocol/protocols/ctl.proto
christophermaier: Yup

components/sup/src/ctl_gateway/server.rs
christophermaier: Have to think on that a bit. On first glance, it looks like those functions' return values would have to contend with actual Ok/Err values, as well as signaling whether to break or continue from the main loop they'd be embedded in. I'm not seeing an obvious refactoring that wouldn't be more complicated as a result. There may be an alternative way of structuring this entire function to make it more clear, but that may be a bigger (and more risky) job at the moment.

components/hab/src/main.rs
```diff
@@ -1038,13 +1038,18 @@ fn sub_svc_load(m: &ArgMatches<'_>) -> Result<()> {
...
+fn sub_svc_unload(m: &ArgMatches<'_>, feature_flags: FeatureFlag) -> Result<()> {
     let ident = PackageIdent::from_str(m.value_of("PKG_IDENT").unwrap())?;
     let cfg = config::load()?;
     let listen_ctl_addr = listen_ctl_addr_from_input(m)?;
     let secret_key = ctl_secret_key(&cfg)?;
-    let mut msg = protocol::ctl::SvcUnload::default();
-    msg.ident = Some(ident.into());
+
+    let timeout = maybe_get_shutdown_timeout(m, feature_flags)?.map(Into::into);
+    let signal = maybe_get_shutdown_signal(m, feature_flags)?;
```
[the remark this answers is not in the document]
christophermaier: Ah, yes... looks like this was just an incomplete refactoring.

components/hab/src/cli.rs
```diff
@@ -1316,6 +1324,47 @@ fn non_empty(val: String) -> result::Result<(), String> {
...
+
+    app
+}
+
+#[cfg(windows)]
+fn maybe_add_configurable_shutdown_options(mut app: App<'static, 'static>,
+                                           feature_flags: FeatureFlag)
+                                           -> App<'static, 'static> {
+    // Identical to the Unix implementation, but does not include the
+    // SHUTDOWN_SIGNAL option, since that doesn't exist on Windows
```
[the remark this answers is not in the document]
christophermaier: Hrmm, good catch. I'll need to expose the signal types and defaults for Windows, but it would only get used on the CLI side; all the Supervisor-side code can stay the same.

components/hab/src/main.rs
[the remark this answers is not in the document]
christophermaier: Issue filed: https://github.com/habitat-sh/habitat/issues/6471

components/core/src/os/process/mod.rs
```diff
@@ -22,19 +22,197 @@ mod windows;
...
-                     Pid,
-                     Signal};
+                     Pid};
+#[cfg(windows)]
+pub use self::windows::{become_command,
+                        current_pid,
+                        handle_from_pid,
+                        is_alive,
+                        Pid};
+
```
raskchanky: Delete this space?

components/hab/src/cli.rs
```diff
@@ -1316,6 +1323,36 @@ fn non_empty(val: String) -> result::Result<(), String> {
...
+/// Adds extra configuration options for shutting down a service with
+/// a customized shutdown signal and timeout.
+///
+/// These are currently feature-flagged. Eventually, we hope to have
+/// this be definable in a package or at load time.
+fn maybe_add_configurable_shutdown_options(mut app: App<'static, 'static>,
+                                           feature_flags: FeatureFlag)
+                                           -> App<'static, 'static> {
+    if feature_flags.contains(FeatureFlag::CONFIGURE_SHUTDOWN) {
+        app = app.arg(Arg::with_name("SHUTDOWN_SIGNAL").help("The signal to send to a service \
```
mwrock: I think we only want this to be a viable option on linux. So only add to clapp on linux platforms.

christophermaier: I had that initially, but @baumanj brought up a good point in https://github.com/habitat-sh/habitat/pull/6450#discussion_r276860946; namely, shouldn't Windows `hab` users be able to interact with Linux Supervisors?

I added https://github.com/habitat-sh/habitat/pull/6450/commits/e13c48ad2760de9fb503429b35aff9ac9cdb366d to address this.

components/core/src/os/process/mod.rs
christophermaier: D'oh!

christophermaier: Fixed: https://github.com/habitat-sh/habitat/pull/6450/commits/2d9a8a182abf655d03914dea1ffe9c7d97656584

components/sup/src/ctl_gateway/server.rs
```diff
@@ -461,46 +468,42 @@ enum SrvHandlerState {
...
+pub fn run(listen_addr: SocketAddr, secret_key: String, mgr_sender: MgrSender) {
+    let tb = thread::Builder::new().name("ctl-gateway".to_string());
+    tb.spawn(move || {
+          let mut core = reactor::Core::new().unwrap();
+          let handle = core.handle();
+          let state = SrvState { secret_key,
+                                 mgr_sender };
+          let state = Rc::new(RefCell::new(state));
+          let server =
+              TcpListener::bind(&listen_addr).unwrap()
```
raskchanky: I realize this was existing code but is it worth at least an `expect` here?  I know we've gone to great lengths to handle binding errors gracefully in other parts of the code.

components/hab/src/cli.rs
mwrock: oh ok that makes sense, but maybe we should add something to the arg help text like "applies to linux supervisors only."

mwrock: oh duh! it does.

components/sup/src/ctl_gateway/server.rs
christophermaier: Probably can't hurt! 👍

components/hab/src/main.rs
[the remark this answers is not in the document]
baumanj: My decision process is:
1. Avoid `Into::into` because the redundancy bothers me; any space that's being taken up in program text should convey meaningful info
1. If both `Foo:from` and `Bar::into` are possible, choose whichever provides more useful information in that particular context

I agree that these are like type annotations, and if I could eliminate them altogether, I'd probably prefer doing that since leaving information that isn't relevant out enhances simplicity, but at long as we have to say *something*, I prefer to say something meaningful.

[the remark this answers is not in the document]
christophermaier: @baumanj That sounds good... I don't have a particularly strong opinion, but having a sensible rubric one way or the other is definitely better than nothing at all.

Now, if only we could automate this 😅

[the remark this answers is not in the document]
davidMcneil: @baumanj @christophermaier I went through the codebase and converted `Into::into` to `from`. Two inconveniences of the `from` version are:

1. If we do not use the name of the type we are converting into anywhere but the `from` statement we have to add a `use` statement we would not otherwise need.
1. If we change the name of the type we have to change the name in any `from` statements.

Not saying we should change our decision, but I think it is good to be aware of the tradeoffs.

PR #6452, opened 2019-04-17
Move habitat_butterfly_test into habitat_butterfly

baumanj:
I'm not sure why this was a separate crate, as it only contained test code for exercising butterfly. However, it's existence as a separate crate led to a circular dependency issue between the two which is making it difficult to add features for testing.


baumanj:
A+ GIF game all around 😂

PR #6453, opened 2019-04-18
Small health check interval refactoring

christophermaier:
Just a few simplifications uncovered while working on making health check hooks run asynchronously.

Read commit messages for details.

### DIFF
--- a/components/core/src/service.rs
+++ b/components/core/src/service.rs
@@ -396,12 +396,12 @@ impl HealthCheckInterval {
     pub fn immediately() -> Self { Self::from(0) }
 }
 
-impl From<u32> for HealthCheckInterval {
-    fn from(seconds: u32) -> Self { Self(Duration::from_secs(seconds.into())) }
+impl From<u64> for HealthCheckInterval {
+    fn from(seconds: u64) -> Self { Self(Duration::from_secs(seconds)) }
 }
 
-impl AsRef<Duration> for HealthCheckInterval {
-    fn as_ref(&self) -> &Duration { &self.0 }
+impl Into<u64> for HealthCheckInterval {
+    fn into(self) -> u64 { self.0.as_secs() }
 }
 
 impl fmt::Display for HealthCheckInterval {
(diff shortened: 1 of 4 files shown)


components/core/src/service.rs
```diff
@@ -392,6 +392,14 @@ impl FromStr for ApplicationEnvironment {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct HealthCheckInterval(Duration);
 
+impl HealthCheckInterval {
+    pub fn immediately() -> Self { Self::from(0) }
+}
+
+impl From<u32> for HealthCheckInterval {
+    fn from(seconds: u32) -> Self { Self(Duration::from_secs(seconds.into())) }
+}
```
baumanj: Why this change? `HealthCheckInterval(Duration::from_secs(30))` (or `Self(Duration::from_secs(30))`)  seems clearer to me than than `Self::from(30)`, even if the latter is shorter. I generally prefer to only pass primitive types to functions that make their units clear.

christophermaier: My goal was to ensure that these kind of transformations all go through the same logic flow. That way you know that all numbers represent seconds, as opposed to being seconds in one impl block, but minutes in another.

christophermaier: Actually @baumanj what do you think of this: instead of having `HealthCheckInterval` wrap a `Duration`, just have it wrap a `u32`?

baumanj: I think
```rust
#[derive(Copy, Clone, Debug)]
pub struct HealthCheckInterval {
    pub in_secs: u64 // to match `Duration`, this could be smaller if we don't care
}
```
would be a good compromise. Make it easy to convert `HealthCheckInterval` → `Duration` for the sake of arithmetic, but letting the internal field be public simplifies consumption and makes the units clear, while preserving the type safety benefits (so long as we are diligent about passing around `HealthCheckInterval`s everywhere).

christophermaier: I swapped out `Duration` for `u64` as the underlying type and ended up simplifying a bunch of other implementations as a result.

I'm not 100% sold on making that inner type public, though, so it's still private for now.

christophermaier: Actually, I'm leaving `Duration` as the underlying type for now... `Serialize` and `Deserialize` are derived on this type, and that affects how this data is rendered into a spec file. Changing out the underlying type changes this serialization logic, and would end up breaking existing services, absent additional changes.

Since that's decidedly out of scope for this PR, I'm going to file an issue to revisit this later. If nothing else, there's no reason to be serializing `Duration` nanoseconds into a spec file.

christophermaier: See #6469

PR #6466, opened 2019-04-19
modernize build.ps1

### LINKED ISSUE #6072: Make build.ps1 better
Now that we have native habitat deps for all of the hab things, we can get rid of a bunch of stuff in this script.

mwrock:
fixes #6072 

`build.ps1` is the windows equivilent of `Makefile`. It performs common build and test tasks and also configures an environment with the needed toolchains to work with the habitat codebase. We have not updated this since moving away from the vendored dependencies in favor of native habitat dependencies. This retrofits build.ps1 to use the new habitat deps.

Now that it does that we can reduce duplicate code and replace `run_clippy` and `run_cargo_test` with this.


build.ps1
```diff
@@ -19,23 +19,32 @@ param (
...
+    # Options to pass to the cargo test command
+    [string]$TestOptions,
+    # Use nightly toolchain
+    [switch]$Nightly
 )
+$ErrorActionPreference="stop"
+
+$toolchain = "stable"
+if($Nightly) { $toolchain = "nightly" }
+elseif($Fmt) { $toolchain = "nightly-2019-04-09" }
```
christophermaier: We should make sure the instructions for bumping the nightly rustfmt during a release point out this line as something to update.

Is there a way we can define this in one place, and then pull from that source in the various build tooling we have lying around?

build.ps1
```diff
@@ -55,10 +64,6 @@ function New-PathString([string]$StartingPath, [string]$Path) {
     }
 }
 
-function Test-AppVeyor {
-    (test-path env:\APPVEYOR) -and ([bool]::Parse($env:APPVEYOR))
-}
-
```
christophermaier: tenor-238363121

build.ps1
```diff
@@ -70,162 +75,163 @@ function Invoke-Configure {
...
-$env:OPENSSL_LIB_DIR            = $ChocolateyHabitatLibDir
-$env:OPENSSL_INCLUDE_DIR        = $ChocolateyHabitatIncludeDir
-$env:LIBZMQ_PREFIX              = Split-Path $ChocolateyHabitatLibDir -Parent
+function Get-CargoCommand($toolchain) {
+    "cargo +$toolchain"
+}
 
-Write-RustToolVersion
+If($Features) {
+    $featureString = "--features $Features"
```
christophermaier: Is there special quoting we need to do for multiple features, either here or when invoking this script?

Is there a better way to accept multiple option values rather than (what I'm assuming is) a single string?

build.ps1
```diff
@@ -70,162 +75,163 @@ function Invoke-Configure {
...
-    Write-Host "Downloading cacerts.pem"
-    $current_protocols = [Net.ServicePointManager]::SecurityProtocol
-    try {
-        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
-        Invoke-WebRequest -UseBasicParsing -Uri "http://curl.haxx.se/ca/cacert.pem" -OutFile "$env:TEMP\cacert.pem"
-    }
-    finally {
-        [Net.ServicePointManager]::SecurityProtocol = $current_protocols
-    }
+# On buildkite, the rust binaries will be directly in C:
```
christophermaier: Would it be worthwhile to add a specific Buildkite flag to the script?

I'm just wondering if this would misbehave if a user happened to have Rust installed like we do in Buildkite?

build.ps1
```diff
@@ -19,23 +19,32 @@ param (
...
     # Use a optimized release build
     [switch]$Release,
     # Run a cargo check instead of build
     [switch]$Check,
-    # Add protocol feature
-    [switch]$protocols
+    # Features to pass to cargo
+    [string]$Features,
+    # Options to pass to the cargo test command
+    [string]$TestOptions,
```
christophermaier: Is there any special quoting or anything that needs to be done here for multiple values?

build.ps1
```diff
@@ -19,23 +19,32 @@ param (
...
+    # Use nightly toolchain
+    [switch]$Nightly
```
christophermaier: Minor, but would there be benefit to making a `-Toolchain` option instead to allow you to set arbitrary toolchains?

(I'm really just thinking out loud... what's here now is certainly fine, and it's not like it'd be difficult to change in the future if we needed to.)

build.ps1
mwrock: shouldn't be since it just emits the options as is. For example `-TestOptions "--test-threads=1"`

build.ps1
```diff
@@ -70,162 +71,154 @@ function Invoke-Configure {
...
-    if ((choco list habitat_native_dependencies --local-only) -match '^1 packages installed\.$') {
-        choco upgrade habitat_native_dependencies --confirm -s https://www.myget.org/F/habitat/api/v2  --allowemptychecksums
-    }
-    else {
-        choco install habitat_native_dependencies --confirm -s https://www.myget.org/F/habitat/api/v2  --allowemptychecksums
+    if (!((choco list habitat --local-only) -match '^1 packages installed\.$')) {
+        choco install habitat -y
     }
 
     if(!(Get-Command git -ErrorAction SilentlyContinue)) {
```
scotthain: This is unneeded, we are guaranteed to have the correct version of git in our base images.

mwrock: yeah in the `BuildKite` case that should never be the case. This is here mainly for the clean VM or container scenario.

mwrock:
ok @shain I think this is now pretty well aligned with what we do on linux. I created a `shared.ps1` to mirror `shared.sh`. `build.ps1` now calls `run_clippy` and `run_cargo_tests` for linting and tests. This also includes working tests for the former core components @christophermaier.

mwrock:
ugh more core tests are failing that passed locally. Will investigate

mwrock:
Note this includes some minor test code alterations to get windows tests passing in core.

Issue #6412, opened 2019-04-11
Unable to docker export or enter a docker studio on windows 2019


mwrock:
We may need to add the ability to use alternate base images based on OS. On windows server SKUs, the kernel of the host and container mush match which means we need to use the appropriate version tag according to the host we are on.

stuartpreston:
Could a possible implementation not just relate to the version allow a plan to specify something like ```$pkg_parent_image=mcr.microsoft.com/windows/servercore/iis:windowsservercore-ltsc2019``` rather than be baked into the tooling?  Selecting an appropriate base/parent image would allow smaller layer deltas when the supervisor and source payload are added, for example.

mwrock:
I'm thinking likely an argument passed to `hab pkg export docker` since you may not want a 1 to 1 relationship between plan and base image.

stuartpreston:
Yes, I guess it would need to be that way. :( Implementing a CI for image creation at build time would mean you need to be running the same kernel as your target on your build server so that it can be built with the install hooks so putting it in the plan just opens up more opportunity for unexpected results.

mwrock:
closed by #6643

PR #6470, opened 2019-04-20
add ltsc2016 tag for windowsservercore base images

mwrock:
This week Microsoft stopped providing a `latest` tag for `microsoft/windowsservercore`. This adds the `ltsc2016` tag to the 2 references where we pull this image. This is a temporary stop-gap measure until we tackle #6412 where we need to accomodate the ability to provide version specific studios and allow the docker exporter to pass in a user supplied tag.


components/studio/build-docker-image.ps1
```diff
@@ -1,97 +0,0 @@
-param ([String[]]$InstallHarts = $null)
```
christophermaier: Did you mean to delete this file as part of this commit? It wasn't mentioned in the PR message.

mwrock: I did. Thanks for asking! It was used by the `bintray-publish` that was previously removed. I just added a separate commit for this deletion.

PR #6482, opened 2019-04-26
Run health checks asynchronously

christophermaier:
Previously, health check hooks were run synchronously by the
Supervisor, with logic for managing the scheduling of the next
invocation at some time in the future. Each time through the service's
"tick", we would have to determine whether or not to run a health
check based on this. Not only does this add complexity, but it also
ends up blocking the Supervisor if a health check hook takes any
appreciable amount of time.

Here, we reformulate health checking to happen in an endlessly
repeating future that runs asynchronously. Rather than tracking when
the next time to run a hook is, this is now baked into the logic of
the future itself, giving us less to explicitly manage.

Now, when we start a service, we spawn the health checking future. It
executes the health check, then waits for the appropriate amount of
time; these two steps constitute a single iteration of the overall
health checking process. This iteration is then repeated endlessly.

The future is constructed such that we retain a "handle" to it, which
can be used to terminate the future when, e.g., the service shuts
down. This handle is the only explicit "state" that the Service itself
needs to keep track of, and even then, all it really has to do is just
maintain ownership of it.

In order to make this approach work with how Services are currently
constructed, we needed to wrap the `HealthCheckHook` itself in an
`Arc`, as well as the service's process supervisor and health check
result fields. As we continue to move toward a more asynchronous
internal architecture, we hope to be able to eliminate this shared
state.

tenor-203861434

### DIFF
--- a/components/sup/src/manager/service/health.rs
+++ b/components/sup/src/manager/service/health.rs
@@ -121,7 +121,7 @@ impl State {
 
         // Use an Arc to avoid having to have full clones everywhere. :/
         let service_group = Arc::new(service_group);
-        let service_group_copy = service_group.clone();
+        let service_group_ref = Arc::clone(&service_group);
 
         if let Some(hook) = hook {
             let hr = hook_runner::HookRunner::new(hook,
@@ -140,7 +140,7 @@ impl State {
             Either::B(lazy(move || Ok(status)))
         }.map_err(move |e| {
              error!("Error running health check hook for {}: {:?}",
-                    service_group_copy, e)
+                    service_group_ref, e)
          })
          .and_then(move |check_result| {
              debug!("Caching HealthCheck = '{}' for '{}'",
--- a/components/sup/src/manager/service/health.rs
+++ b/components/sup/src/manager/service/health.rs
@@ -193,19 +193,16 @@ impl State {
             // it.
             let service_group = state.service_group.clone();
             state.clone().single_iteration().then(move |res| {
-                                                match res {
-                                                    Ok(_) => {
-                                                        trace!("Health check future for {} \
-                                                                succeeded; continuing loop",
-                                                               service_group);
-                                                        Ok(Loop::Continue(state))
-                                                    }
-                                                    Err(_) => {
-                                                        trace!("Health check future for {} \
-                                                                failed failed; continuing loop.",
-                                                               service_group);
-                                                        Ok(Loop::Continue(state))
-                                                    }
+                                                if res.is_ok() {
+                                                    trace!("Health check future for {} \
+                                                            succeeded; continuing loop",
+                                                           service_group);
+                                                    Ok(Loop::Continue(state))
+                                                } else {
+                                                    trace!("Health check future for {} failed \
+                                                            failed; continuing loop.",
+                                                           service_group);
+                                                    Ok(Loop::Continue(state))
                                                 }
                                             })
         })
--- a/components/sup/src/manager/service/health.rs
+++ b/components/sup/src/manager/service/health.rs
@@ -1,5 +1,24 @@
-use std::fmt;
+use crate::manager::{service::{hook_runner,
+                               hooks::HealthCheckHook,
+                               supervisor::Supervisor},
+                     GatewayState};
+use futures::{future::{self,
+                       lazy,
+                       Either,
+                       Future,
+                       Loop},
+              IntoFuture};
+use habitat_common::templating::package::Pkg;
+use habitat_core::service::{HealthCheckInterval,
+                            ServiceGroup};
+use std::{fmt,
+          ops::Deref,
+          sync::{Arc,
+                 Mutex,
+                 RwLock},
+          time::Instant};
 
+/// The possible results of running a health check hook.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize)]
 pub enum HealthCheck {
     Ok,
@@ -12,6 +31,7 @@ impl Default for HealthCheck {
     fn default() -> HealthCheck { HealthCheck::Unknown }
 }
 
+/// Convert health check hook exit codes into `HealthCheck` statuses.
 impl From<i8> for HealthCheck {
     fn from(value: i8) -> HealthCheck {
         match value {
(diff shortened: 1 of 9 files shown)


components/sup/src/manager/service/health.rs
```diff
@@ -1,19 +1,24 @@
...
+use habitat_core::service::{HealthCheckInterval,
+                            ServiceGroup};
+use std::{fmt,
+          ops::Deref,
+          sync::{Arc,
+                 Mutex,
+                 RwLock},
+          time::Instant};
 
+/// The possible results of running a health check hook.
```
baumanj: Maybe this implies that `HealthCheck` isn't really the best name. Perhaps `ServiceHealth` or `HealthCheckResponse` would be better.

components/sup/src/manager/service/health.rs
```diff
@@ -40,12 +46,177 @@ impl From<i8> for HealthCheck {
...
+
+    /// A reference to the process supervisor for the service. This is
+    /// used to create a "proxy health check" for services that do not
+    /// provide their own health check hook.
+    supervisor: Arc<Mutex<Supervisor>>,
+
+    /// The configured interval at which to run health checks for this
+    /// service. The interval actually used may differ based on the
+    /// previous health check result.
+    configured_interval: HealthCheckInterval,
```
baumanj:
```suggestion
    nominal_interval: HealthCheckInterval,
```
May be a slightly more illustrative name. To me, "configured" implies that it's been explicitly set, whereas this will usually just be the default value of 30 seconds. Nominal focuses on the fact that it's the interval applied in *normal* operation, and that erroneous circumstances may use a different one.

components/sup/src/manager/service/health.rs
```diff
@@ -40,12 +46,177 @@ impl From<i8> for HealthCheck {
...
+    /// Creates a future that runs the health check and then waits for
+    /// a suitable interval. Multiple such iterations will then be
+    /// chained together for an unending stream of health checks.
+    fn single_iteration(self) -> impl Future<Item = (), Error = ()> {
+        let State { hook,
+                    service_group,
+                    package,
+                    svc_encrypted_password,
+                    supervisor,
+
```
baumanj: Why a blank line?

baumanj:
Is this ready for re-review?

christophermaier:
@baumanj Yeah, should be good to go... There are still some `TODO` comments, but they're about pre-existing aspects of health check behavior that I have bigger questions about. I'll file issues for them and remove the comments. They should be addressed, but are beyond the scope of this change, I feel. Don't bother commenting on those 😉

components/sup/src/manager/service/health.rs
christophermaier: That is a _fabulous_ question 🤔

components/sup/src/manager/service/health.rs
christophermaier: I like it!

components/sup/src/manager/service/health.rs
christophermaier: Indeed... I'll see what the implications of such a rename would be.

components/sup/src/manager/service/health.rs
```diff
@@ -26,12 +46,167 @@ impl From<i8> for HealthCheck {
...
+                    package,
+                    svc_encrypted_password,
+                    supervisor,
+                    nominal_interval,
+                    service_health_result,
+                    gateway_state, } = self;
+
+        // Use an Arc to avoid having to have full clones everywhere. :/
+        let service_group = Arc::new(service_group);
+        let service_group_copy = service_group.clone();
```
baumanj:
```suggestion
        let service_group_ref = Arc::clone(&service_group);
```
Use [the preferred idiom](http://doc.rust-lang.org/1.34.1/std/sync/struct.Arc.html#cloning-references), plus I think `_copy` is misleading since we're explicitly trying to avoid a copy.

components/sup/src/manager/service/health.rs
```diff
@@ -26,12 +46,167 @@ impl From<i8> for HealthCheck {
...
+                                                               service_group);
+                                                        Ok(Loop::Continue(state))
+                                                    }
+                                                    Err(_) => {
+                                                        trace!("Health check future for {} \
+                                                                failed failed; continuing loop.",
+                                                               service_group);
+                                                        Ok(Loop::Continue(state))
+                                                    }
+                                                }
```
baumanj: Since this is already pretty extremely indented and we're not doing anything with the *contents* of the `Result` variants, we could simplify:
```rust
                                                if res.is_ok() {
                                                    trace!("Health check future for {} \
                                                            succeeded; continuing loop",
                                                           service_group);
                                                    Ok(Loop::Continue(state))
                                                } else {
                                                    trace!("Health check future for {} failed \
                                                            failed; continuing loop.",
                                                           service_group);
                                                    Ok(Loop::Continue(state))
                                                }
```

components/sup/src/sup_futures.rs
```diff
@@ -0,0 +1,48 @@
...
+/// A handle attached to a future generated using
+/// `cancelable_future`. That future will be dropped when this handle
+/// is dropped.
+#[derive(Debug)]
+pub struct FutureHandle(oneshot::Sender<()>);
+
+impl FutureHandle {
+    /// A more explicit way to cancel a future instead of relying on
+    /// dropping the handle.
+    pub fn terminate(self) { let _ = self.0.send(()); }
```
baumanj: I find this to be a non-obvious way of ignoring the result, the shortest, simplest idiom seems to be
```suggestion
    pub fn terminate(self) { self.0.send(()).ok(); }
```
or just be explicit and throw in `#[allow(unused_must_use)]` since the function has only one call.

components/sup/src/manager/service/health.rs
```diff
@@ -26,12 +46,167 @@ impl From<i8> for HealthCheck {
...
-            HealthCheck::Critical => "CRITICAL",
-            HealthCheck::Unknown => "UNKNOWN",
-        };
-        write!(f, "{}", msg)
+        match *self {
+            HealthCheck::Ok => "OK".fmt(f),
+            HealthCheck::Warning => "WARNING".fmt(f),
+            HealthCheck::Critical => "CRITICAL".fmt(f),
+            HealthCheck::Unknown => "UNKNOWN".fmt(f),
+        }
```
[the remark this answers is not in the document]
christophermaier: It struck me as a bit cleaner, but I'm happy to back it out, given that it's not essential to anything else in this PR.

components/sup/src/manager/service/mod.rs
```diff
@@ -222,7 +222,7 @@ impl Service {
                      bldr_url: spec.bldr_url,
                      channel: spec.channel,
                      desired_state: spec.desired_state,
-                     health_check_result: Arc::new(Mutex::new(HealthCheck::default())),
+                     health_check_result: Default::default(),
```
christophermaier: Check it out @baumanj ☝️

components/sup/src/manager/service/health.rs
christophermaier: All good!

components/sup/src/manager/service/health.rs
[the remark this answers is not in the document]
baumanj: Not a big deal either way; I just wanted to be sure I wasn't missing some subtlety. I like my code like I like my martinis…

PR #6493, opened 2019-04-30
Clean up some workarounds to the release process

scotthain:
This removes some workarounds regarding promotion of artifacts with different package targets. This requires 0.80.0.

### DIFF
--- a/.buildkite/scripts/build_component.sh
+++ b/.buildkite/scripts/build_component.sh
@@ -47,10 +47,6 @@ ${hab_binary} pkg upload \
     --auth="${HAB_AUTH_TOKEN}" \
     "results/${pkg_artifact}"
 
-${hab_binary} pkg promote \
-    --auth="${HAB_AUTH_TOKEN}" \
-    "${pkg_ident}" "${channel}" "${pkg_target}"
-
 echo "--- :writing_hand: Recording Build Metadata"
 case "${component}" in
     "hab")
--- a/.buildkite/scripts/build_component.sh
+++ b/.buildkite/scripts/build_component.sh
@@ -41,11 +41,18 @@ else
 fi
 source results/last_build.env
 
-echo "--- :habicat: Uploading ${pkg_ident} to Builder in the '${channel}' channel"
+# TODO (SM): The 0.59.0 hab cli that we rely on for x86_64-linux builds 
+# doesn't emit pkg_target. Until we've sufficiently bootstrapped ourselves
+# we need to set it. This can be removed when studio-ci-common pulls 0.63.0 
+# or newer. This is safe to do because the x86_64-linux-kernel2 builds will
+# already have this value set.
+: "${pkg_target:=x86_64-linux}"
+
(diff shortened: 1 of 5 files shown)


.buildkite/scripts/build_component.sh
```diff
@@ -41,37 +41,15 @@ else
...
-
-promote "${pkg_ident}" "${pkg_target}" "${channel}"
-set_target_metadata "${pkg_ident}" "${pkg_target}"
+echo "--- :habicat: Uploading ${pkg_ident} to Builder in the '${channel}' channel"
+${hab_binary} pkg upload \
+    --channel="${channel}" \
+    --auth="${HAB_AUTH_TOKEN}" \
+    "results/${pkg_artifact}"
+
+${hab_binary} pkg promote \
```
smacfarlane: I know this is an uncomment of previous behavior,  but I thought `hab pkg upload --channel` promoted the package to the channel on upload.  Is there a reason for this promote that I'm missing context on?

.buildkite/scripts/build_component.sh
```diff
@@ -41,37 +41,15 @@ else
...
-     --header "Content-Type: application/octet-stream" \
-     --header "Authorization: Bearer $HAB_AUTH_TOKEN" \
-     --data-binary "@results/${pkg_artifact:?}" \
-     --fail \
-     --verbose \
-    "https://bldr.habitat.sh/v1/depot/pkgs/${pkg_ident}?checksum=${pkg_blake2bsum:?}&target=${pkg_target}"
-
-promote "${pkg_ident}" "${pkg_target}" "${channel}"
-set_target_metadata "${pkg_ident}" "${pkg_target}"
+echo "--- :habicat: Uploading ${pkg_ident} to Builder in the '${channel}' channel"
```
smacfarlane: should be `"--- :habicat: :partyparrot:`

.buildkite/scripts/build_component.sh
christophermaier: We had to do them separately when using `curl`, but before that workaround, we uploaded straight to the channel using `hab`.

I can't think of a reason to not go back to that.

scotthain: Fixed!

.buildkite/scripts/build_component.sh
```diff
@@ -41,37 +41,11 @@ else
...
+${hab_binary} pkg upload \
+    --channel="${channel}" \
+    --auth="${HAB_AUTH_TOKEN}" \
+    "results/${pkg_artifact}"
```
christophermaier: We should double-check that this behaves properly with respect to package target before getting this merged. We can do a "fake release", as detailed in https://github.com/habitat-sh/habitat/pull/6510.

baumanj:
What's the next step here?

scotthain:
Needs to be updated - I ran it through a fake release and it had some issues. I'll resolve them.

smacfarlane:
This work has been incorporated into #6601.  Thanks @scotthain!

Issue #6267, opened 2019-03-05
Output a full plan.sh when initializing a new project


dmccown:
## Description
When creating a new plan with `hab plan init` we lay down a base plan file. Suppose you are building a Java project and run your first build. You'll start seeing error related to building a C project. You'll then look at your plan file and have no idea what build step you need to override or where that C related command is even coming from. You'll need to find our source and then see the specific step you need to override.

 It would be really helpful to do something like [default_build_call_backs.sh · GitHub](https://gist.github.com/ericcalabretta/a2609425ae9e51c3deba7687f4fa2715) which in-lines all the default behaviors, so you can quickly determine what steps to override and when without referring to source. 

We should also link directly to documentation where appropriate.

## Acceptance Criteria
1. Run `hab plan init` to create a new project
2. Verify the plan is complete with a default implementation, so it's easy to update.

davidMcneil:
It looks like there are currently options to `init` to output a complete plan `--with-all`, `--with-docs`, `--with-callbacks`. `--with-all` was broken so I am adding a pull request to fix that. These options do not list the implementation of the `do_default_*` functions but the output does give a [link](https://github.com/habitat-sh/habitat/tree/master/components/plan-build/bin/hab-plan-build.sh) to the implementations. Is that sufficient?

dmccown:
If possible it would be great to get the `do_default_*` implementations in-lined in the initial plan. That would save users from having to look up the default implementation. @christophermaier any thoughts on this?

I'm open  if this would take a couple of steps to achieve as well. E.g. first cut would be doing as you described and from there we could look at in-lining these functions.

PR #6495, opened 2019-04-30
Streamline plan initialization

davidMcneil:
The `--with-all` would not output the docs or the callbacks.


baumanj:
It helps to find the original issue (and automatically closes it when the PR merges) if you add

Resolves https://github.com/habitat-sh/habitat/issues/6267

to the PR. See https://help.github.com/en/articles/closing-issues-using-keywords for more.

davidMcneil:
@baumanj Thanks for the feedback. I agree that what I did here is probably not what we want. I like the idea of making a trait and adding tests, and agree that the ux could be rethought (I was certainly confused as demonstrated by this pull request). I think the first step would be to determine all the permutations of initial plan we want to support. Would this be the same as what is currently possible? So there are three variables `use full template`, `insert callbacks`, and `insert docs` for a total of 8 possibilities.

baumanj:
> I was certainly confused as demonstrated by this pull request

You are not alone there!

> I think the first step would be to determine all the permutations of initial plan we want to support

Agreed

> Would this be the same as what is currently possible? So there are three variables `use full template`, `insert callbacks`, and `insert docs` for a total of 8 possibilities

16 if you also consider `--windows`.

As for a path forward, I think we could stand to simplify this dramatically. Advanced users can generate the plans themselves from scratch, so I see `hab plan init` as a beginner-oriented command to show people all what's available. Given that, I think we should keep it as simple as possible and always generate the full template will all callbacks and docs. Additionally, I don't really see too much value in having a `--windows` argument, when we can bake that behavior into the binary itself through conditional compilation. If there's really a use case for generating a `plan.ps1` from a linux `hab` binary, we can always add it back later (along with a `--linux` and probably `--linux2` option for doing the reverse).

Removing all this may simplify things enough that there's not really anything worth testing.

davidMcneil:
> we should keep it as simple as possible and always generate the full template will all callbacks and docs

That sounds great! Is it ok to remove the command line flags `--windows`, `--with-all`, `--with-docs`, and `--with-callbacks` and break backwards compatibility?

In [#6267](https://github.com/habitat-sh/habitat/issues/6267#issuecomment-488107999) @dmccown mentioned in-lining the default implementation of the callbacks. What is the history of the default callbacks? Is this as straightforward as copying the implementations from [hab-plan-build.sh](https://github.com/habitat-sh/habitat/tree/master/components/plan-build/bin/hab-plan-build.sh) and putting them in the [full_template_plan.sh](https://github.com/habitat-sh/habitat/blob/master/components/hab/static/default_template_plan.sh)?

baumanj:
> That sounds great! Is it ok to remove the command line flags `--windows`, `--with-all`, `--with-docs`, and `--with-callbacks` and break backwards compatibility?

Since this isn't the sort of command that can imagine being used in automation, I think so. Plus we're technically not 1.0 yet, so it's legal from a semver perspective.

As for in-lining the default callbacks, I'm less certain about that. I wouldn't want to just copy/paste them since that would lead to a confusing inconsistency if we ever changed the default implementations. One option would be to add a link to the implementation like:
```bash
do_download() {
    # See https://github.com/habitat-sh/habitat/blob/0.79.1/components/plan-build/bin/hab-plan-build.sh
    do_default_download
}
```
If all the default callback implementations are in that one file, perhaps there could just be one link. The only other complexity I can think of would be making sure the appropriate version number is injected.

davidMcneil:
How does the `--scaffolding` flag fit into all of this? Now that I see this flag, I realize this is a reason we cannot just copy the `do_default_*` implementations into the `plan.sh`. These scaffolding options override the `do_default_*` implementations.

baumanj:
> How does the `--scaffolding` flag fit into all of this?

I don't know much about scaffolding. @eeyun or @smacfarlane are more knowledgeable there.

> I realize this is a reason we cannot just copy the do_default_* implementations into the plan.sh

Yeah, I don't think that's the right approach anyway for reasons I mentioned in https://github.com/habitat-sh/habitat/pull/6495#issuecomment-488457334

smacfarlane:
A couple bits of feedback:    As an experienced plan author, it would be nice to have a way to initialize a plan with the minimum amount of generated code.  With this change, the first thing I would do is to delete the majority of the output. Even though I know what that minimum required metadata is, I still use `hab plan init` as it's much quicker to get started than generating a plan from scratch, and it gives consistent ordering of plan metadata.   I'm not sure what the correct default behavior should be as that's going to depend on the experience level of the user. 

The output when passing `--scaffolding` to it I suspect will also need some additional work.  The available scaffolding packages override a lot of the build phase functions (`do_build`, etc.) and having `plan init` output those functions as well will likely cause the scaffolding to not function.  I'd have to re-read the scaffolding load code to be certain, but that is my suspicion.   

This is a take-it-or-leave-it suggestion.  It might be better to comment out the build phase functions in the output to treat them as examples.  I'm certainly biased in my view, but it feels awkward to have the default for a plan be

```
do_build() {
  do_default_build
}
```

but there's also no harm in that.

components/hab/src/command/plan/init.rs
```diff
@@ -107,19 +92,11 @@ pub fn start(ui: &mut UI,
...
-        }
-    } else if with_all || scaffold.is_none() {
-        let rendered_plan = handlebars.template_render(FULL_PLAN_TEMPLATE, &data)?;
-        create_with_template(ui, &format!("{}/plan.sh", root), &rendered_plan)?;
+    if cfg!(windows) {
+        let rendered_plan = handlebars.template_render(FULL_PLAN_PS1_TEMPLATE, &data)?;
+        create_with_template(ui, &format!("{}/plan.ps1", root), &rendered_plan)?;
     } else {
-        let rendered_plan = handlebars.template_render(DEFAULT_PLAN_TEMPLATE, &data)?;
+        let rendered_plan = handlebars.template_render(FULL_PLAN_TEMPLATE, &data)?;
```
christophermaier: Wonder if we should take this opportunity to rename this to `FULL_PLAN_BASH_TEMPLATE` instead.

components/hab/src/command/plan/init.rs
```diff
@@ -107,19 +92,11 @@ pub fn start(ui: &mut UI,
...
-            let rendered_plan = handlebars.template_render(FULL_PLAN_PS1_TEMPLATE, &data)?;
-            create_with_template(ui, &format!("{}/plan.ps1", root), &rendered_plan)?;
-        } else {
-            let rendered_plan = handlebars.template_render(DEFAULT_PLAN_PS1_TEMPLATE, &data)?;
-            create_with_template(ui, &format!("{}/plan.ps1", root), &rendered_plan)?;
-        }
-    } else if with_all || scaffold.is_none() {
-        let rendered_plan = handlebars.template_render(FULL_PLAN_TEMPLATE, &data)?;
-        create_with_template(ui, &format!("{}/plan.sh", root), &rendered_plan)?;
+    if cfg!(windows) {
```
christophermaier: This means that we can only create Windows plans on Windows machines, and Linux plans on Linux machines, corret? That is probably a bit too strict. There may not be a lot of Linux users that want to make Windows plans, but our (enterprise) Windows users often also have Linux services they run.

Granted, those same Windows users could easily drop into WSL for most of their Linux needs. 

If it's not too difficult, it'd be nice to have a way for users to create whatever kind of plan they need to, regardless of whatever platform they happen to currently be on. It should probably default to the platform they're currently on, but still be modifiable.

That being said, I'm happy to leave that for a subsequent PR until we can really nail down the UX of this work.

christophermaier:
As for embedding the default implementations into the template, I think that could get messy real quick, particularly if those change in the future. I like @baumanj's suggestion of linking.

components/hab/src/command/plan/init.rs
davidMcneil: That is correct. There used to be a `--windows` flag that forced a powershell template, but I removed that. I made a separate issue (#6516) to discuss the UX. If we come to a decision on that I can go ahead and integrate it into this PR.

davidMcneil:
Yes, embedding the default implementations would be a mess and it breaks scaffolding. There is currently one [reference](https://github.com/habitat-sh/habitat/blob/a82f16409792579559475f7e199886237ebed3cb/components/hab/static/full_template_plan.sh#L254) to where the `do_default_*` implementations are. Do we want more than that?

components/hab/src/command/plan/init.rs
baumanj: I'm inclined to take a [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) approach on this one. It's always easy to add this later, but much harder to remove support once we've added it. I'm willing to bet the development of a plan for an arch other than the one currently running is more of an imagined use case than a reality.

baumanj:
The only other thought I'd have here is that if possible, it might be nice to have just one plan template per platform rather than a full and minimal version to reduce duplication. Are there any differences that can't be expressed via the handlebars conditionals we have available?

If that seems like a big headache though, I don't have particularly strong feelings on the matter.

components/hab/src/command/plan/init.rs
christophermaier: @davidMcneil Ah, cool, you've got a separate issue.

Let's keep the current PR as it is, and if that issue attracts any additional input, we can revisit later.

davidMcneil:
Initially, I started down the path of just one template, but I felt like all the added conditionals made it too hard to manage. I am happy to put up a version with a single template to compare against.

baumanj:
> Initially, I started down the path of just one template, but I felt like all the added conditionals made it too hard to manage. I am happy to put up a version with a single template to compare against.

I'd say they're already pretty hard to visually parse, so I'd be in favor of either

1. One template per platform, with gnarly conditionals that we only have to get right once
1. 2ⁿ  templates per platform corresponding to the _n_ switches (just `--minimal`, currently) which determine the output and zero handlebars templating for maximum readability (and code simplicity).

PR #6520, opened 2019-05-07
Add FQDN to EventCore Metadata

afiune:
This is the first end-to-end iteration of the JIRA Task [#A2-762](https://chefio.atlassian.net/browse/A2-762) 
-- **Implement FQDN Detection in Habitat** 

From the [`hostname` man page](https://www.systutorials.com/docs/linux/man/1-hostname/):
>The FQDN (Fully Qualified Domain Name) of the system is the
name that the resolver(3) returns for the host name, such as,
**ursula.example.com**.  It is usually the hostname followed by the
DNS domain name (the part  after the first dot).  You can check
the FQDN using `hostname --fqdn` or the domain name using
`dnsdomainname`.

As the man page says it, the FQDN is the combination of the
Hostname and the Domain Name. But technically speaking:
>The FQDN is the name getaddrinfo(3) returns for the
host name returned by gethostname(2). The DNS domain name
is the part after the first dot.

This change is adding two new `core::os::net` functions:
- `lookup_fqdn(hostname)`: returns a vector of fqdn that resolves
the provided hostname.
- `fqdn()`: returns the FQDN (fully qualified domain name) of the
running machine.

These functions are leveraging the [dns-lookup crate](https://docs.rs/dns-lookup/1.0.0/dns_lookup/index.html) that implements
the C function called `getaddrinfo()` that we technically need to get
FQDN of the running machine. 

Habitat already have a function that returns the hostname of the
machine[ called `hostname()`](https://github.com/habitat-sh/habitat/blob/master/components/core/src/os/net/windows.rs#L22), we are leveraging that function inside
the new `fqdn()` one.

I have created an issue https://github.com/habitat-sh/habitat/issues/6531 to explore the possibility of refactoring the
existing `hostname()` function in favor of `dns_lookup::lookup_host()`

Signed-off-by: Salim Afiune <afiune@chef.io>

### DIFF
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -34,7 +34,7 @@ use std::io;
 // Since the underlying crate is platform agnostic, this function is as well,
 // we only implement a single function `ai_canonname()` that returns the right
 // hint flag for the running operating system.
-pub fn lookup_fqdn(hostname: &str) -> io::Result<Vec<String>> {
+pub fn lookup_fqdn(hostname: &str) -> io::Result<String> {
     #[cfg(not(windows))]
     let flags = libc::AI_CANONNAME;
 
@@ -45,11 +45,15 @@ pub fn lookup_fqdn(hostname: &str) -> io::Result<Vec<String>> {
     let addrinfos =
         getaddrinfo(Some(hostname), None, Some(hints))?.collect::<std::io::Result<Vec<_>>>()?;
 
-    let canonnames = addrinfos.into_iter()
-                              .filter_map(|info| info.canonname)
-                              .collect();
+    // If 'hints.ai_flags' includes the AI_CANONNAME flag, then the ai_canonname
+    // field of the first of the addrinfo structures in the returned list is set
+    // to point to the official name of the host.
+    if addrinfos.len() > 0 {
+        let addrinfo = addrinfos[0].clone();
+        return Ok(addrinfo.canonname.unwrap_or(hostname.to_string()))
+    }
 
-    Ok(canonnames)
+    Ok(hostname.to_string())
 }
 
 // fqdn returns the fully qualified domain name of the running machine
@@ -58,7 +62,7 @@ pub fn fqdn() -> Option<String> {
     match hostname() {
         Ok(hostname) => {
             match lookup_fqdn(&hostname) {
-                Ok(fqdns) => fqdns.into_iter().find(|ref h| h.contains(&hostname)),
+                Ok(fqdn) => Some(fqdn),
                 // @afiune if the lookup_fqdn returns an Err(), should we
                 // return the hostname instead of None?
                 Err(_) => None,
@@ -84,7 +88,7 @@ fn test_fqdn_lookup() {
     let fqdn = lookup_fqdn("localhost");
     assert!(fqdn.is_ok());
     assert_eq!(fqdn.unwrap(),
-               vec![String::from("localhost")],
+               String::from("localhost"),
                "the fqdn of localhost should be localhost");
 }
 
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -35,8 +35,13 @@ use std::io;
 // we only implement a single function `ai_canonname()` that returns the right
 // hint flag for the running operating system.
 pub fn lookup_fqdn(hostname: &str) -> io::Result<Vec<String>> {
-    let hints = AddrInfoHints { flags: ai_canonname(),
-                                ..AddrInfoHints::default() };
+    #[cfg(not(windows))]
+    let flags = libc::AI_CANONNAME;
+
+    #[cfg(windows)]
+    let flags = winapi::shared::ws2def::AI_CANONNAME;
+
+    let hints = AddrInfoHints { flags, ..AddrInfoHints::default() };
     let addrinfos =
         getaddrinfo(Some(hostname), None, Some(hints))?.collect::<std::io::Result<Vec<_>>>()?;
 
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -29,10 +29,14 @@ use dns_lookup::{getaddrinfo,
                  AddrInfoHints};
 use std::io;
 
-// lookup_fqdn returns a vector of fqdn that resolves the provided hostname
+// lookup_fqdn returns a vector of fqdn that resolves the provided hostname.
+//
+// Since the underlying crate is platform agnostic, this function is as well,
+// we only implement a single function `ai_canonname()` that returns the right
+// hint flag for the running operating system.
 pub fn lookup_fqdn(hostname: String) -> io::Result<Vec<String>> {
   let hints = AddrInfoHints {
-      flags: libc::AI_CANONNAME,
+      flags: ai_canonname(),
       .. AddrInfoHints::default()
   };
   let addrinfos =
@@ -81,6 +85,7 @@ fn test_fqdn() {
         "should match with the configured fqdn in the running machine");
 }
 
+#[cfg(not(windows))]
 #[test]
 fn test_fqdn_lookup() {
     let fqdn = lookup_fqdn(String::from("localhost"));
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -59,16 +59,8 @@ pub fn lookup_fqdn(hostname: &str) -> io::Result<String> {
 
 // fqdn returns the fully qualified domain name of the running machine
 pub fn fqdn() -> Option<String> {
-    // Implementation 1 - Using match statements
     match hostname() {
-        Ok(hostname) => {
-            match lookup_fqdn(&hostname) {
-                Ok(fqdn) => Some(fqdn),
-                // @afiune if the lookup_fqdn returns an Err(), should we
-                // return the hostname instead of None?
-                Err(_) => None,
-            }
-        }
+        Ok(hostname) => lookup_fqdn(&hostname).ok(),
         Err(_) => None,
     }
 }
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -22,3 +22,79 @@ mod imp;
 mod imp;
 
 pub use self::imp::*;
+
+extern crate dns_lookup;
+
+use dns_lookup::{getaddrinfo,
+                 AddrInfoHints};
+use std::io;
+
+// lookup_fqdn returns a vector of fqdn that resolves the provided hostname
+pub fn lookup_fqdn(hostname: String) -> io::Result<Vec<String>> {
+  let hints = AddrInfoHints {
+      flags: libc::AI_CANONNAME,
+      .. AddrInfoHints::default()
+  };
+  let addrinfos =
+      getaddrinfo(Some(&hostname), None, Some(hints))?
+      .collect::<std::io::Result<Vec<_>>>()?;
+
+  let canonnames = addrinfos
+      .into_iter()
+      .filter_map(|info| info.canonname)
+      .collect();
+
+  Ok(canonnames)
+}
+
+// fqdn returns the fully qualified domain name of the running machine
+pub fn fqdn() -> Option<String> {
+    // Implementation 1 - Using match statements
+    match hostname() {
+        Ok(hostname) => {
+            // we clone the hostname since we need it to find the
+            // right fqdn that the lookup fn returns
+            let host_to_lookup = hostname.clone();
+
+            match lookup_fqdn(host_to_lookup) {
+                Ok(fqdns) => {
+                    fqdns.into_iter()
+                        .find(| ref h| h.contains(&hostname))
+                }
+                // @afiune if the lookup_fqdn returns an Err(), should we
+                // return the hostname instead of None?
+                Err(_) => None
+            }
+        },
+        Err(_) => None,
+    }
+}
+
+#[test]
+#[ignore]
+fn test_fqdn() {
+    // @afiune This test is ignore because it is testing the actual
+    // fqdn of the running machine, mine has 'afiune-ubuntu-vb.lala.com'
+    assert_eq!(
+        fqdn().unwrap(),
+        String::from("afiune-ubuntu-vb.lala.com"),
+        "should match with the configured fqdn in the running machine");
+}
+
+#[test]
+fn test_fqdn_lookup() {
+    let fqdn = lookup_fqdn(String::from("localhost"));
+    assert!(fqdn.is_ok());
+    assert_eq!(fqdn.unwrap(),
+       vec![String::from("localhost")],
+       "the fqdn of localhost should be localhost");
+}
+
+#[test]
+fn test_fqdn_lookup_err() {
+    let fqdn = lookup_fqdn(String::from(""));
+    assert!(fqdn.is_err(), "Should be an Err()");
+    assert_eq!(
+        format!("{}",fqdn.unwrap_err()),
+        "failed to lookup address information: Name or service not known");
+}
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -32,8 +32,8 @@ use std::io;
 // lookup_fqdn returns a vector of fqdn that resolves the provided hostname.
 //
 // Since the underlying crate is platform agnostic, this function is as well,
-// we only implement a single function `ai_canonname()` that returns the right
-// hint flag for the running operating system.
+// we only implement a single variable `flags` that has the right hint for
+// the running operating system.
 pub fn lookup_fqdn(hostname: &str) -> io::Result<String> {
     #[cfg(not(windows))]
     let flags = libc::AI_CANONNAME;
@@ -41,16 +41,17 @@ pub fn lookup_fqdn(hostname: &str) -> io::Result<String> {
     #[cfg(windows)]
     let flags = winapi::shared::ws2def::AI_CANONNAME;
 
-    let hints = AddrInfoHints { flags, ..AddrInfoHints::default() };
+    let hints = AddrInfoHints { flags,
+                                ..AddrInfoHints::default() };
     let addrinfos =
         getaddrinfo(Some(hostname), None, Some(hints))?.collect::<std::io::Result<Vec<_>>>()?;
 
     // If 'hints.ai_flags' includes the AI_CANONNAME flag, then the ai_canonname
     // field of the first of the addrinfo structures in the returned list is set
     // to point to the official name of the host.
-    if addrinfos.len() > 0 {
+    if !addrinfos.is_empty() {
         let addrinfo = addrinfos[0].clone();
-        return Ok(addrinfo.canonname.unwrap_or(hostname.to_string()))
+        return Ok(addrinfo.canonname.unwrap_or_else(|| hostname.to_string()));
     }
 
     Ok(hostname.to_string())
--- a/components/core/src/os/net/mod.rs
+++ b/components/core/src/os/net/mod.rs
@@ -34,11 +34,11 @@ use std::io;
 // Since the underlying crate is platform agnostic, this function is as well,
 // we only implement a single function `ai_canonname()` that returns the right
 // hint flag for the running operating system.
-pub fn lookup_fqdn(hostname: String) -> io::Result<Vec<String>> {
-    let hints = AddrInfoHints { flags: flags: ai_canonname(),
+pub fn lookup_fqdn(hostname: &str) -> io::Result<Vec<String>> {
+    let hints = AddrInfoHints { flags: ai_canonname(),
                                 ..AddrInfoHints::default() };
     let addrinfos =
-        getaddrinfo(Some(&hostname), None, Some(hints))?.collect::<std::io::Result<Vec<_>>>()?;
+        getaddrinfo(Some(hostname), None, Some(hints))?.collect::<std::io::Result<Vec<_>>>()?;
 
     let canonnames = addrinfos.into_iter()
                               .filter_map(|info| info.canonname)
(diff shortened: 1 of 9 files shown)


components/sup/protocols/event.proto
```diff
@@ -14,6 +14,7 @@ message EventMetadata {
   string application = 4;
   string environment = 5;
   map<string, string> meta = 6;
+  string fqdn = 7;
```
baumanj: I'd prefer
```suggestion
  string fully_qualified_domain_name = 7;
```
and a comment describing which fully qualified domain name this is and when it's resolved/updated

components/sup/src/manager/mod.rs
```diff
@@ -494,6 +494,9 @@ impl Manager {
             let es_config =
                 cfg.event_stream_config
                    .expect("Config should be present if the EventStream feature is enabled");
+            // TODO: @afiune We could use a public crate to gather the hostname
```
baumanj: I'm not a fan of TODOs in checked-in code. If it's worth doing, file and issue (and optionally reference it here), but TODOs that live only in code tend to either be fixed right away or stick around forever and don't get included in work prioritization and planning.

afiune: Agree! 💯 I am working on it as we type 😸

components/sup/protocols/event.proto
afiune: I agree that fqdn might not have a true meaning and would be good to use the full name as you suggest, but also now I am wondering if instead of FQDN we should just pass it as `hostname`. Thoughts @christophermaier @baumanj ??

components/sup/src/manager/mod.rs
afiune: If we can't implement the crate I mentioned, I will definitely remove this TODO. Thanks @baumanj

components/sup/protocols/event.proto
sdelano: FQDN is a common reference to Fully Qualified Domain Name and is especially common to the various products we ship at Chef. It's even referred to as such in the `man hostname`:

```
   THE FQDN
       The FQDN (Fully Qualified Domain Name) of the system is the name that the resolver(3) returns for the host name, such as, ursula.example.com.  It is usually the hostname followed by the DNS domain name (the part  after the first dot).  You can check the FQDN using hostname --fqdn or the domain name using dnsdomainname.
```

I'm definitely for leaving this as is.

sdelano: Additionally, `hostname` and `fqdn` are not the same, so they can't be used interchangeably.

baumanj: I agree that `hostname` and `fqdn` aren't interchangeable. And for anyone that knows what `fqdn` means without any additional context, `fully_qulaified_domain_name` should be just as clear, but for those who don't, the latter is a much more informative name than the former. I'm well aware of what an FQDN is myself, but the goal is to make the code as easy to understand and reason about for as many people as possible, right? The tradeoff here is clarity versus concision. If it were a local variable, I'd be more inclined to `fqdn`, but I prefer more explicitness in field names.

As for whether what we want to store here is the FQDN or just the hostname, that's a question of what the needs are that we're trying to satisfy with this change, so I don't really know. Perhaps, @christophermaier does.

components/sup/src/manager/mod.rs
baumanj: We should almost certainly be able to use https://crates.io/crates/hostname. I'd file an issue, with a comment referencing it at https://github.com/habitat-sh/habitat/blob/master/components/core/src/os/net/unix.rs#L20 and https://github.com/habitat-sh/habitat/blob/master/components/core/src/os/net/windows.rs#L22, but remove the TODO here, since it's not where the change needs to be made.

components/sup/protocols/event.proto
afiune: @baumanj  that crate is abandoned.

components/sup/src/manager/mod.rs
afiune: @baumanj that crate is abandoned.

components/sup/protocols/event.proto
baumanj: What makes you say that? It hasn't been updated in awhile, but that could just be that there's no need to make changes to a very simple crate that does the job it's designed to.

components/sup/src/manager/mod.rs
afiune: I was thinking that we might need to implement the C function `getnameinfo()` instead of `gethostname()`. I was going to work on that next but I’ll difer to you @baumanj or @christophermaier to agree or disagree.

components/sup/protocols/event.proto
afiune: I say it because it says it in the tittle of the repository.

816596FD-11B9-4756-98CA-1A722DFD1FB8

afiune: @baumanj Also this: (https://github.com/fengcen/hostname/pull/4#issuecomment-455735989)
>fengcen **(REPO-OWNER)** commented on Jan 18
Sorry guys. I have no time and interest in Rust right now. This project is going to Abandoned status.

raskchanky: In my opinion, `fqdn` is acceptable for a field name.  The field name gets translated into a struct field name in code, so it's something you'll end up typing with some frequency.  I would venture to guess that any programmers working in this code will know what a `fqdn` is.  For those that don't, it feels like an abbreviation worth learning, because it comes up often.  Hypothetically, if there were a field on a protobuf message that indicated whether TCP was being used or not, it would be reasonable to call it `tcp` instead of `transmission_control_protocol`, even though the latter is more explicit.  This feels like a similar situation to me.  If more clarity is warranted, a comment could be added above the name indicating that it's a fully qualified domain name.  I try not to worship at the altar of brevity too much, but in this particular case, it feels like a reasonable tradeoff to me.

components/core/src/os/net/unix.rs
```diff
@@ -30,6 +30,20 @@ pub fn hostname() -> io::Result<String> {
...
+    let ptr = buf.as_mut_slice().as_mut_ptr();
+    match unsafe { getdomainname(ptr as *mut libc::c_char, len as libc::size_t) } {
+        0 => {
+            let c_str = unsafe { CStr::from_ptr(ptr as *const libc::c_char) };
+            Ok(c_str.to_string_lossy().into_owned())
+        }
+        code => Err(io::Error::from_raw_os_error(code)),
+    }
+}
+
```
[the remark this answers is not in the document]
afiune: tenor-109240101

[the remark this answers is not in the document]
afiune: I'll go use that crate! Thanks for your feedback cm.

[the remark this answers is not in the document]
afiune: Implemented it! 🎉 

tenor-32184341

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,71 @@ mod imp;
...
+// hint flag for the running operating system.
+pub fn lookup_fqdn(hostname: &str) -> io::Result<Vec<String>> {
+    let hints = AddrInfoHints { flags: ai_canonname(),
+                                ..AddrInfoHints::default() };
+    let addrinfos =
+        getaddrinfo(Some(hostname), None, Some(hints))?.collect::<std::io::Result<Vec<_>>>()?;
+
+    let canonnames = addrinfos.into_iter()
+                              .filter_map(|info| info.canonname)
+                              .collect();
```
christophermaier: Reading the [manpage for getaddrinfo](http://man7.org/linux/man-pages/man3/getaddrinfo.3.html) reveals this nugget:

> If hints.ai_flags includes the AI_CANONNAME flag, then the ai_canonname field of the first of the addrinfo structures in the returned list is set to point to the official name of the host.

This means that we just need to look at the first item in the list, grab the canonical name out of it, and return a `String` rather than a `Vec<String>`, so that should simplify a lot of stuff here.

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,71 @@ mod imp;
...
+                 AddrInfoHints};
+use std::io;
+
+// lookup_fqdn returns a vector of fqdn that resolves the provided hostname.
+//
+// Since the underlying crate is platform agnostic, this function is as well,
+// we only implement a single function `ai_canonname()` that returns the right
+// hint flag for the running operating system.
+pub fn lookup_fqdn(hostname: &str) -> io::Result<Vec<String>> {
+    let hints = AddrInfoHints { flags: ai_canonname(),
```
christophermaier: Since the `ai_cannonname()` functions are so small, and only used here, it might be a bit cleaner to model things a bit differently:

```
    #[cfg(unix)]
    let flags = libc::AI_CANONNAME;

    #[cfg(windows)]
    let flags = winapi::shared::ws2def::AI_CANONNAME;

    let hints = AddrInfoHints { flags, ..AddrInfoHints::default() };
```

components/core/src/os/net/mod.rs
afiune: @christophermaier Would it be ok to return an `Option<String>` since that is what `canonname` is already returning?

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,73 @@ mod imp;
...
+    // If 'hints.ai_flags' includes the AI_CANONNAME flag, then the ai_canonname
+    // field of the first of the addrinfo structures in the returned list is set
+    // to point to the official name of the host.
+    if !addrinfos.is_empty() {
+        let addrinfo = addrinfos[0].clone();
+        return Ok(addrinfo.canonname.unwrap_or_else(|| hostname.to_string()));
+    }
+
+    Ok(hostname.to_string())
+}
```
christophermaier: If you make `addrinfos` above `mut`, then you can replace this bit with the following:

```rust
    let name = if let Some(addrinfo) = addrinfos.pop() {
        addrinfo.canonname.unwrap_or_else(|| hostname.to_string())
    } else {
        hostname.to_string()
    };
    Ok(name)
```
That simplifies things by removing a clone, and also removes a `return` call.

I think there are maybe some additional simplifications we can make (we ought to be able to just use `unwrap` or `expect`), but I'd want to do some additional testing first. This is good for now, though.

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,73 @@ mod imp;
...
+
+// fqdn returns the fully qualified domain name of the running machine
+pub fn fqdn() -> Option<String> {
+    match hostname() {
+        Ok(hostname) => lookup_fqdn(&hostname).ok(),
+        Err(_) => None,
+    }
```
christophermaier: I'd be interested to see what the error cases here are (both in the `ok()` case, as well as the `Err(_)` case).

Logging those at a debug level would be nice.

(I'm also curious as to the various failure modes involved here, and what they imply.)

components/sup/src/manager/mod.rs
```diff
@@ -495,7 +495,12 @@ impl Manager {
             let es_config =
                 cfg.event_stream_config
                    .expect("Config should be present if the EventStream feature is enabled");
-            let ec = EventCore::new(&es_config, &sys);
+
+            // Collect the FQDN of the running machine
+            let fqdn = habitat_core::os::net::fqdn().unwrap_or_default();
```
christophermaier: Since `fqdn()` returns an `Option<String>`, `unwrap_or_default()` is going to return an empty string if `fqdn()` returns `None`... I'm not sure how useful that will be for our purposes here.

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,73 @@ mod imp;
...
 
 pub use self::imp::*;
+
+extern crate dns_lookup;
+
+use dns_lookup::{getaddrinfo,
+                 AddrInfoHints};
+use std::io;
+
+// lookup_fqdn returns a vector of fqdn that resolves the provided hostname.
```
christophermaier: In Rust, documentation comments on functions/methods use `///` instead of `//`.

Additionally, this is no longer what the function returns 😄

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,73 @@ mod imp;
...
+//
+// Since the underlying crate is platform agnostic, this function is as well,
+// we only implement a single variable `flags` that has the right hint for
+// the running operating system.
```
christophermaier: I don't really think this comment is needed.

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,73 @@ mod imp;
...
+        let addrinfo = addrinfos[0].clone();
+        return Ok(addrinfo.canonname.unwrap_or_else(|| hostname.to_string()));
+    }
+
+    Ok(hostname.to_string())
+}
+
+// fqdn returns the fully qualified domain name of the running machine
+pub fn fqdn() -> Option<String> {
+    match hostname() {
```
christophermaier: Interesting... the `dns_lookup` crate has its own `get_hostname` function... might be a good opportunity to get rid of some more code 😄

components/sup/src/manager/mod.rs
afiune: I thought we would send an empty fqdn inside the protobuf mesasge. By sending the default string will cause the protobuf to be "empty".

christophermaier: If that's the desired behavior on the Automate side, then you'd just keep it as `None`, which would ultimately give you an empty field in the protobuf. There's a difference between `None` and `Some("")`.

afiune: Oops. In golang, an empty string field is an empty string. I can update this as you recommend.

components/core/src/os/net/mod.rs
baumanj: There's no need to do a `collect` on the return of `getaddrinfo` at all, I don't think. I would suggest:
```rust
    if let Some(Ok(first_addr_info)) = getaddrinfo(Some(hostname), None, Some(hints))?.next() {
        Ok(first_addr_info.canonname
                       .expect("Some(canonname) if requested"))
    } else {
        Ok(hostname.to_string())
    }
```
or
```rust
    if let Some(first_result) = getaddrinfo(Some(hostname), None, Some(hints))?.next() {
        Ok(first_result?.canonname
                        .expect("Some(canonname) if requested"))
    } else {
        Ok(hostname.to_string())
    }
```
Depending on whether we want an `Err` variant of [`Result<AddrInfo>`](https://docs.rs/dns-lookup/1.0.0/dns_lookup/struct.AddrInfoIter.html) to return a `Result::Err` or the `hostname`.

afiune: Updated!

components/sup/src/manager/mod.rs
afiune: I chose to use hostname instead for now, I really don't know how to implement the `Option` in protobuf. 😞

components/core/src/os/net/mod.rs
afiune: Thank you both, updated it.

components/core/src/os/net/mod.rs
afiune: deleted

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,71 @@ mod imp;
...
+        }
+        Err(_) => None,
+    }
+}
+
+#[test]
+#[ignore]
+fn test_fqdn() {
+    // @afiune This test is ignore because it is testing the actual
+    // fqdn of the running machine, mine has 'afiune-ubuntu-vb.lala.com'
```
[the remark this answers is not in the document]
afiune: I think @christophermaier was trying to use the `resolv_wrapper` tool so we can mock the hostname of the machine. I just left this test since I used it to test the implementation and as a reminder that we actually need to test. If you think is worthless I don't mind deleting it.

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,73 @@ mod imp;
 mod imp;
 
 pub use self::imp::*;
+
+extern crate dns_lookup;
+
+use dns_lookup::{getaddrinfo,
+                 AddrInfoHints};
```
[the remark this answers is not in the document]
afiune: Ok. Updated.

components/core/src/os/net/mod.rs
afiune: Good idea! Added the debug statements 💯

components/core/src/os/net/mod.rs
afiune: Done done done!!!! Really simple and great implementation. 🥇

components/core/src/os/net/mod.rs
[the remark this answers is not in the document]
baumanj: Yeah, let's remove

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,76 @@ mod imp;
...
+}
+
+/// Returns the fully qualified domain name of the running machine.
+pub fn fqdn() -> Option<String> {
+    match dns_lookup::get_hostname() {
+        Ok(hostname) => {
+            match lookup_fqdn(&hostname) {
+                Ok(fqdn) => Some(fqdn),
+                Err(e) => {
+                    debug!("fqdn() was unabled to lookup the machine fqdn. {:?}", e);
```
baumanj:
```suggestion
                    debug!("fqdn() was unable to lookup the machine fqdn. {:?}", e);
```

components/core/src/os/net/mod.rs
baumanj: We can factor out the redundancy in the debug statements here with [`and_then`](http://doc.rust-lang.org/1.34.1/std/result/enum.Result.html#method.and_then):
```rust
    match dns_lookup::get_hostname().and_then(|hostname| lookup_fqdn(&hostname)) {
        Ok(fqdn) => Some(fqdn),
        Err(e) => {
            debug!("fqdn() was unable to lookup the machine fqdn. {:?}", e);
            None
        }
    }
```
Or, even a bit more concisely if we convert the `Result` -> `Option` using [`Result::ok`](http://doc.rust-lang.org/1.34.1/std/result/enum.Result.html#method.ok):
```rust
    let result = dns_lookup::get_hostname().and_then(|hostname| lookup_fqdn(&hostname));
    if let Err(ref e) = result {
        debug!("fqdn() was unable to lookup the machine fqdn. {:?}", e);
    }
    result.ok()
```

components/core/src/os/net/mod.rs
```diff
@@ -22,3 +22,76 @@ mod imp;
 mod imp;
 
 pub use self::imp::*;
+
+extern crate dns_lookup;
```
christophermaier: With recent changes to Rust for the 2018 Edition, `extern crate` is [hardly ever required](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-extern-crate) anymore.

components/core/src/os/net/mod.rs
christophermaier: Reading through the [source code](https://docs.rs/dns-lookup/1.0.0/src/dns_lookup/addrinfo.rs.html#96) it looks like errors of the kind returned by `first_result?` would result from null pointers or non-v4 or v6 socket families (things that should be exceedingly rare and pathological).

I think we could probably log those errors and return `Ok(hostname.to_string())`

At that point, the only error this function would be returning would be the case where the lookup call itself fails. Over in the manager, we currently up falling back to the hostname from the `Sys` struct. That hostname comes from our existing homegrown `hostname` implementation, which I now think that we should replace with the one from `dns_lookup` for consistency. That can be a followup PR, though.

components/core/src/os/net/mod.rs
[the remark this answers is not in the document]
afiune: Ok.

components/core/src/os/net/mod.rs
afiune: I'll try to update it to log a debug message and return the hostname as you mentioned.

components/core/src/os/net/mod.rs
afiune: I used the super concisely way (the second one). Though I have to say that even though my implementation was a bit more code (a bit more ugly 😅 ) it was a bit more helpful since you could have known whether this function was unable to get the machine `hostname` or get the machine `fqdn`, at least more clear and helpful for the end-user or developers that are troubleshooting anything at this level of the code base.

baumanj: If you think that distinction will be helpful, you don't *need* to make this change. It was a suggestion of how I'd have done it.

Looking at http://man7.org/linux/man-pages/man2/gethostname.2.html, it seems pretty unlikely that `get_hostname` will fail, and I tend towards simplicity and non-redundancy in the first revision since complexity is harder to reduce later. If we discover it's useful to log these errors distinctly, we can add it later (though that first time may be painful).

In any case, you're the author here and I want you to do what feels best to you. The fact that you made this comment tells me you're not entirely happy with my suggestion. If I see something that I feel is very important to change, I'll make that clear. Otherwise, you can interpret my comments as suggestions and alternative perspectives, not mandates.

afiune: This was absolutely a misunderstanding from my side Jon, I took your feedback very seriously and I shouldn't have done that. I promise to do what you said in your comment above next time and take it as a suggestion. My apologies.

afiune: Though, it is almost impossible for me not to change my code with your suggestions since they are so badass!!! Thank you so much for all your feedback. ❤️

baumanj: No apologies necessary! I'm happy to give lots of feedback. It helps me learn the new code and everyone gets better though seeing more different approaches and perspectives.

It's awesome to have another new rustacean!

christophermaier:
tenor-237683291

Thanks a ton, @afiune !

afiune:
tenor-222042211

Issue #6509, opened 2019-05-06
Investigate the possibility of making `core/hab-backline` a formal dependency of the Studio


christophermaier:
The `core/hab-backline` package is really a "meta-package", containing only dependencies, and is only used in `core/hab-studio`. However, it is not a direct dependency in the traditional Habitat sense (i.e., it's not in the `pkg_deps` or `pkg_build_deps` lists of `core/hab-studio`), but is pulled in dynamically in the actual executing code of the Studio.

If possible, we should try and make `core/hab-backline` a direct runtime dependency of the Studio. It would would eliminate a non-standard "indirect" dependency, making Habitat (and by extension, Builder) aware of the relationship between the Studio and its dependencies. It would simplify our release process (and possibly help to make it possible for more of our release to run out of Builder itself). It would also stand to simplify the code of the Studio, since it would eliminate a bunch of runtime logic that would now be handled by core Habitat logic.

With all that said, it remains to be seen whether it is actually possible to modify the Studio this way. We need to ensure that any changes that are made are compatible with the needs of the bootstrapping Studio, which is used for building the initial foundation of Habitat in a non-Habitat environment. This work may also reveal ways to tease bootstrapping requirements apart from "normal" Studio requirements, which may help clarify and simplify code for both of these usecases.

christophermaier:
Pinging @smacfarlane and @scotthain

smacfarlane:
One advantage backline provides is that it gives you a single package that contains all of studios "internal" dependencies that is then controlled by a single variable: `HAB_STUDIO_BACKLINE_PKG`. 

The backline package is also used by the [rootless studio](https://github.com/habitat-sh/habitat/blob/master/components/rootless_studio/default/Dockerfile#L5) that makes up our Docker image, so we'd need to track any changes in two places. These changes rarely happen (in the past 3 years the only change has been the pinning and then un-pinning of dependencies to pre-core-plans-refresh versions), but because it is rare it would be easy to forget to make both updates. 

That being said, I do think as a first pass making backline a run dependency for the studio is the right thing to do. I suspect that if we had that, the need for `HAB_STUDIO_BACKLINE_PKG` would go away.  Then, if we're able to address #6056, we should be able remove the need for backline completely. 

Random thought: Would we want/need to always build a studio with `--ignore-local` if we're moving items to direct dependencies of the studio?

christophermaier:
[...]

PR #6528, opened 2019-05-08
First pass at test framework to validate studio changes

smacfarlane:
As we investigate #6509 we will be making changes to the studio.  This is a first pass at adding some tests to verify that the behavior of the studio doesn't change as we make those modifications. 

There's currently no straightforward way to specify a studio package, as `hab studio enter` will use the latest `core/hab-studio` package that shares the same version as the `hab` command, so the `studio-from-package` tests assume that the `hab` cli matches the version of the studio you wish to test. 

I've also include a POC for Windows, though it depends on a powershell module that appears unmaintained.  

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -339,6 +339,19 @@ steps:
       automatic:
         limit: 1
 
+  - label: "[unit] :linux: studio"
+    command:
+      - ./test/run_studio_test.sh "studio-from-source"
+    expeditor:
+      executor:
+        linux: 
+          privileged: true
+          single-use: true 
+    timeout_in_minutes: 10
+    retry:
+      automatic:
+        limit: 1
+
  #################################################################
 
 
@@ -493,6 +506,16 @@ steps:
   - label: "[unit] :windows: sup-protocol"
     command:
       - ./test/run_cargo_test.ps1 sup-protocol
+    agents:
+      queue: 'default-windows-privileged'
+    timeout_in_minutes: 20
+    retry:
+      automatic:
+        limit: 1
+
+  - label: "[unit] :windows: studio"
+    command:
+      - ./test/run_studio_test.ps1 "studio-from-source"
     agents:
       queue: 'default-windows-privileged'
     timeout_in_minutes: 20
@@ -684,6 +707,7 @@ steps:
     retry:
       automatic:
         limit: 1
+
 
 #######################################################################
 # Things that have no tests but should be built to make sure they
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -347,7 +347,7 @@ steps:
         linux: 
           privileged: true
           single-use: true 
-    timeout_in_minutes: 10
+    timeout_in_minutes: 5
     retry:
       automatic:
         limit: 1
@@ -518,7 +518,7 @@ steps:
       - ./test/run_studio_test.ps1 "studio-from-source"
     agents:
       queue: 'default-windows-privileged'
-    timeout_in_minutes: 20
+    timeout_in_minutes: 5
     retry:
       automatic:
         limit: 1
--- a/components/studio/Makefile
+++ b/components/studio/Makefile
@@ -1,0 +1,15 @@
+HAB_BLDR_CHANNEL ?= stable
+
+export HAB_INTERNAL_BLDR_CHANNEL = ${HAB_BLDR_CHANNEL}
+export CI_OVERRIDE_CHANNEL = ${HAB_BLDR_CHANNEL}
+export HAB_STUDIO_SECRET_HAB_INTERNAL_BLDR_CHANNEL = ${HAB_BLDR_CHANNEL}
+export HAB_STUDIO_SECRET_CI_OVERRIDE_CHANNEL = ${HAB_BLDR_CHANNEL}
+
+test: studio-from-package studio-from-source
+
+studio-from-package:
+	test/studio-from-package/test.sh 
+
+studio-from-source: 
+	test/studio-from-source/test.sh
+
--- a/components/studio/test/shared/studio-enter.exp
+++ b/components/studio/test/shared/studio-enter.exp
@@ -1,0 +1,40 @@
+#!/usr/bin/env expect
+
+set env(HAB_NOCOLORING) true
+set studio_enter_command $::env(STUDIO_ENTER_COMMAND)
+set studio_test [lindex $argv 0]
+# Print out some helpful tracing messages in the test output.
+
+proc log {message} {
+    puts "LOG >>> $message"
+}
+
+# Cleanup after the test
+exit -onexit {
+    exec hab studio rm
+}
+
+# Installing packages can take time
+set timeout 30
+
+# {*} Syntax explanation:
+# https://www.tcl.tk/man/tcl/TclCmd/Tcl.htm [5] Argument Expansion
+spawn {*}$studio_enter_command
+expect {
+  {\[default:/src:0]#} { 
+    log "Studio entered successfully -alpha "
+  }
+  eof {
+    error "Unable to enter studio"
+  }
+  timeout {
+    error "Timeout entering studio"
+  } 
+}
+
+send "$studio_test\n"
+expect {
+  {\[PASS]} { }
+  {\[FAIL]} { error "Studio internals tests failed" }
+  timeout { error "Studio internals tests timed out" }
+}
--- a/components/studio/test/shared/studio-enter.exp
+++ b/components/studio/test/shared/studio-enter.exp
@@ -22,7 +22,7 @@ set timeout 30
 spawn {*}$studio_enter_command
 expect {
   {\[default:/src:0]#} { 
-    log "Studio entered successfully -alpha "
+    log "Studio entered successfully"
   }
   eof {
     error "Unable to enter studio"
--- a/components/studio/test/shared/studio-enter.ps1
+++ b/components/studio/test/shared/studio-enter.ps1
@@ -1,0 +1,41 @@
+# Depends on https://www.powershellgallery.com/packages/Await/0.8
+param(
+    [String]$studio_command,
+    [String]$test_case
+)
+
+$env:HAB_NOCOLORING = "true"
+# Await has trouble parsing non-ascii glyphs
+$env:HAB_GLYPH_STYLE = "ascii"  
+$exit_code = 0
+
+try { 
+    Start-AwaitSession 
+    Send-AwaitCommand "$studio_command enter" 
+    $prompt = Wait-AwaitResponse "[HAB-STUDIO] Habitat:\src>"
+    Write-Host $prompt
+
+    Send-AwaitCommand "$test_case" 
+    $result = Wait-AwaitResponse "test finished"
+
+    Write-Host $result
+    if($result -match "\[FAIL\]") {
+        $exit_code = 1
+    }
+
+} finally { 
+    Write-Host "Cleaning up"
+    Send-AwaitCommand "exit"
+    
+    # Await won't block on "Waiting for supervisor to finish..." 
+    # and there's no output to match afterwords to wait on.
+    # This sleep is required to give the studio a chance to finish 
+    # exiting before we issue the `rm`.
+    sleep 10
+
+    Send-AwaitCommand "$studio_command rm"
+    sleep 2
+    Stop-AwaitSession
+}
+
+exit $exit_code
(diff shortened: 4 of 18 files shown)


components/studio/test/shared/studio-enter.exp
```diff
@@ -0,0 +1,38 @@
...
+  {default} { log "Something odd" }
+  eof {
+    error "Unable to enter studio"
+  }
+  timeout {
+    error "Timeout entering studio"
+  } 
+}
+
+send "test/shared/studio-internals/test-all.sh\n"
```
christophermaier: Not quite sure what's going on yet, but it appears that this is being run both inside the studio, as well as outside the studio when the first run is done 😱

components/studio/test/shared/studio-internals/test-all.sh
```diff
@@ -0,0 +1,30 @@
...
+    metadata
+    hab-version
+)
+
+exit_code=0 
+
+for test_case in "${test_cases[@]}"; do
+  echo "Testing '$test_case'"
+  test/shared/studio-internals/test-studio-"$test_case".sh 
+  if [[ $? -ne 0 ]]; then
```
christophermaier: to simplify a bit, you could redo this as:

```bash
if test/shared/studio-internals/test-studio-"${test_case}".sh; then
  echo "PASS - ${test_case}"
else
  echo "FAIL - ${test_case}"
  exit_code = 1
fi
```

components/studio/test/shared/studio-internals/test-all.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo "FAIL - $test_case"
+    exit_code = 1
+  else
+    echo "PASS - $test_case" 
+  fi 
+done 
+
+
+# This is for expect, as I don't know how to get the result 
+# of a command passed to "send"
```
christophermaier: You're not going to be able to get an exit code from something passed to `send`, since that is really just typing the commands for you.

Instead, I think you'll want to look at some combination of `spawn` and `wait` (e.g., https://stackoverflow.com/a/23632210)

components/studio/test/shared/studio-internals/test-studio-metadata.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if ! grep -q 'studio_build_environment=""' /.studio; then 
+  echo "Studio build environment incorrect" 
+  exit_code=1
+fi
+
+if ! grep -q 'studio_run_environment=""' /.studio; then 
+  echo "Studio run environment incorrect" 
+  exit_code=1
+fi
```
christophermaier: You need an `exit ${exit_code}` down here.

components/studio/test/shared/studio-internals/test-all.sh
smacfarlane: It wasn't entirely clear to me if spawn was an option. We want all the test commands to run in the context (inside) of the `spawn`ed studio,  can a subsequent spawn be executed in that context, or will it it always execute on the host (external to the studio)?   This is something I need to look into still 😄

components/studio/test/shared/studio-enter.exp
smacfarlane: Interesting, I wasn't observing that behavior.   How were you invoking the tests?  It's also entirely possible I messed something up when I extracted this from my studio work branch 😂  I'll look into it.

components/studio/test/shared/studio-internals/test-studio-metadata.sh
smacfarlane: 👍  I never finished that test as I wasn't sure if it was actually useful.  I was originally worried about accidentally changing the contents of the file as part of my investigations into #6509 , but I don't know that we actually use that file anywhere.

components/studio/Makefile
```diff
@@ -0,0 +1,16 @@
...
+	test/studio-from-package/test.sh
+
+studio-from-source: 
+	test/studio-from-source/test.sh
+
+targets:
+	@echo "Targets: "
+	@echo "  studio-from-package"
+	@echo "  studio-from-source"
+	@echo "  test: studio-from-package, studio-from-source"
```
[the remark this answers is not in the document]
smacfarlane: Yeah,  I'm going to toss this I think.  I had the same thought as I was writing it and I don't know that it provides much value.

components/studio/test/shared/studio-internals/test-studio-metadata.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [ ! -f /.studio ]; then 
+  echo "Unable to find .studio metadata"
+  exit_code=1
+fi
+
+if ! grep -q 'studio_type="default"' /.studio; then 
+  echo "Studio type is not 'default'"
+  exit_code=1
+fi
```
baumanj: Might make things a bit clearer to reduce the duplication:
```bash
check_dot_studio_for() {
  if ! grep -q "$1" /.studio; then
    echo "Expected '$1' not found in .studio"
    exit_code=1
  fi
}

check_dot_studio_for 'studio_type="default"'
check_dot_studio_for 'studio_path="/hab/bin"'
check_dot_studio_for 'studio_build_environment=""'
check_dot_studio_for 'studio_run_environment=""'
```

components/studio/test/studio-from-source/test.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+export HAB_LICENSE="accept-no-persist"
+
+sudo hab pkg install core/busybox-static core/hab core/hab-backline
+
+# Current studio has the expectation that busybox and hab live in the libexec directroy
+# These two lines should be removed at a later date to validate this is no longer a requirement
```
baumanj: Seems like this should be an issue with a link to it here

components/studio/test/studio-from-source/test.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
```
baumanj:
```suggestion
set -euo pipefail 
```

components/studio/test/shared/studio-enter.exp
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env expect
+
+set env(HAB_NOCOLORING) true
+set studio_enter_command $::env(studio_enter_command)
```
baumanj: It's conventional for variables that will be exported into the environment to be all caps. Any reason not to make this `STUDIO_ENTER_COMMAND`?

components/studio/test/studio-from-source/test.ps1
```diff
@@ -0,0 +1,29 @@
...
+mkdir "bin/7zip" | Out-Null
+
+Copy-Item "$(hab pkg path core/powershell)/bin/*" "bin/powershell" -Recurse
+Copy-Item "$(hab pkg path core/hab)/bin/*" "bin/hab"
+Copy-Item "$(hab pkg path core/7zip)/bin/*" "bin/7zip"
+Copy-Item "$(hab pkg path core/hab-plan-build-ps1)/bin/*" "bin/"
+
+try {
+    test/shared/studio-enter.ps1 -studio_command "bin/hab-studio.ps1"
+    sleep 5
```
mwrock: do you need this sleep in addition to what is in `studio-enter.ps1`?

components/studio/test/studio-from-package/test.ps1
```diff
@@ -0,0 +1,7 @@
+$env:HAB_LICENSE = "accept-no-persist"
+
+hab studio rm
+
+& test/shared/studio-enter.ps1 -studio_command "hab studio"
+
+exit $LastExitStatus
```
mwrock:
```suggestion
exit $LASTEXITCODE
```

components/studio/test/studio-from-source/test.ps1
```diff
@@ -0,0 +1,29 @@
...
+    $exit_code = $LastExitStatus
```
mwrock:
```suggestion
    $exit_code = $LASTEXITCODE
```

components/studio/test/shared/studio-enter.ps1
```diff
@@ -0,0 +1,34 @@
...
+    Send-AwaitCommand "$studio_command enter" 
+    $prompt = Wait-AwaitResponse "[HAB-STUDIO] Habitat:\src>"
+    Write-Host $prompt
+
+    Send-AwaitCommand "test/shared/studio-internals/test-all.ps1"
+    Write-Host "Send test comand"
+     $result = Wait-AwaitResponse "All tests finished"
+
+    Write-Host $result
+    if($result -match "[FAIL]") {
```
mwrock: where is `[FAIL]` emitted?

components/studio/test/shared/studio-enter.exp
smacfarlane: No reason to not make the change, it's left over from an earlier iteration. Will update.

components/studio/test/shared/studio-enter.ps1
```diff
@@ -0,0 +1,34 @@
+# Depends on https://www.powershellgallery.com/packages/Await/0.8
+param([String]$studio_command) #
+
+$env:HAB_NOCOLORING = "true"
+# Await has trouble parsing non-ascii glyphs
+$env:HAB_GLYPH_STYLE = "ascii"  
+$exit_code = 0
+
+Install-Module -Name Await
```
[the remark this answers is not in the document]
smacfarlane: 👍  I've moved this out to the wrapper, so I'll be removing this instance of it.

test/run_studio_test.ps1
```diff
@@ -0,0 +1,14 @@
+#Requires -Version 5
+
+$ErrorActionPreference="stop"
+
+Install-Module -Name Await -Force
+
+Push-Location "components/studio"
+try {
+  Invoke-Expression test/studio-from-source/test.ps1
```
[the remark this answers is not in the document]
smacfarlane: Good to know, thank you!

components/studio/test/studio-from-source/test.ps1
smacfarlane: I'll double check this to be certain it's still required, but what I was encountering was failures removing files in the `finally` block because the studio was still partially mounted. The sleep gave it a chance for any open handles to the copied files to be closed.  

If it's still required I'll add a comment to that effect, or if there's a better way to handle it I can implement that. Not a fan of arbitrary sleeps, but I'm not aware of another way on Windows to handle that.

components/studio/test/shared/studio-enter.ps1
smacfarlane: `[FAIL]` should be emitted by each individual test. I'm doing some small refactoring and I'll double check that that is true.   I've also added a README (not pushed yet) explaining how these tests work and what they should do.

mwrock: yeah I see that in the bash/exp but don't see anything on the powershell side that would output `[FAIL]`

smacfarlane: This should be fixed now.

.expeditor/verify.pipeline.yml
```diff
@@ -500,6 +513,16 @@ steps:
       automatic:
         limit: 1
 
+  - label: "[unit] :windows: studio"
+    command:
+      - ./test/run_studio_test.ps1
+    agents:
+      queue: 'default-windows-privileged'
+    timeout_in_minutes: 20
+    retry:
```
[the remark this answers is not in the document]
smacfarlane: I think having  a single retry is beneficial some or most of the time.  If I know it takes 10+ minutes to complete a test suite, then I'm probably going to switch gears and respond to email, check slack, etc.  If a transient failure happens in the first few minutes then the auto retry will have saved me time over all.  The repo I have most experience with tends to fail fast or be more susceptible to transient issues, so that certainly influences my opinion, and it may not hold true to this code base.  It would certainly be interesting to see data on failure types and if retries are saving us time. 

That being said, I've reduced the timeouts to 5 minutes for these tests as they should never take more than a couple minutes. We can adjust upwards later if that proves to be too short.

smacfarlane:
@cm @baumanj @mwrock   This should be good to go now. I've made some structural changes the code to ensure each test is run in a fresh studio instance, which bubbled out into how each script is invoked.  I think it ended in a much better spot though, and should be easier to add additional tests as we find cases we want to test.

.expeditor/verify.pipeline.yml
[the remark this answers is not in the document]
baumanj: That sounds reasonable. The downside of retrying is directly proportional to the length of the timeout. The piece of data I think we lack (and which probably varies somewhat according to the particular step) is the prevalence of transient issues that have nothing to do with our code so the retry is a pure time-saver. If it's a transient issue in *our* code, I'd much rather have it fail and know about it.

PR #6535, opened 2019-05-09
Update tutorials with license acceptance information

smacfarlane:
This adds information on how to accept the license when running the exported docker containers.  It also adds a notice about the license on install with a link to our licenseing page. 

I haven't verified the Windows docker export lines are correct yet, but the docker docs indicate that `--env` behaves exactly the same as it does on Linux. 


mwrock:
Thanks so much for adding this!

PR #6540, opened 2019-05-10
remove trailing space from env vars set in export docker build

mwrock:
This fixes broken docker export on windows. I've definitely used this before without issue and perhaps we should also `trim` the `HAB_LICENSE` to protect against these kinds of scenarios.

### DIFF
--- a/components/pkg-export-docker/defaults/Dockerfile_win.hbs
+++ b/components/pkg-export-docker/defaults/Dockerfile_win.hbs
@@ -5,7 +5,7 @@ EXPOSE 9631 {{exposes}}
 RUN SET HAB_FEAT_OFFLINE_INSTALL=ON && \
     {{~ #if environment}}
     {{~ #each environment}}
-        SET {{@key}}={{{this}}} && \
+        SET {{@key}}={{{this}}}&& \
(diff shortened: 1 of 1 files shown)


components/pkg-export-docker/defaults/Dockerfile_win.hbs
```diff
@@ -5,7 +5,7 @@ EXPOSE 9631 {{exposes}}
 RUN SET HAB_FEAT_OFFLINE_INSTALL=ON && \
     {{~ #if environment}}
     {{~ #each environment}}
-        SET {{@key}}={{{this}}} && \
+        SET {{@key}}={{{this}}}&& \
```
baumanj: Would
```suggestion
        SET {{@key}}="{{{this}}}" && \
```
work as well? It's strange to have to space before the `&&`, so if this handles the issue, quoting the RHS might be a good convention to adopt.

mwrock: no that would include the quotes in the value

baumanj: Oh, wild. TIL.

PR #6547, opened 2019-05-13
make sure all build prereqs exist on windows, run cargo test against core/rust and use a local studio for builds

mwrock:
This will install the hab cli, visual studio tools and rustup if they are not present. This will also run `cargo test` against the latest core/rust instead of whatever toolchain was installed with rustup. Finally this moves to using a local studio instead of a docker studio for the CI component builds. Because we control the host and our builds do not impact machine state, we don't get value from a dockerized studio and we should also get faster build times and less resource consumption with local builds.

### DIFF
--- a/support/ci/build_component.ps1
+++ b/support/ci/build_component.ps1
@@ -7,6 +7,9 @@ param (
     [string]$Component
 )
 
+. $PSScriptRoot\shared.ps1
+Install-Habitat
+
 # Since we are only verifying we don't have build failures, make everything
 # temp!
 $env:HAB_ORIGIN="throwaway"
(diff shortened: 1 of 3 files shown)


mwrock:
@baumanj its a little tricky to get that working just right in powershell. Typically you would use `[parameter(ValueFromRemainingArguments=$true)]` to just aggregate all remaining args passed tro the script but it will have quotes and what powershell thinks are empty args removed. For example:
```
.\test.ps1 component --features "vvvvv vvvvvvv" -- --test-options blah
```
becomes:
```
running 'command --features vvvvv vvvvvvv --test-options blah'
```
I'm sure I could get something working but would rather do that later.

baumanj:
> @baumanj its a little tricky to get that working just right in powershell

Don't worry about it then. I was just throwing it out there in case it was simple. Too bad `pwsh` doesn't have anything like `sh`'s `"$@"` which passes through all arguments as they were received.

PR #6549, opened 2019-05-13
Make studio backline dependency explicit

### LINKED ISSUE #6509: Investigate the possibility of making `core/hab-backline` a formal dependency of the Studio

smacfarlane:
Ref: #6509 

This is initial set of work to make studio dependencies explicit, starting with backline.  

In order for the studio to rely on the RUNTIME_ENVIRONMENT in the package metadata, we need to treat the `studio enter` command the same way we run `hab pkg exec`.  This  is currently a copy/paste from `hab pkg exec` and will need to be refactored in the future. When I started investigating the possibility of doing that work in this PR, it ended up being a much a much larger slice of work that I felt should be captured in its own PR. 

### DIFF
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -63,7 +63,8 @@ mod inner {
                         fs::{am_i_root,
                              find_command},
                         os::process,
-                        package::{PackageIdent,PackageInstall},
+                        package::{PackageIdent,
+                                  PackageInstall},
                         users::linux as group},
                 VERSION};
     use std::{env,
@@ -86,9 +87,10 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
-                    // TODO(SM): This is the minimum change needed to allow the studio to specify
-                    // its dependencies. This is a duplicate of the code in `hab pkg exec` and
-                    // should be refactored.
+                    // This is a duplicate of the code in `hab pkg exec` and
+                    // should be refactored as part of or after:
+                    // https://github.com/habitat-sh/habitat/issues/6633
+                    // https://github.com/habitat-sh/habitat/issues/6634
                     let pkg_install = PackageInstall::load(&ident, None)?;
                     let cmd_env = pkg_install.environment_for_command()?;
                     for (key, value) in cmd_env.into_iter() {
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -63,7 +63,7 @@ mod inner {
                         fs::{am_i_root,
                              find_command},
                         os::process,
-                        package::PackageIdent,
+                        package::{PackageIdent,PackageInstall},
                         users::linux as group},
                 VERSION};
     use std::{env,
@@ -86,6 +86,23 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
+                    // TODO(SM): This is the minimum change needed to allow the studio to specify
+                    // its dependencies. This is a duplicate of the code in `hab pkg exec` and
+                    // should be refactored.
+                    let pkg_install = PackageInstall::load(&ident, None)?;
+                    let cmd_env = pkg_install.environment_for_command()?;
+                    for (key, value) in cmd_env.into_iter() {
+                        debug!("Setting: {}='{}'", key, value);
+                        env::set_var(key, value);
+                    }
+
+                    let mut display_args = super::STUDIO_CMD.to_string();
+                    for arg in args {
+                        display_args.push(' ');
+                        display_args.push_str(arg.to_string_lossy().as_ref());
+                    }
+                    debug!("Running: {}", display_args);
+
                     exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?
                 }
             };
--- a/components/studio/habitat/plan.sh
+++ b/components/studio/habitat/plan.sh
@@ -4,7 +4,9 @@ pkg_origin=core
 pkg_version=$(cat "$SRC_PATH/../../VERSION")
 pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_license=('Apache-2.0')
-pkg_deps=()
+pkg_deps=(
+  core/hab-backline/$pkg_version
+)
 pkg_build_deps=(core/coreutils
                 core/tar
                 core/xz
@@ -12,6 +14,10 @@ pkg_build_deps=(core/coreutils
                 core/busybox-static
                 core/hab)
 pkg_bin_dirs=(bin)
+
+do_prepare() {
+  set_runtime_env "HAB_STUDIO_BACKLINE_PKG" "$(< $(pkg_path_for core/hab-backline)/IDENT)"
+}
 
 do_build() {
   cp -v "$SRC_PATH"/bin/hab-studio.sh hab-studio
--- a/components/studio/habitat/plan.sh
+++ b/components/studio/habitat/plan.sh
@@ -5,7 +5,7 @@ pkg_version=$(cat "$SRC_PATH/../../VERSION")
 pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_license=('Apache-2.0')
 pkg_deps=(
-  core/hab-backline/$pkg_version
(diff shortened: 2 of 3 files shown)


components/hab/src/command/studio/enter.rs
```diff
@@ -86,6 +86,23 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
+                    // TODO(SM): This is the minimum change needed to allow the studio to specify
+                    // its dependencies. This is a duplicate of the code in `hab pkg exec` and
+                    // should be refactored.
```
baumanj: Rather than a `TODO`, let's file an issue with a link back here so it's discoverable

baumanj: Or just refactor now; it doesn't look like it would be too disruptive. What am I missing that makes it so?

components/studio/libexec/hab-studio-type-default.sh
```diff
@@ -88,17 +85,7 @@ finish_setup() {
...
-      info "Override channel set; retrieving ${pkg} from ${CI_OVERRIDE_CHANNEL}"
-      _hab install --channel="${CI_OVERRIDE_CHANNEL}" "${pkg}" || {
-        info "Package not found in ${CI_OVERRIDE_CHANNEL}; falling back to stable"
-        _hab install "${pkg}"
-      }
-    else
-      _hab install "$pkg"
-    fi
-  done
+  _hab install $HAB_STUDIO_BACKLINE_PKG
```
baumanj: Can you explain a little why `CI_OVERRIDE_CHANNEL` is no longer necessary? I'm happy to see it go, I just don't understand why.

components/hab/src/command/studio/enter.rs
smacfarlane: Refactoring this would involve updating code in `hab pkg exec` and `core/os/process` or `PackageInstall` which I feel should be a separate PR. Additionally, when I started looking at making the same change for the Windows studio, I found a lot of duplication and not very easy to grok logic to determine how we enter a studio. Refactoring that will be a larger slice of work that I again feel it should be a separate PR. 

Given that this change is to help unblock some changes we're making to our release process and it's updating a component that is both critical and not as well tested as we'd like, I'd like to keep this change as minimal as possible to make discovery of any potential issues easy to debug.

I've filed #6633 and #6634 to track the changes needed.

components/studio/libexec/hab-studio-type-default.sh
smacfarlane: Before this change, core/hab-backline was an implicit dependency of the studio used when the studio is invoked.  Since it's not declared, the studio would have to determine what backline package to use every time it is run.  This would always use the latest stable version of the package that matches `core/hab-backline/VERSION`. In the context of a release, we don't want the latest stable but we can't rely on HAB_BLDR_CHANNEL as that has a different intent and overloading it to test backline caused issues (Note: I wasn't around at the time so I don't know the specifics) so `CI_OVERRIDE_CHANNEL` was introduced.

By adding `core/hab-backline` to `pkg_deps`, we can leverage the plan-build logic to always resolve that to a fully qualified ident at build time, which allows us to set `$HAB_STUDIO_BACKLINE_PKG` at the time of build. With the change to `studio enter` to set up the environment as specified by the studio package, we know exactly what backline package the studio was built with and can correctly install it on the _inside_ of the studio. 

This is in alignment with how packages in Habitat are intended to work. Rather than the existing behavior where we've built a back door in order to test changes to a component, we now rebuild the dependent packages against the new version and `CI_OVERRIDE_CHANNEL` can go away.

components/hab/src/command/studio/enter.rs
```diff
@@ -86,6 +86,23 @@ mod inner {
...
+                        debug!("Setting: {}='{}'", key, value);
+                        env::set_var(key, value);
+                    }
+
+                    let mut display_args = super::STUDIO_CMD.to_string();
+                    for arg in args {
+                        display_args.push(' ');
+                        display_args.push_str(arg.to_string_lossy().as_ref());
+                    }
+                    debug!("Running: {}", display_args);
```
christophermaier: This could be a note for the future refactorings, rather than a change for this PR, but this is a lot of code for just printing out what's going to be run. It could be done more concisely (e.g., with some help from [SliceConcatExt::join](https://doc.rust-lang.org/std/slice/trait.SliceConcatExt.html#tymethod.join)).

(As an aside, it's currently doing all that concatenation even if debug logging is disabled, making it wasted effort.)

Reading more of the surrounding code, it also seems out of place here, given that this block of code is just trying to figure out which Studio binary to run; the arguments only really get used later on.

Similarly, jamming the environment variables directly into the running environment here, rather than somehow conveying it out to the point where the command is actually run, strikes me as a bad idea. Sure, it ends up working out to the same thing in this case, since we `exec` to become the studio process, but messing with global state at a point so distant from the actual execution just seems like begging for bugs.

components/studio/habitat/plan.sh
```diff
@@ -13,6 +15,10 @@ pkg_build_deps=(core/coreutils
                 core/hab)
 pkg_bin_dirs=(bin)
 
+do_prepare() {
+  set_runtime_env "HAB_STUDIO_BACKLINE_PKG" "$(< $(pkg_path_for core/hab-backline)/IDENT)"
```
christophermaier: Is this going to be a permanent thing, or is this basically some support code to bridge us to the glorious future when all this code is sane? 😅

components/studio/libexec/hab-studio-type-default.sh
christophermaier: This is my favorite thing!

components/hab/src/command/studio/enter.rs
smacfarlane: > (As an aside, it's currently doing all that concatenation even if debug logging is disabled, making it wasted effort.)
> 
> Reading more of the surrounding code, it also seems out of place here, given that this block of code is just trying to figure out which Studio binary to run; the arguments only really get used later on.

Yup 😭  I couldn't see a low-impact way of doing this only when necessary, `studio/enter.rs` is need of a lot of cleanup.  Ideally we'd do that work up front, but this will unblock a lot of the improvements to our release process. I'm open to suggestions, but as I noted above I feel that work should be it's own PR and keep this isolated to changes necessary to make backline a runtime dependency.

components/studio/habitat/plan.sh
smacfarlane: I suspect semi-permanent as `RUNTIME_ENVIRONMENT` is our interface for influencing the runtime (but not the internals!) of the studio. Though...  I just had a thought that may allow us to remove this. I'll run some experiments, but I think for now this should stay as-is.

components/hab/src/command/studio/enter.rs
christophermaier: @smacfarlane Yup, I totally get the constraints you're operating under. Feel free to take my suggestions as todo items on that future PR.

I _would_ like for us to be able to tackle that PR in the very near future, however, just so we can not let bad code lie any longer than absolutely necessary.

components/hab/src/command/studio/enter.rs
baumanj: Ok, can we replace the TODO with references to those issues?

components/hab/src/command/studio/enter.rs
baumanj: I had to abandon https://github.com/habitat-sh/habitat/pull/6347, but I think that direction would take us towards something cleaner and more maintainable.

Issue #6500, opened 2019-05-01
Add support for runtime configuration of logging levels


christophermaier:
Our current logging configuration is based around environment variables, meaning that any changes in that configuration requires a restart of the Supervisor in such a way that requires all services it manages to be restarted. This isn't great for debugging live scenarios.

We should switch over to a logging backend that supports runtime configuration changes. [log4rs](https://docs.rs/log4rs/0.8.3/log4rs/) is one such backend to investigate (through the `Handle` returned from its [`init_config` method](https://docs.rs/log4rs/0.8.3/log4rs/fn.init_config.html)) (there may also be other libraries to look into, as well.)

christophermaier:
Addendum: it looks like `log4rs` provides the building blocks to dynamically change logging levels, but how we choose to use that is up to us. For instance, it looks like we could watch changes to a config file and react by updating the configuration, we could reload by responding to a signal (though that has cross-platform implications), we could have some kind of supervisor command, etc. Of those, the first sounds the easiest to implement, and to interact with, but other solutions can be possible. I bring it up to emphasize that this wouldn't just be a drop-in replacement that gives us magical logging superpowers for free; there will need to be a bit of extra design work. However, the end result would be incredibly beneficial for all Habitat users.

christophermaier:
Addendum to the addendum: I lied, there is a built in way to monitor a file for changes. If you're using a configuration file, and that file contains a `refresh_rate` key, and you use `log4rs::init_file` to start everything up, a refresh thread will be spawned that will sleep for the configured `refresh_rate` interval and wake up to check for modifications.

We may want to tweak how that operates in the future, but for now, it will definitely do the trick, and will unlock an important new feature for us.

PR #6565, opened 2019-05-15
Add thread heartbeats to log errors for likely deadlocks

baumanj:
The 0.80.0 release incident showed that deadlock scenarios can be very confusing to debug in production, so this adds calls to the various threads which run a continuous work loop to indicate that they are still processing and adds a new thread which checks the heartbeats and logs an error indicating the thread which is likely deadlocked if one does not occur for too long a period.

`ThreadAliveThreshold` (how long a heartbeat must go missing before an error is logged, default 5 minutes) and `ThreadAliveCheckDelay` (how frequently the checker thread looks for updated heartbeats, default 1 minute) can be configured with the `HAB_THREAD_ALIVE_THRESHOLD_SECS`
and `HAB_THREAD_ALIVE_THRESHOLD_SECS` environment variables.

### DIFF
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -1169,23 +1169,11 @@ impl fmt::Display for Server {
 }
 
 fn persist_loop(server: &Server) {
-    struct PersistLoopPeriod(Duration);
+    habitat_core::env_config_duration!(PersistLoopPeriod,
+                                       HAB_PERSIST_LOOP_PERIOD_SECS,
+                                       Duration::from_secs(30));
 
-    impl EnvConfig for PersistLoopPeriod {
-        const ENVVAR: &'static str = "HAB_PERSIST_LOOP_PERIOD_SECS";
-    }
-
-    impl Default for PersistLoopPeriod {
-        fn default() -> Self { Self(Duration::from_secs(30)) }
-    }
-
-    impl std::str::FromStr for PersistLoopPeriod {
-        type Err = std::num::ParseIntError;
-
-        fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
-            Ok(Self(Duration::from_secs(s.parse()?)))
-        }
-    }
+    let min_loop_period = PersistLoopPeriod::configured_value().0;
 
     loop {
         habitat_common::sync::mark_thread_alive();
@@ -1194,9 +1182,7 @@ fn persist_loop(server: &Server) {
         server.persist_data();
         let time_to_persist = before_persist.elapsed();
         trace!("persist_data took {:?}", time_to_persist);
-        match PersistLoopPeriod::configured_value().0
-                                                   .checked_sub(time_to_persist)
-        {
+        match min_loop_period.checked_sub(time_to_persist) {
             Some(time_to_wait) => thread::sleep(time_to_wait),
             None => {
                 warn!("Persisting data took longer than expected: {:?}",
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -40,8 +40,7 @@ use crate::{error::{Error,
             trace::{Trace,
                     TraceKind}};
 use habitat_common::FeatureFlag;
-use habitat_core::{crypto::SymKey,
-                   env::Config as EnvConfig};
+use habitat_core::crypto::SymKey;
 use prometheus::{HistogramTimer,
                  HistogramVec,
                  IntGauge};
@@ -1173,7 +1172,7 @@ fn persist_loop(server: &Server) {
                                        HAB_PERSIST_LOOP_PERIOD_SECS,
                                        Duration::from_secs(30));
 
-    let min_loop_period = PersistLoopPeriod::configured_value().0;
+    let min_loop_period: Duration = PersistLoopPeriod::configured_value().into();
 
     loop {
         habitat_common::sync::mark_thread_alive();
--- a/components/butterfly/src/server/mod.rs
+++ b/components/butterfly/src/server/mod.rs
@@ -40,7 +40,8 @@ use crate::{error::{Error,
             trace::{Trace,
                     TraceKind}};
 use habitat_common::FeatureFlag;
-use habitat_core::crypto::SymKey;
+use habitat_core::{crypto::SymKey,
+                   env::Config as EnvConfig};
 use prometheus::{HistogramTimer,
                  HistogramVec,
                  IntGauge};
@@ -1168,16 +1169,34 @@ impl fmt::Display for Server {
 }
 
 fn persist_loop(server: &Server) {
-    // TODO: Make this configurable with EnvConfig. That trait needs to move
-    // to common or core first
-    const MIN_LOOP_PERIOD: Duration = Duration::from_secs(30);
+    struct PersistLoopPeriod(Duration);
+
+    impl EnvConfig for PersistLoopPeriod {
+        const ENVVAR: &'static str = "HAB_PERSIST_LOOP_PERIOD_SECS";
+    }
+
+    impl Default for PersistLoopPeriod {
+        fn default() -> Self { Self(Duration::from_secs(30)) }
+    }
+
+    impl std::str::FromStr for PersistLoopPeriod {
+        type Err = std::num::ParseIntError;
+
+        fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+            Ok(Self(Duration::from_secs(s.parse()?)))
+        }
+    }
 
     loop {
+        habitat_common::sync::mark_thread_alive();
+
         let before_persist = Instant::now();
         server.persist_data();
         let time_to_persist = before_persist.elapsed();
         trace!("persist_data took {:?}", time_to_persist);
-        match MIN_LOOP_PERIOD.checked_sub(time_to_persist) {
+        match PersistLoopPeriod::configured_value().0
+                                                   .checked_sub(time_to_persist)
+        {
             Some(time_to_wait) => thread::sleep(time_to_wait),
             None => {
                 warn!("Persisting data took longer than expected: {:?}",
--- a/components/butterfly/src/server/pull.rs
+++ b/components/butterfly/src/server/pull.rs
@@ -53,7 +53,11 @@ impl Pull {
         socket.bind(&format!("tcp://{}", self.server.gossip_addr()))
               .expect("Failure to bind the ZMQ Pull socket to the port");
         'recv: loop {
-            habitat_common::sync::mark_thread_alive();
+            if let Ok(-1) = socket.get_rcvtimeo() {
+                trace!("Skipping thread liveliness checks due to infinite recv timeout");
+            } else {
+                habitat_common::sync::mark_thread_alive();
+            }
 
             if self.server.paused() {
                 thread::sleep(Duration::from_millis(100));
@@ -63,6 +67,8 @@ impl Pull {
             let msg = match socket.recv_msg(0) {
                 Ok(msg) => msg,
                 Err(e) => {
+                    // We intentionally set a timeout above so that `mark_thread_alive` can be
+                    // used to show this thread is alive even when there's no data to receive.
                     if e != zmq::Error::EAGAIN {
                         error!("Error receiving message: {:?}", e);
                     }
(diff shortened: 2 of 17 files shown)


components/butterfly/src/server/mod.rs
```diff
@@ -1168,16 +1169,34 @@ impl fmt::Display for Server {
 }
 
 fn persist_loop(server: &Server) {
-    // TODO: Make this configurable with EnvConfig. That trait needs to move
-    // to common or core first
-    const MIN_LOOP_PERIOD: Duration = Duration::from_secs(30);
+    struct PersistLoopPeriod(Duration);
```
christophermaier: We've got so many of these specialized duration types now (both in this PR and elsewhere) that it might be worth creating a macro that can generate all these implementations for us.

It would ensure consistency if nothing else.

(Just a thought, and not necessarily for this PR.)

components/butterfly/src/server/mod.rs
```diff
@@ -1168,16 +1169,34 @@ impl fmt::Display for Server {
...
+        habitat_common::sync::mark_thread_alive();
+
         let before_persist = Instant::now();
         server.persist_data();
         let time_to_persist = before_persist.elapsed();
         trace!("persist_data took {:?}", time_to_persist);
-        match MIN_LOOP_PERIOD.checked_sub(time_to_persist) {
+        match PersistLoopPeriod::configured_value().0
+                                                   .checked_sub(time_to_persist)
+        {
```
christophermaier: Given that this will never change once the binary has started, why not grab the configured value outside of the loop, so you only create it once?

components/common/src/lib.rs
```diff
@@ -146,7 +146,171 @@ impl FeatureFlag {
...
+    }
+
+    struct ThreadAliveThreshold(Duration);
+
+    impl EnvConfig for ThreadAliveThreshold {
+        const ENVVAR: &'static str = "HAB_THREAD_ALIVE_THRESHOLD_SECS";
+    }
+
+    impl Default for ThreadAliveThreshold {
+        fn default() -> Self { Self(Duration::from_secs(5 * 60)) }
```
christophermaier: TIL there's no `Duration::from_minutes`

components/common/src/lib.rs
```diff
@@ -146,7 +146,171 @@ impl FeatureFlag {
...
+
+    /// Call periodically from a thread which has a work loop to indicate that the thread is
+    /// still alive and processing its loop. If this function is not called for more than
+    /// `ThreadAliveThreshold`, it will be error logged as a likely deadlock.
+    pub fn mark_thread_alive() {
+        mark_thread_alive_impl(&mut THREAD_HEARTBEATS.lock()
+                                                     .expect("THREAD_HEARTBEATS poisoned"));
+    }
+
+    fn mark_thread_alive_impl(heartbeats: &mut HeartbeatMap) {
```
christophermaier: I like this `_impl` pattern to separate the mutex from the thing inside it.

components/common/src/lib.rs
```diff
@@ -146,7 +146,171 @@ impl FeatureFlag {
...
+    }
+
+    /// Call once per binary to start the thread which will check that all the threads that
+    /// call `mark_thread_alive` continue to do so.
+    pub fn spawn_thread_alive_checker() {
+        thread::Builder::new().name("thread-alive-check".to_string())
+                              .spawn(|| {
+                                  loop {
+                                      check_thread_heartbeats();
+                                      thread::sleep(ThreadAliveCheckDelay::configured_value().0);
```
christophermaier: Similarly, can we figure out what our delay is once before entering the loop since it will never change?

components/common/src/lib.rs
```diff
@@ -146,7 +146,171 @@ impl FeatureFlag {
...
+                                  }
+                              })
+                              .expect("Error spawning thread alive checker");
+    }
+
+    fn check_thread_heartbeats() {
+        for (name, last_heartbeat) in
+            threads_missing_heartbeat(&THREAD_HEARTBEATS.lock()
+                                                        .expect("THREAD_HEARTBEATS poisoned"),
+                                      ThreadAliveThreshold::configured_value().0)
```
christophermaier: Maybe this threshold could be passed in as an argument to `check_thread_heartbeats` as well.

components/common/src/lib.rs
```diff
@@ -146,7 +146,171 @@ impl FeatureFlag {
...
+        }
+    }
+
+    fn threads_missing_heartbeat(heartbeats: &HeartbeatMap,
+                                 threshold: Duration)
+                                 -> Vec<NameAndLastHeartbeat> {
+        heartbeats.iter()
+                  .filter_map(|(thread_id, (thread_name, last_heartbeat))| {
+                      let time_since_last_heartbeat = last_heartbeat.elapsed();
+                      trace!("{:?} {:?} last heartbeat: {:?} ago",
```
christophermaier: I really want #6500 now.

components/common/src/lib.rs
```diff
@@ -146,7 +146,171 @@ impl FeatureFlag {
...
+                                  })
+                                  .unwrap()
+                                  .join()
+                                  .unwrap();
+            thread::spawn(move || {
+                loop {
+                    mark_thread_alive_impl(&mut HEARTBEATS.lock().unwrap());
+                    thread::sleep(TEST_THRESHOLD / 2)
+                }
+            });
```
christophermaier: Do you need to clean this thread up? Or does it end up killing itself at the end of the function when `HEARTBEATS` gets dropped?

Also, does the `HEARTBEATS` mutex in these tests need to be in a `lazy_static!` block in the first place?

components/butterfly/src/server/mod.rs
baumanj: I had the same thought

components/butterfly/src/server/mod.rs
baumanj: Good point; it was originally a constant during my development and I added in the `EnvConfig` later. Will change.

components/common/src/lib.rs
baumanj: This thread should get cleaned up when the test runner exits. I suppose it would be better to have it shut down right at the end of this particular test. I'll see how much messier that makes things.

components/butterfly/src/server/mod.rs
baumanj: See https://github.com/habitat-sh/habitat/pull/6565/commits/419d60dbe29be4c931108a60fd36f1f51574fd6a

components/common/src/lib.rs
baumanj: See https://github.com/habitat-sh/habitat/pull/6565/commits/419d60dbe29be4c931108a60fd36f1f51574fd6a#diff-ce8ce0bdcb96f29e33454e96fcc78dafR275

baumanj:
Please take a look at https://github.com/habitat-sh/habitat/pull/6565/commits/f19fd6cbbc25273ddfc1e083a251bafa16fff931, it's the only real behavior changes I had to make outside the heartbeat code itself and I want to make sure it's safe.

components/core/src/env.rs
```diff
@@ -72,6 +72,75 @@ pub fn var_os<K: AsRef<OsStr>>(key: K) -> std::option::Option<OsString> {
...
+    ) => {
+        use $crate::env::Config as _;
+
+        struct $wrapping_type($wrapped_type);
+        // A little trickery to avoid env var name collisions:
+        // This enum can't ever be instantiated, but the compiler will give
+        // an error if two invocations in a namespace give the same env_var.
+        // It'd be nice to make this work globally, but I don't see a good way
+        #[allow(non_camel_case_types, dead_code)]
+        enum $env_var {}
```
baumanj: Here's what it generates if I made the same mistake I made in https://github.com/habitat-sh/habitat/pull/6565/commits/2bd5ad6ca5c5b25553e894d0322bda23b43c1fc4:
```
error[E0428]: the name `HAB_THREAD_ALIVE_THRESHOLD_SECS` is defined multiple times
   --> components/common/src/lib.rs:162:5
    |
162 | /     habitat_core::env_config_duration!(ThreadAliveThreshold,
163 | |                                        HAB_THREAD_ALIVE_THRESHOLD_SECS,
164 | |                                        Duration::from_secs(5 * 60));
    | |____________________________________________________________________^ `HAB_THREAD_ALIVE_THRESHOLD_SECS` redefined here
165 | /     habitat_core::env_config_duration!(ThreadAliveCheckDelay,
166 | |                                        HAB_THREAD_ALIVE_THRESHOLD_SECS,
167 | |                                        Duration::from_secs(60));
    | |________________________________________________________________- previous definition of the type `HAB_THREAD_ALIVE_THRESHOLD_SECS` here
    |
    = note: `HAB_THREAD_ALIVE_THRESHOLD_SECS` must be defined only once in the type namespace of this module
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

christophermaier: This is _really_ slick!
tenor-73439784

components/core/src/env.rs
```diff
@@ -128,6 +128,19 @@ macro_rules! env_config_duration {
     };
 }
 
+#[macro_export]
+macro_rules! env_config_int {
```
christophermaier: It would be nice to add some documentation about what this does, since a quick glance at the implementation doesn't really help all that much.

components/butterfly/src/server/pull.rs
```diff
@@ -59,7 +63,9 @@ impl Pull {
             let msg = match socket.recv_msg(0) {
                 Ok(msg) => msg,
                 Err(e) => {
-                    error!("Error receiving message: {:?}", e);
+                    if e != zmq::Error::EAGAIN {
```
christophermaier: I think that it _might_ be useful to add a comment here about why we're doing this. If you're not aware of this current deadlock monitoring situation, you could be forgiven for wondering what the heck we're doing here.

Also, somebody that's not catching the fact that this particular channel is a ZMQ one, as opposed to one of the the increasing number of Rust channels in the rest of the codebase might not realize that there's an inherent timeout on this `recv` message.

I know this is planning _way_ ahead, but I see a day in our future when we can drop ZMQ for pure Rust everywhere, and it would be good to make sure that the intention of this change is very clear, so we can faithfully translate it.

Based on my ZMQ knowledge, this looks good. @markan would be a good second opinion to get here to confirm, though.

markan: It looks reasonable to me. I assume this means you will be limiting the maximum value for the RecvTimeoutMillis to be less than the maximum heartbeat interval?  The default 1 minute heartbeat check and 5 s Recv timeout should be fine, but might as well block a footgun.

components/core/src/env.rs
baumanj: Good call. I'll add docs to the macro definition

components/butterfly/src/server/pull.rs
christophermaier: oooh, that's a great point @markan

baumanj: I don't think it's worth exposing `ThreadAliveThreshold` (which is private to the `habitat_common::sync` module) and making this more complicated. The defaults should work well and the consequences of setting the timeout too high relative to `ThreadAliveThreshold` would just be some spurious logging.

To account for the case that we may want to disable the receive timeout altogether (by setting it to the default of -1), I'll add a check for that value and opt this thread out of the liveliness checking altogether.

If we run into problems, we can always make this more complex later (perhaps by letting the calling thread specify how long must pass without a heartbeat to consider it dead), but for now I'd like to keep it simple.

baumanj: Does https://github.com/habitat-sh/habitat/pull/6565/commits/83d38218a723c2dc8758817c709dfde6bea7f65c work for y'all?

components/core/src/env.rs
baumanj: Done in Does https://github.com/habitat-sh/habitat/pull/6565/commits/83d38218a723c2dc8758817c709dfde6bea7f65c

components/butterfly/src/server/pull.rs
christophermaier: tenor-242646827

Ship it!

PR #6568, opened 2019-05-16
Implement Site field in EventMetadata message

afiune:
The user will be able to set this parameter through an environment
variable or an option in the habitat CLI, just as you will be able to do
with application and environment.

Ticket: https://chefio.atlassian.net/browse/A2-838

Signed-off-by: Salim Afiune <afiune@chef.io>


components/sup/src/event/types.rs
```diff
@@ -75,6 +75,7 @@ impl EventCore {
                         fqdn:          self.fqdn.clone(),
                         application:   self.application.clone(),
                         environment:   self.environment.clone(),
+                        site:          self.site.clone().unwrap_or_default(),
```
afiune: I couldn't leave the field `site` as an `Option<String>` because the proto message requires you to have it as a `String`. 🤔 I wonder if this is the best way to do this optional field here.

afiune: If I change this to be only `self.site.clone()` then I have the following compile error:
```
error[E0308]: mismatched types
  --> components/sup/src/event/types.rs:78:40
   |
78 |                         site:          self.site.clone(),
   |                                        ^^^^^^^^^^^^^^^^^ expected struct `std::string::String`, found enum `std::option::Option`
   |
   = note: expected type `std::string::String`
              found type `std::option::Option<std::string::String>`
```

christophermaier: This is a "feature" of v3 protobufs... the default value of a String is `""`, but v3 protobuf doesn't have a way to distinguish between an unset value and a default value (that it doesn't transmit over the wire).

In our internal types we can use `Option`, and then convert to `""` at the protobuf layer when necessary.

components/hab/src/cli.rs
```diff
@@ -1755,6 +1762,46 @@ mod tests {
...
+                "MY_TOKEN",
+                "--event-stream-url",
+                "127.0.0.1:4222",
+                "--event-stream-site",
+                "",
+            ]);
+            assert!(matches.is_err());
+            let error = matches.unwrap_err();
+            assert_eq!(error.kind, clap::ErrorKind::ValueValidation);
+        }
```
baumanj: It's awesome that there are tests!

A couple of suggestions for making them more focused and robust:
1. Since you're only testing one arg, extract the definition of `Arg::with_name("EVENT_STREAM_SITE")` to it's own function so you can focus on what you're testing and not all the setup required. There are 17 lines in each test and 10 of them aren't really relevant to what's being tested.
1. Try to avoid over-constraining your tests to details of clap that aren't relevant. Should that library change the name of structure of its errors, this test will break, but all we really care about is that `--event-stream-site` requires a non-empty value, so asserting that getting matches results in an error should be sufficient.
1. If we're testing that no value or an empty value is an error, we should also be testing that a non-empty value and omitting the option entirely are non-errors.

By the way, I realize you're following the template of the other tests that are already here, but I haven't reviewed those previous or I missed them.

components/sup/src/event/types.rs
afiune: Which is what I just did, right? Also I think I asked you on my previous PR so I think we are good. 💯

components/hab/src/cli.rs
```diff
@@ -1213,6 +1213,13 @@ fn maybe_add_event_stream_options(mut app: App<'static, 'static>,
                                                         .required(true)
                                                         .takes_value(true)
                                                         .validator(non_empty));
+        app = app.arg(Arg::with_name("EVENT_STREAM_SITE").help("The name of the site where this \
+                                                                Supervisor is running. It is used \
+                                                                for event stream purposes.")
+                                                         .long("event-stream-site")
+                                                         .required(false)
```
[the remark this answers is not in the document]
afiune: Yeah makes sense. I think I sow one or two more like this one in the code, I’ll update them as well!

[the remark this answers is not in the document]
afiune: Updated! 💥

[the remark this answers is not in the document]
baumanj: If you're updating the other ones, two other little tweaks would be nice:
1. Assuming there will only ever be the one caller, change
    ```rust
    maybe_add_event_stream_options(sub, feature_flags)
    ```
    to
    ```rust
    if feature_flags.contains(FeatureFlag::EVENT_STREAM) {
        add_event_stream_options(sub)
    } else {
        sub
    }
    ```
1. Within `[maybe_]add_event_stream_options`, change
    ```rust
    fn maybe_add_event_stream_options(mut app: App<'static, 'static>,
                                          feature_flags: FeatureFlag)
                                          -> App<'static, 'static> {
        if feature_flags.contains(FeatureFlag::EVENT_STREAM) {
            app = app.arg(Arg::with_name("EVENT_STREAM_APPLICATION") ...
            app = app.arg(Arg::with_name("EVENT_STREAM_ENVIRONMENT") ...
            app = app.arg(Arg::with_name("EVENT_STREAM_URL") ...
            …
        }

        app
    }
    ```
    to
    ```rust
    fn add_event_stream_options(app: App<'static, 'static>) -> App<'static, 'static> {
	    app.arg(Arg::with_name("EVENT_STREAM_APPLICATION") ...
	       .arg(Arg::with_name("EVENT_STREAM_ENVIRONMENT") ...
	       .arg(Arg::with_name("EVENT_STREAM_URL") ...
	}
    ```

Issue #6555, opened 2019-05-14
Remove description implementation for std::error::Error


christophermaier:
The `description` function for the `std::error::Error` trait has been ["soft-deprecated"](https://doc.rust-lang.org/1.34.0/std/error/trait.Error.html#method.description).

Historically, this implementation has always been a bit "noisy" in our codebase, since it largely duplicates the `Display` trait implementation for our error types.

The docs say that we no longer need to implement `description`, and should favor implementing `Display` on new types. Since we already do the latter, we should now to the former and remove all `description` implementations in all the crates of this repository.

menoude:
Hi,
I'm looking for a first contribution to a project, do you only need to remove the method implementation everywhere or is there a need for some replacement? If it's the former, I can try to handle that :)

christophermaier:
@menoude Yes, this is just about removing the current method implementations; we don't need anything to replace them.

(Rather, it _should_ be just about removing the current code... I don't think we're explicitly using the `description` method ourselves, and have only implemented it to conform to the then-current definition of the `Error` trait. If you get into it and find otherwise, let's chat 😅 )

Thanks for your interest in Habitat; we're very glad to have you around!

PR #6580, opened 2019-05-23
Remove fn description definitions in every error::Error impl block

menoude:
This deletes every implementation of the `fn description(&self) -> &str` method in the `impl error:Error for ... {}` blocks. The description method [has been recently soft-deprecated ](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.description) and its implementation is not needed anymore when assigning the std::error::Error trait.
This should solve #6555.
@christophermaier


christophermaier:
@menoude Thank you so much for this contribution! It looks like we need a Developer's Certificate of Origin (DCO) sign-off on the commit, though, before we can legally accept it. You can see how to do this [here](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Alternatively, you could invoke the [obvious fix policy](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#chef-obvious-fix-policy).

As soon as we get that, we'll be happy to merge this. Let us know if you have any additional questions.

christophermaier:
@menoude Any update on this? If we don't hear back soon, we'll have to close this 😢

baumanj:
Feel free to reopen if you'd like to continue work on this, @menoude.

PR #6581, opened 2019-05-23
Add limit option to hab pkg search

davidMcneil:
This adds a `--limit` option to `hab pkg search` that lets users set an arbitrary limit when searching for packages. Before it was restricted to 50. 

This uses the existing builder api. Future work could improve the builder api cleaning up the logic of the cli. For example, it is currently necessary to do multiple http requests to get the results.

I also made output that is not directly printing package idents pipe to stderr instead of stdout. I have found this makes tools easier to use in conjunction with other command line utilities.

Resolves [#56](https://github.com/habitat-sh/builder/issues/56)

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -216,6 +216,52 @@ impl BuilderAPIClient {
     }
 }
 
+impl BuilderAPIClient {
+    fn seach_package_with_range(&self,
+                                search_term: &str,
+                                token: Option<&str>,
+                                range: usize)
+                                -> Result<(PackageResults<PackageIdent>, bool)> {
+        let req = self.0
+                      .get_with_custom_url(&package_search(search_term), |url| {
+                          url.set_query(Some(&format!("range={:?}&distinct=true", range)));
+                      });
+        let mut res = self.maybe_add_authz(req, token).send()?;
+        let mut encoded = String::new();
+        res.read_to_string(&mut encoded)
+           .map_err(Error::BadResponseBody)?;
+        let package_results = serde_json::from_str(&encoded)?;
+        match res.status {
+            StatusCode::Ok => Ok((package_results, false)),
+            StatusCode::PartialContent => Ok((package_results, true)),
+            _ => Err(err_from_response(res)),
+        }
+    }
+
+    fn search_package_impl(&self,
+                           search_term: &str,
+                           limit: usize,
+                           token: Option<&str>,
+                           search_with_range: impl Fn(&BuilderAPIClient,
+                              &str,
+                              Option<&str>,
+                              usize)
+                              -> Result<(PackageResults<PackageIdent>, bool)>)
+                           -> Result<(Vec<PackageIdent>, usize)> {
+        let mut packages = Vec::new();
+        loop {
+            let (mut package_results, more_to_come) =
+                search_with_range(self, search_term, token, packages.len())?;
+            packages.append(&mut package_results.data);
+
+            if packages.len() >= limit || !more_to_come {
+                packages.truncate(limit);
+                return Ok((packages, package_results.total_count as usize));
+            }
+        }
+    }
+}
+
 impl BuilderAPIProvider for BuilderAPIClient {
     type Progress = Box<dyn DisplayProgress>;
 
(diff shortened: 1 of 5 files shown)


components/hab/src/main.rs
```diff
@@ -784,8 +784,12 @@ fn sub_pkg_provides(m: &ArgMatches<'_>) -> Result<()> {
 fn sub_pkg_search(m: &ArgMatches<'_>) -> Result<()> {
     let url = bldr_url_from_matches(&m)?;
     let search_term = m.value_of("SEARCH_TERM").unwrap(); // Required via clap
+    let limit = m.value_of("LIMIT").unwrap().parse().unwrap(); // Required via clap
```
baumanj: I learned my lesson awhile back about using `expect` even when failure should be impossible, since it makes zeroing in on the issue in release builds *way* easier. I'd suggest changing both this and the above line.
```suggestion
    let limit = m.value_of("LIMIT")
                 .expect("required opt LIMIT")
                 .parse()
                 .expect("valid LIMIT");
```

components/builder-api-client/src/lib.rs
```diff
@@ -1104,21 +1104,35 @@ impl Client {
...
+                        serde_json::from_str(&encoded)?;
+                    packages.extend(package_results.data);
+                    total = package_results.total_count;
+                    status_code = res.status;
+                }
+                _ => return Err(err_from_response(res)),
+            };
         }
+        packages.truncate(limit);
+        Ok((packages, total))
```
baumanj: In order to avoid the nonsensical `isize` leaking out to callers, I think it would be better to clean thing up here (if we can't change `PackageResults` itself):
```suggestion
        total = if package_results.total_count > 0 {
           packages.len().max(package_results.total_count as usize)
        } else {
            packages.len()
        };
        Ok((packages, total))
```

components/builder-api-client/src/lib.rs
```diff
@@ -1104,21 +1104,35 @@ impl Client {
...
+                              url.set_query(Some(&format!("range={:?}&distinct=true",
+                                                          packages.len())));
+                          });
+            let mut res = self.maybe_add_authz(req, token).send()?;
+            match res.status {
+                StatusCode::Ok | StatusCode::PartialContent => {
+                    let mut encoded = String::new();
+                    res.read_to_string(&mut encoded)
+                       .map_err(Error::BadResponseBody)?;
+                    let package_results: PackageResults<PackageIdent> =
```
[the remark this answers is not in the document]
chefsalim: There's likely a big difference between debug and release timings... from what I've found in testing perf in the past

components/builder-api-client/src/lib.rs
chefsalim: The isize is due to PG/Diesel.. I _think_ it might be fine to change PackageResult itself.. but it's probably safest to just adjust the return here to be usize.  It should also be safe to cast to usize as we're not actually ever going to return negative counts.

components/hab/src/command/pkg/search.rs
```diff
@@ -16,9 +16,10 @@ pub fn start(st: &str, bldr_url: &str, token: Option<&str>) -> Result<()> {
                     println!("{}/{}", p.origin, p.name);
                 }
             }
-            if more {
-                println!("Search returned too many items, only showing the first {}",
-                         packages.len());
+            if packages.len() != total as usize {
```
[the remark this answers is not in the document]
chefsalim: I think the cast is safe, but Jon's suggestion is better overall

components/hab/src/main.rs
chefsalim: 👍

components/builder-api-client/src/lib.rs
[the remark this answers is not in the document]
davidMcneil: @baumanj I was unable to reproduce these results both versions ran at around 4.25 seconds on my machine. How are you doing the test?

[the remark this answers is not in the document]
baumanj: I was just running
```sh
time env RUST_LOG= ./target/debug/hab pkg search --limit=1000 core
```

But the proper way to compare would be to add [benchmark tests](https://doc.rust-lang.org/unstable-book/library-features/test.html) that compare the performance. Probably one that includes the network call and one that doesn't would be the most illustrative.

components/builder-api-client/src/builder.rs
```diff
@@ -1142,4 +1175,100 @@ mod tests {
...
+
+        let searcher = seach_generator(sample_data.as_slice(), 10);
+        let r = client.search_package_impl("does_not_exist", 100, None, searcher)
+                      .expect("valid search");
+        assert_eq!(r.0.iter().map(|i| i.name.clone()).collect::<Vec<_>>(),
+                   Vec::<String>::new());
+        assert_eq!(r.1, 0);
+    }
+
+    #[ignore]
```
baumanj: Why is this ignored? I'm guessing it's because the test tends to take over 60s, and gives the `package_search_large has been running for over 60 seconds` warning, but on my machine it completes in ~90 seconds, so I think it's fine to put it in the normal suite. It doesn't ever fail, does it?

I'd say `#[ignored]` is for things that are inconsistent, but we'd still like to fix them up. Tests like that won't be noticed if they catch a regression, I'd want to know if this started failing.

davidMcneil: No, it does not every fail. This relates to our conversation on benchmarking. Ideally, this would be a benchmark and could be run separately from the tests. I like tests to have as fast a feedback loop as possible, but this is probably abusing the `ignore` attribute.

bixu:
🎉

PR #6589, opened 2019-05-24
Two minor fixes for vagrant based bootstrap

markan:
1) Pass accept license through sudo.

2) Fixup bad path for install.sh. This seems reasonable for the vagrant
environment, and the docker environment still seems to work for me. Tested this on MacOS and Ubuntu 18.10 hosts. This might need a bit more review because I'm not sure how wide ranging the usage of this script is.

Thanks to @smacfarlane for his help with this and getting my dev-env bootstrapped.

### DIFF
--- a/support/linux/install_dev_9_linux.sh
+++ b/support/linux/install_dev_9_linux.sh
@@ -45,5 +45,5 @@ else
   sudo -E addgroup --system hab || true
 fi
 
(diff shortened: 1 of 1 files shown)


markan:
I don't have merge rights here; how should I proceed to finish this up?

PR #6590, opened 2019-05-24
Decouple logging concerns from Supervisor errors

### LINKED ISSUE #6585: Decouple logging concerns from Supervisor Error implementation
As detailed in #6584, the `StructuredOutput` struct replicates some concerns of a logging framework by capturing location information for an event.

It is _also_ used in the `Display` implementations of our Supervisor errors, which needlessly couples the printing of error information with logging concerns (if we really want location information, we can just configure our logging framework to output that information when we log the error).

This is of particular concern in light of #6555, which will require us to use our `Display` implementation for printing errors, which will effectively result in "double logging": once for the log message itself, and once for the formatted error message it embeds.
     
We should decouple these concerns. 

This will in turn make implementing #6559 easier and also more valuable.

christophermaier:
Fixes #6585

### DIFF
--- a/components/sup/src/util/pkg.rs
+++ b/components/sup/src/util/pkg.rs
@@ -1,6 +1,5 @@
 use crate::{error::{Error,
-                    Result,
-                    SupError},
+                    Result},
             PRODUCT,
             VERSION};
 use habitat_common::{self,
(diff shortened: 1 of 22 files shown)


components/sup/src/util/pkg.rs
```diff
@@ -37,33 +36,41 @@ pub fn install<T>(ui: &mut T,
...
+                                                     // current
+                                                     // function, though, should always have a
+                                                     // channel. We
+                                                     // should
+                                                     // push this "Option-ness" as far down the
+                                                     // stack as we
+                                                     // can,
+                                                     // with the ultimate goal of eliminating it
+                                                     // altogether.
+                                                     channel,
```
baumanj: Maybe re-wrap these comments so they're less awkwardly line-broken?

christophermaier: oof, yeah, that's bad... will fix.

PR #6592, opened 2019-05-24
Macroize env config

baumanj:
Convert all the implementations of the `env::Config` trait to use the boilerplate-reducing macros. Also, make various cleanups along the way. See individual commits for more details.


components/core/src/lib.rs
```diff
@@ -25,42 +25,30 @@ use serde_derive::{Deserialize,
 pub use crate::os::{filesystem,
                     users};
 
+/// A type which can't be instantiated
+/// Use this when a generic requires a type which will never be used.
+/// For example, FromStr::Err on an infallible conversion
+pub enum Impossible {}
```
christophermaier: You can use [std::convert::Infallible](http://doc.rust-lang.org/1.34.1/core/convert/enum.Infallible.html) for this, instead.

baumanj: Thank you!

I knew I had seen that somewhere but I couldn't find it again. Updated in https://github.com/habitat-sh/habitat/pull/6592/commits/15d6f630797d6defdbfe240c96e751ca24667289.

christophermaier: I kinda wish they'd gone with a slightly different name, though....

tenor-41037299

PR #6593, opened 2019-05-25
Add toolchain file and use its version instead of "stable"

baumanj:
Since we don't control when "stable" changes, this gives more predictable behavior across platforms and machines where "stable" may mean different versions.

This should address the failure we're currently seeing where ["stable" is resolving to 1.35 on Linux and 1.34 on Windows in clippy runs](https://buildkite.com/chef/habitat-sh-habitat-master-verify/builds/1996) resulting in a situation where it's impossible for both to pass.

See also https://github.com/rust-lang/rustup.rs#the-toolchain-file


baumanj:
> It may be worth incorporating this into the version of `core/rust` that we pull into the Habitat builds, too, which would ensure that we're truly consistent everywhere. That could be another PR, though.

Agreed on both counts. I'll create an issue to track both phases of this work.

baumanj:
See https://github.com/habitat-sh/habitat/issues/6594

mwrock:
I think we should handle clippy runs the same as tests: Use the most recent hab rust package for `cargo` this way we don't have to track that in a separate file here but instead it just syncs to core-plans. Note how the linux based tests `--binlink` core/rust and windows does something similar by adding the core/rust bin to the path. I think we should do the same for the clippy scripts.

baumanj:
> I think we should handle clippy runs the same as tests: Use the most recent hab rust package for `cargo` this way we don't have to track that in a separate file here but instead it just syncs to core-plans. Note how the linux based tests `--binlink` core/rust and windows does something similar by adding the core/rust bin to the path. I think we should do the same for the clippy scripts.

I'll look into this approach. I'm 💯 on using the `core/rust` package for our installs, but I don't think we necessarily want to make all our infra use the newest stable package as the indicator of what version we build/test habitat with, or else we have a similar problem (though more tightly scoped) of unintended consequences. Just because we promote a new `core/rust` package version doesn't mean that everything in the `habitat` repo is ready to do that upgrade. And I don't think we want to block updating the `core/rust` package until the `habitat` repo is ready for update since other consumers may want to use `core/rust`, and I don't think that's a good reason to delay them.

Even if we're using habitat to install `core/rust` rather than `rustup`, I thinking changing* rust versions for `habitat` consumption should be an explicit operation. I do still like the idea of using habitat to install our rust toolchain though, since I don't foresee the `habitat` repo ever needing to get *ahead* of `core/rust`.

Finally, since most habitat development uses a `rustup`-based installation locally, a rust-toolchain file leverages a standard practice to help increase parity between local development and CI.

\* Should always be an upgrade, but imagine what a mess it would be if habitat needed to downgrade the rust version it used and it was just based on the latest stable `core/rust` package.

scotthain:
As long as we are pinning the rust version consistently all over I'm good with this - I'd also advocate for *not* binlinking and being very explicit as to which rust binary/toolchain/etc we are using (which i know kinda goes against the rust way of things). Either that or maybe setting separate RUST_HOME and CARGO_HOME directories per task? I don't know all the ramifications of doing that but might be worth some thinking.

christophermaier:
Just for reference, right now all our packages in this repo are using `core/rust` as their build dependency, and not something like `core/rust/1.34.0`. This is what I was alluding to in https://github.com/habitat-sh/habitat/pull/6593#pullrequestreview-242695383... if that version is pulled from a toolchain file, then I think we get the best of all worlds.

support/ci/shared.ps1
```diff
@@ -20,7 +20,7 @@ function Install-Rustup($Toolchain) {
     if (get-command -Name rustup.exe -ErrorAction SilentlyContinue) {
         Write-Host "rustup is currently installed"
         rustup set default-host x86_64-pc-windows-msvc
-        rustup default stable-x86_64-pc-windows-msvc
+        rustup default "$(Rust-Toolchain)-x86_64-pc-windows-msvc"
```
mwrock: you will want to keep this `stable` because there is no guarantee that `$(Rust-Toolchain)-x86_64-pc-windows-msvc` will exist until `Install-RustToolchain` runs. Making it `stable` will not interfere with our desires to execute with the specific toolchain we are at since all `cargo` commands specify that toolchain. During the docker work, I found that setting the default to stable prevented some weird errors when running `Install-RustToolchain` when there was no default set at all which seems to be the case in our docker images.

support/ci/shared.ps1
```diff
@@ -30,6 +30,10 @@ function Install-Rustup($Toolchain) {
     }
 }
 
+function Rust-Toolchain {
```
mwrock: The idiomatic Powershell naming here would be `Get-RustToolChain`. Ideally you want to only use `verb`s listed in `get-verb`.

support/ci/shared.ps1
baumanj: That's very helpful; will fix.

baumanj:
Thanks for the review, @mwrock! I think all the issues should be resolved with https://github.com/habitat-sh/habitat/pull/6593/commits/172f868b27a31069206a6b994a136b35d5b94095

support/ci/shared.ps1
```diff
@@ -30,6 +30,10 @@ function Install-Rustup($Toolchain) {
     }
 }
 
+function Get-Rust-Toolchain {
```
mwrock: actually `Get-RustToolchain` (just 1 dash). Might want to make it `Get-Toolchain` to align with the bash function naming which works.

baumanj: Ah, good point. I went with `Rust` in the name to match `Install-RustToolchain` in the same file. Any preference between `Get-RustToolchain` and `Get-Toolchain`?

mwrock: I like `Get-Toolchain`

baumanj: [Done](https://github.com/habitat-sh/habitat/pull/6593/commits/e17eccc6aed51c8486fe901d987d9ba1e4f4e976)!

PR #6603, opened 2019-05-30
Swap windows jobs to docker (as much as possible)

scotthain:

There are a few jobs that still fail.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -49,12 +49,12 @@ steps:
       - .\test\run_clippy.ps1 stable .\test\unexamined_lints.txt .\test\allowed_lints.txt .\test\lints_to_fix.txt .\test\denied_lints.txt
     agents:
       queue: 'default-windows-privileged'
-    plugins:
-      docker#v3.2.0:
-        image: "chefes/buildkite-windows"
-        shell: [ "powershell", "-Command" ]
-        always-pull: true
-        propagate-environment: true
+    # plugins:
+    #   docker#v3.2.0:
+    #     image: "chefes/buildkite-windows"
+    #     shell: [ "powershell", "-Command" ]
+    #     always-pull: true
+    #     propagate-environment: true
     timeout_in_minutes: 25
 
 #######################################################################
@@ -410,12 +410,13 @@ steps:
       - ./test/run_cargo_test.ps1 butterfly -Nightly -Features "lock_as_rwlock deadlock_detection" -TestOptions "--test-threads=1"
     agents:
       queue: 'default-windows-privileged'
-    plugins:
-      docker#v3.2.0:
-        image: "chefes/buildkite-windows"
-        shell: [ "powershell", "-Command" ]
-        always-pull: true
-        propagate-environment: true
+    # TODO: enable once we've figured out why this fails in docker
+    # plugins:
+    #   docker#v3.2.0:
+    #     image: "chefes/buildkite-windows"
+    #     shell: [ "powershell", "-Command" ]
+    #     always-pull: true
+    #     propagate-environment: true
     timeout_in_minutes: 20
     retry:
       automatic:
@@ -458,12 +459,13 @@ steps:
       - ./test/run_cargo_test.ps1 core
     agents:
       queue: 'default-windows-privileged'
-    plugins:
-      docker#v3.2.0:
-        image: "chefes/buildkite-windows"
-        shell: [ "powershell", "-Command" ]
-        always-pull: true
-        propagate-environment: true
+    # TODO: enable once we've figured out why this fails in docker
+    # plugins:
+    #   docker#v3.2.0:
+    #     image: "chefes/buildkite-windows"
+    #     shell: [ "powershell", "-Command" ]
+    #     always-pull: true
+    #     propagate-environment: true
     timeout_in_minutes: 20
     retry:
       automatic:
@@ -474,12 +476,13 @@ steps:
       - ./test/run_cargo_test.ps1 hab
     agents:
       queue: 'default-windows-privileged'
-    plugins:
-      docker#v3.2.0:
-        image: "chefes/buildkite-windows"
-        shell: [ "powershell", "-Command" ]
-        always-pull: true
-        propagate-environment: true
+    # TODO: enable once we've figured out why this fails in docker
+    # plugins:
+    #   docker#v3.2.0:
+    #     image: "chefes/buildkite-windows"
+    #     shell: [ "powershell", "-Command" ]
+    #     always-pull: true
+    #     propagate-environment: true
     timeout_in_minutes: 20
     retry:
       automatic:
@@ -635,12 +638,13 @@ steps:
       - ./test/run_cargo_test.ps1 sup-protocol
     agents:
       queue: 'default-windows-privileged'
-    plugins:
-      docker#v3.2.0:
-        image: "chefes/buildkite-windows"
-        shell: [ "powershell", "-Command" ]
-        always-pull: true
-        propagate-environment: true
+    # TODO: enable once we've figured out why this fails in docker
+    # plugins:
+    #   docker#v3.2.0:
+    #     image: "chefes/buildkite-windows"
+    #     shell: [ "powershell", "-Command" ]
+    #     always-pull: true
+    #     propagate-environment: true
     timeout_in_minutes: 20
     retry:
       automatic:
@@ -651,12 +655,13 @@ steps:
       - ./test/run_studio_test.ps1 "studio-from-source"
     agents:
       queue: 'default-windows-privileged'
-    plugins:
-      docker#v3.2.0:
-        image: "chefes/buildkite-windows"
-        shell: [ "powershell", "-Command" ]
-        always-pull: true
-        propagate-environment: true
+    # TODO: enable once we've figured out why this fails in docker
+    # plugins:
+    #   docker#v3.2.0:
+    #     image: "chefes/buildkite-windows"
+    #     shell: [ "powershell", "-Command" ]
+    #     always-pull: true
+    #     propagate-environment: true
     timeout_in_minutes: 5
     retry:
       automatic:
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -49,13 +49,13 @@ steps:
       - .\test\run_clippy.ps1 stable .\test\unexamined_lints.txt .\test\allowed_lints.txt .\test\lints_to_fix.txt .\test\denied_lints.txt
     agents:
       queue: 'default-windows-privileged'
-    # plugins:
-    #   docker#v3.2.0:
-    #     image: "chefes/buildkite-windows"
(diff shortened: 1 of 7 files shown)


support/ci/shared.ps1
```diff
@@ -50,6 +50,11 @@ function Install-Habitat {
         Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1')) | out-null
     }
 
+    choco source Add -Name artifactory -Source http://artifactory-preprod.chef.co/api/nuget/chocolatey
+    choco source enable --name artifactory
+    choco source disable --name chocolatey
```
mwrock: we should wral these 3 lines in `if($env:BUILDKITE) {}`

mwrock:
I've gone and made a few changes to allow tests to succeed in a windows container environment. Here is a summary:

* Remove use of `canonicalize` in `fs` test code. Its a known issue that this method is not friendly to certain Windows file systems including docker mounts.
* Interactive `promote` test should not try to use `cat` as a UI editor. I do not know how this could have ever worked and fails for me locally outside of docker.
* Use `copy` and `remove_file` instead of `rename` when generating the protocols. Rename will not work over a mount.
* set the default rustup host to `x86_64-pc-windows-msvc`. It was `x86_64-pc-windows-gnu` which was breaking some butterfly test compilations that was trying to find `gcc.exe`.
* Make sure the studio tests have needed prereqs (habitat and nuget provider)

.expeditor/verify.pipeline.yml
```diff
@@ -460,12 +460,12 @@ steps:
     agents:
       queue: 'default-windows-privileged'
     # TODO: enable once we've figured out why this fails in docker
```
scotthain: Extra comment

scotthain:
This looks great @mwrock - I think you need to reapprove it and then I can merge it if you're ok with it. Thanks for picking up the ball on this!!

components/hab/src/command/bldr/job/promote.rs
```diff
@@ -240,12 +243,25 @@ mod test {
                                                    "Properly formated timestamp".to_string(),
                                                project_name: "Test Project".to_string(),
                                                target:       "x86_64-linux".to_string(), };
+
+        #[cfg(windows)]
+        let tmp_file_path = {
+            let mut t = env::temp_dir();
```
baumanj: Why not use the [`tempfile`](https://docs.rs/tempfile/3.0.8/tempfile/) crate here? It would avoid the need to manually remove it later (which could get missed if `get_ident_list` fails)

baumanj: After looking at `test_get_ident_list_interactive` for a bit, I think we should remove this test. It's not actually testing the interactive-specific code (and this isn't really the right level to do such testing anyhow), so it's entirely redundant with `test_get_ident_list`, but adds unnecessary complexity.

components/sup-protocol/build.rs
```diff
@@ -38,9 +38,10 @@ fn generate_protocols() {
           .expect("protocols");
     compile_proto_impls(&protocol_files(), &protocol_includes()).expect("protocol-impls");
     for file in generated_files() {
-        fs::rename(&file,
-                   format!("src/generated/{}",
-                           file.file_name().unwrap().to_string_lossy())).unwrap();
+        fs::copy(&file,
+                 format!("src/generated/{}",
+                         file.file_name().unwrap().to_string_lossy())).unwrap();
+        fs::remove_file(&file).unwrap();
```
baumanj: Why is this change necessary? It seems like we could end up failing to do the `remove_file` if the `copy` fails, and then I'm not sure what the outcome would be for the build script if we have stale files sitting around.

.expeditor/verify.pipeline.yml
```diff
@@ -48,7 +48,13 @@ steps:
     command:
       - .\test\run_clippy.ps1 stable .\test\unexamined_lints.txt .\test\allowed_lints.txt .\test\lints_to_fix.txt .\test\denied_lints.txt
     agents:
-      queue: 'single-use-windows-privileged'
+      queue: 'default-windows-privileged'
```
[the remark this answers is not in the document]
scotthain: All steps are single use, with a brand new isolated container, there is no reuse.

scotthain:
I'm going to defer to @mwrock to answer questions on the rust changes.

components/hab/src/command/bldr/job/promote.rs
mwrock: Cool. I'll remove it, but just to answer why I did not use TempFile: TempFile holds a handle to the file which gets removed on `drop` but I deenen the file handle to be closed bedore calling the file and `close` of TempFile will delete the file.

components/core/src/fs.rs
```diff
@@ -822,8 +821,8 @@ mod test_find_command {
     fn setup_path() {
```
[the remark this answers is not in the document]
mwrock: `locked_env_var` is defined in `common` so we would have to move it to core and adjust all callers which feels like it belongs in a separate PR.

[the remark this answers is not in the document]
baumanj: Yeah, it's overdue to be moved into `core` (probably common/core should be merged entirely), but I think it's worth doing now since these tests are prone to non-deterministic failures without it. A separate PR is good, but it could merge before this one so we can base the fix on it.

How about taking the dependency-injection approach? That's the preferable solution in my view and is appropriate to this PR scope.

components/sup-protocol/build.rs
mwrock: The `expect sounds good. Windows cannot rename a file across different drives. The error emitted by rename is `The system cannot move the file to a different disk drive.`

components/core/src/fs.rs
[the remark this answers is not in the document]
scotthain: Including that in here feels to me kinda like scope creep for this PR but I'll defer to you both on it!

components/hab/src/command/bldr/job/promote.rs
baumanj: Sounds good; I think removal is the best option. For my own understanding, could we have used `TempFile` (or probably `NamedTempFile::new_in` since it needed to be called `type.bat`) if we made the scope of `tmp_file_path` the entire function? Something like:
```rust
        #[cfg_attr(not(windows), allow(unused_variables))]
        let tmp_file_path: std::path::PathBuf;
        #[cfg(windows)]
        {
            // assign `tmp_file_path`
        }
```

components/sup-protocol/build.rs
baumanj: Oh! The rename doesn't work *across* mounts. I get it now. But what I don't understand is why this *is* a move across mounts. Why are the files generated in one mount and then moved to another? It looks like we can choose where we do the generation, so why not do it within the same mount?

components/core/src/fs.rs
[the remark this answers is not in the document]
baumanj: I agree that moving `locked_env_var` deserves its own PR if we go that way. If we use dependency injection instead, we can make those changes in this PR and leave moving `locked_env_var` for later, but either way I think we should not leave this in a subtly broken state which may crop up some unknown time in the future.

components/hab/src/command/bldr/job/promote.rs
mwrock: yeah that would have worked as long as we called `persist` to close it without deleting it but then we  would still needed to delete is when done.

components/sup-protocol/build.rs
mwrock: Thats a good point. It was generating them in `OUT_DIR`. I just changed it to generate directly to `src/generated` and it works fine and also removes several more lines of code which is always a win!

components/core/src/fs.rs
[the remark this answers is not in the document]
mwrock: In this case, I think even if multiple tests concurrently change this, its harmless. worse that will happen id the fixtures will get added to the path multiple times.

components/hab/src/command/bldr/job/promote.rs
baumanj: The original code is gone, so I can't confirm this, but wasn't `tmp_file_path` deleted before the end of that function? Unlike [`tempfile`](https://docs.rs/tempfile/3.0.8/tempfile/fn.tempfile.html), which
> will be automatically removed by the OS when the last handle to it is closed

[`NamedTempFile`](https://docs.rs/tempfile/3.0.8/tempfile/struct.NamedTempFile.html) stays around until it goes out of scope, so I think that would've fit the use case here. Just FYI for the future.

components/core/src/fs.rs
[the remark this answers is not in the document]
baumanj: I'm not so sure we can make that strong an assumption about the way `env::set_var` works. The documentation says "concurrent access to environment variables is safe in Rust", but that doesn't mean that it will do what you expect. I'm not sure we're guaranteed one thread accessing this in the midst of another thread setting it couldn't result in an empty result.

In any case, it's not a huge deal, but it's an error-prone pattern to rely on global state like this, so when we can switch to something more robust (such as injecting global dependencies as arguments) without a huge hassle, I think we'd be well served to do so.

components/sup-protocol/build.rs
```diff
@@ -77,13 +59,7 @@ fn protocol_includes() -> Vec<String> { vec!["protocols".to_string()] }
...
     where P: AsRef<Path>
 {
-    let target: PathBuf = env::var_os("OUT_DIR").ok_or_else(|| {
-                                                    Error::new(ErrorKind::Other,
-                                                               "OUT_DIR environment variable is \
-                                                                not set")
-                                                })?
-                                                .into();
-
+    let target = PathBuf::from("src/generated");
```
christophermaier: Why did this protobuf stuff need to get changed in this PR?

(Also, I swear that I've changed this in the past...)

mwrock: @christophermaier see discussion above with @baumanj . Basically it started with the fact that we can't use `rename` to move a file to a mounted directory on windows followed by the question, why don't we just generate the files in place instead of doing a `copy` and `remove_file`

christophermaier: Ah, the switch to Docker cause this... gotcha.

We've fixed this before in other crates, but apparently haven't over here 😦 Also, this entire code generation logic is _waaaaaaaaaaaaaaaaay_ more complex than it should be. I'm going to file an issue to fix it up once and for all.

christophermaier: Filed #6628

components/core/src/fs.rs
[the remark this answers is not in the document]
mwrock: Looking at this I still think it is out of scope. These tests while not ideal has been in place for 3 years without raising stability issues. I'd just not go into refactoring non-test code unrelated to the intent of this PR and get it merged.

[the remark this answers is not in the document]
baumanj: In test code, I'm equally or more worried about false positives, but you're right that we can probably let this one go. The value of getting this merged is quite high.

PR #6609, opened 2019-06-03
Move deadlock error to trace - stop spam

chefsalim:
This error is continuously spamming Builder and making it very hard to debug anything else.

### DIFF
--- a/components/common/src/lib.rs
+++ b/components/common/src/lib.rs
@@ -199,7 +199,7 @@ pub mod sync {
         let heartbeats = &THREAD_HEARTBEATS.lock()
                                            .expect("THREAD_HEARTBEATS poisoned");
         for (name, last_heartbeat) in threads_missing_heartbeat(heartbeats, threshold) {
-            error!("No heartbeat from {} in {} seconds; deadlock likely",
+            warn!("No heartbeat from {} in {} seconds; deadlock likely",
                    name.unwrap_or_else(|| { "unnamed thread".to_string() }),
                    last_heartbeat.elapsed().as_secs());
         }


components/common/src/lib.rs
```diff
@@ -199,7 +199,7 @@ pub mod sync {
         let heartbeats = &THREAD_HEARTBEATS.lock()
                                            .expect("THREAD_HEARTBEATS poisoned");
         for (name, last_heartbeat) in threads_missing_heartbeat(heartbeats, threshold) {
-            error!("No heartbeat from {} in {} seconds; deadlock likely",
+            trace!("No heartbeat from {} in {} seconds; deadlock likely",
```
christophermaier: I think this may be better at `warn`, rather than `trace`.

By default, `error` is the default level that is printed, but this really should be a higher level of urgency than `trace`.

chefsalim: The problem is that warn will still spam when we have debug on, eg, acceptance

christophermaier: FYI, you can configure different modules individually.

For instance:

```
RUST_LOG=debug,habitat_common::sync=error
```

will set everything to `DEBUG`, but set the level to `ERROR` for code coming from the `habitat_common::sync` module.

christophermaier: (I don't think there's a way to selectively _disable_ logging, apart from setting the level, so something at the `error` level seems destined to always log. Moving it to `warn` seems like a good compromise between the "urgency"  of the message and the ability to control when it is output.)

chefsalim: Ok, so changing it to warn and using the RUST_LOG config should work for now (we'll have to confirm after the change is merged that its working as expected).

christophermaier: @chefsalim Turns out I was wrong; you _can_ selectively disable logging:

```
RUST_LOG=debug,habitat_common::sync=off
```
That's buried deep in the code of `env_logger`, but not clearly documented. With that, this PR wouldn't be necessary at all.

tenor-73718303

PR #6610, opened 2019-06-03
Allow clippy::identity_conversion due to clippy bug

davidMcneil:


davidMcneil:
The change this addressed was reverted so this is no longer needed.

PR #6611, opened 2019-06-03
Adds service update started event

gpeers:
This PR fires a service update started event when the supervisor sees that a new hab package is available and thus starts the package update process. Automate uses this event to help determine and display the deployment status of a service.

Note: there's work in progress to refactor the supervisor to perform certain operations, including starting packages, asynchronously. Until this work is done, providing additional deployment events (service update completed, service update failed) is awkward because we don't have a concrete handle on when these events occur without adding more state as a stopgap. I'm thinking it might be best to let Automate derive completed and failure states on its side until we get the supervisor refactored and can implement additional events in the correct way. One possible way Automate might accomplish this would be to compare the update fully-qualified package identifier from the service update started event with the fully-qualified package identifier  from the most recent service started event for a given service. If they match, Automate could assume a successful update. If this is not the case (and a specified period of time has passed), Automate could assume an update failure.

There are nuances to this we'd need to discuss, e.g., multiple entities trying to update the same service. Additionally, depending upon the level of effort on the Automate side, we may wish to revisit adding state on the Habitat side, bumping up the priority of the supervisor work, etc. Another option would be just using the service update started event on the Automate side without trying to derive any additional information. This needs discussion with the Automate team and Product.

Shake 'n' Bake!
image
 
Signed-off-by: Gina Peers <gpeers@chef.io>

### DIFF
--- a/components/sup/Cargo.toml
+++ b/components/sup/Cargo.toml
@@ -51,7 +51,7 @@ prost = "*"
 prost-derive = "*"
 prost-types = "0.5.0" # This is current stable, but for some reason gets pulled in as 0.4.0 if we use "*" (2019-03-28)
 rand = "*"
-ratsio = "*"
+ratsio = "^0.2"
 regex = "*"
 rustls = "*"
 serde = { version = "*", features = ["rc"] }
--- a/components/sup/Cargo.toml
+++ b/components/sup/Cargo.toml
@@ -51,7 +51,7 @@ prost = "*"
 prost-derive = "*"
 prost-types = "0.5.0" # This is current stable, but for some reason gets pulled in as 0.4.0 if we use "*" (2019-03-28)
 rand = "*"
-ratsio = "^0.2"
+ratsio = "*"
 regex = "*"
 rustls = "*"
 serde = { version = "*", features = ["rc"] }
--- a/components/sup/protocols/event.proto
+++ b/components/sup/protocols/event.proto
@@ -72,6 +72,48 @@ message ServiceStoppedEvent {
   ServiceMetadata service_metadata = 2;
 }
 
+message ServiceUpdateStartedEvent {
+  EventMetadata event_metadata = 1;
+  ServiceMetadata service_metadata = 2;
+  // Details of the service pkg to which 
+  // we're attempting to update.These
+  // are implemented as strings for now,
+  // but at some point we'll likely want
+  // to create a type for this set of
+  // attributes (PackageIdent)
+  string update_origin = 3;
+  string update_name = 4;
+  string update_version = 5;
+  string update_release = 6;
+}
+
+message ServiceUpdateCompleteEvent {
+  EventMetadata event_metadata = 1;
+  ServiceMetadata service_metadata = 2;
+  // Details of the service pkg to
+  // which we've successfully updated.
+  // These are implemented as strings for now,
+  // but at some point we'll likely want
+  // to create a type for this set of
+  // attributes (PackageIdent)
+  string update_origin = 3;
+  string update_name = 4;
+  string update_version = 5;
+  string update_release = 6;
+}
+
+message FailureDetail {}
+
+message ServiceUpdateFailedEvent {
+  EventMetadata event_metadata = 1;
+  ServiceMetadata service_metadata = 2;
+  // Details of the service pkg to
+  // which we've failed to update
+  ServiceMetadata update_metadata = 3;
+  // service update failure detail
+  FailureDetail failure_detail = 4;
+}
+
 message HealthCheckEvent {
   EventMetadata event_metadata = 1;
   ServiceMetadata service_metadata = 2;
--- a/components/sup/protocols/event.proto
+++ b/components/sup/protocols/event.proto
@@ -87,33 +87,6 @@ message ServiceUpdateStartedEvent {
   string update_release = 6;
 }
 
-message ServiceUpdateCompleteEvent {
-  EventMetadata event_metadata = 1;
-  ServiceMetadata service_metadata = 2;
-  // Details of the service pkg to
-  // which we've successfully updated.
-  // These are implemented as strings for now,
-  // but at some point we'll likely want
-  // to create a type for this set of
-  // attributes (PackageIdent)
-  string update_origin = 3;
-  string update_name = 4;
-  string update_version = 5;
-  string update_release = 6;
-}
-
-message FailureDetail {}
-
-message ServiceUpdateFailedEvent {
-  EventMetadata event_metadata = 1;
-  ServiceMetadata service_metadata = 2;
-  // Details of the service pkg to
-  // which we've failed to update
-  ServiceMetadata update_metadata = 3;
-  // service update failure detail
-  FailureDetail failure_detail = 4;
-}
-
 message HealthCheckEvent {
   EventMetadata event_metadata = 1;
   ServiceMetadata service_metadata = 2;
--- a/components/sup/protocols/event.proto
+++ b/components/sup/protocols/event.proto
@@ -76,15 +76,12 @@ message ServiceUpdateStartedEvent {
   EventMetadata event_metadata = 1;
   ServiceMetadata service_metadata = 2;
   // Details of the service pkg to which 
-  // we're attempting to update.These
-  // are implemented as strings for now,
-  // but at some point we'll likely want
+  // we're attempting to update.This is
+  // implemented as a string for now,
+  // but at some point we'll may want
   // to create a type for this set of
-  // attributes (PackageIdent)
-  string update_origin = 3;
-  string update_name = 4;
-  string update_version = 5;
-  string update_release = 6;
+  // attributes (FullyQualifiedPackageIdent)
+  string update_package_ident = 3;
 }
 
 message HealthCheckEvent {
--- a/components/sup/src/event.rs
+++ b/components/sup/src/event.rs
@@ -185,7 +185,7 @@ pub fn service_update_started(service: &Service, update: &PackageIdent) {
         publish(ServiceUpdateStartedEvent { event_metadata:       None,
                                             service_metadata:
                                                 Some(service.to_service_metadata()),
-                                            update_package_ident: format!("{}", update.clone()), });
+                                            update_package_ident: update.clone().to_string(), });
     }
 }
 
--- a/components/sup/src/event.rs
+++ b/components/sup/src/event.rs
@@ -27,7 +27,8 @@ use self::types::{EventMessage,
                   EventMetadata,
                   HealthCheckEvent,
                   ServiceStartedEvent,
-                  ServiceStoppedEvent};
+                  ServiceStoppedEvent,
+                  ServiceUpdateStartedEvent};
 use crate::manager::{service::{HealthCheckResult,
                                Service},
                      sys::Sys};
@@ -37,7 +38,8 @@ pub use error::{Error,
 use futures::sync::mpsc::UnboundedSender;
 use habitat_common::types::{AutomateAuthToken,
                             EventStreamMetadata};
-use habitat_core::env::Config as EnvConfig;
+use habitat_core::{env::Config as EnvConfig,
+                   package::ident::PackageIdent};
 use state::Container;
 use std::{net::SocketAddr,
           num::ParseIntError,
@@ -177,6 +179,22 @@ pub fn service_stopped(service: &Service) {
     }
 }
 
+/// Send an event at the start of a Service update.
+pub fn service_update_started(service: &Service, update: &PackageIdent) {
+    if stream_initialized() {
+        publish(ServiceUpdateStartedEvent { event_metadata:   None,
+                                            service_metadata: Some(service.to_service_metadata()),
+                                            update_origin:    update.origin.clone(),
+                                            update_name:      update.name.clone(),
+                                            update_version:   update.version
+                                                                    .clone()
+                                                                    .unwrap_or_default(),
+                                            update_release:   update.release
+                                                                    .clone()
+                                                                    .unwrap_or_default(), });
+    }
+}
+
 // Takes metadata directly, rather than a `&Service` like other event
 // functions, because of how the asynchronous health checking
 // currently works. Revisit when async/await + Pin is all stabilized.
@@ -219,6 +237,7 @@ fn publish(mut event: impl EventMessage) {
         // one.
         //
         // The ugliness is at least contained, though.
+        debug!("Publishing to event stream: event {:?} ", event);
         event.event_metadata(EventMetadata { occurred_at:
                                                  Some(std::time::SystemTime::now().into()),
                                              ..EVENT_CORE.get::<EventCore>().to_event_metadata() });
--- a/components/sup/src/event.rs
+++ b/components/sup/src/event.rs
@@ -186,12 +186,16 @@ pub fn service_update_started(service: &Service, update: &PackageIdent) {
                                             service_metadata: Some(service.to_service_metadata()),
                                             update_origin:    update.origin.clone(),
                                             update_name:      update.name.clone(),
-                                            update_version:   update.version
-                                                                    .clone()
-                                                                    .unwrap_or_default(),
-                                            update_release:   update.release
-                                                                    .clone()
-                                                                    .unwrap_or_default(), });
+                                            update_version:
+                                                update.version
+                                                      .clone()
+                                                      .expect("Can't send service update started \
+                                                               event; invalid update version"),
+                                            update_release:
+                                                update.release
+                                                      .clone()
+                                                      .expect("Can't send service update started \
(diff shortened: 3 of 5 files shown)


components/sup/protocols/event.proto
```diff
@@ -72,6 +72,48 @@ message ServiceStoppedEvent {
...
+  // Details of the service pkg to which 
+  // we're attempting to update.These
+  // are implemented as strings for now,
+  // but at some point we'll likely want
+  // to create a type for this set of
+  // attributes (PackageIdent)
+  string update_origin = 3;
+  string update_name = 4;
+  string update_version = 5;
+  string update_release = 6;
```
christophermaier: Is there a reason in Automate to have these as separate fields, rather than a single string?

components/sup/protocols/event.proto
```diff
@@ -72,6 +72,48 @@ message ServiceStoppedEvent {
...
+  // Details of the service pkg to
+  // which we've successfully updated.
+  // These are implemented as strings for now,
+  // but at some point we'll likely want
+  // to create a type for this set of
+  // attributes (PackageIdent)
+  string update_origin = 3;
+  string update_name = 4;
+  string update_version = 5;
+  string update_release = 6;
```
christophermaier: I think when this is sent, the identifier would be the same as what's in the `service_metadata`, so we probably don't need to send it.

components/sup/protocols/event.proto
```diff
@@ -72,6 +72,48 @@ message ServiceStoppedEvent {
...
+}
+
+message FailureDetail {}
```
christophermaier: An empty message?

components/sup/src/event.rs
```diff
@@ -177,6 +179,22 @@ pub fn service_stopped(service: &Service) {
...
+        publish(ServiceUpdateStartedEvent { event_metadata:   None,
+                                            service_metadata: Some(service.to_service_metadata()),
+                                            update_origin:    update.origin.clone(),
+                                            update_name:      update.name.clone(),
+                                            update_version:   update.version
+                                                                    .clone()
+                                                                    .unwrap_or_default(),
+                                            update_release:   update.release
+                                                                    .clone()
+                                                                    .unwrap_or_default(), });
```
christophermaier: I think the `update` `PackageIdent` will always be a fully-qualified one at this point, so these could be made into `expect` calls.

However, this just drives home my desire to create a formal `FullyQualifiedPackageIdent` variant, though, so we can be 100% assured that this is the case 😭

components/sup/src/event/types.rs
```diff
@@ -105,6 +105,12 @@ impl EventMessage for ServiceStoppedEvent {
     }
 }
 
+impl EventMessage for ServiceUpdateStartedEvent {
+    fn event_metadata(&mut self, event_metadata: EventMetadata) {
+        self.event_metadata = Some(event_metadata);
+    }
+}
+
```
christophermaier: If you're feeling ambitious and want to get into Rust macros, now might be a good time to tackle the suggestion on lines 93--94 😄

components/sup/protocols/event.proto
christophermaier: Also, it doesn't look like this message is used yet, so we should go ahead and remove it.

components/sup/protocols/event.proto
```diff
@@ -72,6 +72,48 @@ message ServiceStoppedEvent {
...
+message ServiceUpdateFailedEvent {
+  EventMetadata event_metadata = 1;
+  ServiceMetadata service_metadata = 2;
+  // Details of the service pkg to
+  // which we've failed to update
+  ServiceMetadata update_metadata = 3;
+  // service update failure detail
+  FailureDetail failure_detail = 4;
+}
+
```
christophermaier: This also appears currently unused.

components/sup/Cargo.toml
```diff
@@ -51,7 +51,7 @@ prost = "*"
 prost-derive = "*"
 prost-types = "0.5.0" # This is current stable, but for some reason gets pulled in as 0.4.0 if we use "*" (2019-03-28)
 rand = "*"
-ratsio = "*"
+ratsio = "^0.2"
```
christophermaier: Why add this constraint?

components/sup/protocols/event.proto
gpeers: @cm i think the identifier in the service_metadata corresponds to the currently running version of the service. the update fields are the values for the package to which the supervisor will update the service. lmk if i'm mistaken on this one.

components/sup/protocols/event.proto
gpeers: it's a placeholder; i can remove if that's preferable.

components/sup/protocols/event.proto
gpeers: ok, i'll remove all currently unused code.

components/sup/protocols/event.proto
gpeers: @cm i thought the ergonomics might be easier on the automate side instead of that code having to parse out what it needs for comparisons, etc. lmk if you disagree.

components/sup/protocols/event.proto
christophermaier: @gpeers it depends on when this event would be fired. If it is after the service has come back up (as I suspect / assume), then the updated identifier would indeed be the currently-running version, so they would be the same.

TL;DR - we're both right 😆

components/sup/Cargo.toml
gpeers: because the client library states in its doc that we should make use of 0.2+. i know * should grab the most recent version, but it's not always entirely deterministic, is it?

components/sup/protocols/event.proto
christophermaier: @gpeers It depends on whether or not this PR is intended as an early look at work in progress, or something that you're ready to merge now. If it's the former, we can leave it in, but if it's the latter, let's take it out.

components/sup/src/event.rs
gpeers: i thought i'd seen that only origin and name were required, but i'll double-check this and make the change if appropriate.

components/sup/src/event/types.rs
gpeers: thanks @cm, i'll check it out!

components/sup/Cargo.toml
christophermaier: It should synonymous in this case. For good or bad, we've settled on using `*` for everything _unless_ there's a concrete reason to lock to a specific version. In that case, we'll leave a comment as to why we're locking.

(Though now I see the `prost-types` version pin that _I_ added just above... I should revisit that one, but were you seeing something similar?)

components/sup/protocols/event.proto
christophermaier: @gpeers maybe we should chat about that usecase. It may be time to create some protobuf types to represent the different types of identifiers that we could have.

components/sup/src/event.rs
christophermaier: In our current `PackageIdent` type that is true. However, in practice, identifiers without versions and releases (i.e., non-fully-qualified identifiers) only exist at the periphery of the system; once they make it down into the core, they're usually resolved to a concrete, fully-qualified identifier.

(If we actually had a `FullyQualifiedPackageIdent` type, then we could have the compiler _prove_ that for us, but alas, such a thing does not exist yet.)

components/sup/protocols/event.proto
gpeers: done

components/sup/protocols/event.proto
gpeers: done

components/sup/protocols/event.proto
gpeers: @christophermaier nope, i insist... i read your comment too fast and didn't see that it referred to the service update complete event. thanks : )

components/sup/Cargo.toml
gpeers: this wasn't pinned, though i did see other pins and followed that example. i'll remove the pin.

components/sup/protocols/event.proto
gpeers: @christophermaier i think i'll create a separate ticket and pr for any new types we decide on if that works for you.

christophermaier: @gpeers I guess that depends on when Automate will start consuming this. If they're not going to do anything with this until those new types get added (and if the new types get added as the immediate next PR), then that's probably OK. I'd just hate for them to write code to deal with one form, only to have it changed right out from under them.

Alternatively, if it's not too much work, adding a new type to this PR would also be fine.

gpeers: @christophermaier ok, sure. let's discuss later today if you've got time.

components/sup/src/event/types.rs
gpeers: done!

components/sup/src/event.rs
```diff
@@ -177,6 +179,26 @@ pub fn service_stopped(service: &Service) {
     }
 }
 
+/// Send an event at the start of a Service update.
+pub fn service_update_started(service: &Service, update: &PackageIdent) {
+    if stream_initialized() {
+        publish(ServiceUpdateStartedEvent { event_metadata:   None,
+                                            service_metadata: Some(service.to_service_metadata()),
```
baumanj: Not new code, but it seems like instead of having `Service::to_service_metadata`, it would be more idiomatic to have `impl Into<ServiceMetadata> for &Service`, then this becomes:
```suggestion
                                            service_metadata: Some(service.into()),
```

components/sup/src/event.rs
```diff
@@ -177,6 +179,26 @@ pub fn service_stopped(service: &Service) {
...
+                                            update_origin:    update.origin.clone(),
+                                            update_name:      update.name.clone(),
+                                            update_version:
+                                                update.version
+                                                      .clone()
+                                                      .expect("Can't send service update started \
```
baumanj: This seems like an awkward way to panic. The ideal would be to have this function take a `FullyQualifiedPackageIdent` type, but until we have that, I'm not sure we want to crash the entire supervisor. Would something like this be preferable?
```rust
        if let PackageIdent { origin: update_origin,
                              name: update_name,
                              version: Some(update_version),
                              release: Some(update_release), } = update.clone()
        {
            publish(ServiceUpdateStartedEvent { event_metadata: None,
                                                service_metadata: service.into(),
                                                update_origin,
                                                update_name,
                                                update_version,
                                                update_release });
        } else {
            error!(…);
        }
    }
```

components/sup/protocols/event.proto
gpeers: @baumanj appreciate the review : ) i don't believe so... these constitute the fully qualified identifier of the package we're updating to rather than the that of the current package. i've also refactored the code so we are consolidating the individual fields into one concatenated string which represents the FQI of the update package.

components/sup/src/event.rs
gpeers: @baumanj yep, makes total sense. i'll refactor in a separate pr.

components/sup/src/event.rs
gpeers: @baumanj thanks, i believe this is a non-issue now that i've refactored the fully-qualified ident for the update package into a string in the proto (the same way the current package ident is expressed). i'll keep this in mind going forward, however!

components/sup/protocols/event.proto
baumanj: The current and updating-to package must be the same in terms of `origin/name`, though, right?

gpeers: @baumanj yes, that's true. at this point in code we also have the version and release of the update package as well. are you ok with sending the update package ident to automate as a single string (origin/name/version/release) in the same way the current package ident is sent over?

components/sup/src/event.rs
```diff
@@ -177,6 +179,16 @@ pub fn service_stopped(service: &Service) {
     }
 }
 
+/// Send an event at the start of a Service update.
+pub fn service_update_started(service: &Service, update: &PackageIdent) {
+    if stream_initialized() {
+        publish(ServiceUpdateStartedEvent { event_metadata:       None,
+                                            service_metadata:
+                                                Some(service.to_service_metadata()),
+                                            update_package_ident: format!("{}", update.clone()), });
```
christophermaier: This would be more idiomatically rendered as `update.to_string()`, rather than going through the `format!` macro.

I thought Clippy would have caught this with [useless_format](https://rust-lang.github.io/rust-clippy/master/#useless_format), but now I see that only would apply if `update` was a `String`.

components/sup/src/event/types.rs
```diff
@@ -90,23 +90,19 @@ pub trait EventMessage: Message + Sized {
...
-    fn event_metadata(&mut self, event_metadata: EventMetadata) {
-        self.event_metadata = Some(event_metadata);
-    }
-}
+event_msg_impl! (
+    ServiceStartedEvent
+    ServiceStoppedEvent
+    ServiceUpdateStartedEvent
+    HealthCheckEvent
+);
```
christophermaier: I think my preference here would be for the macro to take a single type, rather than multiple.

It'd be a bit more typing to invoke them, but it also means the macro itself has a single responsibility (just adding the code, rather than adding code and also iterating). It simplifies the macro (very slightly), but I tend to favor the simplest macro possible. It's also more apparent from a glance at the code below exactly what is going on (you don't need to ask what it needs all the types for). 

(As an macro implementation aside, it's also a little odd to _not_ have commas between the types, since they're basically being treated as arguments here.)

```rust
event_msg_impl!(ServiceStartedEvent);
event_msg_impl!(ServiceStoppedEvent);
event_msg_impl!(ServiceUpdateStartedEvent);
event_msg_impl!(HealthCheckEvent);
```

I think this particular example is arguably in a gray area, though, so I'm open to other thoughts. @baumanj @raskchanky, what do you think?

components/sup/src/event.rs
gpeers: 👍

components/sup/src/event/types.rs
gpeers: 👍

components/sup/protocols/event.proto
baumanj: I'm torn. On one hand, I generally prefer to avoid mandatory duplication (especially in protocols) since it opens the door to inconsistent states. How should consumers behave if the `ServiceMetadata` says `foo/bar/XXX/YYY` but the update package says `baz/qux/WWW/ZZZ`? On the other, we don't have anywhere else where we have a release and version without origin and name, so it's a bit odd. I'd still say my preference would be populating `ServiceMetadata` with the pre-update FQPI and having just `update_version` and `update_release` in this message. If that leads to major awkwardness, I'll defer to your discretion.

components/sup/src/event/types.rs
baumanj: I'm fine with either approach, but if it takes multiple types, I agree that comma separation would be good (and should be straightforward to add).

Focusing more on this code, I think it's kind of confusing that the method is named `event_metadata`, but also assigns to a field named `event_metadata`. I'd expect the setter method to have a name like `set_metadata` instead.

Thinking about it a bit more, if `EventMetadata` is part of every event,  could we restructure a bit and avoid the need to have it as part of the message types at all? It's only ever set in the `publish` method, and since the argument to `publish` is moved, it can't be accessed afterward. Couldn't `publish` itself create and send the `EventMetadata` bytes and then the particular message type (which no longer has `EventMetadata` embedded)? If we can go this route, then we wouldn't need the macro at all.

christophermaier: @baumanj I agree that the `EventMetadata` bit is a little weird. However, `publish` is using the `to_bytes` method from `EventMessage` to transform the entire event type into a protobuf byte string. Splitting that byte generation up would probably be pretty error prone, and more complex than the current (admittedly weird) code.

I'm open to other alternatives, though. This is probably the part of this code that I like the least.

components/sup/protocols/event.proto
christophermaier: Hrmm... it's a good point.

Given that we don't (yet?) have a concept of just "version + release", I'm inclined to keep the code as it is now, sending the fully-qualified identifier of the new version.

As we dig into additional refactorings around our identifier types, though, this distinction could be good to keep in mind.

components/sup/src/event/types.rs
baumanj: Sending the event on the wire together with its metadata is fine, but I think there must be some way to have `publish` handle the population of the bits that are identical across all types. Possibly it would require more types in the .proto, but I think that's preferable to the shenanigans we've got currently.

components/sup/protocols/event.proto
baumanj: We should be careful; changing protocols down the road is much harder than changing code. If we send FQPIs for both now, we're likely locked into that for awhile.

Issue #6588, opened 2019-05-23
Consider owning our own logging configuration refresh


christophermaier:
In #6500, we transitioned over to `log4rs` to manage our logging, in part because it offers the ability to change logging configuration at runtime. Out of the box, it will do this if and only if your configuration provides a `refresh_rate`. If it does, `log4rs` will spin up a thread to periodically check for modifications to that file.

However, if you start with a configuration that does not have `refresh_rate` set, this thread does _not_ get started, and any subsequent changes to the file will be ignored for the rest of the process' lifetime. Similarly, if you start with `refresh_rate`, but later _remove_ that configuration, the refresh thread is terminated, thus depriving you of any further chance to modify your configuration.

Additionally, this refresh logic only applies if a configuration file is present at the start; if you were to start up without a configuration file and use some default or programmatic configuration, the subsequent addition of a configuration file would be ignored.

All this makes sense, but we may want to provide additional capabilities beyond what is available out of the box in order to allow our users to always be able to modify their logging configuration, regardless of how their Supervisor process was originally started. If we managed our own configuration refresh, we would have free reign to determine exactly how we wanted it to work. It would also decouple us from any particular logging framework, allowing us to more easily migrate away from `log4rs` in the future, should we ever desire to do so.

PR #6622, opened 2019-06-05
Switch to log4rs in the Supervisor

### LINKED ISSUE #6500: Add support for runtime configuration of logging levels

### LINKED ISSUE #6583: Ease transition from env_logger to log4rs
See #6500 

`log4rs` allows _much_ more flexibility of logging than we currently have. Initially, at least, we'll take advantage of this by using a `log4rs` configuration file.

However, throughout our entire project history, we've conditioned users to configure logging using the `RUST_LOG` environment variable (the default for Rust's `env_logger`). While using a configuration file allows for considerable latitude in configurations, it can be a lot to ask if you just want to quickly enable debug logging. And while it isn't impossible to mount a configuration file into a Docker container running Habitat, it adds extra friction if you don't need that level of configurability.

To help transition users over to the new logging implementation, we can take the contents of any `RUST_LOG` environment variable that is present and programmatically create a logging configuration for `log4rs`. If `RUST_LOG` is present, but a `log4rs` configuration file is not, then we can use the environment variable. If both are present, however, we will only use the configuration file, as that would be a signal that the user is ready to use the new system.

christophermaier:
By switching away from `env_logger` to `log4rs`, we gain a lot of
configurability, including _dynamic_ configuration changes, allowing
us to selectively toggle log verbosity at runtime.

Out of the box, and without any additional configuration, logging will
behave largely as it does now. We will even honor `RUST_LOG`
environment variable-based configuration strings as a way to
transition users to the new logging system.

However, by adding a `/hab/sup/default/config/log.yml` configuration
file (whose format is detailed at
https://docs.rs/log4rs/0.8.3/log4rs/#configuration), users will be
able to greatly influence how logging is done.

A key feature to notice is that if the file contains a `refresh_rate`
key, then the file will be monitored for changes at runtime. There is
a caveat in the current default implementation, however; if the
Supervisor starts with a configuration file that does _not_ have a
`refresh_rate` key, then it will _never_ update the active
configuration based on changes to the file. Similarly, if you later
update the file and _remove_ the `refresh_rate` key, then monitoring
will _stop_ altogether. For now, if you want runtime monitoring, just
always have a `refresh_rate` present. We can change this in the
future, though; follow #6588 for details.

A basic, refreshing configuration file would look like this:

```yaml
refresh_rate: 5 seconds

appenders:
  stdout:
    kind: console
    encoder:
      pattern: "[{d(%Y-%m-%dT%H:%M:%SZ)(utc)} {l} {module}] {message}{n}"

root:
  level: error
  appenders:
    - stdout
```

To selectively filter on modules, add some `loggers` to that, like so:

```yaml
loggers:
  tokio_threadpool:
    level: trace
  habitat_sup::manager::service_updater:
    level: debug
```

Be sure to read the code comments for further details.

Fixes #6500
Fixes #6583

### DIFF
--- a/components/sup/src/logger.rs
+++ b/components/sup/src/logger.rs
@@ -1,0 +1,96 @@
+use log::LevelFilter;
+use log4rs::{append::console::ConsoleAppender,
+             config::{Appender,
+                      Config,
+                      Root},
+             encode::pattern::PatternEncoder,
+             file::Deserializers};
+use std::path::PathBuf;
+
+mod env_logger_compatibility;
+
+/// A `log4rs`
+/// [PatternEncoder](https://docs.rs/log4rs/0.8.3/log4rs/encode/pattern/index.html)
+/// format to mimic that of out-of-the-box `env_logger`.
+const DEFAULT_PATTERN: &str = "[{d(%Y-%m-%dT%H:%M:%SZ)(utc)} {l} {module}] {message}{n}";
+
+/// Initialize a log4rs-based logging system.
+///
+/// Absent any other configuration, a basic logging configuration will
+/// be used. Users can also specify configuration using a YAML
+/// configuration file (see
+/// https://docs.rs/log4rs/0.8.3/log4rs/#configuration).
+///
+/// We also provide a migration path from our previous
+/// `env_logger`-based implementation. If a configuration file is
+/// absent, but `RUST_LOG` is present in the environment, we will
+/// coerce a `log4rs` configuration from that `env_logger`
+/// configuration string.
+pub fn init() {
+    let file = configuration_file();
+    if file.exists() {
+        if let Err(e) = log4rs::init_file(&file, Deserializers::default()) {
+            eprintln!("Logging configuration file '{}' not valid: {}; using default logging \
+                       configuration",
+                      file.display(),
+                      e);
+            log4rs::init_config(default_config()).expect("Tried setting the log configuration, \
+                                                          but the global logger had already been \
+                                                          set!");
+        }
+    } else {
+        let config = env_logger_compatibility::from_env().unwrap_or_else(|| {
+                                                             eprintln!("Logging configuration \
+                                                                        file '{}' not found; \
+                                                                        using default logging \
+                                                                        configuration",
+                                                                       file.display());
+                                                             default_config()
+                                                         });
+
+        log4rs::init_config(config).expect("Tried setting the log configuration, but the global \
+                                            logger had already been set!");
+    }
+}
+
+/// The logging configuration that will be used if a configuration
+/// file is not found when the Supervisor is started.
+///
+/// Logs everything at ERROR and above to standard output. This is
+/// kept deliberately "bare bones" (for now), since `log4rs` allows for so much
+/// flexibility to end users via its configuration file.
+///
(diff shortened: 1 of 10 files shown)


components/sup/src/logger/env_logger_compatibility.rs
```diff
@@ -0,0 +1,246 @@
...
+use std::{collections::HashMap,
+          str::FromStr};
+
+pub(super) fn from_env() -> Option<Config> {
+    std::env::var("RUST_LOG").ok().map(|env_var| {
+                                      eprintln!("RUST_LOG environment variable found; using it \
+                                                 to configure log4rs");
+                                      env_var.parse::<EnvLogConfig>()
+                                             .expect("RUST_LOG parsing can't fail")
+                                             .into()
```
baumanj: Since it's infallible, why not implement `From<String> for EnvLogConfig` instead of `FromStr` and simplify this to:
```rust
EnvLogConfig::from(env_var).into()
```
?

components/sup/src/logger.rs
```diff
@@ -0,0 +1,96 @@
...
+/// https://github.com/habitat-sh/habitat/issues/6587, but
+/// particularly https://github.com/habitat-sh/habitat/issues/6584),
+/// we _may_ want to further customize this default configuration, in
+/// order to more closely adhere to our existing non-log-based output
+/// (at that point, though, it may be better to define the default in
+/// a YAML file and `include!` it here for readability)
+fn default_config() -> Config {
+    let stdout = ConsoleAppender::builder().encoder(Box::new(PatternEncoder::new(DEFAULT_PATTERN)))
+                                           .build();
+    // Calling `unwrap()` is OK because we have full control over this
```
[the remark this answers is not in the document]
christophermaier: 😅

components/sup/src/logger/env_logger_compatibility.rs
christophermaier: I'd rather leave this as `FromStr`, even though it is infallible, since this is inherently a parsing activity.

This question did cause me to dig in and figure out exactly what the distinction between `FromStr` and `From<String>` actually is, though; I actually wasn't too clear on that myself! After looking at the `From<String>` implementations in the standard library, it looks like `From<String>` is just for doing things like creating contained strings (or string-like things), as in `From<String> for Box<String>`, or for converting from a string into another inherently string-like thing, as with `From<String> for PathBuf`. 

That feels like a very distinct use case from what we're trying to do here.

baumanj: I mostly suggested it because it eliminated the need for both a turbo fish and an `expect`, but if you feel `FromStr` is more appropriate, that's fine with me.

Issue #6620, opened 2019-06-05
Upgrade to actix 1.0


davidMcneil:
Upgrade to actix 1.0.

A first attempt at the upgrade can be found [here](https://github.com/habitat-sh/habitat/tree/dmcneil/upgrade-actix-web).

The two main blockers are the `Middleware` and `Predicate` traits. The way middleware is handled has been completely changed. There are [examples](https://github.com/actix/examples/blob/master/middleware/src/simple.rs) of the new system. However, there seems to be many questions by the community on how to best implement middlware so I am going to hold off and see how the actix-web team responds. I can not find any equivalent of `Predicate` or filtering requests in the 1.0 release. I asked a question on their Gitter channel and am waiting to hear back.

Note: Also upgrade builder.

PR #6630, opened 2019-06-06
0.82.0 Post-release tasks

smacfarlane:
This updates Cargo.lock, and fixes any issues introduced. 

* Temporarily pin libc to 0.2.54. Newer versions are not compatible with the current release of palaver.
* Temporarily pin actix-web until #6620 is resolved

This also updates RUSTFMT_VERSION to 2019-06-06


components/common/src/command/package/install.rs
```diff
@@ -63,7 +64,7 @@ use crate::{error::{Error,
             ui::{Status,
                  UIWriter}};
 
-pub const RETRIES: u64 = 5;
+pub const RETRIES: usize = 5;
 pub const RETRY_WAIT: u64 = 3000;
```
baumanj: As long as we're making modifications around here, I abhor constants without unit indications
```suggestion
pub const RETRY_WAIT_MS: u64 = 3000;
```
or now that `Duration` constructors are const functions, we could do even better:
```suggestion
pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
```
but because the crate's implementation of `Fixed` only provides a `from_millis(u64)` constructor and `Duration`'s `as_millis` returns `u128`, we'd need a little helper function:
```rust
fn wait_for(delay: Duration, times: usize) -> impl IntoIterator<Item = Duration> {
    vec![delay].into_iter().cycle().take(times)
}
```
But I think this makes for the most ergonomic result:
```rust
retry(wait_for(RETRY_WAIT, RETRIES), …)
```

baumanj: It was such a small thing, I submitted [a PR to the `retry` crate](https://github.com/jimmycuadra/retry/pull/21).

smacfarlane: I like this suggestion, but it feels like `command/package/install.rs` is the wrong place for the `wait_for` function. It would be also used in key upload/download and Builder would also want to take advantage of it.  Somewhere in the core crate seems like the best place for it but I don't see an obvious (to me) spot for it.  What do you think?  

Would it be best to file an issue to change the type to a Duration once the retry crate accepts your pr?

baumanj: Yeah, I wouldn't put `wait_for` in install.rs. I'd suggest https://github.com/habitat-sh/habitat/blob/master/components/core/src/util/mod.rs as a good home that allows `builder` to take advantage.

No need to file an issue, I don't think. In the event that my PR is accepted into the `retry` crate, I'll get a notification and make the appropriate change once it merges. If you like, you could add a link to the PR in a doc comment for `wait_for` as a reminder.

smacfarlane: @baumanj I think I've addressed your feedback,  please let me know if there are any additional changes needed.

baumanj: Looks great! Thanks for your thoroughness

PR #6643, opened 2019-06-11
allow windows docker studio and exported images to run on kernels later than 2016ltsc

mwrock:
Today when we build windows containers, we use `microsoft/windowsservercore:2016ltsc` as a base image. This works (both for building and consuming) on a 2016 Server host or any host using Hyper-V isolation which will run a container inside of a light weight VM. The bottom line is that a container can only be run in a host that supports the kernel version of the image. This PR allows us to build windows containers for all docker supported kernels.

## Exporter changes
The docker exporter now takes an optional arg (only on windows) `--base-image` that a user can use to specify a specific base image for building. If none is provided, the exporter will use the appropriate base image based on the running host.

## Plans for building non-2016 studios
As this code stands with our existing build infrastructure, we will continue to build 2016 studios since our build queues are all 2016 images. However this gives us the ability to use either different build images for building other versions or use a single bare metal hyper-v host to build a collection of different versions.

## General Refactorings
There are a couple refactorings included here:
* Finding the `docker` command and determining the base image tag based on host are extracted to the `core` crate since they are used by both `hab` and `pkg-exporter-docker`.
* When the `hab` cli delegates to an exporter, it now uses `command::pkg::exec::start` so that the runtime deps of the exporter are added to the path which include `docker` for the docker exporter.


mwrock:
fixes #6412

components/core/src/util/docker.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::{path::PathBuf,
+          process::Command};
+
+const DOCKER_CMD: &str = "docker";
+const DOCKER_CMD_ENVVAR: &str = "HAB_DOCKER_BINARY";
+
+pub fn command() -> Result<PathBuf> {
+    let docker_cmd = env::var(DOCKER_CMD_ENVVAR).unwrap_or_else(|_| DOCKER_CMD.to_string());
+
+    match find_command(&docker_cmd) {
```
davidMcneil: This 
```rust
match find_command(&docker_cmd) {
    Some(docker_abs_path) => Ok(docker_abs_path),
    None => Err(Error::DockerCommandNotFound(docker_cmd.into())),
}
```
could be expressed as
```rust
find_command(&docker_cmd).ok_or_else(|| Error::DockerCommandNotFound(docker_cmd.into()))
```

components/pkg-export-docker/src/build.rs
```diff
@@ -24,6 +24,7 @@ use crate::{accounts::{EtcGroupEntry,
                     package::{PackageArchive,
                               PackageIdent,
                               PackageInstall},
+                    util::docker,
```
davidMcneil: On linux I get a warning that this is unused.
```rust
#[cfg(windows)]
use crate::hcore::util::docker;
```

mwrock: great suggestions. will fix!

components/pkg-export-docker/src/build.rs
```diff
@@ -82,6 +104,8 @@ pub struct BuildSpec<'a> {
     pub idents_or_archives: Vec<&'a str>,
     /// The Builder Auth Token to use in the request
     pub auth: Option<&'a str>,
+    /// Base image used in From of dockerfile
+    pub base_image: Option<&'a str>,
```
baumanj: The more work I do in rust, the more I've found `Option`s inside types (as opposed to parameters or return values) to be an anti-pattern.

In this case, I'd suggest
```suggestion
    pub base_image: String,
```
This allows `docker_base_image` to be simplified to:
```rust
fn default_docker_base_image() -> String {
    #[cfg(unix)]
    {
        DEFAULT_BASE_IMAGE.to_string()
    }

    #[cfg(windows)]
    {
        format!("{}:{}",
                DEFAULT_BASE_IMAGE,
                docker::default_base_tag_for_host()?)
    }
}
```
(And perhaps inlining those `DEFAULT_BASE_IMAGE` constants since they're not used anywhere else and to make it clear they're different across the different platforms would be even better)

With that change, the handling of the optional case can go _before_ the `BuildSpec` is created in `new_from_cli_matches`:
```rust
                    base_image:         m.value_of("BASE_IMAGE")
                                         .map(str::to_string)
                                         .unwrap_or_else(default_docker_base_image), }
```

components/pkg-export-docker/src/cli.rs
```diff
@@ -279,6 +279,16 @@ impl<'a, 'b> Cli<'a, 'b> {
 
         Cli { app }
     }
+
+    pub fn add_base_image_arg(self) -> Self {
+        let app = self.app
+                      .arg(Arg::with_name("BASE_IMAGE").value_name("BASE_IMAGE")
+                                                       .long("base-image")
```
baumanj: An example of a valid input and/or a validator would probably be helpful here

components/core/src/util/docker.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{env,
+            error::{Error,
+                    Result},
+            fs::find_command};
+use std::{path::PathBuf,
+          process::Command};
+
+const DOCKER_CMD: &str = "docker";
+const DOCKER_CMD_ENVVAR: &str = "HAB_DOCKER_BINARY";
```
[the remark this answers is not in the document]
mwrock: I actually just moved it from `hab/src/command/studio/docker.rs`. I'm actually thinking we just delete it. There is only a single reference to it in `building.md` but I really don't think anything uses it.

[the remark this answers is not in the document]
baumanj: I'm in favor of removing it. Though we don't really have any way of knowing if some user is making use of it, do we?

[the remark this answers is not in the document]
mwrock: no. However, I feel confident its not used (of course now that I say it, it will probably be wrong).

components/core/src/util/docker.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub fn command() -> Result<PathBuf> {
```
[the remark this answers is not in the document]
mwrock: I was changing to return a `Command` since on the surface that seemed like the right thing to do but ended up going with `command_path` because we may need to create several `command` instances in a single operation but only want to run through this logic to get the path once.

mwrock:
thanks @baumanj for the review. Just pushed up changes addressing all of your comments.

PR #6644, opened 2019-06-11
Load config variables from cli.toml when entering hab-studio

### LINKED ISSUE #4908: Studio should import `auth_token` from cli.toml
After running `hab setup`, there will be an `auth_token` key in `~/.hab/etc/cli.toml`. This is automatically used for operations outside the studio:
```
➤ hab pkg promote jbauman/np-mongodb/3.2.9/20180412155926 stable
» Promoting jbauman/np-mongodb/3.2.9/20180412155926 to channel 'stable'
✓ Promoted jbauman/np-mongodb/3.2.9/20180412155926
```
but not inside the studio:
```
[2][default:/src:0]# hab pkg promote jbauman/np-mongodb/3.2.9/20180412155926 stable
✗✗✗
✗✗✗ No auth token specified
✗✗✗
```
This was surprising to me and seems seems inconsistent since the studio *does* access the `origin` key from `cli.toml` to import keys and set `HAB_ORIGIN`:
```
[4][default:/src:1]# echo $HAB_ORIGIN
jbauman
```
This was on Mac OS with hab 0.55.0/20180321215236.

qubitrenegade:
This commit loads from `cli.toml` and sets:

* HAB_AUTH_TOKEN
* HAB_BLDR_URL
* HAB_CTL_SECRET

If these variables are not set before `hab studio enter` is executed.

This resolves #4908

### DIFF
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -13,25 +13,44 @@ use crate::{config,
 
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 
-const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
 const AUTH_TOKEN_ENVVAR: &str = "HAB_AUTH_TOKEN";
+const BLDR_URL_ENVVAR: &str = "HAB_BLDR_URL";
+const CTL_SECRET_ENVVAR: &str= "HAB_CTL_SECRET";
+const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
 pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
+    if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(auth_token) = config.auth_token {
+            debug!("Setting {}={} via config file", AUTH_TOKEN_ENVVAR, &auth_token);
+            env::set_var("HAB_AUTH_TOKEN", auth_token);
+        }
+    }
+
+    if henv::var(BLDR_URL_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(bldr_url) = config.bldr_url {
+            debug!("Setting {}={} via config file", BLDR_URL_ENVVAR, &bldr_url);
+            env::set_var("HAB_BLDR_URL", bldr_url);
+        }
+    }
+
+    if henv::var(CTL_SECRET_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(ctl_secret) = config.ctl_secret {
+            debug!("Setting {}={} via config file", CTL_SECRET_ENVVAR, &ctl_secret);
+            env::set_var("CTL_SECRET_ENVVAR", ctl_secret);
+        }
+    }
+
     if henv::var(ORIGIN_ENVVAR).is_err() {
         let config = config::load()?;
         if let Some(default_origin) = config.origin {
             debug!("Setting default origin {} via CLI config", &default_origin);
             env::set_var("HAB_ORIGIN", default_origin);
-        }
-    }
-    if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
-        let config = config::load()?;
-        if let Some(auth_token) = config.auth_token {
-            debug!("Setting default origin {} via CLI config", &auth_token);
-            env::set_var("HAB_AUTH_TOKEN", auth_token);
         }
     }
 
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -22,13 +22,17 @@ const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
-fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive: bool) {
+enum Sensitivity {
+    PrintValue,
+    NoPrintValue,
+}
+
+fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive: Sensitivity) {
     if henv::var(env_var).is_err() {
         if let Some(val) = config_val {
-            if sensitive {
-                debug!("Setting {}=REDACTED (sensitive) via config file", env_var)
-            } else {
-                debug!("Setting {}={} via config file", env_var, val)
+            match sensitive {
+                Sensitivity::NoPrintValue =>  debug!("Setting {}=REDACTED (sensitive) via config file", env_var),
+                Sensitivity::PrintValue => debug!("Setting {}={} via config file", env_var, val),
             }
             env::set_var(env_var, val);
         }
@@ -38,10 +42,10 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
 pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
     let config = config::load()?;
 
-    set_env_var_from_config(AUTH_TOKEN_ENVVAR, config.auth_token, true);
-    set_env_var_from_config(BLDR_URL_ENVVAR, config.bldr_url, false);
-    set_env_var_from_config(CTL_SECRET_ENVVAR, config.ctl_secret, true);
-    set_env_var_from_config(ORIGIN_ENVVAR, config.origin, false);
+    set_env_var_from_config(AUTH_TOKEN_ENVVAR, config.auth_token, Sensitivity::NoPrintValue);
+    set_env_var_from_config(BLDR_URL_ENVVAR, config.bldr_url, Sensitivity::PrintValue);
+    set_env_var_from_config(CTL_SECRET_ENVVAR, config.ctl_secret, Sensitivity::NoPrintValue);
+    set_env_var_from_config(ORIGIN_ENVVAR, config.origin, Sensitivity::PrintValue);
 
     if henv::var(CACHE_KEY_PATH_ENV_VAR).is_err() {
         let path = fs::cache_key_path(None::<&str>);
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -14,7 +14,7 @@ use crate::{config,
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 
 const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
-const AUTH_TOKEN_ENVVAR: &str = "HAB_AUTH_TOKEN"
+const AUTH_TOKEN_ENVVAR: &str = "HAB_AUTH_TOKEN";
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
@@ -31,7 +31,7 @@ pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
         let config = config::load()?;
         if let Some(auth_token) = config.auth_token {
             debug!("Setting default origin {} via CLI config", &auth_token);
-            env::set_var("HAB_AUTH_TOKEN", auth_token;
+            env::set_var("HAB_AUTH_TOKEN", auth_token);
         }
     }
 
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -14,6 +14,7 @@ use crate::{config,
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 
 const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
+const AUTH_TOKEN_ENVVAR: &str = "HAB_AUTH_TOKEN"
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
@@ -24,6 +25,13 @@ pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
         if let Some(default_origin) = config.origin {
             debug!("Setting default origin {} via CLI config", &default_origin);
             env::set_var("HAB_ORIGIN", default_origin);
+        }
+    }
+    if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(auth_token) = config.auth_token {
+            debug!("Setting default origin {} via CLI config", &auth_token);
+            env::set_var("HAB_AUTH_TOKEN", auth_token;
         }
     }
 
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -21,36 +21,47 @@ const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
+// fn set_env_var(env_var: &str) -> Result<()> {
+//   let config = config::load()?;
+//   if henv::var(env_var).is_err() {
+//     if let Some(config_val) = match env_var {
+//       AUTH_TOKEN_ENVVAR => config.auth_token,
+//       _ => println!("uh oh, this shouldn't happen"),
+//     };
+//     println!("config var is: {}", config_val);
+//     env::set_var(env_var, config_val)
+//   }
+// }
+
+fn set_env_var(env_var: &str, val: String) -> Result<()> {
+    debug!("Setting {}={} via config file", env_var, &val);
+    env::set_var(env_var, val);
+    Ok(())
+}
+
 pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
+    let config = config::load()?;
     if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
-        let config = config::load()?;
         if let Some(auth_token) = config.auth_token {
-            debug!("Setting {}={} via config file", AUTH_TOKEN_ENVVAR, &auth_token);
-            env::set_var("HAB_AUTH_TOKEN", auth_token);
+          set_env_var(AUTH_TOKEN_ENVVAR, auth_token)?;
         }
     }
 
     if henv::var(BLDR_URL_ENVVAR).is_err() {
-        let config = config::load()?;
         if let Some(bldr_url) = config.bldr_url {
-            debug!("Setting {}={} via config file", BLDR_URL_ENVVAR, &bldr_url);
-            env::set_var("HAB_BLDR_URL", bldr_url);
+            set_env_var(BLDR_URL_ENVVAR, bldr_url)?;
         }
     }
 
     if henv::var(CTL_SECRET_ENVVAR).is_err() {
-        let config = config::load()?;
         if let Some(ctl_secret) = config.ctl_secret {
-            debug!("Setting {}={} via config file", CTL_SECRET_ENVVAR, &ctl_secret);
-            env::set_var("CTL_SECRET_ENVVAR", ctl_secret);
+            set_env_var(CTL_SECRET_ENVVAR, ctl_secret)?;
         }
     }
 
     if henv::var(ORIGIN_ENVVAR).is_err() {
-        let config = config::load()?;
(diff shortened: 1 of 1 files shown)


qubitrenegade:
Ok, how do I test this?  I for the life of me can't get into the hab studio...  however, I do see:

```
ubuntu@ubuntu:~/habitat$ RUST_LOG=debug cargo run -p hab studio enter 2>&1 | grep hab::command::studio::enter
[2019-06-11T02:51:15Z DEBUG hab::command::studio::enter] Setting HAB_AUTH_TOKEN=FOOBARBIZBAZ via config file
[2019-06-11T02:51:15Z DEBUG hab::command::studio::enter] Setting HAB_BLDR_URL=https://foo.bar.com via config file
[2019-06-11T02:51:15Z DEBUG hab::command::studio::enter] Setting HAB_CTL_SECRET=BAZBOZBOOZE via config file
[2019-06-11T02:51:15Z DEBUG hab::command::studio::enter] Setting default origin originfoo via CLI config
[2019-06-11T02:51:15Z DEBUG hab::command::studio::enter] Setting HAB_CACHE_KEY_PATH=/home/ubuntu/.hab/cache/keys
[2019-06-11T02:51:15Z DEBUG hab::command::studio::enter] Setting ARTIFACT_PATH=/home/ubuntu/.hab/cache/artifacts
```

When I try to set the HAB_STUDIO_BINARY, it fails being unable to find `/etc/passwd` but there is one in `components/studio/defaults/etc/passwd` and my user has permissions to read the system `/etc/passwd`, so I don't understand the error...

```
ubuntu@ubuntu:~/habitat$ env HAB_STUDIO_BINARY=/home/ubuntu/habitat/components/studio/bin/hab-studio.sh target/debug/hab studio enter
   hab-studio.sh: Creating Studio at /hab/studios/home--ubuntu--habitat (default)
ERROR: Tried to copy default file for '/etc/passwd', but could not find one! Please report this error.
ubuntu@ubuntu:~/habitat$ target/debug/hab studio enter
∵ Missing package for core/hab-studio/0.83.0-dev
» Installing core/hab-studio/0.83.0-dev
☁ Determining latest version of core/hab-studio/0.83.0-dev in the 'stable' channel
✗✗✗
✗✗✗ Package not found. 
✗✗✗
ubuntu@ubuntu:~/habitat$ HAB_STUDIO_BINARY=/home/ubuntu/habitat/components/studio/bin/hab-studio.sh cargo run -p hab studio enter
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/hab studio enter`
   hab-studio.sh: Creating Studio at /hab/studios/home--ubuntu--habitat (default)
ERROR: Tried to copy default file for '/etc/passwd', but could not find one! Please report this error.
```

baumanj:
> Ok, how do I test this?

Have you taken a look at https://github.com/habitat-sh/habitat/blob/master/BUILDING.md#testing-changes ?

components/hab/src/command/studio/enter.rs
```diff
@@ -13,12 +13,39 @@ use crate::{config,
...
+            debug!("Setting {}={} via config file", BLDR_URL_ENVVAR, &bldr_url);
+            env::set_var("HAB_BLDR_URL", bldr_url);
+        }
+    }
+
+    if henv::var(CTL_SECRET_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(ctl_secret) = config.ctl_secret {
+            debug!("Setting {}={} via config file", CTL_SECRET_ENVVAR, &ctl_secret);
+            env::set_var("CTL_SECRET_ENVVAR", ctl_secret);
```
baumanj: I believe this should be
```suggestion
            env::set_var("HAB_CTL_SECRET", ctl_secret);
```

components/hab/src/command/studio/enter.rs
```diff
@@ -13,12 +13,39 @@ use crate::{config,
 
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 
+const AUTH_TOKEN_ENVVAR: &str = "HAB_AUTH_TOKEN";
```
qubitrenegade: This is defined in [`components/core/src/lib.rs`](https://github.com/habitat-sh/habitat/blob/master/components/core/src/lib.rs#L28), how do I use that value instead of defining it here?  (e.g.: if I comment this line out, I get a "not found in this scope" error)

components/hab/src/command/studio/enter.rs
```diff
@@ -13,12 +13,39 @@ use crate::{config,
...
+const BLDR_URL_ENVVAR: &str = "HAB_BLDR_URL";
```
qubitrenegade: same question, defined in [`components/core/src/url.rs`](https://github.com/habitat-sh/habitat/blob/master/components/core/src/url.rs#L4), how do I load that pub const instead of defining it here.

components/hab/src/command/studio/enter.rs
```diff
@@ -13,12 +13,39 @@ use crate::{config,
...
 const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
 pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
+    if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(auth_token) = config.auth_token {
+            debug!("Setting {}={} via config file", AUTH_TOKEN_ENVVAR, &auth_token);
```
qubitrenegade: Seems like, with the exception of the origin--and maybe we make the origin match--the debug output is "Setting <env var name>=<env var value>", this could be abstracted to a `env_debug_output()`?

components/hab/src/command/studio/enter.rs
```diff
@@ -13,12 +13,39 @@ use crate::{config,
...
+const BLDR_URL_ENVVAR: &str = "HAB_BLDR_URL";
+const CTL_SECRET_ENVVAR: &str= "HAB_CTL_SECRET";
 const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
 pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
+    if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
+        let config = config::load()?;
```
qubitrenegade: I debated on this one here...  I thought this might have been done for a reason.  

Potentially, if all of the env vars are set, you never need to call `config::load()`, but if none of the environment variables are set, we're doing a `config::load()` four times...  Does rust cache those reads, or does that cause four reads?

components/hab/src/command/studio/enter.rs
```diff
@@ -13,17 +13,55 @@ use crate::{config,
...
+const CTL_SECRET_ENVVAR: &str= "HAB_CTL_SECRET";
 const ORIGIN_ENVVAR: &str = "HAB_ORIGIN";
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
+// fn set_env_var(env_var: &str) -> Result<()> {
+//   let config = config::load()?;
+//   if henv::var(env_var).is_err() {
+//     if let Some(config_val) = match env_var {
```
qubitrenegade: Maybe I don't understand what `config.bldr_url` is doing...  it's not a function call, right? (`config.bldr_url()`)

I want to set `config_val` based on `config.<something>` depending on `env_var`...  I know how I'd do it in ruby...

The other version might be the most elegant?

components/hab/src/command/studio/enter.rs
```diff
@@ -13,17 +13,55 @@ use crate::{config,
...
+    }
+
+    if henv::var(CTL_SECRET_ENVVAR).is_err() {
+        if let Some(ctl_secret) = config.ctl_secret {
+            set_env_var(CTL_SECRET_ENVVAR, ctl_secret)?;
+        }
+    }
+
     if henv::var(ORIGIN_ENVVAR).is_err() {
-        let config = config::load()?;
```
qubitrenegade: For this commit I moved it out, so we're only executing `config::load()` once.

components/hab/src/command/studio/enter.rs
```diff
@@ -13,12 +13,39 @@ use crate::{config,
...
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
 pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
+    if henv::var(AUTH_TOKEN_ENVVAR).is_err() {
+        let config = config::load()?;
+        if let Some(auth_token) = config.auth_token {
+            debug!("Setting {}={} via config file", AUTH_TOKEN_ENVVAR, &auth_token);
+            env::set_var("HAB_AUTH_TOKEN", auth_token);
```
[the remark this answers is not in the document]
qubitrenegade: Haha, because that's how it was done originally.  This makes more sense.

components/hab/src/command/studio/enter.rs
qubitrenegade: Actually, what I really wanted was `env::set_var(CTL_SECRET_ENVVAR)` :D

qubitrenegade:
> > Ok, how do I test this?
> 
> Have you taken a look at https://github.com/habitat-sh/habitat/blob/master/BUILDING.md#testing-changes ?

Yes :)

I set the `HAB_STUDIO_BINARY`, but then it errors that it can't find `/etc/passwd`.

qubitrenegade:
> There was already a bit of duplication here, but with the additions of these 3 items, I think it's worth abstracting it out. If we add something like:
> 
> ```rust
> fn set_env_var_from_config(env_var: &str, config_val: Option<String>) {
>     if henv::var(env_var).is_err() {
>         if let Some(val) = config_val {
>             debug!("Setting {}={} via config file", env_var, val);
>             env::set_var(env_var, val);
>         }
>     }
> }
> ```
> 
> we could simplify several of them to
> 
> ```rust
>     let config = config::load()?;
> 
>     set_env_var_from_config(AUTH_TOKEN_ENVVAR, config.auth_token);
>     set_env_var_from_config(BLDR_URL_ENVVAR, config.bldr_url);
>     …
> ```
> 
> But `artifact_path` breaks that pattern. We could change `set_env_var_from_config` to return an `Option<String>` and make a generalization that could accommodate `artifact_path` as well, but I'll leave it to you whether you want to go down that road or not.

Ah ha!  I was still thinking we might `config::load();` in `set_env_var_from_config()`.  Your name is a bit more verbose than mine ("`set_env_var`"), but is more explicit/obvious so I used that :)

There's a comment somewhere, but what is `config.auth_token`?  Why is in an `Option<String>`?

I was totally not expecting a reply so late!

components/hab/src/command/studio/enter.rs
qubitrenegade: we did.

components/hab/src/command/studio/enter.rs
baumanj: The compiler error gives good advice here:
```
error[E0425]: cannot find value `AUTH_TOKEN_ENVVAR` in this scope
  --> components/hab/src/command/studio/enter.rs:36:29
   |
36 |     set_env_var_from_config(AUTH_TOKEN_ENVVAR, config.auth_token);
   |                             ^^^^^^^^^^^^^^^^^ not found in this scope
help: possible candidate is found in another module, you can import it into scope
   |
1  | use habitat_core::AUTH_TOKEN_ENVVAR;
```
Adding
```rust
```
is what you want.

components/hab/src/command/studio/enter.rs
baumanj: Again the compiler gives good advice:
```
error[E0425]: cannot find value `BLDR_URL_ENVVAR` in this scope
  --> components/hab/src/command/studio/enter.rs:41:29
   |
41 |     set_env_var_from_config(BLDR_URL_ENVVAR, config.bldr_url);
   |                             ^^^^^^^^^^^^^^^ not found in this scope
help: possible candidates are found in other modules, you can import them into scope
   |
1  | use crate::BLDR_URL_ENVVAR;
   |
1  | use habitat_core::url::BLDR_URL_ENVVAR;
```
And it even shows us that we've got duplicates of `BLDR_URL_ENVVAR`; one in [core/src/url.rs](https://github.com/habitat-sh/habitat/blob/master/components/core/src/url.rs#L4) and one in [hab/src/lib.rs](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/lib.rs#L41). Either `use` statement will work, so take your pick. We should probably remove the duplication and move all these `*_ENVVAR` constants to one place, but it's probably best to leave that for a follow-up PR (which you're welcome to tackle if interested).

baumanj:
As for testing, I didn't look closely enough at what you're doing. There's no need to use the `HAB_STUDIO_BINARY` override, since you're not modifying [hab-studio.sh](https://github.com/habitat-sh/habitat/blob/master/components/studio/bin/hab-studio.sh). You should be able to test your changes with just
```sh
cargo run -p hab studio enter
```

baumanj:
> There's a comment somewhere, but what is `config.auth_token`? Why is in an `Option<String>`?

It's the value of `auth_token` as read from `cli.toml`. It's an `Option` because it may or may not be present.

components/hab/src/command/studio/enter.rs
```diff
@@ -13,19 +13,30 @@ use crate::{config,
...
-pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
-    if henv::var(ORIGIN_ENVVAR).is_err() {
-        let config = config::load()?;
-        if let Some(default_origin) = config.origin {
-            debug!("Setting default origin {} via CLI config", &default_origin);
-            env::set_var("HAB_ORIGIN", default_origin);
+fn set_env_var_from_config(env_var: &str, config_val: Option<String>) {
+    if henv::var(env_var).is_err() {
+        if let Some(val) = config_val {
+            debug!("Setting {}={} via config file", env_var, val);
```
baumanj: It's helpful for other settings, but it may not be the best idea to print out _sensitive_ values like `HAB_AUTH_TOKEN` here. Perhaps adding an additional parameter to this function to control whether the value itself is printed, or just a notice that the env var was set via config file without printing the value would be a good approach.

components/hab/src/command/studio/enter.rs
baumanj: They won't be cached. I had the same thought about the fact that `config::load` may not be called at all and whether it made sense to change it to always be called, but I think that's the best approach. It shouldn't be expensive, and in the rare case that it causes an error (`cli.toml` is corrupt, or there are permissions problems or something), I think it's easier to track down if we *always* call `config::load`.

components/hab/src/command/studio/enter.rs
baumanj: `config.bldr_url` is an `Option<String>` value defined in [hab/src/config.rs](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/config.rs#L23). I'm not quite sure I understand the rest of your question. Does knowing the type explain it?

components/hab/src/command/studio/enter.rs
baumanj: Making them all match (with the exception of not printing the value of `HAB_AUTH_TOKEN`) sounds good.

components/hab/src/command/studio/enter.rs
qubitrenegade: doh.  I even saw that...

components/hab/src/command/studio/enter.rs
qubitrenegade: Yep, fair enough.  This may have been where my confusion came from.

I used everything I could from `crate::*`

components/hab/src/command/studio/enter.rs
qubitrenegade: Ah ha!  And `Config` is a struct so `.bldr_url` is a... "property"(?).  :+1:

qubitrenegade:
> ```shell
> cargo run -p hab studio enter
> ```

That bombs out unable to find `core/hab-studio/0.83.0-dev`

```
ubuntu@ubuntu:~/habitat$ cargo run -p hab studio enter 2>&1
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/hab studio enter`
[sudo hab-studio] password for ubuntu: 
∵ Missing package for core/hab-studio/0.83.0-dev
» Installing core/hab-studio/0.83.0-dev
☁ Determining latest version of core/hab-studio/0.83.0-dev in the 'stable' channel
✗✗✗
✗✗✗ failed to lookup address information: Name or service not known
✗✗✗
```

components/hab/src/command/studio/enter.rs
qubitrenegade: That's funny because I just read [this yesterday](http://www.informit.com/articles/article.aspx?p=1392524).

I think I like the [previous version](https://github.com/qubitrenegade/habitat-1/commit/fa88dfd75530b264711103e21ae6b6cc689ef0de)

```
fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive: Option<bool>) {
...
            match sensitive {
              Some(_) => debug!("Setting Sensitive value {} via config file", env_var),
              None => debug!("Setting {}={} via config file", env_var, val),
            }
...
```

But I think the if/else version is easier to reason about, so that's what I pushed.

components/hab/src/command/studio/enter.rs
qubitrenegade: Cool, so along the same lines of thinking :+1:

components/hab/src/command/studio/enter.rs
baumanj: One thing I really like in rust is using simple custom `enums` instead of booleans. For example
```rust
enum Sensitivty {
    OkToPrintValue,
    DoNotPrintValue,
}
```
Allows the call site to be clearer:
```rust
    set_env_var_from_config(AUTH_TOKEN_ENVVAR, config.auth_token, Sensitivty::DoNotPrintValue);
    set_env_var_from_config(BLDR_URL_ENVVAR, config.bldr_url, Sensitivty: OkToPrintValue);
```
And then `match` ensures exhaustive handling:
```rust
            match sensitive {
                Sensitivty::DoNotPrintValue =>  debug!("Setting {}=REDACTED (sensitive) via config file", env_var)
                Sensitivty::OkToPrintValue => debug!("Setting {}={} via config file", env_var, val),
            }
```

components/hab/src/command/studio/enter.rs
baumanj: I believe the common term in rust is ["field"](https://doc.rust-lang.org/rust-by-example/custom_types/structs.html)

baumanj:
> ```shell
> ```shell
> cargo run -p hab studio enter
> ```
> 
> 
> 
> That bombs out unable to find core/hab-studio/0.83.0-dev
> ```

That's ok though because it's *after* your code runs, right? I mean, it's not perfect, but it should be enough to run your code.

You might be able to get further with setting HAB_STUDIO_BINARY, but I think looking at the debug output is enough to validate your code is working.

components/hab/src/command/studio/enter.rs
qubitrenegade: Ta da!  There's always a way to do what I want (well, what I _really_ want is default values/optional parameters) just couldn't quite get there.

I went with

```
enum Sensitivity {
    PrintValue,
    NoPrintValue,
}
```

Because it's a bit less typing ;)

qubitrenegade:
> > ```shell
> > ```shell
> > cargo run -p hab studio enter
> > ```
> > 
> > 
> > That bombs out unable to find core/hab-studio/0.83.0-dev
> 
> That's ok though because it's _after_ your code runs, right? I mean, it's not perfect, but it should be enough to run your code.
> 
> You might be able to get further with setting HAB_STUDIO_BINARY, but I think looking at the debug output is enough to validate your code is working.

I'm like... 75%...? certain this does what we want/expect and will leave the studio in the desired state.

components/hab/src/command/studio/enter.rs
```diff
@@ -9,23 +9,49 @@ use crate::{common::ui::UI,
...
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
-pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
-    if henv::var(ORIGIN_ENVVAR).is_err() {
-        let config = config::load()?;
-        if let Some(default_origin) = config.origin {
-            debug!("Setting default origin {} via CLI config", &default_origin);
-            env::set_var("HAB_ORIGIN", default_origin);
+enum Sensitivity {
```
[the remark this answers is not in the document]
qubitrenegade: I can't see anything, buildkite seems to require a login.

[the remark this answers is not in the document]
baumanj: Oops, sorry about that. That's also part of the work in progress. I think this last change should do the trick, though. I'll run it through CI and either merge or let you know if there's anything else needed.

baumanj:
Thanks again for a great submission, @qubitrenegade!

bdangit:
Agree! This is awesome!

PR #6645, opened 2019-06-11
[Studio][Tests] Use absolute drive path for windows studio name

smacfarlane:
When exiting the studio inside an Await session, the CWD would remain at the `[Habitat]` junction location.  This caused the Linux style path of `/hab/studios/studio-internals-test` to not be resolvable and the studio rm would fail, resulting in a failing test.  

This changes the studio name to use Windows-style drive letter paths, allowing the `hab studio rm` to behave as expected. 

### DIFF
--- a/components/studio/test/shared/studio-enter.ps1
+++ b/components/studio/test/shared/studio-enter.ps1
@@ -8,7 +8,7 @@ $env:HAB_NOCOLORING = "true"
 # Await has trouble parsing non-ascii glyphs
 $env:HAB_GLYPH_STYLE = "ascii"  
 $exit_code = 0
-$studio_name = "/hab/studios/studio-internals-test"
+$studio_name = "c:\hab\studios\studio-internals-test"
 try { 
     Start-AwaitSession 
     Send-AwaitCommand "$studio_command enter -o $studio_name" 
@@ -25,8 +25,10 @@ try {
 
 } finally { 
     Write-Host "Cleaning up"
-    Send-AwaitCommand "exit"
     
+    Send-AwaitCommand "exit"
+    Receive-AwaitResponse
+
     # Await won't block on "Waiting for supervisor to finish..." 
     # Copy the studio shutdown behavior and block for a bit until the supervisor has finished
     $retry = 0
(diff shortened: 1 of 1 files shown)


components/studio/test/shared/studio-enter.ps1
```diff
@@ -8,7 +8,7 @@ $env:HAB_NOCOLORING = "true"
 # Await has trouble parsing non-ascii glyphs
 $env:HAB_GLYPH_STYLE = "ascii"  
 $exit_code = 0
-$studio_name = "/hab/studios/studio-internals-test"
+$studio_name = "c:\hab\studios\studio-internals-test"
```
mwrock: This is fine our purposes here and indeed almost any scenario. I always like to use:
```
$ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath("/some/path/I/want/to/cannonacalize")
```
This will always transform the given string to a Windows friendly absolute path whether or not the path actually exists and applicable to one's current context. In otherwords, if you were in the `d:\` drive, this would resolve to the `d` drive.

Note there is also the simpler and more intuitively named `Resolve-Path` which does the same thing but requires the given path to exist. If I know or expect the path to already exist, I prefer this command.

components/studio/test/shared/studio-enter.ps1
```diff
@@ -36,12 +38,13 @@ try {
...
-    Send-AwaitCommand "$studio_command rm"
-    
+    Send-AwaitCommand "$studio_command rm -o $studio_name"
+    Receive-AwaitResponse -Stream
+
     # Add the same behavior for `hab studio rm` to ensure that the studio is fully cleaned up before 
     # stopping the Await session.
     $retry = 0
-    while(($retry -lt 10) -and (Test-Path "$studio_name")) {
+    while(($retry -lt 5) -and (Test-Path "$studio_name")) {
```
mwrock: I still like keeping this to a higher number because if we reach the limit, subsequently deleting the studio environment is guaranteed to fail because of open file handles from the hab binary.

PR #6648, opened 2019-06-12
allow windows service to set arbitrary environment variables

### LINKED ISSUE #6647: Not possible to set RUST_LOG to 'warn' for supervisor in windows service
Currently the Windows Service exposes a application setting `debug` and when this is set, it sets the `RUST_LOG` environment variable to `debug` otherwise it removes the variable.

We should allow the setting of any arbitrary environment variable through the service application settings which would allow the user to set `RUST_LOG` to any desired value.

mwrock:
fixes #6647 

Now users can create any application setting prefixed with `ENV_` and that setting (sans the `ENV_`) will be added to the environment of the service.

This also deprecates the debug setting which we will remove in a future release.

### DIFF
--- a/components/windows-service/HabService.cs
+++ b/components/windows-service/HabService.cs
@@ -74,7 +74,7 @@ namespace HabService
             try
             {
                 ConfigureDebug();
-                ConfigureSupSignal();
+                ConfigureEnvironment();
 
                 // DataReceivedEventArgs.Data will return text in the default system
                 // locale. We can change that via System.Console.OutputEncoding for a console
@@ -108,6 +108,17 @@ namespace HabService
             }
         }
 
+        private static void ConfigureEnvironment()
+        {
+            var envPrefix = "ENV_";
+            foreach(var key in ConfigurationManager.AppSettings.AllKeys) {
+                var i = key.Trim().ToUpper();
+                if (i.StartsWith(envPrefix)) {
+                    Environment.SetEnvironmentVariable(i.Substring(envPrefix.Length), ConfigurationManager.AppSettings[key]);
+                }
+            }
+        }
+
         private static void ConfigureDebug()
         {
             if (ConfigurationManager.AppSettings["debug"] != null)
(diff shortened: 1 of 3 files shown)


components/windows-service/HabService.cs
```diff
@@ -108,6 +108,17 @@ protected override void OnStart(string[] args)
             }
         }
 
+        private static void ConfigureEnvironment()
+        {
+            var envPrefix = "ENV_";
+            foreach(var key in ConfigurationManager.AppSettings.AllKeys) {
+                var i = key.Trim().ToUpper();
```
markan: Curious about when whitespace can appear (and Trim is needed). Is there always whitespace, or is that only when someone is doing something wrong? The effect here is to fold "ENV_WIERD " (with space) and "ENV_WIERD" (without space) together; which takes effect depends on the order from AllKeys. 

There's a similar effect from the ToUpper; if both "env_case" and "ENV_CASE" are present only one will passed.

Would it make sense to be more restrictive, and only pass 'well formed' ENV_* variables (no whitespace, all upper, etc.)

It's been a while since I've done C#/windows, so very well could be missing something.

mwrock: The `trim` is really defensive handling of a config file content like:

```
<add key="   HAB_FEAT_IGNORE_SIGNALS   " value="true" />
```

If some one had multiple entries for the same key. For example one with spaces and the other without, it seems like it would be unintentional. However even if we omitted the trim, windows  would effectively trim the key when we added it to the environment.

Same holds true for the `ToUpper`. On Windows, environment var keys are case insensitive. So it would likely be unintentional to add 2 keys that only vary in casing.

The main benefit of trimmimg and upper casing the key is it just makes it simpler to do the `StartsWith("ENV_")` check.

markan: Thanks, makes sense to me.

chefsalim:
Will this change require updates to the Windows build worker setup?

mwrock:
no this should not require any builder changes @chefsalim

PR #6653, opened 2019-06-12
Add Artifactory as BuilderAPIProvider

chefsalim:
This PR adds in a new Artifactory provider for the Builder API.  This allows the use of Artifactory directly as a Depot, albeit with certain limitations:
* Artifactory repos (one for each origin, named "habitat-artifact-store-local.<ORIGIN>") need to be pre-created, and users given permission appropriately by the Artifactory admin - ie, there is no origin/user management via hab
* Only package, public/private key management, and promote/demote channel operations are supported
* Job management or build related commands are not supported
* All calls are authenticated - ie, a HAB_AUTH_TOKEN is required
* Only works with Artifactory Pro version 

In order to enable this feature:
```
export HAB_BLDR_PROVIDER=artifactory
export HAB_BLDR_URL=<your artifactory url>
export HAB_AUTH_TOKEN=<your artifactory API key>
```

This feature should be considered Alpha level for early adopter usage and feedback purposes.

tenor-68370775

NOTE TO REVIEWERS: not looking for a deep review on this, mostly please eyeball for anything obviously broken I might have blindly overlooked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -699,6 +699,11 @@ version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "env"
+version = "0.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "env_logger"
 version = "0.5.13"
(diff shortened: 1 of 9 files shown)


davidMcneil:
This looks reasonable to me.

PR #6659, opened 2019-06-14
Fix conflict in search interface

chefsalim:
Fixing a conflict between a couple of PRs that broke the build


mwrock:
Seems like all but the last line of `new` and `create` could be extracted to a separate function to eliminate duplication but I find the interface confusing. When would I use `new` over `create` it feeles like `create` should be `create_boxed` just to make that obvious.

davidMcneil:
That is a good idea. This [PR](https://github.com/habitat-sh/habitat/pull/6660) removes the duplication.

PR #6660, opened 2019-06-14
Remove duplication in BuilderAPI create

davidMcneil:


davidMcneil:
We could take this a step further and add `new` and `create` to the `BuilderAPIProvider` api then we could then have a blanket `create` for all `BuilderAPIProvider`s.

PR #6663, opened 2019-06-14
bump travis ruby and build www in a pr

mwrock:

### DIFF
--- a/support/ci/deploy_website.sh
+++ b/support/ci/deploy_website.sh
@@ -2,14 +2,26 @@
 
 set -eu
 
-if [[ "${TRAVIS_PULL_REQUEST}" = "false" ]] && [[ "${TRAVIS_BRANCH}" = "master" ]]; then
+# TRAVIS_PULL_REQUEST is set to the pull request number if the current job is a pull request build, or false if it’s not.
+#  https://docs.travis-ci.com/user/environment-variables/#convenience-variables
+if [[ "${TRAVIS_PULL_REQUEST}" = "false" ]];then
+  pr_triggered_build=false
+else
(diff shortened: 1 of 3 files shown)


support/ci/deploy_website.sh
```diff
@@ -6,10 +6,14 @@ if [[ "${TRAVIS_PULL_REQUEST}" = "false" ]] && [[ "${TRAVIS_BRANCH}" = "master"
...
+  echo "We are on a PR. Going to just build the site."
+  cd www
+  make build
 else
-  if [[ "${TRAVIS_PULL_REQUEST}" = "false" ]] && [[ "${TRAVIS_BRANCH}" =~ ^acceptance_deploy ]]; then
-    echo "We are on a PR or against the master branch. Deploying to Acceptance."
-    cd www
-    BUILDER_WEB_URL="https://bldr.acceptance.habitat.sh" GITHUB_APP_URL="https://github.com/apps/habitat-builder-acceptance" make acceptance
-  fi
+  echo "Not building web site."
```
baumanj: What case does this represent? It's not a triggered by a PR and the branch is neither `master` nor `acceptance_deploy`, so I guess it's a manually-triggered run for some other branch. Would it be bad to go ahead and do `make build` in this case?

mwrock: yeah. I mainly put it is there because it would tell me I did something wrong above. It would only fire merging to another branch.

mwrock: also good call on the above, I had to find the travis docs to explain why looking for `true` was not working with `TRAVIS_PULL_REQUEST`

PR #6670, opened 2019-06-18
cross platform binstubs with package environment

mwrock:
This refactors #5696 to normalize the binstubs we currently create for windows to apply to linux as well and then adds in the setting of the environment variables. This includes a couple deviations from the behavior originally in #5696:

* It does not leverage `hab pkg exec`. While it is convenient to delegate to that existing logic. It introduces the dilemma of requiring an existing hab to be on the path. This is a fairly reasonable expectation but not guaranteed. Fortunately, `hab pkg exec` does not really do much at all so we are not missing out on much.
* Instead of creating a binstub in a `/hab/binstubs` folder and then creating a symlink in `dest_path`, this just creates the binstub in `dest_path` and creates no symlink which is exactly what we do for windows.

### DIFF
--- a/components/hab/src/command/pkg/binlink.rs
+++ b/components/hab/src/command/pkg/binlink.rs
@@ -1,11 +1,9 @@
-#[cfg(windows)]
-use std::fs::File;
-#[cfg(windows)]
-use std::io::{BufRead,
-              BufReader};
-use std::{env,
-          fs,
-          io::Write,
+use std::{collections::HashMap,
+          env,
+          fs::{self,
+               File},
+          io::{BufRead,
+               BufReader},
           path::{Path,
                  PathBuf}};
 
@@ -13,7 +11,6 @@ use crate::{common::ui::{Status,
                          UIWriter,
                          UI},
             hcore::{fs as hfs,
-                    fs::ROOT_PATH,
                     package::{PackageIdent,
                               PackageInstall}}};
 
@@ -21,7 +18,13 @@ use crate::error::{Error,
                    Result};
 
 #[cfg(windows)]
-const BAT_COMMENT_MARKER: &str = "REM";
+const COMMENT_MARKER: &str = "REM";
+#[cfg(unix)]
+const COMMENT_MARKER: &str = "#";
+#[cfg(windows)]
+const SET_OR_EXPORT: &str = "SET";
+#[cfg(unix)]
+const SET_OR_EXPORT: &str = "export";
 
 pub fn start(ui: &mut UI,
              ident: &PackageIdent,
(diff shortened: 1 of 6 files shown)


components/hab/src/command/pkg/binlink.rs
```diff
@@ -208,16 +197,65 @@ impl Binlink {
...
+        let mut exports = String::new();
+        for (key, value) in env.into_iter() {
+            exports.push_str(&format!("export {}={}\n", key, value));
+        }
+        format!(
+                "\
+#!/bin/sh
+# source='{0}'
+{1}
+{0} \"$@\"",
```
fnichol: You might want to consider using [exec](https://en.wikipedia.org/wiki/Exec_(system_call)) to turn the shell script process into the target program (`exec {0} \"$@\""`)

fnichol: If you use Rust's [raw string literals](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) here, then you can avoid needing to escape the double quotes (it's an odd looking string like `r#"I can use "double quotes""#`). The only catch is that the `"\` on line 211 would have to become `"#!/bin/sh` as the backslash would be added into the string.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -208,16 +197,65 @@ impl Binlink {
...
+    }
+
+    #[cfg(target_os = "windows")]
+    fn stub_template(&self, env: HashMap<String, String>) -> String {
+        let mut exports = String::new();
+        for (key, value) in env.into_iter() {
+            exports.push_str(&format!("SET {}={}\n", key, value));
+        }
+        format!(
+                "\
```
fnichol: If you'd like to be able to indent this multi-line string, I finally found a solution that works for me in the [indoc](https://crates.io/crates/indoc) crate which provides an `indoc!` macro to help with this problem. Here's an [example](https://github.com/fnichol/mtoc/blob/aebbfe6ada535dcb8eede38f3f2084c0760f31fb/mtoc-parser/src/write.rs#L1031-L1047) with a raw string to boot.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -208,16 +197,65 @@ impl Binlink {
...
+    }
+
+    #[cfg(not(target_os = "windows"))]
+    fn write_binstub(&self, content: &[u8]) -> Result<()> {
+        use std::{io::Write,
+                  os::unix::fs::OpenOptionsExt};
+        fs::OpenOptions::new().create(true)
+                              .write(true)
+                              .truncate(true)
+                              .mode(0o770)
```
fnichol: Is it intended to not add `rx` to the other group (i.e. `0o775`)?

mwrock:
thanks for the review @fnichol !!

baumanj:
Will plan to look at this today

mwrock:
fixes #4493

mwrock:
This includes all of your comments @baumanj which I super appretiate. I am much happier with the platform specific templating.

@chefsalim I know you had some concerns regarding disruption in #5696 and I am wondering if you have the same level of concern now. I feel this is less potentially disruptive since it has no assumption that `hab` is in the path and the stubs are pretty vanilla.

Also any thoughts from @jamessewell or @themightychris ?

jamessewell:
Great!

PR #6673, opened 2019-06-20
Disable artifactory mirror for choco

scotthain:


mwrock:
my only concern here is that this was added because all our builds were failing getting rate limited by chocolatey. I wonder if we should just add chocolatey to the container image @scotthain ?

Issue #6386, opened 2019-04-06
Bump core/powershell to 6.2.1


mwrock:
https://github.com/PowerShell/PowerShell/releases/tag/v6.2.1

mwrock:
closed by #6676

PR #6676, opened 2019-06-21
use explicit version of powershell

mwrock:
This links to a specific version of powershell. I built everything and validated this version performs as expected.


mwrock:
fixes #6386

PR #6679, opened 2019-06-21
update widestring crate

### LINKED ISSUE #6456: migrate to latest widestring
We had to pin to 0.3.0 (latest is currently 0.4.0) because after making a couple modifications to compile against the latest version, some of the windows API calls using wide strings started failing.

mwrock:
fixes #6456 

I thought I remembered this being more convoluted in March but I was able to update the crate this time around without any code changes or failing tests. I even added one here for good measure.

### DIFF
--- a/components/core/src/os/users/windows.rs
+++ b/components/core/src/os/users/windows.rs
@@ -68,6 +68,12 @@ mod tests {
     use super::*;
 
     #[test]
+    fn get_uid_of_current_user() {
+        assert!(get_current_username().map(|s| get_uid_by_name(&s))
+                                      .is_some())
+    }
+
+    #[test]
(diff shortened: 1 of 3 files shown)


components/core/src/os/users/windows.rs
```diff
@@ -67,6 +67,11 @@ mod tests {
 
     use super::*;
 
+    #[test]
+    fn get_uid_of_current_user() {
+        assert!(get_uid_by_name(&get_current_username().unwrap()).is_some());
```
baumanj: Functionally equivalent, but I find this clearer (and prefer to avoid unwraps)
```suggestion
        assert!(get_current_username().map(|s| get_uid_by_name(&s)).is_some())
```

PR #6683, opened 2019-06-24
resolve build deps before run deps

### LINKED ISSUE #6637: build deps should appear on PATH before runtime deps
While it is rare for different deps to include the same binary names. `dotnet-core-sdk` and `dotnet-core` are good examples of two deps that have a binary with the same name `dotnet.exe` and are included in the same plan. During a build the sdk is used and the runtime dep just exists for...well...runtime. It seems logical to expect that if runtime and buildtime deps have a binary with the same name and it is called in a build, the buildtime dep should take precedence.

mwrock:
fixes #6637 

With this, build deps appear on the path before runtime deps.

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1144,8 +1144,8 @@ _populate_dependency_arrays() {
   # Build `${pkg_all_deps_resolved[@]}` containing all direct build and run
   # dependencies. The build dependencies appear before the run dependencies.
   pkg_all_deps_resolved=(
-    "${pkg_deps_resolved[@]}"
     "${pkg_build_deps_resolved[@]}"
+    "${pkg_deps_resolved[@]}"
   )
 
   # Build an ordered set of all build and run dependencies (direct and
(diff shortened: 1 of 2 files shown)


components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1144,24 +1144,24 @@ _populate_dependency_arrays() {
...
+  # 2. All direct run dependencies
+  # 3. All unique transitive build dependencies that aren't already added
+  # 4. All unique transitive run dependencies that aren't already added
   pkg_all_tdeps_resolved=(
-    "${pkg_deps_resolved[@]}"
     "${pkg_build_deps_resolved[@]}"
+    "${pkg_deps_resolved[@]}"
   )
-  for dep in "${pkg_tdeps_resolved[@]}" "${pkg_build_tdeps_resolved[@]}"; do
+  for dep in "${pkg_build_tdeps_resolved[@]}" "${pkg_tdeps_resolved[@]}"; do
```
christophermaier: Is it possible to have some tests for this behavior? It's hard to determine from reading the code if this does the right thing (or if this is reversed by the code adding these to the `$PATH`)

😭

PR #6684, opened 2019-06-24
provide a better error when no value is passed to Set-BuildtimeEnv or Set-RuntimeEnv

### LINKED ISSUE #6487: Confusing error message for invalid parameters provided to `Set-RuntimeEnv`
Here is an example of improper parameters provided to the `Set-RuntimeEnv` function in plan.ps1. 

```
  Set-RuntimeEnv GEM_PATH" $pkg_prefix\vendor\bundle\ruby\2.6.0"
```

You can see that there is no space between `GEM_PATH` and the value. The error message from the studio when this code is evaluated isn't very straight forward and doesn't really tell you what is wrong.  

```
New-Item : Cannot process argument because the value of argument "value" is null. Change the value of argument "value" to a non-null value.
At C:\hab\pkgs\core\hab-studio\0.79.1\20190410223129\bin\hab-plan-build.ps1:1270 char:13
+             New-Item -Name $k -Value $env["Runtime"][$k] -ItemType Va ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidArgument: (GEM_PATH C:\hab...ndle\ruby\2.6.0:String) [New-Item], PSArgumentNullException
+ FullyQualifiedErrorId : NewItemValueNotSpecified,Microsoft.PowerShell.Commands.NewItemCommand
```

mwrock:
fixes #6487

### DIFF
--- a/components/plan-build-ps1/bin/hab-plan-build.ps1
+++ b/components/plan-build-ps1/bin/hab-plan-build.ps1
@@ -1523,11 +1523,17 @@ function __assemble_legacy_paths($DepPath, $pathFile) {
     $path
 }
 
-function Set-BuildtimeEnv($VarName, $VarValue, [switch]$force) {
+function Set-BuildtimeEnv(
+  $VarName,
+  $VarValue = $(throw "Must provide a value to Set-BuildtimeEnv for key '$VarName'"),
+  [switch]$force) {
     set_env "BuildTime" @PSBoundParameters
 }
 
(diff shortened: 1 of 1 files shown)


mwrock:
No that is a great question @davidMcneil ! The problem with using `mandatory` params here is that those will prompt the user if not provided which is not what we want in a studio build.

PR #6687, opened 2019-06-25
Refactor hook runner

davidMcneil:
_Note: This is change is purely a refactor to simplify code. I may have missed the necessity of the previous method._

Instead of spawning a new thread and using a one shot channel to communicate when the hook has completed, can we use [`future::lazy`](https://docs.rs/futures/0.1.27/futures/future/fn.lazy.html) to run the closure directly with the current tokio runtime?


components/sup/src/manager/service/hook_runner.rs
```diff
@@ -41,40 +41,28 @@ impl<H> HookRunner<H> where H: Hook + Sync
...
-        }
+        future::lazy(Box::new(move || {
+                         // _timer is for Prometheus metrics, but we also want
+                         // the runtime for other purposes. Unfortunately,
+                         // we're not able to use the same timer for both :(
+                         let _timer = hook_timer(H::file_name());
+                         let start = Instant::now();
+                         let exit_value =
+                             self.hook
+                                 .run(&self.service_group, &self.pkg, self.passwd.as_ref());
```
christophermaier: The problem here is that a long-running hook will gum up the executor, because this future will only yield once the hook is done running. Though the previous formulation is more verbose, it has the advantage of keeping the heavy lifting off to the side in another thread; the oneshot channel is the only real future bit there, and polling it is a very quick operation.

See https://docs.rs/futures/0.1.27/futures/future/trait.Future.html#runtime-characteristics for details.

The closure that `future::lazy` takes is just run normally (see [here](https://docs.rs/futures/0.1.27/src/futures/future/lazy.rs.html#64)), so there's no opportunity to break out of it and come back later if it's taking a long time to run.

The comment on Line 4 above points the way forward to a cleaner implementation.

davidMcneil:
See this [comment](https://github.com/habitat-sh/habitat/pull/6687#discussion_r297340053).

PR #6694, opened 2019-06-27
Bump version to 0.83.0

christophermaier:
Also, remove PR #6670 from the changelog, since that was reverted with #6689.


smacfarlane:
tenor

PR #6697, opened 2019-06-28
do not overwrite PSModulePath when setting up build environment

mwrock:
The `PSModulePath` environment variable is where Powershell looks for modules. When entering a powershell shell, powershell will ensure that its own `modules` directory which include the built in modules that ship with powershell are added to this variable when the shell starts.

If a plan includes something like:
```
function Invoke-SetupEnvironment {
    Push-RuntimeEnv "PSModulePath" "\hab\pkgs\$pkg_origin\$pkg_name\$pkg_version\$pkg_release\Modules"
}
```
This will overwrite the existing `PSModulePath` and potentially cause powershell commands that leverage the distribution bundled modules to fail.

An example would be any plan that takes a dep on `core/dsc-core` which includes the above code.

This PR does 3 things:
1. Sets the `PSModulePath` in the Studio creation code so that any module paths other than those included with the powershell vendored in the studio are not included.
2. Add `PSModulePath` to the list of variables in `hab-plan-build` that should not be overwritten.
3. Removes a function I found to be entirely dead code.

### DIFF
--- a/components/plan-build-ps1/bin/hab-plan-build.ps1
+++ b/components/plan-build-ps1/bin/hab-plan-build.ps1
@@ -1267,7 +1267,7 @@ function Invoke-SetupEnvironmentWrapper {
     # the real environment, except for PATH; for that, push the
     # runtime path onto the front of the system path
     foreach($k in $env["Runtime"].keys) {
-        if(@("PATH", "LIB", "INCLUDE") -contains $k) {
+        if(@("PATH", "LIB", "INCLUDE", "PSMODULEPATH") -contains $k) {
             $currentVal = ""
             if(Test-path env:\$k) {
                 $currentVal = Get-Content env:\$k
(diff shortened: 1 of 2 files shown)


components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1267,7 +1267,7 @@ function Invoke-SetupEnvironmentWrapper {
     # the real environment, except for PATH; for that, push the
     # runtime path onto the front of the system path
     foreach($k in $env["Runtime"].keys) {
```
christophermaier: Is it significant that this "Runtime", while above we have "RunTime" (same with "Buildtime" below)?

components/studio/bin/hab-studio.ps1
```diff
@@ -231,6 +231,7 @@ function New-Studio {
   )
 
   $env:PATH = [String]::Join(";", $pathArray)
+  $env:PSModulePath = "$PSScriptRoot\powershell\Modules"
```
christophermaier: Similarly, is it significant that this is "PSModulePath", while above we have "PSMODULEPATH"?

components/plan-build-ps1/bin/hab-plan-build.ps1
mwrock: No. Powershell hash keys and variables are not case sensitive.

components/studio/bin/hab-studio.ps1
mwrock: same as above applies. Generally `PSModulePath` is used and what is automatically set but I kept it uppercase above just to make it look like the others. To windows/powershell they are all the same.

components/plan-build-ps1/bin/hab-plan-build.ps1
christophermaier: Hrmm... maybe it'd be better to be consistent.

If nothing else, it'd stop me from asking these questions :wink:

components/studio/bin/hab-studio.ps1
christophermaier: Same suggestion above re: consistency... if `PSModulePath` is customary in Powershell world, we should probably stick with that.

Issue #6657, opened 2019-06-13
In Windows, hab pkg exec should add to PATH instead of replace it


mwrock:
This is definitely a problem on Windows and **might** be on Linux too. For instance, any windows program pretty much expects `c:\windows\system32` to be on the path and may fail if it is not. `hab pkg exec` adds the package deps and any other vars from `Invoke-SetupEnvironment` to the environment. Ideally it should add to the `PATH` and not replace it.

smacfarlane:
On Linux, at least, this is correct behavior. If we were to add to the path rather than replace we would lose some of the isolation guarantees. Depending on ordering, we would either potentially use system commands instead of the desired hab commands or mask a missing dependency.

@mwrock Would it be fair to say for `hab pkg exec` the behavior should be 

Windows: `$env:PATH=C:\windows\system32:$env:RUNTIME_ENVIRONMENT_PATH`
Linux: `PATH=$RUNTIME_ENVIRONMENT_PATH`   

Where $RUNTIME_ENVIRONMENT_PATH comes from the metadata file `RUNTIME_ENVIRONMENT` ?

mwrock:
close. There are a couple other windows system dirs to add as well. Ultimately I'd expect the same behavior that the supervisor uses which is to prepend the dep paths to the current path. That way those deps should take precedence over anything else in the path.

mwrock:
Thinking more on this I am leaning that this behavior (appending existing `PATH` entries) would be correct linux behavior. I think the isolation guarantees are largely the responsibility of the studio environment. Even in a studio, the build system and binlinks will honor the existing `PATH` entries which can be edited inside a studio environment.

smacfarlane:
@mwrock I don't think you want system paths to come first in Linux.  I kinda don't feel they should be present at all, but you definitely don't want them to come first.  

Contrived example: consider a package that relies on busybox behavior for core utility packages (or even bsd versions of them if they were to be made available). By putting the system paths first, you'll always get those versions of the packages,coupling the behavior of the package to the distro it's installed on and in the worst case breaking the package.

mwrock:
totally not first @smacfarlane . I think you always want the packages runtime dep `PATH` entries to be first. In the supervisor and plan-build we prepend the runtime path entries.

smacfarlane:
Ahh, I must be misunderstanding what you wrote, sorry.  

[...]

PR #6700, opened 2019-06-28
redux: cross platform binstubs with package environment

mwrock:
This can be considered a continuation of #6670 which was reverted after we found that the binstubs in a linux studio were resulting in recursive symlinks. This is because the binstub interpreter is `/bin/sh` and if you binlinked a package including `sh`, it would link to itself as an interpreter.

Note: To tease out what changed between this PR and the original #6670, you may just want to look at the commit(s) after the first one that reverts the revert.

In summary, this makes the following changes:
* changes the interpreter from `/bin/sh` to the source of the link IF the link destination is also `/bin/sh`.
* Look for symlinks when checking for existing links. This will catch links from previous versions or one that was maybe manually created. Because we were not looking for these before, we were not removing the existing the link and thus writing the binstub was following the link and overwriting the source binary.
* overwrite existing links with `--force` even if the source binary path matches. This will allow a user to overwrite the old style links with these new ones if they want.
* Append the existing `PATH` to the `PATH` being set in the link

### Notes on appending existing `PATH`
There are some scenarios where a binary may need something on the path that could never be anticipated at build time. Examples would be:
* `sh`/`bash` where a script needs stuff on the path unrelated to bash itself
* `rust` where a `build.rs` expects something on the path unrelated to `cargo`
* `inspec` where it calls into `Windows Powershell` (not PS Core) that installs with windows and is not packaged.

The current binlinking behavior simply symlinks binaries to a known directory on the `PATH` (`/bin` by default) and therefore they include whatever is currently on the `PATH` so limiting the `PATH` to the binary's runtime deps could potentially be a breaking change but should work in the vast majority of cases.

Note that both the Supervisor and the build system also retain the existing `PATH` entries when running package hooks and building a plan. `hab pkg exec` does not but there is an [issue](https://github.com/habitat-sh/habitat/issues/6657) proposing that we append the existing `PATH` entries for `hab pkg exec` as well.

### DIFF
--- a/components/hab/src/command/pkg/binlink.rs
+++ b/components/hab/src/command/pkg/binlink.rs
@@ -22,9 +22,15 @@ const COMMENT_MARKER: &str = "REM";
 #[cfg(unix)]
 const COMMENT_MARKER: &str = "#";
 #[cfg(windows)]
+const PATH_SEPARATOR: char = ';';
+#[cfg(unix)]
+const PATH_SEPARATOR: char = ':';
+#[cfg(windows)]
 const SET_OR_EXPORT: &str = "SET";
 #[cfg(unix)]
 const SET_OR_EXPORT: &str = "export";
+
+const DEFAULT_INTERPRETER: &str = "/bin/sh";
 
 pub fn start(ui: &mut UI,
              ident: &PackageIdent,
@@ -35,10 +41,10 @@ pub fn start(ui: &mut UI,
              -> Result<()> {
     let dst_path = fs_root_path.join(dest_path.strip_prefix("/")?);
     ui.begin(format!("Binlinking {} from {} into {}",
-                     &binary,
-                     &ident,
+                     binary,
+                     ident,
                      dst_path.display()))?;
-    let pkg_install = PackageInstall::load(&ident, Some(fs_root_path))?;
+    let pkg_install = PackageInstall::load(ident, Some(fs_root_path))?;
     let mut src = match hfs::find_command_in_pkg(binary, &pkg_install, fs_root_path)? {
         Some(c) => c,
         None => {
@@ -55,29 +61,28 @@ pub fn start(ui: &mut UI,
                   format!("parent directory {}", dst_path.display()))?;
         fs::create_dir_all(&dst_path)?
     }
-    let binlink = Binlink::new(&src, &dst_path)?;
+
+    let binlink = Binlink::new(&src, &dst_path, &PathBuf::from(DEFAULT_INTERPRETER))?;
     let ui_binlinked = format!("Binlinked {} from {} to {}",
-                               &binary,
-                               &pkg_install.ident(),
-                               &binlink.dest.display(),);
-    match Binlink::from_file(&binlink.dest) {
+                               binary,
+                               pkg_install.ident(),
+                               binlink.link.display(),);
+    match Binlink::from_file(&binlink.link) {
         Ok(link) => {
-            if force && link.src != src {
-                fs::remove_file(&link.dest)?;
+            if force {
+                fs::remove_file(link.link)?;
                 binlink.link(pkg_install.environment_for_command()?)?;
-                ui.end(&ui_binlinked)?;
-            } else if link.src != src {
+                ui.end(ui_binlinked)?;
+            } else {
                 ui.warn(format!("Skipping binlink because {} already exists at {}. Use --force \
                                  to overwrite",
-                                &binary,
-                                &link.dest.display(),))?;
-            } else {
-                ui.end(&ui_binlinked)?;
+                                binary,
+                                link.link.display(),))?;
             }
         }
         Err(_) => {
             binlink.link(pkg_install.environment_for_command()?)?;
-            ui.end(&ui_binlinked)?;
+            ui.end(ui_binlinked)?;
         }
     }
 
@@ -96,7 +101,7 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
                           fs_root_path: &Path,
                           force: bool)
                           -> Result<()> {
-    let pkg_path = PackageInstall::load(&pkg_ident, Some(fs_root_path))?;
+    let pkg_path = PackageInstall::load(pkg_ident, Some(fs_root_path))?;
     for bin_path in pkg_path.paths()? {
         for bin in fs::read_dir(fs_root_path.join(bin_path.strip_prefix("/")?))? {
             let bin_file = bin?;
@@ -134,7 +139,7 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
                     continue;
                 }
             };
-            self::start(ui, &pkg_ident, &bin_name, dest_path, &fs_root_path, force)?;
+            self::start(ui, pkg_ident, &bin_name, dest_path, fs_root_path, force)?;
         }
     }
     Ok(())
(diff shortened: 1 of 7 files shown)


baumanj:
Currently reviewing; thanks for your patience

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+                             src:                 lnk,
+                             default_interpreter: PathBuf::from(DEFAULT_INTERPRETER), })
+            }
+            Err(_) => {
+                let file = File::open(path)?;
+                for line in BufReader::new(file).lines() {
+                    let ln = line?;
+                    if ln.to_uppercase().starts_with(COMMENT_MARKER) {
+                        let (_, rest) = ln.split_at(COMMENT_MARKER.len());
+                        if let Ok(Table(toml_exp)) = rest.parse() {
```
baumanj: The cascade of `if let`s may be more readable re-cast with an FP approach:
```rust
                rest.parse()
                    .map(|Table(toml_exp)| toml_exp.get("source"))
                    .map(toml::value::Value::as_str)
                    .map(|val| ...);
```
And then we might be able to reuse `Self::from_link` at the bottom.

baumanj: I poked at this a bit, and here's an approach that will actually compile:
```rust
                if let Some(source) = Self::get_source_from_toml(rest) {
                    return Ok(Self::with_default_interpreter(source.into(), path.into()));
                }
```
with
```rust
    fn with_default_interpreter(src: PathBuf, dest: PathBuf) -> Self {
        Self { dest,
               src,
               default_interpreter: PathBuf::from(DEFAULT_INTERPRETER) }
    }

    fn get_source_from_toml(toml: &str) -> Option<String> {
        toml.parse()
            .ok()
            .as_ref()
            .and_then(toml::value::Value::as_table)
            .and_then(|toml_table| toml_table.get("source"))
            .and_then(toml::value::Value::as_str)
            .map(String::from)
    }

```

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+                                     bat"),
+                       src = self.src.display(),
+                       env = exports))
+        }
+
+        #[cfg(unix)]
+        {
+            // We need to prevent the binstub from using itself as an interpreter.
+            // On linux we use `/bin/sh` as the binstub interpreter and if this package
+            // includes a `sh` binary, we risk a circularly linked interpretter. We can
```
baumanj:
```suggestion
            // includes a `sh` binary, we risk a circularly linked interpreter. We can
```

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+            // detect if the link path is the same as the interpreter and if so, use the
+            // src binary as the interpreter.
+            let interpreter = {
+                if self.default_interpreter == self.dest {
```
baumanj: I wonder if we need to [`canonicalize`](http://doc.rust-lang.org/1.36.0/std/path/struct.Path.html#method.canonicalize) here. It's not uncommon for `/bin/sh` to be a link to some other interpreter. On my Ubuntu system: for instance:
```
➤ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Dec 13  2017 /bin/sh -> dash*
```

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
-                     src:  fs::read_link(&dest)?, })
+    pub fn from_file(path: &Path) -> Result<Self> {
+        use toml::Value::Table;
+
+        // its possible the link could already exist as a symlink either
+        // because a previous version of habitat did the binlinking or
+        // the symlink created outside of habitat
+        match fs::read_link(&path) {
+            Ok(lnk) => {
+                Ok(Binlink { dest:                path.to_path_buf(),
```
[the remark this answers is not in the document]
mwrock: `dest` is the destination of the binlink file. The `PathBuf` from `read_link` gets added to `src`. I can see now how these names can be confusing.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+        Ok(Self { dest:                Self::binstub_path(&src, dest_dir)?,
+                  src:                 src.to_path_buf(),
+                  default_interpreter: default_interpreter.to_path_buf(), })
     }
 
-    pub fn from_file<T: AsRef<Path>>(dest: T) -> Result<Self> {
-        Ok(Binlink { dest: dest.as_ref().to_path_buf(),
-                     src:  fs::read_link(&dest)?, })
+    pub fn from_file(path: &Path) -> Result<Self> {
+        use toml::Value::Table;
```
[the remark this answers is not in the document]
mwrock: oh good catch. Thats a remnant of when this was windows only behavior.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+    pub fn new(src: &Path, dest_dir: &Path, default_interpreter: &Path) -> Result<Self> {
+        Ok(Self { dest:                Self::binstub_path(&src, dest_dir)?,
+                  src:                 src.to_path_buf(),
+                  default_interpreter: default_interpreter.to_path_buf(), })
     }
 
-    pub fn from_file<T: AsRef<Path>>(dest: T) -> Result<Self> {
-        Ok(Binlink { dest: dest.as_ref().to_path_buf(),
-                     src:  fs::read_link(&dest)?, })
+    pub fn from_file(path: &Path) -> Result<Self> {
```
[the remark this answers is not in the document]
mwrock: unfortunately this gets mucked up with the confusion I refer to above. Because `src` throughout this code is referring to the actual `pkgs` binary and `dest` is the actual binlink like this `path` arg. Gonna think about how to make this better.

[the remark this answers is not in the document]
mwrock: oh ha! just now getting to your comment below where you are getting at the same point on these confusing field names.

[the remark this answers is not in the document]
mwrock: I think in the `Binlink` struct, I'm going to globally rename `dest` to `link_path` and `src` to `target_path`.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +141,159 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
-        fs::write(&self.dest, template)?;
-        Ok(())
+    fn env_var(name: &str) -> String {
+        #[cfg(windows)]
+        {
+            format!("%{}%", name)
+        }
+        #[cfg(unix)]
+        {
+            format!("${}", name)
```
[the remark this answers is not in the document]
mwrock: I think you actually meant `format!(r#""${}""#, name)` to result in `"$PATH"` instead of `$"PATH"`

mwrock:
@baumanj this should have all your comments applied. Thanks for the complete review!

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +145,165 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
     }
 }
 
+#[allow(dead_code)]
 struct Binlink {
-    dest: PathBuf,
-    src:  PathBuf,
+    link_path:           PathBuf,
+    target_path:         PathBuf,
+    default_interpreter: PathBuf,
```
baumanj: 💯 on the change from `dest`/`src`. I think this is *much* clearer. Do we need the `_path` suffix, though? The type already tells us we're dealing with a path, and it seems inconsistent with `default_interpreter`.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +145,165 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
-        let bin_name = match src.as_ref().file_stem() {
-            Some(name) => name,
-            None => return Err(Error::CannotParseBinlinkSource(src.as_ref().to_path_buf())),
-        };
-        let mut path = dest_dir.as_ref().join(bin_name);
-        path.set_extension("bat");
-
-        Ok(Binlink { dest: path,
-                     src:  src.as_ref().to_path_buf(), })
+    fn with_default_interpreter(target_path: PathBuf, link_path: PathBuf) -> Self {
```
baumanj: This function and `from_link` are identical in implementation. I'd suggest keeping this one and losing the other.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +145,165 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
-        use crate::hcore::os::filesystem;
+    fn from_link(target_path: PathBuf, link_path: PathBuf) -> Self {
+        Self { link_path,
+               target_path,
+               default_interpreter: PathBuf::from(DEFAULT_INTERPRETER) }
+    }
 
-        filesystem::symlink(&self.src, &self.dest)?;
-        Ok(())
+    fn try_from_script(path: &Path) -> Result<Self> {
```
baumanj: Maybe call this `script` or `binstub`? I think it makes things clearer later when there are multiple paths being juggled around.

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +145,165 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+        let file = File::open(path)?;
+        for line in BufReader::new(file).lines() {
+            let ln = line?;
+            if ln.to_uppercase().starts_with(COMMENT_MARKER) {
+                let (_, rest) = ln.split_at(COMMENT_MARKER.len());
+                if let Some(source) = Self::get_source_from_toml(rest) {
+                    return Ok(Self::with_default_interpreter(source.into(), path.into()));
```
baumanj: To be consistent with the clearer naming, perhaps `source` → `target`?

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +145,165 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
+                }
+            }
+        }
+        Err(Error::CannotParseBinlinkSource(path.to_path_buf()))
```
baumanj: Again for the sake of naming consistency, perhaps something other than `Source` here

components/hab/src/command/pkg/binlink.rs
```diff
@@ -135,89 +145,165 @@ pub fn binlink_all_in_pkg(ui: &mut UI,
...
     }
 
-    pub fn from_file<T: AsRef<Path>>(path: T) -> Result<Self> {
-        use toml::Value::Table;
+    fn get_source_from_toml(toml: &str) -> Option<String> {
+        toml.parse()
+            .ok()
+            .as_ref()
+            .and_then(toml::value::Value::as_table)
+            .and_then(|toml_table| toml_table.get("source"))
```
baumanj: If we really want to be consistent, we should even change this `source` (and in the stub templates)

PR #6702, opened 2019-07-01
fail build on docker push failures

mwrock:

### DIFF
--- a/.buildkite/release_pipeline.yaml
+++ b/.buildkite/release_pipeline.yaml
@@ -357,7 +357,7 @@ steps:
     env:
       BUILD_PKG_TARGET: "x86_64-linux-kernel2"
 
-  - label: ":docker: :windows: Upload windows container to Bintray"
+  - label: ":docker: :windows: Upload windows container to Docker Hub"
     command: .buildkite/scripts/publish_docker_studio.ps1
     agents:
       queue: 'single-use-windows-privileged'
--- a/.buildkite/scripts/publish_docker_studio.ps1
+++ b/.buildkite/scripts/publish_docker_studio.ps1
@@ -21,6 +21,9 @@ Write-Host "--- Publishing the Windows Docker Studio"
 . ./results/last_image.ps1
 
 docker login -u="$env:DOCKER_LOGIN_USER" -p="$env:DOCKER_LOGIN_PASSWORD"
+if ($LASTEXITCODE -ne 0) {
+  Write-Error "Failed to login to docker hub"
+}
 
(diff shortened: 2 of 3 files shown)


.buildkite/release_pipeline.yaml
```diff
@@ -357,7 +357,7 @@ steps:
     env:
       BUILD_PKG_TARGET: "x86_64-linux-kernel2"
 
-  - label: ":docker: :windows: Upload windows container to Bintray"
+  - label: ":docker: :windows: Upload windows container to Docker Hub"
```
christophermaier: eeeeexcellent

.buildkite/scripts/publish_docker_studio.ps1
```diff
@@ -21,6 +21,9 @@ Write-Host "--- Publishing the Windows Docker Studio"
 . ./results/last_image.ps1
```
christophermaier: I can't comment directly, but what if line 18 fails?

.buildkite/scripts/publish_docker_studio.ps1
```diff
@@ -29,14 +32,24 @@ if(is_fake_release) {
...
     Write-Host "Pushing ${docker_image}:$docker_base_image_tag-$docker_image_short_version tag for $docker_image_version"
     docker push "${docker_image}:$docker_base_image_tag-$docker_image_short_version"
+    if ($LASTEXITCODE -ne 0) {
+      Write-Error "Failed to tag image"
+    }
     if($docker_base_image_tag -eq "ltsc2016") {
       Write-Host "Pushing latest tag for $docker_image_version"
       docker push "${docker_image}:latest"
+      if ($LASTEXITCODE -ne 0) {
+        Write-Error "Failed to tag image"
```
christophermaier: Would it be worth differentiating this error message from the one on line 41, just to give us a little extra context if something fails?

Outputting which tagging failed would be nice.

mwrock: I think the `Write-Host` should give that context

.buildkite/scripts/publish_docker_studio.ps1
mwrock: we have a check on the docker build but although the `docker tag`s are pretty safe, we might as well guard them.

.buildkite/scripts/publish_docker_studio.ps1
christophermaier: Derp... so it does!

Ignore me :sweat_smile:

PR #6704, opened 2019-07-02
Kg/departialinator

kagarmoe:
This removes the singe-use partials from the documentation, flattening the docs structure. Cleans up Markdown and catches a few obvious errors.


www/source/docs/reference.html.md.erb
```diff
@@ -23,24 +24,1480 @@ This section will cover as many of the modules and APIs that Chef Habitat is bui
...
+
+
+**strConcat**
+: The `concat` helper can be used to connect multiple strings into one string without a separator. For example, `{{strConcat "foo" "bar" "baz"}}` would return `"foobarbaz"`.\
+
+You cannot concatenate an object (e.g. `{{strConcat web}}`), but you could concatenate the variables in an object (e.g. `{{strConcat web.list}}`).
+
 ---
-<%= partial "/partials/docs/reference-template-data"%>
+## <a name="template-data" id="template-data" data-magellan-target="template-data">Template Data</a>
```
kagarmoe: I deleted these lines from the partials file, because they don't work and appear in the text:

```
 ARGV tmp/render_context_schema.json	
<!-- This is a generated file, do not edit it directly. See https://github.com/habitat-sh/habitat/blob/master/www/scripts/generate-template-reference.js -->
```

https://www.habitat.sh/docs/reference/#concat-helper

baumanj:
With over 5,000 lines of changes, it would make it much easier to review if you directed attention towards what parts should be scrutinized and what parts can be generally ignored as mechanical changes.

Additionally, pushing the branch to acceptance to make sure the docs are correctly rendered in their target environment would be useful.

raskchanky:
I haven't done a thorough review of this, but I did do a cursory side-by-side click through of the production site and this branch running locally, and found a few issues.

There are a number of sections where the heading font size has changed.  Compare the following two screenshots taken from the Glossary page.

First, production:

Screen Shot 2019-07-03 at 11 30 00 AM

Next, local:

Screen Shot 2019-07-03 at 11 30 08 AM

Note how `Artifacts` in production is much bigger than this branch renders.  This is an issue on the following top level pages:
* Glossary
* Internals
* Developing Packages
* Managing Services
* Chef Habitat and other software
* Best Practices

There's also an issue with in correct text showing up on the Reference page.  Note the following screenshot:

Screen Shot 2019-07-03 at 11 33 34 AM

"Working with Builder" shouldn't show at the top of the Reference page.

kagarmoe:
@raskchanky Thanks for taking a look! The changes to the heading sizes is intentional We currently have multiple H1s on each page, which is really badly formed markdown. I'll check the "Working with builder" problem.

@baumanj Yes, it is an enormous PR. Could you take a look at what the `ARGV` was supposed to do? Is it something that we should recreate?

www/source/docs/reference.html.md.erb
baumanj: I'm not really sure what the intent was here. It looks like diagnostic logging that wasn't supposed to end up in the generated output. It was added in https://github.com/habitat-sh/habitat/pull/6427, which @raskchanky reviewed, so he may have more context.

raskchanky: The `ARGV` line appears to be a bug in how the CLI documentation script generates its output.  That's worth reviewing.  The HTML comment is generated automatically by the node.js script that generates the template reference docs.  This change to eliminate all the partials breaks that script, which overwrote a specific partial every time it ran.

Either the script will need to be adjusted to account for this new layout, which is likely a fairly complicated change, or we should make an exception in this case and keep this partial.

raskchanky:
I'm curious about the motivation behind this PR.  Clearly a significant amount of time and effort went into this work, and the last thing I want to do is trivialize that.  To my eyes though, at least at first glance and without the proper context, I'm not sure this is a win.

One example is `www/source/docs/reference.html.md.erb`.  Before, it was 46 lines, consisting of a Table of Contents and 11 partials.  Now it's 1500 lines of text.  Having the partials made it easier (again, from my perspective) to edit things in small chunks and easier for the scripts we have to auto-generate documentation and put it in the right place, without needing complicated find/replace logic.

I'm interested to hear more about the problems this PR solves.

kagarmoe:
@raskchanky Partials are a way of sharing content across pages to avoid duplication. In the case of the partials in the Habitat documentation, nearly all of the partials were single-use, and so by definition don't share content. However, they did add a level of complexity. The result is badly-formed markdown, which, while problematic both aesthetically and editorially, really confounds all reasonable efforts at search and discovery. 

Aesthetically, the use of partials inserted multiple H1s into the markdown. This is simply _wrong_. 
The generation of multiple H1s then causes a cascade of wrongness in each document.

Editorially, it is nearly impossible to gain a clear perspective of the whole of each document from its discrete, distributed, and disguised parts. Comprehending the entire document as a whole--not in its parts--is vital to understanding a user's experience in our documentation and, subsequently, to editing the document for clarity, correctness, and conciseness.

Document search cannot work in an environment that does not distinguish between a document's topic (H1) and its parts (H2 and below). This is, of course, directly related to the aesthetics of partials.

Finally, we will eventually convert the Habitat documentation into Hugo, and the presence of single-use partials will complicate and extend our work. We're a small team and this is a big project.

kagarmoe:
Regarding the idea that a document was previously was 46 lines, but now 1500 lines of text, I can only assert that it was always 1500 lines of text for our users. As one of the developers, you had the privilege of understanding it, experiencing it, and knowing how to find information in its parts.

Regarding auto-generation: I was not aware that the reference doc contained auto-generated parts. I find this quite thrilling. We can look at different organizational strategies to build on the auto-generation that's already in place. Hugo definitely has procedures that we can use, and I assume that Middleman has something similar.

kagarmoe:
I've thought more about our use of generated content in the reference page. Right now, only the section on [templates](https://www.habitat.sh/docs/reference/#template-data) is generated from code. Ideally, I would like to see all of the reference page content generated from code using [Middleman data files](https://middlemanapp.com/advanced/data-files/) for holding the content and page templates + CSS for presentation. The obvious benefit is that these sections would "refresh" themselves with each new build. This is clearly a larger project and would need to be part of the team plans. 

Without a greater automation plan, we have two paths:

1. Revert the reference page and the partials to their original condition. This will likely impede progress on improving searching for our users.
2. Keep the reference page layout that is in this PR. This will break the content generation for the templates section until we are able to take up the larger project of generating documentation from code.

raskchanky:
@kagarmoe 

> Partials are a way of sharing content across pages to avoid duplication. In the case of the partials in the Habitat documentation, nearly all of the partials were single-use, and so by definition did not share content. However, they did add a level of complexity. The result is badly-formed markdown, which, while problematic both aesthetically and editorially, really confounds all reasonable efforts at search and discovery.

Partials are just a content manipulation tool, and they have many uses, including, but not limited to:
* sharing content across pages to avoid duplication
* breaking up long content into sub-sections for ease of editing (as was done here)
* allowing the generation of content via an external tool for easy inclusion in the main site (as is done with our CLI docs)

They certainly add a level of complexity, as does any abstraction.  That's one of the tradeoffs in using them.  Partials themselves, however, have nothing to do with markdown.  What's contained inside a partial is blindly substituted into the spot where the partial is rendered.  I'm not saying that our site didn't contain badly-formed markdown, because I'm sure it did.  But partials aren't to blame for that problem.

> Aesthetically, the use of partials inserted multiple H1s into the markdown. This is simply wrong.
The generation of multiple H1s then causes a cascade of wrongness in each document.

Again, the use of partials doesn't automatically insert any kind of content.  The only thing that's inserted is what's contained inside the partial.  If the partial contains an H1, then an H1 will get inserted.  If it doesn't, it won't.  I'm confident we could solve all of the wrong H1 tags without touching partials at all.  We are using the [slim templating language](http://slim-lang.com/) for laying out our pages, and it is indentation sensitive.  It's entirely possible that errant indentation is to blame for these issues, although I haven't investigated.

> Editorially, it is nearly impossible to gain a clear perspective of the whole of each document from its discrete, distributed, and disguised parts. Comprehending the entire document as a whole is vital to understanding a user's experience in our documentation and, subsequently, to editing the document for clarity, correctness, and conciseness.

This makes a lot of sense, and as my contributions to the docs have been fairly minimal, I trust your better judgement on this.

> Document search cannot work in an environment that does not distinguish between a document's topic (H1) and its parts (H2 and below). This is, of course, directly related to the aesthetics of partials.

I'm not certain what `Document search` refers to in this context, but if it's referring to the ability of end-users to search our site via some kind of built-in search tool, partials are again irrelevant.  All of the partials get rendered into a single document that's deployed to the web server.  End users and any kind of search tools we use shouldn't know or care whether we're using partials or not.

> Finally, we will eventually convert the Habitat documentation into Hugo, and the presence of single-use partials will complicate and extend our work. We're a small team and this is a big project.

This is something I think we're all looking forward to.

> Regarding the idea that a document was previously was 46 lines, but now 1500 lines of text, I can only assert that it was always 1500 lines of text for our users. As one of the developers, you had the privilege of understanding it, experiencing it, and knowing how to find information in its parts.

Right, it was always the same to our users, but my comment was more geared towards the ease of editing and maintaining the docs themselves.  As I mentioned before, my work on the docs has not been extensive, and while I find it easier to maintain partials, I recognize that not everyone does.

> Regarding auto-generation: I was not aware that the reference doc contained auto-generated parts. I find this quite thrilling. We can look at different strategies to build on the auto-generation that's already in place. Hugo definitely has procedures that we can use, and I assume that Middleman has something similar.

Our auto-generation process is separate from Middleman.  It's a node.js script that runs through the `Makefile` and creates a partial that gets rendered when the site is built.

> I've thought more about our use of generated content in the reference page. Right now, only the section on templates is generated from code. Ideally, I would like to see all of the reference page content generated from code using Middleman data files  for holding the content and page templates + CSS for presentation. The  obvious benefit is that these sections would "refresh" themselves with  each new build. This is clearly a larger project and would need to be  part of the team plans.

> Without a greater automation plan, we have two paths:
> 1. Revert the reference page and the partials to their original  condition. This will likely impede progress on improving searching for  our users.
> 2. Keep the reference page layout that is in this PR. This will break  the content generation for the templates section until we are able to  take up the larger project of generating documentation from code.

I think adding more Middleman specific functionality at this time, when we've been trying to migrate away from Middleman to Hugo for at least a year now, is probably not the best use of time.  I don't see a downside to leaving our generated content as partials, generated from the node.js scripts that we already have.

Of the two options you presented, my preference would be option 1.  As I stated before, I don't think this will have any impact on search for our users.

www/source/docs/reference.html.md.erb
kagarmoe: This ARGV line exists in the live docs: 
Screen Shot 2019-08-15 at 4 05 24 PM

kagarmoe: The template-data partial is the only auto-generated part of the reference docs and it is infrequently updated.

eeyun:
I have some thoughts on this one before I give it a thumbs up. I see you all have covered some significant ground in the backscroll so I'll try not to tread where you've already been.

First, for the sake of future work, the project has a policy not to merge PRs of this size. This much churn is nearly impossible to fully review. Even in this case we have site changes mixed with content changes and 86 pages of change is just too much overhead to get an effective review. I won't gatekeep the merge on this as its clear an astronomical amount of work has gone in here but in the future lets keep changes more closely scoped than this.

Second in some of the backscroll there was some back and forth on the removal of the partials and in taking a peek through the PR it looks like some of that content has ended up in very large flat files now. I share the same initial concerns expressed by @raskchanky on this and I think based on reading his feedback I'm in agreement, however I want to add some context and a caveat to what he had to say. 

Close to 20 people have worked on writing, refactoring, and reorganizing these docs over the course of the project and much of those organizational structure decisions were made to optimize a couple of things that are worth mentioning:

 - The structure of the docs content was determined by previous technical writers, UX folks and product owners based on accessibility research in the realm in which our project exists.  That structure evolved over time after hitting a plethora of issues with multiple search providers until ultimately we landed on this most current iteration. That is to say, our docs have been fully rewritten multiple times in the last 3 years so there is somewhat of a precedent for this change!
- Secondly and possibly more importantly - one of the main things the team has had to optimize for with the site's code and content structure is community/developer workflow. Effectively the issue we've run into in the past is not having enough support from folks outside of the team with keeping the docs maintained. One of the examples of this is with the CLI doc generation that Josh mentioned above. The team optimized to make it as easy as possible for ourselves to manage the docs because for a very long time, the team were the only people actually managing the docs.

Now we've come a long way and I'm ready to give this a :+1: to merge so long as we're not moving backwards. What I mean by that is, if we remove auto-generation of CLI docs then someone has to maintain those and to a larger extent this applies to any QoL rollback we make to the docs. With the amount of work the team already has on their plates I feel concerned enough to say that if we actively roll back work that was performed to make project engineers/maintainers lives better we cannot obligate them to take on that extra burden of maintenance. Which is to say - this change is very large in scope and changes things about the way site content has been managed thus far. I'm totally good on all of that, so long as we have organizational support to keep them updated and managed without lots of future docs work by the engineers who have had to make it work so far!

kagarmoe:
I couldn't squash branch, so I went with `git reset --soft` instead. I then had to resolve a conflict with the `gemfile.lock` and I stayed with the structure on master. Once the current build is done, I'd like to see this in acceptance (I don't think I can push it myself). Thanks!

chefsalim:
setting do-not-merge flag until rebase issues are resolved

Issue #5957, opened 2018-12-10
Run post-run hook asynchronously


christophermaier:

davidMcneil:
Resolved by #6705

Issue #6701, opened 2019-07-01
Add delay before starting healthchecks


davidMcneil:
It would be a good idea to tweak the health check future to wait for some delay (say 5 seconds?) and _then_ launch into its check-wait-check cycle. This allows the service some warm up time before the first health check. You would get this same initial delay behaviour for restarts as well as for reconfigure (as well as plain old starts, of course). Reducing the amount of special cases seems like a good thing. We'd still want to ensure that the "handle" we keep to the overall health checking future still works, such that it would still kill the future if for some reason we needed to kill the service before that initial delay expired.

stevendanna:
@davidMcneil Just for certainty, the proposal here is still have an initial health check that happens sooner than the typical interval, but to move it to a slightly longer delay as opposed to happening on the first "server tick"?

davidMcneil:
Right. The thinking is to give the services some time to reach a "steady state" before they run their first health check. The delay amount is arbitrary but in practice, it might work well enough to prevent odd behavior for the first health check.

PR #6705, opened 2019-07-02
Make post-run async and add retry logic

### LINKED ISSUE #2364: post-run hook doesn't re-run on exit 1
Running hab from hab-0.22.1-20170509234454-x86_64-linux on CentOS 7.2, deploying into Docker.

It looks like when a unsuccessful code is encountered from the `post-run` hook script the script is not run again, like we see with `init`, `run` etc...

The use case for this would be:

1. Start the application in `run`
2. Perform some once off config which requires the application to be up in `post-run`

The two options I can see here are block in `run` until my service becomes available for config, or have the `post-run` script accept an `exit 1` as requiring another `post-run` pass.

If this is the intended functionality ideas are welcome!

davidMcneil:
Resolves #2364 #5957 

### DIFF
--- a/components/common/src/templating/hooks.rs
+++ b/components/common/src/templating/hooks.rs
@@ -43,7 +43,7 @@ pub fn stderr_log_path<T>(package_name: &str) -> PathBuf
     fs::svc_logs_path(package_name).join(format!("{}.stderr.log", T::file_name()))
 }
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub struct ExitCode(pub i32);
 
 impl Default for ExitCode {
@@ -290,7 +290,8 @@ pub trait Hook: fmt::Debug + Sized + Send {
                        status: ExitStatus)
                        -> Self::ExitValue;
 
-    fn retry(&self, _exit_value: Self::ExitValue) -> bool { false }
+    /// Return true if this hook should be retried provided the exit value of the previous run.
+    fn should_retry(_exit_value: &Self::ExitValue) -> bool { false }
 
     fn path(&self) -> &Path;
 
--- a/components/common/src/templating/hooks.rs
+++ b/components/common/src/templating/hooks.rs
@@ -290,6 +290,8 @@ pub trait Hook: fmt::Debug + Sized + Send {
                        status: ExitStatus)
                        -> Self::ExitValue;
 
+    fn retry(&self, _exit_value: Self::ExitValue) -> bool { false }
+
     fn path(&self) -> &Path;
 
     fn renderer(&self) -> &TemplateRenderer;
--- a/components/sup/src/manager/service.rs
+++ b/components/sup/src/manager/service.rs
@@ -405,10 +405,11 @@ impl Service {
         // See https://github.com/habitat-sh/habitat/issues/6739
     }
 
-    /// Called when stopping the Supervisor for an update. Should
-    /// *not* stop the service process itself, but should stop any
-    /// associated processes, futures, etc., that would otherwise
-    /// prevent the Supervisor from shutting itself down.
+    /// Called when stopping the Supervisor for an update and
+    /// before stopping a service. Should *not* stop the service
+    /// process itself, but should stop any associated processes,
+    /// futures, etc., that would otherwise prevent the Supervisor
+    /// from shutting itself down.
     ///
     /// Currently, this means stopping any associated long-running
     /// futures.
@@ -425,9 +426,7 @@ impl Service {
     pub fn stop(&mut self,
                 shutdown_config: ShutdownConfig)
                 -> impl Future<Item = (), Error = Error> {
-        // If the post run hook is still being retried, make sure we stop it.
-        self.stop_post_run();
-        self.stop_health_checks();
+        self.detach();
 
         let service_group = self.service_group.clone();
         let gs = Arc::clone(&self.gateway_state);
--- a/components/sup/src/manager/service.rs
+++ b/components/sup/src/manager/service.rs
@@ -18,6 +18,7 @@ mod supervisor;
 mod terminator;
 
 use self::{context::RenderContext,
+           hook_runner::HookRunner,
            hooks::{HookCompileTable,
                    HookTable},
            supervisor::Supervisor};
@@ -225,6 +226,7 @@ pub struct Service {
     /// health checks on this service. This is the means by which we
     /// can stop that future.
     health_check_handle: Option<sup_futures::FutureHandle>,
+    post_run_handle: Option<sup_futures::FutureHandle>,
 }
 
 impl Service {
@@ -276,6 +278,7 @@ impl Service {
                      health_check_interval: spec.health_check_interval,
                      gateway_state,
                      health_check_handle: None,
+                     post_run_handle: None,
                      shutdown_timeout: spec.shutdown_timeout })
     }
 
@@ -418,6 +421,7 @@ impl Service {
     pub fn stop(&mut self,
                 shutdown_config: ShutdownConfig)
                 -> impl Future<Item = (), Error = Error> {
+        self.stop_post_run();
         self.stop_health_checks();
 
         let service_group = self.service_group.clone();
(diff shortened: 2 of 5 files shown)


components/sup/src/manager/service.rs
```diff
@@ -810,25 +814,36 @@ impl Service {
...
+    ///
+    /// Consumes the handle to that future in the process.
+    fn stop_post_run(&mut self) {
+        if let Some(h) = self.post_run_handle.take() {
+            h.terminate();
         }
     }
 
     // This hook method looks different from all the others because
     // it's the only one that runs async right now.
```
christophermaier: We can probably remove this particular comment, since it's no longer the only async hook :joy:

components/sup/src/manager/service/hook_runner.rs
```diff
@@ -40,7 +57,28 @@ impl<H> HookRunner<H> where H: Hook + Sync
...
+
+    // TODO (DM): Make `HookRunner` and this method allow arbirary configuration of how to run a
+    // hook. For example, a configurable backoff, timeout, or failure tracking. It would also be
+    // nice to merge this with the healthcheck hook retry logic.
+    pub fn loop_future(&self) -> (FutureHandle, impl Future<Item = (), Error = ()>) {
+        let f = future::loop_fn(self.clone(), |hook_runner| {
+            hook_runner.clone()
+                       .into_future()
+                       .map(move |(exit_value, _duration)| {
+                           if hook_runner.hook.retry(exit_value) {
```
christophermaier: I like how you added a `retry` function to the `Hook` trait to allow for the (temporal) future where we can start to generalize this kind of hook behavior, rather than tightly-coupling this to the `post-run` hook specifically. We don't lose  anything, but we have a nice foundation from which to build.

christophermaier: It would be helpful to add some logging here (probably `debug!`-level) indicating that we're rerunning the offending hook.

components/sup/src/manager/service.rs
baumanj: This comment isn't true any longer, is it?

components/sup/src/manager/service.rs
```diff
@@ -810,25 +814,36 @@ impl Service {
...
+            let (handle, f) = hook_runner.loop_future();
+            self.post_run_handle = Some(handle);
+            executor.spawn(f);
+        }
+    }
+
+    /// Stop the `post-run` future. The `post-run` hook has retry logic based on its exit code. This
+    /// will stop this retry loop regardless of `post-run`'s exit code.
+    ///
+    /// Consumes the handle to that future in the process.
```
baumanj: Is there a way to indicate that in the name of the method rather than the comment? Or maybe a more interesting question is: does the caller of the method need to know this?

components/sup/src/manager/service.rs
```diff
@@ -810,25 +814,36 @@ impl Service {
...
+                                              self.pkg.clone(),
+                                              self.svc_encrypted_password.clone());
+            let (handle, f) = hook_runner.loop_future();
+            self.post_run_handle = Some(handle);
+            executor.spawn(f);
+        }
+    }
+
+    /// Stop the `post-run` future. The `post-run` hook has retry logic based on its exit code. This
+    /// will stop this retry loop regardless of `post-run`'s exit code.
```
baumanj: I'd hesitate to comment about the behavior of _other_ code in comments, as they're extra likely to become incorrect. See the comment on `post_stop` for instance. With the second sentence removed, I think this comment is great.

components/sup/src/manager/service/hook_runner.rs
```diff
@@ -40,7 +57,28 @@ impl<H> HookRunner<H> where H: Hook + Sync
                      pkg,
                      passwd }
     }
+
+    // TODO (DM): Make `HookRunner` and this method allow arbirary configuration of how to run a
+    // hook. For example, a configurable backoff, timeout, or failure tracking. It would also be
+    // nice to merge this with the healthcheck hook retry logic.
```
baumanj: Just a note to make sure this TODO is resolved before merge

baumanj: Personally, this sounds a bit like YAGNI. Do we have a use case that demands this today? If not, I'd leave it out.

components/common/src/templating/hooks.rs
```diff
@@ -285,6 +285,8 @@ pub trait Hook: fmt::Debug + Sized + Send {
                        status: ExitStatus)
                        -> Self::ExitValue;
 
+    fn retry(&self, _exit_value: Self::ExitValue) -> bool { false }
```
christophermaier: Always good to add some documentation comments :smile:

components/sup/src/manager/service/hooks.rs
```diff
@@ -213,6 +213,14 @@ impl Hook for PostRunHook {
         }
     }
 
+    fn retry(&self, exit_value: Self::ExitValue) -> bool {
+        match exit_value.0 {
+            0 => false,
+            1 => true,
+            _ => false,
+        }
```
baumanj: Why not
```rust
        exit_value.0 == 1
```
?

Even better, define a constant (at the `fn` or type level, either seems good) and you get:
```rust
        exit_value == SHOULD_RETRY
```

You'll have to derive `PartialEq` for `ExitCode`, but that seems worth it to me.

Also, unrelated, if feels like this method would more accurately be named `should_retry`.

components/sup/src/manager/service/hooks.rs
```diff
@@ -213,6 +213,14 @@ impl Hook for PostRunHook {
         }
     }
 
+    fn retry(&self, exit_value: Self::ExitValue) -> bool {
+        match exit_value.0 {
+            0 => false,
+            1 => true,
+            _ => false,
```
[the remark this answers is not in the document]
davidMcneil: I am not sure what should be done here from a product perspective. I can think of three potential, reasonable options:

Option 1
- 0 no retry
- else retry

Option 2
- 0 or 1 no retry
- else retry

Option 3
- 1 retry
- else no retry

There are obviously infinite other possibilities. Is there currently a way for the user to access the exit code of the `post-run` hook and use it? If not, and the exit code is solely used to determine retry behaviour, I now think I would change this to the simpler Option 1. Are there preferences for a particular set of logic here?

components/sup/src/manager/service.rs
```diff
@@ -931,7 +946,7 @@ impl Service {
             self.initialize();
             if self.initialized {
                 self.start(launcher, executor);
-                self.post_run();
+                self.post_run(executor);
```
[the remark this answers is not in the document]
davidMcneil: :thinking: I do not think I fully understand the purpose of `busy_services`. It makes sense that stopping a service would require special logic. We want to make sure we completely stop before starting again. It seems like the purpose of `busy_services` is to avoid race conditions, and I am having trouble coming up with race conditions with `post-run`.  Not saying they don't exist. I just cannot come up with a concrete example.

I completely agree that we need a top-level `Service` future. Right now we are executing futures at two levels: in the individual services and in the manager. This causes quite a bit of pain. For example, it would be nice to use the `wrap_async_service_operation` for the `post-run` hook to add it to the `busy_services`, but that is defined at the manager level whereas we run `post-run` at the service level. Another example is the need for the `needs_restart` flag.

components/sup/src/manager/service/hook_runner.rs
davidMcneil: I agree. I will remove this comment. I wanted a way to signify to reviewers that this is where we could add this functionality and get their feedback. In that way the comment has served its purpose :smile: . Maybe a better method would have been to just leave a comment on the PR?

I do think it would be useful to merge this logic with the health check retry logic, and there is an [issue](https://github.com/habitat-sh/habitat/issues/6701) about adding an initial delay before starting a service and many of those other features are currently TODOs in other parts of the code. Ultimatly, I think we will end up implementing some of this logic, but it is outside the scope of this PR.

components/sup/src/manager/service.rs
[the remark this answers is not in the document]
christophermaier: I think you're probably correct that `post-run` doesn't require extra bookkeeping... I thought about it some more, and I wasn't able to come up with a concrete example myself.

components/sup/src/manager/service.rs
```diff
@@ -363,6 +366,7 @@ impl Service {
     pub fn stop(&mut self,
                 shutdown_config: ShutdownConfig)
                 -> impl Future<Item = (), Error = Error> {
+        self.stop_post_run();
```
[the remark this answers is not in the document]
baumanj: The name `stop_post_run` makes it clear _what_ is happening, so I think the comment here should be focused on _why_ that's necessary.

components/sup/src/manager/service/hook_runner.rs
```diff
@@ -40,7 +57,26 @@ impl<H> HookRunner<H> where H: Hook + Sync
                      pkg,
                      passwd }
     }
+
+    pub fn retriable_future(&self) -> (FutureHandle, impl Future<Item = (), Error = ()>) {
```
baumanj: "retriable" is a totally valid spelling, though it might be worth considering using "retryable", so it turns up in a search for "retry"

components/sup/src/manager/service/hook_runner.rs
```diff
@@ -40,7 +57,26 @@ impl<H> HookRunner<H> where H: Hook + Sync
...
+        let f = future::loop_fn(self.clone(), |hook_runner| {
+            hook_runner.clone()
+                       .into_future()
+                       .map(move |(exit_value, _duration)| {
+                           if <H as Hook>::should_retry(&exit_value) {
```
baumanj: Why not just
```suggestion
                           if H::should_retry(&exit_value) {
```
? It seems to work. Similarly with `file_name`. If you think `H` is unclear, it's fine to be explicit and call it `HOOK`. There's no need for template variables to be limited to single characters.

baumanj:
> Alternatively, we could just not restart post-run hooks when the Supervisor reattaches to a service.

I'm generally in favor of doing the simplest thing/YAGNI. The only thing to consider is whether it would be hard to change later. What's the impact on folks who get accustomed to it not restarting `post-stop` after reattach? My guess is nothing, but that's the thing I would think more about.

davidMcneil:
> Alternatively, we could just not restart post-run hooks when the Supervisor reattaches to a service.
This is a tricky problem. [This](https://github.com/habitat-sh/habitat/pull/6705/commits/82838cc58d9fb9883a2afafb683d81a922d453b5) should add the logic

components/sup/src/manager/service.rs
```diff
@@ -399,26 +400,33 @@ impl Service {
...
+        self.stop_health_checks();
+    }
 
     /// Return a future that will shut down a service, performing any
     /// necessary cleanup, and run its post-stop hook, if any.
     pub fn stop(&mut self,
                 shutdown_config: ShutdownConfig)
                 -> impl Future<Item = (), Error = Error> {
-        self.stop_health_checks();
+        self.detach();
```
christophermaier: Nice.

components/sup/src/manager/service.rs
```diff
@@ -932,19 +939,23 @@ impl Service {
                      executor: &TaskExecutor,
                      template_update: &TemplateUpdate)
                      -> bool {
+        let up = self.check_process();
         if !self.initialized {
-            if self.check_process() {
+            // If the service is not initialized and the process is still running, the Supervisor
+            // was restarted and we just have to reattach to the process.
+            if up {
```
christophermaier: This is a welcome refactoring :+1:

PR #6714, opened 2019-07-09
Improve liveliness checker

baumanj:
Add mechanisms for indicating a thread has ended, either with an error (will be logged for a period of time, alongside potentially deadlocked threads) or expectedly. Also, add special `must_use` types to help ensure correct usage. See individual commits for details.

Also, add liveliness checking to the launcher.


baumanj:
> I think 1 week might be a tad excessive

I'm open to other suggestions. I only added the culling to prevent runaway memory consumption, but if that's happening, it implies the same threads are ending with errors over and over, so maybe a shorter time would be appropriate.

However, if there was a rare occurrence of an error, it would be a shame to miss it because it got culled before anyone looked. We could make the culling logic only trigger if there are a large number of items in the `ThreadStatusMap`, but I'm hesitant to make the logic more complex before we see a problem in practice.

christophermaier:
I like the idea of culling, and I definitely wouldn't want to make the logic any more complex. I was thinking maybe a few days would be a good threshold, but it's really no big deal. It _is_ configurable, after all.

components/sup/src/manager.rs
```diff
@@ -855,7 +855,7 @@ impl Manager {
         // TODO (CM): Investigate the appropriateness of capturing any
         // errors or panics generated in this loop and performing some
         // kind of controlled shutdown.
-        let shutdown_mode = loop {
+        let shutdown_mode: habitat_common::sync::ThreadReturn<ShutdownMode> = loop {
```
christophermaier: I think we might not want type the loop like this, given that we still do some work after the loop finishes. That's actually where we were hanging with the regression fixed by #6717.

If this code was in place earlier, it would look like the main loop hadn't deadlocked at all.

christophermaier: In general though, the explicit approach of typing the thread exits looks really slick! :+1:

baumanj: > I think we might not want type the loop like this, given that we still do some work after the loop finishes.

Sure, that's no problem. Though if we are calling `mark_thread_alive` only _in_ the loop, we want to make sure all the paths that exit from the loop opt out of liveliness checking, right?

> If this code was in place earlier, it would look like the main loop hadn't deadlocked at all.

I'm not sure I understand. Can you elaborate? I'm guessing the place where we were stuck prior to https://github.com/habitat-sh/habitat/pull/6717 is https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager.rs#L1086, yes? If that's the case, getting a deadlock warning from this thread was more like a happy accident. For correctly identifying future-based deadlock, I think we should be replacing the `wait` there with a `poll` that calls `mark_thread_alive`, no? Or maybe, I'm mistaken because my `Futures` knowledge is pretty minimal as yet.

components/common/src/liveliness_checker.rs
```diff
@@ -0,0 +1,467 @@
...
+/// particular one being overlooked.
+///
+/// Additionally, since code paths which the threads pass through while unregistering may wish
+/// to communicate a Result back to the caller, this type wraps an arbitrary `Result` which can
+/// be then recovered via `into_result`, or if an arbitrary value is desired, via `into_ok`.
+#[must_use]
+pub struct ThreadUnregistered<T = (), E = std::convert::Infallible>(Result<T, E>);
+
+impl<T, E> ThreadUnregistered<T, E> {
+    pub fn into_result(self) -> Result<T, E> { self.0 }
```
baumanj: In case you're wondering why we can't use `Into`/`From` traits: https://users.rust-lang.org/t/generic-conversion-from-newtypes/16247

components/sup/src/manager.rs
christophermaier: I think I'm OK with not having anything in here do any unregistering of threads from the deadlock detector. Once we get out of the loop, we should be shutting down in a timely fashion anyway. I don't think we want any kind of polling and periodic marking of the main thread as still being alive once we get down to the point where we're draining the reactor of futures. 

#6717 fixes things in the near term, but I'll think a bit more about ways that we can more completely ensure that our "long-running" futures are terminated before we get to the "drain the reactor" part.

components/butterfly/src/server/push.rs
```diff
@@ -38,7 +37,7 @@ lazy_static! {
 
 pub fn spawn_thread(name: String, server: Server, timing: Timing) -> std::io::Result<()> {
     thread::Builder::new().name(name)
-                          .spawn(move || run_loop(&server, &timing))
+                          .spawn(move || -> ! { run_loop(&server, &timing) })
```
christophermaier: I notice that removing the `-> !` doesn't seem to have an effect, as long as `run_loop` is defined to return either nothing, or `!`.

As soon as I modified `run_loop` to return something, though, this refuses to compile. Neat!

components/common/src/liveliness_checker.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// ```
+    /// use habitat_common::liveliness_checker;
+    ///
+    /// fn run_loop() -> ! {
+    ///     loop {
+    ///         liveliness_checker::mark_thread_alive().and_divergent();
+    ///     }
+    /// }
+    /// ```
+    pub fn and_divergent(self) {}
```
christophermaier: This is a nice touch.

components/common/src/liveliness_checker.rs
```diff
@@ -0,0 +1,467 @@
...
+                              let max_time_since_death =
+                                  ThreadDeadIgnoreDelay::configured_value().into();
+                              loop {
+                                  let statuses =
+                                      &mut THREAD_STATUSES.lock()
+                                                          .expect("THREAD_STATUSES poisoned");
+                                  check_thread_heartbeats(statuses, threshold);
+                                  log_dead_threads(statuses);
+                                  cull_dead_threads(statuses, max_time_since_death);
+                                  thread::sleep(delay);
```
christophermaier: Wait, how will we know _this_ thread is still alive?

:trollface: _Quis custodiet ipsos custodes?_ :trollface:

components/common/src/liveliness_checker.rs
```diff
@@ -0,0 +1,467 @@
...
+
+/// Call periodically from a thread which has a work loop to indicate that the thread is
+/// still alive and processing its loop. If this function is not called for more than
+/// `ThreadAliveThreshold`, it will be error logged as a likely deadlock.
+///
+/// If the thread exits the loop where mark_thread_alive will be called, it must be unregistered
+/// from the liveliness checker or else be subject to false positives. In general, the idiom is
+/// ```
+/// use habitat_common::liveliness_checker;
+/// # let error = false;
```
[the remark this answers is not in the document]
baumanj: Yes, without that line, `cargo test --doc -p habitat_common` would fail. 

The `#` is to hide it from the documentation display. See, for example: https://github.com/rust-lang/rust/blob/master/src/liballoc/string.rs#L953 vs http://doc.rust-lang.org/1.36.0/alloc/string/struct.String.html#method.try_reserve.

components/common/src/liveliness_checker.rs
baumanj: I nearly added it, before considering how pointless that would be

components/sup/src/manager.rs
baumanj: I'm torn on this one. In the current state of the world, you're absolutely right. What I worry about is down the line and how changes to the code are likely to cause either false positives or false negatives.

We can remove the unregister calls here, but *only* because it's the main thread for the entire supervisor and *only* because once it leaves this particular loop, it should shut down in a timely fashion. That seems like a lot of assumptions that are fairly brittle. I could certainly imagine this loop getting moved to a separate thread in the future or an additional loop coming after which represented a different state of operation, but which isn't expected to end in a short period of time.

As we use futures more and more, I think the value of having the liveliness checker work with them becomes more important. Right now, removing the unregister calls upon exiting the main loop here would give us "deadlock likely" messages for the `main` loop, but it wouldn't tell us where it was stuck. My first thought would be false positive, and that wold be reinforced by the lack of unregister calls. Understanding this *isn't* the case requires realizing that `main` is a special case. That may be clear in retrospect, but if we can avoid making special cases, I think that'll reduce the cognitive load at a time we may be already tired and/or stressed (i.e., troubleshooting erroneous behavior in production).

The two improvements I think would be of significant value are making the liveliness checker work for tracking futures progress and adding an indicator of where in the code `mark_thread_alive` is called (more than just the thread name). If we had these capabilities, it would've allowed us to zero in on where we were stuck directly. I anticipate this won't be the last time futures code is associated with getting unexpectedly stuck.

For the time being, let's go with removing the unregister calls and I can file issues for the other enhancements if you agree about where we want to go with this and we can prioritize getting to a place that would truly expedite the discovery of issues like this in the future.

christophermaier: I agree with all this. In order to really work with futures, we're going to have to make some additional changes to the tracker anyway, since they get passed around on a threadpool. (i.e., just the thread name won't be useful; otherwise we'll see stuff like "'tokio-threadpool-3' hasn't checked in in X seconds; deadlock likely"). 

It would be interesting to get some kind of introspection on the threadpool, though. If a future ends up blocking because it's somehow deadlocked on one of these read-write locks, then that would block up a thread. Even if it wasn't deadlocked, the fact that a future would jam up a threadpool thread for an appreciable amount of time is itself a sign of a poorly-structured future, and that would be a good thing to know anyway. Perhaps that's a good avenue of exploration, if there's not already some off-the-shelf crates that address that.

While it's not going to be 100% applicable to this exact use case, I'm very keen on figuring out how to incorporate [tokio-trace](https://docs.rs/tokio-trace/0.1.0/tokio_trace/index.html) into our codebase. I suspect that will be useful in helping us get a clearer picture of everything that's going on internally.

components/common/src/liveliness_checker.rs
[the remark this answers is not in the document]
christophermaier: Ah, gotcha... :+1:

components/sup/src/manager.rs
baumanj: Filed https://github.com/habitat-sh/habitat/issues/6723

PR #6719, opened 2019-07-10
Add unrunnable services

### LINKED ISSUE #4494: Updating a running service does not obey new pkg_binds
When habitat auto-updates a package, it ignores newly required bindings. Once the package
has been updated, it will restart the service. If that service is unloaded, or the supervisor
is restarted, the service will fail to come up because of missing binds. In this case, `hab svc status` says no services are loaded, however trying to load the service again fails until it is unloaded.

This was done with the at-once update strategy.

davidMcneil:
Resolves #4494

This does not fix the core issue of #4494. That was already fixed by previous refactoring. Instead, this PR fixes `hab status` not reporting all loaded services. This problem is alluded to in #4494. The problem is explained concretely below.  

## The Problem
If I have a package `dem/test` that has one bind `database`, after running `hab svc load dem/test` I get the following messages
- the supervisor says `Unable to start dem/test, Missing required bind(s), database`
- `hab svc status` says `No services loaded.`
- `hab svc load dem/test --bind database:redis.default` fails with `Service already loaded, unload 'dem/test' and try again`

After explicitly unloading `dem/test` or loading with `--force` the service successfully starts. However, it seems incorrect that I get contradictory messages: one stating there are no services loaded and another that `dem/test` is already loaded.

## Solution
Previously, I was under the impression a service could be in one of three states: `LoadedUp`, `LoadedDown`, and `Unloaded`. This PR acknowledges that there is really a forth state `LoadedUnrunnable`. This state occurs when a service's spec file cannot be loaded. One way to get into this state is unsatisfied binds like in the example above. The majority of this PR is simply tracking what services are in the `LoadedUnrunnable` state and adding the plumbing necessary to get this information to those who care about it. Previously `hab status` provided no information on services in the `LoadedAndUnrunnable` state.

I am not overly enthusiastic about this solution. I would have preferred to simplify the problem and completely disallow the `LoadedUnrunnable` state. At first, this seems straightforward. We simply fail the load if binds are not met. Unfortunately, we can also get in this state following a package update if the new version added binds. One solution would be to automatically unload a service that entered this state, but I thought from a UX perspective that was too surprising. I would love to hear other solutions and thoughts.

## New Behavior
For the example above, `hab status` would now output the following:
```
package                           type        desired  state   elapsed (s)  pid     group           error
core/etcd/v3.3.10/20190305234020  standalone  up       up      52           24866   etcd.default    <none>
core/consul/1.5.2/20190701155502  standalone  up       up      52           24892   consul.default  <none>
dem/test                          standalone  <none>   <none>  <none>       <none>  <none>          Missing required bind(s), database
```
This addes a new `error` column that allows services in the `LoadedUnrunnable` state a place to report why they are unrunnable

## Questions for Reviewers
- This [comment](https://github.com/habitat-sh/habitat/blob/b57cb638461c9db4511667c277874f61a28b9412/components/hab/src/main.rs#L1597) implies that the output of `hab status` needs to be machine-readable. Is this the case? Do special considerations need to be made for the addition of the `error` column?
- I needed to add [this line](https://github.com/habitat-sh/habitat/blob/b57cb638461c9db4511667c277874f61a28b9412/components/sup-protocol/build.rs#L39) to make the build correctly output protocol buffer types. Could someone verify that this is in fact necessary? This should probably have been in a different commit?
- I added a `/services/unrunnable` http endpoint [here](https://github.com/habitat-sh/habitat/blob/b57cb638461c9db4511667c277874f61a28b9412/components/sup/src/http_gateway.rs#L283). I have no specific use for this so maybe it should be removed, but I was mirroring what the `/services` endpoint does. 

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -101,7 +101,8 @@ lazy_static! {
              "state",
              "elapsed (s)",
              "pid",
-             "group",]
+             "group",
+             "error"]
     };
 }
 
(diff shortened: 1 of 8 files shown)


mwrock:
Yeah I was thinking it would be better for packages in this scenario to end up in the `unloaded` state instead of unrunnable but your comment about upgrade certainly complicates that approach. Out of curiosity, what if you do add the `--bind` argument and that bind is not available and the service ends up in the waiting state. Is that service `down` or in this same limbo state?

davidMcneil:
If a service has all of its binds specified and it is simply waiting on them to be present in the census data, the service is in the `LoadedDown` state.

davidMcneil:
That is a good thought @mwrock. Instead of entering the `LoadedUnrunnable` state the service could enter the `LoadedDown` state and the supervisor could output a message such as `Missing binding 'database'. Please reload the service with the binding specified.`

The original issue is that we were swallowing errors when trying to [load a services](https://github.com/habitat-sh/habitat/blob/128f5429d54031f563f0c57fd076930836a9cede/components/sup/src/manager.rs#L1276). This change would make missing binds no longer an error condition, and make an error automatically unload the service (if we don't do this services can still be in the `LoadedUnrunnable` state just not for unsatisfied binds). An example error condition is a corrupted spec file. Does it seem reasonable to automatically unload a service in this case?

I think it would be useful if `hab status` could report why a service is in the `LoadedDown` state if the desired state is up. Currently, that information is hidden in the supervisor logs. Would this be a feature worth adding?

baumanj:
> * This [comment](https://github.com/habitat-sh/habitat/blob/b57cb638461c9db4511667c277874f61a28b9412/components/hab/src/main.rs#L1597) implies that the output of `hab status` needs to be machine-readable. Is this the case? Do special considerations need to be made for the addition of the `error` column?

I read that comment as "don't remove the `standalone` column" specifically. I don't think there should be an issue with adding a new column as long as we leave the rest alone. Anyone who's parsing this output in scripts (a bad idea, but we haven't provided a better alternative) is unlikely to be perturbed by the addition of another column at the end.

That said, I'd love if we replaced that TODO with a linked issue to provide a machine-friendly output version for `hab sup status`.

> * I added a `/services/unrunnable` http endpoint [here](https://github.com/habitat-sh/habitat/blob/b57cb638461c9db4511667c277874f61a28b9412/components/sup/src/http_gateway.rs#L283). I have no specific use for this so maybe it should be removed, but I was mirroring what the `/services` endpoint does.

I would err on the side of not adding something until we're sure we need it, because once something like this is added, it becomes much harder to remove later.

Also, I think we're trending toward the view that we may be able to use `LoadedDown` instead of `LoadedUnrunnable`, which I think is a good call.

components/sup-protocol/build.rs
```diff
@@ -22,6 +22,8 @@ use prost_types::{DescriptorProto,
                   FileDescriptorProto,
                   FileDescriptorSet};
 
+const GENERATED_OUT_DIR: &'static str = "src/generated";
```
baumanj:
```suggestion
const GENERATED_OUT_DIR: &str = "src/generated";
```
`const` lifetimes can be inferred now

components/sup/src/manager.rs
```diff
@@ -206,6 +206,23 @@ impl ShutdownConfig {
...
+    pub spec_ident: PackageIdent,
+    pub error:      String,
+}
+
+impl From<UnrunnableService> for habitat_sup_protocol::types::UnrunnableServiceStatus {
+    fn from(other: UnrunnableService) -> Self {
+        let mut proto = habitat_sup_protocol::types::UnrunnableServiceStatus::default();
+        proto.ident = other.spec_ident.into();
+        proto.error = other.error;
+        proto
```
baumanj: I'm not sure why we tend to use the paradigm with our proto conversions, but we're free to use the shorter and more rustic approach:
```rust
        Self { ident: other.spec_ident.into(),
               error: other.error, }
```

components/hab/src/main.rs
```diff
@@ -1563,60 +1564,73 @@ fn print_svc_status<T>(out: &mut T,
...
     where T: io::Write
 {
-    let status = match reply.message_id() {
+    match reply.message_id() {
         "ServiceStatus" => {
-            reply.parse::<sup_proto::types::ServiceStatus>()
-                 .map_err(SrvClientError::Decode)?
+            let status = reply.parse::<sup_proto::types::ServiceStatus>()
+                              .map_err(SrvClientError::Decode)?;
+            let svc_desired_state = status.desired_state
```
baumanj: Since this is a big chunk of code that's mostly moved, but also has some other changes, I find it makes it easier to review if broken into separate commits:
1. Move existing code to new location with no logic changes
1. Apply logic changes

davidMcneil:
Closing in favor of #6730.

PR #6729, opened 2019-07-13
Habitat API docs

kagarmoe:
Working Habitat API docs using Redoc

## How to use

1. Checkout this branch
2. `cd www`
3. `bundle exec middleman serve`
4. go to `http://localhost:4567/docs/api/builder-api`
OR
    go to `http://localhost:4567/docs/api/supervisor`

Screen Shot 2020-02-14 at 1 36 32 PM


chefsalim:
Closing for now so we don't have this in our review backlog - pls reopen when ready for a merge review

mwrock:
These pages render but how would a user navigate to them? I dont see entries for them in the left nav. Also, I'm curious how the json is populated? It seems like we should have scripts to do that in the repo.

www/source/layouts/api_builder.slim
```diff
@@ -0,0 +1,4 @@
+<a class="return" style="background-color:#007FAB; display:block; text-align:center; width:228px; margin:1rem 0 1.5rem 16px; padding:8px 0; color:white; text-decoration:none ;font-size:14px; text-transform:uppercase" href='https://www.habitat.sh/docs/'>Return to Docs</a>
```
mwrock: can we make this just `/docs/` so that it remains in whatever environment you are viewing the api?

www/source/layouts/api_sup.slim
```diff
@@ -0,0 +1,3 @@
+<a class="return" style="background-color:#007FAB; display:block; text-align:center; width:228px; margin:1rem 0 1.5rem 16px; padding:8px 0; color:white; text-decoration:none; font-size:14px; text-transform:uppercase" href='https://www.habitat.sh/docs/'>Return to Docs</a>
```
mwrock: same comment as above

mwrock:
I think we want a shell script that automates the json generation otherwise I'd be concerned the docs would easily go out of date.

kagarmoe:
I'll be honest @mwrock , I didn't think that a relative path would work, but it does.

We're on the same page on automating the JSON generation. Let's do that next.

www/source/layouts/api_builder.slim
kagarmoe: done

www/source/layouts/api_sup.slim
kagarmoe: done

PR #6730, opened 2019-07-15
Do not allow unrunnable services

### LINKED ISSUE #4494: Updating a running service does not obey new pkg_binds

davidMcneil:
Resolves #4494

For background please see my [first attempt](https://github.com/habitat-sh/habitat/pull/6719). Overall, I believe this is a much better solution. However, it does change existing behavior as described below.

The change consists of two steps that can be followed in the commit history.

1. Disallow the `LoadedAndUnrunnable` service state. This state would occur when [loading a service](https://github.com/habitat-sh/habitat/blob/128f5429d54031f563f0c57fd076930836a9cede/components/sup/src/manager.rs#L1276) would fail, but we would swallow the error and keep the package loaded. Now if the load fails the service is automatically unloaded. The two primary conditions in which a load can fail are 1. if the spec file cannot be found and 2. if the service has unsatisfied binds.
2. Do not allow loading a service with unsatisfied binds. As a result of 1, if a service was loaded with unsatisfied binds it would be immediately unloaded, but the cli would still show the message `service was successfully loaded`. This seemed like very bad UX. Now the cli gives an error when trying to load a service with unsatisfied binds.

Note, a service can still be automatically unloaded for unsatisfied binds if a package update adds binds. Does this seem like reasonable behavior?

### DIFF
--- a/components/sup/src/manager.rs
+++ b/components/sup/src/manager.rs
@@ -765,7 +765,7 @@ impl Manager {
                   self.butterfly.gossip_addr());
         self.butterfly.start_mlw(&Timing::default())?;
         debug!("gossip-listener started");
-        self.persist_state_mlr();
+        self.persist_state_mlr(&mut runtime);
         let http_listen_addr = self.sys.http_listen();
         let ctl_listen_addr = self.sys.ctl_listen();
         let ctl_secret_key = ctl_gateway::readgen_secret_key(&self.fs_cfg.sup_root)?;
@@ -948,27 +948,7 @@ impl Manager {
                     }
                     SupervisorAction::UnloadService { service_spec,
                                                       shutdown_input, } => {
-                        let file = self.state.cfg.spec_path_for(&service_spec);
-                        if let Err(err) = fs::remove_file(&file) {
-                            warn!("Tried to unload '{}', but couldn't remove the file '{}': {:?}",
-                                  service_spec.ident,
-                                  file.display(),
-                                  err);
-                        };
-                        if let Some(future) =
-                            self.remove_service_from_state(&service_spec)
-                                .map(|service| {
-                                    let shutdown_config =
-                                        ShutdownConfig::new(&shutdown_input, &service);
-                                    self.stop_with_config(service, shutdown_config)
-                                })
-                        {
-                            runtime.spawn(future);
-                        } else {
-                            warn!("Tried to unload '{}', but couldn't find it in our list of \
-                                   running services!",
-                                  service_spec.ident);
-                        }
+                        self.unload(&service_spec, &shutdown_input, &mut runtime);
                     }
                 }
             }
@@ -1016,11 +996,11 @@ impl Manager {
                                         &self.butterfly.service_file_store);
 
             if self.check_for_changed_services() {
-                self.persist_state_mlr();
+                self.persist_state_mlr(&mut runtime);
             }
 
             if self.census_ring.changed() {
-                self.persist_state_mlr();
+                self.persist_state_mlr(&mut runtime);
             }
 
             for service in self.state
@@ -1235,13 +1215,13 @@ impl Manager {
     /// # Locking
     /// * `MemberList::entries` (read) This method must not be called while any MemberList::entries
     ///   lock is held.
-    fn persist_state_mlr(&self) {
+    fn persist_state_mlr(&mut self, runtime: &mut Runtime) {
         debug!("Updating census state");
         self.persist_census_state();
         debug!("Updating butterfly state");
         self.persist_butterfly_state_mlr();
         debug!("Updating services state");
-        self.persist_services_state();
+        self.persist_services_state(runtime);
     }
 
     fn persist_census_state(&self) {
@@ -1267,20 +1247,20 @@ impl Manager {
             .butterfly_data = json;
     }
 
-    fn persist_services_state(&self) {
+    fn persist_services_state(&mut self, runtime: &mut Runtime) {
         let config_rendering = if self.feature_flags.contains(FeatureFlag::REDACT_HTTP) {
             ConfigRendering::Redacted
         } else {
             ConfigRendering::Full
         };
 
-        let services = self.state
-                           .services
-                           .read()
-                           .expect("Services lock is poisoned!");
-        let existing_idents: Vec<PackageIdent> =
-            services.values().map(|s| s.spec_ident.clone()).collect();
-
+        let existing_idents: Vec<PackageIdent> = self.state
+                                                     .services
+                                                     .read()
+                                                     .expect("Services lock is poisoned!")
+                                                     .values()
+                                                     .map(|s| s.spec_ident.clone())
+                                                     .collect();
         // Services that are not active but are being watched for changes
         // These would include stopped persistent services or other
         // persistent services that failed to load
@@ -1288,15 +1268,34 @@ impl Manager {
             self.spec_dir
                 .specs()
                 .iter()
-                .filter(|spec| !existing_idents.contains(&spec.ident))
-                .flat_map(|spec| {
-                    Service::load(self.sys.clone(),
-                                  spec.clone(),
-                                  self.fs_cfg.clone(),
-                                  self.organization.as_ref().map(|org| &**org),
-                                  self.state.gateway_state.clone()).into_iter()
+                .filter_map(|spec| {
+                    if existing_idents.contains(&spec.ident) {
+                        None
+                    } else {
+                        match Service::load(self.sys.clone(),
+                                            spec.clone(),
+                                            self.fs_cfg.clone(),
+                                            self.organization.as_ref().map(|org| &**org),
+                                            self.state.gateway_state.clone())
+                        {
+                            Ok(s) => Some(s),
+                            Err(e) => {
+                                outputln!("Failed to load spec file for '{}' with error '{}'. \
+                                           This service will be unloaded.",
+                                          spec.ident,
+                                          e);
+                                self.unload(spec, &ShutdownInput::default(), runtime);
+                                None
+                            }
+                        }
+                    }
                 })
                 .collect();
+        let services = self.state
+                           .services
+                           .read()
+                           .expect("Services lock is poisoned!");
+
         let watched_service_proxies: Vec<ServiceProxy<'_>> =
             watched_services.iter()
                             .map(|s| ServiceProxy::new(s, config_rendering))
@@ -1305,7 +1304,6 @@ impl Manager {
             services.values()
                     .map(|s| ServiceProxy::new(s, config_rendering))
                     .collect();
-
         services_to_render.extend(watched_service_proxies);
 
         let json = serde_json::to_string(&services_to_render).unwrap();
@@ -1319,6 +1317,31 @@ impl Manager {
     /// Check if any elections need restarting.
     fn restart_elections_mlr(&mut self, feature_flags: FeatureFlag) {
         self.butterfly.restart_elections_mlr(feature_flags);
+    }
+
+    fn unload(&mut self,
+              service_spec: &ServiceSpec,
+              shutdown_input: &ShutdownInput,
+              runtime: &mut Runtime) {
+        let file = self.state.cfg.spec_path_for(&service_spec);
+        if let Err(err) = fs::remove_file(&file) {
+            warn!("Tried to unload '{}', but couldn't remove the file '{}': {:?}",
+                  service_spec.ident,
+                  file.display(),
+                  err);
+        };
+        if let Some(future) = self.remove_service_from_state(&service_spec)
+                                  .map(|service| {
+                                      let shutdown_config =
+                                          ShutdownConfig::new(&shutdown_input, &service);
+                                      self.stop_with_config(service, shutdown_config)
+                                  })
+        {
+            runtime.spawn(future);
+        } else {
+            warn!("Tried to unload '{}', but couldn't find it in our list of running services!",
+                  service_spec.ident);
+        }
     }
 
     /// Create a future for stopping a Service. The Service is assumed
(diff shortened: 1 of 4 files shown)


baumanj:
> Note, a service can still be automatically unloaded for unsatisfied binds if a package update adds binds. Does this seem like reasonable behavior?

To make sure I'm understanding correctly:
1. `foocorp/bar/1.0` is loaded with auto-update from `stable` channel enabled
2. `foocorp/bar/2.0` is promoted to `stable`, this version adds a required bind
3. The running service `foocorp/bar` is updated to version 2.0, and due to the fact the bind cannot be satisfied, it is unloaded

Is that the scenario you're referring to?

davidMcneil:
@baumanj Yes, that is a correct scenario.

baumanj:
I think that's reasonable. The only alternative I see to that is leaving 1.0 running if starting 2.0 isn't successful, but that has different and more complicated consequences. While the above scenario may be unpleasant, it's straightforward and consistent.

stevendanna:
Regarding the additional binds: I think unloading the service is reasonable behavior given the other features, but it might be worth it to open an issue investigating whether there might be a feature that could allow users to handle that case better.  Changing binds is one of a handful of issues that makes it hard to correctly use auto-upgrade without human intervention.

components/sup/src/manager/commands.rs
```diff
@@ -185,60 +184,39 @@ pub fn service_load(mgr: &ManagerState,
...
-            // desired package identifier, it will be used;
-            // otherwise, we'll install the latest suitable
-            // version from the specified Builder channel.
-            util::pkg::satisfy_or_install(req, &source, &bldr_url, &bldr_channel)?;
-
-            mgr.cfg.save_spec_for(&spec)?;
-            req.info(format!("The {} service was successfully loaded", spec.ident))?;
+    let mut spec = if let Some(spec) = mgr.cfg.spec_for_ident(source.as_ref()) {
+        let force = opts.force.unwrap_or(false);
+        // We've seen this service  before. Thus `load`
```
stevendanna: [nit] Extra space in comment here.
```suggestion
        // We've seen this service before. Thus `load`
```

components/sup/src/manager/commands.rs
```diff
@@ -185,60 +184,39 @@ pub fn service_load(mgr: &ManagerState,
...
-            util::pkg::satisfy_or_install(req, &source, &spec.bldr_url, &spec.channel)?;
+    opts.into_spec(&mut spec);
 
-            mgr.cfg.save_spec_for(&spec)?;
-            req.info(format!("The {} service was successfully loaded", spec.ident))?;
-        }
+    // If a package exists on disk that satisfies the desired package identifier, it will be used.
+    // Otherwise, we'll install the latest suitable version from the specified Builder channel.
+    let package = util::pkg::satisfy_or_install(req, &source, &spec.bldr_url, &spec.channel)?;
+    if let Err(e) = spec.validate(&package) {
```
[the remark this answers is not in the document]
davidMcneil: Thanks for catching this. I agree that would have been very confusing. I was doing this to make the error code be `InvalidPayload` instead of `Internal`, but the proper way to do this is just fix the `From` implementation. This [commit](https://github.com/habitat-sh/habitat/pull/6730/commits/48db0b9e4ccd657f374ac08f8d3c8308bbdfb4aa) should do it.

components/sup/src/manager.rs
```diff
@@ -1267,36 +1247,55 @@ impl Manager {
...
+                                            self.organization.as_ref().map(|org| &**org),
+                                            self.state.gateway_state.clone())
+                        {
+                            Ok(s) => Some(s),
+                            Err(e) => {
+                                outputln!("Failed to load spec file for '{}' with error '{}'. \
+                                           This service will be unloaded.",
+                                          spec.ident,
+                                          e);
+                                self.unload(spec, &ShutdownInput::default(), runtime);
```
christophermaier: This refactoring highlights one of the things I've always felt was a little "weird" about this function, as well as how our other functions around Services are named. Perhaps now is the time to address it.

This function is just about persisting our state for the HTTP gateway. Thus, one should expect that it wouldn't change the broader state of the Supervisor. The existing state of the Supervisor should be "read-only" for this function - it should only be changing the state of the HTTP gateway (but really, we should have a function that just reads Supervisor state and creates the new gateway state, which we then pass to another function that modifies the gateway itself, but I digress). Having to pass a Tokio Runtime into this function is thus _super_ suprising and unexpected.

Then we come to the issue of naming and how existing functions are being used, with a side-trip into "ubiquitous language". From a user's perspective, "load" means "make the Supervisor aware of a given service", making it possible to run that service; "unload" means the opposite. However, once we get into the code, `Service::load` is really a constructor... I think it should be `Service::new`, really. The "load" operation is really something that takes place well before we get into the `Service` code, since "loading" is "create a spec file on disk". That we can even create a `Service` instance implies that that information has _already_ been loaded into the Supervisor.

Thus, if `Service::load` is really a constructor, then this `unload` operation isn't symmetric, because it is actually stopping a service as well as removing a spec file from disk.

Stepping back a bit, it seems that the only reason we are going through this "load" logic (very intentionally using air-quotes here), is because we don't have an in-memory representation of services that are loaded, but not running (e.g., because they are supposed to be stopped). If we had _that_, then we wouldn't need to constantly construct these `Service` instances. The persistence call could then just simply read Supervisor state. Additionally, we could push the handling of unloadable services out toward the periphery, and handle them when we first encounter them, rather than deeper in the code like this.

baumanj:
> I think unloading the service is reasonable behavior given the other features, but it might be worth it to open an issue investigating whether there might be a feature that could allow users to handle that case better. Changing binds is one of a handful of issues that makes it hard to correctly use auto-upgrade without human intervention.

100% agree. I think in the long run we'll want to make auto-update configuration more sophisticated than just on or off. The thing that leaps to mind is allowing semver specifications of what versions should be auto-updated _to_, though that would be tricky in the current state where we can't make many assumptions about the structure of the version field. In general, I think opening an issue to gather information about how auto update is really used and what additional features would be worth the additional complexity would be valuable.

components/sup/src/manager/commands.rs
```diff
@@ -205,7 +205,12 @@ pub fn service_load(mgr: &ManagerState,
             // desired package identifier, it will be used;
             // otherwise, we'll install the latest suitable
             // version from the specified Builder channel.
-            util::pkg::satisfy_or_install(req, &source, &bldr_url, &bldr_channel)?;
+            let package = util::pkg::satisfy_or_install(req, &source, &bldr_url, &bldr_channel)?;
+            if let Err(e) = spec.validate(&package) {
```
christophermaier: In retrospect, it's very odd that this _wasn't_ here before... weird.

Good catch!

components/sup/src/manager/commands.rs
```diff
@@ -238,6 +238,16 @@ pub fn service_load(mgr: &ManagerState,
...
+
+            // TODO (DM): Why do we use `spec.bldr_url` and `spec.channel` here?
+            // `opts.into_spec` will override these fields if they are defined in the `ServiceLoad`
+            // so we are essentially always using `bldr_url` and `bldr_channel`. In fact `bldr_url`
+            // and `bldr_channel` are not even neccessary, we could just always use what
+            // is in `spec` because `opts.into_spec` will set it appropriatly.
+            // Potentially, I could see the reason for the use of `spec` here trying to
+            // prevent updating the `bldr_url` and `channel` on a --force reload, but that
+            // is not the current behavior. With this change this method could be greatly
+            // simplified.
```
christophermaier: I've never noticed this before :thinking: 

I can't think of a compelling reason to differentiate between `&spec.bldr_url` and `&bldr_url` (same for channel). 

I think the _real_ question is why aren't we using the `spec.*` data in the case above? Seems like we should be taking the potentially-resolved-from-defaults `bldr_url` and channel and shoving them into spec and then just using those. Actually, the *real* real question is why aren't those defaults already resolved in the incoming `SvcLoad` instance :thinking: 

I'm all for simplifying this.

components/sup/src/manager/service/spec.rs
```diff
@@ -164,6 +164,7 @@ impl ServiceSpec {
     pub fn default_for(ident: PackageIdent) -> Self {
         let mut spec = Self::default();
         spec.ident = ident;
+        spec.bldr_url = habitat_sup_protocol::DEFAULT_BLDR_URL.to_string();
         spec
```
christophermaier: You can use this idiom to avoid using `mut`:

```rust
Self {ident: ident,
      bldr_url: DEFAULT_BLDR_URL.to_string(),
      ..Default::default()}
```

components/sup/src/manager/commands.rs
```diff
@@ -185,81 +184,39 @@ pub fn service_load(mgr: &ManagerState,
...
-            if !force {
-                return Err(net::err(ErrCode::Conflict,
-                                    format!("Service already loaded, unload '{}' \
-                                             and try again",
-                                            ident)));
-            }
+        spec
+    } else {
+        ServiceSpec::default()
+    };
```
christophermaier: One way to consolidate some of this might be:

```rust
let mut spec = mgr.cfg.spec_for_ident(source.as_ref()).unwrap_or_default();
if spec.force.unwrap_or(false) { // or spec.force.unwrap_or_default()
  return Err(...);
}
```

components/sup/src/manager/commands.rs
```diff
@@ -185,81 +184,39 @@ pub fn service_load(mgr: &ManagerState,
...
-            // is not the current behavior. With this change this method could be greatly
-            // simplified.
-            let package =
-                util::pkg::satisfy_or_install(req, &source, &spec.bldr_url, &spec.channel)?;
-            if let Err(e) = spec.validate(&package) {
-                return Err(net::err(ErrCode::InvalidPayload,
-                                    format!("Failed to validate '{}' due to '{}'",
-                                            ident, e)));
-            }
+    opts.into_spec(&mut spec);
```
christophermaier: Digging into this `into_spec` function / trait... it's a little weird.

One possibly interesting and more idiomatic refactoring might be to implement `From<SvcLoad> for ServiceSpec`, and then construct a way to "layer" two `ServiceSpec`s together.

That might also allow you to avoid a `mut` variable, which is often nice.

That's not a blocker or anything, but could be interesting to explore for a bit.

components/sup/src/manager/commands.rs
davidMcneil: :thinking: I am not sure this will work? We only want to check for `force` if the spec already existed. This would always check for it and require the user to always specify `--force`. Are you saying we should store `force` in the spec? Currently, it is set in `opts`.

davidMcneil:
> This PR constitutes a change in functionality, but there aren't any changes to tests. That implies that the existing behavior wasn't covered by tests. Let's take this opportunity to get this (rather subtle) area of the code under test. 

That would be great. I will work on adding tests as part of the larger refactoring alluded to [here](https://github.com/habitat-sh/habitat/pull/6730#discussion_r303999105). 

> Additionally, to be clear on terminology, [f57e049](https://github.com/habitat-sh/habitat/commit/f57e04919a523f86edf03584378d1383f3211017) is described as "Refactor service_load". My understanding of the term "refactor" is to change the internal implementation _without_ modifying the external behavior. This change seems to be a change in behavior, so the commit itself was not what I expected based on the description. Is there a different sense of the term that I'm not familiar with?

[f57e04](https://github.com/habitat-sh/habitat/pull/6730/commits/f57e04919a523f86edf03584378d1383f3211017) should be a pure refactoring when compared against [ed536a](https://github.com/habitat-sh/habitat/pull/6730/commits/ed536a8df11c19856b1c39061a71b9d6a3170dba) which adds the validation check (i.e. that commit was a refactoring while the entire PR is a change in functionality). The questions posed [here](https://github.com/habitat-sh/habitat/pull/6730#discussion_r304029601) are what is addressed in the refactoring. @baumanj was there a specific change in external behavior you where referring to beyond what was added by [ed536a](https://github.com/habitat-sh/habitat/pull/6730/commits/ed536a8df11c19856b1c39061a71b9d6a3170dba)?

components/sup/src/manager/commands.rs
baumanj: I agree that this won't really work. You'd have to pass the `force` opt into `mgr.cfg.spec_for_ident` or something to get the appropriate behavior, but I'm not sure that's what we want.

Alternatively, if there were a function that just checked for the existence of a spec, we might be able to make things nicer, but overall this logic is so convoluted that I don't think can be easily simplified. Before we even _try_, I'd want to get the existing behavior documented via tests so it can be more safely refactored.

baumanj:
> [f57e04](https://github.com/habitat-sh/habitat/pull/6730/commits/f57e04919a523f86edf03584378d1383f3211017) should be a pure refactoring when compared against [ed536a](https://github.com/habitat-sh/habitat/pull/6730/commits/ed536a8df11c19856b1c39061a71b9d6a3170dba) which adds the validation check (i.e. that commit was a refactoring while the entire PR is a change in functionality).
…
@baumanj was there a specific change in external behavior you where referring to beyond what was added by [ed536a](https://github.com/habitat-sh/habitat/pull/6730/commits/ed536a8df11c19856b1c39061a71b9d6a3170dba)?

Maybe it's a sufficiently complex refactoring that I'm having trouble seeing that they're logically identical. Looking at the pre-`ed536a` behavior of `service_load`, it seems that `satisfy_or_install` and `validate` is only called in the case that `mgr.cfg.spec_for_ident` returns `Some`, whereas afterwards both are called for `None` returns as well. Since `service_load` may return an error value based on what happens with `validate`, this seems like a behavioral difference, but maybe in practice it's not?

components/sup/src/manager/commands.rs
christophermaier: ah, correct... :grimacing:

davidMcneil:
Closing in favor of #6794

PR #6731, opened 2019-07-15
Add liveliness checker regression test

baumanj:
This should cover the issue fixed in https://github.com/habitat-sh/habitat/pull/6724

The first commit is expected to fail CI in order to validate the test. The behavior is fixed in the subsequent commit.

Additionally, improve the logging for the liveliness checker and switch its locking to `parking_lot` from the stdlib `Mutex`.

### DIFF
--- a/components/common/src/liveliness_checker.rs
+++ b/components/common/src/liveliness_checker.rs
@@ -1,6 +1,5 @@
 use parking_lot::Mutex;
 use std::{collections::HashMap,
-          path::PathBuf,
           thread::{self,
                    ThreadId},
           time::{Duration,
@@ -171,16 +170,15 @@ pub fn spawn_thread_alive_checker() {
 }
 
 fn spawn_thread_alive_checker_impl(delay: Duration) {
-    let executable = std::env::args_os().next()
-                                        .map(PathBuf::from)
-                                        .and_then(|first_arg| {
-                                            first_arg.components().last().map(|basename| {
-                                                                             basename.as_os_str()
+    let executable = std::env::current_exe().ok()
+                                            .and_then(|exe_path| {
+                                                exe_path.components().last().map(|basename| {
+                                                                                basename.as_os_str()
                                                                                  .to_string_lossy()
                                                                                  .into_owned()
-                                                                         })
-                                        })
-                                        .unwrap_or_else(|| "unknown-executable".into());
+                                                                            })
+                                            })
+                                            .unwrap_or_else(|| "unknown-executable".into());
 
     debug!("Spawning liveliness checker thread for {} PID {}",
            executable,
--- a/components/common/src/liveliness_checker.rs
+++ b/components/common/src/liveliness_checker.rs
@@ -194,11 +194,11 @@ fn spawn_thread_alive_checker_impl(delay: Duration) {
                               let max_time_since_death =
                                   ThreadDeadIgnoreDelay::configured_value().into();
                               loop {
+                                  thread::sleep(delay);
                                   let statuses = &mut THREAD_STATUSES.lock();
                                   check_thread_heartbeats(statuses, threshold);
                                   log_dead_threads(statuses);
                                   cull_dead_threads(statuses, max_time_since_death);
-                                  thread::sleep(delay);
                               }
                           })
                           .expect("Error spawning thread alive checker");
--- a/components/common/src/liveliness_checker.rs
+++ b/components/common/src/liveliness_checker.rs
@@ -1,5 +1,6 @@
+use parking_lot::Mutex;
 use std::{collections::HashMap,
-          sync::Mutex,
+          path::PathBuf,
           thread::{self,
                    ThreadId},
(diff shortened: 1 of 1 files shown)


baumanj:
Open question: is it worth modifying `mark_thread_alive` to make use of `parking_lot`'s [`try_lock_for`](https://docs.rs/lock_api/0.3.1/lock_api/struct.Mutex.html#method.try_lock_for), ensuring that real work isn't ever delayed for more than a given threshold in order to facilitate liveliness tracking? If so, what should that threshold be?

davidMcneil:
> Open question: is it worth modifying `mark_thread_alive` to make use of `parking_lot`'s [`try_lock_for`](https://docs.rs/lock_api/0.3.1/lock_api/struct.Mutex.html#method.try_lock_for)

While this seems like a good idea, I wonder if it adds complexity that could make it more confusing to interpret results. For example, the liveness checker could think a thread was deadlocked when it is actually fine correct? It was just not able to acquire the lock. I guess this confussion could be alleviated with sufficient logging. If the `try_for_lock` ever failed something would be seriously wrong correct? This is not something we foresee happening frequently?

Ultimately, I dont have a strong preference either way.

components/common/src/liveliness_checker.rs
```diff
@@ -165,22 +167,38 @@ fn unregister_thread_impl<T, E: ToString>(_: CheckedThread,
...
-                                  {
-                                      let statuses =
-                                          &mut THREAD_STATUSES.lock()
-                                                              .expect("THREAD_STATUSES poisoned");
-                                      check_thread_heartbeats(statuses, threshold);
-                                      log_dead_threads(statuses);
-                                      cull_dead_threads(statuses, max_time_since_death);
-                                  }
                                   thread::sleep(delay);
+                                  let statuses = &mut THREAD_STATUSES.lock();
```
christophermaier: heh... sneaky order swap here to avoid the explicit scoping :smile:

baumanj: The benefit of adding tests is it justifies more sneakiness 😉

components/common/src/liveliness_checker.rs
```diff
@@ -165,22 +167,38 @@ fn unregister_thread_impl<T, E: ToString>(_: CheckedThread,
...
+    let executable = std::env::args_os().next()
+                                        .map(PathBuf::from)
+                                        .and_then(|first_arg| {
+                                            first_arg.components().last().map(|basename| {
+                                                                             basename.as_os_str()
+                                                                                 .to_string_lossy()
+                                                                                 .into_owned()
+                                                                         })
+                                        })
+                                        .unwrap_or_else(|| "unknown-executable".into());
```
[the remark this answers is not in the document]
baumanj: Good call! I'll look into that

baumanj:
> While this seems like a good idea, I wonder if it adds complexity that could make it more confusing to interpret results.

Indeed. It would lead to false positives, but the justification would be that false positives are preferable to slowing down the operation of all the threads that call `mark_thread_alive`. That seems uncontroversial. The question is, what's the likelihood of this actually helping us out in practice, and does the benefit outweigh the added complexity?

> If the `try_for_lock` ever failed something would be seriously wrong correct? This is not something we foresee happening frequently?

It's not something we foresee at all. But then again, we didn't foresee the locking bug you fixed in https://github.com/habitat-sh/habitat/pull/6724, so 🤷‍♂

> What would using `try_lock_for` look like?

```rust
pub fn mark_thread_alive() -> CheckedThread { mark_thread_alive_impl(&mut THREAD_STATUSES.lock()) }
```
becomes
```rust
pub fn mark_thread_alive() -> CheckedThread {
    if let Some(mut lock) = THREAD_STATUSES.try_lock_for(TIMEOUT) {
        mark_thread_alive_impl(&mut lock)
    } else {
        error!("THREAD_STATUSES lock not acquired within {:?}", TIMEOUT);
        CheckedThread(std::ptr::null())
    }
}
```
with `TIMEOUT` defined somewhere and possibly a better error message.

> Would that mean that the `mark_thread_alive` call itself would only take a maximum of X seconds before moving on?

Exactly.

If none of us feel strongly, I'll err on the side of YAGNI/simplicity and leave it out; I think it's fairly well covered by tests now.

components/common/src/liveliness_checker.rs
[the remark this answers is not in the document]
baumanj: [Not a huge reduction in code](https://github.com/habitat-sh/habitat/pull/6731/commits/5aa83bdc56a601c81a0d1669ef612178935a04e8), but a clear win for simplicity and intentionality

christophermaier:
@baumanj I'm cool with the YAGNI route for now, but we can keep it in mind as a potential refinement in the future.

components/common/src/liveliness_checker.rs
christophermaier: Sneakiness-Driven Development!

PR #6743, opened 2019-07-18
Remove templated data when installing a new package

### LINKED ISSUE #4925: Lifecycle hooks removed from a package are not removed from service directory
If a package is updated and one of its lifecycle hooks are removed as part of the change, the supervisor does not remove the hook from the service directory.

davidMcneil:
Resolves #4925 

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -44,7 +44,8 @@ use crate::{api_client::{self,
                     fs::{cache_key_path,
                          pkg_install_path,
                          svc_hooks_path,
-                         AtomicWriter},
+                         AtomicWriter,
+                         SvcDir},
                     package::{list::temp_package_directory,
                               Identifiable,
                               PackageArchive,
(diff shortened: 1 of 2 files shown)


stevendanna:
Apologies if this is off base, it has been a while since I've been in this code base.

I'm not sure if this is safe.  The `svc_config_path` contains the rendered templates of a loaded habitat service.  It may be the case that the service expects to access those files at runtime. A package installation may happen while the service is running, deleting files that the service may attempt to access.  There is no guarantee that an installed package is ever going to be started as a service, so the act of installing a package feels like the wrong place to put this cleanup.

davidMcneil:
> Apologies if this is off base, it has been a while since I've been in this codebase.

Not at all.

It sounds like my understanding of how installing and loading packages while a service is running needs to be refined. I agree that with the points you brought up this seems very unsafe. I will have to investigate further and come back to this.

stevendanna:
> It sounds like my understanding of how installing and loading packages while a service is running needs to be refined.

I think a key thing in this case is that `hab pkg install` could be run by a user on that machine independent of a particular service's upgrade cycle.

davidMcneil:
That makes sense. I could be running `some_origin/test/1.0` and then install `some_origin/test/2.0` (independent of an upgrade) and that would break `some_origin/test/1.0`'s execution. I will close this and rethink where the logic should go. Thanks for catching this.

PR #6748, opened 2019-07-20
[hab-plan-build] Normalize file checksums before comparison

### LINKED ISSUE #5427: Hab studio download_file checksum comparison should be case-insensitive
- hab-studio version 0.57 on Mac OS X

This is a fairly easy one to demonstrate:

From `plan.sh`:

```
pkg_shasum="7C4C34E567959677C1EF3A450992AFCE43FB4B8C4BD5B98C5E9D48AFD9CA2223"

...

  if [ ! -f $HAB_CACHE_SRC_PATH/$pkg_filename ]; then
    download_file "${pkg_source}" "${HAB_CACHE_SRC_PATH}/${pkg_filename}"
  fi
```

And then, when running `hab studio run build`:

```
   itm: Downloaded '/hab/cache/src/itm6.3.0.tar.gz'
   itm: Verifying itm6.3.0.tar.gz
   itm: WARN Checksum invalid for itm6.3.0.tar.gz:
   itm: WARN    Expected: 7C4C34E567959677C1EF3A450992AFCE43FB4B8C4BD5B98C5E9D48AFD9CA2223
   itm: WARN    Computed: 7c4c34e567959677c1ef3a450992afce43fb4b8c4bd5b98c5e9d48afd9ca2223
   itm: Build time: 1m13s
   itm: Exiting on error
```

Notice that the checksum matches. The only distinction is the case insensitivity. This should be an easy fix; `ucase` or `lcase` both expected and found checksums before comparison.

stevendanna:
This changes our checksum comparison to be case-insensitive, allowing
user to use hexadecimal representations of the checksum that use
uppercase letters.

Fixes #5427

### DIFF
--- a/components/plan-build/bin/public.bash
+++ b/components/plan-build/bin/public.bash
@@ -382,19 +382,38 @@ download_file() {
 # Will return 0 if the shasums match, and 1 if they do not match. A message
 # will be printed to stderr with the expected and computed shasum values.
 verify_file() {
-  build_line "Verifying $1"
-  local checksum
+  local filename=$1
+  local expected_checksum=$2
+  local computed_checksum
+
+  build_line "Verifying $filename"
+  expected_checksum=$(normalize_checksum "$expected_checksum")
+
   # shellcheck disable=2154
-  read -r checksum _ < <($_shasum_cmd "$HAB_CACHE_SRC_PATH"/"$1")
-  if [[ $2 = "$checksum" ]]; then
+  read -r computed_checksum _ < <($_shasum_cmd "$HAB_CACHE_SRC_PATH"/"$filename")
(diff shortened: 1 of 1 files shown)


stevendanna:
Was going through some older tickets and saw this issue.  This isn't a priority for me personally, but it seemed like an easy enough thing to fix.  If we decide we'd rather not, let's also close the related ticket.

christophermaier:
@mwrock The Windows version of `hab-plan-build` needs this too, correct?

christophermaier:
Also, thanks for this @stevendanna !

mwrock:
Nope @christophermaier . Powershell string comparisons are case insensitive.

christophermaier:
@mwrock That's surprising. Good for us, but surprising. :man_shrugging: 

Thanks!

PR #6758, opened 2019-07-24
Adds origin creation support to CLI

eeyun:

### DIFF
--- a/components/builder-api-client/src/artifactory.rs
+++ b/components/builder-api-client/src/artifactory.rs
@@ -702,6 +702,8 @@ impl BuilderAPIProvider for ArtifactoryClient {
         Err(Error::NotSupported)
     }
 
+    fn create_origin(&self, _origin: &str, _token: &str) -> Result<()> { Err(Error::NotSupported) }
+
     fn create_origin_secret(&self,
                             _origin: &str,
                             _token: &str,
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -462,6 +462,32 @@ impl BuilderAPIProvider for BuilderAPIClient {
                       progress)
     }
 
+    /// Create an origin
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Unable to authenticate
+    fn create_origin(&self, origin: &str, token: &str) -> Result<()> {
+        let body = json!({
+            "name": origin,
+        });
+
+        let sbody = serde_json::to_string(&body)?;
+
+        let res = self.add_authz(self.0.post("depot/origins"), token)
+                      .body(&sbody)
+                      .header(Accept::json())
+                      .header(ContentType::json())
+                      .send()?;
+
+        if res.status != StatusCode::Created {
+            return Err(err_from_response(res));
+        }
+
+        Ok(())
+    }
+
     /// Create secret for an origin
     ///
     /// # Failures
--- a/components/hab/src/command/origin/create.rs
+++ b/components/hab/src/command/origin/create.rs
@@ -1,0 +1,32 @@
+use crate::{api_client::{self,
+                         Client},
+            common::ui::{Status,
+                         UIWriter,
+                         UI},
+            error::{Error,
+                    Result},
+            PRODUCT,
+            VERSION};
+use hyper::status::StatusCode;
+
+pub fn start(ui: &mut UI, bldr_url: &str, token: &str, origin: &str) -> Result<()> {
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
+
+    ui.status(Status::Creating, format!("origin {}.", origin))?;
+
+    match api_client.create_origin(origin, token) {
+        Ok(_) => {
+            ui.status(Status::Created, format!("origin {}.", origin))?;
+            Ok(())
+        }
+        Err(api_client::Error::APIError(StatusCode::Conflict, _msg)) => {
+            ui.status(Status::Skipping,
+                      format!("creation of origin {}. Origin already exists!", origin))?;
+            Ok(())
+        }
+        Err(e) => {
+            ui.fatal(format!("Failed to create origin {}, {:?}.", origin, e))?;
+            Err(Error::from(e))
+        }
+    }
+}
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -204,6 +204,7 @@ fn start(ui: &mut UI, feature_flags: FeatureFlag) -> Result<()> {
                         _ => unreachable!(),
                     }
                 }
+                ("create", Some(m)) => sub_origin_create(ui, m)?,
                 ("delete", Some(m)) => sub_origin_delete(ui, m)?,
                 _ => unreachable!(),
             }
@@ -451,8 +452,15 @@ fn sub_origin_secret_list(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     command::origin::secret::list::start(ui, &url, &token, &origin)
 }
 
+fn sub_origin_create(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
+    let origin = m.value_of("ORIGIN").expect("required ORIGIN");
+    let url = bldr_url_from_matches(&m)?;
+    let token = auth_token_param_or_env(&m)?;
+    command::origin::create::start(ui, &url, &token, &origin)
+}
+
 fn sub_origin_delete(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
-    let origin = origin_param_or_env(&m)?;
+    let origin = m.value_of("ORIGIN").expect("required ORIGIN");
     let url = bldr_url_from_matches(&m)?;
     let token = auth_token_param_or_env(&m)?;
     command::origin::delete::start(ui, &url, &token, &origin)
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -276,6 +276,8 @@ pub trait BuilderAPIProvider: Sync + Send {
                                           progress: Option<Self::Progress>)
                                           -> Result<PathBuf>;
 
+    fn create_origin(&self, origin: &str, token: &str) -> Result<()>;
+
     fn create_origin_secret(&self,
                             origin: &str,
                             token: &str,
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -242,6 +242,16 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
             (about: "Commands relating to Habitat origin keys")
             (aliases: &["o", "or", "ori", "orig", "origi"])
             (@setting ArgRequiredElseHelp)
+            (@subcommand create =>
+                (about: "Creates a new Builder origin")
+                (aliases: &["cre", "crea"])
+                (@arg ORIGIN: +required {valid_origin} "The origin to be created")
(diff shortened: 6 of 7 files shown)


components/builder-api-client/src/builder.rs
```diff
@@ -462,6 +462,35 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
 
+    /// Create an origin
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin already exists
+    ///  * Y'aint got no auth
+    fn create_origin(&self, origin: &str, token: &str) -> Result<()> {
+        let path = format!("depot/origins");
```
raskchanky: Since you're only ever using this as a `&str` in the call to `self.0.post()` below, you shouldn't need this line at all.  You should be able to just pass `"depot/origins"` to the call to `self.0.post()` directly.

components/hab/src/command/origin/create.rs
```diff
@@ -0,0 +1,33 @@
...
+    ui.status(Status::Creating, format!("origin {}.", origin))?;
+
+    match api_client.create_origin(origin, token) {
+        Ok(_) => {
+            ui.status(Status::Created, format!("origin {}.", origin))
+              .map_err(Error::from)
+        }
+        Err(api_client::Error::APIError(StatusCode::Conflict, msg)) => {
+            ui.fatal(format!("Unable to create origin {}", origin))?;
+            ui.fatal("Origins may only be created if they don't already exist")?;
```
raskchanky: It would probably be good to be consistent with punctuation/line endings for these messages and include periods at the end.  Similarly with the error messages below.

components/hab/src/command/origin/create.rs
```diff
@@ -0,0 +1,33 @@
...
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
+
+    ui.status(Status::Creating, format!("origin {}.", origin))?;
+
+    match api_client.create_origin(origin, token) {
+        Ok(_) => {
+            ui.status(Status::Created, format!("origin {}.", origin))
+              .map_err(Error::from)
+        }
+        Err(api_client::Error::APIError(StatusCode::Conflict, msg)) => {
```
chefsalim: I don't think we should fail on a conflict return - instead we can just indicate that the origin already exists, similar to how origin key upload handles conflicts.  This will also make it much easier to use in the on-prem scripts.

components/hab/src/command/origin/create.rs
```diff
@@ -0,0 +1,32 @@
...
+            ui.status(Status::Skipping,
+                      format!("creation of origin {}. Origin already exists!", origin))?;
+            Err(Error::APIClient(api_client::Error::APIError(StatusCode::Conflict, msg)))
```
chefsalim: This should return Ok(())

components/hab/src/command/origin/create.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub fn start(ui: &mut UI, bldr_url: &str, token: &str, origin: &str) -> Result<()> {
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
+
+    ui.status(Status::Creating, format!("origin {}.", origin))?;
+
+    match api_client.create_origin(origin, token) {
+        Ok(_) => {
+            ui.status(Status::Created, format!("origin {}.", origin))
+              .map_err(Error::from)
```
chefsalim: This should return Ok(())

components/hab/src/command/origin/create.rs
```diff
@@ -0,0 +1,32 @@
...
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
+
+    ui.status(Status::Creating, format!("origin {}.", origin))?;
+
+    match api_client.create_origin(origin, token) {
+        Ok(_) => {
+            ui.status(Status::Created, format!("origin {}.", origin))?;
+            Ok(())
+        }
+        Err(api_client::Error::APIError(StatusCode::Conflict, msg)) => {
```
chefsalim: Unused var clippy err:

22 \|         Err(api_client::Error::APIError(StatusCode::Conflict, msg)) => {
--
  | \|                                                               ^^^ help: consider prefixing with an underscore: `_msg`

components/builder-api-client/src/builder.rs
```diff
@@ -462,6 +462,33 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+                      .body(&sbody)
+                      .header(Accept::json())
+                      .header(ContentType::json())
+                      .send()?;
+
+        if res.status != StatusCode::Created {
+            return Err(err_from_response(res));
+        }
+
+        Ok(())
```
baumanj: I find code a bit simpler to understand when
1. The common/happy case comes first
1. The conditions are positive rather than negated
1. Explicit returns are avoided

Sometimes these conflict, but in this case, I think we can have them all and it leads to a structure that I find more pleasing:
```rust
        if res.status == StatusCode::Created {
            Ok(())
        } else {
            Err(err_from_response(res))
        }
```

components/builder-api-client/src/builder.rs
```diff
@@ -462,6 +462,33 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+    ///  * Y'aint got no auth
+    fn create_origin(&self, origin: &str, token: &str) -> Result<()> {
+        let body = json!({
+            "name": origin,
+        });
+
+        let sbody = serde_json::to_string(&body)?;
+
+        let res = self.add_authz(self.0.post("depot/origins"), token)
+                      .body(&sbody)
```
baumanj: You can eliminate an additional binding* and avoid having to handle an error that can't really occur by using `serde_json::Value`'s `to_string` method:
```suggestion
                      .body(&body.to_string())
```

\* Not that additional bindings are bad, but the fewer things to keep in one's head, the easier I find it to reason about code. Another alternative is to re-bind `body` as a `String` rather than introduce `body`. Shadowing may be something to avoid in other languages, but it's [idiomatic rust](https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html?highlight=shadow#shadowing).

components/hab/src/main.rs
```diff
@@ -451,6 +452,13 @@ fn sub_origin_secret_list(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     command::origin::secret::list::start(ui, &url, &token, &origin)
 }
 
+fn sub_origin_create(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
+    let origin = origin_param_or_env(&m)?;
```
baumanj: `<ORIGIN>` is a required argument [per the clap definition](https://github.com/habitat-sh/habitat/pull/6758/files#diff-ee10e75b553edc813cf07f6a5036374cR248), so falling back to the env var seems contradictory. Our typical pattern here has been something like:
```suggestion
    let origin = m.value_of("ORIGIN").expect("required ORIGIN");
```

components/hab/src/main.rs
```diff
@@ -451,6 +452,13 @@ fn sub_origin_secret_list(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+    let url = bldr_url_from_matches(&m)?;
```
baumanj: `bldr_url_from_matches` is unfortunately named, especially when viewed next to `auth_token_param_or_env`. It makes it appear that `bldr_url_from_matches ` will *not* look in environment variables for a value, but it will. I'd suggest renaming `bldr_url_from_matches` to have a name more indicative of its behavior. It'll touch a decent number of places, but it should be a find/replace with essentially zero risk.

components/hab/src/command/origin/create.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    match api_client.create_origin(origin, token) {
+        Ok(_) => {
+            ui.status(Status::Created, format!("origin {}.", origin))?;
+            Ok(())
+        }
+        Err(api_client::Error::APIError(StatusCode::Conflict, _msg)) => {
+            ui.status(Status::Skipping,
+                      format!("creation of origin {}. Origin already exists!", origin))?;
+            Ok(())
```
baumanj: I think it makes more sense to make this an error from the standpoint of automation. Outputting a different exit code would also be useful. It's easier to intentionally ignore a particular kind of error than find one based on output our code considers "ok". If people really desire the behavior that trying to create an existing origin is not an error, we can always add an additional flag that lets people opt into that behavior.

components/builder-api-client/src/builder.rs
chefsalim: I'll fix up some of this in my PR after this merges... although I decided to go with the negated case first as it ended up reading a bit cleaner to me

components/builder-api-client/src/builder.rs
chefsalim: This and a couple other serde handings will get fixed in my PR with reqwest

components/hab/src/main.rs
chefsalim: Let's do that separately to minimize any merge conflicts with my PR

components/hab/src/command/origin/create.rs
chefsalim: This is consistent with commands like origin key upload, and the automation we're adding this for will be much easier if we just follow that pattern.  We can do more complicated things later..

components/hab/src/main.rs
chefsalim: I'm guessing this is copy-paste from some other command

components/builder-api-client/src/builder.rs
```diff
@@ -462,6 +462,33 @@ impl BuilderAPIProvider for BuilderAPIClient {
                       progress)
     }
 
+    /// Create an origin
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin already exists
+    ///  * Y'aint got no auth
```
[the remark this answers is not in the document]
chefsalim: Maybe just remove the comments rather than trying to wordsmith this stuff..

components/builder-api-client/src/artifactory.rs
```diff
@@ -702,6 +702,8 @@ impl BuilderAPIProvider for ArtifactoryClient {
         Err(Error::NotSupported)
     }
 
+    fn create_origin(&self, _origin: &str, _token: &str) -> Result<()> { Err(Error::NotSupported) }
```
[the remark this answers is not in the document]
chefsalim: We can save for later - the future of the Artifactory plug in is not clear atm

components/hab/src/command/origin/create.rs
baumanj: 100% agree that consistency is the best thing. Just wanted to point out that in general it's *way* harder to make something that succeeds fail later than the other way around. Since we're adding new functionality here, let's be careful we avoid YAGNI, but also don't make choices that paint us into corners.

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
baumanj: These seem like good, useful comments to me. I'd only suggest changing the last one so the tone/grammar is consistent with the rest of our codebase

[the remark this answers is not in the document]
chefsalim: The second comment is not correct... (currently)..

components/hab/src/main.rs
eeyun: Corrected.

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
eeyun: Corrected.

Issue #6682, opened 2019-06-24
Enable building on newer Ubuntu versions


chefsalim:
This is a counterpart of Builder issue: https://github.com/habitat-sh/builder/issues/1105

The latest versions of Ubuntu (18.10, 19.04, etc) have migrated to openssl 1.1.1. Unfortunately the Rust crates we are locked to fail to build with this version of openssl. Downgrading openssl is not recommended or a good idea. We need to upgrade some locked crates - basically reqwest from 0.8.x -> 0.9.x, and deprecate direct usage of hyper. This will also simplify our stack and give additional flexibility in dealing with proxies and certs.

chefsalim:
This will be resolved by https://github.com/habitat-sh/habitat/pull/6759
Closing so we don't need to track both on the board

Issue #6460, opened 2019-04-18
Upgrade hyper crate to current, async version APIs.


fnichol:
This chore involves upgrading to the newer APIs, starting after version `0.10.*` in the `http-client` component. There are 4 considerations we need to address:

* At a minimum, provide a synchronous API for existing call sites
* Consider an alternate, asynchronous API for calling code that can support this (most likely, this would be an additional change)
* Upgrade (possibly needing to reimplement) the related OpenSSL code which has a custom strategy for finding CA certificates on the 3 major platforms (i.e. Windows, Linux, macOS)
* Upgrade (possibly needing to reimplement) the HTTP proxy code which implements an expected behavior of proxy support that didn't ship with hyper at the time (it may now, it may not)

PR #6759, opened 2019-07-25
Update http-client and builder-api-client from hyper to reqwest.

chefsalim:
This change has two significant components: 
1. Migrate http-client and builder-api-client crates from hyper to reqwest
2. Modify the cert handling to be more flexible and sane

The reqwest migration allows us to remove custom handlers for proxy and SSL support and use the library implementations instead. It also simplifies the handling in the builder api client crate. It also allows us to build on newer versions of Ubuntu which have updated openssl libraries.

The cert handling changes allow more robustness and flexibility in the client. We now add the cacerts to the root certificate list instead of replacing it. We also allow dropping new SSL certs into the hab cache ssl directory that also get added to the root certificate list. Finally we can now support both PEM and DER cert formats.

The behavior in cert handling for the Mac platform has also been brought in line with the other platforms. Previously, we did basically left the Mac to its system certs. Now the same logic is used consistently.

Resolves https://github.com/habitat-sh/habitat/issues/6682 and https://github.com/habitat-sh/habitat/issues/6460

---
This change is [Reviewable](https://reviewable.io/reviews/habitat-sh/habitat/6759)


mwrock:
I have not reviewed the code yet but I have validated that this builds and successfully interacts with bldr on Windows.

components/builder-api-client/src/artifactory.rs
```diff
@@ -32,8 +37,8 @@ use crate::{error::{Error,
             OriginKeyIdent,
             SchedulerResponse};
```
raskchanky: We typically delete all the blank lines in between `use` clauses now and let `rustfmt` sort them.

components/builder-api-client/src/builder.rs
```diff
@@ -31,8 +30,7 @@ use crate::{error::{Error,
                                  PackageIdent,
                                  PackageTarget},
                        ChannelIdent},
-            hab_http::{util::decoded_response,
-                       ApiClient},
```
raskchanky: Same comment here as above re: deleting all the blank lines in between the `use` statements.

components/builder-api-client/src/builder.rs
```diff
@@ -173,57 +170,60 @@ impl BuilderAPIClient {
...
 
-        let file_name = res.headers
-                           .get::<XFileName>()
-                           .expect("XFileName missing from response")
-                           .to_string();
+        let file_name = resp.headers()
+                            .get(X_FILENAME)
+                            .expect("XFileName missing from response")
+                            .to_str()
+                            .expect("Invalid X-Filename");
```
raskchanky: I notice this pattern is repeated a lot, where in order to retrieve the value of some header, we have to call `headers().get(NAME_OF_HEADER).expect("something's missing").to_str().expect("Invalid something or other")`.

It'd be nice if there was a way to abstract that out into something that could be reused every time we need to get a header value out, ideally without the need to call `expect()` twice, but at the very least just to avoid all the repetition.

components/builder-api-client/src/builder.rs
```diff
@@ -369,22 +378,23 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
 
-        if res.status != StatusCode::Ok {
-            return Err(err_from_response(res));
-        }
+        if resp.status() != StatusCode::OK {
+            Err(err_from_response(resp))
+        } else {
+            let mut encoded = String::new();
+            resp.read_to_string(&mut encoded).map_err(Error::IO)?;
+            trace!("Body: {:?}", encoded);
```
raskchanky: Good call on moving this to `trace!`.

components/builder-api-client/src/builder.rs
```diff
@@ -499,15 +508,18 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
 
-        // We expect NoContent, because the origin must be empty to delete
-        if res.status == StatusCode::NoContent {
-            Ok(())
+        // We expect NO_CONTENT, because the origin must be empty to delete
+        if resp.status() != StatusCode::NO_CONTENT {
+            Err(err_from_response(resp))
         } else {
-            Err(err_from_response(res))
+            Ok(())
```
raskchanky: Slight nitpick, but I think the logic for this read better how it originally was:

`if status == NoContent` { Ok } else { Err }`

components/builder-api-client/src/builder.rs
```diff
@@ -878,22 +923,23 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
 
-        match result {
-            Ok(Response { status: StatusCode::Created,
-                          .. }) => Ok(()),
-            Ok(response) => Err(err_from_response(response)),
-            Err(e) => Err(Error::from(e)),
+        if resp.status() != StatusCode::OK {
+            Err(err_from_response(resp))
+        } else {
+            Ok(())
```
raskchanky: I left a similar comment elsewhere, and this is totally personal preference, but I think this reads better as:

```rust
if resp.status() == StatusCode::Ok {
    Ok(())
} else {
    Err(err_from_response(resp))
}
```

Similarly elsewhere.

components/builder-api-client/src/builder.rs
```diff
@@ -667,29 +692,34 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
-            Ok(Response { status: StatusCode::Created,
-                          .. }) => Ok(()),
-            Ok(response) => Err(err_from_response(response)),
-            Err(e) => Err(Error::from(e)),
+        debug!("Status: {:?}", resp.status());
+
+        if resp.status() != StatusCode::OK {
+            Err(err_from_response(resp))
+        } else {
+            Ok(())
```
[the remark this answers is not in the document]
christophermaier: You could also add something like this to the trait to start abstracting the common "parse the response as a JSON and return the type" pattern:

```rust
    fn parse_json_if_status<T: serde::de::DeserializeOwned>(mut self,
                                                            code: reqwest::StatusCode)
                                                            -> Result<T> {
        if self.status() == code {
            Ok(self.json()?)
        } else {
            Err(err_from_response(self))
        }
    }
```
then just call

```rust
resp.parse_json_if_status(StatusCode::OK)
```
etc.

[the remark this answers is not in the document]
christophermaier: (these suggestions can all be modified, of course; I just wanted to provide an example)

components/builder-api-client/src/artifactory.rs
chefsalim: Will do

components/builder-api-client/src/builder.rs
chefsalim: :+1:

components/builder-api-client/src/builder.rs
chefsalim: Good point, will look at that

components/builder-api-client/src/builder.rs
chefsalim: I went back and forth on this, and ended up keeping the negated checks first (as there were already many of those) and after a while it seemed to read a bit better that way to me. I'll look at it again...

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
-                   SSL_VERIFY_NONE};
+
 use url::Url;
 
-use crate::{error::{Error,
-                    Result},
-            net::ProxyHttpsConnector,
-            proxy::{proxy_unless_domain_exempted,
-                    ProxyInfo},
-            ssl};
```
raskchanky: Same comment as previously re: deleting blank lines in the `use` statements.

components/builder-api-client/src/artifactory.rs
```diff
@@ -32,8 +37,8 @@ use crate::{error::{Error,
...
 
-header! { (XFileName, "X-Artifactory-Filename") => [String] }
-header! { (XJFrogArtApi, "X-JFrog-Art-Api") => [String] }
+const X_JFROG_ART_API: &str = "x-jfrog-art-api";
+const X_ARTIFACTORY_FILENAME: &str = "x-artifactory-filename";
```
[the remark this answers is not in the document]
chefsalim: HeaderName::from_static requires the string be lowercase

components/builder-api-client/src/artifactory.rs
```diff
@@ -90,37 +95,42 @@ impl ArtifactoryClient {
...
         fs::create_dir_all(&dst_path)?;
 
-        let file_name = res.headers
-                           .get::<XFileName>()
+        let file_name = res.headers()
+                           .get(X_ARTIFACTORY_FILENAME)
                            .expect("X-Artifactory-Filename missing from response")
-                           .to_string();
+                           .to_str()
+                           .expect("Invalid X-Artifactory-Filename");
```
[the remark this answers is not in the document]
chefsalim: Mostly convenience.. I'll look at encapsulating out this API as its a pain

components/builder-api-client/src/artifactory.rs
```diff
@@ -90,37 +95,42 @@ impl ArtifactoryClient {
...
              match progress {
                  Some(mut progress) => {
-                     let size: u64 = res.headers
-                                        .get::<hyper::header::ContentLength>()
-                                        .map_or(0, |v| **v);
+                     let cl = res.headers()
+                                 .get(CONTENT_LENGTH)
+                                 .expect("Content length missing")
+                                 .to_str()
+                                 .expect("Content length invalid");
```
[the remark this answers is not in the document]
chefsalim: Yep

components/builder-api-client/src/builder.rs
```diff
@@ -536,23 +551,26 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
 
-        let mut res = self.add_authz(self.0.get(&path), token).send()?;
+        let path = format!("depot/origins/{}/secret", origin);
+        let mut resp = self.add_authz(self.0.get(&path), token).send()?;
+        debug!("Status: {:?}", resp.status());
 
-        if res.status != StatusCode::Ok {
-            return Err(err_from_response(res));
+        if resp.status() != StatusCode::OK {
+            Err(err_from_response(resp))
```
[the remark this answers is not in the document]
chefsalim: Will look at switching

components/builder-api-client/src/builder.rs
```diff
@@ -536,23 +551,26 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+        } else {
+            let mut encoded = String::new();
+            resp.read_to_string(&mut encoded)
+                .map_err(Error::BadResponseBody)?;
+            trace!("Body: {:?}", encoded);
+
+            let secret_keys: Vec<String> =
```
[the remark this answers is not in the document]
chefsalim: Will update

components/builder-api-client/src/builder.rs
```diff
@@ -597,19 +615,23 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
-        Ok(revisions)
+        if resp.status() != StatusCode::OK {
+            Err(err_from_response(resp))
+        } else {
+            let mut encoded = String::new();
+            resp.read_to_string(&mut encoded)
+                .map_err(Error::BadResponseBody)?;
+            trace!("Body: {:?}", encoded);
+
+            let revisions: Vec<OriginKeyIdent> =
```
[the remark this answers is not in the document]
chefsalim: OK

components/builder-api-client/src/builder.rs
```diff
@@ -622,32 +644,35 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
-            return Err(err_from_response(res));
+        if resp.status() != StatusCode::OK {
+            Err(err_from_response(resp))
+        } else {
+            let mut encoded = String::new();
+            resp.read_to_string(&mut encoded)
+                .map_err(Error::BadResponseBody)?;
+            trace!("Body: {:?}", encoded);
+
+            let channels: Vec<String> = serde_json::from_str::<Vec<String>>(&encoded)?.into_iter()
```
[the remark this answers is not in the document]
chefsalim: I did that and ended up reverting that change since the ability to trace the body could be pretty valuable for debugging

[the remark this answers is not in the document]
chefsalim: Will look at a helper

baumanj:
Wasn't able to get to this today, but will plan to take a look first thing tomorrow.

components/builder-api-client/src/builder.rs
christophermaier: An idea just occurred to me for this stuff. It would also be something that justifies its own PR, but I figured I'd mention it.

One interesting thing we _might_ want to consider here (at least until we can move our logging over to some kind of tagged / structured logging system, that is), is to add a target to the log call, like so:

```rust
trace!(target: "builder_api_traffic::fetch_rdeps::body", "{:?}", encoded);
```
If you don't explicitly set a target, it defaults to the module you're currently in (thus the module-like `::`-delimited path). _This_ is what you're actually operating on when you do something like `RUST_LOG=foo::bar::baz=trace`.

However, nothing stops you from adding your own targets. This basically lets us establish our own custom parallel logging hierarchy. If you want to see all traffic at trace level, use `builder_api_traffic=trace`. If you only want to see the info for this particular API call, narrow it to `builder_api_traffic::fetch_rdeps`. I added `::body` as an example of how to further refine, but that's not needed.

Ideally, we'd have structured logging so that we could just annotate what's happening here (tack on IDs, status codes, header information, responses, etc.). Depending on how things are actually happening under the hood, we might already be able to use [log_mdc](https://github.com/sfackler/rust-log-mdc) for this (which I've been using on some other in-progress work; however using it with futures requires a bit of extra consideration). However, until we get there, we could use this parallel hierarchy technique to squeeze a bit more functionality and usefulness out of the base logging facilities to extract more information out of our code. 

If we were to go this route, I'd recommend putting the logic behind some helper functions, both to ensure we're logging consistently and with the same overall structure, and to ease a future transition to a real structured logging setup.

components/builder-api-client/src/builder.rs
```diff
@@ -399,28 +409,27 @@ impl BuilderAPIProvider for BuilderAPIClient {
                                    token: &str,
                                    promote: bool)
                                    -> Result<()> {
+        debug!("Promote/demote for group: {}, channel: {}",
+               group_id, channel);
```
christophermaier: It would be useful to include the concrete operation being performed here (either promote or demote), rather than being ambiguous.

components/builder-api-client/src/builder.rs
```diff
@@ -399,28 +409,27 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
-                   if promote { "promote" } else { "demote" },
-                   res.status);
-            return Err(err_from_response(res));
-        }
 
-        Ok(())
+        let resp = self.add_authz(self.0.post(&url), token)
+                       .json(&body)
+                       .send()?;
+        debug!("Status: {:?}", resp.status());
```
christophermaier: Going back to my comment above re: structured logging, will logging _just_ the status here, absent any other information about a specific request, be useful?

components/builder-api-client/src/builder.rs
```diff
@@ -473,24 +482,24 @@ impl BuilderAPIProvider for BuilderAPIClient {
                             key: &str,
                             secret: &WrappedSealedBox)
                             -> Result<()> {
+        debug!("Creating origin secret: {}, {}", origin, key);
```
christophermaier: Seeing "key" in a logging statement (in a function named `create_origin_secret`, no less) got me concerned about leaking confidential information in logs.

It looks like "key" in this context refers to a hash map key though, so that's fine for logging. However, it might be good to change the argument name from `key` to something a bit more descriptive (and less concerning!) like `name`.

components/builder-api-client/src/builder.rs
```diff
@@ -499,15 +508,18 @@ impl BuilderAPIProvider for BuilderAPIClient {
     ///
     /// * Remote Builder is not available
     fn delete_origin_secret(&self, origin: &str, token: &str, key: &str) -> Result<()> {
+        debug!("Deleting origin secret: {}, {}", origin, key);
```
christophermaier: Same suggestion regarding the name "key".

components/builder-api-client/src/builder.rs
christophermaier: Thinking a bit more, I get why you might want to get the negated scenario out of the way first. If you know it's bad, just get rid of it now; otherwise move on.

I was messing around a bit, and a nice alternative might be a macro like this: 

```rust
macro_rules! fail_if_not {
    ($resp:expr, $status:expr) => {
        if $resp.status() != $status {
            return Err(err_from_response($resp));
        }
    };
}
```

This is nice because you can recast the above code like this:

```rust
fail_if_not!(resp, StatusCode::OK);
Ok(())
```
When the non-error case is more complex, the benefits are greater.

This lets you quickly get the errors out of the way, and lets you do so in a pretty descriptive and low-overhead way. You don't need to add an extra level of indentation to the code with an `if`/`else` block. Because it's a macro, you can only use this in functions that return a `Result`, but that's clearly not an issue here.

I played around with a function formulation, but then you have to concern yourself with return types and potentially reassigning the response, which can be done and is valid, but adds a bit of unnecessary ceremony and line noise in my opinion.

components/builder-api-client/src/builder.rs
```diff
@@ -902,8 +948,9 @@ impl BuilderAPIProvider for BuilderAPIClient {
         let ident = pa.ident()?;
         let target = pa.target()?;
 
-        let mut file =
-            File::open(&pa.path).map_err(|e| Error::PackageReadError(pa.path.clone(), e))?;
+        debug!("Uploading package {}, target {}", ident, target);
```
christophermaier: I am _soooooo_ looking forward to a day when target and ident can be part of a single unified identifier.

components/builder-api-client/src/builder.rs
christophermaier: And because it's a macro, you could also modify it to accept multiple status codes, a pattern I see in at least one other place in this module.

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
-use std::{path::Path,
+use std::{fs::{self,
+               File},
+          io::{Read,
+               Write},
+          iter::FromIterator,
+          path::{Path,
+                 PathBuf},
+          str::FromStr,
           time::Duration};
```
christophermaier: Here's another newline we can remove, allowing `rustfmt` to keep everything sorted for us.

components/builder-api-client/src/builder.rs
chefsalim: Interesting idea... I didn't realize you could add a target to the trace command... adding that could indeed be very useful... lets start with that and we can discuss the future state of logging separately

components/builder-api-client/src/builder.rs
chefsalim: Nice... a set of macros for these common patterns could be useful..

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
chefsalim: Cool... I like it!

components/builder-api-client/src/builder.rs
chefsalim: Ha, yes...

components/builder-api-client/src/builder.rs
chefsalim: It's semi useful atm... I did some work to make the logging a bit cleaner and consistent but that definitely hasn't been the main thrust of this PR...

components/builder-api-client/src/builder.rs
chefsalim: OK

components/builder-api-client/src/builder.rs
chefsalim: That's part of "v1" right? :)

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+fn cert_from_file(file_path: &PathBuf) -> Result<Certificate> {
+    let mut buf = Vec::new();
+    File::open(file_path)?.read_to_end(&mut buf)?;
+
+    let ext = file_path.extension().unwrap(); // unwrap Ok
+    if ext == "pem" {
+        Certificate::from_pem(&buf).map_err(Error::ReqwestError)
+    } else {
+        Certificate::from_der(&buf).map_err(Error::ReqwestError)
+    }
```
christophermaier: To reduce the duplication here, you could do something like this:

```rust
let fun = if ext == "pem" {
    Certificate::from_pem
} else {
    Certificate::from_der
};

fun(&buf).map_err(Error::ReqwestError)
```

components/sup-protocol/src/lib.rs
```diff
@@ -74,7 +74,7 @@ lazy_static! {
 
 /// Generate a new secret key used for authenticating clients to the `CtlGateway`.
 pub fn generate_secret_key(out: &mut String) {
-    let mut rng = rand::rngs::OsRng::new().unwrap();
+    let mut rng = rand::rngs::OsRng;
```
christophermaier: Why did this change?

components/sup/src/manager.rs
```diff
@@ -1785,19 +1785,38 @@ fn track_memory_stats() {
...
+                .set(jemalloc_ctl::stats::resident::mib().unwrap()
+                                                         .read()
+                                                         .unwrap()
+                                                         .to_i64());
     MEMORY_STATS.with_label_values(&["retained"])
-                .set(jemalloc_ctl::stats::retained().unwrap().to_i64());
+                .set(jemalloc_ctl::stats::retained::mib().unwrap()
+                                                         .read()
+                                                         .unwrap()
+                                                         .to_i64());
```
christophermaier: Why did all these change?

I'm a bit uneasy with all the unwraps... can we be certain they're safe to do?

At the _very_ least, they should be `expect` calls, so we can get additional context if one of them fires.

components/builder-api-client/src/builder.rs
christophermaier: It'd be real nice!

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
     }
 
-    Ok(conn.build())
+    if let Ok(entries) = fs::read_dir(cache_ssl_path(fs_root_path)) {
+        for entry in entries {
+            let entry = entry?;
+            let path = entry.path();
+
+            if let Some(ext) = path.extension() {
+                if path.is_file() && ((ext == "pem") || (ext == "der")) {
```
christophermaier: What happens if someone has symlinked their certs from somewhere else?

baumanj:
Does this address https://github.com/habitat-sh/habitat/issues/5597 and the need for the `export SSL_CERT_FILE=/usr/local/etc/openssl/cert.pem` macOS workaround generally?

christophermaier:
@baumanj That's my suspicion and hope.

components/sup/src/manager.rs
chefsalim: The version of jemalloc got bumped in the dep chain... in general seems fine to pick up an updated version, the only change is that the stats is under a mib namespace and needed another unwrap than what was already there.

components/sup-protocol/src/lib.rs
chefsalim: Updated rand deprecates the old way - the new one seems better as there is no unwrap

danielcbright:
This looks great to me, minor nag though is I don't see any documentation/readme/examples on how to use your own cacert bundle both in-and-out of studio, am I missing it somewhere?

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+/// 2. If there is no 'core/cacerts packages, then a copy of `cacert.pem` will be written in an SSL
+///    cache directory (by default `/hab/cache/ssl` for a root user and `$HOME/.hab/cache/ssl` for
+///    a non-root user) and this will be used. The contents of this file will be inlined in this
+///    crate at build time as a fallback, which means that if the program using this code is
+///    operating in a minimal environment which may not contain any system certificates, it can
+///    still operate.
+/// 3. Other certs files (for example self-signed certs) that are found in the SSL cache directory
+///    will also get loaded into the root certs list. Both PEM and DER formats are supported (the
+///    extensions should be '.pem' or '.der' respectively)
+fn certificates(fs_root_path: Option<&Path>) -> Result<Vec<Certificate>> {
```
baumanj: Since it's used 3 times, it might be nice to add a binding:
```rust
    let cert_cache_dir = cache_ssl_path(fs_root_path);
```

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+        certificates.push(cert);
+    } else {
+        debug!("No installed cacerts package found");
+
+        let cached_certs = cache_ssl_path(fs_root_path).join("cert.pem");
+        if !cached_certs.exists() {
+            fs::create_dir_all(cache_ssl_path(fs_root_path))?;
+            debug!("Creating cached cacert.pem: {}", cached_certs.display());
+            let mut file = File::create(&cached_certs)?;
+            file.write_all(CACERT_PEM.as_bytes())?;
```
baumanj: These two lines can be simplified to:
```rust
            fs::write(cached_certs, CACERT_PEM)?;
```
Or, if you're into even more brevity combined with the dir creation:
```rust
            fs::create_dir_all(&cert_cache_dir).and_then(|_| fs::write(cached_certs, CACERT_PEM))?;
```

components/http-client/src/api_client.rs
baumanj: This function is fairly complex. It would be good to get some test coverage on it. For that, though, it would probably be helpful to break it up into its 3 logical segments described in the doc comment.

components/http-client/src/api_client.rs
christophermaier: I kinda like the idea of ignoring the extension and just trying both conversions in sequence. I think that would simplify quite a bit of code.

components/builder-api-client/src/artifactory.rs
```diff
@@ -90,37 +95,42 @@ impl ArtifactoryClient {
...
+                     let size = cl.parse::<u64>().unwrap_or_else(|_| 0);
```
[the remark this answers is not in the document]
chefsalim: Will fix

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,120 @@ impl ApiClient {
...
+///    will also get loaded into the root certs list. Both PEM and DER formats are supported (the
+///    extensions should be '.pem' or '.der' respectively)
+fn certificates(fs_root_path: Option<&Path>) -> Result<Vec<Certificate>> {
+    let mut certificates = Vec::new();
+
+    // MacOS is not yet fully consistent with other platforms,
+    // as it cannot handle PEM files with multiple certs.
+    // We can enable this when the following issue is resolved:
+    // https://github.com/sfackler/rust-native-tls/issues/132
+    if cfg!(not(target_os = "macos")) {
```
baumanj: The macro form is evaluated at runtime, but we can use the attribute form to exclude this code at compile time
```suggestion
    #[cfg(not(target_os = "macos"))]
    {
```

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
chefsalim: Between these two approaches, I'm inclined to go with the extension trait. It solves a couple of different issues - the error handling, as well as abstracting out things like headers in a nicer way. It also lets us return error for specific status codes which the error_for_status does not. Also the From<reqwest::Error> error logic doesn't seem right when moving err_from_response there, since we would have to handle both error from status case as well as a regular reqwest error case.

components/builder-api-client/src/builder.rs
chefsalim: Addressed by other fixes

components/builder-api-client/src/builder.rs
chefsalim: Addressed by other fixes

components/builder-api-client/src/builder.rs
```diff
@@ -173,57 +170,60 @@ impl BuilderAPIClient {
...
                  Some(mut progress) => {
-                     let size: u64 = res.headers
-                                        .get::<hyper::header::ContentLength>()
-                                        .map_or(0, |v| **v);
+                     let cl = resp.headers()
+                                  .get(CONTENT_LENGTH)
+                                  .expect("Content length missing")
+                                  .to_str()
+                                  .expect("Content length invalid");
+                     let size = cl.parse::<u64>().unwrap_or_else(|_| 0);
```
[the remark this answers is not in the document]
chefsalim: Fixed

components/http-client/src/api_client.rs
chefsalim: OK

components/builder-api-client/src/builder.rs
chefsalim: OK

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
     /// Creates and returns a new `ApiClient` instance.
     ///
-    /// # Errors
-    ///
-    /// * If the underlying Hyper client cannot be created
-    /// * If a suitable SSL context cannot be established
-    /// * If an HTTP proxy cannot be correctly setup
-    /// * If a `User-Agent` HTTP header string cannot be constructed
+    /// Builds a new Reqwest HTTP client with appropriate SSL configuration and HTTP/HTTPS proxy
+    /// support.
```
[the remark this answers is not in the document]
chefsalim: OK I'll add them back in

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
-///
-/// For example:
-///
-/// ```text
-/// hab/0.6.0/20160606153031 (x86_64-darwin; 14.5.0)
-/// ```
-///
-/// # Errors
-///
-/// * If system information cannot be obtained via `uname`
```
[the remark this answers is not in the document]
chefsalim: I'll add back in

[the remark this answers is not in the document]
chefsalim: Will add back

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
chefsalim: Fixed

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+/// 3. Other certs files (for example self-signed certs) that are found in the SSL cache directory
+///    will also get loaded into the root certs list. Both PEM and DER formats are supported (the
+///    extensions should be '.pem' or '.der' respectively)
+fn certificates(fs_root_path: Option<&Path>) -> Result<Vec<Certificate>> {
+    let mut certificates = Vec::new();
+    let cacerts_ident = PackageIdent::from_str(CACERTS_PKG_IDENT)?;
+
+    if let Ok(pkg_install) = PackageInstall::load(&cacerts_ident, fs_root_path) {
+        let pkg_certs = pkg_install.installed_path().join("ssl/cert.pem");
+        debug!("Found installed certs: {}", pkg_certs.display());
```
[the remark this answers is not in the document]
chefsalim: I'll update the debug to be clearer - that line is to output debug that an installed core/cacerts Hab package was found, which is true if PackageInstall::load succeeds

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+        let cert = cert_from_file(&pkg_certs)?;
+        certificates.push(cert);
+    } else {
+        debug!("No installed cacerts package found");
```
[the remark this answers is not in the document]
chefsalim: This is not meant to be text for end users.. will look at updating

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+
+        let cached_certs = cache_ssl_path(fs_root_path).join("cert.pem");
+        if !cached_certs.exists() {
+            fs::create_dir_all(cache_ssl_path(fs_root_path))?;
+            debug!("Creating cached cacert.pem: {}", cached_certs.display());
```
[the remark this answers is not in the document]
chefsalim: It would be nice to expose cert info as part of the regular UI output (which is where I think you're going with this) but non-trivial enough that I think its out of scope for this PR

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
-    Ok(conn.build())
+    if let Ok(entries) = fs::read_dir(cache_ssl_path(fs_root_path)) {
+        for entry in entries {
+            let entry = entry?;
+            let path = entry.path();
+
+            if let Some(ext) = path.extension() {
+                if path.is_file() && ((ext == "pem") || (ext == "der")) {
+                    debug!("Found cached cert: {}", path.display());
+                    let cert = cert_from_file(&path)?;
```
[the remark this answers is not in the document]
chefsalim: Well thats a great question - we can certainly log the error and move on. That could also "work" for the OSX case where we currently are not able to load the PEM file atm.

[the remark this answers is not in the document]
baumanj: My suggestion (to minimize painting ourselves into a corner) would be to fail the whole operation currently. Later, we can always add a flag (or configuration item) to continue on error, but we'd have a much harder time changing it in the other direction.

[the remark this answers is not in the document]
chefsalim: Gracefully handling errors and not failing the entire operation allows us to remove the check for PEM/DER entirely, which is a definite simplification, and also better as we can now handle PEM formatted CRT files as well.

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+    }
+
+    Ok(certificates)
+}
+
+fn cert_from_file(file_path: &PathBuf) -> Result<Certificate> {
+    let mut buf = Vec::new();
+    File::open(file_path)?.read_to_end(&mut buf)?;
+
+    let ext = file_path.extension().unwrap(); // unwrap Ok
```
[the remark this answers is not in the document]
chefsalim: Happily, needing to look at the extension is no longer needed

components/http-client/src/api_client.rs
chefsalim: Yep, going with Jon's suggestion on this...

components/http-client/src/api_client.rs
chefsalim: Yep, agreed

components/builder-api-client/src/artifactory.rs
```diff
@@ -351,30 +361,27 @@ impl BuilderAPIProvider for ArtifactoryClient {
...
-                .send()
+            let reader = TeeReader::new(file, progress);
+            let body = Body::sized(reader, file_size);
+            self.add_authz(self.0.put(&path), token).body(body).send()?
         } else {
-            self.add_authz(self.0.put(&path), token)
-                .body(Body::SizedBody(&mut file, file_size))
-                .send()
+            let body = Body::sized(file, file_size);
+            self.add_authz(self.0.put(&path), token).body(body).send()?
```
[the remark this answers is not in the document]
chefsalim: Good catch on the duplicated code - will fix.

Re: the result/response - I'm updating so that Result = "res" and Response = "resp" in accordance with the Reqwest samples so we are consistent in the usage.

components/http-client/src/api_client.rs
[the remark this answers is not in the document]
baumanj: Seems like we could try all the available formats and *then*, if they all gave errors, fail the operation, no?

components/builder-api-client/src/artifactory.rs
```diff
@@ -514,24 +516,20 @@ impl BuilderAPIProvider for ArtifactoryClient {
 
         let path = self.url_path_for_package(&ident, target);
 
-        let mut reader: Box<dyn Read> = if let Some(mut progress) = progress {
+        let reader: Box<dyn Read + Send> = if let Some(mut progress) = progress {
```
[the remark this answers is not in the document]
chefsalim: Yep will fix

components/builder-api-client/src/builder.rs
```diff
@@ -173,57 +170,60 @@ impl BuilderAPIClient {
...
     }
 
-    fn maybe_add_authz<'a>(&'a self,
-                           rb: RequestBuilder<'a>,
-                           token: Option<&str>)
-                           -> RequestBuilder<'_> {
+    fn maybe_add_authz(&self, rb: RequestBuilder, token: Option<&str>) -> RequestBuilder {
         if token.is_some() {
-            rb.header(Authorization(Bearer { token: token.unwrap().to_string(), }))
+            rb.bearer_auth(token.unwrap().to_string())
```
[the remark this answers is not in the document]
chefsalim: Yes indeed

components/builder-api-client/src/builder.rs
```diff
@@ -173,57 +170,60 @@ impl BuilderAPIClient {
...
         } else {
             rb
         }
     }
 
-    fn add_authz<'a>(&'a self, rb: RequestBuilder<'a>, token: &str) -> RequestBuilder<'_> {
-        rb.header(Authorization(Bearer { token: token.to_string(), }))
+    fn add_authz(&self, rb: RequestBuilder, token: &str) -> RequestBuilder {
+        rb.bearer_auth(token.to_string())
```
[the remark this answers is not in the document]
chefsalim: Agreed, I like that better as well...

components/builder-api-client/src/builder.rs
```diff
@@ -173,57 +170,60 @@ impl BuilderAPIClient {
...
-
-        if res.status != hyper::status::StatusCode::Ok {
-            return Err(err_from_response(res));
+    fn download(&self,
+                rb: RequestBuilder,
+                dst_path: &Path,
+                token: Option<&str>,
+                progress: Option<<BuilderAPIClient as BuilderAPIProvider>::Progress>)
+                -> Result<PathBuf> {
+        debug!("Downloading file to path: {:?}", dst_path);
```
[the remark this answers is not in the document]
chefsalim: OK

components/builder-api-client/src/builder.rs
```diff
@@ -232,19 +232,25 @@ impl BuilderAPIClient {
...
         let package_results = serde_json::from_str(&encoded)?;
-        match res.status {
-            StatusCode::Ok => Ok((package_results, false)),
-            StatusCode::PartialContent => Ok((package_results, true)),
-            _ => Err(err_from_response(res)),
+
+        if resp.status() == StatusCode::OK || resp.status() == StatusCode::PARTIAL_CONTENT {
+            Ok((package_results, resp.status() == StatusCode::PARTIAL_CONTENT))
+        } else {
+            Err(err_from_response(resp))
```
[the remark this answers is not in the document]
chefsalim: The updated code has some changes..

components/builder-api-client/src/builder.rs
```diff
@@ -793,10 +824,8 @@ impl BuilderAPIProvider for BuilderAPIClient {
                                     u.set_query(Some(&format!("target={}", target)))
                                 });
 
-        match self.download(req_builder, dst_path.as_ref(), token, progress) {
-            Ok(file) => Ok(PackageArchive::new(file)),
-            Err(e) => Err(e),
-        }
+        let file = self.download(req_builder, dst_path.as_ref(), token, progress)?;
+        Ok(PackageArchive::new(file))
```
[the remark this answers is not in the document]
chefsalim: Re: error_from_status - as mentioned in a different comment, there are implications of compatibility and higher level changes, so I think it will be better in a future PR rather than undertaking this now

components/builder-api-client/src/builder.rs
```diff
@@ -904,22 +945,23 @@ impl BuilderAPIProvider for BuilderAPIClient {
         };
         debug!("Reading from {}", &pa.path.display());
 
-        let mut reader: Box<dyn Read> = if let Some(mut progress) = progress {
+        let reader: Box<dyn Read + Send> = if let Some(mut progress) = progress {
```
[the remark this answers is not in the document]
chefsalim: Fixed

components/builder-api-client/src/builder.rs
```diff
@@ -998,25 +1044,27 @@ impl BuilderAPIProvider for BuilderAPIClient {
                        channel: &ChannelIdent,
                        token: &str)
                        -> Result<()> {
+        debug!("Promoting package {}, target {}", ident, target);
```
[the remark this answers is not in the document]
chefsalim: Fixed

components/core/Cargo.toml
```diff
@@ -49,8 +49,7 @@ winapi = { version = "*", features = ["dpapi", "ioapiset", "namedpipeapi", "user
 windows-acl = "*"
 
 [dev-dependencies]
-# Pending upgrade activities in https://github.com/habitat-sh/core/issues/72
-hyper = "0.10"
+hyper = "*"
```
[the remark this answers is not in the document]
chefsalim: Good question - looks like we can safely remove it

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
-    /// using an HTTP proxy.
-    inner: HyperClient,
-    /// Proxy information, if a proxy is being used.
-    proxy: Option<ProxyInfo>,
-    /// The URL scheme of the endpoint.
-    target_scheme: String,
-    /// The `User-Agent` header string to use for HTTP calls.
-    user_agent_header: UserAgent,
+    /// An instance of a `reqwest::Client`
+    inner: ReqwestClient,
```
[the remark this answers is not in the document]
chefsalim: Seems reasonable - will update

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
     {
-        let endpoint = endpoint.into_url().map_err(Error::UrlParseError)?;
-        Ok(ApiClient { inner: new_hyper_client(&endpoint, fs_root_path)?,
-                       proxy: proxy_unless_domain_exempted(Some(&endpoint))?,
-                       target_scheme: endpoint.scheme().to_string(),
-                       endpoint,
-                       user_agent_header: user_agent(product, version)? })
+        let endpoint = endpoint.into_url().map_err(Error::ReqwestError)?;
+
+        let timeout_in_secs = match env::var("HAB_CLIENT_SOCKET_TIMEOUT") {
```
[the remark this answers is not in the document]
chefsalim: For this change, I'd prefer to leave the existing env vars the same in order to preserve as much backward compatibility as possible

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
+            Ok(t) => {
+                match t.parse::<u64>() {
+                    Ok(n) => n,
+                    Err(_) => CLIENT_SOCKET_RW_TIMEOUT_SEC,
+                }
+            }
+            Err(_) => CLIENT_SOCKET_RW_TIMEOUT_SEC,
+        };
```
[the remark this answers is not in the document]
chefsalim: Thanks, I think I'll tackle env switchover separately and not refactor as part of this PR

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
+        debug!("Client socket timeout: {} secs", timeout_in_secs);
+
+        let skip_cert_verify = env::var("HAB_SSL_CERT_VERIFY_NONE").is_ok();
+        debug!("Skip cert verification: {}", skip_cert_verify);
+
+        let header_values = vec![(USER_AGENT, user_agent(product, version)?)];
+        let headers = HeaderMap::from_iter(header_values.into_iter());
```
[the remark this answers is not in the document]
chefsalim: Done

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
+
+        let mut client = client_with_proxy(&endpoint).default_headers(headers)
+                                                     .timeout(Duration::from_secs(timeout_in_secs))
+                                                     .danger_accept_invalid_certs(skip_cert_verify);
+
+        let certs = certificates(fs_root_path)?;
+        for cert in certs {
```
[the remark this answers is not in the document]
chefsalim: OK

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
-fn new_hyper_client(url: &Url, fs_root_path: Option<&Path>) -> Result<HyperClient> {
-    let connector = ssl_connector(fs_root_path)?;
-    let ssl_client = OpensslClient::from(connector);
-
-    let timeout_in_secs = match env::var("HAB_CLIENT_SOCKET_TIMEOUT") {
-        Ok(t) => {
-            match t.parse::<u64>() {
-                Ok(n) => n,
-                Err(_) => CLIENT_SOCKET_RW_TIMEOUT_SEC,
+fn client_with_proxy(url: &Url) -> ClientBuilder {
```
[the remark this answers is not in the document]
chefsalim: Agreed, will update

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+            }
+        }
+    }
+
+    Ok(certificates)
+}
+
+fn cert_from_file(file_path: &PathBuf) -> Result<Certificate> {
+    let mut buf = Vec::new();
+    File::open(file_path)?.read_to_end(&mut buf)?;
```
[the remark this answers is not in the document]
chefsalim: Done

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+                    certificates.push(cert);
+                }
+            }
+        }
+    }
+
+    Ok(certificates)
+}
+
+fn cert_from_file(file_path: &PathBuf) -> Result<Certificate> {
```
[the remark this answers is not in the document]
chefsalim: Fixed

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+///    still operate.
+/// 3. Other certs files (for example self-signed certs) that are found in the SSL cache directory
+///    will also get loaded into the root certs list. Both PEM and DER formats are supported (the
+///    extensions should be '.pem' or '.der' respectively)
+fn certificates(fs_root_path: Option<&Path>) -> Result<Vec<Certificate>> {
+    let mut certificates = Vec::new();
+    let cacerts_ident = PackageIdent::from_str(CACERTS_PKG_IDENT)?;
+
+    if let Ok(pkg_install) = PackageInstall::load(&cacerts_ident, fs_root_path) {
+        let pkg_certs = pkg_install.installed_path().join("ssl/cert.pem");
```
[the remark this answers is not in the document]
chefsalim: We seem to use the plural by default as in the "cacerts" package... leaving it as is for now

components/http-client/src/api_client.rs
```diff
@@ -191,121 +188,113 @@ impl ApiClient {
...
+        debug!("Found installed certs: {}", pkg_certs.display());
+        let cert = cert_from_file(&pkg_certs)?;
+        certificates.push(cert);
```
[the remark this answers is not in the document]
chefsalim: Updated

components/http-client/src/api_client.rs
chefsalim: Done

components/http-client/src/api_client.rs
chefsalim: Done

components/http-client/src/api_client.rs
chefsalim: Simplified the function. For testing I'll open a separate issue to add cert cases to our e2e tests.

components/http-client/src/api_client.rs
chefsalim: Done

components/http-client/src/api_client.rs
```diff
@@ -1,82 +1,96 @@
...
+                match t.parse::<u64>() {
+                    Ok(n) => n,
+                    Err(_) => CLIENT_SOCKET_RW_TIMEOUT_SEC,
+                }
+            }
+            Err(_) => CLIENT_SOCKET_RW_TIMEOUT_SEC,
+        };
+        debug!("Client socket timeout: {} secs", timeout_in_secs);
+
+        let skip_cert_verify = env::var("HAB_SSL_CERT_VERIFY_NONE").is_ok();
```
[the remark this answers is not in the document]
chefsalim: As mentioned elsewhere, I'll save the env update for a separate PR

chefsalim:
@davidMcneil re: your docs comment, I'm going to open a separate issue for updating the docs with the new behavior.  There will also be a separate issue to inject cache cert files into the studio similar to how we are injecting keys.

PR #6763, opened 2019-07-26
hab: don't start analytics thread if analytics is disabled

stevendanna:
This will introduce a few calls to stat(2) into the startup of Habitat
in exchange for not starting the analytics thread.

The actual motivation here is that this thread is involved in the
conditions that appear to be triggering:

https://github.com/habitat-sh/habitat/issues/6252

While we should fix that bug, not starting a thread we don't need
seems like an easy way to lower the impact of the bug quickly.

Typically, I can reproduce 6252 in between 10k-250k invocations of
hab. With this fix in place, I've been able to run 5M+ invocations of
hab without issue.

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -109,7 +109,11 @@ fn main() {
     env_logger::init();
     let mut ui = UI::default_with_env();
     let flags = FeatureFlag::from_env(&mut ui);
-    thread::spawn(analytics::instrument_subcommand);
+
+    if analytics::analytics_enabled() {
+        thread::spawn(analytics::instrument_subcommand);
+    }
+
     if let Err(e) = start(&mut ui, flags) {
         ui.fatal(e).unwrap();
         std::process::exit(1)
--- a/components/hab/src/analytics.rs
+++ b/components/hab/src/analytics.rs
@@ -419,7 +419,7 @@ pub fn is_opted_in<T>(analytics_path: T) -> Option<bool>
 }
 
 /// Returns true if analytics are enabled and false otherwise.
-fn analytics_enabled() -> bool {
+pub fn analytics_enabled() -> bool {
     match is_opted_in(hcore::fs::cache_analytics_path(None::<String>)) {
         // If the value is explicitly true or false, return the unwrapped value
         Some(val) => val,


stevendanna:
Alternatively: https://github.com/habitat-sh/habitat/pull/6764

components/hab/src/main.rs
```diff
@@ -109,7 +109,11 @@ fn main() {
     env_logger::init();
     let mut ui = UI::default_with_env();
     let flags = FeatureFlag::from_env(&mut ui);
-    thread::spawn(analytics::instrument_subcommand);
+
+    if analytics::analytics_enabled() {
+        thread::spawn(analytics::instrument_subcommand);
+    }
```
baumanj: Seems like if we're doing the check on the outside we shouldn't [*also* do the check in the inside](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/analytics.rs#L235-L238).

stevendanna: Happy to make that change, I left it to protect against callers who called it without checking.  But since calls to that function are minimal (this is the only one) I suppose that isn't much of a risk.

stevendanna:
> The idea that it's merely spawning this thread that is causing #6252 seems hard to understand. I could see that its's a timing/alignment issue, but it still seems pretty wild.

My very unproven theory is that there is a bookkeeping error in musl. Note that this code (purposefully according to the initial commit) don't wait on this thread, so we exit with this thread still running.  It has been a while since I looked at this, but there is code in musl specifically around tracking whether we are multi-threaded or not right around the loop that I think we are stuck in.

baumanj:
> My very unproven theory is that there is a bookkeeping error in musl. Note that this code (purposefully according to the initial commit) don't wait on this thread, so we exit with this thread still running.

And quite a few others! If that's the problem, I'm surprised all those others don't have an effect.

Out of curiosity, does changing this thread from using [`std::thread::spawn`](http://doc.rust-lang.org/1.36.0/std/thread/fn.spawn.html) to [`std::thread::Builder::spawn`](http://doc.rust-lang.org/1.33.0/std/thread/struct.Builder.html#method.spawn) (and giving it a name, perhaps) make any difference? I only ask because that's what we do for *most* of our threads.

christophermaier:
@baumanj The key difference is that this is for the `hab` binary, and not the `hab-sup` binary... only the former is built using `musl`. This is probably the only thread that the `hab` binary starts.

chefsalim:
Superceded by https://github.com/habitat-sh/habitat/pull/6764

PR #6764, opened 2019-07-26
hab: remove analytics

stevendanna:
This removes analytics from hab. Discussions with the Habitat team
indicate that this information is not well used and that the removal
of the http_client dependency would be welcome. This is also likely
one less code change that would be required by any community
distributions of Habitat.

However, our real motives are more sinister.

Namely, the thread started to instrument_subcommands is involved in
the conditions that appear to be triggering:

https://github.com/habitat-sh/habitat/issues/6252

Removing this thread completely also substantially reduces the impact
of that bug.

A less intrusive alternative is to disable the instrument_subcommands
thread in some cases. This approach has also been submitted as a PR
for discussion.


stevendanna:
Alternatively: https://github.com/habitat-sh/habitat/pull/6763

PR #6767, opened 2019-07-29
Cleanup expeditor config

scotthain:

- We should only have 1 subscription section.
- We want to ensure the pipeline triggers post merge to master
- We no longer need a version constraint, as the `release_branch` defaults to the master branch. This was used during development to use a separate release branch a while ago and is no longer needed.
- We also want to ensure that the changelog and version get bumped correctly before we do the merge and kick off the pipeline, so moved those to a subscription.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -56,6 +56,7 @@ subscriptions:
           - 'Expeditor: Exclude from Changelog'
           - 'Expeditor: Skip All'
       - trigger_pipeline:release_habitat:
+          only_if: built_in:bump_version
           post_commit: true
 
   - workload: buildkite_build_passed:{{agent_id}}:release_habitat:*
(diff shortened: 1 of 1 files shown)


christophermaier:
still looks :+1:

.expeditor/config.yml
```diff
@@ -61,9 +41,22 @@ staging_areas:
...
-      - trigger_pipeline:release_habitat
+      - built_in:bump_version:
+          ignore_labels:
+          - 'Expeditor: Skip Version Bump'
+          - 'Expeditor: Skip All'
+      - built_in:update_changelog:
+          ignore_labels:
+          - 'Expeditor: Exclude from Changelog'
+          - 'Expeditor: Skip All'
+      - trigger_pipeline:release_habitat:
```
tduffield: Do you want to put a filter on this action to only trigger if you bump the version? 

```
- trigger_pipeline:release_habitat:
    only_if: built_in:bump_version
    post_commit: true
```

Issue #6634, opened 2019-06-07
Extract `hab pkg exec` behavior to reusable function


smacfarlane:
When `hab pkg exec` runs a package, it carefully sets up the environment that the package expects to have before exec'ing the command. `hab studio enter` also needs this behavior as it is essentially execing the studio. We should look at refactoring the code [here](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/pkg/exec.rs#L19-L35) to its own function.  

I suspect we'll want to add a `become_command` method on PackageInstall to set up the environment for us, but there may be other approaches that will be more appropriate.

smacfarlane:
I took a swing at this but had to set it down due to time contraints.  

My initial approach was to hang a `command_with_environment(&str cmd_name) -> std::process::Command` method off of `PackageInstall`. However, `process::become_command` expects a `PathBuf` and is used quite widely throughout the code base.  I wasn't sure if it would be better to have different code paths in studio enter for a `Command` vs `PathBuf` or if all PathBufs should be turned into `Command`s before executing them, or perhaps there is a different approach I hadn't considered.  

It also wasn't as testable as I had hoped because Command doesn't (as far as I was able to determine) give you any mechanism to make assertions about changes to the environment it would execute in, set via `env(key,value)`.

PR #6772, opened 2019-07-30
Ensure studio is installed

### LINKED ISSUE #6771: 0.83.0 hab studio enter no longer auto-installs core/hab-studio (chroot studio)
### Scope 
This bug only affects the 0.83.0 Linux chroot studio.  Windows, MacOS and Linux Docker studios appear to be unaffected.  

### Workaround 
As a workaround, you can run `hab pkg install core/hab-studio/0.83.0` on any Linux machines that require the chroot studio before executing those commands. 

### Details
With the 0.83.0 release, there was a small change to how the studio is invoked on Linux in order to enable us to make the hab-backline package an explicit dependency of the studio. #6549. 

Unfortunately, the process for entering a studio attempts to read the RUNTIME_ENVIRONMENT metadata [[link](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/studio/enter.rs#L122)] before the package is installed [[link](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/studio/enter.rs#L135)].

This appears to have been missed in the release process because we always explicitly install the studio [link](https://github.com/habitat-sh/habitat/blob/master/.buildkite/scripts/shared.sh#L123-L126). 

### Tasks:
 - [ ] Ensure studio is installed before attempting to read package metadata
 - [ ] Add end-to-end test to verify studio can be invoked from a clean system

smacfarlane:
Closes #6771 

Previously, `studio enter` would ensure that the appropriate `core/hab-studio` package was installed before attempting to invoke it.  With the change in #6549, we attempt to read the metadata from the studio package, assuming it is already installed which isn't the case after installing a new release of Habitat. 

This corrects that by ensuring the studio package is installed before attempting to read the metadata. 

### DIFF
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -133,7 +133,7 @@ mod inner {
                     }
                     debug!("Running: {}", display_args);
 
-                    command 
+                    command
                 }
             };
 
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -114,6 +114,7 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
+                    let command = exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?
                     // This is a duplicate of the code in `hab pkg exec` and
                     // should be refactored as part of or after:
                     // https://github.com/habitat-sh/habitat/issues/6633
@@ -132,7 +133,7 @@ mod inner {
                     }
                     debug!("Running: {}", display_args);
 
-                    exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?
+                    command 
                 }
             };
 
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -114,7 +114,7 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
-                    let command = exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?
+                    let command = exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?;
                     // This is a duplicate of the code in `hab pkg exec` and
                     // should be refactored as part of or after:
                     // https://github.com/habitat-sh/habitat/issues/6633


components/hab/src/command/studio/enter.rs
```diff
@@ -114,6 +114,7 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
+                    let command = exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?
```
smacfarlane: This continues the trend in this code path of "spooky action at a distance".  This not only gets the command to install (which returns a PathBuf rather than a Cmd), but it also manages the installation of the studio package. 

This should be the minimum change necessary to fix the underlying issue (still testing to verify), but depending on the timing of shipping this fix in a release  I'd like to refactor this to clean up the responsibilities of these functions to make it clear how/when a studio package is installed.

components/hab/src/command/studio/enter.rs
```diff
@@ -114,6 +114,7 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
+                    let command = exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident)?;
                     // This is a duplicate of the code in `hab pkg exec` and
                     // should be refactored as part of or after:
                     // https://github.com/habitat-sh/habitat/issues/6633
```
christophermaier: Is it maybe time for this refactoring?

smacfarlane: Yes :joy_cat:    I started down this path a couple weeks ago, but had to set it aside due to time constraints.  I've updated my notes on the tracking issue #6634 

In addition to the above refactor, I think we'd possibly want to pull the "ensure package is installed" code out of `command_from_min_pkg` so that we can isolate the responsibilities of that code. That function appears to only be used in a few places through the code base, but it is used by the launcher and supervisor, in addition to studio enter.

scotthain:
This will resolve https://github.com/habitat-sh/habitat/pull/6775

components/hab/src/command/studio/enter.rs
baumanj: > I think we'd possibly want to pull the "ensure package is installed" code out of `command_from_min_pkg`

Strong agree.

Replacing the `PackageInstall::load` on L121 with something like `::install_and_load` makes this more intuitive, I think. In fact, how often do we really want `PackageInstall::load` to fail if the package isn't installed versus trying the install? At a quick glance, it appears to be nearly always called with `?`. This might be the time to do a little audit of the callers and see if we have more latent bugs which are similar to this.

smacfarlane:
@baumanj   Let's go with this as-is so at minimum we have a fix in the next release.    I can swing back to this once the core-plans refresh is complete, or someone else can pick this work up in the mean time.

components/hab/src/command/studio/enter.rs
smacfarlane: I was thinking, based on how `PackageIdent` and `PackageInstall` are used here, that it would make sense to have method on `PackageIdent` that  returns a `PackageInstall` and handles the installation if needed.  

Perhaps there's a more Rusty way to handle it, but given that `PackageInstall::load` takes a PackageIdent, we probably have this pattern throughout the code and it, maybe naively, would make sense to have that behavior baked into the transformation of ident into install.

baumanj: If what we have is a `PackageIdent`, and want is a `PackageInstall`, the most rustic approaches to me seem to be adding `PackageInstall::from_ident() -> Self` or `PackageIdent::install(&self) -> PackageInstall`. I like the way the latter reads, but since `PackageInstall` has non-`pub` fields, you'll probably need to call on `PackageInstall` to create it for you. In that case implementing `PackageIdent::install` by having it call `PackageInstall::from_ident` might be appropriate.

But now I see that we also have an `InstallTask` type, which already has a `with_ident` method that returns a `Result<PackageInstall>`, so perhaps that's what we should be using. It's currently private, but it probably makes sense to expose at least part of that interface so we have a cleaner way to go from `PackageIdent` → `PackageInstall`. The complication is that [a number of other inputs are necessary](https://github.com/habitat-sh/habitat/blob/master/components/common/src/command/package/install.rs#L419-L427).

PR #6775, opened 2019-07-30
Work around 0.83.0 studio not installing bug

scotthain:


.expeditor/scripts/shared.sh
```diff
@@ -38,6 +38,8 @@ install_latest_hab_binary() {
     rm -rf /hab/pkgs/core/hab/0.82.0
     curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -t "$pkg_target"
     hab_binary="/bin/hab"
+    # TODO: workaround for https://github.com/habitat-sh/habitat/issues/6771
```
christophermaier: It would be good to add this `TODO` note to the other explicit studio installs you have above.

PR #6780, opened 2019-08-01
Fixing up some End-to-End Supervisor Tests

christophermaier:
Enables the following end-to-end tests in our end-to-end pipeline:

* [`test_launcher_restarts_supervisor.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_launcher_restarts_supervisor.sh) 
* [`test_socket_file_cleanup.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_socket_file_cleanup.sh) 
* [`test_tar_export.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_tar_export.sh)

PR #6781, opened 2019-08-02
Add e2e test to ensure studio auto-installs

smacfarlane:
Addresses #6771 

This is a first pass at adding end-to-end tests for the studio. The goal of this test is to ensure that the auto-install behavior of the chroot-based studio is preserved when making changes to the `hab studio *` commands. 

### DIFF
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -126,3 +126,15 @@ steps:
           privileged: true
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
+
+  - label: "[:linux: test_studio_auto_installs]"
+    command:
+      - .expeditor/scripts/end_to_end/setup_environment.sh DEV
+      - test/end-to-end/test_studio_auto_installs.sh
+    expeditor:
+      executor:
+        docker:
+          privileged: true
+          environment:
+            - BUILD_PKG_TARGET=x86_64-linux
+            - HAB_BLDR_URL=https://bldr.acceptance.habitat.sh
--- a/test/end-to-end/test_studio_auto_installs.sh
+++ b/test/end-to-end/test_studio_auto_installs.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# This test is designed to catch the regression described in 
+# https://github.com/habitat-sh/habitat/issues/6771
+# 
+# When a user runs `hab studio enter` for the first time after installing a
+# new Habitat release, the `core/hab-studio` package won't be present on the 
+# system and the cli will automatically download and install the appropirate 
+# package. Since we always install the studio as part of our build process to 
+# ensure we're using the correct version, this behavior needs to be exercised 
+# as its own test. 
+
+set -euo pipefail
+
+# Ensure there are no studios installed
+while [ -d /hab/pkgs/core/hab-studio ]; do 
+  hab pkg uninstall core/hab-studio
+done
+
+# 'studio enter' requires a signing key to be present for the current origin
+echo "--- Generating signing key for $HAB_ORIGIN"
+hab origin key generate "$HAB_ORIGIN" 
+
+echo "--- Creating new studio"
+hab studio new
+
+echo "--- $(hab studio version)"


smacfarlane:
#6780 will contain fixes needed for this to run.

.expeditor/end_to_end.pipeline.yml
```diff
@@ -112,3 +112,15 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
             - HAB_BLDR_URL=$ACCEPTANCE_HAB_BLDR_URL
+  - label: "[:linux: test_studio_auto_installs]"
+    command:
+      - .expeditor/scripts/setup_environment.sh DEV
+      - test/end-to-end/test_studio_auto_installs.sh
+    # TODO - FIX ME
+    soft_fail: true
```
christophermaier: We can remove this now.

.expeditor/end_to_end.pipeline.yml
```diff
@@ -112,3 +112,15 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
             - HAB_BLDR_URL=$ACCEPTANCE_HAB_BLDR_URL
+  - label: "[:linux: test_studio_auto_installs]"
```
christophermaier: This is a personal preference, but I find these pipeline stage definitions to be easier to read if they're separated from each other by newlines.

test/end-to-end/test_studio_auto_installs.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -euo pipefail
+
+export HAB_BLDR_URL="${ACCEPTANCE_HAB_BLDR_URL}"
```
christophermaier: If we can try and rely as much as possible on standard environment variables here, I think that would be easier to reason about, as well as easier to run locally (fewer "special" variables to mentally juggle).

To that end, relying on `HAB_BLDR_URL` already being set from outside would be nice to do if we can; I know I ran into some situations where code that runs before the test was trying to download expeditor itself based on this variable, and ultimately failing. If that's _not_ the case here, simplifying would be nice.

test/end-to-end/test_studio_auto_installs.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
christophermaier: Documenting the intent and reason behind the test up here would be good, along with links (PRs, issues) to any regressions it's trying to catch. Any assumptions the script makes about system state, environment, etc. would also be good to document.

If any of the above can be verified / proven by using code, then that would be even better.

test/end-to-end/test_studio_auto_installs.sh
smacfarlane: I've tried adding HAB_BLDR_URL to the `environment` hash in the pipeline definition and it appears to have no effect, so there is something wrong with how we're using that that I need to dig into.

.expeditor/end_to_end.pipeline.yml
smacfarlane: Agreed,  that was a copy/paste error on my part.  Fixed :)

test/end-to-end/test_studio_auto_installs.sh
smacfarlane: Turns out I can't spell `acceptance` :man_facepalming:

.expeditor/end_to_end.pipeline.yml
```diff
@@ -112,3 +112,16 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
             - HAB_BLDR_URL=$ACCEPTANCE_HAB_BLDR_URL
+
+  - label: "[:linux: test_studio_auto_installs]"
+    command:
+      - .expeditor/scripts/setup_environment.sh DEV
+      - hab origin key generate $HAB_ORIGIN
```
scotthain: Why isn't this in the e2e script?

smacfarlane: Originally it was, but I found that iterating on it cluttered up my local workstation with fake keys.  Since we don't have a great way to check for a key or conditionally generate one at the moment, I decided to move it into the pipeline steps.  

Locally, you're almost always guaranteed to have a key generated for your default origin and I'd argue that we should leave it up to a user to do that.  In the pipeline however we can't make that guarantee and a key is required, otherwise the studio creates but exits `1` because it has no keys to import. 

With https://github.com/habitat-sh/habitat/pull/6855 this may not be an issue anymore in which case I'd agree it should go back into the script. I'd need to verify that keys won't make it out of that container. I don't think they will, but haven't looked closely at that PR yet.

scotthain: If the assumption is there will always be a key available, we may want to move this to `setup_environment` - thoughts?

smacfarlane: I think that assumption is only true in `studio` or `pkg build` cases.  Other cases keys will be fetched as needed from builder. Are we worried about having keys on disk when we don't need them?  The only case I can think of is if we add e2e tests around key generation.   There also may be cases where it would mask behavior like we saw with the studio that lead to this test :smile: , but I can't think of what they might be so are probably aren't real concerns. 

Is the intent that a user runs setup_environment locally when adding/iterating on things? If so, my comments about key generation spam and #6855 remain the same. 

Otherwise, that change seems reasonable to me if we don't care that there is a key generation step when we don't need it.

PR #6789, opened 2019-08-05
Extract a curlbash_hab function

christophermaier:
The original `install_latest_hab_binary` is doing more work than
needed; it both bootstraps `hab` onto a system, and then does extra work
to possibly install `hab` and `hab-studio` in the context of a release.

To help tease these concerns apart, we extract a `curlbash_hab`
function that just handles the bootstrapping logic.

Now, in the places where we really don't care what version of `hab`
we're using, we can just use `curlbash_hab`. Note that we also don't
need to use the `hab_binary` global variable that
`install_latest_hab_binary` set; since the "curlbash" method adds a
link to `/bin/hab`, we can just reference the binary by name directly.

This will allow us to expand our usage into additional pipelines,
without implicitly relying on logic that was only useful in the
context of the release pipeline.

### DIFF
--- a/.expeditor/scripts/remove_build_channel.sh
+++ b/.expeditor/scripts/remove_build_channel.sh
@@ -10,11 +10,11 @@ source .expeditor/scripts/shared.sh
 export HAB_AUTH_TOKEN="${ACCEPTANCE_HAB_AUTH_TOKEN}"
 export HAB_BLDR_URL="${ACCEPTANCE_HAB_BLDR_URL}"
 
-install_latest_hab_binary "$BUILD_PKG_TARGET"
+curlbash_hab "$BUILD_PKG_TARGET"
 
 channel="$(get_release_channel)"
 echo "--- Destroying release channel '${channel}'"
 
-${hab_binary} bldr channel destroy \
+hab bldr channel destroy \
     --origin=core \
     "${channel}"
--- a/.expeditor/scripts/shared.sh
+++ b/.expeditor/scripts/shared.sh
@@ -25,18 +25,31 @@ get_latest_pkg_version_in_channel() {
     echo "${version}"
 }
 
+# Install the latest stable `hab` binary using our standard
+# "curl|bash" approach.
+#
+# This is the currently recommended way to get Habitat onto a system
+# that does not already have it.
+#
+# As usual, this will install `hab` at `/bin/hab`
+curlbash_hab() {
+    local pkg_target="${1:-$BUILD_PKG_TARGET}"
+    echo "--- :habicat: Bootstrap installation of the current stable hab binary for $pkg_target using curl|bash"
+    # TODO:
+    # really weird corner case on linux2 because the 0.82.0 versions of both
+    # are the same. let's just delete it
+    rm -rf /hab/pkgs/core/hab/0.82.0
+    curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -t "$pkg_target"
+}
+
 # Always install the latest hab binary appropriate for your linux platform
 #
 # Accepts a pkg target argument if you need to override it, otherwise
 # will default to the value of `BUILD_PKG_TARGET`
 install_latest_hab_binary() {
     local pkg_target="${1:-$BUILD_PKG_TARGET}"
-    echo "--- Installing latest hab binary for $pkg_target using curl|bash"
-    # TODO:
-    # really weird corner case on linux2 because the 0.82.0 versions of both
-    # are the same. let's just delete it
-    rm -rf /hab/pkgs/core/hab/0.82.0
-    curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -t "$pkg_target"
+    curlbash_hab "${pkg_target}"
+
     hab_binary="/bin/hab"
     # TODO: workaround for https://github.com/habitat-sh/habitat/issues/6771
     ${hab_binary} pkg install core/hab-studio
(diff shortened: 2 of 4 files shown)


.expeditor/scripts/shared.sh
```diff
@@ -25,18 +25,31 @@ get_latest_pkg_version_in_channel() {
...
+# As usual, this will install `hab` at `/bin/hab`
+curlbash_hab() {
     local pkg_target="${1:-$BUILD_PKG_TARGET}"
-    echo "--- Installing latest hab binary for $pkg_target using curl|bash"
+    echo "--- :habicat: Bootstrap installation of the current stable hab binary for $pkg_target using curl|bash"
     # TODO:
     # really weird corner case on linux2 because the 0.82.0 versions of both
     # are the same. let's just delete it
     rm -rf /hab/pkgs/core/hab/0.82.0
     curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -t "$pkg_target"
```
smacfarlane: Could this be simplified by running the script from the repo?    

If we were to break the install script, hopefully running it from the repo would catch it before it makes it to master. If it were to make it to master, then we'd be broken with `curl|bash` until we merged a fix back in.   

The curl adds an additional network hop in the setup of each step, but if curl were to fail it more than likely means GitHub is having issues so avoiding the curl probably wouldn't net us any benefits.

I went down this path in our existing pipeline and ended up not using the install.sh script at all. Unfortunately, I don't appear to have captured this information. From memory, it had to do with the install script not properly updating symlinks to the requested version if there was already a hab on the system.  Something to keep our eye on as we iterate on this.

.expeditor/scripts/remove_build_channel.sh
```diff
@@ -10,11 +10,11 @@ source .expeditor/scripts/shared.sh
...
 export HAB_BLDR_URL="${ACCEPTANCE_HAB_BLDR_URL}"
 
-install_latest_hab_binary "$BUILD_PKG_TARGET"
+curlbash_hab "$BUILD_PKG_TARGET"
 
 channel="$(get_release_channel)"
 echo "--- Destroying release channel '${channel}'"
 
-${hab_binary} bldr channel destroy \
+hab bldr channel destroy \
```
scotthain: I would argue that we should continue to use the `hab_binary` variable, even though we don't need it on linux - this way we have consistency on the binlinked vs install.sh based hab binary on OSX. (one installs to /usr/local/bin, the other to /usr/bin IIRC, which is an issue in and of itself, in my opinion)

.expeditor/scripts/shared.sh
christophermaier: Hrmm... if nothing else, this points out that we should probably have a suite of tests around our `install.sh` script itself.

If you can find the concrete issue's with `install.sh` that you ran into, that would be useful @smacfarlane. At the moment, my inclination is to leave this change "as is", since it's a pretty minimal refactoring, but I'm certainly not opposed to further changes in the future.

scotthain: Yes - let's not conflate these two feature/issues here. That would be a great improvement to log in an issue in the releng repo though.

.expeditor/scripts/remove_build_channel.sh
christophermaier: My motivation here was to streamline the code. When you see `${hab_binary}` instead of `hab`, your first thought is "Why are we not using `hab` directly... what is it about this code that means we have to care?" That _is_ a concern in the release pipeline, where we want to ensure that we're building with the Habitat toolchain that we're currently building.

However, in other cases, we really don't care as much; we just need a `hab` binary that will let us do simple things like interact with the Builder API. In those cases, we can just use `curlbash_hab` to install the latest stable `hab`, just like any normal user would, and then just use `hab`. Whether that's in `/usr/bin` or `/usr/local/bin` doesn't matter; the fact that the `install.sh` script that `curlbash_hab` is using under the hood ensures that `hab` is on `$PATH` is the key thing.

scotthain: While I agree on a technical level that we don't have to care in some places, it feels sloppy to rely on the whatever is on the path, especially since we should be using the "newly built hab" for as many things as possible after studio is built, just to exercise it. I my mind the first thing i think when I see `${hab_binary}` is "Ok I can trust this is set correctly" rather than thinking "which hab is on the path right now?" :)

christophermaier: Even that won't work because we're pulling from the "current release channel" which is tied to Buildkite job. It's fine in a release pipeline, but will be a no-op in any downstream pipeline.

That is, absent any other changes, the current refactoring actually captures what is really going on now in those downstream pipelines.

Whether or not we want to always use the `hab` binary from `DEV` in those downstream pipelines is probably something to chat about, since I can see various arguments pro and con. Depending on the outcome of that, we can make subsequent changes to more explicitly capture that in our helper functions.

scotthain: That makes sense - I'm totally for splitting up the functions - I think we should probably rename the 2nd one to like "install_release_channel_hab" or even "install_hab_from_channel" with an argument (which would then work for DEV in downstream pipelines.

christophermaier: @scotthain Yeah that sounds like a good follow-on refactoring. I'll probably add something like that in my ongoing work on https://github.com/habitat-sh/habitat/pull/6780

christophermaier: @scotthain Yup, that sounds good. I'll probably end up adding something like that in my ongoing work for https://github.com/habitat-sh/habitat/pull/6780.

Issue #6791, opened 2019-08-05
Too many layers of symlinks


dmccown:
# Description

When using the chroot studio with 0.84 development versions, studio commands fail with `/hab/pkgs/core/coreutils/8.30/20190115012313/bin/env: ‘/hab/bin/hab’: Too many levels of symbolic links`.

# Acceptance Criteria
1. `hab pkg build` should not crash a chroot studio due to the binlink in the Studio.

dmccown:
https://github.com/habitat-sh/habitat/pull/6700

smacfarlane:
I believe the issue is here: https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/pkg/binlink.rs#L269-L288

We check to verify that the path that we intend to write the script to is not the same as what the `#!` would render as. However, when creating a chroot studio, this is insufficient.  The binlinks are created from outside the studio, before we enter it, so the resulting path will look like `/hab/studios/<normalized_path>/bin/sh`. Since we allow the user to override the `STUDIO_NAME` this is realistically `/some/valid/path/bin/sh` though.   

I can't think of a safe way to test for this when the prefix can be some arbitrary path (not to say one doesn't exist :smile:) , so my gut feeling say we need to provide opt-in/opt-out capabilities for the binlink command.  My preference would be for opt-in as that would preserve existing behavior, while allowing for our users to adopt it as they're ready.  Where this gets tricky though is, from my understanding, that this is required behavior on Windows.  That would leave us with different behaviors on platforms which will inevitably lead to confusion and a bad experience for our users. 

I think in the long run, binstubs are a much safer option for our Linux experience compared to binlinks, but I think it would be best to take an iterative approach to adopting them.

baumanj:
> We check to verify that the path that we intend to write the script to is not the same as what the `#!` would render as. However, when creating a chroot studio, this is insufficient. The binlinks are created from outside the studio, before we enter it, so the resulting path will look like `/hab/studios/<normalized_path>/bin/sh`

Could we compare inode numbers (`stat -f "%i" /path/to/file`) to identify whether they refer to the same file?

christophermaier:
Is there any way we can programmatically access the full studio prefix to us in comparisons? If not, would that be possible to easily add?

Can we add some tests that exercise this functionality and use that to further characterize the problem?

smacfarlane:
> Is there any way we can programmatically access the full studio prefix to us in comparisons? If not, would that be possible to easily add?

[...]

PR #6796, opened 2019-08-06
Revert "redux: cross platform binstubs with package environment"

smacfarlane:
Reverts habitat-sh/habitat#6700

Ref: #6791 

There an edge case with the studio and generating the binstubs that is preventing us from performing release builds,  see #6791 for details.  

For now, I think it's best we back this out and take this to a design doc to discuss the goals of the `binlink` command and explore any additional potential edge cases our users might encounter.

### DIFF
--- a/components/hab/src/command/pkg/binlink.rs
+++ b/components/hab/src/command/pkg/binlink.rs
@@ -1,9 +1,10 @@
-use std::{collections::HashMap,
-          env,
-          fs::{self,
-               File},
-          io::{BufRead,
-               BufReader},
+#[cfg(windows)]
+use std::fs::File;
+#[cfg(windows)]
+use std::io::{BufRead,
+              BufReader};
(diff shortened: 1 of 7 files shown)


christophermaier:
@smacfarlane are there any backward compatibility issues that this revert might introduce that would need to be addressed?

smacfarlane:
@christophermaier We haven't released this change yet so from that perspective we should be safe.  There are builds of 0.84.x from the release pipeline work on builder, which gives me a little pause, but they've remained in the unstable or 'fake-0.84' channels. 

I'm re-reading the commit history since this change, but a quick examination didn't show any additional changes in these files.

mwrock:
I'm just catching up from vacation and have not read in to this yet. Before we merge this is like the chance to have a look. And at the least just revert the Linux implementation.

PR #6802, opened 2019-08-07
Add mac hab build to release process

scotthain:
Still needs commit cleanup, rebase, and final verification.


.expeditor/scripts/build_mac_component.sh
```diff
@@ -0,0 +1,84 @@
...
+channel=$(get_release_channel)
+
+echo "--- Channel: $channel - bldr url: $HAB_BLDR_URL"
+
+echo "--- Installing buildkite agent"
+brew tap buildkite/buildkite
+brew install --token="$BUILDKITE_AGENT_ACCESS_TOKEN" buildkite-agent
+
+echo "--- Installing mac bootstrap package"
+bootstrap_package_version="1.0.7"
```
smacfarlane: Do we want this hard-coded rather than a "latest" package?   Future iterations on the bootstrap package would require us to remember to bump this pipeline.  

Not a blocker ( I assume that would require work on the mac omnibus pipeline ), just a thought.

.expeditor/scripts/build_mac_component.sh
```diff
@@ -0,0 +1,84 @@
...
+sudo hdiutil attach "mac-bootstrapper-${bootstrap_package_version}-1.dmg"
+sudo installer -verbose -pkg "/Volumes/Habitat macOS Bootstrapper/mac-bootstrapper-${bootstrap_package_version}-1.pkg" -target /
+sudo hdiutil detach "/Volumes/Habitat macOS Bootstrapper"
+brew install wget
+export PATH=/opt/mac-bootstrapper/embedded/bin:/usr/local/bin:$PATH
+
+install_latest_hab_binary "$BUILD_PKG_TARGET"
+import_keys
+
+# Because we are not running as root on mac, move our keys
```
smacfarlane: Perhaps a little additional clarity 
```
# We invoke hab-plan-build.sh directly via sudo, so we don't get the key management that studio provides. 
# Copy keys from the user account Habitat cache to the system Habitat cache so that they are present for root.
```

.expeditor/scripts/build_mac_component.sh
scotthain: The future of this will be in a file, similar to the `rust-version` files so we can set up a subscription for when the mac-bootstrapper gets updated. So yes!

.expeditor/scripts/build_mac_component.sh
scotthain: 👍

.expeditor/scripts/shared.sh
```diff
@@ -50,12 +66,13 @@ install_latest_hab_binary() {
     local pkg_target="${1:-$BUILD_PKG_TARGET}"
     curlbash_hab "${pkg_target}"
 
-    hab_binary="/bin/hab"
-    # TODO: workaround for https://github.com/habitat-sh/habitat/issues/6771
-    ${hab_binary} pkg install core/hab-studio
+    if [ "$pkg_target" != "x86_64-darwin" ]; then
+        # TODO: workaround for https://github.com/habitat-sh/habitat/issues/6771	
+        ${hab_binary} pkg install core/hab-studio
+    fi
```
christophermaier: There's actually no reason this function should be called on macOS, because it's really about ensuring that the hab and studio version that we're currently building are installed on the machine. Since that's explicitly not something that we ever do on macOS, we don't need this function.

I'd argue that we should probably error if this is called on macOS at the moment, just to help keep us honest.

.expeditor/scripts/shared.sh
```diff
@@ -40,6 +40,22 @@ curlbash_hab() {
...
+            ;;
+        x86_64-darwin)
+            hab_binary="/usr/local/bin/hab"
+            ;;
+        *)
+            echo "--- :no_entry_sign: Unknown PackageTarget: ${pkg_target}"
+            exit 1
+            ;;
+    esac
+    declare -g hab_binary
```
christophermaier: I'm still not sure about this... the whole idea of the curlbash install is that you get `hab` on your path.

.expeditor/scripts/shared.sh
```diff
@@ -68,7 +85,6 @@ install_latest_hab_binary() {
     else
```
christophermaier: Somehow I think the original intent of `hab_binary` got lost... it was supposed to be the concrete path to a `/hab/pkgs/core/hab/XXX/XXXXXXXXXXXX/bin/hab` binary, so we could ensure we used the one we just installed, rather than whatever was curlbashed into the `PATH`.

After all, if `hab_binary` is always pointing to the on-PATH location that it gets curlbashed to, there isn't much value in setting it in a variable... you'd just always call `hab` directly.

As such, if we're going to be setting `hab_binary` it should be in this function, and it should be the complete path to that `hab` binary.

I think we should probably also rename this function to better reflect what it's doing. Thinking a bit more, we should probably completely remove the call to `curlbash_hab` from inside this function. That will further help to clarify the distinct responsibilities the two functions have.

.expeditor/scripts/build_mac_component.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/local/bin/bash
+
+set -euo pipefail
+
+source .expeditor/scripts/shared.sh
+
+# Get secrets!
+eval "$(vault-util fetch-secret-env)"
```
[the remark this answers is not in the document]
scotthain: This is pulling/injecting our HAB_AUTH_TOKEN and AWS creds from vault

.expeditor/scripts/shared.sh
scotthain: Good call - I didn't see the forest for the trees :D

.expeditor/scripts/build_mac_component.sh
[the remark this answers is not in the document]
christophermaier: We override `HAB_AUTH_TOKEN` immediately below, and I don't see where we're interacting with AWS. 

Could we just remove this line, then?

.expeditor/scripts/shared.sh
christophermaier: :palm_tree: :christmas_tree: :tanabata_tree: :deciduous_tree: :evergreen_tree:

.expeditor/scripts/build_mac_component.sh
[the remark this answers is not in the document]
scotthain: I wasn't clear, sorry - for right now we're pulling the ACCEPTANCE_HAB_AUTH_TOKEN from vault (which is my acceptaance token) from vault and overriding it. When we go live with this we can skip the `export` line here because it'll already be in the env.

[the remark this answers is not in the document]
christophermaier: Is there a reason why we're not setting that in the pipeline like the rest of the build steps? If not, it would be nice to have them all be as self-consistent as possible.

(it would also be good to add a comment that indicates which secrets in particular we're intending to pull in)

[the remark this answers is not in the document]
scotthain: They are actually: https://github.com/habitat-sh/habitat/blob/master/.expeditor/release_habitat.pipeline.yml#L1-L6

We only have this extra "eval" because of this bug: https://github.com/chef/ci-studio-common/issues/200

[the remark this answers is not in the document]
scotthain: I will totally add some comments to this though explaining the situation :)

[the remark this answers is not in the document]
christophermaier: Great! A link to that issue will definitely help :smile: 

Thanks!

christophermaier:
The changes thus far look great! 👍

.expeditor/scripts/build_component.sh
```diff
@@ -20,6 +20,7 @@ channel=$(get_release_channel)
 
 echo "--- Channel: $channel - bldr url: $HAB_BLDR_URL"
 
+declare -g hab_binary
```
christophermaier: Might be nice to wrap this in a helper function and call _that_ instead, just to help centralize the logic, and make it easier to evolve over time (I'd love it if we could get away from a global variable if possible... but I realize this is `bash`, after all :frowning: )

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
```
christophermaier: This is just a random thought... would it be worth prefixing the names of functions in this (and similar) shared "library files" with the "namespace" the function comes from, as an additional clue to readers where code is defined and how it is intended to be used?

e.g., `release.import_keys`, `release.get_latest_pkg_version`, etc?

I'd know exactly where the function was defined, and if I'm using a `release.*` function in one of my end-to-end tests, I know I've done something very wrong.

.expeditor/scripts/build_mac_hab_binary.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/local/bin/bash
+
+set -euo pipefail
+
+source .expeditor/scripts/shared_release_habitat.sh
+
+# Get secrets! (our auth token and aws creds should be auto-injected but there's a bug:
+# https://github.com/chef/ci-studio-common/issues/200)
+eval "$(vault-util fetch-secret-env)"
```
baumanj: I'm alway very wary of `eval` statements. What does this return?

.expeditor/scripts/build_mac_hab_binary.sh
```diff
@@ -0,0 +1,81 @@
...
+channel=$(get_release_channel)
+
+echo "--- Channel: $channel - bldr url: $HAB_BLDR_URL"
+
+echo "--- Installing buildkite agent"
+brew tap buildkite/buildkite
+brew install --token="$BUILDKITE_AGENT_ACCESS_TOKEN" buildkite-agent
+
+echo "--- Installing mac bootstrap package"
+# TODO: subscribe to releases: https://github.com/habitat-sh/release-engineering/issues/84
```
baumanj: I think we agreed to use `TODO` to mark things that should happen _before_ merge (such as create the issue for future fork). Since that issue is already created and linked here, I think we can remove the `TODO` and just leave the issue link.

.expeditor/scripts/build_mac_hab_binary.sh
```diff
@@ -0,0 +1,81 @@
...
+bootstrap_package_version="$(cat MAC_BOOTSTRAPPER_VERSION)"
+bootstrap_package_name="mac-bootstrapper-${bootstrap_package_version}-1"
+curl "https://packages.chef.io/files/current/mac-bootstrapper/${bootstrap_package_version}/mac_os_x/10.12/${bootstrap_package_name}.dmg" -o "${bootstrap_package_name}.dmg"
```
baumanj: Seem like we could eliminate the 
```suggestion
curl "https://packages.chef.io/files/current/mac-bootstrapper/${bootstrap_package_version}/mac_os_x/10.12/${bootstrap_package_name}.dmg" -O"
```

.expeditor/scripts/build_mac_hab_binary.sh
scotthain: It exposes and exports the secrets from the `vault-util` utility.

.expeditor/scripts/build_mac_hab_binary.sh
```diff
@@ -0,0 +1,81 @@
...
+
+echo "--- :rust: Installing Rust"
+curl https://sh.rustup.rs -sSf | sh -s -- -y
+# This ensures the appropriate binaries are on our path
+source "${HOME}/.cargo/env"
+
+# set the rust toolchain
+rust_toolchain="$(cat rust-toolchain)"
+echo "--- :rust: Using Rust toolchain ${rust_toolchain}"
+rustup default "${rust_toolchain}"
```
baumanj: I think this will be done automatically via the existence of [the `rust-toolchain` file](https://github.com/rust-lang/rustup.rs#the-toolchain-file), no?

.expeditor/scripts/build_mac_hab_binary.sh
scotthain: That seems counterintuitive to me but if that's what consensus is I'm fine with it!

.expeditor/scripts/build_mac_hab_binary.sh
```diff
@@ -0,0 +1,81 @@
...
+rustc --version # just 'cause I'm paranoid and I want to double check
+
+echo "--- :habicat: Building components/hab"
+
+HAB_BLDR_CHANNEL="${channel}" sudo -E bash \
+        components/plan-build/bin/hab-plan-build.sh \
+        components/hab
+source results/last_build.env
+
+echo "--- :habicat: Uploading ${pkg_ident:?} to ${HAB_BLDR_URL} in the '${channel}' channel"
```
baumanj: The `:?` on `pkg_ident` is redundant with `set -eu`, so it's unclear to me why it's applied to `pkg_ident` (and a couple other variables, but not to others.

.expeditor/scripts/shared.sh
```diff
@@ -40,148 +19,20 @@ curlbash_hab() {
     # are the same. let's just delete it
     rm -rf /hab/pkgs/core/hab/0.82.0
     curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -t "$pkg_target"
+    case "${pkg_target}" in
+        x86_64-linux)
```
baumanj: These cases can be combined; it makes it clearer that they do the same thing
```suggestion
        x86_64-linux | x86_64-linux-kernel2)
```
or
```suggestion
        x86_64-linux*)
```

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,176 @@
...
+#
+# Requires the `hab_binary` global variable is already set!
+#
+# Accepts a pkg target argument if you need to override it, otherwise
+# will default to the value of `BUILD_PKG_TARGET`
+install_release_channel_hab_binary() {
+    local pkg_target="${1:-$BUILD_PKG_TARGET}"
+    curlbash_hab "${pkg_target}"
+
+    # TODO: workaround for https://github.com/habitat-sh/habitat/issues/6771
```
baumanj: Same `TODO` note as before

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,176 @@
...
+    # TODO: workaround for https://github.com/habitat-sh/habitat/issues/6771	
+    ${hab_binary} pkg install core/hab-studio
+
+    echo "--- :habicat: Installed latest stable hab: $(${hab_binary} --version)"
+    # now install the latest hab available in our channel, if it and the studio exist yet
+
+    hab_version=$(get_latest_pkg_version_in_channel "hab")
+    studio_version=$(get_latest_pkg_version_in_channel "hab-studio")
+
+    if [[ -n $hab_version && -n $studio_version && $hab_version == "$studio_version" ]]; then
```
baumanj: Would `hab_version` or `studio_version` ever be set to the empty string? If not
```suggestion
    if [[ $hab_version == "$studio_version" ]]; then
```
should suffice.

.expeditor/scripts/build_mac_hab_binary.sh
baumanj: More specifically, what does the format look like? I'm trying to understand why it _needs_ to be an `eval` as opposed to running a regular command because `eval` is incredibly easy to mess up in very dangerous ways and should only be used as a las resort.

.expeditor/scripts/build_mac_hab_binary.sh
scotthain: So the `rustup default` command isn't needed?

.expeditor/scripts/shared_release_habitat.sh
scotthain: Yes there is a very real possibility that either or both could be empty string.

.expeditor/scripts/build_mac_hab_binary.sh
scotthain: It's going to be exported as such: https://github.com/chef/ci-studio-common/blob/master/cmd/vault-util/commands/fetch-secret-env.go#L66, key value pairs. There is an issue (linked) where it doesn't behave the same way on linux as it does on macos.

.expeditor/scripts/build_mac_hab_binary.sh
baumanj: The idea is that `TODO`s in code (rather than as issues) make it hard to keep track of the work we want to do in the future because it leads to fragmentation. We want all our planned work to exist in issues to capture context and be discoverable. It's fine to link issues in code, but we're trying to get away from merging the text `TODO` specifically, because we want it to serve a different purpose.

The idea there is that putting `TODO` in a bit of code is a fine practice to keep your momentum so you don't have to to search for or create the issue right away and can keep working on the code task at hand (perhaps the TODO gets resolved without needing an issue at all). Then later, when things are ready to merge, we can go back and make sure all the `TODO` get dealt with appropriately (either by code changes or issue creation) _before_ merge.

In this example, while working on the code, writing
```bash
# TODO: subscribe to releases
```
and then before merging, creating the issue changing that `TODO` line to:
```bash
# subscribe to releases: https://github.com/habitat-sh/release-engineering/issues/84
```
is the expected workflow. `TODO` is just a marker that can help us remember to do things before merge without getting to distracted. Does that make sense?

scotthain: I understand. That's not a workflow I've used, but like I said if that's how we're doing it, I'm fine with it! :) I feel like having a marker in the code to link to an issue (todo, fixme, whatever) is easier to search for unless you're linking directly in the issue to the code in question, which may be what y'all are doing already.

.expeditor/scripts/build_mac_hab_binary.sh
baumanj: I don't think so, no.

.expeditor/scripts/build_mac_hab_binary.sh
baumanj: I see. That's fairly reasonable as `eval` calls go. Still makes me nervous, but probably not worth the hassle to do things more safely.

scotthain: We need to do some debugging and figure out what the real issue is, but now didn't seem like the time.

.expeditor/scripts/build_mac_hab_binary.sh
baumanj: If the issue is searchability, grepping for `"issues/\d+"` works pretty well

.expeditor/scripts/shared_release_habitat.sh
baumanj: I understand now. Still, 
```suggestion
    if [[ -n $hab_version && $hab_version == "$studio_version" ]]; then
```
should be enough, no?

scotthain: We could do `-n $studio_version && $hab_version == "$studio_version"` I suppose. I like being explicit about the fact that we want both to exist and match, and I think it makes the intent clearer.

baumanj: As a reader of the code, if a test seems logically redundant, I tend to think I must've misunderstood it. Reading this, I puzzled over whether there was a way for `-n $hab_version` and `$hab_version == "$studio_version"` to be true while `-n $studio_version` was false. If the goal is maximizing understandability, I tend to find minimizing the number of terms in a test to be generally helpful.

That said, it's not a biggie, so go with what you think it best; just wanted to share my experience as a reader.

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,175 @@
...
+    echo "--- :habicat: Installed latest stable hab: $(${hab_binary} --version)"
+    # now install the latest hab available in our channel, if it and the studio exist yet
+    hab_version=$(get_latest_pkg_version_in_channel "hab")
+    studio_version=$(get_latest_pkg_version_in_channel "hab-studio")
+
+    if [[ -n $hab_version && -n $studio_version && $hab_version == "$studio_version" ]]; then
+        echo "-- Hab and studio versions match! Found hab: ${hab_version:-null} - studio: ${studio_version:-null}. Upgrading :awesome:"
+        channel=$(get_release_channel)
+        ${hab_binary:?} pkg install --binlink --force --channel "${channel}" core/hab
+        ${hab_binary:?} pkg install --binlink --force --channel "${channel}" core/hab-studio
```
christophermaier: Are these binlinks really needed? The whole idea of using `$hab_binary` is to unambiguously refer to a specific version of `hab`, and binlinking is just going to set the global `hab` to that specific version.

I also don't think there's a need to link to anything in `core/hab-studio`, since we don't call it directly, but rely the `hab` binary to resolve it correctly.

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,175 @@
...
+    # workaround for https://github.com/habitat-sh/habitat/issues/6771	
+    ${hab_binary} pkg install core/hab-studio
+
+    echo "--- :habicat: Installed latest stable hab: $(${hab_binary} --version)"
+    # now install the latest hab available in our channel, if it and the studio exist yet
+    hab_version=$(get_latest_pkg_version_in_channel "hab")
+    studio_version=$(get_latest_pkg_version_in_channel "hab-studio")
+
+    if [[ -n $hab_version && -n $studio_version && $hab_version == "$studio_version" ]]; then
+        echo "-- Hab and studio versions match! Found hab: ${hab_version:-null} - studio: ${studio_version:-null}. Upgrading :awesome:"
```
christophermaier: If we get down here, `$hab_version` and `$studio_version` should always be set, so I don't think we need the `:-null` bits.

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,175 @@
...
+# Accepts a pkg target argument if you need to override it, otherwise
+# will default to the value of `BUILD_PKG_TARGET`
+install_release_channel_hab_binary() {
+    local pkg_target="${1:-$BUILD_PKG_TARGET}"
+    curlbash_hab "${pkg_target}"
+
+    # workaround for https://github.com/habitat-sh/habitat/issues/6771	
+    ${hab_binary} pkg install core/hab-studio
+
+    echo "--- :habicat: Installed latest stable hab: $(${hab_binary} --version)"
```
christophermaier: This should probably be moved into `curlbash_hab` for consistency.

.expeditor/scripts/shared_release_habitat.sh
christophermaier: I think it would probably be more readable and correct to do install using e.g. `core/hab/${hab_version}` and `core/hab-studio/${studio_version}`.

.expeditor/scripts/shared_release_habitat.sh
```diff
@@ -0,0 +1,175 @@
...
+has_hab_ident() {
+    local target=$1
+    buildkite-agent meta-data exists "hab-ident-${target}"
+}
+
+set_hab_ident() {
+    local target=$1
+    local ident=$2
+    buildkite-agent meta-data set "hab-ident-${target}" "${ident}"
+}
```
christophermaier: I think these metadata functions aren't being used anymore.  `get_hab_ident` and `has_hab_ident` aren't, at least, despite `set_hab_ident` being used.

It would be good to audit the rest and remove whatever's not being used anymore.

PR #6835, opened 2019-08-09
Fix "warning: unused variable: `curr_path`" on linux

baumanj:
This variable was only used in two Windows-specific lines, so using the literal instead should silence this warning without affecting Windows.

### DIFF
--- a/components/hab/src/command/pkg/binlink.rs
+++ b/components/hab/src/command/pkg/binlink.rs
@@ -292,7 +292,6 @@ mod test {
         let hypnoanalyze_link = "hypnoanalyze.exe";
         #[cfg(target_os = "windows")]
         let hypnoanalyze_link = "hypnoanalyze.bat";
-        let curr_path = ";%PATH%";
 
         start(&mut ui,
               &ident,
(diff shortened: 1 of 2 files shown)


components/hab/src/command/pkg/binlink.rs
```diff
@@ -301,7 +300,7 @@ mod test {
               rootfs.path(),
               force).unwrap();
         #[cfg(windows)]
-        assert!(fs::read_to_string(rootfs_bin_dir.join(magicate_link)).unwrap().contains(&format!("PATH={}{}", rootfs_src_dir.to_string_lossy(), curr_path)));
+        assert!(fs::read_to_string(rootfs_bin_dir.join(magicate_link)).unwrap().contains(&format!("PATH={}{}", rootfs_src_dir.to_string_lossy(), ";%PATH%")));
```
christophermaier: I wonder if we need to start running `rustfmt` in CI on Windows, too... Is this line allowed to stay this long because it's only on Windows? :thinking:

mwrock: when I run rustfmt locally on windows ^^^ is the result

baumanj: `rustfmt` doesn't interpret `#[cfg(windows)]`. When I run it locally, it still reformats things in those blocks. I'm pretty sure it operates at the AST level and ignores conditional compilation other than `rustfmt`-specific directives.

Issue #6817, opened 2019-08-08
Address all existing tests that are annotated `#[ignore]`


christophermaier:
We have a few existing unit and integration tests in Rust that are annotated with the `#[ignore]` attribute, which means they are not run. Historically, this has been added to tests that became "flaky" for some reason, but we were not able to investigate further. When we have dug into some of these tests in the pasts, the flakiness was revealed to be an actual legitimate issue.

We should not have ignored tests. If they are valuable tests, they should be fixed. If they provide no value, they should be either modified so they do provide value, or removed altogether.

# Acceptance Criteria
Running `git grep '#\[ignore\]'` returns 0 matches.

This doesn't mean that there are no ignored tests; it means that any ignored tests have the reason for being ignored are properly annotated as to why.

PR #6836, opened 2019-08-09
Reenable or properly annotate and run ignored tests

baumanj:
Resolves https://github.com/habitat-sh/habitat/issues/6817

See individual commit messages for details.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -80,7 +80,7 @@ steps:
       automatic:
         limit: 1
 
-- label: "[unit] :linux: builder-api-client ignored"
+  - label: "[unit] :linux: builder-api-client ignored"
     command:
       - ./test/run_cargo_test.sh builder-api-client -- --ignored
     agents:
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -80,6 +80,22 @@ steps:
       automatic:
         limit: 1
 
+- label: "[unit] :linux: builder-api-client ignored"
+    command:
+      - ./test/run_cargo_test.sh builder-api-client -- --ignored
+    agents:
+      queue: 'default-privileged'
+    plugins:
+      docker#v3.0.1:
+        always-pull: true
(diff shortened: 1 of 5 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -392,6 +408,22 @@ steps:
       automatic:
         limit: 1
 
+  - label: "[unit] :windows: builder-api-client ignored"
```
christophermaier: I must have gotten mixed up with other languages' testing annotations, because I thought Rust had a way to let you actually add arbitrary groupings to tests, so you could, say, concretely annotate long-running tests as such. But no, [per The Book](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested) `#[ignore]` is how you canonically denote long-running tests. Weird.

To that end, what do you think about at least naming the CI stage to reflect the long-running nature of the tests?

Do you have thoughts on how we might make it difficult / impossible to _not_ run ignored / long-running tests in CI? The pattern you have here is good, in that it lets us break out long-running tests in each crate, but it also requires us to remember to add a new stage if we add additional long-running tests to other crates. Maybe a stage that just runs ignored tests found anywhere in the project?

baumanj: I'm wary to change the name of the CI step because if other tests were added as `ignore`, they'd be run by this step too and there's nothing to enforce _why_ the tests would be marked `ignore`. If having long-running unit tests were a bigger part of our system, I'd be in favor of investing more complexity in switching to a feature-based inclusion mechanism, which we could control the naming of*. But since after this PR, there will just be the one test, I'm not sure it's worth it.

As a follow-up, I will look at switching to a single job that covers ignore tests for the whole workspace. That sounds like a nice safety net to avoid marking tests `ignore` and forgetting about them.

\* We had something like this previously, but since it was just being used to ignore flaky tests, switching to using `#[ignore]` seemed simpler.

baumanj: See https://github.com/habitat-sh/habitat/pull/6845

PR #6843, opened 2019-08-12
Add ability to disable job group creation from the cli

### LINKED ISSUE #6762: Add flag to disable builds on package upload
When a package is uploaded to builder, if the `Automated Dependency Rebuilds` is turned on for that package, it will automatically start a build group based on the uploaded package. There are some scenarios where this isn't desirable, such as when you are uploading a set of related packages as is done in a base-plans refresh.

Builder already has this support in its api, so in order to support this we would need to add a flag to `pkg upload` in order to expose this behavior to users.

smacfarlane:
Closes #6762 

This adds the ability for the cli to disable the creation of a build group when uploading packages.  This is useful when building sets of package outside of builder, such as we do with a base-plans refresh.  

Because builder tests presence of the `builder` flag to determine if a build group should be created, I've chosen to represent this as an Option, rather than a boolean. By representing it as an Option, it felt more in-line with what we end up passing to Builder. 

## Behavior change
Previously, if any dependencies were uploaded along with the target artifact they would each trigger a build group (the Thundering Herd problem).  Now, dependencies will always disable job creation. 


components/hab/src/main.rs
```diff
@@ -822,6 +822,14 @@ fn sub_pkg_upload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     // before allowing a write to the backend, this bypasses the check
     let force_upload = m.is_present("FORCE");
 
+    // Builder checks for presence of the 'builder' param to determine
+    // if a job group should be created. 
+    let disable_build_group = if m.is_present("NO_BUILD") {
```
chefsalim: It might be simpler/less code to just pass and use this as a bool like `force_upload`

components/hab/src/cli.rs
```diff
@@ -526,6 +526,8 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                 (@arg FORCE: --force "Skips checking availability of package and \
                     force uploads, potentially overwriting a stored copy of a package. \
                     (default: false)")
+                (@arg DISABLE_BUILD: --("disable-build")  "Do not create a build group on package upload. This flag has no \
```
chefsalim: @apriofrost may want to wordsmith, but I might just keep the phrasing simple and say "Do not auto build on package upload"

components/hab/src/command/pkg/upload.rs
```diff
@@ -52,6 +52,7 @@ pub fn start(ui: &mut UI,
              token: &str,
              archive_path: &Path,
              force_upload: bool,
+             disable_build_group: Option<&str>,
```
chefsalim: Getting into the nitpicky weeds.... `disable_build` naming would be shorter and simpler

components/builder-api-client/src/builder.rs
```diff
@@ -828,6 +829,10 @@ impl BuilderAPIProvider for BuilderAPIClient {
                .append_pair("checksum", &checksum)
                .append_pair("target", &target.to_string())
                .append_pair("forced", &force_upload.to_string());
+            
+            if let Some(val) = disable_build_group {
+              url.query_pairs_mut().append_pair("builder", val);
+            }
```
baumanj: Just an interesting rust FYI that `Option` and `Result` implement `IntoIterator`, which would allow you to do this as:
```rust
            url.query_pairs_mut()
               .extend_pairs(disable_build_group.map(|v| ("builder", v)));
```
Per my other comments, I think we should use a non-`Option` type here, but I thought this was handy to know.

components/hab/src/cli.rs
baumanj: I think boolean-type flags conform to the `--foo`/`--no-foo` pattern enhances  consistency and usability, making this `--no-build` (which is also a bit shorter).

components/builder-api-client/src/artifactory.rs
```diff
@@ -447,6 +447,7 @@ impl BuilderAPIProvider for ArtifactoryClient {
                    pa: &mut PackageArchive,
                    token: &str,
                    _force_upload: bool,
+                   _disable_build_group: Option<&str>,
```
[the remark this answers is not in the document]
chefsalim: I wouldn't optimize for the provider model as that is likely going away soon (since we;re not going the Artifactory plug in direction any longer)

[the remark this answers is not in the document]
baumanj: Good to know! Any reason not to take the Artifactory code out right away?

components/hab/src/cli.rs
apriofrost: Agree with @chefsalim . I'd suggest "Disable auto-build for all packages in this upload." to align with the language we use in the Builder UI.

components/builder-api-client/src/builder.rs
baumanj: Just to be clear, if there's a `builder=XXX` item in the query parameters, then that **dis**ables the auto-build functionality? That's really unexpected. Why is the builder API like that?

smacfarlane:
@baumanj (GH not letting me respond directly on your feedback :disappointed: )   I like the idea of an `enum` rather than a boolean to describe what that parameter is.  When I started to go down that road for implementation though, in some cases it started to read awkwardly: `disable_build = BuildOnUpload::Enable`. 

I thought about changing the language from `disable_build` to `auto_build`, but that would be misleading when reading the code as that parameter can't enable builds, it can only disable them. 

I've gone with a simple boolean for now,  what are your thoughts?

baumanj:
>I like the idea of an `enum` rather than a boolean to describe what that parameter is. When I started to go down that road for implementation though, in some cases it started to read awkwardly: `disable_build = BuildOnUpload::Enable`.

Can you show me what you mean? With the enum option, I think your code would look like:
```rust
if let BuildOnUpload::Disable = build_on_upload {
    url.query_pairs_mut().append_pair("builder", "true");
}
```
or
```rust
match build_on_upload {
    BuildOnUpload::Enable => {},
    BuildOnUpload::Disable => url.query_pairs_mut().append_pair("builder", "true"),
}
```
Neither of those look particularly awkward to me. Though maybe it's a question of being used to `if let` statements and how they use pattern matching. It's also possible to use a regular `if`:
```rust
if build_on_upload == BuildOnUpload::Disable {
    url.query_pairs_mut().append_pair("builder", "true");
}
```
Though, you'd have to derive `PartialEq` for the `BuildOnUpload` enum, and I'd consider it somewhat less idiomatic rust.

If you dislike those approaches, I think using `pub struct BuildOnUpload(pub bool)` is just as good as the enum approach. That gives:
```rust
if let BuildOnUpload(false) = build_on_upload {
    url.query_pairs_mut().append_pair("builder", "true");
}
```

> I thought about changing the language from `disable_build` to `auto_build`, but that would be misleading when reading the code as that parameter can't enable builds, it can only disable them.

That could be addressed by changing the enum to
```rust
pub enum BuildOnUpload {
    PackageDefault,
    Disable,
}
```
Possibly with a doc comment explaining things in more detail.

> I've gone with a simple boolean for now, what are your thoughts?

It's definitely better than the `Option<&str>` approach, and I'm not going to stand in your way if you're happy with it. I'd still prefer the readability and additional safety of having a specific type, though.

components/builder-api-client/src/artifactory.rs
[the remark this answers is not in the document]
chefsalim: The only reason to not take it out right away is to avoid complicating things in case we have to revert any of the other PRs that also touch that code (paranoia).

components/builder-api-client/src/builder.rs
chefsalim: When Builder does an upload it wants to disable the auto build (since it is doing those builds, and we don't want an infinite loop). The query param naming is historical.

smacfarlane:
> ```rust
> pub enum BuildOnUpload {
>     PackageDefault,
>     Disable,
> }
> ```

@baumanj That makes a lot more sense and have reworked it to use this pattern.  I was previously fixated on it coming from a boolean and being an enable/disable toggle.  Thanks for this advice :smiley_cat:

components/builder-api-client/src/builder.rs
baumanj: That makes more sense; thanks for the context.

smacfarlane:
In lieu of code tests at this time, I did a quick behavior test and verified if a job group was created in the builder UI and added #6857 to track the need for tests: 

Default behavior:
```
env RUST_LOG=debug cargo run --bin hab -- pkg upload ~/src/smacfarlane/habitat-empty-test-plans/results/smacfarlane-empty-plan-0.0.3-20190815204152-x86_64-linux.hart
...
[2019-08-15T21:04:59Z DEBUG habitat_http_client::api_client] POST https://bldr.habitat.sh/v1/depot/pkgs/smacfarlane/empty-plan/0.0.3/20190815204152?checksum=1421a9f398f8b014c83a0680fd6c6153e57a926d8be4f7329b29b33ff9577179&target=x86_64-linux&forced=false with ApiClient { endpoint: "https://bldr.habitat.sh/v1", inner: Client }
```
With `--no-build`:
```
env RUST_LOG=debug cargo run --bin hab -- pkg upload --no-build ~/src/smacfarlane/habitat-empty-test-plans/results/smacfarlane-empty-plan-0.0.3-20190815210934-x86_64-linux.hart
...
[2019-08-15T21:10:00Z DEBUG habitat_http_client::api_client] POST https://bldr.habitat.sh/v1/depot/pkgs/smacfarlane/empty-plan/0.0.3/20190815210934?checksum=8055508fc9ae5fdd839e6aa2c38214e4d8e63719160d4547858ab43de0733ee6&target=x86_64-linux&forced=false&builder=true with ApiClient { endpoint: "https://bldr.habitat.sh/v1", inner: Client }
```

PR #6844, opened 2019-08-13
Copy SSL certs into Windows studio

chefsalim:
This change copies over SSL certs from a hab cert cache from outside the studio when creating a new Windows studio. This allows certs to be seamlessly propagated into the Studio.

Signed-off-by: Salim Alam <seasalim>

### DIFF
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -1,4 +1,5 @@
-use crate::{command::studio::enter::ARTIFACT_PATH_ENVVAR,
+use crate::{command::studio::enter::{ARTIFACT_PATH_ENVVAR,
+                                     CERT_PATH_ENVVAR},
             common::ui::UI,
             error::{Error,
                     Result},
@@ -6,7 +7,8 @@ use crate::{command::studio::enter::ARTIFACT_PATH_ENVVAR,
                     env as henv,
                     fs::{cache_key_path,
                          CACHE_ARTIFACT_PATH,
-                         CACHE_KEY_PATH},
+                         CACHE_KEY_PATH,
+                         CACHE_SSL_PATH},
                     os::process,
                     package::target,
                     util::docker},
@@ -68,8 +70,13 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
                                    mnt_prefix,
                                    CACHE_KEY_PATH),];
     if let Ok(cache_artifact_path) = henv::var(ARTIFACT_PATH_ENVVAR) {
+        // Don't use Path::join here as "\" can cause problems in Docker mounts
         volumes.push(format!("{}:{}/{}",
                              cache_artifact_path, mnt_prefix, CACHE_ARTIFACT_PATH));
+    }
+    if let Ok(cache_ssl_path) = henv::var(CERT_PATH_ENVVAR) {
+        // Don't use Path::join here as "\" can cause problems in Docker mounts
+        volumes.push(format!("{}:{}/{}", cache_ssl_path, mnt_prefix, CACHE_SSL_PATH));
     }
     if !using_windows_containers
        && (Path::new(DOCKER_SOCKET).exists() || cfg!(target_os = "windows"))
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -17,6 +17,7 @@ use crate::{config,
 use habitat_core::AUTH_TOKEN_ENVVAR;
 
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
+pub const CERT_PATH_ENVVAR: &str = "CERT_PATH";
 
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
@@ -74,6 +75,19 @@ pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
         stdfs::create_dir_all(&artifact_path)?;
     }
 
+    let ssl_path = match henv::var(CERT_PATH_ENVVAR) {
+        Ok(p) => PathBuf::from(p),
+        Err(_) => {
+            let path = fs::cache_ssl_path(None::<&str>);
+            debug!("Setting {}={}", CERT_PATH_ENVVAR, path.display());
+            env::set_var(CERT_PATH_ENVVAR, &path);
+            path
+        }
+    };
+    if !ssl_path.is_dir() {
+        debug!("Creating ssl_path at: {}", ssl_path.display());
+        stdfs::create_dir_all(&ssl_path)?;
+    }
     inner::start(ui, args)
 }
 
--- a/components/studio/bin/hab-studio.ps1
+++ b/components/studio/bin/hab-studio.ps1
@@ -254,6 +254,16 @@ function New-Studio {
     $env:HAB_CACHE_KEY_PATH = Join-Path $env:FS_ROOT "hab\cache\keys"
   }
 
+  $env:HAB_CACHE_SSL_PATH = Join-Path $env:FS_ROOT "hab\cache\ssl"
+  if (!(Test-Path $env:HAB_CACHE_SSL_PATH)) {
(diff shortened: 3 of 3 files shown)


components/hab/src/command/studio/docker.rs
```diff
@@ -71,6 +73,9 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
         volumes.push(format!("{}:{}/{}",
                              cache_artifact_path, mnt_prefix, CACHE_ARTIFACT_PATH));
     }
+    if let Ok(cache_ssl_path) = henv::var(SSL_PATH_ENVVAR) {
+        volumes.push(format!("{}:{}/{}", cache_ssl_path, mnt_prefix, CACHE_SSL_PATH));
```
christophermaier: Instead of "{}/{}", should we use `Path::join` instead for cross-platform pathing (both here and above)?

components/hab/src/command/studio/enter.rs
```diff
@@ -74,6 +75,19 @@ pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
         stdfs::create_dir_all(&artifact_path)?;
     }
 
+    let ssl_path = match henv::var(SSL_PATH_ENVVAR) {
+        Ok(p) => PathBuf::from(p),
+        Err(_) => {
+            let path = fs::cache_ssl_path(None::<&str>);
+            debug!("Setting {}={}", SSL_PATH_ENVVAR, path.display());
+            env::set_var(SSL_PATH_ENVVAR, &path);
```
christophermaier: Setting the environment variable like this is a bit surprising, and potentially fragile. If we want to make sure there's a value for this environment variable set before we spawn the actual studio process, I think it would be better to set it in `inner::start`, where we manipulate the rest of the environment we're creating for the exec'd process.

I can see someone easily refactoring the environment code in there for better isolation, etc., and totally removing this SSL path in the process, since it was handled remotely.

components/hab/src/command/studio/enter.rs
```diff
@@ -17,6 +17,7 @@ use crate::{config,
 use habitat_core::AUTH_TOKEN_ENVVAR;
 
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
+pub const SSL_PATH_ENVVAR: &str = "SSL_PATH";
```
christophermaier: Should this be prefixed with `HAB_`?

components/studio/bin/hab-studio.ps1
```diff
@@ -254,6 +254,16 @@ function New-Studio {
...
+  $env:HAB_CACHE_SSL_PATH = Join-Path $env:FS_ROOT "hab\cache\ssl"
+  if (!(Test-Path $env:HAB_CACHE_SSL_PATH)) {
+    mkdir $env:HAB_CACHE_SSL_PATH | Out-Null
+  }
+
+  $sslCacheSourcePath = Join-Path $env:SystemDrive "hab\cache\ssl"
+  if (Test-Path $sslCacheSourcePath) {
+    Write-HabInfo "Populating SSL certificate cache at $env:HAB_CACHE_SSL_PATH"
+    Copy-Item -Path $sslCacheSourcePath\* -Destination $env:HAB_CACHE_SSL_PATH
+  }
```
christophermaier: Why was the `SSL_PATH` environment variable needed? It doesn't seem to be referenced in this code.

components/hab/src/command/studio/docker.rs
chefsalim: That seems reasonable, however I'm trying to make the minimum change possible in order to minimize the risks of additional refactoring.

components/hab/src/command/studio/enter.rs
chefsalim: Ah, this needs to be "CERT_PATH" to match what the studio is actually expecting

components/hab/src/command/studio/enter.rs
chefsalim: Good point, there's probably room to refactor all this, however I'd prefer to not do refactoring or make flow changes in this PR

components/studio/bin/hab-studio.ps1
chefsalim: ARTIFACT_PATH and CERT_PATH are currently only used in the Linux Studio.  We don't have exact parity between Windows and Linux Studios atm.

components/hab/src/command/studio/enter.rs
christophermaier: This is all new code, though.

components/hab/src/command/studio/enter.rs
```diff
@@ -17,6 +17,7 @@ use crate::{config,
 use habitat_core::AUTH_TOKEN_ENVVAR;
 
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
+pub const SSL_PATH_ENVVAR: &str = "CERT_PATH";
```
christophermaier: Might as well call this `CERT_PATH_ENVVAR` now.

components/studio/bin/hab-studio.ps1
christophermaier: If this PR is all about Windows, and Windows doesn't use `CERT_PATH`, why were the Rust changes above to add `CERT_PATH` required for this?

components/hab/src/command/studio/docker.rs
mwrock: If I am remembering correctly, `\` can cause problems in docker mounts.

baumanj: If that's the case, it would probably help to document that fact and make a function which takes care of this so that some later well-intentioned person doesn't break things later.

chefsalim: Cool, I'll add a comment for now

components/hab/src/command/studio/enter.rs
chefsalim: It's essentially exactly the same pattern as the code above is my point - I'm happy to open an issue to refactor as a separate PR

components/hab/src/command/studio/enter.rs
chefsalim: Fixed

components/studio/bin/hab-studio.ps1
chefsalim: The CERT_PATH changes are not strictly required for Windows (atm) - those look like a missed use case to supplement the Linux PR that was done earlier.  If and when we bring Windows into parity with Linux for the env, we will be able to leverage those for Windows also.

components/hab/src/command/studio/enter.rs
chefsalim: Ok

components/studio/bin/hab-studio.ps1
christophermaier: In that case, it might be nice to separate these into different PRs.

components/hab/src/command/studio/enter.rs
christophermaier: I took a look at the code above this... yup, same pattern :scream: 

I think we really do need to refactor all this (separate PR is fine), and put some effort into wrapping up https://github.com/habitat-sh/habitat/issues/6314 as well.

components/hab/src/command/studio/docker.rs
mwrock: before adding the comment we should verify. I could be misremembering or it is also possible docker has fixed this in recent versions.

mwrock: Either I misremembered or (more likely in this case) docker has fixed this because I just validated `\` works fine in volume mounts.

components/studio/bin/hab-studio.ps1
chefsalim: OK

chefsalim: Well nvm, I think this is going to be needed for the Windows Docker Studio scenario after all - specifically, in the error case in enter.rs (where there is no CERT_PATH env defined) we set the env var, and then in docker.rs we use that var for the volume mount.

components/hab/src/command/studio/docker.rs
chefsalim: Still probably safer to leave it as is than risk some regression scenario

PR #6845, opened 2019-08-13
Add job to run all `#[ignore]`d tests

baumanj:
Also, fix up broken doc tests and remove their `ignore` annotation.

### DIFF
--- a/test/run_cargo_test.ps1
+++ b/test/run_cargo_test.ps1
@@ -31,8 +31,15 @@ Install-Rustup $toolchain
 Install-RustToolchain $toolchain
 $CargoTestCommand = "cargo +$toolchain test $FeatureString -- $TestOptions"
 
-Write-Host "--- Running cargo +$toolchain test on $Component with command: '$CargoTestCommand'"
-cd components/$Component
+
+if($Component -eq "") {
+    $scope = "habitat workspace"
+} else {
+    $scope = $Component
+    cd components/$Component
+}
+Write-Host "--- Running cargo +$toolchain test on $scope with command: '$CargoTestCommand'"
+
 cargo +$toolchain version
 Invoke-Expression $CargoTestCommand
 
--- a/test/run_cargo_test.ps1
+++ b/test/run_cargo_test.ps1
@@ -32,7 +32,7 @@ Install-RustToolchain $toolchain
 $CargoTestCommand = "cargo +$toolchain test $FeatureString -- $TestOptions"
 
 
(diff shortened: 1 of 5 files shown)


test/run_cargo_test.ps1
```diff
@@ -31,8 +31,15 @@ Install-Rustup $toolchain
 Install-RustToolchain $toolchain
 $CargoTestCommand = "cargo +$toolchain test $FeatureString -- $TestOptions"
 
-Write-Host "--- Running cargo +$toolchain test on $Component with command: '$CargoTestCommand'"
-cd components/$Component
+
+if($Component -eq "") {
```
mwrock: if no `-Component` is passed, then `$Component` will be `$null` and not empty. So should be `if(!$Component)`.

baumanj: That looks better, thanks! I'll make that change shortly.

For my own edification, testing `-eq ""` [_seems_ to work as well](https://tio.run/##hZFBSwMxEIXv@RVjXGiLXb0LniriSUEKHqSHNJ3dBLNJTGYpRfrb12y6y7aCmEvCzPveDC/e7TFEhcZ0nRdBNDBnkM41rBWCFQ2Cq4DSW7rGO4uWgBxsEbatNnQLj1iJ1lDsqzb1M/0RKWhbb4rVCC0H1ycU1AbMci9ivqUItbvkRtmIvXrSzl5QeaeeBMJI/XqNsLtLm3XqDOhy6Ow1SbUpXnStyBzYgjFdzac9ocQv4HwB31n/HjRh@ezSAD5pdARsPB04O2b8auL/BW1rzMj9ObaI0nmEB@BKbDUJgr0Ln9ELiYkFNBF/KyevXJe76cPi3VnzyM5XK8sS3lprU1pDmDcFOWekEtqegnV2HJKiU2PO9zArVj3QR7w61WacdV15lnnvr2vrAu74Dw). Is that expected, or is there something I'm missing?

baumanj: Fixed in https://github.com/habitat-sh/habitat/pull/6845/commits/6098da247213d856147a65871d4532caff2af3ec

mwrock: Arg. thats right. It is initialized as a `[string]` and thus has `""` as a default value and not `$null`. sorry!!

baumanj: This approach works too, and seems more robust. And I learned something, so no need to apologize! Can I get an approval though?

PR #6846, opened 2019-08-13
Begin cleaning up scripts and locations of things

scotthain:

The purpose of this PR is to clean up the scripts and organize them in a pattern that is easily recognizable and makes it very clear which scripts belong to which pipeline, as well as separating out pipeline specific details from tests that run against the artifacts themselves.


.expeditor/scripts/shared.sh
```diff
@@ -33,3 +33,7 @@ curlbash_hab() {
     esac
     echo "--- :habicat: Hab binary set to $hab_binary"
 }
+
+get_toolchain() {
+  cat "rust-toolchain"
```
baumanj: This seems like it depends on the working directory of the caller to be the root of the repo. I'm not sure we can assume that. It might make it more robust to employ the source dir trick:
```bash
source_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
And then access `rust-toolchain` via its relative path.

scotthain: It does depend on running from the root fo the repo, are there any situations where we wouldn't be? I'm taking into account Makefile based execution, buildkite based execution, and running manually.

baumanj: This path resolution will depend on the working directory of the user, regardless of where any of the code doing the sourcing lives. So, if a user were to manually run any script which sourced this function from a directory outside the repo or from a sub directory within the repo, it wouldn't work.

scotthain: Right, I understand that, but is that a valid use case that we want to support?

baumanj: I've run into breakages exactly like this elsewhere in our project, so yes, I would. It's not terribly hard to support and I'd say it makes the code more robust, so why not?

build.ps1
```diff
@@ -32,7 +32,8 @@ param (
     [string]$Toolchain
 )
 $ErrorActionPreference="stop"
-. $PSScriptRoot\support\ci\shared.ps1
+. $PSScriptRoot\.expeditor\scripts\verify\shared.ps1
+. $PSScriptRoot\.expeditor\scripts\release_habitat\shared.ps1
```
mwrock: It should just need `verify\shared.ps1`

.expeditor/scripts/verify/shared.sh
```diff
@@ -17,17 +20,7 @@ get_rustfmt_toolchain() {
...
-
-# Get the version of the nightly toolchain we use for compiling,
-# running, tests, etc.
-get_nightly_toolchain() {
-    cat "$dir/../../RUST_NIGHTLY_VERSION"
-}
-
-get_toolchain() {
-  cat "$dir/../../rust-toolchain"
+  cat "RUSTFMT_VERSION"
```
baumanj: I think it would be more robust to use the relative pathing here (like `get_toolchain` in `.expeditor/scripts/shared.sh`)

build.ps1
scotthain: Oh gotcha oops!

.expeditor/scripts/verify/shared.sh
```diff
@@ -52,10 +51,10 @@ install_rust_toolchain() {
...
 }
 
-install_rustfmt() {
-  local toolchain="${1?toolchain argument required}"
-  install_rust_toolchain "$toolchain"
-  rustup component add --toolchain "$toolchain" rustfmt
+# Get the version of the nightly toolchain we use for compiling,
+# running, tests, etc.
+get_nightly_toolchain() {
+    cat "RUST_NIGHTLY_VERSION"
```
baumanj: I'd suggest relative pathing here as well

.expeditor/scripts/verify/run_cargo_test.sh
```diff
@@ -2,8 +2,7 @@
 
 set -eou pipefail
 
-# shellcheck source=./support/ci/shared.sh
-source ./support/ci/shared.sh
+source .expeditor/scripts/verify/shared.sh
```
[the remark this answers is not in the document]
scotthain: Is it needed? shellcheck passes without it so i'm not sure what it gains us in this instance.

.expeditor/end_to_end.pipeline.yml
```diff
@@ -20,7 +20,7 @@ steps:
 
   - label: "[:linux: test_hab_help_doesnt_install_hab_sup]"
     command:
-      - .expeditor/scripts/setup_environment.sh DEV
+      - .expeditor/scripts/end_to_end/setup_environment.sh DEV
```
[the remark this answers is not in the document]
scotthain: unfortunately not :(

.expeditor/scripts/verify/shared.sh
scotthain: I'll  fix this

.expeditor/scripts/verify/shared.sh
scotthain: Will fix

.expeditor/scripts/verify/run_cargo_test.sh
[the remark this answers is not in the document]
baumanj: It improves shellcheck's analysis by letting it access the sourced file. This can be used to resolving things like undefined variables and the like. Ideally we'd use it everywhere we source static files, but that's unrelated to this PR. I just don't think we should remove it since it does add value.

.expeditor/scripts/verify/shared.sh
```diff
@@ -52,10 +52,11 @@ install_rust_toolchain() {
...
 }
 
-install_rustfmt() {
-  local toolchain="${1?toolchain argument required}"
-  install_rust_toolchain "$toolchain"
-  rustup component add --toolchain "$toolchain" rustfmt
+# Get the version of the nightly toolchain we use for compiling,
+# running, tests, etc.
+get_nightly_toolchain() {
+    dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
```
baumanj: Calculating the relative dir can be done just once at the top of the file and shared between these various functions if you like. For example:
```bash
source_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
project_root="$source_dir/../../.."
```
Simplifying this function to
```bash
    cat "$project_root/RUST_NIGHTLY_VERSION"
```

Again, not necessary for correctness, just a suggestion if you find that more readable/maintainable.

scotthain:
@baumanj Updated the overall PR comment.

Issue #6816, opened 2019-08-08
Add existing end-to-end tests to pipeline


christophermaier:
Now that the new release pipeline is nearing completion, we have the framework in place to add more automated tests of the Supervisor in more real-world scenarios. 

We have some [existing scripted scenarios](https://github.com/habitat-sh/habitat/tree/master/test/end-to-end) that should be added to the [pipeline](https://github.com/habitat-sh/habitat/blob/master/.expeditor/end_to_end.pipeline.yml) 

There should be stub pipeline steps for most, if not all, of these tests already, but they don't all necessarily work yet, as they were not necessarily written expecting to run in this CI pipeline. If changes need to be made to run the tests in CI, try to make it possible to easily run the tests both in CI and locally for ease of developer usage. However, they _must_ run in CI.

# Acceptance Criteria
Existing tests are automated and run in the pipeline. The fact that they are all separate test files, or run in pipeline stages of their own, is less important than the fact that they just run. We can refine the overall framework in subsequent work. We just want the robots running them now.

These are the existing tests to automate; the issue is complete when all have been addressed:
- [ ] [`hup-does-not-abandon-services.exp`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/hup-does-not-abandon-services.exp)
- [ ] [`test_hab_help_doesnt_install_hab_sup.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_hab_help_doesnt_install_hab_sup.sh)
- [ ] [`test_launcher_checks_supervisor_version.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_launcher_checks_supervisor_version.sh)
- [ ] [`test_launcher_exits_on_supervisor_connection_failure.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_launcher_exits_on_supervisor_connection_failure.sh)
- [ ] [`test_launcher_exits_on_supervisor_startup_failure.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_launcher_exits_on_supervisor_startup_failure.sh)
- [ ] [`test_launcher_restarts_supervisor.sh`](https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_launcher_restarts_supervisor.sh)
[...]

PR #6849, opened 2019-08-14
End to end tests: hab-svc-load and hup-does-not-abandon-services

baumanj:
Add/fix these two tests for our end-to-end CI. This will merge back into https://github.com/habitat-sh/habitat/pull/6780.

See also https://github.com/habitat-sh/habitat/issues/6816

### DIFF
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -33,8 +33,8 @@ steps:
   - label: "[:linux: hup-does-not-abandon-services]"
     command:
       - .expeditor/scripts/setup_environment.sh DEV
-      - hab pkg install --binlink core/expect/5.45.4/20190115014137
-      - expect test/end-to-end/hup-does-not-abandon-services.exp
+      - hab pkg install --binlink --channel=stable core/expect
+      - test/end-to-end/hup-does-not-abandon-services.exp
     expeditor:
       executor:
         docker:
@@ -45,8 +45,8 @@ steps:
   - label: "[:linux: hab-svc-load]"
     command:
       - .expeditor/scripts/setup_environment.sh DEV
-      - hab pkg install --binlink core/expect/5.45.4/20190115014137
-      - expect test/end-to-end/hab-svc-load.exp
+      - hab pkg install --binlink --channel=stable core/expect
+      - test/end-to-end/hab-svc-load.exp
     expeditor:
       executor:
         docker:
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -42,6 +42,18 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
 
+  - label: "[:linux: hab-svc-load]"
+    command:
+      - .expeditor/scripts/setup_environment.sh DEV
+      - hab pkg install --binlink core/expect/5.45.4/20190115014137
+      - expect test/end-to-end/hab-svc-load.exp
+    expeditor:
+      executor:
+        docker:
+          privileged: true
+          environment:
+            - BUILD_PKG_TARGET=x86_64-linux
+
   - label: "[:linux: test_launcher_checks_supervisor_version]"
     command:
       - .expeditor/scripts/setup_environment.sh DEV
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -9,7 +9,7 @@ expeditor:
       timeout_in_minutes: 30
       env:
         HAB_ORIGIN: "core"
-        ACCEPTANCE_HAB_BLDR_URL: "https://bldr.acceptance.habitat.sh"
+        HAB_BLDR_URL: "https://bldr.acceptance.habitat.sh"
         HAB_BLDR_CHANNEL: "DEV"
         HAB_INTERNAL_BLDR_CHANNEL: "DEV"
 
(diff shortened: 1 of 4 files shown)


.expeditor/end_to_end.pipeline.yml
```diff
@@ -29,6 +29,30 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
 
+  - label: "[:linux: hup-does-not-abandon-services]"
+    command:
+      - .expeditor/scripts/setup_environment.sh DEV
+      - hab pkg install --binlink core/expect/5.45.4/20190115014137
```
christophermaier: The reason the fully-qualified ident is needed here is because the environment contains `HAB_BLDR_CHANNEL=DEV` (see line 13 above), and there isn't a `core/expect` in that channel.

Since the exact release of `expect` isn't really important here, we could simplify this to 

```sh
hab pkg install --binlink --channel=stable core/expect/5.45.4
```
or, simpler still, to

```sh
hab pkg install --binlink --channel=stable core/expect
```

test/end-to-end/hup-does-not-abandon-services.exp
```diff
@@ -99,7 +99,7 @@ expect {
         error "Redis not ready to accept connections!"
     }
 }
-set redis_pid [exec pgrep redis-server]
+set redis_pid [exec cat /hab/svc/redis/PID]
```
christophermaier: When I run this test locally, it will sometimes fail at this part; it looks like there might be a race between when the Supervisor writes this PID file and when the test tries to read it.

If I add a `sleep 1` immediately before this check, for instance, I see no failures.

test/end-to-end/hab-svc-load.exp
```diff
@@ -0,0 +1,73 @@
...
+        error "Supervisor appears not to have started, correctly"
+    }
+}
+
+# Let the control gateway come up. :/
+# Alternatively, use `nc -wv localhost 9632`
+sleep 1
+
+# Load up Redis on that Supervisor
+spawn hab svc load core/redis
```
christophermaier: Initially, I was surprised to see that this package installation was coming from `stable` and not `DEV`, but then I discovered that the `HAB_BLDR_CHANNEL` environment variable is ignored in `hab svc load`.

christophermaier:
Incidentally, I think you can rebase your work directly on `master` and merge from there. I don't think any of your changes are dependent on the ones from my PR.

.expeditor/end_to_end.pipeline.yml
baumanj: What's the advantage? I don't think `expect` is likely to change in a way that's relevant to this test. Pinning it has the advantage to eliminating a potential difference when running this in different environments.

test/end-to-end/hup-does-not-abandon-services.exp
baumanj: https://github.com/habitat-sh/habitat/pull/6849/commits/0d5487c41d036bf913827988d68fa58fa4cf3dd1 should address this

baumanj:
> Incidentally, I think you can rebase your work directly on `master` and merge from there. I don't think any of your changes are dependent on the ones from my PR.

Any reason *not* to merge back into your PR? All the e2e work is captured in the same issue, and there are some things in my changes that may be useful to the other e2e tests. Also, it'll just be less work for me to merge it as is since my branch was based on yours from the beginning.

.expeditor/end_to_end.pipeline.yml
christophermaier: It's just that when I see a pin, I immediately wonder what the significance of that specific release is.

test/end-to-end/hup-does-not-abandon-services.exp
christophermaier: That works.

Out of curiosity, what prompted you to move away from the `pgrep` approach to the PID file? Not having to worry about whether `pgrep` would be there?

christophermaier:
> > Incidentally, I think you can rebase your work directly on `master` and merge from there. I don't think any of your changes are dependent on the ones from my PR.
> 
> Any reason _not_ to merge back into your PR? All the e2e work is captured in the same issue, and there are some things in my changes that may be useful to the other e2e tests. Also, it'll just be less work for me to merge it as is since my branch was based on yours from the beginning.

If I were to rework and force-push my branch then that could mess yours up. They're not inherently coupled. If your PR has something I (or anyone else) can use, then I'll rebase mine when yours merges.

test/end-to-end/hup-does-not-abandon-services.exp
baumanj: As I mentioned in https://github.com/habitat-sh/habitat/pull/6849/commits/c53bfa1eaa8eee7a165a11e501fe82788b384e78:
> there may be other `redis-server`s running locally.

To wit:
```
jbauman@ubuntu:~/habitat$ pgrep -afl redis-server
77398 /opt/opscode/embedded/bin/redis-server 127.0.0.1:16379                         
```

.expeditor/end_to_end.pipeline.yml
baumanj: Fair enough. I'll try changing it.

baumanj:
I think I've addressed all the feedback now.

scotthain:
This is drive-by post-merge commentary that may have been resolved somewhere else, but it looks like `ACCEPTANCE_HAB_BLDR_URL` was changed to `HAB_BLDR_URL` and not all instances were change later in the pipeline.

PR #6850, opened 2019-08-14
Propperly root SetupEnvironment variables with file system paths at run time

mwrock:

### DIFF
--- a/components/plan-build-ps1/bin/hab-plan-build.ps1
+++ b/components/plan-build-ps1/bin/hab-plan-build.ps1
@@ -302,6 +302,21 @@ function Invoke-Begin {
 function Invoke-DefaultBegin {
 }
 
+function _Get-UnrootedPath($path) {
+  Push-Location $originalPath
+  try {
+    $path = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($path)
+  }
+  finally { Pop-Location }
+  $prefixDrive = (Resolve-Path $originalPath).Drive.Root
+  $strippedPrefix = $path
+  if($path.StartsWith($prefixDrive)) {
+    $strippedPrefix = $path.Substring($prefixDrive.length)
+  }
+  if(!$strippedPrefix.StartsWith('\')) { $strippedPrefix = "\$strippedPrefix" }
+  $strippedPrefix
+}
+
 # **Internal**  Build and export `$PATH` containing each path in our own
 # `${pkg_bin_dirs[@]}` array, and then any dependency's `PATH` entry (direct or
 # transitive) if one exists. The ordering of the path is specific to
@@ -311,8 +326,6 @@ function Invoke-DefaultBegin {
 # Reference implementation:
 # https://github.com/habitat-sh/habitat/blob/3d63753468ace168bbbe4c52e600d408c4981b03/components/plan-build/bin/hab-plan-build.sh#L1584-L1638
 function _Set-BuildPath {
-  $prefixDrive = (Resolve-Path $originalPath).Drive.Root
-
   $paths=@()
 
   # Add element for each entry in `$pkg_bin_dirs[@]` first
@@ -338,8 +351,7 @@ function _Set-BuildPath {
       # a `PATH` entry. This is necessary for packages created using a release
       # of Habitat between 0.53.0 (released 2018-02-05) and up to including
       # 0.55.0 (released 2018-03-20).
-      $strippedPrefix = $dep_prefix.Substring($prefixDrive.length)
-      if(!$strippedPrefix.StartsWith('\')) { $strippedPrefix = "\$strippedPrefix" }
+      $strippedPrefix = _Get-UnrootedPath $dep_prefix
 
       foreach ($line in (Get-Content (Join-Path $dep_prefix "RUNTIME_ENVIRONMENT"))) {
           $varval = $line.split("=")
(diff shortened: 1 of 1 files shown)


components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1467,71 +1506,34 @@ function __env_var_type($VarName) {
...
-    if(Test-Path $path_to_dep) {
-        $dep_path = (Get-Content $path_to_dep | Out-String).Trim()
-        if($dep_path.length -gt 0) {
-            $path = push_to_path $dep_path $path
-        }
-    }
-
-    # Finally, return the path
-    $path
-}
```
mwrock: This function was dead code. Callers were removed in https://github.com/habitat-sh/habitat/commit/30248d4ea7157daab2bce300473f8f3ff9af3eaa#diff-286d17c574b12110264d8a48b995792f

build.ps1
```diff
@@ -106,7 +106,7 @@ function Invoke-Build([string]$Path, [switch]$Clean, [string]$Command, [switch]$
             $nightly = ($toolchain -eq "nightly")
             Push-Location $PSScriptRoot
             try {
-                & $PSScriptRoot\test\run_cargo_test.ps1 -Component (Get-Component $path) -Features $Features -TestOptions $TestOptions -Nightly:$nightly
+                & $PSScriptRoot\.expeditor\scripts\verify\run_cargo_test.ps1 -Component (Get-Component $path) -Features $Features -TestOptions $TestOptions -Nightly:$nightly
```
mwrock: fallout from recent script cleanup

components/plan-build-ps1/habitat/plan.ps1
```diff
@@ -44,12 +43,6 @@ function Invoke-Build {
...
+    Copy-Item "$PLAN_CONTEXT\..\bin\*" "$pkg_prefix\bin" -Force
     Copy-Item "$bin" "$pkg_prefix\bin\$bin" -Force
 }
-
-# Turn the remaining default phases into no-ops
-function Invoke-Download {}
-
-function Invoke-Verify {}
-
-function Invoke-Unpack {}
```
mwrock: we use to have to provide empty functions and a dummy source file. We don't need to do that anymore.

components/core/src/package/install.rs
```diff
@@ -541,14 +561,15 @@ impl PackageInstall {
...
-    /// If no value of `RUNTIME_ENVIRONMENT` is found, return an empty `HashMap`.
-    fn runtime_environment(&self) -> Result<HashMap<String, String>> {
-        match self.read_metafile(MetaFile::RuntimeEnvironment) {
+    /// If no value of file is found, return an empty `HashMap`.
+    fn runtime_environment_file_map(&self, metafile: MetaFile) -> Result<HashMap<String, String>> {
+        match self.read_metafile(metafile) {
             Ok(ref body) => Self::parse_runtime_environment_metafile(body),
-            Err(Error::MetaFileNotFound(MetaFile::RuntimeEnvironment)) => Ok(HashMap::new()),
+            Err(Error::MetaFileNotFound(MetaFile::RuntimeEnvironment))
+            | Err(Error::MetaFileNotFound(MetaFile::RuntimeEnvironmentPaths)) => Ok(HashMap::new()),
```
davidMcneil: Could this just be `Err(Error::MetaFileNotFound(_)) => Ok(HashMap::new())` instead of checking for the specific file types?

components/core/src/package/install.rs
```diff
@@ -231,15 +231,27 @@ impl PackageInstall {
...
+
+        // release 0.85.0 introduces the RUNTIME_ENVIRONMENT_PATHS metadata
+        // that are environment variables containing file paths that are to be
+        // rooted under FS_ROOT. For backwards compatibility the key/value
+        // pairs are duplicated in RUNTIME_ENVIRONMENT so we will remove previously
+        // stored values with the same key.
+        let environment_paths =
+            self.runtime_environment_file_map(MetaFile::RuntimeEnvironmentPaths)?;
+        for (key, value) in environment_paths.into_iter() {
+            let mut split_path: Vec<PathBuf> = env::split_paths(&value).collect();
```
davidMcneil: I think `Vec<PathBuf>` could be just `Vec<_>`.

components/core/src/package/install.rs
```diff
@@ -231,15 +231,27 @@ impl PackageInstall {
...
+            let rooted_path = self.root_paths(&mut split_path)?;
+            if !rooted_path.is_empty() {
+                env.remove(&key);
+                env.insert(key, rooted_path);
```
davidMcneil: It looks like [`insert`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert) will replace the old value so I don/t think the `remove` is necessary.

components/core/src/package/install.rs
```diff
@@ -231,15 +231,27 @@ impl PackageInstall {
...
+        // that are environment variables containing file paths that are to be
+        // rooted under FS_ROOT. For backwards compatibility the key/value
+        // pairs are duplicated in RUNTIME_ENVIRONMENT so we will remove previously
+        // stored values with the same key.
+        let environment_paths =
+            self.runtime_environment_file_map(MetaFile::RuntimeEnvironmentPaths)?;
+        for (key, value) in environment_paths.into_iter() {
+            let mut split_path: Vec<PathBuf> = env::split_paths(&value).collect();
+            let rooted_path = self.root_paths(&mut split_path)?;
+            if !rooted_path.is_empty() {
```
davidMcneil: Do we ever expect this to be empty, and if it is, is there harm in adding it to `env`? Won't it already be set as empty in `RUNTIME_ENVIRONMENT` so it would already be set as empty in `env`? Just curious and trying to understand the code better.

mwrock: That's a great point. I was basically reproducing the code above that works with `PATH` but was wondering the same thing here. And you bring up a good point that even if it is empty the empty value would have gotten propagated in `RUNTIME_ENVIRONMENT`. I can't think of why it would be empty unless someone either manually manipulated the file or explicitly set the variable to an empty string in `Invoke-SetupEnvironment`. And if they did, they probably did so for a reason. So I'm going to remove this check.

mwrock:
great comments @davidMcneil I have applied them all.

robbkidd:
I can't speak to the implementation, but as an author of plans, I've been using this bugfix/feature all week from an experimental channel and I freaking love it.

all of my love

PR #6851, opened 2019-08-14
Pin to reqwest 0.9.17

christophermaier:
A change was added to the 0.9.18 release of reqwest that throws an
error if a blocking client is used in an asynchronous context (see
https://github.com/seanmonstar/reqwest/commit/5096e12fa21ffdf62ce953c7514e772959d1e4e1).

While this is technically correct, we need to augment our HTTP client
usage to be able to switch to an asynchronous mode of operation in the
Supervisor. Otherwise, we won't be able to, say, run `hab svc load`
and install a package, because that's performed asynchronously.

Pinning to the last good version of reqwest will allow us to proceed
until we can do the bigger fix.


christophermaier:
Note: the issue this fixes is not one that affects any current users of Habitat, since it was introduced in https://github.com/habitat-sh/habitat/pull/6759, which is unreleased at this time.

PR #6761, opened 2019-07-25
Fix event stream reconnect

davidMcneil:
This address three issues with the forked [ratsio library's](https://github.com/habitat-sh/ratsio) reconnect logic:

- [`CONNECT`](https://nats-io.github.io/docs/nats_protocol/nats-protocol.html#connect) messages were not being sent on reconnect
- The STAN client reconnect logic is not completed before we actually consider the client reconnected. If we try and publish messages before this is complete it causes the client to be invalidated by the server.
- The STAN reconnect logic occasionally hangs. A timeout was added as a temporary workaround.

The [ratsio library](https://github.com/habitat-sh/ratsio) may need additional vetting and cleanup. I am not sure if this should be completed as part of this PR or as follow on work.

Resolves #6737

Issue #6740, opened 2019-07-17
Make Habitat supervisor startup behavior with event stream reporting to Automate configurable 


ericcalabretta:
Currently if the Habitat Event Stream flags are fed to the supervisor, it required Automate to be reachable and available to successfully start or load services. 

Automate could be unavailable due to upgrades, network outages, or other issues. In these situations new supervisors would not be able to start & and if any existing supervisors had a reboot or restart event they would not be able to load/manage services. This could make a simple Automate outage cause production workloads to be impacted.

I recommend we make this behavior configurable so a user can choose if Habitat should fail to start or successfully start when Automate is unavailable 

The user could indicate their preference by an environment variable like this: 

`HAB_REQUIRE_AUTOMATE=false` 

If set the behavior should be that the habitat supervisor should always start regardless of Automate's availability. This will allow services managed by Habitat to function even if a problem exists with Automate at the moment. 

This will also allow user to use Habitat and its upgrade capabilities to ship fixes to problems that could resolve any Automate misconfigurations. For example if this is combined with the Effortless Infra pattern a user may need to ship a cookbook change via Habitat to remedy the solution. If the Habitat Supervisor doesn't start that may not be possible.

christophermaier:
One subtlety is that if a Supervisor is allowed to start in event stream mode, but without being able to connect to Automate, how do we (and should we?) differentiate between the following cases:

* User provided incorrect connection parameters
* User provided correct parameters, but Automate is unavailable for some reason

OkJaybird:
I know that our team would prefer something more like errors in the logs that indicate connection failures. Our big criteria is because we are going the Effortless design route, the hab supervisor is our hand into managing the system. So the sup & our services have to be able to come up, even if reporting isn't working properly. If the supervisor is up, we can ship updates and address the issue, but if not we could find ourselves dead in the water. Basically reporting is important, but it's secondary to being able to get the system up and running.

I think a different error logged based on the kind of issue that occurred would work well for us - i.e. authentication issue with Automate, Automate server wasn't reachable at all, etc.

If there is a strong desire to keep a blocking startup, maybe this could be flag-driven, so the consumer can pick the behavior needed.

Issue #6770, opened 2019-07-30
Spike: evaluate NATS client library alternatives


gpeers:
Evaluate alternatives to our NATS client library [mnetship/ratsio](https://github.com/mnetship/ratsio), and decide whether to adopt a new library or keep patching the current library. Must-have features include TLS (with support for self-signed certificates), which the ratsio library lacks. A library with streaming as well as messaging features would be ideal, but if we can't find this, decide in conjunction with the Automate EAS Observability team whether or not to proceed with a messaging-only implementation of the Habitat-Automate integration for now.

Related to #6514, #6769, #6737, #6761, #6605

gpeers:
07/30/19: I've contacted the EAS Observability team and Product regarding the messaging-only option.

PR #6717, opened 2019-07-10
Stop health check futures when upgrading the Supervisor

christophermaier:
When we shut down the Supervisor normally, we stop the long-running
control gateway future, and then spawn a bunch of service shutdown
futures. Those aren't "infinite" futures, so they'll eventually be
driven to completion. Then, we just wait for all the futures on the
reactor to finish and we shut down.

When we shut down for a Supervisor upgrade, however, we _don't_ shut
down the services. Since their health checks now run in an infinite
future, these futures will effectively prevent the Supervisor from
shutting down.

This commit introduces a "detach" method for services, which is now
called in the Supervisor upgrade scenario. It currently just stops
health checks, but can be augmented to do more as we transition more
functionality over to asynchronous futures.

Similarly, the existing "reattach" logic is now extracted into a
"reattach" method, which will now restart health checking.

Fixes #6712

Signed-off-by: Christopher Maier <cmaier@chef.io>

PR #6853, opened 2019-08-14
Update event stream

davidMcneil:
Resolves #6761 #6740 

This PR first addresses #6740 using the existing `natsio` and `nitox` nats clients. The PR then removes those clients and switches to using a non-streaming client `rust-nats`. In initial testing, `rust-nats` and non-streaming in general appear much more reliable. However, testing was done on a standalone nats server instead of directly with automate. This is blocked on automate supporting a plain [nats connection](https://github.com/chef/automate/issues/1124).

There are several TODOs in this PR. Those should be straightforward to address once we verify this is the direction we want to go for our nats client. #6770 is the spike to evaluate the nats clients.


davidMcneil:
This resolves #6740 by adding the `event-stream-connect-timeout` cli option. This can also be set with the `HAB_EVENT_STREAM_CONNECT_TIMEOUT` environment variable. This option takes a numeric value that represents the number of seconds to wait for an event stream connection before exiting the supervisor. The value `0` is treated specially. It indicates there is no timeout and the supervisor should immediately start regardless of the state of the event stream connection.

davidMcneil:
Unfortunately, we were not able to avoid forking the `rust-nats` library. We needed to make the `connect` method public and add support for auth token credentials. The fork can be found [here](https://github.com/habitat-sh/rust-nats)

components/sup/src/event/stream.rs
```diff
@@ -3,31 +3,81 @@ use crate::event::{Error,
...
-    let connect_method_is_timeout = connect_method.is_timeout();
+    let connection_is_timeout = connect_method.is_timeout();
+
+    // Note: With the way we are using the client, we will not respond to pings from the server
+    // (https://nats-io.github.io/docs/nats_protocol/nats-protocol.html#pingpong).
+    // Instead, we rely on publishing events to keep us connected to the server. This is completely
+    // valid according to the protocol, and in fact, the server will not send pings if there is
+    // other traffic from a client. If we have no events for an extended period of time, we will
+    // be automatically disconnected (because we do not respond to pings). When the next event comes
+    // in we will try to reconnect.
```
christophermaier: If we later decide we need to participate in the `PING/PONG` interaction, we can always compose the `event_handler` future with something like an [`Interval`](https://docs.rs/tokio-timer/0.2.11/tokio_timer/struct.Interval.html) to send a `PING` periodically if we don't get an actual event to send.

christophermaier: But just as a point of clarification: if we disconnect, the underlying NATS client library will handle the reconnect for us, correct? That's nothing that we need to manage ourselves?

components/sup/src/event/stream.rs
```diff
@@ -3,31 +3,81 @@ use crate::event::{Error,
...
+        }
+
+        let event_handler = event_rx.for_each(move |event: Vec<u8>| {
+                                        if let Err(e) = client.publish(HABITAT_SUBJECT, &event) {
+                                            error!("Failed to publish event, '{}'", e);
+                                        }
+                                        Ok(())
+                                    });
+
+        Runtime::new().expect("Couldn't create event stream runtime!")
```
christophermaier: It's probably worth noting here: the reason all this was initially running in a thread in the first place is that the `nitox` library had an issue where it didn't play well with other futures on its reactor. To work around that, I just put it off on its own reactor on a separate thread.

Since `rust-nats` presumably doesn't have that issue, we _could_ theoretically move all this into running directly on the Supervisor's main reactor. If we were to do that, however, we'd need to do it in such a way that we could cleanly shut it down when the Supervisor needs to come down, or run into the same underlying issue that was behind https://github.com/habitat-sh/habitat/issues/6712 and fixed by https://github.com/habitat-sh/habitat/pull/6717.

(It's perfectly fine to leave it as is; I'm just thinking it would be good to leave a comment here to ensure that any well-meaning refactoring engineer that comes after us knows what the trade-offs are.)

components/sup/src/event/stream.rs
```diff
@@ -11,7 +11,7 @@ use tokio::{prelude::Stream,
             runtime::current_thread::Runtime};
 
 /// All messages are published under this subject.
-const HABITAT_SUBJECT: &str = "habitat";
+const HABITAT_SUBJECT: &str = "habitat.event.healthcheck";
```
christophermaier: We send out more events than just healthchecks, though... if we're going to have different subjects per event, we'll need to handle that a little differently.

components/sup/src/event.rs
```diff
@@ -237,47 +231,20 @@ fn publish(mut event: impl EventMessage) {
...
+struct EventStream(Sender<Vec<u8>>);
 
 impl EventStream {
     /// Queues an event to be sent out.
-    fn send(&self, event: Vec<u8>) {
+    fn send(&mut self, event: Vec<u8>) {
         trace!("About to queue an event: {:?}", event);
-        if let Err(e) = self.0.unbounded_send(event) {
-            error!("Failed to queue event: {:?}", e);
+        if let Err(e) = self.0.try_send(event) {
```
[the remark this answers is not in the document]
davidMcneil: The current error we message we give when `try_send` fails due to a full channel is `Failed to queue event: send failed because channel is full`. Is there something more you would like to see?

components/sup/src/event/stream.rs
davidMcneil: Thanks for catching that! I will follow up with the automate team and see what they expect.

components/sup/src/event/stream.rs
davidMcneil: > But just as a point of clarification: if we disconnect, the underlying NATS client library will handle the reconnect for us, correct?

Correct, if a publish fails the library will try and reconnect and then republish.

components/sup/src/event.rs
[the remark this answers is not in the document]
christophermaier: Nope, that's good!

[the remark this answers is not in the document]
christophermaier: I think a comment in the code about dropping messages would still be useful as a documentation of intent.

components/sup/src/event/stream.rs
davidMcneil: These commits [here](https://github.com/habitat-sh/habitat/pull/6853/commits/15228661d18f3ff4adcde8d5072e66934b7e941c) and [here](https://github.com/habitat-sh/habitat/pull/6853/commits/e62b6ef813b84ecc5b302431eaad2325bf0a1918) address this.

It seems like we need a more robust and ergonomic solution for ensuring all futures are stopped before shutting down. This is outside the scope of this PR, but just to get some ideas out there.

What if we made a "wrapper" that wraps two runtimes. This "wrapper" could have a `spawn_divergent` (or something to indicate that this future does not end) call. This would spawn the future on a runtime that calls `shutdown_now` instead of `shutdown_on_idle` when we shutdown.

Not sold on this solution, but it would be nice to not have to keep a handle for every divergent future. I wonder how other tokio projects handle correctly ending unbounded futures. Thoughts?

christophermaier: Neat :smile: 

I'm not sure what the best way going forward should be for this. I like the "handle" approach, since it's explicit, but it does require a little bookkeeping. I haven't seen other approaches for this, though (which is what motivated that handle solution initially).

Your "two Runtimes" approach is also an interesting one, and is worth digging into, I think.

As long as the code in this PR doesn't get us back into a 0.83.0 bug situation, I'm :+1: on merging it.

ericcalabretta:
@davidMcneil if a user sets `HAB_EVENT_STREAM_CONNECT_TIMEOUT=0` Would the supervisor attempt to connect to Automate if it was not initially available? 

Automate may be unavailable because of an upgrade, failure, etc and users may want to still start the supervisor & services. It sounds like setting the value to `0` would accomplish this but users would also want the supervisor to connect to Automate when/if it becomes available. When the upgrade is complete or service was restored from whatever the failure was.

davidMcneil:
@ericcalabretta Regardless of the value of `HAB_EVENT_STREAM_CONNECT_TIMEOUT` habitat will always try and connect to automate when it goes to publish an event (if it is disconnected). `HAB_EVENT_STREAM_CONNECT_TIMEOUT` only impacts startup behavior. So if a value of `0` is used habitat will eventually connect to automate given it comes up at the correct url with the correct auth token.

ericcalabretta:
@davidMcneil That's perfect, thanks for the clarifications.

PR #6855, opened 2019-08-15
Add support for running e2e tests locally, for rapid iteration

christophermaier:
Adds an `e2e_local.sh` script that can be used to run an individual
end-to-end test script in an isolated Docker environment that mimics
what we use in CI.

Documentation is included, but the TL;DR is

```sh
./e2e_local.sh test/end-to-end/${YOUR_TEST_FILE}
```

### DIFF
--- a/e2e_local.sh
+++ b/e2e_local.sh
@@ -2,25 +2,32 @@
 
 set -euo pipefail
 
+if [[ ${#} == 0 ]]; then
+    cat <<EOF
+    You must supply arguments! For example:
+
+        ${0} test/end-to-end/my_test.sh
+
+EOF
+    exit 1
+fi
+
 # These are the commands that will be sequentially run in the
 # container. The final one will be the test itself, which is supplied
-# as the first argument to this script
+# as the arguments of the script.
 #
 # Not every test requires `expect`, but it's not hurting anything to
 # install it for everything.
 #
 # Note that the `ci-studio-common.sh` file is baked into the image we
 # use.
-raw_commands=(". /opt/ci-studio-common/buildkite-agent-hooks/ci-studio-common.sh"
-              ".expeditor/scripts/setup_environment.sh DEV"
-              "hab pkg install --binlink --channel=stable core/expect"
-              "${*}")
+commands=(". /opt/ci-studio-common/buildkite-agent-hooks/ci-studio-common.sh"
+          ".expeditor/scripts/setup_environment.sh DEV"
+          "hab pkg install --binlink --channel=stable core/expect"
+          "${*}")
 
-# Add a newline after every command, for feeding into the container.
-commands="${raw_commands[*]/%/\\n}"
-
-# The `${variable@E}` notation interprets the `\n` sequences as actual
-# newlines.
+# Add a `;` after every command, for feeding into the container. This
+# allows them to run in sequence.
 docker run \
        --rm \
        --interactive \
@@ -29,4 +36,4 @@ docker run \
        --env-file="$(pwd)/e2e_env" \
        --volume="$(pwd):/workdir" \
        --workdir=/workdir \
-       chefes/buildkite /bin/bash -e -c "${commands@E}"
+       chefes/buildkite /bin/bash -e -c "${commands[*]/%/;}"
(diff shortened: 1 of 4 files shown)


e2e_local.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Note that the `ci-studio-common.sh` file is baked into the image we
+# use.
+raw_commands=(". /opt/ci-studio-common/buildkite-agent-hooks/ci-studio-common.sh"
+              ".expeditor/scripts/setup_environment.sh DEV"
+              "hab pkg install --binlink --channel=stable core/expect"
+              "${*}")
+
+# Add a newline after every command, for feeding into the container.
+commands="${raw_commands[*]/%/\\n}"
```
baumanj: I think this should use `[@]` rather than `[*]`. Additionally, would it be simpler to append `;` to separate commands? I think that would eliminate the need for `${commands@E}` later. Eliminating that dependency would be nice, since it's a feature that doesn't exist in the version of `bash` that ships with macOS.

e2e_local.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -euo pipefail
+
```
baumanj: Some sort of help output when run without args (or with `-h` or `--help`) would be nice

e2e_local.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Not every test requires `expect`, but it's not hurting anything to
+# install it for everything.
+#
+# Note that the `ci-studio-common.sh` file is baked into the image we
+# use.
+raw_commands=(". /opt/ci-studio-common/buildkite-agent-hooks/ci-studio-common.sh"
+              ".expeditor/scripts/setup_environment.sh DEV"
+              "hab pkg install --binlink --channel=stable core/expect"
+              "${*}")
```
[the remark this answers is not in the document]
christophermaier: I was trying to keep it as close to CI as possible, while also encouraging us to move toward things being as self-contained and consistent as possible. Ideally, I think I'd like each test to just be a single executable file.

To that end, this actually started as `${1}`, but some test files currently take arguments.

e2e_local.sh
christophermaier: Looks like `;` works fine, too. The `\n` approach was in keeping with what Buildkite actually executes, but it's functionally equivalent and (as you observe) a bit more cross-platform friendly.

I'll make the change.

e2e_local.sh
christophermaier: Will do

e2e_local.sh
christophermaier: I looked at `*` vs `@` and it looks like it really does have to be `*`;

``` sh
 ${commands[@]/%/;}  # -> first 'command;' second 'command;' third 'command;'
 ${commands[*]/%/;}  # -> first 'command;' second 'command;' third 'command;'
"${commands[@]/%/;}" # -> 'first command;' 'second command;' 'third command;'
"${commands[*]/%/;}" # -> 'first command; second command; third command;'
```
Only the last one gives us what we want; a single string of delimited commands to pass into `bash -c`.

baumanj: If you only want to allow one command (with arguments) as input to `e2e_local.sh` and not have to put quotes around them; I agree. Using `@` is necessary if you want to allow more than one command on the CLI, but it would have to be invoked like `e2e_local.sh "cmd1 arg1" "cmd2 arg2"`

christophermaier: I think for this iteration, I'll leave the code as-is, since I think it'll be a good forcing function to move us toward more standardization.

If that's trickier to do in practice, then we can easily modify things as laid out in these comments.

baumanj:
> [would it make sense to move most of this to the `test/end-to-end` directory?](https://github.com/habitat-sh/habitat/pull/6855#pullrequestreview-275610061)

I never saw a response to this. Any thoughts, @christophermaier ?

christophermaier:
@baumanj it could... we'd need to juggle some pathing internally to get the right docker mounts, but that's just typing :joy:

baumanj:
Weigh that against whether people who work on the e2e tests are more likely to discover this stuff if it's more local to the directory, I guess. I'm fine with either.

PR #6862, opened 2019-08-18
Add conditional compliation flags for unsupported targets

### LINKED ISSUE #6899: Add conditional compliation flags for unsupported targets
See #6862 for previous work.

davidMcneil:
Resolves #6899 #6932 

This is the code portion of what is needed for supporting the development of unsupported targets. Once we agree that this is the correct approach, #6932 should be addressed.

This PR adds a feature flag to the `habitat_core` crate for each target (eg `x86_64-linux`, `x86_64-windows`...). It also adds a `supported_targets` feature which enables all supported targets, and `aarch64-linux` as an unsupported target.

There are currently some idiosyncracies with rust's feature flags:

1. You can not currently check if one of you of your dependencies has enabled a feature (see [here](https://stackoverflow.com/questions/57792943/can-you-test-if-the-feature-of-a-dependency-is-set-using-the-cfg-macro)). This requires us to also have feature flags for crates which depend on the conditional symbols of `habitat_core`. Currently, the `hab` crate is the only crate.
2. (See **Edit**) Because of point 1 and because the `--no-default-features` flag only applies to the current crate. `supported_targets` is not set by default in the `habitat_core` crate. This prevents the case that the default features are disabled in `hab` but would still be enabled in `habitat_core` leading to more complicated conditional compilation logic. `hab` also has a `supported_targets` feature and it _does_ set it as a default feature. So as long as you are using `habitat_core` in the habitat workspace all supported targets will be on by default.
3. Building a workspace cannot use the `--features` flag (see [here](https://github.com/rust-lang/cargo/issues/5015)). It would be nice if you could run `cargo build --features "hab/aarch64-linux"` from the workspace directory to enable the `aarch64-linux` target. Unfortunately, that is not currently supported. This means the easiest way to build with an unsupported target is to go to a crate that uses `hab. `, `habitat_sup` for example, and add `features = ["aarch64-linux"]` to the `hab` dependency. Another option would be to add `"aarch64-linux"` temporarily as a default feature of the `hab` crate.

**Edit** I needed to add `supported_targets` as default to feature to `habitat_core` in order to make ci pass. This means we lose the ability to easily disable `supported_targets` (you would need to set `default-features = false` each time `habitat_core` is used). I do not think this is a huge loss. I can not think of a time it would be useful to disable `supported_targets`.


components/core/src/package/target.rs
```diff
@@ -254,7 +269,7 @@ supported_package_targets! {
     /// [x]: https://en.wikipedia.org/wiki/X86
     /// [isa]: https://en.wikipedia.org/wiki/Instruction_set_architecture
     /// [x86_64]: https://en.wikipedia.org/wiki/X86-64
-    ("x86_64-darwin", X86_64_Darwin, X86_64_DARWIN, "x86_64", "macos");
+    ("x86_64-darwin", X86_64_Darwin, X86_64_DARWIN, "x86_64", "macos", cfg(not(feature = "")));
```
baumanj: This `cfg(not(feature = ""))` stuff is kind of weird. I wonder if it would be necessary if we took a slightly different approach. Rather than a general `unsupported_targets` feature, would we have specific features for the various targets that we don't support by default? Or, have every target be an opt-in feature, but make our default supported targets part of the default feature set.

Also, I wonder how much added complexity the significant macroification of components/core/src/package/target.rs is saving us from. I didn't do anything about it before because we didn't need to change it. But if we're making modifications I'm inclined to at least consider a move away from that.

davidMcneil: Completely agree with both of these points. We could even have family of targets (eg `arm_targets`) that could enable all the arm targets, but in reality, I think the macro should just be removed which simplifies a lot of this.

davidMcneil: After polishing this PR more, I decided to leave the macro. It does remove some duplication and does group all of the related settings in the code. By leaving it, it also greatly reduced the complexity of this PR. If there is a consensus that it should be removed, I would be happy to tackle that but maybe as a different PR.

davidMcneil:
@baumanj Thanks for looking it over! As I was making these changes, I discovered a crates feature flags are much more nuanced than I initially thought. Did you have specific questions? I am sure others will have similar questions. I would like to make it more clear if possible.

components/core/src/package/target.rs
```diff
@@ -83,13 +83,15 @@ macro_rules! supported_package_targets {
     ) => {
         const SUPPORTED_PACKAGE_TARGETS: &'static [PackageTarget] = &[
```
christophermaier: This name might need to be changed... it seems to mean "all targets we were compiled to know about", but now that probably needs clarification.

baumanj:
> Did you have specific questions? I am sure others will have similar questions. I would like to make it more clear if possible.

No specific questions. It's just not code that I'm very familiar with and the density of the macros make it hard to read.

components/hab/src/command/studio/docker.rs
```diff
@@ -300,19 +300,39 @@ fn image_identifier_for_active_target(using_windows_containers: bool) -> Result<
...
+        #[cfg(feature = "supported_targets")]
+        target::X86_64_DARWIN => target::X86_64_LINUX,
+        #[cfg(feature = "supported_targets")]
+        target::X86_64_LINUX => target::X86_64_LINUX,
+        #[cfg(feature = "supported_targets")]
+        target::X86_64_LINUX_KERNEL2 => target::X86_64_LINUX_KERNEL2,
+        #[cfg(feature = "supported_targets")]
+        target::X86_64_WINDOWS => target::X86_64_LINUX,
+        #[cfg(feature = "aarch64-linux")]
+        target::AARCH64_LINUX => panic!("{} is not supported", target::AARCH64_LINUX),
```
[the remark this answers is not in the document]
davidMcneil: I thought this was used to determine what docker image to use when we start a studio (I could easily be wrong). So this is saying we do not support a studio for `aarch64-linux`. We don't support Docker exports for ARM either, but it does not appear that that part of the code uses conditionals around the `target` enum.

christophermaier:
It would be good to have some definitive documentation for how this is supposed to be used by a developer before merging.

For instance, I just tried adding `aarch64-linux` to `supported_targets` to build, and set `PLAN_PACKAGE_TARGET=aarch64-linux` and got compilation errors in the Docker studio code because the `aarch64-linux` feature wasn't enabled. Even adding that to the default features seemed to not work for me :frowning:

davidMcneil:
@christophermaier Agreed documentation is needed. I tried to provide some initial directions in point number 3 of my first comment. This will definitely need to be cleaned up and copied to a more visible location. Did you specify `aarch64-linux` as a `default`/`supported_target` in the `habitat_core` or `hab` crate? Unfortunately, you have to specify it in the `hab` crate. I realize this seems backward. The reason is, you can not detect if a dependency of the current crate has a feature enabled. You have to create a feature that wraps the dependencies feature so you can check it. Certainly not ideal, and I am very open to ways to improve this.

components/hab/src/command/studio/docker.rs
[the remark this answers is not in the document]
christophermaier: Right... we don't support ARM studios right now; I misread the code :smile: Might be good to add a bit of "studio-specificity" to that panic message, just to be super explicit.

Does the docker export code need some conditionals then?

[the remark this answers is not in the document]
davidMcneil: Currently, the docker export code uses `#[cfg(unix)]` and `#[cfg(windows)]`. This should potentially be changed to use the target feature flags.

davidMcneil:
> I think it would be a good idea to just pull in work from #6932

[6fbd5b](https://github.com/habitat-sh/habitat/pull/6862/commits/6fbd5b2cfb019092abdb8e3fd1315a8363d9294a) adds documentation on how to produce a build that has symbols for an unsupported target as well as how to add an unsupported target.

PR #6887, opened 2019-08-21
Use propogate environment to ensure hab url gets passed -

scotthain:
into docker containers, rather than explicitly specifying it on every step.

More clarity on what this fixes:
Since the `ACCEPTANCE_HAB_BLDR_URL` variable was never removed from the steps, it was passing an empty string to the `HAB_BLDR_URL`, which made it default to prod builder.


scotthain:
Please ignore the E2E failure - it's not part of the testing here.

nellshamrell:
LGTM

PR #6908, opened 2019-08-27
Add script to repackage binaries and push to our S3 bucket

scotthain:
This will leverage our default distribution channels.


.expeditor/scripts/release_habitat/shared.sh
```diff
@@ -22,13 +26,20 @@ import_keys() {
         core
 }
 
-get_latest_pkg_version_in_channel() {
+get_latest_pkg_release_in_channel() {
```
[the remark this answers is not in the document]
scotthain: Yeah these two names are not great, the idea is to get the full release name `0.XX.Y/0339494949` vs `0.XX.Y`. I don't think `get_latest_pkg_in_release_channel` makes that clearer since it's not getting a package, it's getting a version.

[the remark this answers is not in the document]
smacfarlane: Ah, yeah that makes sense.   `get_latest_(something)_in_release_channel` where `something` indicates we're getting a `version + release` would be ideal but I don't know that we have a term to describe that pair of the identifier.   Maybe a short `# Returns the 'version/release' of the requested package in the release channel` comment would be sufficient?

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,125 @@
...
+tail -n+6 "${target_hart}" | \
+    tar --directory "${extract_dir}" \
+        --extract \
+        --xz \
+        --strip-components=6
+
+extracted_hab_binary="$(find "$extract_dir" \( -name hab -or -name hab.exe \) -type f)"
+pkg_target="$(tr --delete '\r' < "${extract_dir}"/TARGET)"
+pkg_arch="$(echo "$pkg_target" | cut -d '-' -f 1)"
+pkg_kernel="$(echo "$pkg_target" | cut -d '-' -f 2)"
```
smacfarlane: Is this intended to spit out `windows`, `linux` etc?  

If so, I don't think this will work for `x86_64-linux-kernel2`. I believe it can be safely changed to `-f2,3` to get both `linux` and `linux-kernel2`.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,125 @@
...
+
+if [[ $pkg_target == *"windows" ]]; then
+for file in "$(dirname "$extracted_hab_binary")"/*; do 
+  cp -p "$file" "$pkg_dir/"
+done
+else
+  cp -p "$extracted_hab_binary" "$pkg_dir/$(basename "$extracted_hab_binary")"
+fi
+
+echo "Compressing \`hab' binary"
```
smacfarlane: Looks like mixed quotes and/or smart quotes here?

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,125 @@
...
+# Name of the file to upload
+upload_artifact="$(basename "$pkg_artifact")"
+# Now strip the `release` so we have a 'latest'
+latest_artifact="$(basename "$pkg_artifact" | sed -E 's/[0-9]{14}/latest/')"
+
+echo "--- Uploading $upload_artifact and associated artifacts to S3"
+pushd "$(dirname "$pkg_artifact")" >/dev/null
+# FYI - the bucket name is not just for automate artifacts, and this will be fixed up later
+# Upload unstable/latest
+aws --profile chef-cd s3 cp "$upload_artifact" "s3://chef-automate-artifacts/unstable/latest/habitat/$latest_artifact" --acl public-read
```
dmccown: I'd recommend extracting a variable for this path. It will be easier to change in the future.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,125 @@
...
+aws --profile chef-cd s3 cp "$upload_artifact.asc" "s3://chef-automate-artifacts/unstable/latest/habitat/$latest_artifact.asc" --acl public-read
+aws --profile chef-cd s3 cp "$upload_artifact.sha256sum" "s3://chef-automate-artifacts/unstable/latest/habitat/$latest_artifact.sha256sum" --acl public-read
+# Upload versioned
+aws --profile chef-cd s3 cp "$upload_artifact" "s3://chef-automate-artifacts/files/habitat/$release_version/$upload_artifact" --acl public-read
```
dmccown: I'd recommend extracting a variable for this as well.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,125 @@
...
+${hab_binary} pkg install core/gzip
+${hab_binary} pkg install core/tar
+${hab_binary} pkg install core/wget
+${hab_binary} pkg install core/zip
+
+# Import packages@chef.io GPG signing key
+aws s3 cp s3://chef-cd-citadel/packages_at_chef.io.pgp packages_at_chef.io.pgp --profile=chef-cd
+gpg --import packages_at_chef.io.pgp
+
+#TODO make this nicer
```
dmccown: Remove the TODO unless you plan to update this in another PR in the near future.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
scotthain: good call. this code was taken directly from the bintray uploader  :D

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
scotthain: Yep! copy pasted from bintray uploader :P

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
scotthain: good call, tho we shouldn't need to change it.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
scotthain: Good eye - that was before I finished :)

.expeditor/scripts/release_habitat/shared.sh
```diff
@@ -22,13 +26,22 @@ import_keys() {
...
+        | jq -r '.ident | .version + "/" + .release')
     echo "${version}"
 }
 
+# Returns the semver version in the form of X.Y.Z
+get_latest_pkg_version_in_release_channel() {
+    local pkg_name="${1:?}"
+    local release
+    release=$(get_latest_pkg_release_version_in_release_channel "$pkg_name")
+    echo "$release" | cut -f1 -d"/"
```
christophermaier: In an effort to keep the `cut` calls to a minimum, which can get hard to reason about and are prone to breakage if the data going into them changes, can we instead put the retrieval of the "latest" JSON blob from builder in a function, and then have this function call that, piping the results through the more readable 

```sh
jq -r '.ident | .release'
```
?
(This is also subject to the same question above: could we just use a fully-qualified identifier here?)

VERSION_FAKE
```diff
@@ -1 +1 @@
-0.84.51
\ No newline at end of file
+0.86.0
```
christophermaier: Are we burning the 86 release line like we did with 84?

scotthain: Yep.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+extract_dir="$tmp_root/extract"
+mkdir -p "$extract_dir"
+
+echo "--- Downloading $release_version for $BUILD_PKG_TARGET"
+if [[ $BUILD_PKG_TARGET == *"darwin" ]]; then
+  artifact_name=$(buildkite-agent meta-data get MACOS_ARTIFACT)
+  buildkite-agent artifact download "$artifact_name" "$tmp_root/" --step "[:macos: build hab]"
+  mv "$tmp_root/$artifact_name" "$tmp_root/hab-$channel.hart"
+else
+  hab pkg exec core/wget wget "${HAB_BLDR_URL}/v1/depot/pkgs/core/hab/$release_version/download?target=$BUILD_PKG_TARGET" -O "$tmp_root/hab-$channel.hart"
```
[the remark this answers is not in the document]
scotthain: Sounds good!

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+popd
+
+# Name of the file to upload
+upload_artifact="$(basename "$pkg_artifact")"
+# Now strip the `release` so we have a 'latest'
+latest_artifact="$(basename "$pkg_artifact" | sed -E 's/[0-9]{14}/latest/')"
+
+echo "--- Uploading $upload_artifact and associated artifacts to S3"
+pushd "$(dirname "$pkg_artifact")" >/dev/null
+# FYI - the bucket name is not just for automate artifacts, and this will be fixed up later
```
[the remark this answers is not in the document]
scotthain: It's actually unknown if the bucket will change names, so I'd rather burn that bridge when we get to it.

.expeditor/scripts/release_habitat/shared.sh
```diff
@@ -22,13 +26,22 @@ import_keys() {
         core
 }
 
-get_latest_pkg_version_in_channel() {
+# Returns the full "release" version in the form of X.Y.Z/DATESTAMP
+get_latest_pkg_release_version_in_release_channel() {
```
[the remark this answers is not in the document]
scotthain: Unfortunately we need that to compare the major versions of the studio and hab to ensure we have both at this stage, so we need separate functions that get both the major and major/release.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+# Sign our artifact
+popd >/dev/null
+pushd "$(dirname "$pkg_artifact")" >/dev/null
+gpg --armor --digest-algo sha256 --default-key 2940ABA983EF826A --output "$(basename "$pkg_artifact").asc" --detach-sign "$(basename "$pkg_artifact")"
+popd
+
+# Name of the file to upload
+upload_artifact="$(basename "$pkg_artifact")"
+# Now strip the `release` so we have a 'latest'
+latest_artifact="$(basename "$pkg_artifact" | sed -E 's/[0-9]{14}/latest/')"
```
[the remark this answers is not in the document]
scotthain: I'd rather have that fail at this point. To cm's point - that would required a pretty big reshuffle of the order of code that could introduce more complications at this point.

.expeditor/scripts/release_habitat/build_component.ps1
```diff
@@ -53,17 +53,6 @@ Write-Host "--- Running hab pkg promote for $pkg_ident to channel $Env:HAB_BLDR_
...
-    Write-Host "--- :buildkite: Recording metadata $pkg_ident"
-    Invoke-Expression "buildkite-agent meta-data set 'hab-version-x86_64-windows' '$pkg_ident'"
-    Invoke-Expression "buildkite-agent meta-data set 'hab-release-x86_64-windows' '$pkg_release'"
-    Invoke-Expression "buildkite-agent meta-data set 'hab-artifact-x86_64-windows' '$pkg_artifact'"
-} Elseif ($component -eq 'studio') {
-    Write-Host "--- :buildkite: Recording metadata for $pkg_ident"
-    Invoke-Expression "buildkite-agent meta-data set 'studio-version-x86_64-windows' $pkg_ident"       
-} Else {
-    Write-Host "Not recording any metadata for $pkg_ident, none required."
-}
```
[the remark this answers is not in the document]
scotthain: The metadata isn't used anywhere else in the pipeline

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+pkg_dir="$build_dir/${archive_name}"
+
+echo "Copying $extracted_hab_binary to $(basename "$pkg_dir")"
+mkdir -p "$pkg_dir"
+mkdir -p "$tmp_root/results"
+
+if [[ $pkg_target == *"windows" ]]; then
+for file in "$(dirname "$extracted_hab_binary")"/*; do 
+  cp -p "$file" "$pkg_dir/"
+done
```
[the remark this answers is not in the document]
scotthain: Fixed!

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+if [[ $BUILD_PKG_TARGET == *"darwin" ]]; then
+  artifact_name=$(buildkite-agent meta-data get MACOS_ARTIFACT)
+  buildkite-agent artifact download "$artifact_name" "$tmp_root/" --step "[:macos: build hab]"
+  mv "$tmp_root/$artifact_name" "$tmp_root/hab-$channel.hart"
+else
+  hab pkg exec core/wget wget "${HAB_BLDR_URL}/v1/depot/pkgs/core/hab/$release_version/download?target=$BUILD_PKG_TARGET" -O "$tmp_root/hab-$channel.hart"
+fi
+
+target_hart="$tmp_root/hab-$channel.hart"
+tail -n+6 "${target_hart}" | \
```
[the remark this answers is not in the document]
scotthain: Fixed

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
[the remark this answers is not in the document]
baumanj: Would it fail cleanly, though? If we were do add more digits to the `release` component (something we've talked about), this would result in `latest_artifact` having something line `latest123` it it.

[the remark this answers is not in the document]
scotthain: I guess my question is - how likely would it be to change the datestamp for the release? that seems like a pretty fundamental thing that we would not change.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,133 @@
...
+else
+  # once https://github.com/habitat-sh/habitat/issues/6878 is released,
+  # we can use hab pkg download rather than this wget
+  hab pkg exec core/wget wget \
+    "${HAB_BLDR_URL}/v1/depot/pkgs/core/hab/$release_version/download?target=$BUILD_PKG_TARGET" \
+    -O "$tmp_root/hab-$channel.hart"
+fi
+
+target_hart="$tmp_root/hab-$channel.hart"
+if [[ $(head -n 1 "${target_hart}") == HART-1 ]]; then
```
baumanj: I think this check is backwards; HART-1 _is_ the format we know how to handle
```suggestion
if [[ $(head -n 1 "${target_hart}") != HART-1 ]]; then
```

scotthain: Derp. Thank you!

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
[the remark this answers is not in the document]
baumanj: We've been actively discussing it as a way to remediate the several times we've had collisions. I'm not aware of the current format being formally specified anywhere, so I think it definitely *could* change.

I don't think this code need to be maximally general or anything, but validating that what we find *is* the format we expect and producing a clear error otherwise would be more robust than assuming that format is inviolable.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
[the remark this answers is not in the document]
christophermaier: Might update the comment to reflect that, then.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+if [[ $pkg_target == *"windows" ]]; then
+for file in "$(dirname "$extracted_hab_binary")"/*; do 
+  cp -p "$file" "$pkg_dir/"
+done
+else
+  cp -p "$extracted_hab_binary" "$pkg_dir/$(basename "$extracted_hab_binary")"
+fi
+
+echo "Compressing 'hab' binary"
+pushd "$build_dir" >/dev/null
```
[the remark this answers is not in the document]
christophermaier: I think this would help increase the readability and modularity of the code, and would like to see it in the final code.

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,126 @@
...
+case "$pkg_target" in
+*-linux | *-linux-kernel2)
+    pkg_artifact="$tmp_root/results/${archive_name}.tar.gz"
+    tarball="$build_dir/$(basename "${pkg_artifact%.gz}")"
```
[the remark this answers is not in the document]
christophermaier: I _always_ have to look these up :sob:

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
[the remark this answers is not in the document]
christophermaier: Ping :point_up:

.expeditor/scripts/release_habitat/shared.sh
```diff
@@ -22,13 +26,22 @@ import_keys() {
...
     local pkg_name="${1:?}"
     version=$(curl -s "${HAB_BLDR_URL}/v1/depot/channels/core/$(get_release_channel)/pkgs/${pkg_name}/latest?target=${BUILD_PKG_TARGET}" \
-        | jq -r '.ident | .version')
+        | jq -r '.ident | .version + "/" + .release')
     echo "${version}"
```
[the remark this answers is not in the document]
christophermaier: Ping :point_up:

.expeditor/scripts/release_habitat/shared.sh
christophermaier: Any thought on this suggestion?

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -82,37 +82,36 @@ else
...
-*)
-    exit_with "$target_hart has unknown TARGET=$BUILD_PKG_TARGET" 3
-    ;;
-esac
-
+(cd "$build_dir"
+  case "$BUILD_PKG_TARGET" in
+  *-linux | *-linux-kernel2)
+      pkg_artifact="$tmp_root/results/${archive_name}.tar.gz"
+      tarball="$build_dir/$(basename "${pkg_artifact}.gz")"
```
christophermaier: I think this needed to be `basename "${pkg_artifact}" .gz` (note the space).

Or, perhaps more readably, `basename --suffix=".gz" "${pkg_artifact}"`

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
```diff
@@ -0,0 +1,135 @@
...
+else
+  cp -p "$extracted_hab_binary" "$pkg_dir/$(basename "$extracted_hab_binary")"
+fi
+
+echo "Compressing 'hab' binary"
+(cd "$build_dir"
+  case "$BUILD_PKG_TARGET" in
+  *-linux | *-linux-kernel2)
+      pkg_artifact="$tmp_root/results/${archive_name}.tar.gz"
+      tarball="$build_dir/$(basename "${pkg_artifact}.gz")"
```
baumanj: This won't strip the `.gz`; the suggestion I made in https://github.com/habitat-sh/habitat/pull/6908#discussion_r321530190 was slightly different:
```suggestion
      tarball="$build_dir/$(basename "${pkg_artifact} .gz")"
```

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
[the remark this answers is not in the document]
baumanj: I think it would be good to address this in some way

.expeditor/scripts/release_habitat/package_and_upload_binary.sh
scotthain: good call

baumanj: [That syntax isn't POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html), but maybe it doesn't matter here

christophermaier: Well, poo... I like long options, but if we wanna stay POSIX, then that's cool too.

Issue #6799, opened 2019-08-07
[certs] Expose cert info in the hab cli UI


chefsalim:
Related to https://github.com/habitat-sh/habitat/pull/6759

Currently we only expose information about the certs that the hab cli is using via debug statements. This might be good to expose to users as part of the regular UI status.

smacfarlane:
After some initial research, we'll need to add the `habitat-common` crate as a dependency to `http-client` in order to use the UI components.  This adds a circular dependency `http-client -> common -> builder-api-client -> http-client`.

It appears that the `builder-api-client` is a dependency of `habitat-common` in order for `common::command::package::install` to manage package installs.  This file is depended on by `hab-sup`, `hab-pkg-export-tar` and `hab-pkg-export-docker` so any refactoring to remove the circular dependency would likely be a pretty impactful change.

dmccown:
This issue is looking a lot bigger than we thought, so we're going to remove this from this iteration.

krasnow:
We are unlikely do this work. Closing the issue.

PR #6909, opened 2019-08-27
Copy SSL_CERT_FILE to studio internal and persist env var

### LINKED ISSUE #6900: [certs] Auto copy SSL_CERT_FILE target to hab ssl cache
This is a continuation of the SSL cert cache changes for hab.

This change will allow auto propagation of a custom cert that is pointed to by the SSL_CERT_FILE environment variable into the /hab/cache/ssl or ~/.hab/cache/ssl folder, and then into the studio.

smacfarlane:
Closes #6900 

This will copy an ssl certificate specified by the SSL_CERT_FILE env variable to the inside of the studio, and set the variable on the inside of the studio to the  internal cache.

@chefsalim  I think we'll want to pull the certificate copy back into `studio::enter` in the Rust code.  It will cut down on the duplication between shell/powershell and I _think_ it will need to live there anyways in order for the Docker studio to have the same functionality. 

I'm going to work on the powershell implementation and add some tests in the mean time. 

### DIFF
--- a/components/hab/src/command/studio/docker.rs
+++ b/components/hab/src/command/studio/docker.rs
@@ -1,5 +1,6 @@
 use crate::{command::studio::enter::{ARTIFACT_PATH_ENVVAR,
-                                     CERT_PATH_ENVVAR},
+                                     CERT_PATH_ENVVAR,
+                                     SSL_CERT_FILE_ENVVAR},
             common::ui::UI,
             error::{Error,
                     Result},
@@ -99,7 +100,8 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
                             String::from("HAB_UPDATE_STRATEGY_FREQUENCY_MS"),
                             String::from("http_proxy"),
                             String::from("https_proxy"),
-                            String::from("RUST_LOG"),];
+                            String::from("RUST_LOG"),
+                            String::from("SSL_CERT_FILE"),];
 
     for (key, _) in env::vars() {
         if key.starts_with(HAB_STUDIO_SECRET) {
@@ -114,6 +116,10 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
         args.remove(index);
     }
 
+    // When a user sets SSL_CERT_FILE, we need to modify the absolute
+    // path to the file to reflect the location of the file inside the studio
+    update_ssl_cert_file_envvar(&mnt_prefix);
+
     // Windows containers do not use filesystem sharing for
     // local mounts
     if !using_windows_containers {
@@ -125,6 +131,23 @@ pub fn start_docker_studio(_ui: &mut UI, args: &[OsString]) -> Result<()> {
                   env_vars.iter(),
                   image,
                   using_windows_containers)
+}
+
+fn update_ssl_cert_file_envvar(mnt_prefix: &str) {
+    if let Ok(ssl_cert_file) = env::var(SSL_CERT_FILE_ENVVAR) {
+        if let Some(cert_file_name) = Path::new(&ssl_cert_file).file_name() {
+            if let Some(cert_file_name) = cert_file_name.to_str() {
+                // Don't use Path::join here in order to work around platform
+                // differences with paths on Windows with linux containers enabled
+                env::set_var(SSL_CERT_FILE_ENVVAR,
+                             format!("{}/{}/{}", mnt_prefix, CACHE_SSL_PATH, cert_file_name));
+            } else {
+                warn!("Unable to format {:?} for use inside studio", ssl_cert_file);
+            }
+        } else {
+            warn!("Invalid SSL_CERT_FILE value: {:?}", ssl_cert_file);
+        }
+    }
 }
 
 fn is_image_present(docker_cmd: &Path, image: &str) -> bool {
@@ -316,11 +339,17 @@ fn image_identifier(windows_base_tag: Option<&str>, target: target::PackageTarge
 #[cfg(test)]
 mod tests {
     use super::{image_identifier,
+                update_ssl_cert_file_envvar,
                 DOCKER_IMAGE,
                 DOCKER_WINDOWS_IMAGE};
     use crate::VERSION;
 
-    use crate::hcore::package::target;
+    use crate::{command::studio::enter::SSL_CERT_FILE_ENVVAR,
+                hcore::{fs::CACHE_SSL_PATH,
+                        package::target}};
+
+    use lazy_static::lazy_static;
+    habitat_common::locked_env_var!(SSL_CERT_FILE, lock_ssl_cert_file_env_var);
 
     #[test]
     fn retrieve_image_identifier() {
@@ -338,5 +367,33 @@ mod tests {
         assert_eq!(image_identifier(Some("ltsc2016"), target::X86_64_LINUX),
                    format!("{}-{}:{}-{}",
                            DOCKER_WINDOWS_IMAGE, "x86_64-windows", "ltsc2016", VERSION));
+    }
+
+    #[test]
+    fn update_ssl_cert_file_envvar_not_set() {
+        let mnt_prefix = "";
+        let env_var = lock_ssl_cert_file_env_var();
+        env_var.unset();
+
+        update_ssl_cert_file_envvar(mnt_prefix);
+        assert!(std::env::var(SSL_CERT_FILE_ENVVAR).is_err());
+    }
+
+    #[test]
+    fn update_ssl_cert_file_envvar_has_value() {
+        let mnt_prefix = "/some/prefix";
+        let key_name = "ssl-test-cert.pem";
+
+        let env_var = lock_ssl_cert_file_env_var();
+        env_var.set(format!("/path/to/{}", key_name));
+
+        update_ssl_cert_file_envvar(mnt_prefix);
+
+        // Don't use Path::join here because we format! the path above,
+        // in order to work around platform differences with paths on
+        // windows with linux containers enabled
+        let internal_cert_path = format!("{}/{}/{}", mnt_prefix, CACHE_SSL_PATH, key_name);
+
+        assert_eq!(std::env::var(SSL_CERT_FILE_ENVVAR), Ok(internal_cert_path));
     }
 }
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -1,7 +1,8 @@
 use std::{env,
           ffi::OsString,
           fs as stdfs,
-          path::PathBuf};
+          path::{Path,
+                 PathBuf}};
 
 use crate::{common::ui::UI,
             hcore::{crypto::CACHE_KEY_PATH_ENV_VAR,
@@ -9,15 +10,18 @@ use crate::{common::ui::UI,
                     fs}};
 
 use crate::{config,
-            error::Result,
+            error::{Error,
+                    Result},
             BLDR_URL_ENVVAR,
             CTL_SECRET_ENVVAR,
             ORIGIN_ENVVAR};
 
 use habitat_core::AUTH_TOKEN_ENVVAR;
+use same_file::is_same_file;
 
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 pub const CERT_PATH_ENVVAR: &str = "CERT_PATH";
+pub const SSL_CERT_FILE_ENVVAR: &str = "SSL_CERT_FILE";
 
 const STUDIO_CMD: &str = "hab-studio";
 const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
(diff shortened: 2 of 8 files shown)


components/hab/src/command/studio/enter.rs
```diff
@@ -43,6 +44,29 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
...
+                debug!("Unable to determine file name for SSL_CERT_FILE={}",
+                       &cert_file);
+                return;
+            }
+        };
+        let cache_file = cert_cache_dir.join(&cert_filename);
+        debug!("Caching SSL_CERT_FILE {}", &cert_file);
+
+        if let Err(err) = stdfs::copy(cert_file, &cache_file) {
+            debug!("Unable to cache SSL_CERT_FILE: {}", err);
```
mwrock: seems like this should be a warn or even error

components/hab/src/command/studio/enter.rs
```diff
@@ -43,6 +44,29 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
...
+        }
+    } else {
+        debug!("Invalid path for SSL_CERT_FILE: {}", &cert_file);
```
mwrock: I think this should be a warn or error too and I'd also add something like "check to make sure that file exists"

components/studio/bin/hab-studio.ps1
```diff
@@ -455,6 +455,17 @@ function Set-Secrets {
   }
 }
 
+function Update-SslCertFile {
+  if($env:SSL_CERT_FILE) {
+    try {
+      $cert_filename = (Get-Item $env:SSL_CERT_FILE).Name
+      $env:SSL_CERT_FILE = Join-Path "Habitat:\hab\cache\ssl" $cert_filename
```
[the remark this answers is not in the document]
mwrock: similar to linux it wil be scoped for the current process and all child processes. When you enter a studio it starts a new pwsh process so this will not leak into the parent (non-studio) environment.

mwrock: Should be `$env:SSL_CERT_FILE = Join-Path $env:HAB_CACHE_SSL_PATH $cert_filename`
Only powershell understands the `Habitat:\` drive but `$env:HAB_CACHE_SSL_PATH` is the cannonical path.

components/hab/src/command/studio/enter.rs
```diff
@@ -43,6 +44,29 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
...
+    if cert_path.exists() && cert_path.is_file() {
+        let cert_filename = match cert_path.file_name() {
+            Some(name) => name,
+            None => {
+                debug!("Unable to determine file name for SSL_CERT_FILE={}",
+                       &cert_file);
+                return;
+            }
+        };
+        let cache_file = cert_cache_dir.join(&cert_filename);
```
baumanj: Since we only execute this code if `cert_path.is_file()` returns true, `cert_path.file_name()` is guaranteed to return `Some(_)`, meaning we can eliminate the whole `match` block and just do:
```rust
        let cache_file = cert_cache_dir.join(cert_path.file_name().unwrap());
```
or, to be a bit more careful:
```rust
        let cert_filename = cert_path.file_name().expect("is file");
        let cache_file = cert_cache_dir.join(cert_filename);
```

components/hab/src/command/studio/enter.rs
```diff
@@ -43,6 +44,29 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
...
+        debug!("Caching SSL_CERT_FILE {}", &cert_file);
```
baumanj:
```suggestion
        debug!("Caching SSL_CERT_FILE {}", cert_file);
```

components/hab/src/command/studio/enter.rs
baumanj:
```suggestion
        debug!("Invalid path for SSL_CERT_FILE: {}", cert_file);
```

smacfarlane: @mwrock @baumanj Do you think it is worth passing the `UI` into this function so we can emit this information as part of the normal output, or is `warn!` sufficient?

mwrock: You could pass the `UI` which I think is better than `warn!` but perhaps even better whould be to have `cache_ssl_cert_file` return a `Result` and then `start` deals with the err using its `UI`.

chefsalim: Note there is a separate card to plumb in the cert status into the UI so might be better to do that work together instead of in this PR

components/studio/bin/hab-studio.sh
```diff
@@ -882,6 +881,12 @@ chroot_env() {
     env="$env no_proxy=$(echo "$no_proxy" | $bb sed 's/, /,/g')"
   fi
 
+  if [ -n "${SSL_CERT_FILE:-}" ] \
+  && [ -z "${NO_MOUNT}" ] \
+  && [ -z "${NO_ARTIFACT_PATH}" ]; then
```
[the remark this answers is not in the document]
chefsalim: Hmm, the NO_ARTIFACT_PATH test seems unnecessary.  I'd be concerned about changing an existing var that might break backward compat.  Maybe just remove the NO_ARTIFACT_PATH line here?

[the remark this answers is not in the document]
baumanj: I can't seem to get `-N` or setting the `NO_ARTIFACT_PATH` environment variable to actually do what it purports to do:
```
➤ hab studio --help
…
COMMON FLAGS:
…
    -N  Do not mount the source artifact cache path into the Studio (default: mount the path)
ENVIRONMENT VARIABLES:
…
    NO_ARTIFACT_PATH       If set, do not mount the source artifact cache path (`-N' flag overrides)
```
```
➤ hab studio rm
…
➤ env DEBUG=1 hab studio -N enter
…
[1][default:/src:0]# ls /hab/cache/artifacts/
core-acl-2.2.53-20190115012136-x86_64-linux.hart             core-gzip-1.9-20190115013612-x86_64-linux.hart               core-openssl-1.0.2r-20190305210149-x86_64-linux.hart
…
```
even though when I grep the debug output, I see:
```
+ NO_ARTIFACT_PATH=true
```

So, we should probably figure out whether that's a bug and if it has impact here. If it's only supposed to apply to `/hab/cache/artifacts` and will leave `/hab/cache/ssl` alone, I agree that we should just omit that check here, but I got the sense that it may have been intended to control the mounting of the entire `/hab/cache` directory and that the name was from when we didn't have other things cached, but I'm really not sure.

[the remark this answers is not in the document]
smacfarlane: This looks like a copy/paste gone wrong,  it should be `NO_CERT_PATH`, which I do think we want in this series of checks.  

@baumanj  I wasn't able to replicate the behavior you were seeing of `NO_ARTIFACT_PATH`, would you mind opening an issue on that?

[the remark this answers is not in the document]
baumanj: What do you see when you pass `-N` or set `NO_ARTIFACT_PATH`, @smacfarlane ?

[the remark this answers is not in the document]
smacfarlane: @baumanj If I enter an existing studio, I see an empty directory. If I destroy and create as you did, then I see the TDEPS of the studio, launcher and supervisor which is what I would expect.   `NO_ARTIFACT_CACHE` only controls the mounting of the cache from the host.  hab will still use `/hab/cache/artifacts` to cache everything it downloads, but the artifacts aren't shared with the host so will need to be fetched from builder and won't saved after the studio is removed.   Sorry I wasn't more clear this morning.

smacfarlane:
> The powershell looks good to me. Does `hab` spit out the `warn!` with the default logging level?

It does not,  but I'd like to address that in #6799

components/hab/src/command/studio/enter.rs
```diff
@@ -272,3 +295,89 @@ mod inner {
...
+
+        cache_ssl_cert_file(&ssl_cert_dir.to_str().unwrap(), &cert_cache_dir);
+
+        assert!(!cert_cache_dir.join(needs_a_name).exists());
+
+        Ok(())
+    }
+
+    #[test]
+    fn cache_ssl_cert_file_invalid_cert_cache_dir() -> std::io::Result<()> {
```
baumanj: Sorry, I'm not sure what I was thinking with suggesting this particular test. Since `cache_ssl_cert_file` doesn't report any errors, there's nothing observable that could happen with passing an invalid `cert_cache_dir`. We can probably remove this test altogether.

components/hab/src/command/studio/docker.rs
```diff
@@ -339,4 +364,30 @@ mod tests {
...
+
+        let env_var = lock_ssl_cert_file_env_var();
+        env_var.set(format!("/path/to/{}", key_name));
+
+        update_ssl_cert_file_envvar(mnt_prefix);
+        let internal_cert_path = Path::new(CACHE_SSL_PATH).join(key_name)
+                                                          .to_str()
+                                                          .unwrap()
+                                                          .to_string();
+        assert_eq!(std::env::var(SSL_CERT_FILE_ENVVAR), Ok(internal_cert_path));
```
baumanj: This test would be more valuable if `mnt_prefix` weren't an empty string. As is it is now, we can break `update_ssl_cert_file_envvar` by changing
```rust
env::set_var(SSL_CERT_FILE_ENVVAR,
             Path::new(mnt_prefix).join(CACHE_SSL_PATH)
                                  .join(cert_file_name));
```
to
```rust
env::set_var(SSL_CERT_FILE_ENVVAR,
             Path::new(CACHE_SSL_PATH).join(cert_file_name));
```
and the test still passes

components/hab/src/command/studio/enter.rs
```diff
@@ -272,3 +295,89 @@ mod inner {
...
+        File::create(&ssl_cert_filepath)?;
+
+        cache_ssl_cert_file(&ssl_cert_filepath.to_str().unwrap(),
+                            &invalid_cert_cache_dir);
+
+        assert!(!invalid_cert_cache_dir.exists());
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
smacfarlane: I've added a guard in the `cache_ssl_cert_file` function, but I'm not sure how we could test that.  Since `fs::copy` isn't guaranteed to truncate the file, inspecting the contents would likely intermittently fail.  There aren't any other side effects that I can think of to inspect, since the function is essentially a no-op in this case.

smacfarlane:
@baumanj  I think I've addressed all your feedback, with the exception of the same-file test.  Any advice on how to proceed on that one would be appreciated.

components/hab/src/command/studio/enter.rs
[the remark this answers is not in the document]
baumanj: A test that intermittently fails in the face of incorrect logic is preferable to no test at all. Even better would be making `cache_ssl_cert_file` return a `Result` so we can test that we positively catch the conditions where the copy would refer to the same file (whether identical paths, symlinks or hard links).

components/hab/src/command/studio/enter.rs
```diff
@@ -43,6 +45,22 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
     }
 }
 
+fn cache_ssl_cert_file(cert_file: &str, cert_cache_dir: &Path) {
+    let cert_path = Path::new(&cert_file);
```
baumanj: I've found myself getting confused when reading the logic of this function, and I think a few naming changes may make it easier to follow:
1. `cert_file` → `file_to_cache` or `src_cert`, etc.
1. `cert_cache_dir` → `dest_dir`
1. When converting `cert_file` of type `&str` → `Path`, re-bind the same name rather than introduce `cert_path`* (this is allowed and idiomatic) or make the function take a `Path` as input rather than `&str`.

These are just suggestions; not required changes

\* Since rust has such strong typing, in general I find it useful to put only the purpose of the data in the variable name and leave out words that describe the type.

components/hab/src/command/studio/enter.rs
```diff
@@ -88,6 +110,14 @@ pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
         debug!("Creating ssl_path at: {}", ssl_path.display());
         stdfs::create_dir_all(&ssl_path)?;
     }
+
+    if let Ok(ssl_cert_file) = env::var(SSL_CERT_FILE_ENVVAR) {
+        match cache_ssl_cert_file(&ssl_cert_file, &ssl_path) {
+            Ok(_) => {}
+            Err(err) => warn!("Unable to cache SSL_CERT_FILE: {}", err),
+        }
```
baumanj:
```rust
        if let Err(err) = cache_ssl_cert_file(&ssl_cert_file, &ssl_path) {
            warn!("Unable to cache SSL_CERT_FILE: {}", err);
        }
```
is a tad simpler/more idiomatic; but the match approach works too

components/hab/src/command/studio/enter.rs
```diff
@@ -1,23 +1,27 @@
...
 use crate::{config,
-            error::Result,
+            error::{Error,
+                    Result},
             BLDR_URL_ENVVAR,
             CTL_SECRET_ENVVAR,
             ORIGIN_ENVVAR};
 
 use habitat_core::AUTH_TOKEN_ENVVAR;
+use same_file::is_same_file;
```
baumanj: I'm getting
> error[E0432]: unresolved import `same_file`

when I try this. Is there a change to the cargo.toml that didn't get pushed up, by any chance?

components/hab/src/command/studio/enter.rs
```diff
@@ -43,6 +47,24 @@ fn set_env_var_from_config(env_var: &str, config_val: Option<String>, sensitive:
...
+    let cert_filename = cert_path.file_name().expect("SSL_CERT_FILE is not a file");
+    let cache_file = cert_cache_dir.join(&cert_filename);
+
+    if cache_file.exists() && cert_path.exists() && is_same_file(&cache_file, &cert_path)? {
+        return Err(Error::CacheSslCertError("Source and destination certificate are the same \
+                                             file"
+                                                  .to_string()));
+    }
+
+    debug!("Caching SSL_CERT_FILE {}", cert_file);
```
baumanj: In case of errors, this would probably be more useful at the top of the function and also inclusive of the destination. Something like
```suggestion
    debug!("Caching SSL_CERT_FILE {:?} -> {:?}",
           cert_file, cert_cache_dir);
```

components/hab/src/command/studio/enter.rs
smacfarlane: :man_facepalming: Fixed!

PR #6934, opened 2019-09-05
Fix typo in 'type' parameter for command 'hab origin key export'

### LINKED ISSUE #6933: Incorrect Help Doc for `type` in `hab origin key export`
Pretty self explanatory:
```
$ hab origin key export --help
hab-origin-key-export 0.83.0/20190712232015
Outputs the latest origin key contents to stdout

USAGE:
    hab origin key export [OPTIONS] <ORIGIN> --cache-key-path <CACHE_KEY_PATH>

FLAGS:

OPTIONS:
        --cache-key-path <CACHE_KEY_PATH>    Path to export origin keys from. Default value is hab/cache/keys if root
                                             and .hab/cache/keys under the home directory otherwise. [env:
                                             HAB_CACHE_KEY_PATH=]
    -t, --type <PAIR_TYPE>                   Export either the 'public' or 'private' key

ARGS:
    <ORIGIN>    

$ hab origin key export --type private sirajrauff
error: Invalid value for '--type <PAIR_TYPE>': PAIR_TYPE: private is invalid, must be one of (public, secret)

$ hab origin key export --type secret sirajrauff
SIG-SEC-1
...
```

This boils down to the command's [help doc](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/cli.rs#L326) no longer matching the function's [parameters](https://github.com/habitat-sh/habitat/blob/21e486ae0a9a4f528b27981a55aebb00bc3845ae/components/hab/src/cli.rs#L1292).

sirajrauff:
Resolves #6933

Signed-off-by: Siraj Rauff <sirajrauff@gmail.com>

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -323,7 +323,7 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                     (aliases: &["e", "ex", "exp", "expo", "expor"])
                     (@arg ORIGIN: +required +takes_value {valid_origin})
                     (@arg PAIR_TYPE: -t --type +takes_value {valid_pair_type}
-                        "Export either the 'public' or 'private' key")
+                        "Export either the 'public' or 'secret' key")
                     (arg: arg_cache_key_path("Path to export origin keys from. \
                         Default value is hab/cache/keys if root and .hab/cache/keys under the home \
                         directory otherwise."))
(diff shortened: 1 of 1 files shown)


components/hab/src/cli.rs
```diff
@@ -323,7 +323,7 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                     (aliases: &["e", "ex", "exp", "expo", "expor"])
                     (@arg ORIGIN: +required +takes_value {valid_origin})
                     (@arg PAIR_TYPE: -t --type +takes_value {valid_pair_type}
-                        "Export either the 'public' or 'private' key")
+                        "Export either the 'public' or 'secret' key")
```
kagarmoe: Thanks! I've suggested a change to define 'secret' with the same vocabulary that we're using int the documentation.

```suggestion
                        "Export either the 'public' or 'secret' key. The 'secret' key is the origin private key")
```

sirajrauff: I've added this in now!

sirajrauff:
@christophermaier amended & force-pushed. Should be good to go?

PR #6940, opened 2019-09-09
Audit `habitat_sup::manager::Manager` locking

baumanj:

Looking back at [the commit that introduced the `RwLock`](https://github.com/habitat-sh/habitat/commit/2e4bcd173701957154f785184c8b0e9c70409c38), it looks like [`UserConfigWatcher` already had a `Mutex`](https://github.com/habitat-sh/habitat/blob/e03695b28234c78aa19b302abbbef602c834c504/components/sup/src/manager/user_config_watcher.rs#L88). Do you know why the additional `RwLock` was added, @christophermaier?


baumanj:
Added `T-DO-NOT-MERGE` label since we're currently in a merge freeze

PR #6943, opened 2019-09-10
Upgrade buildkite agent on macOS instead of install

christophermaier:
If it's running in Buildkite, it already has the agent, by
definition. This just ensures we have the latest version of it.

Without this change, builds will fail with an error like this:

```
Error: buildkite-agent 3.13.2 is already installed
To upgrade to 3.14.0, run `brew upgrade buildkite-agent`.
```


christophermaier:
"Fake" release using this code can be found [here](https://buildkite.com/chef/habitat-sh-habitat-master-release/builds/4876) (sorry, broader world; it's a private internal link :frowning: )

PR #6945, opened 2019-09-10
Ma/hab 6878/pkg download

### LINKED ISSUE #6878: Add 'hab pkg download'
## History and context:

This work is to support:

https://app.zenhub.com/workspaces/habitat-iteration-board-5d0267716553d87b0348bb21/issues/habitat-sh/on-prem-builder/166

The purpose of that ticket is to ease the bootstrapping and updating of on-prem builders. Currently users can download all of the core plans in LATEST.tgz, or by hand download what they need, but it's a very clunky process. A user may well want to use both a subset of core packages and non-core origins (particularly for windows users). We want to make it easy to sync an on-prem builder with a specific set of top-level packages that are needed for a customer deployment. To assist with this, we wanted the on-prem-builder script (scripts/on-prem-archive.sh) to be able to take as input a set of top-level packages in a simple text based file, and create a bootstrap bundle with all the required packages and keys with transitive dependencies accounted for.

During the development process, we realized that expanding the packages transitively and downloading them was much easier implemented in Rust, so we could reuse habitat code that already did this, culminating in pull https://github.com/habitat-sh/on-prem-builder/pull/175. On review, we realized that the work not only should reuse habitat code, but really belonged as part of habitat. That would allow us to leverage even more of the habitat code
base, including UX cues like progress bars, and fold some improvements back into the habitat base.

This caused us to revisit https://github.com/habitat-sh/habitat/issues/4616; which proposes a `hab pkg download` command. Our primary purpose is to provide the functionality required for the on-prem builder script, but it still is worthwhile making it a usable command for general usage.

## Proposed Command API:

    Downloads a Habitat package or packages from Builder

    USAGE:

    hab pkg download [FLAGS] [OPTIONS] [<PKG_IDENT> | --file <FILENAME>]...

    FLAGS:

    OPTIONS:
        -u, --url <BLDR_URL>       Specify an alternate Builder endpoint. If not specified, the value will be taken from the
        -c, --channel <CHANNEL>    Install from the specified release channel (default: stable)
        --cache <DIRECTORY>        Directory to use for the cache. /hab/cache is the default
        --file <FILENAME>          Newline separated list of package identifiers or filepaths
        -t, --target <TARGET>      Use this as the target architecture to fetch (default is host architecture <insert actual list of architectures here>).  
    ARGS:
        <PKG_IDENT>...             One or more Habitat package identifiers (ex: acme/redis)
 
    This command downloads package artifacts and their transitive dependencies into the cache, but does not install 
    them. It also downloads the relevant origin public keys. The packages are specified on the command line or in a
    file. This primary use of this is fetch packages to a local cache as part of bootstrapping a local builder instance.

## Design notes:

This command follows parallel construction with `hab pkg install` wherever it makes sense.

AUTH\_TOKEN, BLDR\_URL, and CHANNEL are taken from `hab pkg install`, as both commands have a similar requirement to select and download packages from a builder.

There are a few points of difference, centered around the requirement to download but not install the packages.

The following options have been removed or changed:

1) We only take package identifiers, not artifacts. The install command can be used to install local artifacts, and that isn't needed here.
2) Binlinking is part of package install, and isn't applicable, so the BINLINK, BINLINK\_DIR, FORCE, and IGNORE\_INSTALL\_HOOK options are not provided.
3) OFFLINE\_INSTALL is not applicable; the whole point is to download packages
4) IGNORE\_LOCAL isn't initially provided; however there may be some value in providing that down the road.

There are a few new flags to support the download use case better:

1) We allow the target to be provided. The TARGET option accommodates the possibility that the machine we are downloading on (say a bastion host) may be running a different OS or architecture than the machines we are building packages for. The actual implementation of the --target feature should list available architectures in the command line help. 

2) The cache directory helps us avoid cluttering the local habitat cache with packages that aren't necessary. 

3) The FILENAME option is a convenience feature. Some synchronization use cases (e.g. cloning an existing on-prem builder) could generate a long list of targets. While these could be provided on the command line, this can be burdensome, particularly on Windows, which limits command line lengths. However, this could be worked around in the script if necessary.

One option that was considered and rejected was to add a 'no recursion' flag, limiting the packages fetched to those provided, not the transitive dependencies. However that didn't seem to add much value, and was dropped.

markan:
This closes #6878

This should probably be squashed a bit before merge, but it was split out into a couple of phases.

An initial attempt to follow parallel construction with the install command was attempted, but it didn't actually lead to much reusable code. A second implementation was done which was much more linear and manageable, and maybe those should be squashed together.

I'm still working on how to test this code. I've added end to end tests.

The  builder API was extended to provide full package information. This might also be a separate PR.

- [x] fix bad choice of flags to rustfmt
- [x] lint
- [x] Clean up error handling for missing files, missing idents etc.
- [x] Fix bug where --cache flag not respected
- [ ] unit tests wherever possible

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -1,9 +1,11 @@
 use crate::command::studio;
+
 use clap::{App,
            AppSettings,
            Arg,
            ArgMatches};
-use habitat_common::{cli::{BINLINK_DIR_ENVVAR,
+use habitat_common::{cli::{file_into_idents,
+                           BINLINK_DIR_ENVVAR,
                            DEFAULT_BINLINK_DIR,
                            PACKAGE_TARGET_ENVVAR,
                            RING_ENVVAR,
@@ -435,14 +437,15 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                 (@arg DEST_DIR: -d --dest +takes_value {non_empty} env(BINLINK_DIR_ENVVAR) default_value(DEFAULT_BINLINK_DIR)
                     "Sets the destination directory")
                 (@arg FORCE: -f --force "Overwrite existing binlinks")
-            )
+             )
+            (subcommand: sub_pkg_build())
             (@subcommand config =>
                 (about: "Displays the default configuration options for a service")
                 (aliases: &["conf", "cfg"])
                 (@arg PKG_IDENT: +required +takes_value {valid_ident}
                     "A package identifier (ex: core/redis, core/busybox-static/1.42.2)")
-            )
-            (subcommand: sub_pkg_build())
+             )
+            (subcommand: sub_pkg_download())
             (@subcommand env =>
                 (about: "Prints the runtime environment of a specific installed package")
                 (@arg PKG_IDENT: +required +takes_value {valid_ident}
@@ -921,6 +924,28 @@ fn sub_pkg_build() -> App<'static, 'static> {
                                               .long("docker"));
     }
 
+    sub
+}
+
+fn sub_pkg_download() -> App<'static, 'static> {
+    let sub = clap_app!(@subcommand download =>
+    (about: "Download Habitat artifacts (including dependencies and keys) from Builder")
+    (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
+    (@arg BLDR_URL: --url -u +takes_value {valid_url} default_value(habitat_core::url::DEFAULT_BLDR_URL)
+        "Specify an alternate Builder endpoint. If not specified, the value will \
+         be taken from the HAB_BLDR_URL environment variable if defined.")
+    (@arg CHANNEL: --channel -c +takes_value default_value[stable] env(ChannelIdent::ENVVAR)
+        "Download from the specified release channel")
+    (@arg DOWNLOAD_DIRECTORY: --("download-directory") +takes_value "The path to store downloaded artifacts")
+    (@arg PKG_IDENT_FILE: --file +takes_value +multiple {valid_ident_file}
+        "File with newline separated package identifiers")
+    (@arg PKG_IDENT: +multiple {valid_ident}
+            "One or more Habitat package identifiers (ex: acme/redis)")
+    (@arg PKG_TARGET: --target -t +takes_value {valid_target}
+            "Target architecture to fetch. E.g. x86_64-linux")
+    (@arg VERIFY: --verify
+            "Verify package integrity after download (Warning: this can be slow)")
+    );
     sub
 }
 
@@ -1404,13 +1429,23 @@ fn valid_ident(val: String) -> result::Result<(), String> {
 }
 
 #[allow(clippy::needless_pass_by_value)] // Signature required by CLAP
+fn valid_ident_file(val: String) -> result::Result<(), String> {
+    file_into_idents(&val).map(|_| ())
+                          .map_err(|e| e.to_string())
+}
+
+#[allow(clippy::needless_pass_by_value)] // Signature required by CLAP
 fn valid_target(val: String) -> result::Result<(), String> {
     match PackageTarget::from_str(&val) {
         Ok(_) => Ok(()),
         Err(_) => {
-            Err(format!("'{}' is not valid. A valid target is in the form \
-                         architecture-platform (example: x86_64-linux)",
-                        &val))
+            let targets: Vec<_> = PackageTarget::targets().map(std::convert::AsRef::as_ref)
+                                                          .collect();
+            Err(format!("'{}' is not valid. Valid targets are in the form \
+                         architecture-platform (currently Habitat allows \
+                         the following: {})",
+                        &val,
+                        targets.join(", ")))
         }
     }
 }
--- a/components/hab/src/command/pkg/download.rs
+++ b/components/hab/src/command/pkg/download.rs
@@ -1,0 +1,432 @@
+//! Downloads a Habitat package from a [depot](../depot).
+//!
+//! # Examples
+//!
+//! ```bash
+//! $ hab pkg download core/redis
+//! ```
+//!
+//! Will download `core/redis` package and all its transitive dependencies from a custom depot:
+//!
+//! ```bash
+//! $ hab pkg download -u http://depot.co:9633 -t x86-64_linux --download-directory download core/redis/3.0.1
+//! ```
+//!
+//! This would download the `3.0.1` version of redis for linux and all
+//! of its transitive dependencies, and the accompanying signing keys
+//! to a directory 'download'
+//!
+//! The most common usage will have a file containing newline separated list of package
+//! identifiers.
+//!
+//! # Internals
+//!
+//! * Resolve the list of partial artifact identifiers to fully qualified idents
+//! * Gather the TDEPS of the list (done concurrently with the above step)
+//! * Download the artifact
+//! * Verify it is un-altered
+//! * Fetch the signing keys
+
+use std::{collections::HashSet,
+          fs::DirBuilder,
+          path::{Path,
+                 PathBuf},
+          time::Duration};
+
+use crate::{api_client::{self,
+                         BoxedClient,
+                         Client,
+                         Error::APIError,
+                         Package},
+            common::Error as CommonError,
+            hcore::{crypto::{artifact,
+                             keys::parse_name_with_rev,
+                             SigKeyPair},
+                    fs::cache_root_path,
+                    package::{PackageArchive,
+                              PackageIdent,
+                              PackageTarget},
+                    ChannelIdent,
+                    Error as CoreError}};
+
+use reqwest::StatusCode;
+use retry::{delay,
+            retry};
+
+use crate::error::{Error,
+                   Result};
+
+use habitat_common::ui::{Status,
+                         UIWriter};
+
+pub const RETRIES: usize = 5;
+pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
+
+/// Download a Habitat package.
+///
+/// If an `PackageIdent` is given, we retrieve the package from the specified Builder
+/// `url`. Providing a fully-qualified identifer will result in that exact package being downloaded
+/// (regardless of `channel`). Providing a partially-qualified identifier will result in the
+/// installation of latest appropriate release from the given `channel`.
+///
+/// Any dependencies of will be retrieved from Builder (if they're not already downloaded locally).
+///
+/// At the end of this function, the specified package and all its
+/// dependencies will be downloaded on the system in the
+/// <download_path>/artifacts directory. Any signing keys will also be
+/// downloaded and put in the <download_path/keys> directory.
+
+/// Also, in the future we may want to accept an alternate builder to 'filter' what we pull down by
+/// That would greatly optimize the 'sync' to on prem builder case, as we could point to that
+/// and only fetch what we don't already have.
+#[allow(clippy::too_many_arguments)]
+pub fn start<U>(ui: &mut U,
+                url: &str,
+                channel: &ChannelIdent,
+                product: &str,
+                version: &str,
+                idents: Vec<PackageIdent>,
+                target: PackageTarget,
+                download_path: Option<&PathBuf>,
+                token: Option<&str>,
+                verify: bool)
+                -> Result<()>
+    where U: UIWriter
+{
+    debug!("Starting download with url: {}, channel: {}, product: {}, version: {}, target: {}, \
+            download_path: {:?}, token: {:?}, verify: {}",
+           url, channel, product, version, target, download_path, token, verify);
+
+    let download_path_default = &cache_root_path::<PathBuf>(None); // Satisfy E0716
+    let download_path_expanded = download_path.unwrap_or(download_path_default).as_ref();
+    debug!("Using download_path {:?} expanded to {:?}",
+           download_path, download_path_expanded);
+
+    // We deliberately use None to specify the default path as this is used for cert paths, which
+    // we don't want to override.
+    let api_client = Client::new(url, product, version, None)?;
+    let task = DownloadTask { idents,
+                              target,
+                              url,
+                              api_client,
+                              token,
+                              channel,
+                              download_path: download_path_expanded,
+                              verify };
+
+    let download_count = task.execute(ui)?;
+
+    debug!("Expanded package count: {}", download_count);
+
+    Ok(())
+}
+
+struct DownloadTask<'a> {
+    idents:        Vec<PackageIdent>,
+    target:        PackageTarget,
+    url:           &'a str,
+    api_client:    BoxedClient,
+    token:         Option<&'a str>,
+    channel:       &'a ChannelIdent,
+    download_path: &'a Path,
+    verify:        bool,
+}
+
+impl<'a> DownloadTask<'a> {
+    fn execute<T>(&self, ui: &mut T) -> Result<usize>
+        where T: UIWriter
+    {
+        // This was written intentionally with an eye towards data parallelism
+        // Any or all of these phases should naturally fit a fork-join model
+
+        ui.begin(format!("Preparing to resolve packages and download transitive \
+                          dependendencies for {} package idents",
+                         self.idents.len()))?;
+        ui.begin(format!("Using channel {} from {}", self.channel, self.url))?;
+        ui.begin(format!("Storing in download directory {:?} ", self.download_path))?;
+
+        self.verify_and_prepare_download_directory(ui)?;
+
+        // Phase 1: Expand to fully qualified deps and TDEPS
+        let expanded_idents = self.expand_sources(ui)?;
+
+        // Phase 2: Download artifacts
+        let downloaded_artifacts = self.download_artifacts(ui, &expanded_idents)?;
+
+        Ok(downloaded_artifacts.len())
+    }
+
+    // For each source, use the builder/depot to expand it to a fully qualifed form
+    // The same call gives us the TDEPS, add those as well.
+    fn expand_sources<T>(&self, ui: &mut T) -> Result<HashSet<(PackageIdent, PackageTarget)>>
+        where T: UIWriter
+    {
+        let mut expanded_packages = Vec::<Package>::new();
+        let mut expanded_idents = HashSet::<(PackageIdent, PackageTarget)>::new();
+
+        // This loop should be easy to convert to a parallel map.
+        for ident in &self.idents {
+            let package = self.determine_latest_from_ident(ui, &ident.clone(), self.target)?;
+            expanded_packages.push(package);
+        }
+
+        // Collect all the expanded deps into one structure
+        // Done separately because it's not as easy to parallelize
+        for package in expanded_packages {
+            for ident in package.tdeps {
+                expanded_idents.insert((ident.clone(), self.target));
+            }
+            expanded_idents.insert((package.ident.clone(), self.target));
+        }
+
+        ui.status(Status::Found,
+                  format!("{} artifacts", expanded_idents.len()))?;
+
+        Ok(expanded_idents)
+    }
+
+    fn download_artifacts<T>(&self,
+                             ui: &mut T,
+                             expanded_idents: &HashSet<(PackageIdent, PackageTarget)>)
+                             -> Result<Vec<PackageArchive>>
+        where T: UIWriter
+    {
+        let mut downloaded_artifacts = Vec::<PackageArchive>::new();
+
+        ui.status(Status::Downloading,
+                  format!("Downloading {} artifacts", expanded_idents.len()))?;
+
+        for (ident, target) in expanded_idents {
+            let archive: PackageArchive = match self.get_downloaded_archive(ui, ident, *target) {
+                Ok(v) => v,
+                Err(e) => {
+                    // Is this the right status? Or should this be a debug message?
+                    debug!("Error fetching archive {} for {}: {:?}", ident, *target, e);
+                    ui.status(Status::Missing,
+                              format!("Error fetching archive {} for {}", ident, *target))?;
+                    return Err(e);
+                }
+            };
+
+            downloaded_artifacts.push(archive);
+        }
+
+        Ok(downloaded_artifacts)
+    }
+
+    fn determine_latest_from_ident<T>(&self,
+                                      ui: &mut T,
+                                      ident: &PackageIdent,
+                                      target: PackageTarget)
+                                      -> Result<Package>
+        where T: UIWriter
+    {
+        // Unlike in the install command, we always hit the online
+        // depot; our purpose is to sync with latest, and falling back
+        // to a local package would defeat that. Find the latest
+        // package in the proper channel from Builder API,
+        ui.status(Status::Determining,
+                  format!("latest version of {} for {} in the '{}' channel",
+                          ident, target, self.channel))?;
+        match self.fetch_latest_package_in_channel_for(ident, target, self.channel, self.token) {
+            Ok(latest_package) => {
+                ui.status(Status::Using,
+                          format!("{} as latest matching {} for {}",
+                                  latest_package.ident, ident, target))?;
+                Ok(latest_package)
+            }
+            Err(Error::APIClient(APIError(StatusCode::NOT_FOUND, _))) => {
+                // In install we attempt to recommend a channel to look in. That's a bit of a
+                // heavyweight process, and probably a bad idea in the context of
+                // what's a normally a batch process. It might be OK to fall back to
+                // the stable channel, but for now, error.
+                ui.warn(format!("No packages matching ident {} for {} exist in the '{}' \
+                                 channel. Check the package ident, target, channel and Builder \
+                                 url ({}) for correctness",
+                                ident, target, self.channel, self.url))?;
+                Err(CommonError::PackageNotFound(format!("{} for {} in channel {}",
+                                                         ident, target, self.channel)).into())
+            }
+            Err(e) => {
+                debug!("Error fetching ident {} for target {}: {:?}",
+                       ident, target, e);
+                ui.warn(format!("Error fetching ident {} for target {}", ident, target))?;
+                Err(e)
+            }
+        }
+    }
+
+    // This function and its sibling get_cached_artifact in
+    // install.rs deserve to be refactored to eke out commonality.
+    /// This ensures the identified package is in the local download directory,
+    /// verifies it, and returns a handle to the package's metadata.
+    fn get_downloaded_archive<T>(&self,
+                                 ui: &mut T,
+                                 ident: &PackageIdent,
+                                 target: PackageTarget)
+                                 -> Result<PackageArchive>
+        where T: UIWriter
+    {
+        let fetch_artifact = || self.fetch_artifact(ui, ident, target);
+        if self.downloaded_artifact_path(ident, target).is_file() {
+            debug!("Found {} in download directory, skipping remote download",
+                   ident);
+            ui.status(Status::Skipping,
+                      format!("because {} was found in the download directory", ident))?;
+        } else if let Err(err) = retry(delay::Fixed::from(RETRY_WAIT).take(RETRIES), fetch_artifact)
+        {
+            return Err(CommonError::DownloadFailed(format!("We tried {} times but could not \
+                                                            download {} for {}. Last error \
+                                                            was: {}",
+                                                           RETRIES, ident, target, err)).into());
+        }
+
+        // At this point the artifact is in the download directory...
+        let mut artifact = PackageArchive::new(self.downloaded_artifact_path(ident, target));
+        self.fetch_keys_and_verify_artifact(ui, ident, target, &mut artifact)?;
+        Ok(artifact)
+    }
+
+    // This function and its sibling in install.rs deserve to be refactored to eke out commonality.
+    /// Retrieve the identified package from the depot, ensuring that
+    /// the artifact is downloaded.
+    fn fetch_artifact<T>(&self,
+                         ui: &mut T,
+                         ident: &PackageIdent,
+                         target: PackageTarget)
+                         -> Result<()>
+        where T: UIWriter
+    {
+        ui.status(Status::Downloading, format!("{} for {}", ident, target))?;
+        match self.api_client.fetch_package((ident, target),
+                                            self.token,
+                                            &self.path_for_artifact(),
+                                            ui.progress())
+        {
+            Ok(_) => Ok(()),
+            Err(api_client::Error::APIError(StatusCode::NOT_IMPLEMENTED, _)) => {
+                println!("Host platform or architecture not supported by the targeted depot; \
+                          skipping.");
+                Ok(())
+            }
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    fn fetch_origin_key<T>(&self,
+                           ui: &mut T,
+                           name_with_rev: &str,
+                           token: Option<&str>)
+                           -> Result<()>
+        where T: UIWriter
+    {
+        ui.status(Status::Downloading,
+                  format!("{} public origin key", &name_with_rev))?;
+        let (name, rev) = parse_name_with_rev(&name_with_rev)?;
+        self.api_client.fetch_origin_key(&name,
+                                          &rev,
+                                          token,
+                                          &self.path_for_keys(),
+                                          ui.progress())?;
+        ui.status(Status::Cached,
+                  format!("{} public origin key", &name_with_rev))?;
+        Ok(())
+    }
+
+    fn fetch_keys_and_verify_artifact<T>(&self,
+                                         ui: &mut T,
+                                         ident: &PackageIdent,
+                                         target: PackageTarget,
+                                         artifact: &mut PackageArchive)
+                                         -> Result<()>
+        where T: UIWriter
+    {
+        // We need to look at the artifact to know the signing keys to fetch
+        // Once we have them, it's the natural time to verify.
+        // Otherwise, it might make sense to take this fetch out of the verification code.
+        let signer = artifact::artifact_signer(&artifact.path)?;
+        if SigKeyPair::get_public_key_path(&signer, &self.path_for_keys()).is_err() {
+            ui.status(Status::Downloading,
+                      format!("Public key for signer {:?}", signer))?;
+            self.fetch_origin_key(ui, &signer, self.token)?;
+        }
+
+        if self.verify {
+            ui.status(Status::Verifying, artifact.ident()?)?;
+            artifact.verify(&self.path_for_keys())?;
+            debug!("Verified {} for {} signed by {}", ident, target, &signer);
+        }
+        Ok(())
+    }
+
+    // This function and its sibling in install.rs deserve to be refactored to eke out commonality.
+    /// Returns the path to the location this package would exist at in
+    /// the local package cache. It does not mean that the package is
+    /// actually *in* the package download directory, though.
+    fn downloaded_artifact_path(&self, ident: &PackageIdent, target: PackageTarget) -> PathBuf {
+        self.path_for_artifact()
+            .join(ident.archive_name_with_target(target).unwrap())
+    }
+
+    fn fetch_latest_package_in_channel_for(&self,
+                                           ident: &PackageIdent,
+                                           target: PackageTarget,
+                                           channel: &ChannelIdent,
+                                           token: Option<&str>)
+                                           -> Result<Package> {
+        self.api_client
+            .show_package_metadata((&ident, target), channel, token)
+            .map_err(Error::from)
+    }
+
+    /// The cache_*_path functions in fs don't let you override a path base with Some(base)
+    /// So we have to build our own paths.
+    fn path_for_keys(&self) -> PathBuf { self.download_path.join("keys") }
+
+    fn path_for_artifact(&self) -> PathBuf { self.download_path.join("artifacts") }
+
+    /// Sanity check the download directory tree. The errors from the api around permissions are
+    /// opaque; this validates the directory in advance to help provide useful feedback.
+    fn verify_and_prepare_download_directory<T>(&self, ui: &mut T) -> Result<()>
+        where T: UIWriter
+    {
+        let system_paths = [self.download_path,
+                            &self.path_for_keys(),
+                            &self.path_for_artifact()];
+
+        ui.status(Status::Verifying,
+                  format!("Checking the download directory ({}) permissions and creating it \
+                           (if needed)",
+                          self.download_path.display()))?;
+
+        let mut builder = DirBuilder::new();
+        builder.recursive(true);
+
+        // Create directories if they don't exist
+        for dir in &system_paths {
+            builder.create(dir).map_err(|_| {
+                                    mk_perm_error(format!("Can't create directory {:?} needed \
+                                                           for download",
+                                                          dir))
+                                })?
+        }
+
+        // Check permissions of directories:
+        for dir in &system_paths {
+            let metadata = std::fs::metadata(dir)?;
+            if !metadata.is_dir() {
+                return Err(mk_perm_error(format!("{} isn't a directory, needed for \
+                                                  download",
+                                                 dir.display())));
+            }
+            if metadata.permissions().readonly() {
+                return Err(mk_perm_error(format!("{} isn't writeable, needed for \
+                                                  download",
+                                                 dir.display())));
+            }
+        }
+        Ok(())
+    }
+}
+
+fn mk_perm_error(msg: String) -> Error { CoreError::PermissionFailed(msg).into() }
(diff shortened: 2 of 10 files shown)


components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+use crate::{error::{Error,
+                    Result},
+            ui::{Status,
+                 UIWriter}};
+
+pub const RETRIES: usize = 5;
+pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum InstallMode {
```
chefsalim: Looks like InstallMode isn't used and can be removed

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+    Online,
+    Offline,
+}
+
+impl Default for InstallMode {
+    fn default() -> Self { InstallMode::Online }
+}
+
+/// Represents a fully-qualified Package Identifier, meaning that the normally optional version and
```
chefsalim: This comment seems out of place

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+//! $ hab pkg install core/redis/3.0.1 redis -u http://depot.co:9633
+//! ```
+//!
+//! This would install the `3.0.1` version of redis.
+//!
+//! # Internals
+//!
+//! * Download the artifact
+//! * Verify it is un-altered
+//! * Unpack it
```
chefsalim: We should not need or want to unpack the downloaded artifact

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+//! ```bash
+//! $ hab pkg install core/redis/3.0.1 redis -u http://depot.co:9633
+//! ```
+//!
+//! This would install the `3.0.1` version of redis.
+//!
+//! # Internals
+//!
+//! * Download the artifact
+//! * Verify it is un-altered
```
chefsalim: Verification is done on install, we should not need to do it on download (it is an expensive operation)

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+
+/// Represents a fully-qualified Package Identifier, meaning that the normally optional version and
+/// release package coordinates are guaranteed to be set. This fully-qualified-ness is checked on
+/// construction and as the underlying representation is immutable, this state does not change.
+
+/// Download a Habitat package.
+///
+/// If an `PackageIdentTarget` is given, we retrieve the package
+/// from the specified Builder `url`. Providing a fully-qualified
+/// identifer will result in that exact package being installed
```
chefsalim: Comments need to be updated to talk about download not install...

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+#[allow(clippy::too_many_arguments)]
+pub fn start<U>(ui: &mut U,
+                url: &str,
+                channel: &ChannelIdent,
+                product: &str,
+                version: &str,
+                idents: Vec<PackageIdent>,
+                target: PackageTarget,
+                fs_root_path: Option<&PathBuf>,
+                token: Option<&str>)
```
chefsalim: I believe start should take a param to the override cache path

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+                -> Result<()>
+    where U: UIWriter
+{
+    debug!("Starting download with url: {}, channel: {}, product: {}, version: {}, target: {}, \
+            fs_root_path: {:?}, token: {:?}",
+           url, channel, product, version, target, fs_root_path, token);
+
+    let key_cache_path = &cache_key_path(fs_root_path);
```
chefsalim: This is probably the bug we were discussing... if the caller specifies a path it should get passed in and used instead

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+    debug!("install key_cache_path: {}", key_cache_path.display());
+
+    let artifact_cache_path = &cache_artifact_path(fs_root_path);
+    debug!("install artifact_cache_path: {}",
+           artifact_cache_path.display());
+
+    // TODO we use the same root path for ssl certs as we do for the rest of the root path,
+    // We shouldn't probably override it here, as this appears to be largely for cert paths
```
chefsalim: Yes, this is specified only for the certs, so it should be passed in as is

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+struct DownloadTask<'a> {
+    idents: Vec<PackageIdent>,
+    target: PackageTarget,
+    url: &'a str,
+    api_client: BoxedClient,
+    token: Option<&'a str>,
+    channel: &'a ChannelIdent,
+    /// The path to the local artifact cache (e.g., /hab/cache/artifacts)
```
chefsalim: Instead of calling these "cache" paths, maybe just call them "download paths"?  Also do we need to pass both artifact and key cache separately?

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+        // depot; our purpose is to sync with latest, and falling back
+        // to a local package would defeat that. Find the latest
+        // package in the proper channel from Builder API,
+        ui.status(Status::Determining,
+                  format!("latest version of {} in the '{}' channel",
+                          &ident, self.channel))?;
+        match self.fetch_latest_package_in_channel_for(ident, self.channel, self.token) {
+            Ok(latest_package) => {
+                ui.status(Status::Using,
+                          format!("{} as lastest matching {}", latest_package.ident, ident))?;
```
chefsalim: Typo: "lastest"

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+                Ok(latest_package)
+            }
+            Err(Error::APIClient(APIError(StatusCode::NOT_FOUND, _))) => {
+                // In install we attempt to recommend a channel to look in. That's a bit of a
+                // heavyweight process, and probably a bad idea in the context of
+                // what's a normally a batch process. It might be ok to fall back to
+                // the stable channel, but for now, error.
+                ui.warn(format!("No releases of {} for exist in the '{}' channel",
+                                ident, self.channel))?;
+                Err(Error::PackageNotFound("".to_string()))
```
chefsalim: Wrap ident instead of "" ?

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+    // install.rs deserve to be refactored to eke out commonality.
+    /// This ensures the identified package is in the local cache,
+    /// verifies it, and returns a handle to the package's metadata.
+    fn get_cached_archive<T>(&self,
+                             ui: &mut T,
+                             package: &PackageIdentTarget)
+                             -> Result<PackageArchive>
+        where T: UIWriter
+    {
+        let fetch_artifact = || self.fetch_artifact(ui, package);
```
chefsalim: This is always downloading the package first which seems incorrect.  I think we want to check to see if it is in the cache first.  If it is, we copy it over to the download path (or, do nothing if download path == cache path).

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+    let task = DownloadTask { idents,
+                              target,
+                              url,
+                              api_client,
+                              token,
+                              channel,
+                              artifact_cache_path,
+                              key_cache_path };
+
+    let downloaded_artifacts: Vec<PackageArchive> = task.execute(ui).unwrap();
```
chefsalim: We don't really need PackageArchives being returned... task.execute() could just return a count - I'm not sure we even need that.

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+        // This was written intentionally with an eye towards data parallelism
+        // Any or all of these phases should naturally fit a fork-join model
+
+        ui.begin(format!("Preparing to download necessary packages for {} idents",
+                         self.idents.len()))?;
+        ui.begin(format!("Using channel {} from {}", self.channel, self.url))?;
+        ui.begin(format!("Storing in cache at {:?} ", self.artifact_cache_path))?;
+
+        // Phase 1: Expand to fully qualified deps and TDEPS
+        let expanded_idents = self.expand_sources(ui)?;
```
chefsalim: I think the code will be simpler if we have the DownloadTask just do a single ident download.  Phase 1 would then get pulled outside of DownloadTask.  DownloadTask also doesn't need to return a PackageArchive (it can just return () - we only care about success/failure(err))

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+        if package.target != artifact_target {
+            return Err(Error::HabitatCore(hcore::Error::WrongActivePackageTarget(
+                package.target,
+                artifact_target,
+            )));
+        }
+
+        let nwr = artifact::artifact_signer(&artifact.path)?;
+        if SigKeyPair::get_public_key_path(&nwr, self.key_cache_path).is_err() {
+            self.fetch_origin_key(ui, &nwr, self.token)?;
```
chefsalim: We don't want to be pulling keys by cracking open each archive - that is prohibitively expensive.
Instead, we should just download all the public keys for the origin(s) into the cache (check and download if they are not already in the cache).

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    fn fetch_origin_key<T>(&self,
+                           ui: &mut T,
+                           name_with_rev: &str,
+                           token: Option<&str>)
+                           -> Result<()>
+        where T: UIWriter
+    {
+        ui.status(Status::Downloading,
```
chefsalim: It would be nice to also have a get_cached_origin_key function (analogous to the get_cached_artifact fn)

components/common/src/command/package/download.rs
chefsalim: This is the only part we need - we should call get_cached_origin_key for each signing key to make sure it is downloaded/populated

components/core/src/package/ident_target.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> { Some(self.cmp(other)) }
+}
+
+impl fmt::Display for PackageIdentTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}/{}", self.ident, self.target)
+    }
+}
+
+// PackageIdentTargets are ORIGIN/NAME/VERSION/DATE/TARGET
```
chefsalim: We should get a design doc and input from the community as well as there are significant decisions.  Might be better to take this out of this PR and do something separately.

components/hab/src/cli.rs
```diff
@@ -905,6 +906,36 @@ fn sub_pkg_build() -> App<'static, 'static> {
...
 
+fn sub_pkg_download(feature_flags: FeatureFlag) -> App<'static, 'static> {
+    let mut sub = clap_app!(@subcommand download =>
+    (about: "Download Habitat artifacts (including dependencies and keys) from Builder")
+    (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
+    (@arg BLDR_URL: --url -u +takes_value {valid_url} default_value(habitat_core::url::DEFAULT_BLDR_URL)
+        "Specify an alternate Builder endpoint. If not specified, the value will \
+         be taken from the HAB_BLDR_URL environment variable if defined. (default: \
+         https://bldr.habitat.sh)")
+    (@arg CACHE_DIRECTORY: --cache +takes_value "The path to store downloaded artifacts")
```
chefsalim: Maybe call this DOWNLOAD_DIRECTORY and default to the user's cache directory if it is not specified.

components/hab/src/cli.rs
```diff
@@ -905,6 +906,36 @@ fn sub_pkg_build() -> App<'static, 'static> {
...
+    (@arg PKG_IDENT_FILE: --file +takes_value +multiple
+        "File with newline separated package identifiers")
+    (@arg PKG_IDENT: +multiple
+            "One or more Habitat package identifiers (ex: acme/redis)")
+    (@arg PKG_TARGET: --target -t +takes_value {valid_target}
+            "Target architecture to fetch. Allowable options are one of (TODO ADD THIS)")
+    );
+    // TODO: Discover the background for this; while it seems reasonable to have, I'd really like to
+    // not cargo cult it in.
+    if feature_flags.contains(FeatureFlag::IGNORE_LOCAL) {
```
chefsalim: Don't think we need this..

components/common/src/command/package/download.rs
markan: That is the intent of the fs_root_path variable. Unfortunately I hit two issues; first the problem described in #6976 where the Some parameter isn't respected. Second is the likely need to at least check the default hab path for ssl certs, or provide a mechanism to copy them. At minimum I hit some issues when overriding the path to Client::new to point to another directory. 
I have a separate branch with some changes to fs.rs to attempt to mitigate some of these issues, but it's a bit bigger (and harder to test) than I think is appropriate to include here.

components/hab/src/cli.rs
markan: I like that. The name seems clearer to me.

components/hab/src/cli.rs
markan: For sure; in no way do we want to fall back to trying to resolve idents from the cache. In other words, IGNORE_LOCAL=true is the only correct semantic I thin.

markan:
Thanks for all the review comments @chefsalim 

There are a few broad issues worth discussing
1) Do we verify the packages after download? We currently get the signing keys as a side effect of this process, but we could split that out.  I do wonder if we at least want to offer the option to verify; I'm thinking about cases where there are expensive steps between download and upload, such as airgaps where no-one wants to discover the stuff they just got through security is corrupted. 
2) @smacfarlane pointed out that most of the other commands live in hab/src/command/ while this command and `pkg install` lives in common/src/command/package. I'm thinking it probably makes sense to move the download command to be with the rest of the commands

components/common/src/command/package/download.rs
chefsalim: Yes, the fs_root_path should be for the certs only, for the Client.  We need a different path for the download (which could either be user specified, or default to the "cache" path)

chefsalim:
@markan on your points: 1) a verify option might be a reasonable middle ground; 2) yes we should move the files to hab/src/command - the install is only in common because the supervisor also uses it.

components/core/src/package/ident_target.rs
markan: Removed PackageIdentTarget from PR

jeremymv2:
Hi @markan! Very nice work - this command will come in handy!
Taking the code for a test drive, I see this UX behavior when no Auth Token is set:
```
/home/ubuntu/habitat (master *%=)$ ./target/debug/hab pkg download --file foo --channel unstable --download_directory /tmp
» Preparing to download necessary packages for 1 idents
» Using channel unstable from https://bldr.habitat.sh
» Storing in cache at "/tmp/artifacts"
☁ Determining latest version of chef/automate-backend-postgresql/0.1.185/20190917192244 for x86_64-linux in the 'unstable' channel
→ Found 0 artifacts
↓ Downloading Downloading 0 artifacts
/home/ubuntu/habitat (master *%=)$
```

Yet the package does indeed exist.

Would it be more helpful to the user to warn or stop if no value for `AUTH_TOKEN` exists? Otherwise the 0 matching results is slightly misleading/confusing.

components/hab/src/cli.rs
```diff
@@ -905,6 +906,29 @@ fn sub_pkg_build() -> App<'static, 'static> {
...
+    let sub = clap_app!(@subcommand download =>
+    (about: "Download Habitat artifacts (including dependencies and keys) from Builder")
+    (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
+    (@arg BLDR_URL: --url -u +takes_value {valid_url} default_value(habitat_core::url::DEFAULT_BLDR_URL)
+        "Specify an alternate Builder endpoint. If not specified, the value will \
+         be taken from the HAB_BLDR_URL environment variable if defined. (default: \
+         https://bldr.habitat.sh)")
+    (@arg CHANNEL: --channel -c +takes_value default_value[stable] env(ChannelIdent::ENVVAR)
+        "Install from the specified release channel (default: stable)")
+    (@arg DOWNLOAD_DIRECTORY: --download_directory +takes_value "The path to store downloaded artifacts")
```
jeremymv2: The underscore here is unlike any other Hab command where dash/hyphen is the norm.  Verifying in the UX Design spec, it should indeed be a hyphen 😃 

It can be remedied easily enough with:

`--("download-directory")`

jeremymv2:
@markan in looking at how `--download-directory` marries up with a potentially new bulk upload from directory argument: `--upload-directory`, I'm thinking we can just slightly tweak the wording of the following message:

`» Storing in cache at "/home/ubuntu/.hab/cache/artifacts"` 
(^^ when `--download_directory` was not specified)

Similarly, 
`» Storing in cache at "/tmp/artifacts"`
(^^ when `--download_directory /tmp` was specified)

I think the `Storing in cache a <DOWNLOAD_DIRECTORY>/artifacts` message might better reflect where items are being stored by indicating `<DIR>` as the *same* value that was specified on the command line or the default. 

This is because `hab pkg download` is _also_ storing other items like the `./keys` directory under the `DOWNLOAD_DIRECTORY`, not just `./artifacts`.

If the message was tweaked to match `DOWNLOAD_DIRECTORY` then on the bulk upload side, the user will mostly likely not be confused and intuitively point `--upload-directory` to the same directory they specified in `--download-directory` and NOT `DOWNLOAD_DIRECTORY/artifacts`.

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+
+        // Phase 2: Download artifacts
+        let downloaded_artifacts = self.download_artifacts(ui, &expanded_idents)?;
+
+        Ok(downloaded_artifacts)
+    }
+
+    // For each source, use the builder/depot to expand it to a fully qualifed form
+    // The same call gives us the TDEPS, add those as
+    fn expand_sources<T>(&self, ui: &mut T) -> Result<HashSet<(Box<PackageIdent>, PackageTarget)>>
```
chefsalim: I think we should be able to eliminate the Box usage...

.expeditor/scripts/verify/run_clippy.sh
```diff
@@ -5,11 +5,11 @@ set -euo pipefail
 source .expeditor/scripts/verify/shared.sh
 
 toolchain="${1:-"$(get_toolchain)"}"
-install_rustup
-install_rust_toolchain "$toolchain"
+#install_rustup
```
chefsalim: I assume we will get these uncommented...

.expeditor/scripts/verify/shared.sh
```diff
@@ -7,7 +7,7 @@ source .expeditor/scripts/shared.sh
 # This script should contain all shared functions for the verify pipeline
 
 # Always accept habitat license
-sudo hab license accept
+#sudo hab license accept
```
chefsalim: Same here on uncommenting

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+//! # Examples
+//!
+//! ```bash
+//! $ hab pkg download core/redis
+//! ```
+//!
+//! Will download `core/redis` package from a custom depot:
+//!
+//! ```bash
+//! $ hab pkg download core/redis/3.0.1 redis -u http://depot.co:9633
```
chefsalim: Is the example command still accurate?

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+//! ```
+//!
+//! This would download the `3.0.1` version of redis.
```
chefsalim: Downloads the latest release for the 3.0.1 version..
Along with tdeps and public keys...
(for the specified target, into the specified folder in a specific format..)

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+use reqwest::StatusCode;
+use retry::{delay,
+            retry};
+
+use crate::{error::{Error,
+                    Result},
+            ui::{Status,
+                 UIWriter}};
+
+pub const RETRIES: usize = 5;
```
chefsalim: Re-use the consts from install.rs
```
                     command::package::install::{RETRIES,
                                                 RETRY_WAIT},
```

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
+
+/// Download a Habitat package.
+///
+/// If an `PackageIdentTarget` is given, we retrieve the package from the specified Builder
```
chefsalim: Update to remove PackageIdentTarget mention

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+/// `url`. Providing a fully-qualified identifer will result in that exact package being installed
+/// (regardless of `channel`). Providing a partially-qualified identifier will result in the
+/// installation of latest appropriate release from the given `channel`.
+///
+/// Any dependencies of will be retrieved from Builder (if they're not already cached locally).
+///
+/// At the end of this function, the specified package and all its dependencies will be downloaded
+/// on the system.
+
+/// Note: it's worth investigating whether
```
chefsalim: Nitpicks... fix the comment line lengths to be somewhat similar across the board

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+    let key_download_path = &path_helper(download_path, "keys", &cache_key_path::<PathBuf>(None));
+    debug!("install key_download_path: {:?}", key_download_path);
+
+    let artifact_download_path = &path_helper(download_path,
+                                              "artifacts",
+                                              &cache_artifact_path::<PathBuf>(None));
+    debug!("install artifact_download_path: {:?}",
+           artifact_download_path);
+
+    // We deliberately use None to specifiy the default path as this is used for cert paths, which
```
chefsalim: Typo - "specify"

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+}
+
+struct DownloadTask<'a> {
+    idents: Vec<PackageIdent>,
+    target: PackageTarget,
+    url: &'a str,
+    api_client: BoxedClient,
+    token: Option<&'a str>,
+    channel: &'a ChannelIdent,
+    /// The path to the local artifact cache (e.g., /hab/cache/artifacts)
```
chefsalim: Update comment... path to the artifact download location 
Alternatively remove comment as the name is self explanatory

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+    fn execute<T>(&self, ui: &mut T) -> Result<Vec<PackageArchive>>
+        where T: UIWriter
+    {
+        // This was written intentionally with an eye towards data parallelism
+        // Any or all of these phases should naturally fit a fork-join model
+
+        ui.begin(format!("Preparing to download necessary packages for {} idents",
+                         self.idents.len()))?;
+        ui.begin(format!("Using channel {} from {}", self.channel, self.url))?;
+        ui.begin(format!("Storing in cache at {:?} ", self.artifact_download_path))?;
```
chefsalim: Maybe omit "cache", or replace with "download path"?

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+    fn get_cached_archive<T>(&self,
+                             ui: &mut T,
+                             ident: &PackageIdent,
+                             target: PackageTarget)
+                             -> Result<PackageArchive>
+        where T: UIWriter
+    {
+        let fetch_artifact = || self.fetch_artifact(ui, ident, target);
+        if self.is_artifact_cached(ident, target) {
+            debug!("Found {} in artifact cache, skipping remote download",
```
chefsalim: Change from "artifact cache" to "download directory"?

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+    {
+        let mut downloaded_artifacts = Vec::<PackageArchive>::new();
+
+        ui.status(Status::Downloading,
+                  format!("Downloading {} artifacts", expanded_idents.len()))?;
+
+        for (ident, target) in expanded_idents {
+            // TODO think through error handling here; failure to fetch, etc
+            // Probably worth keeping statistics
+            let archive: PackageArchive = self.get_cached_archive(ui, ident, *target)?;
```
chefsalim: Maybe debug log the error?

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    fn verify_artifact<T>(&self,
+                          ui: &mut T,
+                          ident: &PackageIdent,
+                          target: PackageTarget,
+                          artifact: &mut PackageArchive)
+                          -> Result<()>
+        where T: UIWriter
+    {
+        let artifact_ident = artifact.ident()?;
```
chefsalim: We should not use any archive functions like ident() or target() that require opening and random accessing the package stream - it is very expensive

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+        if ident != &artifact_ident {
+            return Err(Error::ArtifactIdentMismatch((artifact.file_name(),
+                                                     artifact_ident.to_string(),
+                                                     ident.to_string())));
+        }
+
+        // Is this even possible? We specify the target in fetch_package above, so we should never
+        // be given a
+        let artifact_target = artifact.target()?;
```
chefsalim: Yeah remove this check - see comment above

components/common/src/command/package/download.rs
```diff
@@ -0,0 +1,400 @@
...
+    // This function and it's sibling in install.rs deserve to be refactored to eke out commonality.
+    fn is_artifact_cached(&self, ident: &PackageIdent, target: PackageTarget) -> bool {
+        self.cached_artifact_path(ident, target).is_file()
+    }
+
+    // This function and it's sibling in install.rs deserve to be refactored to eke out commonality.
+    /// Returns the path to the location this package would exist at in
+    /// the local package cache. It does not mean that the package is
+    /// actually *in* the package cache, though.
+    fn cached_artifact_path(&self, ident: &PackageIdent, target: PackageTarget) -> PathBuf {
```
chefsalim: Maybe change 'cached_artifact_path' to 'download_artifact_path' ?

components/core/src/error.rs
```diff
@@ -89,6 +89,8 @@ pub enum Error {
     InvalidBinding(String),
     /// Occurs when a package identifier string cannot be successfully parsed.
     InvalidPackageIdent(String),
+    /// Occurs when a package identifier target string cannot be successfully parsed.
+    InvalidPackageIdentTarget(String),
```
chefsalim: We should be able to remove this error..

components/core/src/package.rs
```diff
@@ -10,6 +11,7 @@ pub use self::{archive::{FromArchive,
                          PackageArchive},
                ident::{Identifiable,
                        PackageIdent},
+               ident_target::PackageIdentTarget,
```
chefsalim: Remove?

components/hab/src/cli.rs
```diff
@@ -905,6 +906,29 @@ fn sub_pkg_build() -> App<'static, 'static> {
...
+    (@arg PKG_IDENT_FILE: --file +takes_value +multiple
+        "File with newline separated package identifiers")
+    (@arg PKG_IDENT: +multiple
+            "One or more Habitat package identifiers (ex: acme/redis)")
+    (@arg PKG_TARGET: --target -t +takes_value {valid_target}
+            "Target architecture to fetch. Allowable options are one of (TODO ADD THIS)")
```
chefsalim: Fix TODO

.expeditor/scripts/verify/run_clippy.sh
markan: Yep, that should have never been committed

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,388 @@
...
+/// If an `PackageIdent` is given, we retrieve the package from the specified Builder
+/// `url`. Providing a fully-qualified identifer will result in that exact package being installed
+/// (regardless of `channel`). Providing a partially-qualified identifier will result in the
+/// installation of latest appropriate release from the given `channel`.
+///
+/// Any dependencies of will be retrieved from Builder (if they're not already downloaded locally).
+///
+/// At the end of this function, the specified package and all its
+/// dependencies will be downloaded on the system in the
+/// <download_path>/archive directory. Any signing keys will also be
```
jeremymv2: nit: `<download_path>/archive` should be `<download_path>/artifacts`

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+use habitat_common::ui::{Status,
+                         UIWriter};
+
+pub const RETRIES: usize = 5;
+pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
+
+/// Download a Habitat package.
+///
+/// If an `PackageIdent` is given, we retrieve the package from the specified Builder
+/// `url`. Providing a fully-qualified identifer will result in that exact package being installed
```
chefsalim: The text regarding "install" should be updated to reflect "download"...

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+
+        // Phase 2: Download artifacts
+        let downloaded_artifacts = self.download_artifacts(ui, &expanded_idents)?;
+
+        Ok(downloaded_artifacts.len())
+    }
+
+    // For each source, use the builder/depot to expand it to a fully qualifed form
+    // The same call gives us the TDEPS, add those as
+    fn expand_sources<T>(&self, ui: &mut T) -> Result<HashSet<(Box<PackageIdent>, PackageTarget)>>
```
chefsalim: I'm not sure why Box is needed here... should be able to eliminate... at least worth a TODO...

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+    {
+        let mut expanded_packages = Vec::<Package>::new();
+        let mut expanded_idents = HashSet::<(Box<PackageIdent>, PackageTarget)>::new();
+
+        // This loop should be easy to convert to a parallel map
+        for ident in &self.idents {
+            let latest = self.determine_latest_from_ident(ui, &ident.clone(), self.target);
+            if let Ok(package) = latest {
+                expanded_packages.push(package);
+            }
```
chefsalim: It might be good to log something here if there is an error.. ?

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+
+        ui.status(Status::Downloading,
+                  format!("Downloading {} artifacts", expanded_idents.len()))?;
+
+        for (ident, target) in expanded_idents {
+            let archive: PackageArchive = match self.get_downloaded_archive(ui, ident, *target) {
+                Ok(v) => v,
+                Err(e) => {
+                    // Is this the right status? Or should this be a debug message?
+                    ui.status(Status::Missing,
```
chefsalim: I don't think we need this status - we can just exit function on error.
ie - 
```
let archive: PackageArchive = self.get_downloaded_archive(ui, ident, *target)?;
```

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+                                  latest_package.ident, ident, target))?;
+                Ok(latest_package)
+            }
+            Err(Error::APIClient(APIError(StatusCode::NOT_FOUND, _))) => {
+                // In install we attempt to recommend a channel to look in. That's a bit of a
+                // heavyweight process, and probably a bad idea in the context of
+                // what's a normally a batch process. It might be OK to fall back to
+                // the stable channel, but for now, error.
+                ui.warn(format!("No packages matching ident {} for {} exist in the '{}' \
+                                 channel. Check the package ident, target, channel and builder \
```
chefsalim: Minor - capitalize "Builder"

components/hab/src/command/pkg/download.rs
chefsalim: Nvm - looks like we do need to print out a message if the download fails
Might not want to debug format "{:?}" the error though

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+            .join(ident.archive_name_with_target(target).unwrap())
+    }
+
+    fn fetch_latest_package_in_channel_for(&self,
+                                           ident: &PackageIdent,
+                                           target: PackageTarget,
+                                           channel: &ChannelIdent,
+                                           token: Option<&str>)
+                                           -> Result<Package> {
+        let origin_package = self.api_client
```
chefsalim: Minor - you can eliminate the let and Ok()... just return show_package_metadata without the "?"

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+    fn verify_and_prepare_download_directory<T>(&self, ui: &mut T) -> Result<()>
+        where T: UIWriter
+    {
+        let system_paths = [self.download_path,
+                            &self.path_for_keys(),
+                            &self.path_for_artifact()];
+
+        ui.status(Status::Verifying,
+                  format!("Checking permissions and creating as needed the download directory \
+                           {:?}",
```
chefsalim: Instead of using a debug formatter {:?}, we should be able to use a normal format {} with self.download_path.display()... same feedback for the other output below

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+                             -> Result<Vec<PackageArchive>>
+        where T: UIWriter
+    {
+        let mut downloaded_artifacts = Vec::<PackageArchive>::new();
+
+        ui.status(Status::Downloading,
+                  format!("Downloading {} artifacts", expanded_idents.len()))?;
+
+        for (ident, target) in expanded_idents {
+            let archive: PackageArchive = match self.get_downloaded_archive(ui, ident, *target) {
```
baumanj: OPTIONAL: Seems like we could simplify this a bit and make it more robust for future changes:
```rust
            match self.get_downloaded_archive(ui, ident, *target) {
                Ok(archive) => downloaded_artifacts.push(archive),
                Err(e) => {
                    // …
                    return Err(e);
                }
            }
```

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+                Ok(v) => v,
+                Err(e) => {
+                    // Is this the right status? Or should this be a debug message?
+                    ui.status(Status::Missing,
+                              format!("Error fetching archive {} for {}: {:?}",
```
baumanj: Using the `Display` trait `{}` rather than `Debug` `{:?}`, should give more user-friendly error messages
```suggestion
                              format!("Error fetching archive {} for {}: {}",
```

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+                Ok(latest_package)
+            }
+            Err(Error::APIClient(APIError(StatusCode::NOT_FOUND, _))) => {
+                // In install we attempt to recommend a channel to look in. That's a bit of a
+                // heavyweight process, and probably a bad idea in the context of
+                // what's a normally a batch process. It might be OK to fall back to
+                // the stable channel, but for now, error.
+                ui.warn(format!("No packages matching ident {} for {} exist in the '{}' \
+                                 channel. Check the package ident, target, channel and builder \
+                                 url for correctness",
```
baumanj: Including the `url` in the output would be helpful for debugging. Since it can be overridden by environment variables, it may not be obvious to the user which URL is being used.

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+                                ident, target, self.channel))?;
+                Err(Error::from(CommonError::PackageNotFound(
+                    format!("{} for {} in channel {}", ident, target, self.channel).to_string(),
```
baumanj: The `format` macro returns a `String` already
```suggestion
                    format!("{} for {} in channel {}", ident, target, self.channel),
```

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+            Err(Error::APIClient(APIError(StatusCode::NOT_FOUND, _))) => {
+                // In install we attempt to recommend a channel to look in. That's a bit of a
+                // heavyweight process, and probably a bad idea in the context of
+                // what's a normally a batch process. It might be OK to fall back to
+                // the stable channel, but for now, error.
+                ui.warn(format!("No packages matching ident {} for {} exist in the '{}' \
+                                 channel. Check the package ident, target, channel and builder \
+                                 url for correctness",
+                                ident, target, self.channel))?;
+                Err(Error::from(CommonError::PackageNotFound(
```
baumanj: OPTIONAL: personally, I find using `.into()` easier to read than `Error::from`:
```rust
                Err(CommonError::PackageNotFound(format!("{} for {} in channel {}",
                                                         ident, target, self.channel)).into())
```
vs
```rust
                Err(Error::from(CommonError::PackageNotFound(
                    format!("{} for {} in channel {}", ident, target, self.channel),
                )))
```

Similarly elsewhere

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+        {
+            return Err(Error::from(CommonError::DownloadFailed(format!(
+                "We tried {} times but could not download {} for {}. Last error was: {}",
+                RETRIES, ident, target, err
+            ))));
+        }
+
+        // At this point the artifact is in the download directory...
+        let mut artifact = PackageArchive::new(self.downloaded_artifact_path(ident, target));
+        self.verify_artifact(ui, ident, target, &mut artifact)?;
```
baumanj: I thought we only wanted to do this conditionally based on `self.verify`, no?

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,439 @@
...
+                          ident: &PackageIdent,
+                          target: PackageTarget,
+                          artifact: &mut PackageArchive)
+                          -> Result<()>
+        where T: UIWriter
+    {
+        // We need to look at the artifact to know the signing keys to fetch
+        // Once we have them, it's the natural time to verify.
+        // Otherwise, it might make sense to take this fetch out of the verification code.
+        let nwr = artifact::artifact_signer(&artifact.path)?;
```
baumanj: OPTIONAL: `nwr` doesn't mean anything to me at first glance, so I think it might improve clarity to consider a different name.

components/hab/src/command/pkg/download.rs
chefsalim: :+1:

components/hab/src/main.rs
```diff
@@ -1572,6 +1603,68 @@ fn install_sources_from_matches(matches: &ArgMatches<'_>) -> Result<Vec<InstallS
...
+                buf_reader.lines()
+                          .map(|x| expand_line(&x.unwrap(), filename))
+                          .collect(); // Collect takes Vec<Result<x>> to Result<Vec<x>>
+            packages_or_nothing?.into_iter()
+                                .filter(std::option::Option::is_some)
+                                .for_each(|y| sources.push(y.unwrap()));
+        }
+    }
+    Ok(sources)
+}
```
baumanj: With the up-front validation I alluded to before, this can be simplified to basically
```rust
    let s = std::fs::read_to_string(path).map_err(habitat_core::error::Error::IO)?;

    s.lines()
     .map(|line| PackageIdent::from_str(line.trim()))
     .collect()
```
eliminating the need for `ident_from_str_helper` and `expand_line`. I'd suggest putting that logic in a shared place (probably `habitat_common`) so that it can be shared by the validation code and the parsing args into `PackageIdent`s code.

components/hab/src/main.rs
```diff
@@ -1572,6 +1603,68 @@ fn install_sources_from_matches(matches: &ArgMatches<'_>) -> Result<Vec<InstallS
...
+
+// Error handling would be nice here; if nothing else dump the offending string
+pub fn expand_line(line: &str, file: &str) -> Result<Option<PackageIdent>> {
```
baumanj: I think the `pub` can be removed

components/hab/src/command/pkg/download.rs
```diff
@@ -0,0 +1,438 @@
...
+        for ident in &self.idents {
+            let package = self.determine_latest_from_ident(ui, &ident.clone(), self.target)?;
+            expanded_packages.push(package);
+        }
+
+        // Collect all the expanded deps into one structure
+        // Done separately because it's not as easy to parallelize
+        for package in expanded_packages {
+            for ident in package.tdeps {
+                // Box is here to avoid E0515; the temporary value
```
chefsalim: Remove comment?

PR #6952, opened 2019-09-11
Add e2e scripts for powershell

smacfarlane:
This adds a setup_environment script for end-to-end testing on Windows, equivalent to the Linux version.   It also adds a `e2e_local.ps1` for running Windows end-to-end tests on a developer workstation.  

Note:  The chefes/buildkite-windows container doesn't appear to set `HAB_LICENSE` so it was added to our e2e_env file. 


e2e_local.ps1
```diff
@@ -0,0 +1,25 @@
+$ErrorActionPreference="stop" 
+
+if ($Args.Count -eq 0) {
```
mwrock: A more powershell idiomatic way to deal with this would be to put this at the top of the script:
```
Param(
   [Parameter(Mandatory=$true)]
   [string[]]$TestArgs
)
```
If the user dose not supply the args, they will be prompted:

```
C:\dev\habitat [named_pipe]> .\e2e_local.ps1                                                                                                                                                                                                         cmdlet test.ps1 at command pipeline position 1
Supply values for the following parameters:
TestArgs[0]: blah
TestArgs[1]:
```

PR #6953, opened 2019-09-11
Error when sys.ip cannot be determined

### LINKED ISSUE #6913: Return error when sys.ip cannot be determined
Epic: TBD
Related Issue: #5079 
Design Doc: #6793 

Right now, when a system's IP address cannot be determined through creating a request to Google DNS, it defaults to 127.0.0.1. This is causing confusion and errors for our users. This task is to return an error when Google DNS is not routable, rather than returning 127.0.0.1.

This is the first of two issues associated with implementing this epic.  The other issue is #6914

davidMcneil:
Resolves #6913

This PR depends on #6949 so the first [commit](https://github.com/habitat-sh/habitat/pull/6953/commits/89a259cfc4d9113c16ca8c81650549ea94125509) can be ignored

If you try and start the supervisor without connectivity to `8.8.8.8`, the supervisor immediately exits with:
```text
hab-sup(MR): core/hab-sup (core/hab-sup/0.85.0)
Failed to discover this hosts outbound IP address: Network is unreachable (os error 101)
[2019-09-11T18:31:56Z ERROR hab_launch] Launcher exiting with code 86
```

Not sure the best way to add automated testing for this. Any suggestions?

### DIFF
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -101,8 +101,6 @@ pub enum Error {
     InvalidPathString(ffi::OsString),
     /// Occurs when making lower level IO calls.
     IO(io::Error),
-    /// Occurs when this hosts outbound IP address cannot be determined
-    IpLookupFailed(io::Error),
     /// Errors when joining paths :)
     JoinPathsError(env::JoinPathsError),
     // When LogonUserW does not have the correct logon type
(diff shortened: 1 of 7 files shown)


christophermaier:
@davidMcneil I haven't read through this PR yet, but for testing you might be able to cook up some kind of Docker-based thing where the networking is really locked down. I think the work I'm currently doing around testing will provide most of the bits needed for that, so I'm OK with waiting until _that_ work lands before having an automated test for this (otherwise I think you would probably end up reinventing the wheel).

davidMcneil:
Thanks for the great feedback @baumanj ! I implemented all of it.

Issue #6903, opened 2019-08-26
Upload set of packages from custom directory


dmccown:

jeremymv2:
`hab pkg upload /path/to/harts/<fileglob>` seems to already provide this functionality - is it missing a feature?

jeremymv2:
Actually, file globbing can be susceptible to the [MAX_ARG_STRLEN](https://unix.stackexchange.com/a/120842) limit and really wouldn't be a reliable cross-platform method when uploading a directory full of thousands of .harts.

I think a better solution would be to add a `--file <HART_PATHS_TO_UPLOAD>` argument, as I proposed [here](https://github.com/habitat-sh/habitat/issues/6878#issuecomment-528600086).

PR #6958, opened 2019-09-13
`hab pkg bulkupload` packages from a directory

jeremymv2:
Closes https://github.com/habitat-sh/habitat/issues/6903
Closes https://github.com/habitat-sh/habitat/issues/6981

This PR adds a new subcommand: `hab pkg bulkupload` which uploads artifacts from a directory created by a [new](https://github.com/habitat-sh/habitat/pull/6945) command: `hab pkg download` .

@chefsalim  @markan FYI

The purpose of this PR is to help simplify the process of bootstrapping a Depot. It takes into account the internal discussion we've had around the UX.

Additionally, it allows for HTTP 201 responses when keys or packages are uploaded for the first time or via the `--force` argument.

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -169,7 +169,7 @@ impl BuilderAPIClient {
                 -> Result<PathBuf> {
         debug!("Downloading file to path: {}", dst_path.display());
         let mut resp = self.maybe_add_authz(rb, token).send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         fs::create_dir_all(&dst_path)?;
         let file_name = resp.get_header(X_FILENAME)?;
@@ -260,7 +260,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
         };
 
         let mut resp = self.0.get_with_custom_url(&path, custom).send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         Ok(resp.json()?)
     }
@@ -282,7 +282,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
         };
 
         let mut resp = self.0.get_with_custom_url(&path, custom).send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         Ok(resp.json()?)
     }
@@ -337,7 +337,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
                                u.set_query(Some(&format!("target={}", &target.to_string())))
                            })
                            .send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         let mut encoded = String::new();
         resp.read_to_string(&mut encoded).map_err(Error::IO)?;
@@ -377,7 +377,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .bearer_auth(token)
             .json(&body)
             .send()?
-            .ok_if(StatusCode::NO_CONTENT)
+            .ok_if(&[StatusCode::NO_CONTENT])
     }
 
     /// Cancel a job group
@@ -394,7 +394,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .post(&url)
             .bearer_auth(token)
             .send()?
-            .ok_if(StatusCode::NO_CONTENT)
+            .ok_if(&[StatusCode::NO_CONTENT])
     }
 
     /// Download a public encryption key from a remote Builder to the given filepath.
@@ -433,7 +433,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .bearer_auth(token)
             .json(&body)
             .send()?
-            .ok_if(StatusCode::CREATED)
+            .ok_if(&[StatusCode::CREATED])
     }
 
     /// Create secret for an origin
@@ -460,7 +460,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .bearer_auth(token)
             .json(&body)
             .send()?
-            .ok_if(StatusCode::CREATED)
+            .ok_if(&[StatusCode::CREATED])
     }
 
     /// Delete a secret for an origin
@@ -478,7 +478,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .delete(&path)
             .bearer_auth(token)
             .send()?
-            .ok_if(StatusCode::NO_CONTENT)
+            .ok_if(&[StatusCode::NO_CONTENT])
     }
 
     /// Delete an origin
@@ -497,7 +497,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .delete(&path)
             .bearer_auth(token)
             .send()?
-            .ok_if(StatusCode::NO_CONTENT)
+            .ok_if(&[StatusCode::NO_CONTENT])
     }
 
     /// List all secrets keys for an origin
@@ -510,7 +510,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
 
         let path = format!("depot/origins/{}/secret", origin);
         let mut resp = self.0.get(&path).bearer_auth(token).send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         let mut encoded = String::new();
         resp.read_to_string(&mut encoded)
@@ -567,7 +567,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
         debug!("Showing origin keys: {}", origin);
 
         let mut resp = self.0.get(&origin_keys_path(origin)).send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         let mut encoded = String::new();
         resp.read_to_string(&mut encoded)
@@ -602,7 +602,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
 
         let mut resp = self.maybe_add_authz(self.0.get_with_custom_url(&path, custom), token)
                            .send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         let mut encoded = String::new();
         resp.read_to_string(&mut encoded)
@@ -647,10 +647,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             Body::sized(file, file_size)
         };
         let mut resp = self.0.post(&path).bearer_auth(token).body(body).send()?;
-        match resp.status() {
-            StatusCode::OK | StatusCode::CREATED => Ok(()),
-            _ => Err(err_from_response(&mut resp)),
-        }
+        resp.ok_if(&[StatusCode::OK, StatusCode::CREATED])
     }
 
     /// Upload a secret origin key to a remote Builder.
@@ -687,7 +684,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             Body::sized(file, file_size)
         };
         let mut resp = self.0.post(&path).bearer_auth(token).body(body).send()?;
-        resp.ok_if(StatusCode::OK)
+        resp.ok_if(&[StatusCode::OK])
     }
 
     /// Download the latest release of a package.
@@ -751,7 +748,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
                                    }),
                              token)
             .send()?
-            .ok_if(StatusCode::OK)
+            .ok_if(&[StatusCode::OK])
     }
 
     /// Returns a package struct for the latest package.
@@ -783,7 +780,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
                                                 }),
                                             token)
                            .send()?;
-        resp.ok_if(StatusCode::OK)?;
+        resp.ok_if(&[StatusCode::OK])?;
 
         let mut encoded = String::new();
         resp.read_to_string(&mut encoded)
@@ -853,10 +850,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
                            .body(body)
                            .send()?;
 
-        match resp.status() {
-            StatusCode::OK | StatusCode::CREATED => Ok(()),
-            _ => Err(err_from_response(&mut resp)),
-        }
+        resp.ok_if(&[StatusCode::OK, StatusCode::CREATED])
     }
 
     fn x_put_package(&self, pa: &mut PackageArchive, token: &str) -> Result<()> {
@@ -885,7 +879,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .bearer_auth(token)
             .body(body)
             .send()?
-            .ok_if(StatusCode::OK)
+            .ok_if(&[StatusCode::OK])
     }
 
     /// Delete a package from Builder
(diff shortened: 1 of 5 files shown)


components/hab/src/main.rs
```diff
@@ -51,10 +51,13 @@ use habitat_core::{crypto::{init,
                    env::{self as henv,
                          Config as _},
                    fs::{cache_artifact_path,
+                        cache_key_path,
+                        cache_path,
```
baumanj: Is there a commit missing? I don't see any definition of `cache_path` in https://github.com/habitat-sh/habitat/blob/master/components/core/src/fs.rs.

components/hab/src/main.rs
```diff
@@ -51,10 +51,13 @@ use habitat_core::{crypto::{init,
...
                         launcher_root_path},
                    os::process::ShutdownTimeout,
                    package::{target,
                              PackageIdent,
+                             PackageIdentTarget,
```
baumanj: Same question; where is `PackageIdentTarget` defined?

eeyun:
The behavior here to me feels like it warrants a different sub-command. I know we want to limit sub-command sprawl in our client so others feel free to chime in and overrule me but this feels like it's logically overloading the `pkg upload` command.

components/hab/src/main.rs
jeremymv2: Ah yes, these are coming from the PR that this PR is dependent upon to merge first.
https://github.com/habitat-sh/habitat/pull/6945

When it merges I planned to rebase on master.

baumanj: In the meantime, can you rebase on top of `ma/hab_6878/pkg_download`? That will allow CI to run and simplify the review process.

jeremymv2: Of course, gladly! Thanks for your patience with me. Rebase is now complete 😄

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+                                       require_literal_leading_dot: true, };
+    let artifact_paths =
+        vec_from_glob_with(&artifact_path.join("*.hart").display().to_string(), options);
+
+    ui.begin(format!("Preparing to upload hartifacts to the '{}' channel on {}",
+                     additional_release_channel.clone()
+                                               .unwrap_or_else(ChannelIdent::unstable),
+                     url))?;
+    ui.status(Status::Using,
+              format!("{} for hartifacts and {} for signing keys.",
```
chefsalim: typo - "hartifacts"

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+                      &artifact_path.display(),
+                      &key_path.display()))?;
+    ui.status(Status::Found,
+              format!("{} hartifacts for upload.", artifact_paths.len()))?;
```
chefsalim: typo - "hartifacts"

jeremymv2: fixed in https://github.com/habitat-sh/habitat/pull/6958/commits/2b025955395c6b9ced76e3219810ebbff89636ec

components/hab/src/main.rs
jeremymv2: fixed in https://github.com/habitat-sh/habitat/pull/6958/commits/2b025955395c6b9ced76e3219810ebbff89636ec

markan:
Did we want to protect this with a feature flag as an experimental command?

jeremymv2:
> Did we want to protect this with a feature flag as an experimental command?

@markan @chefsalim @apriofrost 

The UX looks like this at the moment. The question then is that enough of an attention grabber/disclaimer?

```
vagrant@chef-builder:~/on-prem-builder$ hab pkg bulkupload --help
hab-pkg-bulkupload 0.87.0-dev

!!! ATTENTION !!!
This command is ALPHA and subject to change.

Bulk Uploads Habitat Artifacts to a Depot from a local directory.

USAGE:
    hab pkg bulkupload [FLAGS] [OPTIONS] <UPLOAD_DIRECTORY>

FLAGS:
        --auto-build    Enable auto-build for all packages in this upload. Only applicable to SaaS Builder. (default:
                        false)
        --force         Skips checking availability of package and force uploads, potentially overwriting a stored copy
                        of a package. (default: false)
    -h, --help          Prints help information
    -V, --version       Prints version information

OPTIONS:
    -u, --url <BLDR_URL>       Specify an alternate Depot endpoint. If not specified, the value will be taken from the
    -c, --channel <CHANNEL>    Optional additional release channel to upload package to. Packages are always uploaded to
                               `unstable`, regardless of the value of this option.

ARGS:
    <UPLOAD_DIRECTORY>    Directory Path from which artifacts will be uploaded.
vagrant@chef-builder:~/on-prem-builder$
```

jeremymv2:
I removed the alpha messaging from the help menu.

components/hab/src/cli.rs
```diff
@@ -545,6 +545,27 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
...
+                (aliases: &["bul", "bulk"])
+                (@arg BLDR_URL: -u --url +takes_value {valid_url} "Specify an alternate Depot \
+                    endpoint. If not specified, the value will be taken from the HAB_BLDR_URL \
+                    environment variable if defined. (default: https://bldr.habitat.sh)")
+                (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
+                (@arg CHANNEL: --channel -c +takes_value
+                    "Optional additional release channel to upload package to. \
+                     Packages are always uploaded to `unstable`, regardless \
+                     of the value of this option.")
+                (@arg FORCE: --force "Skips checking availability of package and \
```
apriofrost: Sorry for the typo in the LucidChart doc, "Skips" should be "Skip" here.

components/hab/src/cli.rs
```diff
@@ -545,6 +545,27 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
...
+                    force uploads, potentially overwriting a stored copy of a package. \
+                    (default: false)")
+                (@arg AUTO_BUILD: --("auto-build") "Enable auto-build for all packages in this upload. Only applicable to SaaS Builder. \
+                    (default: false)")
```
apriofrost: Since the `--force` and `--auto-rebuild` are flags, we should drop the "(default: false)" at the end of the descriptions. We initially added these thinking they were options, but then we changed them to flags. I should have removed them in the Lucidchart doc, sorry about that. 😬

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+        BuildOnUpload::PackageDefault
+    } else {
+        BuildOnUpload::Disable
+    };
+    let token = auth_token_param_or_env(&m)?;
+    let options = glob::MatchOptions { case_sensitive:              true,
+                                       require_literal_separator:   true,
+                                       require_literal_leading_dot: true, };
+    let artifact_paths =
+        vec_from_glob_with(&artifact_path.join("*.hart").display().to_string(), options);
```
baumanj: OPTIONAL: I think it would enhance clarity to handle the type conversions inside the helper function, leaving the easier to read call here:
```rust
    let artifact_paths = vec_from_glob_with(artifact_path.join("*.hart"));
```

baumanj: OPTIONAL: the type returned is inherent in the signature, so I think the name could be simplified to `glob_with`.

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+
+    ui.begin(format!("Preparing to upload artifacts to the '{}' channel on {}",
+                     additional_release_channel.clone()
+                                               .unwrap_or_else(ChannelIdent::unstable),
```
baumanj: Is defaulting to the unstable channel an inherent assumption of the builder API? I worry about the fact that this could become misleading if that were to ever change and there's nothing that ties the modification of the two together. Perhaps it would be safer to require the channel specification to be explicit.

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     command::pkg::sign::start(ui, &pair, &src, &dst)
 }
 
+fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
+    let upload_dir = bulkupload_dir_from_matches(m);
+    // upload_dir is required via clap and exists
+    let artifact_path = upload_dir.as_ref().unwrap().join("artifacts");
+    let key_path = upload_dir.as_ref().unwrap().join("keys");
+    let url = bldr_url_from_matches(&m)?;
```
baumanj: `m` is already a reference to `ArgMatches`, so there's no need to take an additional reference:
```suggestion
    let url = bldr_url_from_matches(m)?;
```
Similarly with `channel_from_matches` and `auth_token_param_or_env`.

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+        vec_from_glob_with(&artifact_path.join("*.hart").display().to_string(), options);
+
+    ui.begin(format!("Preparing to upload artifacts to the '{}' channel on {}",
+                     additional_release_channel.clone()
+                                               .unwrap_or_else(ChannelIdent::unstable),
+                     url))?;
+    ui.status(Status::Using,
+              format!("{} for artifacts and {} for signing keys.",
+                      &artifact_path.display(),
+                      &key_path.display()))?;
```
baumanj: The `&`s are unnecessary here. `Path::display()` returns an owned `String`.

In general, I find that I get the simplest code when I leave `&`s out until the compiler makes me put them in.

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+
+    for artifact_path in &artifact_paths {
+        command::pkg::upload::start(ui,
+                                    &url,
+                                    &additional_release_channel,
+                                    &token,
+                                    &artifact_path,
+                                    force_upload,
+                                    auto_build,
+                                    &key_path)?;
```
baumanj: It's unnecessary to have the `&` both before `artifact_paths` (in L877) *and* before `artifact_path`.

The `for` loop implicitly calls [`into_iter`](http://doc.rust-lang.org/1.38.0/std/iter/trait.IntoIterator.html) on the value after `in` which will return an iterator over `PathBuf`s given `artifact_paths` and an iterator over `&PathBuf`s given `&artifact_paths`. There's a good explanation here: http://xion.io/post/code/rust-for-loop.html.

I suggest removing the `&` from  `artifact_paths` and leaving it on `&artifact_path`.

components/hab/src/cli.rs
jeremymv2: Fixed in https://github.com/habitat-sh/habitat/pull/6958/commits/ae2f142bfe52399952e722020275aa16341d6290

components/hab/src/cli.rs
jeremymv2: Fixed in https://github.com/habitat-sh/habitat/pull/6958/commits/ae2f142bfe52399952e722020275aa16341d6290

apriofrost: 👍

components/builder-api-client/src/builder.rs
```diff
@@ -647,7 +647,10 @@ impl BuilderAPIProvider for BuilderAPIClient {
             Body::sized(file, file_size)
         };
         let mut resp = self.0.post(&path).bearer_auth(token).body(body).send()?;
-        resp.ok_if(StatusCode::OK)
+        match resp.status() {
+            StatusCode::OK | StatusCode::CREATED => Ok(()),
+            _ => Err(err_from_response(&mut resp)),
+        }
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/6958/commits/a93ad69be0311206fd13c19bc4d529e8448ccc63

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     command::pkg::sign::start(ui, &pair, &src, &dst)
 }
 
+fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
+    let upload_dir = bulkupload_dir_from_matches(m);
+    // upload_dir is required via clap and exists
+    let artifact_path = upload_dir.as_ref().unwrap().join("artifacts");
```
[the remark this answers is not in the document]
jeremymv2: Addressed in https://github.com/habitat-sh/habitat/pull/6958/commits/849538aa924d3227622e62130ae677be62d665dc

components/hab/src/main.rs
jeremymv2: Addressed in https://github.com/habitat-sh/habitat/pull/6958/commits/bc5b9642d03ce480fd4573ae05eaef724b3c12a7

components/hab/src/main.rs
jeremymv2: Addressed in https://github.com/habitat-sh/habitat/pull/6958/commits/bc5b9642d03ce480fd4573ae05eaef724b3c12a7

components/hab/src/main.rs
jeremymv2: Addressed in https://github.com/habitat-sh/habitat/pull/6958/commits/bc5b9642d03ce480fd4573ae05eaef724b3c12a7

components/hab/src/main.rs
```diff
@@ -841,6 +843,51 @@ fn sub_pkg_sign(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+    let url = bldr_url_from_matches(&m)?;
+    let additional_release_channel = channel_from_matches(&m);
+    let force_upload = m.is_present("FORCE");
+    let auto_build = if m.is_present("AUTO_BUILD") {
+        BuildOnUpload::PackageDefault
+    } else {
+        BuildOnUpload::Disable
+    };
+    let token = auth_token_param_or_env(&m)?;
+    let options = glob::MatchOptions { case_sensitive:              true,
```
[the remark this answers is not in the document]
jeremymv2: Addressed in https://github.com/habitat-sh/habitat/pull/6958/commits/8d25f6ff2db26abbef10200c0d17313f933a8e1a

components/hab/src/main.rs
jeremymv2: I don't think that would ever change though. From my understanding, `unstable` is designed as the default birthing channel for all artifacts and the one in which they all must reside even if no other.

components/hab/src/main.rs
jeremymv2: Since `glob::glob_with` is already a function I'm importing I went with the `vec_from_glob_with`.

baumanj: Yeah, not a big deal either way. My recommendation would be to remove the `use glob::glob_with;`, since you can still refer to `glob::glob_with` using that fully-qualified name. I think it makes things clearer to call into other crates with an at least partially qualified name and leave unqualified function names for only crate local things. Having a local `glob_with` that calls `glob::glob_with` internally is perfectly acceptable in my opinion.

But as I said, not a big deal, so do what you like.

Issue #6798, opened 2019-08-07
[certs] Update docs with new cert usage


chefsalim:
Related to https://github.com/habitat-sh/habitat/pull/6759

We need to update the docs to outline the new cert behavior and examples of various scenarios

chefsalim:
PR for first part of the cert update (for Builder) is here - https://github.com/habitat-sh/on-prem-builder/pull/182

Once Habitat cli releases with the latest changes, there will be a follow-up PR

PR #6959, opened 2019-09-13
Update docs for new cert handling

chefsalim:
Resolves https://github.com/habitat-sh/habitat/issues/6798

This change adds documentation for using the new custom certificate model with hab 0.85.0.

Since the most common usage is to connect to Builder, it is added to the section on working with Builder.

Screenshot from 2019-09-13 11-17-33

### DIFF
--- a/www/source/docs/using-builder.html.md.erb
+++ b/www/source/docs/using-builder.html.md.erb
@@ -517,7 +517,7 @@ If you attempt to perform an operation using the Habitat client that talks to a 
 ✗✗✗
 ```
 
-One way to deal with this issue is to use the `SSL_CERT_FILE` environment variable, and point it to the path of the custom certificate prior to performing the client operation.
+One option to remediate this error is to define a `SSL_CERT_FILE` environment variable pointing to the custom certificate path before performing the client operation.
 
 With the release of Habitat 0.85.0, there is another option that is now available. Habitat will now look for custom certificates in its `~/.hab/cache/ssl` directory (or `/hab/cache/ssl` if running as root). Multiple certificates (for example, a self-signed certificate, and a custom certificate authority certificate) can be copied over to the cache directory and will become automatically avaialable for use by the Habitat client.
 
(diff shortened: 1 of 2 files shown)


www/source/docs/using-builder.html.md.erb
```diff
@@ -504,3 +505,22 @@ Currently, our on-premises Builder depot only stores packages for download and u
...
+✗✗✗
+```
+
+One way to deal with this issue is to use the `SSL_CERT_FILE` environment variable, and point it to the path of the custom certificate prior to performing the client operation.
+
+With the release of Habitat 0.85.0, there is another option that is now available. Habitat will now look for custom certificates in its `~/.hab/cache/ssl` directory (or `/hab/cache/ssl` if running as root). Multiple certificates (for example, a self-signed certificate, and a custom certificate authority certificate) can be copied over to the cache directory and will become automatically avaialable for use by the Habitat client.
+
+The `/hab/cache/ssl` directory is also made available inside a Habitat Studio. So, if you have populated the cache directory before entering the Studio, you will find that its contents are populated with the files from outside the Studio. In addition, when entering the Habitat Studio, the `SSL_CERT_FILE` environment variable will be propagated to the Studio, and the file that it is pointing to will also be copied over to the `/hab/cache/ssl` directory.
```
kagarmoe:
```suggestion
The `/hab/cache/ssl` directory is also available inside a Habitat Studio. As long as the certificates are inside the cache directory before you enter the Studio, you'll also find them inside the Studio. In addition, if you've set the `SSL_CERT_FILE` environment variable, you'll also find both it and the file that it points to inside the Studio`/hab/cache/ssl` directory.
```

I think this is still a little awkward and could be refined more.

www/source/docs/using-builder.html.md.erb
```diff
@@ -504,3 +505,22 @@ Currently, our on-premises Builder depot only stores packages for download and u
...
+
+Note: The filename `cert.pem` is reserved for Habitat, so please do not use that file name when copying certs into the cache directory.
```
kagarmoe:
```suggestion
Note: The `cert.pem` file name is reserved for Habitat. Do not use `cert.pem` as a file name when copying certs into the cache directory.
```

kagarmoe: If I try this, will I overwrite my existing `cert.pem`--and thus my entire cert chain? Or does it simply refuse to be overwritten? I haven't tried it yet.

chefsalim:
@kagarmoe Your suggestions have been committed to the PR

Issue #5098, opened 2018-05-18
Supervisor cannot bind to IPv6 addresses, should bind both v4 and v6 by default


HT154:
I noticed that the supervisor only binds to `0.0.0.0:9631` and `0.0.0.0:9638`. I'm trying to get Habitat working in an IPv6-only environment. When I attempted to set the listen addresses via the `hab sup run` command, I encountered errors. This appears to be a bug in hab-sup. Once this is fixed, the supervisor's default behavior should be to bind to both `0.0.0.0` and `[::]` addresses.

OS: Oracle Linux 7.4
Backtrace:
```
# RUST_LOG=debug RUST_BACKTRACE=1 DEBUG=1 hab sup run --listen-gossip '[::]:9639' --override-name test --listen-http '[::]:9632'
DEBUG 2018-05-18T17:30:23Z: habitat_common::ui: InputStream(stdin): { is_a_terminal(): true }
DEBUG 2018-05-18T17:30:23Z: habitat_common::ui: OutputStream(stdout): { is_colored(): true, supports_color(): true, is_a_terminal(): true }
DEBUG 2018-05-18T17:30:23Z: habitat_common::ui: OutputStream(stderr): { is_colored(): true, supports_color(): true, is_a_terminal(): true }
DEBUG 2018-05-18T17:30:23Z: habitat_core::os::process::imp: Calling execvp(): ("/hab/pkgs/core/hab-launcher/7241/20180321162126/bin/hab-launch") ["run", "--listen-gossip", "[::]:9639", "--override-name", "test", "--listen-http", "[::]:9632"]
DEBUG 2018-05-18T17:30:23Z: habitat_launcher::server: Starting Supervisor...
hab-sup(MN):
hab-sup(MN): CAUTION: Running more than one Habitat Supervisor is not recommended for most
hab-sup(MN): CAUTION: users in most use cases. Using one Supervisor per host for multiple
hab-sup(MN): CAUTION: services in one ring will yield much better performance.
hab-sup(MN):
hab-sup(MN): CAUTION: If you know what you're doing, carry on!
hab-sup(MN):
DEBUG 2018-05-18T17:30:23Z: habitat_sup::manager: Creating data directory: /hab/sup/test/data
DEBUG 2018-05-18T17:30:23Z: habitat_sup::manager: Creating specs directory: /hab/sup/test/specs
DEBUG 2018-05-18T17:30:23Z: habitat_sup::manager: Creating composites directory: /hab/sup/test/composites
[...]

PR #6960, opened 2019-09-13
Allow user to specify `sys.ip` address

### LINKED ISSUE #6914: Allow a user to specify an IP address when starting a Supervisor
Epic: TBD
Related Issue: #5079
Design Doc: #6793

When starting a supervisor, a user should have the option to specify an IP address to broadcast for the supervisor. This will likely be a command line flag that is passed to the appropriate command.

This is the second of two issues associated with implementing this epic, the first is #6913

davidMcneil:
Resolves #6914 #7010

~~This PR builds off of #6953. The commits that pertain to this PR start at [dfe109e](https://github.com/habitat-sh/habitat/pull/6960/commits/dfe109eba7f08c5d4e5fdd9a520db27ead9036e0).~~

This PR adds the `sys-ip-address` cli option, that allows specifying a fallback `sys.ip` in the case that it cannot be dynamically determined.

~~**Note**: This differs slightly from what is currently proposed in design doc #6793. It allows specifying a network interface or an IP address (instead of only a network interface), and uses the `sys-ip-address` cli option opposed to `network-interface`. If these changes are agreed upon the design doc can be updated.~~

~~Determining the `sys.ip` address uses the following order of precedence:~~
1. ~~Try to dynamically determine an outgoing IP address using the routing table trick.~~
2. ~~If `sys-ip-address` matches a network interface name do the following:~~
    - ~~If there is more than one IPv4 address, there is ambiguity and we produce an error. We do not support multiple `sys.ip`s. Note, we do not support IPv6 so it is ok if there are multiple IPv6 addresses.~~
    - ~~If there is no IPv4 address, produce an error.~~
    - ~~Otherwise, use the network interface's IPv4 address.~~
3. ~~Try to parse `sys-ip-address` as an IP address.~~

Based on additional conversation, it was decieded to remove the ability to lookup IP address by network interface. This significanly simplifies the logic for determining the `sys.ip`:

1. Use the IP address provided by the `sys-ip-address` cli option
2. If no `sys-ip-address` is specified, try to dynamically determine an outgoing IP address using the routing table trick. If that fails, exit the supervisor with an error.

**Additional Considerations**
- [ ] ~~This change introduces the `pnet` crate which requires [special configuration](https://github.com/libpnet/libpnet#windows) on windows. Are we ok adding this crate? We need to integrate the special configuration into our build. We could disable allowing a network interface name on windows?~~
- [ ] UX of the `sys-ip-address` cli option
- [ ] end-to-end tests that spoof a variety of network scenarios. Docker will probably help with this once we have that infrastructure in place?
- [ ] Once the approach is agreed upon, we need to add documentation


components/sup/src/main.rs
```diff
@@ -167,7 +167,9 @@ fn sub_run_rsr_imlw_mlw_gsw(m: &ArgMatches,
     set_supervisor_logging_options(m);
 
     let cfg = mgrcfg_from_sup_run_matches(m, feature_flags)?;
-    let manager = Manager::load_imlw(cfg, launcher)?;
+    let sys_ip = util::determine_sys_ip_address(m.value_of("SYS_IP_ADDRESS"))?;
```
baumanj: OPTIONAL: Assuming we do CLAP-level validation first, adding something like
```rust
fn ipaddr_from_matches(matches: &ArgMatches<'_>, arg_name: &str) -> Option<std::net::IpAddr> {
    matches.value_of(arg_name)
           .and_then(|s| std::net::IpAddr::from_str(s).ok())
}
```
would let us pass an `Option<IpAddr>` into determine_sys_ip_address, obviating the need for that function to handle a possibly invalid string.

components/hab/src/cli.rs
```diff
@@ -1076,6 +1076,12 @@ pub fn sub_sup_run(feature_flags: FeatureFlag) -> App<'static, 'static> {
                                                             Implies NO_COLOR")
                             (@arg HEALTH_CHECK_INTERVAL: --("health-check-interval") -i +takes_value {valid_health_check_interval}
                              "The interval (seconds) on which to run health checks [default: 30]")
+                            (@arg SYS_IP_ADDRESS: --("sys-ip-address") +takes_value
+                             "The fallback IPv4 address if we can not dynamically determine an \
```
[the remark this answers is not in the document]
davidMcneil: Yeah, I am not sure what to do here. The [`Sys`](https://github.com/habitat-sh/habitat/blob/332ef8e89ae2abc2af1ec1494d9d92fedb4ce95c/components/sup/src/manager/sys.rs#L18) struct does use `IpAddr`, but from other conversations, I do not believe we support IPv6 (not sure exactly what all the issues are). Also, from discussions around this issue, we only want to look for IPv4 addresses from a network interface.

I see three paths forward:

* Change the types to actually reflect that we only support IPv4. Not sure how much of an impact this would be? Kinda of a step in the wrong direction removing parts of the code that can support either?
* Investigate the amount of work needed for IPv6 support. #5098
* Leave it as is. The help message acts as a warning to the user.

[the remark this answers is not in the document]
davidMcneil: Thinking about it more, I think we should make the IP address specified by the user be v4 and leave the existing code as it is.

components/hab/src/cli.rs
```diff
@@ -1076,6 +1076,12 @@ pub fn sub_sup_run(feature_flags: FeatureFlag) -> App<'static, 'static> {
                                                             Implies NO_COLOR")
                             (@arg HEALTH_CHECK_INTERVAL: --("health-check-interval") -i +takes_value {valid_health_check_interval}
                              "The interval (seconds) on which to run health checks [default: 30]")
+                            (@arg SYS_IP_ADDRESS: --("sys-ip-address") +takes_value
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [be3aa8](https://github.com/habitat-sh/habitat/pull/6960/commits/be3aa875b5d92f8bfbc6d34142579b2ba2e55d0e).

components/hab/src/cli.rs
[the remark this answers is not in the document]
davidMcneil: Addressed in [be3aa8](https://github.com/habitat-sh/habitat/pull/6960/commits/be3aa875b5d92f8bfbc6d34142579b2ba2e55d0e).

components/hab/src/cli.rs
```diff
@@ -1076,6 +1076,12 @@ pub fn sub_sup_run(feature_flags: FeatureFlag) -> App<'static, 'static> {
...
+                             "The fallback IPv4 address if we can not dynamically determine an \
+                             outgoing IPv4 address. This value can be a network interface name \
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [be3aa8](https://github.com/habitat-sh/habitat/pull/6960/commits/be3aa875b5d92f8bfbc6d34142579b2ba2e55d0e). I also updated the initial comment. We are no longer supporting determining an IP address from a network interface.

components/sup/src/main.rs
davidMcneil: Addressed in [be3aa8](https://github.com/habitat-sh/habitat/pull/6960/commits/be3aa875b5d92f8bfbc6d34142579b2ba2e55d0e).

predominant:
I'd always thought this should be a per-service option, not per-supervisor. That doesn't negate the work here, I think this is a good start, but most of my environments deal with servers on multiple networks, with each service assigned to various networks.

A great example is a proxy server than handles inbound and outbound requests, requires each proxy service on a different NIC/Address.

apriofrost:
image
I visualized the new UX flow with the above chart and posted my comments on it. Let me know if I misunderstood anything. Thank you!

predominant:
0.0.0.0 should be valid.

stevendanna:
@predominant I think the challenge with `0.0.0.0` is that this is the address we _broadcast_ to other nodes about who to contact us on.  In that sense, `0.0.0.0` wouldn't mean to make sense to actually render into service configuration for a service on some other machine since it wouldn't contain the information we want (i.e. what address do I contact this service on).  I could see accepting 0.0.0.0 to mean "broadcast all addresses", but at the moment we only allow a single address, so I think at best `0.0.0.0` could mean "use dynamic ip lookup for me" (i.e. "use the left side of that decision tree")

davidMcneil:
@apriofrost thanks for the diagram and feedback! It is very helpful.

On the left side of the tree, the error message is actually `Failed to discover this host's outbound IP address: {}` not `Can't parse IP address or socket address: {}`. I agree that pointing a user to `--sys-ip-address` would be very helpful. Commit [00cf6b](https://github.com/habitat-sh/habitat/pull/6960/commits/00cf6bc3087b9e18e76f5d5609c4468ba94d80f5) adds the message `Consider setting the '--sys-ip-address' option.`

I am not sure it would be helpful for the right side of the tree to suggest removing `--sys-ip-address`. This error indicates a typo. If they are setting this flag it is probably because determining the IP address dynamically does not work or the user has a specific IP address they want to use.

apriofrost:
@davidMcneil All your suggestions make sense to me, and I agree. Thanks for improving the error message!

components/hab/src/cli.rs
```diff
@@ -1326,6 +1331,14 @@ fn file_exists_or_stdin(val: String) -> result::Result<(), String> {
     }
 }
 
+#[allow(clippy::needless_pass_by_value)] // Signature required by CLAP
+fn valid_ipv4_address(val: String) -> result::Result<(), String> {
+    match Ipv4Addr::from_str(&val) {
+        Ok(_) => Ok(()),
+        Err(_) => Err(format!("'{}' is not a valid IPv4 address, eg: '0.0.0.0'", val)),
```
raskchanky: I can't decide if this is a problem worth fixing or not, but my concern with this particular error message is it might lead people who receive it to pass `--sys-ip-address 0.0.0.0` to the supervisor, which, while a valid IP address, is not a good choice for `sys.ip`.  I might consider changing it to something else, perhaps a random address from one of the reserved IP blocks: `10.0.0.0/8`, `192.168.0.0/16`, or `172.16.0.0/12`.

davidMcneil: Great point! Changed in [188452](https://github.com/habitat-sh/habitat/pull/6960/commits/1884529be226c3eb326c13adb2c70a1e49accd87).

components/sup/src/main.rs
```diff
@@ -173,7 +176,15 @@ fn sub_run_rsr_imlw_mlw_gsw_smw_rhw_msw(m: &ArgMatches,
     set_supervisor_logging_options(m);
 
     let cfg = mgrcfg_from_sup_run_matches(m, feature_flags)?;
-    let manager = Manager::load_imlw(cfg, launcher)?;
+
+    let maybe_sys_ip = m.value_of("SYS_IP_ADDRESS")
+                        .and_then(|s| IpAddr::from_str(s).ok());
+    let sys_ip = maybe_sys_ip.unwrap_or_else(|| {
+                     habitat_core::util::sys::ip().unwrap_or(IpAddr::V4(Ipv4Addr::LOCALHOST))
```
stevendanna: Can we issue a warning with the error in the case of failure?

components/sup/src/main.rs
```diff
@@ -173,7 +176,15 @@ fn sub_run_rsr_imlw_mlw_gsw_smw_rhw_msw(m: &ArgMatches,
...
+                 });
```
baumanj: OPTIONAL: This can be simplified and made easier to read and understand:
```rust
let sys_ip = m.value_of("SYS_IP_ADDRESS")
     .and_then(|s| IpAddr::from_str(s).ok())
     .or_else(|| habitat_core::util::sys::ip().ok())
     .unwrap_or_else(|| IpAddr::V4(Ipv4Addr::LOCALHOST));
```
and if we're willing to change `habitat_core::util::sys::ip` to return an `Option` (this is its only caller, so I see no reason not to), we can simplify even further:
```rust
let sys_ip = m.value_of("SYS_IP_ADDRESS")
              .and_then(|s| IpAddr::from_str(s).ok())
              .or_else(habitat_core::util::sys::ip)
              .unwrap_or_else(|| IpAddr::V4(Ipv4Addr::LOCALHOST));
```

gscho:
Gonna ask this question here since the issue was closed. If the purpose of `{{sys.ip}}` is to advertise a supervisors IP address to its peers, why can't we use one of the `--peer` IPs and a known hab UDP port to determine the `{{sys.ip}}`. If I'm not peered with anything, would I ever want my IP to be something other than `127.0.0.1`?

PR #6962, opened 2019-09-16
Update basic local environment setup and documentation

scotthain:

### DIFF
--- a/.expeditor/scripts/verify/run_cargo_test.sh
+++ b/.expeditor/scripts/verify/run_cargo_test.sh
@@ -40,6 +40,8 @@ sudo hab pkg install core/zeromq
 sudo hab pkg install core/protobuf --binlink
 sudo hab pkg install core/rust --binlink
 export SODIUM_STATIC=true # so the libarchive crate links to sodium statically
+export LIBARCHIVE_PREFIX
+LIBARCHIVE_PREFIX="$(hab pkg path core/libarchive)"
 export LIBARCHIVE_STATIC=true # so the libarchive crate *builds* statically
 export OPENSSL_DIR # so the openssl crate knows what to build against
 OPENSSL_DIR="$(hab pkg path core/openssl)"
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -22,8 +22,7 @@ Then, run the system preparation scripts
 
 ```
 sh support/linux/install_dev_0_ubuntu_latest.sh
-sh support/linux/install_dev_9_linux.sh
-```
+``
 
 If you want to run the BATS based integration tests you also need docker installed:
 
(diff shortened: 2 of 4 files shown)


support/setup_local_env.sh
```diff
@@ -0,0 +1,40 @@
...
+export HAB_LICENSE="accept-no-persist"
+sudo -E hab pkg install core/bzip2
+sudo -E hab pkg install core/libarchive
+sudo -E hab pkg install core/libsodium
+sudo -E hab pkg install core/openssl
+sudo -E hab pkg install core/xz
+sudo -E hab pkg install core/zlib
+sudo -E hab pkg install core/zeromq
+sudo -E hab pkg install core/protobuf --binlink
+sudo -E hab pkg install core/rust --binlink
```
christophermaier: As a Rust developer, I don't think I would ever want to binlink `core/rust` on my local development system, since that would be a global change, and `rustup` provides a ton of flexibility for developers.

If someone doing other Rust development were to run this, it could cause headaches for their other projects.

support/setup_local_env.sh
```diff
@@ -0,0 +1,40 @@
...
+LIBZMQ_PREFIX=$(hab pkg path core/zeromq)
+# the libarchive and zmq look to the load lib path (as well as the lib path) for these
+# we get this for free when running hab pkg build
+export LD_LIBRARY_PATH
+LD_LIBRARY_PATH="$(hab pkg path core/zlib)/lib:$(hab pkg path core/bzip2)/lib:$(hab pkg path core/xz)/lib:$(hab pkg path core/libarchive)/lib:$(hab pkg path core/libsodium)/lib:$(hab pkg path core/openssl)/lib:$(hab pkg path core/gcc)/lib:$LIBZMQ_PREFIX/lib"
+# again we get this for free when running hab pkg build, here we have to set it so the lib path is set for cargo
+export LIBRARY_PATH
+LIBRARY_PATH=$LD_LIBRARY_PATH
+export PKG_CONFIG_PATH
+PKG_CONFIG_PATH="$(hab pkg path core/libarchive)/lib/pkgconfig:$(hab pkg path core/libsodium)/lib/pkgconfig:$(hab pkg path core/openssl)/lib/pkgconfig"
```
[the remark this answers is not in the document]
christophermaier: Additionally, it seems like these scripts could be used directly in `verify_cargo_test.sh` to reduce duplication.

Speaking of duplication, should the `LIBARCHIVE_PREFIX` variable you added to `verify_cargo_test.sh` in this PR also be added to this `setup_local_env.sh` script (or what might replace it)?

.expeditor/scripts/verify/run_cargo_test.sh
```diff
@@ -40,6 +40,8 @@ sudo hab pkg install core/zeromq
 sudo hab pkg install core/protobuf --binlink
 sudo hab pkg install core/rust --binlink
```
christophermaier: If we're installing `rustup` and a particular toolchain above, do we need to install `core/rust`?

BUILDING.md
```diff
@@ -22,64 +22,41 @@ Then, run the system preparation scripts
...
 ```
 
-You may also be able to configure your editor to automatically run rustfmt every time you save.
-
-# Compiling habitat binaries
-
 In the root of the `habitat` repo:
 ```
-make
+cargo build
```
christophermaier: One day it would be great to have this be executable documentation and have a pipeline step that executes it on a fresh system to ensure that we can always have a validated way for developers to build the code.

I get that today is probably not that day, though :cry:

support/linux/install_dev_0_ubuntu_latest.sh
```diff
@@ -1,32 +1,12 @@
...
-  sudo \
-  tmux \
-  vim \
-  wget
+# Install gcc and dev tools
+sudo apt-get install -y --no-install-recommends \
+  build-essential
+
+# Install habitat
+export HAB_LICENSE="accept-no-persist"
```
baumanj: I don't think this export will apply to the `sudo bash` below unless you add `-E` to the sudo call.

support/setup_local_env.sh
```diff
@@ -0,0 +1,40 @@
...
+sudo -E hab pkg install core/zlib
+sudo -E hab pkg install core/zeromq
+sudo -E hab pkg install core/protobuf --binlink
+sudo -E hab pkg install core/rust --binlink
+
+# for the libarchive crate
+export LIBARCHIVE_LIB_DIR
+LIBARCHIVE_LIB_DIR="$(hab pkg path core/libarchive)/lib"
+export LIBARCHIVE_INCLUDE_DIR
+LIBARCHIVE_INCLUDE_DIR="$(hab pkg path core/libarchive)/include"
```
baumanj: OPTIONAL: this could be a little less repetitive:
```bash
LIBARCHIVE_PKG_PATH=$(hab pkg path core/libarchive)
export LIBARCHIVE_LIB_DIR="$LIBARCHIVE_PKG_PATH/lib"
export LIBARCHIVE_INCLUDE_DIR="$LIBARCHIVE_PKG_PATH/include"
```
and later where its referenced in `LD_LIBRARY_PATH`.

Similarly for other `hab pkg path` calls.

PR #6982, opened 2019-09-26
Add auth option to tar exporter

chefsalim:
Update tar exporter with auth option
Porting over https://github.com/habitat-sh/habitat/pull/5203/files


chefsalim:
Confirmed this code works

PR #6983, opened 2019-09-26
Add some automated networked tests

christophermaier:
This introduces some `docker-compose` based testing scenarios for
exercising various Habitat behaviors in small networks. Read the
comments in the code for further details.

Right now, the tests are written in Bash, with a very crude test
runner also written in Bash. As a result, they only really run on
Linux right now. However, we are investigating transitioning to
Powershell, as that can run on all our platforms, which means we could
get these same tests running on Windows without a lot of extra work.

One big downside of the current implementation is that it's difficult
to run the tests locally on code you're currently working on (instead,
you provide the name of a channel from which to pull Habitat
artifacts). More work needs to be done on this to ensure that `cargo
build`-built binaries can be self-contained enough to drop into a
container and still work. The benefit of having tests that run in CI,
however, is high, so we're getting this out now.


.expeditor/end_to_end.pipeline.yml
```diff
@@ -18,6 +18,18 @@ steps:
 # E2E
 #######################################################################
 
+  - label: ":docker: Docker End-to-End Supervisor Tests"
+    command:
+      - cd automated_testing
+      # TODO (CM): should be DEV, but somehow it doesn't have the packages I need at the moment?!
+      - ./run_all.sh stable
```
smacfarlane: Testing this now, but my suspicion based on the output of `run_all.sh DEV` is that the `Dockerfile`s need an `ARG` to pass in `HAB_BLDR_URL`.

smacfarlane: Looks like that was it.   I hard-coded `CHANNEL` and `HAB_BLDR_URL` in the `Makefile` to get this to work.  Do you think worth the effort of making `HAB_BLDR_URL` a parameter though at this point? If so, I can work up a patch for you. 

```Step 9/9 : RUN hab pkg install core/hab --binlink --force --channel="${CHANNEL}"
 ---> Running in 5988014cf364
» Installing core/hab
☁ Determining latest version of core/hab in the 'DEV' channel
↓ Downloading core/hab/0.88.0/20191009204957
☛ Verifying core/hab/0.88.0/20191009204957
✓ Installed core/hab/0.88.0/20191009204957
★ Install of core/hab/0.88.0/20191009204957 complete with 1 new packages installed.
» Binlinking hab from core/hab/0.88.0/20191009204957 into /bin
★ Binlinked hab from core/hab/0.88.0/20191009204957 to /bin/hab```

File: automated_testing/defaults/run_test.sh [hunk 2]
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -xeuo pipefail
+
+echo "You should really fix this, you know?"
```
smacfarlane: We probably should :joy_cat:    Is this intended as a placeholder for tests that don't do the right thing? Maybe some verbage indicating that `run_test.sh` for the specific test case needs to be implemented?

automated_testing/testcases/apply_config_from_file/run_test.sh
```diff
@@ -0,0 +1,36 @@
...
+
+set -xeuo pipefail
+
+hab pkg install core/redis
+hab pkg exec core/redis redis-cli --version
+
+hab svc load core/redis --remote-sup=alpha.habitat.dev
+hab svc load core/redis --remote-sup=beta.habitat.dev
+
+# TODO Wait until redis is available
```
smacfarlane: Nothing to change here, just lamenting.  Core-plans has the same issue in that we don't have a great way in our tooling to say 'start this service, and let me know when it's healthy' :(

automated_testing/testcases/new-config-is-gossiped/run_test.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -xeuo pipefail
+
+hab pkg install core/jq-static -b -f
+
+hab svc load core/redis --remote-sup=alpha.habitat.dev
+sleep 15
+hab svc load christophermaier/test-probe --bind=thing_with_a_port:redis.default --remote-sup=beta.habitat.dev
```
smacfarlane: @christophermaier How do you feel about creating a `habitat-testing` or similarly named origin (or maybe just use the `habitat` origin?) and pulling this package into it, along with moving the source to the `habitat-sh` org on github?

automated_testing/Makefile
```diff
@@ -0,0 +1,11 @@
+CHANNEL ?= stable
```
mwrock: I think this `automated_testing` directory would be better located in a folder under `test/end-to-end/` maybe called `multi-supervisor`?

automated_testing/defaults/run_test.sh
mwrock: It looks like this is a test case template. One would run `new_testcase my_test_case` which creates a new directory named `my_test_case` and copy this file and all others in the `defaults` directory to that new directory. So it makes sense that a template for a new test would have a failing test.

automated_testing/testcases/apply_config_from_file/run_test.sh
mwrock: After #7148 is merged, we can convert these to pester tests and leverage `Wait-SupervisorService` that polls a service for a given timeout.

automated_testing/testcases/send_file/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM habitat_integration_base
+
+# Always accept the license when we run this image.
+ENV HAB_LICENSE=accept-no-persist
+
+# We're going to be using docker inside the container.
+#
+# TODO: Consider ensuring that the base container itself has Docker if
+# this becomes more of a pattern.
+RUN hab pkg install core/docker -b -f
```
mwrock: I think it would be nice for this to "just be there". In testing this PR with rolling builds I have used docker in a few tests.

automated_testing/testcases/send_file/docker-compose.override.yml
```diff
@@ -0,0 +1,20 @@
...
+      - alpha
+      - beta
+    volumes:
+      # The test currently requires access to the individual
+      # Supervisor "servers"' file systems. Since this is Docker,
+      # though, we need access to the Docker socket inside the
+      # container, so we can use docker to get into the other docker
+      # containers to check.
+      #
+      # Yo dawg, I heard you like Docker....
```
mwrock: might just add this to `test_base` because dawg, I do like docker

christophermaier: Will do.

automated_testing/testcases/send_file/Dockerfile
christophermaier: Will do.

.expeditor/end_to_end.pipeline.yml
christophermaier: I think it's OK now that we're in production Builder.

automated_testing/testcases/new-config-is-gossiped/run_test.sh
christophermaier: Yes, I want to do this... I think I'll get this after we initially merge, so I can unblock @mwrock

PR #6988, opened 2019-09-30
do not inject studio environment with ctl_secret from cli.toml

mwrock:
This was inspired by [this discussion](https://forums.habitat.sh/t/err-4-secret-key-mismatch/1210/8). If there is a `ctl_secret` in one's `cli.toml` created by `hab setup` then that value is injected into the studio's environment. This only applies to a local studio and not a docker studio since `HAB_CTL_SECRET` is not one of the variables injected into the studio `docker run` call.

Chances are very high (maybe 100%) that this will cause all `hab svc` commands to fail in a local studio because a studio supervisor will run with a generated key that will inevitably be different from the `ctl_secret`.

### DIFF
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -4,7 +4,8 @@ use std::{env,
           path::{Path,
                  PathBuf}};
 
-use crate::{common::ui::UI,
+use crate::{common::ui::{UIWriter,
+                         UI},
             hcore::{crypto::CACHE_KEY_PATH_ENV_VAR,
                     env as henv,
                     fs}};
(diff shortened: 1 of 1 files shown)


components/hab/src/command/studio/enter.rs
```diff
@@ -75,9 +74,6 @@ pub fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
                             config.auth_token,
                             Sensitivity::NoPrintValue);
     set_env_var_from_config(BLDR_URL_ENVVAR, config.bldr_url, Sensitivity::PrintValue);
-    set_env_var_from_config(CTL_SECRET_ENVVAR,
```
christophermaier: It would be good to leave a comment here saying that we are intentionally leaving one of the `cli.toml` fields out, along with why.

It may also be useful to print a message to the user if they have a `CTL_SECRET` set in their environment or `cli.toml` that we are explicitly ignoring it, as well as an explanation why.

(of course, the second suggestion would also do the job of the first suggestion.)

mwrock: Good point! Message added.

Issue #5973, opened 2018-12-12
Overhaul PID handling in the Supervisor


dmccown:
## Description
The way the Supervisor manages PIDs (process identifiers) of the services it manages has been the cause of a handful of bugs and instability, and needs to be revisited.

A number of issues appear to be related to this:
* https://github.com/habitat-sh/habitat/issues/6977
* https://github.com/habitat-sh/habitat/issues/4578
* https://github.com/habitat-sh/habitat/issues/4735
* https://github.com/habitat-sh/habitat/issues/6512
* https://github.com/habitat-sh/habitat/issues/6986
* https://github.com/habitat-sh/habitat/issues/4069
* https://github.com/habitat-sh/habitat/issues/4578
* https://github.com/habitat-sh/habitat/issues/4735
* https://github.com/habitat-sh/habitat/issues/6972

Currently, the Supervisor tracks the PIDs in files on disk. If we no longer had these files, we could instead have the Launcher be the source of truth for PIDs. It already has access to this information, since it is the parent of these processes anyway. If the Supervisor could ask the Launcher what PIDs it should be monitoring, then we could have a better separation of concerns and eliminate a source of instability (corrupted or out-of-date pidfiles).

Additional research needs to be done to fully flesh this idea out.
 
## Acceptance Criteria
* PID 0 should never be tracked as a PID we care about
* When a process is intentionally terminated the Supervisor should never think it needs to restart that process.

PR #6990, opened 2019-10-01
Add some "PID 0" protections

christophermaier:
This work is preliminary to tackling https://github.com/habitat-sh/habitat/issues/5973.

Read commit messages for further details.


christophermaier:
@baumanj Making a non-zero Pid type is actually the next thing on my list, along with potentially creating a type to encapsulate the pid file logic. I wanted to do a bit more digging on the former to ensure it doesn't have unintended consequences, and am weighing the appropriateness of the latter (ultimately I'd like to do away with pid files altogether, but there might be safety and visibility benefits of a type in the interim).

I also thought I remembered types that couldn't be zero in Rust but for some reason wasn't able to find the various `NonZero*` types :man_facepalming:, so thanks for that pointer :joy:

Good suggestions on the logging and casting; I'll make those changes now.

PR #6999, opened 2019-10-07
Migrate to executors and windows 2019 across the board pt 1

scotthain:

(fyi there may not be a part 2)

### DIFF
--- a/.expeditor/scripts/release_habitat/build_mac_hab_binary.sh
+++ b/.expeditor/scripts/release_habitat/build_mac_hab_binary.sh
@@ -15,10 +15,6 @@ channel=$(get_release_channel)
 
 echo "--- Channel: $channel - bldr url: $HAB_BLDR_URL"
 
-echo "--- Installing buildkite agent"
-brew tap buildkite/buildkite
-brew install --token="$BUILDKITE_AGENT_ACCESS_TOKEN" buildkite-agent
-
 echo "--- Install openssl for the certs"
 brew install openssl
 export SSL_CERT_FILE=/usr/local/etc/openssl/cert.pem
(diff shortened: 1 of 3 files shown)


.expeditor/scripts/release_habitat/build_mac_hab_binary.sh
```diff
@@ -15,10 +15,6 @@ channel=$(get_release_channel)
 
 echo "--- Channel: $channel - bldr url: $HAB_BLDR_URL"
 
-echo "--- Installing buildkite agent"
-brew tap buildkite/buildkite
-brew install --token="$BUILDKITE_AGENT_ACCESS_TOKEN" buildkite-agent
```
christophermaier: Ah, crap... sorry I forgot about this when I did https://github.com/habitat-sh/habitat/commit/e09374beebf66cb0b36da396636846c16ff908cc.

Thanks for catching it!

.expeditor/verify.pipeline.yml
```diff
@@ -665,8 +519,10 @@ steps:
       HAB_STUDIO_SECRET_HAB_LICENSE: "accept-no-persist"
     command:
       - .expeditor/scripts/verify/build_component.sh backline
-    agents:
-      queue: 'default-privileged'
+    expeditor:
+      executor:
+        linux:
```
smacfarlane: Can these `build_component` verifications run in a docker executor?  Does it matter?

scotthain: Possibly? It wasn't before so I didn't want to rock the boat in this PR, it's big enough. If you want to test it in a followup, that would be :awesome:

PR #7027, opened 2019-10-09
end 2 end testing for supervisor on windows

mwrock:
resolves #7005, #7006, #7008 and #7009 

Signed-off-by: mwrock <matt@mattwrock.com>

Issue #6469, opened 2019-04-19
Health check intervals are serialized incorrectly in spec files


christophermaier:
Because `HealthCheckInterval` derives `Serialize` and `Deserialize`, and because that affects how this data is rendered into spec files, the internal structure of `HealthCheckInterval` is exposed to the outside world.

Specifically, this is what a `HealthCheckInterval` looks like in a spec file:

```toml
[health_check_interval]
secs = 5
nanos = 0
```
Ideally, the fact that it's a `Duration` under the hood should be irrelevant, and it should instead look like this:

```toml
health_check_interval = 5
```
(I discovered this trying to change the underlying implementation to use a `u64` instead of `Duration`.)

I'm becoming less and less fond of deriving `Serialize` and `Deserialize` like this, particularly when they make it easy to leak private internal details to the world.

At the very least, we should have a custom `Serialize` and `Deserialize` implementation that hides the underlying `Duration` (bonus points for switching to a `u64`). However, we will also need to migrate older spec files that have the `Duration`-based serialization, or we stand to break existing users.

christophermaier:
Still important.

PR #7049, opened 2019-10-16
fixes using a svc_user on windows

### LINKED ISSUE #6834: windows hab svc load with password flag errors out
In my plan.ps1 I set 
```$pkg_svc_user="domain\\user"```
I also tried the format
```$pkg_svc_user="user@domain"```

**Command causing the issue**: 
```
hab svc load --password "*****" domain/installedpackage
```
Error message
```
XXX
XXX [Err: 0] Service spec could not be rendered successfully: values must be emitted before tables
XXX
```

```Windows 10 Version 1709 OS Build 16299.1268```
```Hab version 0.83.0/20190712231714```
- [x] Debug/backtrace of the command you are trying to run. You can set the following environment variables before running the `hab` command to generate a trace:

```[2019-08-09T20:53:40Z DEBUG habitat_common::ui] UI { shell: Shell { input: InputStream { isatty: true }, out: OutputStream { coloring: Auto, isatty: true }, err: OutputStream { coloring: Auto, isatty: true } } }
[2019-08-09T20:53:40Z DEBUG hab] clap cli args: ["C:\\hab\\pkgs\\core\\hab-studio\\0.83.0\\20190712234514\\bin\\hab\\hab.exe", "svc", "load", "--password", "*******", "domain/package"]
[2019-08-09T20:53:40Z DEBUG hab] remaining cli args: []
[2019-08-09T20:53:40Z DEBUG hab::config] No CLI config found, loading defaults
[2019-08-09T20:53:40Z DEBUG tokio_reactor] adding I/O source: 0
[2019-08-09T20:53:40Z DEBUG tokio_reactor::background] starting background reactor
[2019-08-09T20:53:40Z DEBUG tokio_reactor::registration] scheduling Write for: 0
[2019-08-09T20:53:40Z DEBUG tokio_reactor::registration] scheduling Read for: 0
[2019-08-09T20:53:40Z DEBUG tokio_reactor::registration] scheduling Read for: 0
[2019-08-09T20:53:40Z DEBUG tokio_reactor] dropping I/O source: 0
XXX
XXX [Err: 0] Service spec could not be rendered successfully: values must be emitted before tables
XXX
```
		
- [x] Current Habitat environment variables where the `hab` command or supervisor is running.  These can be gathered using:

```
HAB_BLDR_CHANNEL               stable
HAB_BLDR_URL                   https://bldr.habitat.sh
HAB_CACHE_KEY_PATH             C:\hab\studios\aag_hab_devopsagent\hab\cache\keys
HAB_LICENSE                    accept-no-persist
HAB_STUDIO_ENTER_ROOT          C:\hab\studios\aag_hab_devopsagent
```
	- [x] If you're running into problems with the [Habitat Studio](https://www.habitat.sh/docs/concepts-studio/)above to include environment variables:
		- [ ] a. before starting the studio
		- [x] b. inside the studio

mwrock:
fixes #6834 

This addresses two key bugs around using a `svc_user` on windows.

* Fixes the serialization of a spec to toml by moving the health check interval to the last field.
* Ignores `svc_user` in the install hook and runs as the current user instead. This is because the install hook does not have access to the svc_user password.

Note there is an end to end test for this included in #7027

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -45,6 +45,7 @@ use crate::{api_client::{self,
                          pkg_install_path,
                          svc_hooks_path,
                          AtomicWriter},
+                    os::users,
                     package::{list::temp_package_directory,
                               Identifiable,
                               PackageArchive,
@@ -405,9 +406,16 @@ fn run_install_hook<T>(ui: &mut T, package: &PackageInstall) -> Result<()>
         ui.status(Status::Executing,
                   format!("install hook for '{}'", &package.ident(),))?;
         templating::compile_for_package_install(package)?;
-        if !hook.run(&package.ident().name,
-                     &Pkg::from_install(package)?,
-                     None::<&str>)
+        let mut pkg = Pkg::from_install(package)?;
+        // Only windows uses svc_password
+        if cfg!(target_os = "windows") {
+            // Install hooks do not have access to svc_passwords so
+            // we execute them under the current user account.
+            if let Some(user) = users::get_current_username() {
+                pkg.svc_user = user;
+            }
+        }
+        if !hook.run(&package.ident().name, &pkg, None::<&str>)
                 .unwrap_or(false)
         {
             return Err(Error::InstallHookFailed(package.ident().clone()));
--- a/components/sup/src/manager/service/spec.rs
+++ b/components/sup/src/manager/service/spec.rs
@@ -107,8 +107,16 @@ pub struct ServiceSpec {
     #[serde(with = "serde_string")]
     pub desired_state: DesiredState,
     pub shutdown_timeout: Option<ShutdownTimeout>,
-    pub health_check_interval: HealthCheckInterval,
     pub svc_encrypted_password: Option<String>,
+    // it is important that the health check interval
+    // is the last field to be serialized because it
+    // is serialized as a table. Individual values
+    // serialized after the health check interval will
+    // break the parser.
+    // Note that there is an issue to ultimately fix this:
+    // https://github.com/habitat-sh/habitat/issues/6469
+    // and eliminate the need to keep this field last.
+    pub health_check_interval: HealthCheckInterval,
 }
 
 impl ServiceSpec {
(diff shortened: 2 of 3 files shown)


components/common/src/command/package/install.rs
```diff
@@ -405,9 +406,13 @@ fn run_install_hook<T>(ui: &mut T, package: &PackageInstall) -> Result<()>
         ui.status(Status::Executing,
                   format!("install hook for '{}'", &package.ident(),))?;
         templating::compile_for_package_install(package)?;
-        if !hook.run(&package.ident().name,
-                     &Pkg::from_install(package)?,
-                     None::<&str>)
+        let mut pkg = Pkg::from_install(package)?;
+        // Install hooks do not have access to svc_passwords so
+        // we execute them under the current user account.
+        if let Some(user) = users::get_current_username() {
```
christophermaier: What would be the situation where the user doesn't have a username (besides some interesting container scenarios I can think of)?

What kind of feedback would we get if there was no username, but the `pkg.svc_user` was somehow incompatible with this install use case?

Since service passwords are only a thing on Windows, can we make this change Windows-only?

components/sup/src/manager/service/spec.rs
```diff
@@ -107,8 +107,13 @@ pub struct ServiceSpec {
...
     pub desired_state: DesiredState,
     pub shutdown_timeout: Option<ShutdownTimeout>,
-    pub health_check_interval: HealthCheckInterval,
     pub svc_encrypted_password: Option<String>,
+    // it is important that the health check interval
+    // is the last field to be serialized because it
+    // is serialized as a table. Individual values
+    // serialized after the health check interval will
+    // break the parser.
+    pub health_check_interval: HealthCheckInterval,
```
christophermaier: The real fix for this will be in https://github.com/habitat-sh/habitat/issues/6469... it would be good to add this link in a comment.

Actually fixing it here in this PR would require additional work to ensure backward compatibility, which may be out of scope.

components/common/src/command/package/install.rs
mwrock: I honestly cannot imagine there being no username. The only way this could happen is if someone removed the USERNAME environment variable which could wreak all sorts of havoc throughout their system. I'm just trying to avoid the unwrap/expect. There is another issue that will use an API instead of the environment variable that will avoid the environment variable alltogether.

Issue #7044, opened 2019-10-15
hab-sup and hab-launcher deadlock because of unsafe signal handling


stevendanna:
# The Problem

We have observed cases where hab-sup and/or hab-launcher becomes non-responsive to user reqeusts, stops forwarding service logs (in the case of hab-launcher), and fails to reap child processes.

# Suspected Causes

hab-sup and hab-launcher share signal handling code.  That code has at least 2 sources of potential deadlocks or state corruption.  The deadlock can eventually grow to include a large number of hab-sup or hab-launcher threads, leading to a variety of unexpected behavior, including non-responsiveness from either service.

We set the following signal handler:

https://github.com/habitat-sh/habitat/blob/41438356ec7e18811b5314780d1247f0a5a859bb/components/core/src/os/signals/unix.rs#L24-L26

This signal handler has 2 problems:

1) Taking a lock inside a signal handler is inherently unsafe.  There are two ways this could lead to a deadlock:

(a) Thread 1 has signal handler invoked for signal A. Thread 1 takes the lock. Thread 1 has signal handler invoked for signal B. Thread 1 attempts to take lock but is now blocked forever.

(b) Thread 1 calls check_for_signal and acquires lock. While in check_for_signal, Thread 1 has signal handler invoked for signal A. Thread 1 attempts to take lock but is now blocked forever.  We believe this happened recently in hab-launcher:

```
Thread 1 (process 31282):
#0  0x00007f2dc5f63c8c in ?? () from /hab/pkgs/core/glibc/2.22/20170513201042/lib/libpthread.so.0
#1  0x00007f2dc5f5dac5 in pthread_mutex_lock () from /hab/pkgs/core/glibc/2.22/20170513201042/lib/libpthread.so.0
#2  0x000055cc42b0c2bf in lock (self=0x7f2dc461f0f0) at libstd/sys/unix/mutex.rs:66
#3  raw_lock (self=0x7f2dc461f0f0) at libstd/sys_common/mutex.rs:46
#4  lock<alloc::collections::vec_deque::VecDeque<i32>> (self=0x55cc42ee64b8 <_$LT$habitat_core..os..signals..unix..CAUGHT_SIGNALS$u20$as$u20$core..ops..deref..Deref$GT$::deref::__stability::LAZY::ha3d3ee8377affb81>)
    at libstd/sync/mutex.rs:230
[...]

PR #7050, opened 2019-10-16
core: use synchronous signal handling on unix

stevendanna:
## Summary

Issue #7044 describes multiple problems with the current signal
handling approach. These problems were introduced in 8e138274be and
render our current signal handling unsafe. Specifically, taking a
mutex and allocating memory from inside an signal handler can result
in a deadlock or state corruption.

When using asynchronous signal handling via signal handlers, the code
in the signal handler are restricted to limited number of safe
operations (i.e. flipping an atomic, calling async-signal-safe
functions).

This change hopes to solve the existing issues and avoid future
problems caused by signal handlers by moving us to synchronous signal
handling.

## Move to Synchronous signal handling

The strategy used here is described in _The Linux
Programming Interface_[0]:

> - All threads block all of the asynchronous signals that the process
>   might receive. The simplest way to do this is to block the signals
>   in the main thread before any other threads are created. Each
>   subsequently created thread will inherit a copy of the main
>   thread's signal mask.
>
> - Create a single dedicated thread that accepts incoming signals
>   using _sigwaitinfo()_, _sigtimedwait()_, or _sigwait()_...

On Linux, signals::init() will now:

- Block all signals that we plan on handling
- Starts a thread that processes signals synchronously.

Because we want to block the signals early in the process, we move
signals::init() much earlier into main.

## Move to a more explicit interface

This change also moves us from using a VecDequeue under a Mutex for
tracking signals to signal-specific AtomicBools, replacing the generic
`check_for_signal()` function with two more-specific
`pending_sighup()` and `pending_sigchld()` functions.

This has two advantages:

1) No mutexes or memory allocations in our signal handling code. This
   matters less since we have also moved to synchronous handling of
   signals, bug gives us some flexibility to change our approach more
   safely in the future.

2) Clearer alignment between this interface and the application
   behavior.

The old check_for_signal() API looks like it intended to provide the
groundwork for a variety of signal-based behavior; however, in
reality, we are handling a small number of signals. The generic nature
of that interface and the fact that the code is shared between two
processes was obscuring the fact that we actually have a small amount
of signal-related behavior across the two services:

*Linux Behavior*
```
| signal    | hab-launch behavior | hab-sup behavior     |
|-----------+---------------------+----------------------|
| SIGINT    | Graceful shutdown   | Handled but ignored  |
| SIGTERM   | Graceful shutdown   | Handled but ignored  |
| SIGHUP    | Send to hab-sup     | Shutdown for restart |
| SIGCHLD   | Reap zombies        | Handled but ignored  |
| SIGQUIT   | Handled but ignored | Handled but ignored  |
| SIGALRM   | Handled but ignored | Handled but ignored  |
| SIGUSR1   | Handled but ignored | Handled but ignored  |
| SIGUSR2   | Handled but ignored | Handled but ignored  |
| All other | Default disposition | Default disposition  |
```

*Windows Behavior*

Windows uses a library which appears to handle the CTRL_C_EVENT. 

The set of changes should not substantively change the behavior on either platform, 
although we are now installing the handler earlier in the startup of the relevant applications.

This more explicit interface has the disadvantage of not being
well-suited to a larger variety of signal-based behavior. However, it
seems to me that we can always implement something smarter when that
actually becomes an issue.

## Resources

[0] Kerrisk, Michael. _The Linux Programming Interface_, 685. San
Franscisco: No Starch Press, 2010.
[1] http://man7.org/linux/man-pages/man2/sigprocmask.2.html
[2] http://man7.org/linux/man-pages/man3/sigwait.3.html
[3] http://man7.org/linux/man-pages/man3/pthread_sigmask.3.html
[4] http://man7.org/linux/man-pages/man3/sigsetops.3.html
[5] http://man7.org/linux/man-pages/man7/signal.7.html

### DIFF
--- a/components/core/src/os/signals/unix.rs
+++ b/components/core/src/os/signals/unix.rs
@@ -3,9 +3,13 @@
 use crate::os::process::{Signal,
                          SignalCode};
 use std::{collections::VecDeque,
+          mem,
+          ptr,
+          io,
           sync::{atomic::Ordering,
                  Mutex,
-                 Once}};
+                 Once},
+          thread};
 
 static INIT: Once = Once::new();
 
@@ -13,26 +17,12 @@ lazy_static::lazy_static! {
     static ref CAUGHT_SIGNALS: Mutex<VecDeque<SignalCode>> = Mutex::new(VecDeque::new());
 }
 
-// Functions from POSIX libc.
-extern "C" {
-    fn signal(sig: SignalCode,
-              cb: unsafe extern "C" fn(SignalCode))
-              -> unsafe extern "C" fn(SignalCode);
-}
-
-unsafe extern "C" fn handle_signal(signal: SignalCode) {
-    CAUGHT_SIGNALS.lock()
-                  .expect("Signal mutex poisoned")
-                  .push_back(signal);
-}
-
-unsafe extern "C" fn handle_shutdown_signal(_signal: SignalCode) {
-    super::SHUTDOWN.store(true, Ordering::SeqCst);
-}
-
 pub fn init() {
     INIT.call_once(|| {
-            self::set_signal_handlers();
+            // TODO(ssd) 2019-10-16: We could bubble this error up
+            // further if we want, but in either case this should be a
+            // hard failure.
+            self::start_signal_handler().expect("starting signal handler failed");
         });
 }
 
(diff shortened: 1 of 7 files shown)


stevendanna:
Note that I kept the `check_for_signal` function; however, I'm kinda skeptical of that function as well because the process that calls it only dequeues one signal at a time and only calls it every tick which means we can get pretty overwhelmed by signals.

components/sup/src/main.rs
```diff
@@ -82,6 +83,17 @@ fn main() {
     logger::init();
     let mut ui = UI::default_with_env();
     let flags = FeatureFlag::from_env(&mut ui);
+    // On Windows initializng the signal handler will create a ctrl+c handler for the
+    // process which will disable default windows ctrl+c behavior and allow us to
+    // handle via check_for_signal. However, if the supervsor is in a long running
+    // non-run hook, the below loop will not get to check_for_signal in a reasonable
+    // amount of time and the supervisor will not respond to ctrl+c. On Windows, we
```
markan: initializng -> initializing supervsor -> supervisor

stevendanna:
Test failures is:

```
[2019-10-16T17:15:03Z ERROR habitat_butterfly::server::inbound] UDP Receive error: Either the application has not called WSAStartup, or WSAStartup failed. (os error 10093)
```

stevendanna:
TODO:

- [ ] Ensure everyone who forks/exec's is using std::os.  Our signal mask is shared inherited across those calls.  The stdlib is careful to reset it before launching a subprocess, but anyone using this process and calling libc directly might be surprised.

components/core/src/os/signals/unix.rs
```diff
@@ -62,20 +52,122 @@ pub fn check_for_signal() -> Option<SignalEvent> {
...
+                          })?;
+    Ok(())
+}
+
+// Sigset is a wrapper for the underlying libc type.
+struct Sigset {
+    inner: libc::sigset_t
+}
+
+impl Sigset {
```
markan: Is it worth looking at the nix crate (https://github.com/nix-rust/nix/blob/01f4d57eede3ed2f1051778113b1d6a7dff8e9d6/src/sys/signal.rs)
Looks like we already pull it in indirectly, but not directly.

stevendanna:
@markan > but I'm still refreshing my memory on signal handling behavior in threads

If you have TLPI, checkout chapter 33.2.

components/core/src/os/signals/unix.rs
stevendanna: Happy to change it over if y'all prefer.  It seems the main difference in implementation (aside from them creating more types given their larger API surface area) is they use `mem::MaybeUninit::uninit()` whereas I use `mem::zeroed()` and `libc::c_int`.  Based on reading this: 

https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html

It seems like that is a worthwhile change even if we don't move to the library.

stevendanna: Although the documentation here makes me think that I've used mem::zeroed() as intended:

https://doc.rust-lang.org/std/mem/fn.zeroed.html

markan: I looked at the mem::zeroed() and mem::MaybeUninit::uninit() docs, and think that your mem::zeroed usage is correct.

markan:
stevendanna > Note that I kept the `check_for_signal` function; however, I'm kinda skeptical of that function as well because the process that calls it only dequeues one signal at a time and only calls it every tick which means we can get pretty overwhelmed by signals.

I was wondering about that as well. Is there any case where we need to resolve multiple instances of a signal, or the order of their arrival? If not could we simply have a static array of flags, one per signal type, and test/process them in some priority order.  While I think it's safe now, getting rid of the alloc implicit in push_back and replacing it with updating a static mask would simplify things and make it easier to think about the code.

stevendanna:
> Is there any case where we need to resolve multiple instances of a signal, or the order of their arrival? 

I don't think so.  Also, based on my understanding of how signals work, anything that _is_ depending on this would be broken by design since the kernel will "coalesce" signals on you.

components/sup/src/main.rs
stevendanna: I've fixed these typos but also tried to clarify this comment which I found a bit confusing because it didn't really reflect reality.

components/launcher/src/server.rs
```diff
@@ -271,12 +271,11 @@ impl Server {
                     return result;
                 }
             }
-            Some(SignalEvent::Passthrough(signal)) => {
-                self.forward_signal(signal);
+
+            if signals::pending_sighup() {
```
stevendanna: _Tecnically_ this is a small behavior change for the launcher.  Whereas before we would only handle 1 signal on each tick, now we potentially handle two.  If we are concerned we could move this to an `if else`.

stevendanna:
@markan Thanks for the review!  I've moved away from the VecDequeue and just added a couple more AtomicBool's.  I thought about doing some bitmask type thing, but for now, I wanted the signal behavior we actually handle to be much more explicit in the code.  I don't think a bitmask+mutex and a new API would be hard to wire up in the future if we needed it. Ultimately, given the interactions between process, threads, signals, and the complicated 3rd party libraries we use which launch their own threads, I'd rather keep our use of signal-based behavior to a minimum.

On your question of libraries.  I defer to y'all who work on this every day.  Since a huge function of launcher and sup is subprocess management, I don't think we ever escape from having to own the details, but there is still a trade-off. In this PR, I've deferred to stdlib to do all necessary cleanup before calling libexec but taken the signal masking operations in-house. To me the big question around whether to use nix is mostly about whether we want to deal with managing the FFI bits. In the very long run, there is also a good looking library `signal-hook` (that we already have as a transitive dependency) that might be worth looking at as well but I think using it would require more substantive changes to the main loops so I've deferred looking into that for now.

components/core/src/os/signals/unix.rs
```diff
@@ -1,87 +1,149 @@
...
 }
 
-// Functions from POSIX libc.
-extern "C" {
-    fn signal(sig: SignalCode,
-              cb: unsafe extern "C" fn(SignalCode))
-              -> unsafe extern "C" fn(SignalCode);
-}
+pub fn pending_sighup() -> bool { PENDING_HUP.compare_and_swap(true, false, Ordering::SeqCst) }
+pub fn pending_sigchld() -> bool { PENDING_CHLD.compare_and_swap(true, false, Ordering::SeqCst) }
```
stevendanna: Reviewer Note:  I've gone with `SeqCst` here because that is what we did for SHUTDOWN and it leaves no room for doubt.  But I think we may be able to use Relaxed here.

jaym:
Something to discuss: I don't believe we need anything more than relaxed for our atomic memory ordering because the 2 threads are not sharing any other writes/reads. The assumption I'm making here is that the actions taken as a result of the atomic bool are not reordered to happen before the CAS.

markan:
This is directed to the more experienced habitat developers:

So when I look at the history of this code, it looks like circa hab 0.50 we handled signals in a similar fashion to this PR. Then in https://github.com/habitat-sh/core/pull/75 we changed that to a queue, because of concerns that multiple signals might be lost or handled out of order.

But (as @stevendanna mentioned) multiple signals to a process can be freely reordered and coalesced, in particular any sent before the receiver is scheduled. It looks like under the covers it's a set of bit flags per process, indicating whether a signal has been sent to it, and when scheduled the signals are delivered in a canonical order https://github.com/torvalds/linux/blob/master/kernel/signal.c#L217. So we can't actually guarantee the behavior that #75 above is intended to provide, especially under load or other duress. 

So any behavior that depends on in-order or multiple signal delivery is broken and needs to be modified. But from https://github.com/habitat-sh/core/issues/11, I'm not clear on what depends on that behavior. How do we find out?

baumanj:
@stevendanna: excellent work, as always. I was indeed mistaken in using locking and allocating functions inside the signal handler 🤦‍♂️.

To that end, I'd suggest at least considering using [the `signal-hook` crate](https://github.com/vorner/signal-hook), which looks like it was written with a lot of those subtleties in mind.

christophermaier:
@stevendanna Was there anything else you wanted to do with this PR? If not, I'm happy to get it merged; I just didn't want to pull the rug out from under you if you had any other tweaks you wanted to do. :smile: 

I'm going to file an issue to look into `signal-hook` as a further evolution of this work... it's been on my radar for a little while, since signal handling has always been a bit more obfuscated and complicated in our codebase than I'd like. I agree that it would likely require more extensive refactoring to our code, though.

I think what you've done here in this PR is a very definite improvement, though!

stevendanna:
@christophermaier No other urgent changes on my end. I think some of these suggestions can be done as more focused follow-ups.

christophermaier:
Here's that signal-hook issue, for posterity: https://github.com/habitat-sh/habitat/issues/7076

PR #7072, opened 2019-10-21
Install Rust using the "minimal" profile

christophermaier:
This should allow us to avoid doing unnecessary work in CI.

For further details, see
https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles.

### DIFF
--- a/.expeditor/scripts/verify/shared.sh
+++ b/.expeditor/scripts/verify/shared.sh
@@ -29,7 +29,7 @@ install_rustup() {
     echo "--- :rust: rustup is currently installed."
   else
     echo "--- :rust: Installing rustup."
-    curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path -y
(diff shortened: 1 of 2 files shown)


.expeditor/scripts/verify/shared.sh
```diff
@@ -29,7 +29,7 @@ install_rustup() {
     echo "--- :rust: rustup is currently installed."
   else
     echo "--- :rust: Installing rustup."
-    curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path -y
+    curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path -y --profile=minimal
     # shellcheck disable=SC1090
```
mwrock: could you add this same arg to https://github.com/habitat-sh/habitat/blob/master/.expeditor/scripts/shared.ps1#L57

christophermaier: Yessir!

PR #7073, opened 2019-10-22
add windows system paths to environment PATH variable if included

### LINKED ISSUE #6657: In Windows, hab pkg exec should add to PATH instead of replace it

mwrock:
fixes #6657 

On Windows these added paths are set by default on all modern windows systems. Any Windows application expects these paths to be present.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -274,6 +274,23 @@ pub fn user_config_path<T: AsRef<Path>>(service_name: T) -> PathBuf {
     user_path(service_name).join("config")
 }
 
+/// This produces a list of the default windows system paths
+/// that are set by default on any modern version of windows.
+/// This also arranges them in the same order as they would be by
+/// default. The ordering is likely not super important but we
+/// might as well match how windows sets them just in case.
+pub fn windows_system_paths() -> Vec<PathBuf> {
+    let mut paths = Vec::new();
+    if let Some(sys_root) = env::var_os("SystemRoot") {
+        let system32 = Path::new(&sys_root).join("system32");
+        paths.push(system32.clone());
+        paths.push(PathBuf::from(sys_root));
+        paths.push(system32.join("wbem"));
+        paths.push(system32.join("WindowsPowerShell").join("v1.0"));
+    }
+    paths
+}
+
 /// Represents the service directory for a given package.
 pub struct SvcDir<'a> {
     service_name: &'a str,
--- a/components/core/src/package/install.rs
+++ b/components/core/src/package/install.rs
@@ -236,7 +236,17 @@ impl PackageInstall {
 
         // Only insert a PATH entry if the resulting path string is non-empty
         if !rooted_path.is_empty() {
-            env.insert(PATH_KEY.to_string(), rooted_path);
+            if cfg!(windows) {
+                // On Windows we want to make sure to append the system paths.
+                // While this is not critical for every application, any windows
+                // app may expect these to be on the path to access system utilities
+                // and libraries and could possibly fail if they are absent.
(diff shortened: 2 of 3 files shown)


components/core/src/package/install.rs
```diff
@@ -236,7 +236,14 @@ impl PackageInstall {
 
         // Only insert a PATH entry if the resulting path string is non-empty
         if !rooted_path.is_empty() {
-            env.insert(PATH_KEY.to_string(), rooted_path);
+            if cfg!(windows) {
+                let mut paths: Vec<PathBuf> = env::split_paths(&rooted_path).collect();
```
christophermaier: Strictly speaking, you don't have to call `collect` here, since `env::join_paths` can operate on an iterator.

You could get that iterator, `chain` on your Windows system paths, and then pass that to `env::join_paths`.

The `to_string_lossy().to_string()` bit makes me wonder if we might be typing our env map incorrectly, but perhaps that is a question for another day :thinking: 

Granted, this isn't going to be called in a tight loop, or with tons of input data, so extra allocations aren't something to be concerned about, but it could lead to smaller, easier-to-reason-about code.

components/core/src/package/install.rs
```diff
@@ -236,7 +236,14 @@ impl PackageInstall {
 
         // Only insert a PATH entry if the resulting path string is non-empty
         if !rooted_path.is_empty() {
-            env.insert(PATH_KEY.to_string(), rooted_path);
+            if cfg!(windows) {
```
christophermaier: Again, some documentation about why this is important would be very helpful, particularly given all the subtle discussion in #6657. It can be incredibly difficult to understand the reasons for some of these changes when coming to the code a year later if those reasons aren't clearly documented.

components/core/src/package/install.rs
mwrock: yeah it seems like `BTreeMap<String, OsString>` would be better

christophermaier: @mwrock Just to be clear, I think that change (if indeed it is justified) would be something for another PR... pretty sure that'd have a decently sized ripple effect.

components/core/src/fs.rs
```diff
@@ -274,6 +274,24 @@ pub fn user_config_path<T: AsRef<Path>>(service_name: T) -> PathBuf {
...
+pub fn windows_system_paths() -> Vec<PathBuf> {
+    let mut paths = Vec::new();
+    if let Some(sys_root) = env::var_os("SystemRoot") {
+        let str_sys_root = sys_root.to_string_lossy().to_string();
+        let system32 = Path::new(&sys_root).join("system32");
+        paths.push(system32.clone());
+        paths.push(PathBuf::from(str_sys_root));
+        paths.push(system32.join("wbem"));
+        paths.push(system32.join("WindowsPowerShell").join("v1.0"));
+    }
```
christophermaier: Because `Path` and `PathBuf` can deal with `OsStr`s, you can simplify this a bit by getting rid of `str_sys_root`:

```rust
    if let Some(sys_root) = env::var_os("SystemRoot") {
        let system32 = Path::new(&sys_root).join("system32");
        paths.push(system32.clone());
        paths.push(PathBuf::from(sys_root));
        paths.push(system32.join("wbem"));
        paths.push(system32.join("WindowsPowerShell").join("v1.0"));
    }
```

PR #7097, opened 2019-10-23
[CI] Temporarily comment out most of the release pipeline

christophermaier:
Cycle time testing this is a killer, but we can start focusing in on
the things that aren't 100% working yet. Once they're good, we can
reinstate the whole thing.


christophermaier:
Closing in favor of #7098

PR #7099, opened 2019-10-23
stop windows service if launcher exits

### LINKED ISSUE #6859: Habitat Windows Service does not monitor the health of the supervisor
When installing Habitat as a Windows Service using `core/windows-service`, the Habitat service does not seem to monitor the hab-sup process or hab-launcher process for failures and attempt to restart them.

This is a problem when there are intermittent issues that can affect startup (like small network outages), or when the supervisor prematurely exits for some reason.

I think it would be nice to include an auto-restart function something similar to systemd `Restart=always`.

## Scenario 1
The hab-launcher.exe processes exit in error and the windows service does not restart it. It also reports a status of running.

### Steps to Reproduce
1. hab pkg exec core/windows-service install
2. Start the Habitat service
3. Kill the hab-launcher.exe process...something like `Stop-Process -Name hab-launcher`
4. The supervisor is not restarted and the service reports running.

## Scenario 2
The habitat supervisor fails to start when the Windows Service is started.

### Steps to Reproduce
1. hab pkg exec core/windows-service install
2. Provide bad launcher arguments. For example:

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <appSettings>
    <add key="debug" value="false" />
    <add key="ENV_HAB_FEAT_IGNORE_SIGNALS" value="true" />
    <add key="launcherArgs" value="--derp yo />
  </appSettings>
</configuration>
```
3. Start the Habitat Service
4. Type `Get-Service Habitat` and see the status shows as running even though the supervisor failed to start.

mwrock:
fixes #6859 

This ensures that if the launcher process exits for any reason, the windows service shuts itself down. It will not end up in a state where the service is in a running state but the supervisor is not running.

This also adds an `install` hook to the windows-service package to eliminate the need for the extra call to `hab pkg exec core/windows-service install`.


.expeditor/end_to_end.pipeline.yml
```diff
@@ -256,6 +256,17 @@ steps:
             - BUILD_PKG_TARGET=x86_64-linux
             - HAB_BLDR_URL=https://bldr.acceptance.habitat.sh
 
+  - label: "[:windows: test_windows_service_stops_on_launcher_termination]"
+    command:
+      - powershell .expeditor/scripts/end_to_end/run_e2e_test.ps1 DEV test_windows_service_stops_on_launcher_termination
```
christophermaier: You'll want to use "dev" instead of "DEV" from here on out; see https://github.com/habitat-sh/habitat/pull/7100 for more.

components/windows-service/hooks/install
```diff
@@ -0,0 +1,7 @@
+. {{pkg.path}}\bin\habitat.ps1
+
+if((Get-Service Habitat -ErrorAction SilentlyContinue) -ne $null) {
+    UnInstall-HabService
+}
```
christophermaier: If the service already exists, we uninstall the old and install the new, correct?

components/windows-service/plan.ps1
```diff
@@ -1,6 +1,6 @@
 $pkg_name="windows-service"
 $pkg_origin="core"
-$pkg_version="0.5.0"
+$pkg_version="0.6.0"
```
christophermaier: Out of curiosity, do we have any kind of meaningful versioning scheme set up for this particular package yet?

test/end-to-end/test_windows_service_stops_on_launcher_termination.ps1
```diff
@@ -0,0 +1,24 @@
...
+    Start-Sleep -Seconds 2
+
+    It "Stops service" {
+        (Get-Service Habitat).Status | Should -Be "Stopped"
+    }
+}
+
+Describe "Bogus Launcher args" {
+    Stop-Service Habitat
+    (Get-Content c:/hab/svc/windows-service/HabService.dll.config).replace('--no-color', '--poop') |
```
christophermaier: :hankey:

components/windows-service/hooks/install
mwrock: that is right.

components/windows-service/plan.ps1
mwrock: no we don't. Now that we are moving to releasing one with every release, perhaps we should just user the `VERSION` file and pull it in to the overall habitat versioning.

.expeditor/end_to_end.pipeline.yml
mwrock: oh ok. I am also making the replacement in a few files that got missed.

components/windows-service/plan.ps1
mwrock: thinking more on this, I vote to keep this as is for now. If we moved to the larger hab versioning that carries with it the assumption that a 0.90.0 windows-service would be tied to a 0.90.0 supervisor which is not true.

PR #7104, opened 2019-10-24
wire up e2e hab pkg download tests into expeditor

jeremymv2:
This PR accomplishes the following:

 * wires up e2e tests for hab pkg download
 * covers bugs/features recently merged for the new download command
 * finally, adds some minor cleanups and enhancements to the `test/end-to-end/test_pkg_download.sh` script

### DIFF
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -372,7 +372,7 @@ steps:
 
   - label: "[:linux: test_pkg_download]"
     command:
-      - .expeditor/scripts/end_to_end/setup_environment.sh DEV
+      - .expeditor/scripts/end_to_end/setup_environment.sh dev
       - test/end-to-end/test_pkg_download.sh
     expeditor:
       executor:
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -370,6 +370,17 @@ steps:
             - BUILD_PKG_TARGET=x86_64-windows
             - BUILDKITE_AGENT_ACCESS_TOKEN
 
+  - label: "[:linux: test_pkg_download]"
+    command:
+      - .expeditor/scripts/end_to_end/setup_environment.sh DEV
+      - test/end-to-end/test_pkg_download.sh
+    expeditor:
+      executor:
+        docker:
+          privileged: true
+          environment:
+            - ACCEPTANCE_HAB_AUTH_TOKEN
+
   - wait
 
   - label: "[:habicat: Promote to Acceptance]"
--- a/test/end-to-end/test_pkg_download.sh
+++ b/test/end-to-end/test_pkg_download.sh
@@ -25,7 +25,7 @@ IDENT_FILE="ident_file"
 HAB_AUTH_TOKEN=${ACCEPTANCE_HAB_AUTH_TOKEN:-${HAB_AUTH_TOKEN}}
 
 echo
-echo "==========Testing command ${HAB}, using cache dir ${CACHE_DIR}"
+echo "==========Testing with command ${HAB}, using cache dir ${CACHE_DIR}"
 echo
 
 before_test() {
--- a/test/end-to-end/test_pkg_download.sh
+++ b/test/end-to-end/test_pkg_download.sh
@@ -10,7 +10,7 @@
 # command line testing.
 #
 # Assumptions:
-# 1. ACCEPTANCE_HAB_AUTH_TOKEN or HAB_AUTH_TOKEN Environment variables are set and valid
+# 1. HAB_AUTH_TOKEN Environment variables is set and valid
 # 2. ${CACHE_DIR} can be set to a writable location on the filesystem
 
 set -euo pipefail
@@ -22,7 +22,8 @@ set -euo pipefail
 HAB=${HAB_TEST_CMD:-hab}
 CACHE_DIR="test-cache"
 IDENT_FILE="ident_file"
-HAB_AUTH_TOKEN=${ACCEPTANCE_HAB_AUTH_TOKEN:-${HAB_AUTH_TOKEN}}
+# we explicitly define --channel where needed in the tests
+export HAB_BLDR_CHANNEL=
 
 echo
 echo "--- Testing with command ${HAB}, using cache dir ${CACHE_DIR}"
@@ -102,7 +103,7 @@ success_from_command_line() {
 
     echo "--- Testing command line idents"
 
-    CMD="$HAB pkg download --download-directory=${CACHE_DIR} core/gzip"
+    CMD="$HAB pkg download --channel stable --download-directory=${CACHE_DIR} core/gzip"
     echo "Testing command line: ${CMD}"
     ${CMD}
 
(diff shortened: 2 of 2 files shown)


.expeditor/end_to_end.pipeline.yml
```diff
@@ -278,6 +278,17 @@ steps:
             - BUILD_PKG_TARGET=x86_64-windows
             - BUILDKITE_AGENT_ACCESS_TOKEN
 
+  - label: "[:linux: test_pkg_download]"
```
markan: Eyeballing the expeditor test timings, a great deal of time is setting up the environment compared to the actual test execution time. As we add more of these tests this might not scale super well; we might want to just add a batch script to run all / most of the tests in the end_to_end directory in a single test env.

mwrock: what I have noticed is the bulk of time can be spent in `docker pull` but sometimes the image is already there. I don't understand why sometimes it is there and sometimes not but ensuring the docker images are on the test VMs could significantly improve performance.

christophermaier: Just to be clear, this whole "parallelization via Buildkite" approach was always intended at a near-term expedient. Eventually, we need to set up some kind of actual test runner code infrastructure to manage these tests.

.expeditor/end_to_end.pipeline.yml
```diff
@@ -278,6 +278,17 @@ steps:
...
+    command:
+      - .expeditor/scripts/end_to_end/setup_environment.sh DEV
```
christophermaier: Need to use "dev" rather than "DEV"; see #7100 for details

jeremymv2: Fixed up in https://github.com/habitat-sh/habitat/pull/7104/commits/2da2907b008d5d7bcfae15e15cf6504fbbb5d734

test/end-to-end/test_pkg_download.sh
```diff
@@ -7,21 +7,25 @@
...
 
 # Test the Habitat package downloader.
 #
 # Uses the `HAB_INTERNAL_BLDR_CHANNEL` environment variable to control
 # the base packages channel for the exporter.
-HAB=$HAB_TEST_CMD || "hab"
+HAB=${HAB_TEST_CMD:-hab}
 CACHE_DIR="test-cache"
 IDENT_FILE="ident_file"
+HAB_AUTH_TOKEN=${ACCEPTANCE_HAB_AUTH_TOKEN:-${HAB_AUTH_TOKEN}}
```
christophermaier: This doesn't appear to be needed (and also seems to fail if the fallback `HAB_AUTH_TOKEN` isn't present)

test/end-to-end/test_pkg_download.sh
```diff
@@ -108,24 +110,41 @@ success_from_command_line() {
...
-    
     before_test
 
-    echo "==========Testing file idents"
-    
+    echo "--- Testing file idents"
+
     echo "core/gzip" > ${IDENT_FILE}
-    CMD="$HAB pkg download --download-directory=${CACHE_DIR} --file=${IDENT_FILE}"   
+    CMD="$HAB pkg download --download-directory=${CACHE_DIR} --file=${IDENT_FILE}"
```
christophermaier: When running these locally with the [e2e_local.sh](https://github.com/habitat-sh/habitat/blob/559d8dcef2bfd2ac85c1aca464adaeecd0599e6d/e2e_local.sh) script, I get failures because of the `HAB_BLDR_CHANNEL` environment being set to `dev`, because all the packages that are being downloaded aren't present in that channel.

Long term, I'm not sure what the best pattern for this should be, since we still want to ensure that all the packages being built in this pipeline are properly resolved. Adding `--channel=stable` to these download commands would do the trick, as would `unset HAB_BLDR_CHANNEL` before executing the test functions.

test/end-to-end/test_pkg_download.sh
jeremymv2: I can remove that line. From what I can see, this will mean `HAB_AUTH_TOKEN` must be set somehow in the pipeline. 

I was thinking it would be made available from `ACCEPTANCE_HAB_AUTH_TOKEN` via expeditor here: https://github.com/habitat-sh/habitat/pull/7104/files#diff-9f269bb242eb760dee635951a1a1226eR382

test/end-to-end/test_pkg_download.sh
jeremymv2: Ahh, I wasn't aware of that script. Thank you!

I've added the explicit `--channel stable` [here](https://github.com/habitat-sh/habitat/pull/7104/commits/6c8b27a8f8dcb2c3b80b31f1ee7dbe19ce898d25) and `unset HAB_BLDR_CHANNEL` [here](https://github.com/habitat-sh/habitat/pull/7104/commits/12273bbfb71209796afeb1935a06ca7e845d6819)

test/end-to-end/test_pkg_download.sh
jeremymv2: I agree, I don't see this as necessary. I've removed it in this [commit](https://github.com/habitat-sh/habitat/pull/7104/commits/6c8b27a8f8dcb2c3b80b31f1ee7dbe19ce898d25)

test/end-to-end/test_pkg_download.sh
```diff
@@ -7,21 +7,26 @@
 # for cleaner code and more fine grained tests. However, they are a
 # bit of a pain to program in bash. This is intended to provide
 # minimal testing pending our figuring out the best approach for
-# command line testing. 
+# command line testing.
 #
+# Assumptions:
+# 1. HAB_AUTH_TOKEN Environment variables is set and valid
```
christophermaier: I don't even think you need a token at all, since all this is downloading stuff (though downloading _private_ packages is perhaps another worthwhile usecase to test in the future!)

test/end-to-end/test_pkg_download.sh
```diff
@@ -7,21 +7,26 @@
...
 # Test the Habitat package downloader.
 #
 # Uses the `HAB_INTERNAL_BLDR_CHANNEL` environment variable to control
 # the base packages channel for the exporter.
-HAB=$HAB_TEST_CMD || "hab"
+HAB=${HAB_TEST_CMD:-hab}
 CACHE_DIR="test-cache"
 IDENT_FILE="ident_file"
+# we explicitly define --channel where needed in the tests
+unset HAB_BLDR_CHANNEL
```
christophermaier: Also just to be clear, I think you only need this, and not also the `--channel=stable` options. Having both doesn't hurt, of course, but it's a bit of a "belt and suspenders" kind of thing :smile:

test/end-to-end/test_pkg_download.sh
jeremymv2: Agreed. Addressed here https://github.com/habitat-sh/habitat/pull/7104/commits/f6d3c4bb03a461c5282e7607e61682d1cbedbb1d

test/end-to-end/test_pkg_download.sh
jeremymv2: Right, it does seem a little over zealous. I've removed the `unset` preferring the explicit `--channel stable` argument here https://github.com/habitat-sh/habitat/pull/7104/commits/f6d3c4bb03a461c5282e7607e61682d1cbedbb1d

I'll let the build go green then merge since it has two approvals. Thank you!

PR #7108, opened 2019-10-24
Fully qualified package names download irrespective of channel

markan:
Fix for hab 7039

When hab pkg download is provided a fully-qualified identifier, it
will fail if that package isn't in the stable channel. In other words
we would need to provide a channel that contains the package on the
command line.  That's a departure from other CLI commands since a
fully-qualified identifier is enough to find a package regardless of
what channel it might be in.

This adds a function show_package_metadata_for_fully_qualified_ident
to the builder-api-client. In the download command implementation test
for a fully qualified ident and use the appropriate call.

An alternative implementation would be to modify the
show_package_metadata function to ignore the channel and use a
different rest endpoint if the package was fully qualified, but that
might create opaque errors and regressions.

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -796,7 +796,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
         let mut encoded = String::new();
         resp.read_to_string(&mut encoded)
             .map_err(Error::BadResponseBody)?;
-        trace!(target: "habitat_http_client::api_client::show_package", "{:?}", encoded);
+        trace!(target: "habitat_http_client::api_client::show_package_metadata", "{:?}", encoded);
 
         let package: Package = serde_json::from_str::<Package>(&encoded)?;
         Ok(package)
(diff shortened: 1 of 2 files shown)


chefsalim:
@markan I think this may be overkill. We should just be able to check for a fully qualified ident in determine_latest_from_ident and use the unstable channel when calling fetch_latest_package_in_channel_for.  That should be only a few lines of change to handle this case.

markan:
> We should just be able to check for a fully qualified ident in determine_latest_from_ident and use the unstable channel when calling fetch_latest_package_in_channel_for. 
I wasn't sure if that was a guaranteed property or not; I take it *every* package is in unstable?

chefsalim:
Yep, every package is guaranteed to be in stable

chefsalim:
Here's a sample snippet of how we're using that pattern to grab info on a fully qualified ident - https://github.com/habitat-sh/habitat/blob/master/components/builder-api-client/src/builder.rs#L734

PR #7111, opened 2019-10-25
Make the init hook async

### LINKED ISSUE #5954: Run init hook asynchronously
Acceptance Criteria:
* Load a package with an init hook with a long sleep. I can still interact with the supervisor while this is happening.
* Validate that the init hook did what I expect (view the log output to verify)

davidMcneil:
Resolves #5954 

### Questions for Reviewers
* See this [comment](https://github.com/habitat-sh/habitat/pull/7111#discussion_r339025931)

### TODO
* [ ] Add end to end tests

### DIFF
--- a/components/sup/src/manager/service.rs
+++ b/components/sup/src/manager/service.rs
@@ -166,6 +166,8 @@ impl TemplateUpdate {
 #[derive(Debug, PartialEq, Eq)]
 enum ServiceState {
     Uninitialized,
+    Initializing,
+    InitializerFinished,
     Initialized,
 }
 
@@ -840,14 +842,13 @@ impl Service {
         let _timer = hook_timer("initialize");
 
         outputln!(preamble self.service_group, "Initializing");
-        self.state = ServiceState::Initialized;
         if let Some(ref hook) = self.hooks.init {
             self.state = if hook.run(&self.service_group,
                                      &self.pkg,
                                      self.svc_encrypted_password.as_ref())
                                 .unwrap_or(false)
             {
-                ServiceState::Initialized
+                ServiceState::InitializerFinished
             } else {
                 ServiceState::Uninitialized
             };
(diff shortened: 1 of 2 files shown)


components/sup/src/manager/service.rs
```diff
@@ -1217,7 +1252,6 @@ impl<'a> Serialize for ServiceProxy<'a> {
         strukt.serialize_field("desired_state", &s.desired_state)?;
         strukt.serialize_field("health_check", &s.health_check_result)?;
         strukt.serialize_field("hooks", &s.hooks)?;
-        strukt.serialize_field("initialized", &s.initialized)?;
```
davidMcneil: Is this needed? It looks as if we just check if the process is up and use that to determine if the process was initialized.

components/sup/src/manager/service.rs
```diff
@@ -163,6 +163,16 @@ impl TemplateUpdate {
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
+enum ServiceState {
+    Uninitialized,
+    Initialized,
+}
```
christophermaier: I _love_ enums instead of `bool`s... it's so much more descriptive!

components/sup/src/manager/service.rs
christophermaier: Yes and no :frowning_face: 

The fact that this field is here at all is because waaaaay back in the beginning, the HTTP gateway basically made direct JSON serializations of internal datastructures available to users. I can't really think of a legitimate use for this field in such a situation, but it _is_ available to users.

For additional background, check out https://github.com/habitat-sh/habitat/pull/5689 as well as its spiritual predecessor https://github.com/habitat-sh/habitat/pull/4823. At some point I'd like to do something like https://github.com/habitat-sh/habitat/issues/4872, but for the HTTP gateway data.

For now, though, to preserve existing behavior, this should turn into something like

```rust
strukt.serialize_field("initialized", &s.state.initialized())?;
```

components/sup/src/manager/service.rs
```diff
@@ -276,7 +277,8 @@ impl Service {
                      last_election_status: ElectionStatus::None,
                      user_config_updated: false,
                      needs_restart: false,
-                     initialization_state: InitializationState::Uninitialized,
+                     initialization_state:
+                         Arc::new(RwLock::new(InitializationState::Uninitialized)),
```
christophermaier: :thinking: Wonder if it'd be worth implementing `Default` for `InitializationState`...

PR #7113, opened 2019-10-25
e2e tests for hab pkg bulkupload, wired up into expeditor

jeremymv2:
Basic tests for `hab pkg bulkupload`

Caveat: This doesn't create a clean room Builder or origin. It merely tests for the proper exit code of commands. Perhaps a future iteration of these tests could verify against a clean room Builder, ensuring that it was properly mutated by the commands.


test/end-to-end/test_pkg_bulkupload.sh
```diff
@@ -0,0 +1,158 @@
...
+    ${HAB} pkg sign --origin ${HAB_ORIGIN} ${PKG_B_TAR} ${PKG_B_HART}
+    echo
+}
+
+test_expecting_fail() {
+    DESC=$1
+    CMD=$2
+
+    echo
+    echo "==========Expected failure: Testing ${DESC}"
```
smacfarlane: Not sure if it's appropriate on this particular line, but you can prefix your `echo` strings with "--- " to give it a nice fold in the buildkite UI.   I've found this to be useful to break things up logically so that you can quickly find any output if you need to inspect the build logs

jeremymv2: Applied your suggestion in https://github.com/habitat-sh/habitat/pull/7113/commits/4df23afc31faf15a0324e1542128dc6a2f1357ad

.expeditor/end_to_end.pipeline.yml
```diff
@@ -370,6 +370,18 @@ steps:
...
+  - label: "[:linux: test_pkg_bulkupload]"
+    command:
+      - .expeditor/scripts/end_to_end/setup_environment.sh dev
+      - test/end-to-end/test_pkg_bulkupload.sh
+    expeditor:
+      executor:
+        docker:
+          privileged: true
+          environment:
+            - ACCEPTANCE_HAB_AUTH_TOKEN
```
christophermaier: You're going to need a `BUILD_PKG_TARGET` in here.

test/end-to-end/test_pkg_bulkupload.sh
```diff
@@ -0,0 +1,158 @@
...
+PKG_A_TAR="${FIXTURES_DIR}/testbulkupload-testpkg1-0.1.0-20191024190939.tar"
+PKG_B_TAR="${FIXTURES_DIR}/testbulkupload-testpkg2-0.1.0-20191024191005.tar"
+PKG_A_HART="${CACHE_DIR}/artifacts/testbulkupload-testpkg1-0.1.0-20191024190939-x86_64-linux.hart"
+PKG_B_HART="${CACHE_DIR}/artifacts/testbulkupload-testpkg2-0.1.0-20191024191005-x86_64-linux.hart"
+
+echo
+echo "--- Testing with command ${HAB}, using cache dir ${CACHE_DIR}"
+echo
+
+before_upload() {
```
christophermaier: I thing these are fine, but it seems like we're very quickly approaching a point where we need to bring in a more formal framework of some kind. For instance, all this seems like it wants to be `bats`, and I've actually got some current work where getting a test environment for `bats` set up (with all the helper libraries) doesn't actually suck. `aruba` is another nice library in the Ruby world. And we've also discussed some kind of Powershell path to make cross-platform testing a bit easier.

I think these tests that you're writing are a bit different from other ones we've got, in that they're more about exercising CLI activity (others are about Supervisor functionality, which bring with them other constraints). That's a distinction that may be worth considering as we move forward.

christophermaier: Also, this is just an observation / food for thought... I'm not asking for a rewrite in `bats` (or anything else!) for this PR :sweat_smile:

.expeditor/end_to_end.pipeline.yml
jeremymv2: Added in this commit https://github.com/habitat-sh/habitat/pull/7113/commits/e9f1516cefa6d66c5b1b5d082eb7c99ce97cbb95

jeremymv2:
I think this issue is pertinent to tests like the ones in this PR
https://github.com/habitat-sh/habitat/issues/7118

test/end-to-end/test_pkg_bulkupload.sh
```diff
@@ -0,0 +1,159 @@
...
+
+# TODO: Future improvement would be to create a clean room Builder or origin.
+# Doing so would allow for more accurate validation of the bulkupload by interrogating
+# the Builder. Of course this introduces significant setup time cost..
+
+set -euo pipefail
+
+export HAB_NOCOLORING=true
+export HAB_NONINTERACTIVE=true
+export HAB_ORIGIN="testbulkupload"
```
chefsalim: It would be great if we could use a more general test origin instead of specific to bulkupload.. I believe there has been some discussion already of a "habitat-test" origin

test/end-to-end/test_pkg_bulkupload.sh
markan: I'm +1 for merging this as is, but I think we should plan a spike near term to see if it's feasible to write a CLI test in powershell and make it work cross platform. If so, we should make that our standard going forward, and rewrite tests as a tech debt item.

PR #7115, opened 2019-10-25
[CI] Initial Expeditor Promotion trigger

christophermaier:


.expeditor/config.yml
```diff
@@ -95,3 +95,27 @@ schedules:
   - name: cargo_update
     description: "Run 'cargo update' every Monday at 12PM UTC"
     cronline: "0 12 * * 1"
+
+promote:
+  actions:
```
tduffield: So, I'm actually trying to deprecate this. Instead of this, please add the following to your subscriptions:

```
- workload: project_promoted:{{agent_id}}:current:*
  actions:
    - bash:.expeditor/scripts/expeditor_promote.sh
- workload: project_promoted:{{agent_id}}:stable:*
  actions:
    - bash:.expeditor/scripts/expeditor_promote.sh
```

This is a little bit clearer because it's declaring what you _do_ want to happen, rather than what you don't.

christophermaier: Nice... the exclusions I was using felt a little odd; I much prefer this formulation!

christophermaier: @tduffield So the `current` and `stable` in the workload are destinations, and not sources?

PR #7119, opened 2019-10-28
Update install scripts to pull from packages.chef.io for new releases

smacfarlane:
This PR should not be merged until we have promoted to stable from the new release pipeline. 

This PR updates our install scripts to pull from packages.chef.io for packages greater than version 0.89.0, as we migrate off of Bintray.  Older versions will still be stored there for the time being, so we need to ensure that we don't break people who need to install specific versions.

Acceptance criteria: 
- [x] Linux packages can be installed from packages.chef.io
- [x] Linux packages can be installed from Bintray
- [x] MacOS packages can be installed from packages.chef.io
- [x] Macos packages can be installed from Bintray
- [x] Windows packages can be installed from packages.chef.io
- [x] Windows packages can be installed from Bintray
- [x] Tests have been added for each platform

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -291,6 +291,32 @@ steps:
       automatic:
         limit: 1
 
+  - label: "[unit] :linux: install script"
+    env:
+      HAB_LICENSE: "accept-no-persist"
+    command:
+      - .expeditor/scripts/verify/test_install_script.sh
+    expeditor:
+      executor:
+        docker:
+    timeout_in_minutes: 5
+    retry:
+      automatic:
+        limit: 1
+  ###############################################################
+  # MacOS tests
+  - label: "[unit] :darwin: install script"
+    command:
+      - .expeditor/scripts/verify/test_install_script.sh
+    expeditor:
+      executor:
+        macos:
+          inherit-environment-vars: true
+    timeout_in_minutes: 5
+    retry:
+      automatic:
+        limit: 1
+
  #################################################################
 
   # See "[unit] :linux: ignored"; the same approach applies here for windows
@@ -523,6 +549,21 @@ steps:
           host_os: windows
           shell: [ "powershell", "-Command" ]
     timeout_in_minutes: 20
+    retry:
+      automatic:
+        limit: 1
+
+  - label: "[unit] :windows: install script"
+    env:
+      HAB_LICENSE: "accept-no-persist"
+    command:
+      - .expeditor/scripts/verify/test_install_script.ps1
+    expeditor:
+      executor:
+        docker: 
+          host_os: windows
+          shell: [ "powershell", "-Command" ]
+    timeout_in_minutes: 5
     retry:
       automatic:
         limit: 1
--- a/components/hab/install.ps1
+++ b/components/hab/install.ps1
@@ -23,8 +23,10 @@ param (
 
 $ErrorActionPreference="stop"
 
-Function Get-Version($version, $channel) {
-    $jsonFile = Join-Path (Get-WorkDir) "version.json"
+Set-Variable packagesChefioRootUrl -Option ReadOnly -value "https://packages.chef.io/files"
+
+Function Get-BintrayVersion($version, $channel) {
+    $jsonFile = Join-Path ($workdir) "version.json"
 
     # bintray expects a '-' to separate version and release and not '/'
     $version = $version.Replace("/", "-")
@@ -78,17 +80,48 @@ Function Get-File($url, $dst) {
 }
 
 Function Get-WorkDir {
-    Join-Path $env:temp "hab.XXXX"
+  $parent = [System.IO.Path]::GetTempPath()
+  [string] $name = [System.Guid]::NewGuid()
+  New-Item -ItemType Directory -Path (Join-Path $parent $name)
 }
 
-Function Get-Archive($channel, $version) {
+# Downloads the requested archive from packages.chef.io
+Function Get-PackagesChefioArchive($channel, $version) {
+    $url = $packagesChefioRootUrl
+    if(!$version -Or $version -eq "latest") {
+      $hab_url="$url/$channel/habitat/latest/hab-x86_64-windows.zip"
+    } else {
+      $hab_url="$url/habitat/${version}/hab-x86_64-windows.zip"
+    }
+    $sha_url="$hab_url.sha256sum"
+    $hab_dest = (Join-Path ($workdir) "hab.zip")
+    $sha_dest = (Join-Path ($workdir) "hab.zip.shasum256")
+
+    Get-File $hab_url $hab_dest
+    $result = @{ "zip" = $hab_dest }
+
+    # Note that this will fail on versions less than 0.71.0
+    # when we did not upload shasum files to bintray.
+    # NOTE: This is left in place because, while we don't ship <0.71.0
+    # from s3 today, the intent is to move old releases over 
+    try {
+        Get-File $sha_url $sha_dest
+        $result["shasum"] = (Get-Content $sha_dest).Split()[0]
+    }
+    catch {
+        Write-Warning "No shasum exists for $version. Skipping validation."
+    }
+    $result
+}
+
+Function Get-BintrayArchive($channel, $version) {
     $url = "https://api.bintray.com/content/habitat/$channel/windows/x86_64/hab-$version-x86_64-windows.zip"
     $query = "?bt_package=hab-x86_64-windows"
 
     $hab_url="$url$query"
     $sha_url="$url.sha256sum$query"
-    $hab_dest = (Join-Path (Get-WorkDir) "hab.zip")
-    $sha_dest = (Join-Path (Get-WorkDir) "hab.zip.shasum256")
+    $hab_dest = (Join-Path ($workdir) "hab.zip")
+    $sha_dest = (Join-Path ($workdir) "hab.zip.shasum256")
 
     Get-File $hab_url $hab_dest
     $result = @{ "zip" = $hab_dest }
@@ -139,7 +172,7 @@ Function Install-Habitat {
     $habPath = Join-Path $env:ProgramData Habitat
     if(Test-Path $habPath) { Remove-Item $habPath -Recurse -Force }
     New-Item $habPath -ItemType Directory | Out-Null
-    $folder = (Get-ChildItem (Join-Path (Get-WorkDir) "hab-*"))
+    $folder = (Get-ChildItem (Join-Path ($workdir) "hab-*"))
     Copy-Item "$($folder.FullName)\*" $habPath
     $env:PATH = New-PathString -StartingPath $env:PATH -Path $habPath
     $machinePath = [System.Environment]::GetEnvironmentVariable("PATH", "Machine")
@@ -166,7 +199,7 @@ Function New-PathString([string]$StartingPath, [string]$Path) {
 }
 
 Function Expand-Zip($zipPath) {
-    $dest = Get-WorkDir
+    $dest = $workdir
     try {
         # Works on .Net 4.5 and up (as well as .Net Core)
         # Yes on PS v5 and up we have Expand-Archive but this works on PS v4 too
@@ -203,13 +236,28 @@ Function Assert-Habitat($ident) {
     }
 }
 
+Function Test-UsePackagesChefio($version) {
+    # The $_patch may contain the /release string as well.
+    # This is fine because we only care about major/minor for this 
+    # comparison. 
+
+    $_major,$_minor,$_patch = $version -split ".",3,"SimpleMatch"
+    $v1 = New-Object -TypeName Version -ArgumentList $_major,$_minor
+    $v2 = New-Object -TypeName Version -ArgumentList "0.89"
+    !$version -Or ($v1 -ge $v2)
+}
+
 Write-Host "Installing Habitat 'hab' program"
 
 $workdir = Get-WorkDir
 New-Item $workdir -ItemType Directory -Force | Out-Null
 try {
-    $Version = Get-Version $Version $Channel
-    $archive = Get-Archive $channel $version
+    if(Test-UsePackagesChefio($Version)) { 
+      $archive = Get-PackagesChefioArchive $channel $version
+    } else {
+      $Version = Get-BintrayVersion $Version $Channel
+      $archive = Get-BintrayArchive $channel $version
+    }
     if($archive.shasum) {
         Assert-Shasum $archive
     }
--- a/components/hab/install.sh
+++ b/components/hab/install.sh
@@ -7,7 +7,7 @@ if [ -n "${DEBUG:-}" ]; then set -x; fi
 
 BT_ROOT="https://api.bintray.com/content/habitat"
 BT_SEARCH="https://api.bintray.com/packages/habitat"
-
+readonly pcio_root="https://packages.chef.io/files"
 export HAB_LICENSE="accept-no-persist"
 
 main() {
@@ -44,8 +44,12 @@ main() {
   create_workdir
   get_platform
   validate_target
-  get_version
-  download_archive
+  if use_packages_chef_io "$version"; then
+    download_packages_chef_io_archive "$version" "$channel" "$target"
+  else
+    get_bintray_version
+    download_bintray_archive
+  fi
   verify_archive
   extract_archive
   install_hab
@@ -144,7 +148,29 @@ get_platform() {
   fi
 }
 
-get_version() {
+use_packages_chef_io() {
+  need_cmd cut
+
+  local version
+  version="${1:-latest}"
+
+  if [ "$version" == "latest" ]; then
+    info "No version specified, using packages.chef.io"
+    return 0
+  else 
+    local major 
+    local minor
+    major="$(echo "${version}" | cut -d'.' -f1)"
+    minor="$(echo "${version}" | cut -d'.' -f2)"
+    if [ "$major" -ge 1 ] || [ "$minor" -ge 89 ]; then
+      info "Specified recent version >= 0.89, using packages.chef.io"
+      return 0
+    fi
+  fi
+  return 1
+}
+
+get_bintray_version() {
   need_cmd grep
   need_cmd head
   need_cmd sed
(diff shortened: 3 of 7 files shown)


components/hab/install.sh
```diff
@@ -295,7 +362,7 @@ install_hab() {
       # /bin means now you have multiple copies of hab on your system and pathing
       # shenanigans might ensue. Rather than deal with that mess, we do it this
       # way.
-      "${archive_dir}/hab" pkg install --binlink --force --channel "$channel" "$_ident"
+      "${archive_dir}/hab" pkg install --binlink --force --channel "$channel" "$_ident" -u https://bldr.acceptance.habitat.sh
```
smacfarlane: This must be removed before merging

components/hab/tests/test_install_script.bats
```diff
@@ -0,0 +1,100 @@
...
+  linux || skip
+  run components/hab/install.sh -c dev
+
+  [ "$status" -eq 0 ]
+  [ "$(installed_target)" == "x86_64-linux" ]
+}
+
+@test "Install specific version for x86_64-linux" {
+  linux || skip
+  run components/hab/install.sh -v 0.89.48 -c dev
```
smacfarlane: `-c dev` should be removed

components/hab/tests/test_install_script.bats
```diff
@@ -0,0 +1,100 @@
...
+installed_target() {
+  version_release="$(hab --version | cut -d' ' -f2)"
+  version="$(cut -d'/' -f1 <<< "$version_release")"
+  release="$(cut -d'/' -f2 <<< "$version_release")"
+  cat /hab/pkgs/core/hab/$version/$release/TARGET
+}
+
+@test "Install latest for x86_86-linux" {
+  linux || skip
+  run components/hab/install.sh -c dev
```
smacfarlane: `-c dev` should be removed

components/hab/tests/test_install_script.bats
```diff
@@ -0,0 +1,100 @@
...
+  run components/hab/install.sh -v 0.79.1 
+
+  [ "$status" -eq 0 ]
+  [ "$(installed_version)" == "hab 0.79.1" ]
+  [ "$(installed_target)" == "x86_64-linux" ]
+}
+
+@test "Install latest for x86_64-linux-kernel2" {
+  linux || skip
+  run components/hab/install.sh -t "x86_64-linux-kernel2" -c dev
```
smacfarlane: `-c dev` should be removed

components/hab/tests/test_install_script.bats
```diff
@@ -0,0 +1,100 @@
...
+
+  [ "$status" -eq 0 ]
+  [ "$(installed_target)" == "x86_64-linux-kernel2" ]
+}
+
+@test "Install specific version for x86_64-linux-kernel2" {
+  linux || skip
+  run components/hab/install.sh -v 0.89.48 -t "x86_64-linux-kernel2" -c dev
```
smacfarlane: `-c dev` should be removed

components/hab/tests/test_install_script.bats
```diff
@@ -0,0 +1,100 @@
...
+  run components/hab/install.sh -v 0.79.1 -t "x86_64-linux-kernel2"
+
+  [ "$status" -eq 0 ]
+  [ "$(installed_version)" == "hab 0.79.1" ]
+  [ "$(installed_target)" == "x86_64-linux-kernel2" ]
+}
+
+@test "Install latest for x86_86-darwin" {
+  darwin || skip
+  run components/hab/install.sh -c dev
```
smacfarlane: `-c dev` should be removed

components/hab/install.sh
```diff
@@ -202,7 +230,37 @@ validate_target() {
...
+  mv -v "${workdir}/hab-${version}.${ext}" "${archive}"
+  mv -v "${workdir}/hab-${version}.${ext}.sha256sum" "${sha_file}"
+  
+  if command -v gpg >/dev/null; then
+    info "GnuPG tooling found, downloading signiatures"
+    sha_sig_file="${archive}.sha256sum.asc"
+    key_file="${workdir}/habitat.asc"
+
+    dl_file "${url}.sha256sum.asc" "${sha_sig_file}"
+    dl_file "${url}.asc" "${key_file}"
```
smacfarlane: This is currently not the correct file.

components/hab/install.sh
```diff
@@ -144,7 +150,29 @@ get_platform() {
...
+    minor="$(echo "${version}" | cut -d'.' -f2)"
+    if [ "$major" -ge 1 ] || [ "$minor" -ge 89 ]; then
+      info "Specified recent version >= 0.89, using packages.chef.io"
+      return 0
+    fi
+  fi
+  return 1
+}
+
+get_packages_chef_io_version() {
```
smacfarlane: Do we even need this given that we s3 "channels" don't track versions?

components/hab/install.sh
```diff
@@ -7,7 +7,8 @@ if [ -n "${DEBUG:-}" ]; then set -x; fi
 
 BT_ROOT="https://api.bintray.com/content/habitat"
 BT_SEARCH="https://api.bintray.com/packages/habitat"
-
+#PCIO_ROOT="https://packages.chef.io"
+PCIO_ROOT="https://chef-automate-artifacts.s3-us-west-2.amazonaws.com"
```
smacfarlane: This must use packages.chef.io before merging

components/hab/install.ps1
```diff
@@ -81,6 +107,35 @@ Function Get-WorkDir {
     Join-Path $env:temp "hab.XXXX"
 }
 
+Function Get-Archive-Pcio($channel, $version) {
+    # $url = "https://packages.chef.io"
```
smacfarlane: This must use packages.chef.io before merging

components/hab/install.ps1
```diff
@@ -23,6 +23,32 @@ param (
...
+    } else {
+      # This differs slightly from the way we fetch with Bintray. With Bintray,
+      # we need to parse json to determine if the requested version is present.
+      # With packages.chef.io, we need to fetch the manifest for the requested version, 
+      # but we do not need to parse it as presence of the manifest is sufficient.
+      # The downside to this is we can't take $version/$release format identifiers anymore.
+      # A user could theoretically install a version that failed e2e tests as we don't 
+      # track channel in the `/files` directory, and channels in s3 only track head.
+
+      # $url = "https://packages.chef.io"
```
smacfarlane: This must use packages.chef.io before merging

components/hab/tests/test_install_script.ps1
```diff
@@ -0,0 +1,22 @@
+Describe "Install habitat using install.ps1" {
+    It "can install the latest version of Habitat" {
+        components/hab/install.ps1 -c dev
```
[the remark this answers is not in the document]
smacfarlane: Good catch! I missed these two, they should go.

smacfarlane: `-c dev` should be removed

components/hab/tests/test_install_script.ps1
```diff
@@ -0,0 +1,22 @@
...
+        $LASTEXITCODE | Should -Be 0
+    }
+
+    It "can install a specific version of Habitat" {
+        components/hab/install.ps1 -v 0.89.46 -c dev
```
smacfarlane: `-c dev` should be removed

components/hab/install.ps1
```diff
@@ -23,6 +23,32 @@ param (
 
 $ErrorActionPreference="stop"
 
+Function Get-Version-Pcio($version, $channel) {
```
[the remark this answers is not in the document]
smacfarlane: I was having trouble shortening `packages.chef.io` into something that was still meaningful.  I went this route because the longer term goal is to migrate old release out of bintray and then this can revert to `Get-Version`.

components/hab/install.sh
```diff
@@ -144,7 +150,29 @@ get_platform() {
...
-get_version() {
+use_packages_chef_io() {
+  need_cmd cut
+
+  if [ "$version" == "" ]; then
+    info "No version specified, using packages.chef.io"
+    return 0
+  else 
+    major="$(echo "${version}" | cut -d'.' -f1)"
+    minor="$(echo "${version}" | cut -d'.' -f2)"
```
stevendanna: We could consider declaring `major` and `minor`  with `local`

components/hab/install.sh
```diff
@@ -202,7 +230,37 @@ validate_target() {
...
+
+  _version="${version:-latest}"
+
+  if [ "$_version" == "latest" ]; then
+    url="${PCIO_ROOT}/${channel}/latest/habitat/hab-${target}.${ext}"
+  else 
+    url="${PCIO_ROOT}/files/habitat/${version}/hab-${target}.${ext}"
+  fi
+  
+  dl_file "${url}" "${workdir}/hab-${version}.${ext}"
```
stevendanna: Should we use _version here and below?  Otherwise won't we create this with the file name `hab-.EXT`?

components/hab/install.sh
```diff
@@ -202,7 +230,37 @@ validate_target() {
...
+  dl_file "${url}.sha256sum" "${workdir}/hab-${version}.${ext}.sha256sum"
+
+  archive="hab-${target}.${ext}"
+  sha_file="hab-${target}.${ext}.sha256sum"
+
+  mv -v "${workdir}/hab-${version}.${ext}" "${archive}"
+  mv -v "${workdir}/hab-${version}.${ext}.sha256sum" "${sha_file}"
+  
+  if command -v gpg >/dev/null; then
+    info "GnuPG tooling found, downloading signiatures"
```
stevendanna: s/signiatures/signatures/

components/hab/install.sh
```diff
@@ -221,22 +279,27 @@ download_archive() {
   info "Renaming downloaded archive files"
   mv -v "${workdir}/hab-latest.${ext}" "${archive}"
   mv -v "${workdir}/hab-latest.${ext}.sha256sum" "${archive}.sha256sum"
+  
+  if command -v gpg >/dev/null; then
+    info "GnuPG tooling found, downloading signiatures"
```
stevendanna: s/signiatures/signatures/

components/hab/install.sh
```diff
@@ -252,14 +315,18 @@ extract_archive() {
       need_cmd zcat
       need_cmd tar
 
-      zcat "${archive}" | tar x -C "${workdir}"
       archive_dir="${archive%.tar.gz}"
+      mkdir "${archive_dir}"
+      zcat "${archive}" | tar x -C "${archive_dir}" --strip-components=1
+
+      #archive_dir="${archive%.tar.gz}"
```
stevendanna: We can probably remove this commented code

components/hab/install.sh
smacfarlane: Good catch, thanks!

components/hab/install.ps1
```diff
@@ -23,6 +23,32 @@ param (
...
+      # The downside to this is we can't take $version/$release format identifiers anymore.
+      # A user could theoretically install a version that failed e2e tests as we don't 
+      # track channel in the `/files` directory, and channels in s3 only track head.
+
+      $manifest_url="$packagesChefioRootUrl/habitat/$version/manifest.json"
+      try {
+        $response = Invoke-WebRequest -Uri "$manifest_url" -ErrorAction Stop -UseBasicParsing
+      } catch {
+        $StatusCode = $_.Exception.Response.StatusCode.value__
+        Write-Error "Specified version ($version)[$manifest_url] not found."
```
mwrock: Not a bad idea to include the `$StatusCode`  in the `Write-Host` below

components/hab/install.ps1
```diff
@@ -208,8 +262,18 @@ Write-Host "Installing Habitat 'hab' program"
...
 New-Item $workdir -ItemType Directory -Force | Out-Null
 try {
-    $Version = Get-Version $Version $Channel
-    $archive = Get-Archive $channel $version
+    # The $_patch may contain the /release string as well,
+    # This is fine because we only care about major/minor for this 
+    # comparison. 
+    $_major,$_minor,$_patch = $version -split ".",3,"SimpleMatch"
+
+    if( $Version -eq "" -Or $_major -ge 1 -Or $_minor -ge 89 ) {
```
mwrock: if no `$version` is passed it will be `$null` and not empty. You could do:
```
if(!$Version -Or ([Version]::new($_major, $_minor) -ge [Version]::new("1.89")))
```

components/hab/install.sh
smacfarlane: This is something I needed to chat with you about :smile:  It should be the packages.chef.io equivalent of  `https://bintray.com/user/downloadSubjectPublicKey?username=habitat`

components/hab/tests/test_install_script.ps1
```diff
@@ -0,0 +1,22 @@
+Describe "Install habitat using install.ps1" {
```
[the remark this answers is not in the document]
smacfarlane: There is not, because the `Install-Habitat` function performs this action https://github.com/habitat-sh/habitat/blob/master/components/hab/install.ps1#L140

.expeditor/verify.pipeline.yml
```diff
@@ -272,6 +272,33 @@ steps:
...
+    expeditor:
+      executor:
+        docker:
+    timeout_in_minutes: 5
+    retry:
+      automatic:
+        limit: 1
+  ###############################################################
+  # MacOS tests
+  # TODO: Enable when we can run in Public CI
```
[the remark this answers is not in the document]
smacfarlane: They are! Forgot to remove this comment

components/hab/install.ps1
```diff
@@ -23,6 +23,32 @@ param (
 
 $ErrorActionPreference="stop"
 
+Set-Variable packagesChefioRootUrl -Option ReadOnly -value "https://packages.chef.io/files"
+
+Function Get-Version-Pcio($version, $channel) {
```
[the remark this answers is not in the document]
smacfarlane: Done!

components/hab/install.sh
```diff
@@ -7,7 +7,7 @@ if [ -n "${DEBUG:-}" ]; then set -x; fi
 
 BT_ROOT="https://api.bintray.com/content/habitat"
 BT_SEARCH="https://api.bintray.com/packages/habitat"
-
+PCIO_ROOT="https://packages.chef.io/files"
```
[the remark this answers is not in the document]
smacfarlane: Good point.  I was following the above pattern,  but I've adjusted to follow convention instead. We'll be ripping out the BT_ stuff soon enough :smile:

components/hab/install.sh
```diff
@@ -202,7 +231,37 @@ validate_target() {
   fi
 }
 
-download_archive() {
+download_packages_chef_io_archive() {
```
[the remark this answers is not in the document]
smacfarlane: I think I've done what I can in the functions I introduced.   There's still some functions external that expect the `sha_sig_file`, `key_file` to be set by this so I've left those as-is for now.

components/hab/install.sh
```diff
@@ -202,7 +231,37 @@ validate_target() {
...
+  need_cmd mv
+
+  _version="${version:-latest}"
```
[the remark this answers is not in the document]
smacfarlane: I've tried to bubble this as far up as I can.  I wanted to move it all the way up to the optparse layer, but bintray uses that odd "%24latest" string that would have necessitated deeper changes. Given that the bintray bits should go away later, it felt best to just wait until that time to set the default for `version`, like we do `channel`

components/hab/tests/test_install_script.bats
```diff
@@ -0,0 +1,100 @@
+setup() {
+  rm -f /bin/hab
+  rm -rf /hab/pkgs/core/hab
```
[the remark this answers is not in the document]
smacfarlane: Agreed, though if we don't clean up you can get installation errors.  I tried to limit the scope of the blast radius as much as possible, though maybe it's best to just throw an error here if it's not in a CI-like environment?

components/hab/tests/test_install_script.bats
smacfarlane: It would not.  I do think we want to test installations of specific versions, so the final change to this before merge would probably need to be the version we're about to release?  Which seems silly but I can't think of an alternative at this time.

components/hab/tests/test_install_script.ps1
mwrock: maybe a better check here than just checking exit code would be to ensure that `(get-command hab).Path` equals `C:\ProgramData\Habitat\hab.exe` which is where install.ps1 installs to.

components/hab/tests/test_install_script.bats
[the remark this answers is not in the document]
christophermaier: That sounds like a decent interim step, for sure.

components/hab/tests/test_install_script.bats
smacfarlane: After some conversation,  migrating .88 isn't an option just yet.

mwrock:
It would also be a good idea to manually test this on a 2008R2 instance just to be sure we don't have any PSv2 problems.

components/hab/install.ps1
```diff
@@ -23,7 +23,35 @@ param (
...
+      # we need to parse json to determine if the requested version is present.
+      # With packages.chef.io, we need to fetch the manifest for the requested version, 
+      # but we do not need to parse it as presence of the manifest is sufficient.
+      # The downside to this is we can't take $version/$release format identifiers anymore.
+      # A user could theoretically install a version that failed e2e tests as we don't 
+      # track channel in the `/files` directory, and channels in s3 only track head.
+
+      $manifest_url="$packagesChefioRootUrl/habitat/$version/manifest.json"
+      try {
+        $response = Invoke-WebRequest -Uri "$manifest_url" -ErrorAction Stop -UseBasicParsing
```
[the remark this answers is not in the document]
smacfarlane: I think we're going to remove this function.  As we migrate older versions out of bintray to s3,  they won't have a manifest to perform this check.  It's currently using presence of the manifest.json as a test for availability of a version, to mimic the bintray behavior but that behavior is no longer necessary as we can infer the full url path from the version.  

The underlying change is that our filenames will no longer have the release timestamp in them, so we don't need to make a query first.

components/hab/install.ps1
```diff
@@ -203,13 +236,25 @@ Function Assert-Habitat($ident) {
     }
 }
 
+Function Test-UsePackagesChefio($version) {
+    # The $_patch may contain the /release string as well.
+    # This is fine because we only care about major/minor for this 
+    # comparison. 
+    $_major,$_minor,$_patch = $version -split ".",3,"SimpleMatch"
+    !$version -Or $version -ge 1 -Or $version -ge 89
```
mwrock: this seems wrong? I'm assuming you meant `!$version -Or $_major -ge 1 -Or $_minor -ge 89`. Personally I prefer `!$version -Or [Version]::new($_major, $_minor) -ge [Version]::new("0.89")`

components/hab/install.ps1
smacfarlane: I had to revert this change in order to be able to run on Windows 2008r2

components/hab/install.ps1
smacfarlane: :man_facepalming: Yes, that is what it should be, will fix.  

Unfortunately, ` [Version]::new ` threw an error on 2008r2 so I had to revert to this approach

components/hab/install.ps1
mwrock: got it. but the revert seems off because it uses `$version` in all comparisons

components/hab/install.ps1
smacfarlane:
```
PS C:\Users\Administrator\Desktop> .\install.ps1 -v 0.79.1
Installing Habitat 'hab' program
Method invocation failed because [System.Version] doesn't contain a method named 'new'.
At C:\Users\Administrator\Desktop\install.ps1:247 char:5
+     !$version -Or ([Version]::new($_major, $_minor) -ge [Version]::new("0.89"))
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : MethodNotFound
```

components/hab/install.ps1
smacfarlane: "revert" was the wrong word to use.  "Change back" would have been better, and then I copy/paste failed :face_with_head_bandage:

mwrock: oh on 2008R2 you would use:
```
$v1 = New-Object -TypeName Version -ArgumentList $_major,$_minor
$v2 = New-Object -TypeName Version -ArgumentList "0.89"
```

mwrock: ^^^ works on all versions

PR #7123, opened 2019-10-28
poll for service startup in windows e2e tests

mwrock:


mwrock:
It was just padding. But I like the "pre-install" strategy. Gonna go with that and use the polling function to wait just 5 seconds like before.

PR #7130, opened 2019-10-29
[CI] Always clean up release channels

christophermaier:
Previously, we deleted release channels at the end of successful
release pipeline runs. However, failed and canceled runs would leave
channels behind. Over time, this can accumulate a lot of extra
channels, which can (with the current implementation of things) slow
down build operations.

Here, we remove this responsibility from the Buildkite pipeline, and
instead allow Expeditor to handle it. This allows us to remove the
offending channel all the time, regardless of what happens in the
pipeline.


.expeditor/scripts/release_habitat/expeditor_destroy_build_channel.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# We want to delete the channel whether or not the pipeline completes
+# successfully, because otherwise, our channels will build up, causing
+# API slowdowns :(
+
+set -euo pipefail
+
+source .expeditor/scripts/release_habitat/shared.sh
+
+export HAB_BLDR_URL="${temporary_hab_bldr_url}"
```
scotthain: should this be `acceptance_hab_bldr_url`?

christophermaier: It actually is... I changed it because Shellcheck rightly pointed out that it could cause confusion between it and `ACCEPTANCE_HAB_BLDR_URL`.

Once we switch over to prod, though, I figure this will just go away.

Issue #7101, opened 2019-10-24
hab pkg bulkupload should auto create origins if necessary


jeremymv2:
Let's make it relatively effortless to use `hab pkg bulkupload` by auto creating any origins that don't exist on the Builder target.

eeyun:
I like this in concept, I wonder though if this is overloading the behavior of the upload.? I wonder if it wouldn't be better to mirror our existing `hab origin delete` subcommand with a `hab origin create` subcommand ? That way its not a side effect that the origin gets created but you also dont have to deal with logging in to the front end to create the origins?

chefsalim:
Yeah, if we do the auto-create,  let's prioritize getting origin management on the board - at least have the capability to change the origin owner. I like @eeyun 's idea of adding hab origin create as a cli command also.

jeremymv2:
@eeyun ahh, I do see we have that command already!

```
14:56 $ hab origin create --help
hab-origin-create 0.88.0/20191009205151
Creates a new Builder origin

USAGE:
    hab origin create [OPTIONS] <ORIGIN>
```

chefsalim:
Cool, one less thing to do! :)

eeyun:
oh weird... i might have written this even. I might have found a bug then because i didnt see it in the help text output!

PR #7133, opened 2019-10-31
hab pkg bulkupload origin creation option

jeremymv2:
Closes https://github.com/habitat-sh/habitat/issues/7101

## UX ⚙️ 
```
vagrant@chef-builder:~$ ./hab pkg bulkupload --url https://chef-builder.test --create-origins transfer/
» Preparing to upload artifacts to the 'unstable' channel on https://chef-builder.test
→ Using transfer/artifacts for artifacts and transfer/keys for signing keys.
→ Found 48 artifact(s) for upload.
☁ Discovering origin names from artifact metadata..
✓  chef
✓  core
✓  effortless
→ Found 3 origin(s) that may need to be created.
Ø Your Builder id will own any origin created. Be sure this is what you intend!
Ownership transfer is not yet implemented and will require SQL commands. [Yes/no/quit] 
...
```

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -565,7 +565,10 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                      of the value of this option.")
                 (@arg FORCE: --force "Skip checking availability of package and \
                     force uploads, potentially overwriting a stored copy of a package.")
-                (@arg AUTO_BUILD: --("auto-build") "Enable auto-build for all packages in this upload. Only applicable to SaaS Builder.")
+                (@arg AUTO_BUILD: --("auto-build") "Enable auto-build for all packages in this upload. \
+                    Only applicable to SaaS Builder.")
+                (@arg AUTO_CREATE_ORIGINS: --("auto-create-origins") "Skip the confirmation prompt and \
+                    automatically create origins that do not exist in the target Builder.")
                 (@arg UPLOAD_DIRECTORY: +required {dir_exists}
                     "Directory Path from which artifacts will be uploaded.")
             )
--- a/components/hab/src/command/pkg/bulkupload.rs
+++ b/components/hab/src/command/pkg/bulkupload.rs
@@ -1,0 +1,128 @@
+//! Uploads packages from cache to a [Depot](../depot).
+//!
+//! # Examples
+//!
+//! ```bash
+//! $ hab pkg bulkupload /path/to/artifact_download_dir \
+//!     -u http://localhost:9632
+//! ```
+//!
+//! Will upload all packages in cache to Builder.
+
+use crate::{api_client::{self,
+                         BuildOnUpload,
+                         Client},
+            command,
+            common::ui::{Glyph,
+                         Status,
+                         UIReader,
+                         UIWriter,
+                         UI},
+            error::{Error,
+                    Result},
+            hcore::{package::PackageArchive,
+                    ChannelIdent},
+            PRODUCT,
+            VERSION};
+use glob::glob_with;
+use reqwest::StatusCode;
+use std::{collections::BTreeSet,
+          path::{Path,
+                 PathBuf}};
+
+/// Bulk Upload the packages from the cache to a Depot.
+///
+/// # Failures
+///
+/// * Fails if it cannot create a missing origin
+/// * Fails if it cannot upload the artifact
+#[allow(clippy::too_many_arguments)]
+pub fn start(ui: &mut UI,
+             bldr_url: &str,
+             additional_release_channel: &Option<ChannelIdent>,
+             token: &str,
+             artifact_path: &Path,
+             force_upload: bool,
+             auto_build: BuildOnUpload,
+             auto_create_origins: bool,
+             key_path: &Path)
+             -> Result<()> {
+    const OPTIONS: glob::MatchOptions = glob::MatchOptions { case_sensitive:              true,
+                                                             require_literal_separator:   true,
+                                                             require_literal_leading_dot: true, };
+    let artifact_paths =
+        vec_from_glob_with(&artifact_path.join("*.hart").display().to_string(), OPTIONS);
+
+    ui.begin(format!("Preparing to upload artifacts to the '{}' channel on {}",
+                     additional_release_channel.clone()
+                                               .unwrap_or_else(ChannelIdent::unstable),
+                     bldr_url))?;
+    ui.status(Status::Using,
+              format!("{} for artifacts and {} for signing keys",
+                      &artifact_path.display(),
+                      key_path.display()))?;
+    ui.status(Status::Found,
+              format!("{} artifact(s) for upload.", artifact_paths.len()))?;
+    ui.status(Status::Discovering,
+              String::from("origin names from local artifact cache"))?;
+
+    let mut origins = BTreeSet::new();
+    for artifact_path in &artifact_paths {
+        let ident = PackageArchive::new(&artifact_path).ident()?;
+        origins.insert(ident.origin);
+    }
+    let mut origins_to_create: Vec<String> = Vec::new();
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None)?;
+
+    for origin in origins {
+        match api_client.check_origin(&origin, token) {
+            Ok(()) => {
+                ui.status(Status::Custom(Glyph::CheckMark,
+                                         format!("Origin '{}' already exists", &origin)),
+                          String::from(""))?;
+            }
+            Err(api_client::Error::APIError(StatusCode::NOT_FOUND, _)) => {
+                ui.status(Status::Missing, format!("origin '{}'", &origin))?;
+                origins_to_create.push(origin);
+            }
+            Err(err) => return Err(Error::from(err)),
+        }
+    }
+
+    if !origins_to_create.is_empty() {
+        if !auto_create_origins {
+            ui.warn(String::from("Origins are required for uploading the artifacts. The \
+                                  Builder account that creates the origin is the owner."))?;
+            if !ask_create_origins(ui)? {
+                return Ok(());
+            };
+        };
+        for origin_to_create in origins_to_create {
+            command::origin::create::start(ui, &bldr_url, &token, &origin_to_create)?;
+        }
+    };
+
+    for artifact_path in &artifact_paths {
+        command::pkg::upload::start(ui,
+                                    &bldr_url,
+                                    &additional_release_channel,
+                                    &token,
+                                    &artifact_path,
+                                    force_upload,
+                                    auto_build,
+                                    &key_path)?
+    }
+
+    Ok(())
+}
+
+fn vec_from_glob_with(pattern: &str, options: glob::MatchOptions) -> Vec<PathBuf> {
+    glob_with(pattern, options).unwrap()
+                               .map(std::result::Result::unwrap)
+                               .collect()
+}
+
+fn ask_create_origins(ui: &mut UI) -> Result<bool> {
+    Ok(ui.prompt_yes_no("Create the above missing origins under your Builder account?",
+                        Some(true))?)
+}
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -11,7 +11,6 @@ use clap::{ArgMatches,
            Shell};
 use env_logger;
 use futures::prelude::*;
-use glob::glob_with;
 use hab::{cli::{self,
                 parse_optional_arg},
           command::{self,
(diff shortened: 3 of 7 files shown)


apriofrost:
Some nitpicks for the UX in the terminal. The goal is to keep the confirmation message consistent with what's in the GUI. Would love to hear @kagarmoe 's thoughts on it as well!
```
vagrant@chef-builder:~$ ./hab pkg bulkupload --url https://chef-builder.test transfer/
» Preparing to upload artifacts to the 'unstable' channel on https://chef-builder.test
→ Using transfer/artifacts for artifacts and transfer/keys for signing keys.
→ Found 48 artifact(s) for upload.
☁ Discovering origin names from artifact metadata..
✓  chef
✓  core
✓  effortless
→ Found 3 missing origin(s).
Ø Origins are required for uploading the artifacts. The Builder account that created the origin is the owner. Ownership transfer is not yet supported and will require SQL commands (link to docs if we have it?). 
Create the following origins under your Builder account? [Yes/no/quit] 
chef
core
effortless
...
```

components/hab/src/main.rs
```diff
@@ -901,9 +906,37 @@ fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+        }
+        ui.status(Status::Found,
+                  format!("{} origin(s) that may need to be created.", origins.len()))?;
+        if !origins.is_empty() && !accept {
+            ui.warn("Your Builder id will own any origin created. Be sure this is what you \
+                     intend!\nOwnership transfer is not yet implemented and will require SQL \
+                     commands."
+                               .to_string())?;
+            if !ask_create_origins(ui)? {
+                process::exit(1);
```
chefsalim: process exit seems overkill - can we not just return here with an Ok(())?

components/hab/src/main.rs
```diff
@@ -901,9 +906,37 @@ fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+            let ident = PackageArchive::new(&artifact_path).ident()?;
+            origins.push(ident.origin);
+        }
+        origins.sort();
+        origins.dedup();
+        for origin in &origins {
+            ui.status(Status::Custom(Glyph::CheckMark, "".to_string()), origin)?;
+        }
+        ui.status(Status::Found,
+                  format!("{} origin(s) that may need to be created.", origins.len()))?;
```
chefsalim: I think we should only give this warning if we find origins that are NOT present.  This seems like a generic warning being given all the time?

components/hab/src/main.rs
```diff
@@ -901,9 +906,37 @@ fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+        if !origins.is_empty() && !accept {
+            ui.warn("Your Builder id will own any origin created. Be sure this is what you \
+                     intend!\nOwnership transfer is not yet implemented and will require SQL \
```
chefsalim: This message seems a bit heavy to me.  We should fast follow with the origin ownership transfer API/cli so we dont have to put in draconian warnings that will live around for a long time and be incorrect as soon as that API lands.

components/hab/src/main.rs
```diff
@@ -901,9 +906,37 @@ fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
     ui.status(Status::Found,
-              format!("{} artifacts for upload.", artifact_paths.len()))?;
-
-    for artifact_path in artifact_paths {
+              format!("{} artifact(s) for upload.", artifact_paths.len()))?;
+
+    if create_origins {
+        let mut origins: Vec<String> = Vec::new();
+        ui.status(Status::Discovering,
+                  "origin names from artifact metadata..".to_string())?;
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/7133/commits/3d92d7571c18c2d95e72072a930114c5caacc0dc

components/hab/src/main.rs
```diff
@@ -901,9 +906,37 @@ fn sub_pkg_bulkupload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+        for artifact_path in &artifact_paths {
+            let ident = PackageArchive::new(&artifact_path).ident()?;
+            origins.push(ident.origin);
+        }
+        origins.sort();
+        origins.dedup();
```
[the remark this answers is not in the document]
jeremymv2: Great callout!
Handled in https://github.com/habitat-sh/habitat/pull/7133/commits/73624f731d3319c8ed2cd6ef56932c99f0d7ac18

components/hab/src/main.rs
jeremymv2: Handled here https://github.com/habitat-sh/habitat/pull/7133/commits/5c46b2f97331cb937da872a6638e69252395d043

components/hab/src/main.rs
jeremymv2: Agreed that is less useful then what I had intended.
Because the Builder API [origins](https://github.com/habitat-sh/builder/blob/1626c3eec4389329df9044ae5a9075c0faba89e1/components/builder-api/doc/api.raml#L780) endpoint doesn't currently implement a `GET` or list functionality to enumerate the present origins, we just have to assume none of the origins exist and call `command::origin::create::start()` on each. 
Due to all of the above, I've just removed the message entirely in https://github.com/habitat-sh/habitat/pull/7133/commits/0c96916156ce8d0eeb4636319945c48c42d56bc6

jeremymv2: It seems that I am mistaken. The Builder API **does** implement a `GET` response on an origin name. 

**Note**: It is slightly more expensive to implement a `GET` followed by a `POST` as opposed to just blindly issuing a `POST` for the same end result. The existence check also adds some minor logic complexity as an additional cost.

The question then becomes: Is the additional existence check worthwhile to be able to provide the UX of only displaying the origins that need to be created?

components/hab/src/main.rs
jeremymv2: This Builder issue exists for ownership transfer https://github.com/habitat-sh/builder/issues/36

jeremymv2:
> Some nitpicks for the UX in the terminal. The goal is to keep the confirmation message consistent with what's in the GUI. 

@apriofrost I've addressed your feedback wrt utilizing `--auto-create-origins` in this [commit](https://github.com/habitat-sh/habitat/pull/7133/commits/9f726c1fe1d6a3a53fa47c00d9f2ef5340de6c0a)

components/hab/src/main.rs
markan: I think it might be worthwhile to do the extra work, and change the flow a little bit.

Consider the case of a sync after the initial bootstrap; all the origins will be in place (most likely). I think the user would like to be warned if a new origin was needed, rather than have to mentally go through and say 'ok, core is ok, chef is ok, etc'.

If I understand the current behavior, if create_origins is false, and and an origin does not exist, we will go through the upload process until we try to upload a package to a non-existent origin, where we will then error. Given that the user's attention might have shifted to something else while waiting for uploads, that can create unnecessary delay and friction. 

Maybe we should instead always check the origins, and list missing origins. If create_origins is true, we go on and offer to create them, and if it's false we fail early. 

The main downside is the walk through the artifacts path and the 'GET' from builder will impose some extra work. I have some ideas for optimizations if that turns out to be expensive.

jeremymv2: @markan @apriofrost @chefsalim with this [commit](https://github.com/habitat-sh/habitat/pull/7133/commits/9f726c1fe1d6a3a53fa47c00d9f2ef5340de6c0a) there is only one CLI argument related to origin creation ("--auto-create-origins"). That is in alignment with @apriofrost's UX [diagram](https://github.com/habitat-sh/habitat/pull/7133#pullrequestreview-310218276) above. I believe that changes some of your reasoning above.

It also ends up meaning that origins are always auto created, the only difference is whether or not we prompt the user for confirmation.

Keeping in mind that `ommand::origin::create::start()` on each origin issues a `POST` but handles any existing origins gracefully:

```
Ω Creating origin chef.
… Skipping creation of origin chef. Origin already exists!
```

Due to the fact that in the current implementation origins are _always_ created and the creation method is effectively idempotent, is it worth it to implement a check?

jeremymv2: I think I'll go ahead and implement an existence check for the sake of completeness and elimination of any assumptions 😃

markan: Thanks for the clarification of the behavior!

I still lean towards the existence check. I'm thinking about the sync case, where most of time the origins will exist. If we list those every time, even the ones that exist already, it will train the user to hit 'yes' (or set auto-create). If a new origin appears the chances are that reflex will have them say 'yes' because that's become the pattern. It's possible they will create an origin they didn't  intend to that way.

components/hab/src/cli.rs
```diff
@@ -566,8 +566,7 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                 (@arg FORCE: --force "Skip checking availability of package and \
                     force uploads, potentially overwriting a stored copy of a package.")
                 (@arg AUTO_BUILD: --("auto-build") "Enable auto-build for all packages in this upload. Only applicable to SaaS Builder.")
-                (@arg CREATE_ORIGINS: --("create-origins") "Create any origins that do not exist in targeted Builder.")
-                (@arg ACCEPT: --("accept") "Do not prompt for confirmation of auto creation and ownership of any origins that do not exist.")
+                (@arg AUTO_CREATE_ORIGINS: --("auto-create-origins") "Do not prompt for confirmation before auto creating any origins that do not exist.")
```
apriofrost: I'd suggest using a description like this: "Skip the confirmation and automatically create origins that do not exist in the target Builder."
The two "do not"  in a sentence may make it a little bit harder to understand.

jeremymv2: Thanks, adjusted  in https://github.com/habitat-sh/habitat/pull/7133/commits/5f5cafefb4ab581f26d69e4f59770c099d5ae378

apriofrost: Awesome! Thanks so much @jeremymv2 !

components/hab/src/main.rs
jeremymv2: I've implemented the client for the origin API check in https://github.com/habitat-sh/habitat/pull/7133/commits/bc818e341240f108e0210ce9c7c27bd80e522fed

components/hab/src/command/pkg/bulkupload.rs
```diff
@@ -0,0 +1,125 @@
...
+                          String::from(""))?;
+            }
+            Err(api_client::Error::APIError(StatusCode::NOT_FOUND, _)) => {
+                ui.status(Status::Missing, format!("origin '{}'", &origin))?;
+                origins_to_create.push(origin);
+            }
+            Err(err) => return Err(Error::from(err)),
+        }
+    }
+    if !auto_create_origins && !origins_to_create.is_empty() {
```
markan: Tastes vary, but I found this a little bit confusing because of the reliance on the fall through case either being auto create or empty list. You have to remember the invariant when looking at the for loop. Maybe the following would be clearer.
```
if !origins_to_create.is_empty() {
   if !auto_create_origins  {
        ui.warn(String::from("Origins are required for uploading the artifacts. The Builder \
                              account that creates the origin is the owner."))?;
        if !ask_create_origins(ui)? {
            return Ok(());
        };
    };
    for origin_to_create in origins_to_create {
        command::origin::create::start(ui, &bldr_url, &token, &origin_to_create)?;
    }
}

File: components/hab/src/command/pkg/bulkupload.rs [hunk 8]
```diff
@@ -0,0 +1,126 @@
...
+              String::from("origin names from local artifact cache.."))?;
+
+    let mut origins = BTreeSet::new();
+    for artifact_path in &artifact_paths {
+        let ident = PackageArchive::new(&artifact_path).ident()?;
+        origins.insert(ident.origin);
+    }
+    let mut origins_to_create: Vec<String> = Vec::new();
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None)?;
+    for origin in origins {
```
chefsalim: Super tiny nit - add a whitespace blank line to break up the code sections a bit more (readability)

components/hab/src/command/pkg/bulkupload.rs
```diff
@@ -0,0 +1,126 @@
...
+                                               .unwrap_or_else(ChannelIdent::unstable),
+                     bldr_url))?;
+    ui.status(Status::Using,
+              format!("{} for artifacts and {} for signing keys.",
+                      &artifact_path.display(),
+                      key_path.display()))?;
+    ui.status(Status::Found,
+              format!("{} artifact(s) for upload.", artifact_paths.len()))?;
+    ui.status(Status::Discovering,
+              String::from("origin names from local artifact cache.."))?;
```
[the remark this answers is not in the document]
jeremymv2: Thanks, handled in https://github.com/habitat-sh/habitat/pull/7133/commits/d13255c9d570503e6694cb0975e82350426e5314

components/hab/src/command/pkg/bulkupload.rs
```diff
@@ -0,0 +1,126 @@
...
+                          String::from(""))?;
+            }
+            Err(api_client::Error::APIError(StatusCode::NOT_FOUND, _)) => {
+                ui.status(Status::Missing, format!("origin '{}'", &origin))?;
+                origins_to_create.push(origin);
+            }
+            Err(err) => return Err(Error::from(err)),
+        }
+    }
+    if !origins_to_create.is_empty() {
```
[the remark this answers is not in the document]
jeremymv2: Thanks, I've handled in https://github.com/habitat-sh/habitat/pull/7133/commits/d13255c9d570503e6694cb0975e82350426e5314

components/hab/src/command/pkg/bulkupload.rs
jeremymv2: Thank you: addressed in https://github.com/habitat-sh/habitat/pull/7133/commits/d13255c9d570503e6694cb0975e82350426e5314

PR #7139, opened 2019-10-31
Automatically sync packages need for e2e tests with builder instance

davidMcneil:
The objective is to make all plans defined in [`test/fixtures/plans`](https://github.com/habitat-sh/habitat/tree/dmcneil/automatically-upload-e2e-test-packages/test/fixtures/plans) available for use in the end-to-end test pipeline. The current plans are just there for testing and will be removed before merging.

The first step is to make a script that can sync the plans in a directory with the packages on a builder instance. This script is defined [here](https://github.com/habitat-sh/habitat/blob/dmcneil/automatically-upload-e2e-test-packages/test/fixtures/plans/Sync-Packages.ps1). Sync-Packages recursively looks for plan files in `PlansPath`. For each plan, the following is done:

* Extract the package name and origin from the plan file to determine the package identifier
* Fetch the last time the package was updated from `<BuilderUrl>/v1/depot/pkgs/<package-identifier>/latest`
* Compare the most recent write time of files in the plan's directory with the packages last update
  time. If the package update time is more recent, the package is already up to date and we can
  skip this package.
* Build the package
* Install the package
* Upload the package to `BuilderUrl` with an optional `Channel`
* Verify that the packages update time is now more recent than the file modification time

The packages are built and installed from most to least nested. This allows packages to depend on
packages that are nested below it.

*Question*: Because the machines running the pipelines are ephemeral, I do not believe we will be able to use the last write time of the files in the package directory to determine if the package is up to date. What should we use? Probably the git history?

**The next step is currently not implemented, waiting to be sure this is the direction we want to go** The next step is to execute this script as part of the [`end_to_end.pipeline.yaml`](https://github.com/habitat-sh/habitat/blob/dmcneil/automatically-upload-e2e-test-packages/.expeditor/end_to_end.pipeline.yml) . This script will be run once on a linux machine and once on a windows machine to build the appropriate packages for each OS. These steps will run before actually executing the end-to-end tests. 

### DIFF
--- a/test/fixtures/plans/Sync-Packages.ps1
+++ b/test/fixtures/plans/Sync-Packages.ps1
@@ -1,0 +1,154 @@
+<#
+
+.SYNOPSIS
+Sync the packages in a specified directory with packages stored in builder
+
+.DESCRIPTION
+Sync-Packages recursivly looks for plan files in `PlansPath. For each plan the following is done:
+
+* Extract the package name and origin from the plan file to determine the package identifier
+* Fetch the last time the package was updated from `<BuilderUrl>/v1/depot/pkgs/<package-identifier>/latest`
+* Compare the most recent write time of files in the plans directory with the packages last update
+  time. If the package update time is more recent, the package is already up to date and we can
+  skip this package.
+* Build the package
+* Install the package
+* Upload the package to `BuilderUrl` with an optional `Channel`
+* Verify that the new package update time is more recent than the file modification time
+
+The packages are built and installed from most to least nested. This allows packages to depend on
+packages that are nested below it.
+
+#>
+
+[CmdletBinding()]
+Param (
+    [string]
+    # The builder URL to upload packages to 
+    $BuilderUrl = "https://bldr.acceptance.habitat.sh",
+    
+    [Parameter(Mandatory = $true)]
+    [string]
+    # The token to authenticate with when accessing the builder instance
+    $BuilderAuthToken,
+    
+    [string]
+    # The channel to tag packages with when uploading
+    $Channel,
+    
+    [Parameter(Mandatory = $true)]
+    [string]
+    # The path to the directory containing the plans to build and to sync with builder
+    $PlansPath
+)
+
+$ErrorActionPreference = "Stop"
+
+###################################################################################################
+
+function Get-PackageUpdateTime {
+    Param([string]$PackageIdentifier)
+
+    $uri = "$($BuilderUrl)/v1/depot/pkgs/$($PackageIdentifier)/latest"
+    $json = try { 
+        (Invoke-WebRequest -Uri $uri -ErrorAction Stop).Content
+    }
+    catch {
+        $statusCode = $_.Exception.Response.StatusCode.value__
+        if ($statusCode -eq 404) {
+            return $null
+        }
+        else {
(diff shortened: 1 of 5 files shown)


mwrock:
I wonder if it would make more sense to "connect" our fixture plans in acceptance and then builder can be in charge of auto building and managing dependency order.

davidMcneil:
> I wonder if it would make more sense to "connect" our fixture plans in acceptance and then builder can be in charge of auto building and managing dependency order.

I agree connecting the plans feels like the more "correct" solution, but I have two questions with this approach:

1. Is there any way to keep the connected plans in a folder under the habitat project? Or would they each need to have their own repo? I like the idea of keeping the test plans with the code. Git submodules would be one way to work around this, but I am not a huge fan of git submodules.
2. How would we know that builder was done and therfore we could proceed with the actual end-to-end tests? There is probably a way to continuously poll builder, but it is not completely clear how that would work.

mwrock:
You can use a .bldr.toml file to specify the location of a plan and have many in the same repo. This is what we do with the `core-plans` repo. We may be able to use `hab bldr job status --origin <TEST_ORIGIN>` which will report all the build jobs, when they were created and their status. For example:

```
C:\dev\habitat [no_ignore_signals]> hab bldr job status --origin core
⌂ Determining status of job groups in core origin

CREATED AT                        GROUP ID             STATUS    IDENT                  TARGET
2019-10-30T06:03:55.315024+00:00  1368595056055771136  Complete  core/grafana-promtail  x86_64-linux
2019-10-30T06:03:55.295025+00:00  1368595055887990784  Complete  core/grafana-promtail  x86_64-windows
2019-10-30T06:03:55.294064+00:00  1368595055887982592  Complete  core/grafana-promtail  x86_64-linux-kernel2
2019-10-30T05:49:29.421427+00:00  1368587792418799616  Complete  core/grafana-loki      x86_64-linux
2019-10-30T05:49:29.421374+00:00  1368587792418791424  Complete  core/grafana-loki      x86_64-linux-kernel2
2019-10-30T05:49:29.421342+00:00  1368587792418783232  Complete  core/grafana-loki      x86_64-windows
2019-10-30T05:36:59.478796+00:00  1368581501440925696  Complete  core/prometheus        x86_64-linux
2019-10-30T05:36:59.455328+00:00  1368581501239590912  Complete  core/prometheus1       x86_64-linux
2019-10-30T05:36:59.437962+00:00  1368581501096976384  Complete  core/prometheus2       x86_64-linux
2019-10-30T05:30:46.820952+00:00  1368578375367409664  Complete  core/prometheus        x86_64-linux
```

Good point about the pain of "connecting" the plan. I'd love to hear what others think but it just feels like we have built up quite alot to do what this script does and some of it is pretty fragile like parsing the plans and basing the nesting level by path length which we could totally harden but it would be cool to utilize our own kit here.

davidMcneil:
> it would be cool to utilize our own kit here.

Completely, agree. It makes more sense to use our own stuff here. My path forward is:
* write a script (should be pretty simple) that blocks until all jobs on builder for a specific channel are completed
* write a few simple packages and wire them up to builder
* add the script to `end_to_end.pipeline.yaml` before actually running the tests

davidMcneil:
Closing in favor of the implmentation disucessed [here](https://github.com/habitat-sh/habitat/pull/7139#issuecomment-549040947)

PR #7140, opened 2019-10-31
remove IGNORE_SIGNALS feature

### LINKED ISSUE #5908: Remove `IgnoreSignals` feature behavior when init hooks run in their own thread
While we will want to deprecate the feature itself, we should make it a no-op once the supervisor runs `init` hooks in their own threads. This feature just suppresses the supervisor's ability to handle shutdown and allows the default `ctrl+c` beavior to terminate the supervisor. That is mostly OK since the launcher handles most of the cleanup but ideally the supervisor should be able to receive signals that things are shutting down and act if necessary. We only suppress it when this feature is enabled for now so that stopping the supervisor does not hang when long running `init` hooks are running.

mwrock:
resolves #5908 

This also adds a e2e test to ensure a service's PID file is removed on shutdown which could not happen when the `IGNORE_SIGNALS` feature is enabled.

### DIFF
--- a/components/sup/src/manager.rs
+++ b/components/sup/src/manager.rs
@@ -1006,18 +1006,12 @@ impl Manager {
                 break ShutdownMode::Departed;
             }
 
-            // This formulation is gross, but it doesn't seem to compile on Windows otherwise.
-            #[allow(clippy::match_bool)]
-            #[allow(clippy::single_match)]
             #[cfg(unix)]
-            match self.feature_flags.contains(FeatureFlag::IGNORE_SIGNALS) {
-                false => {
-                    if signals::pending_sighup() {
-                        outputln!("Supervisor shutting down for signal");
-                        break ShutdownMode::Restarting;
-                    }
+            {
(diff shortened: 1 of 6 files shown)


components/sup/src/manager.rs
```diff
@@ -1006,18 +1006,12 @@ impl Manager {
                 break ShutdownMode::Departed;
             }
 
-            // This formulation is gross, but it doesn't seem to compile on Windows otherwise.
-            #[allow(clippy::match_bool)]
-            #[allow(clippy::single_match)]
             #[cfg(unix)]
```
markan: Out of curiosity, is there no equivalent signal in windows?  How does a windows service restart cleanly?

mwrock: There really is no windows equivilent. The closest thing to a ubiquitous "restart me" signal would be a Windows Service restart but that only applies to programs running as windows services that must implement a specific interface with `OnStart` and `OnStop` methods. Outside of services apps typically handle this in app specific ways.

PR #7148, opened 2019-11-04
Make `run_e2e_test.ps1` and `shared.ps1` script work on powershell core

davidMcneil:
Windows e2e tests can be run with:
```
powershell .expeditor/scripts/end_to_end/run_e2e_test.ps1 <channel> <test name>
```
Linux e2e tests can be run with:
```
bash .expeditor/scripts/end_to_end/run_e2e_test.sh <channel> <test name>
```

### DIFF
--- a/.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
+++ b/.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
@@ -1,0 +1,116 @@
+# This script expects `setup_environment.sh` or `setup_environment.ps1` to be sourced before execution
+
+param (
+    [string]$TestName,
+    [string]$BuilderUrl = $env:HAB_BLDR_URL
+)
+
+. .expeditor/scripts/shared.ps1
+
+$ErrorActionPreference = "stop"
+
+if (!$IsCoreCLR) {
+    Write-Error "End-to-end tests are only compatiable with Powershell Core"
+}
+
+###################################################################################################
+
+$EndToEndTestingOrigin = "habitat-testing"
+$TestStartTime = Get-Date
+if ($IsLinux -Or $IsMacOS) {
+    $env:SystemDrive = "/"
+}
+
+function Wait-True([ScriptBlock]$TestScript, [ScriptBlock]$TimeoutScript, [int]$Timeout) {
+    $startTime = Get-Date
+    while (!$TestScript.Invoke()) {
+        $timeTaken = [DateTime]::Now.Subtract($startTime)
+        if ($timeTaken.TotalSeconds -ge $Timeout) {
+            $TimeoutScript.Invoke()
+            break
+        }
+        Start-Sleep -Seconds 1
+    }
+}
+
+function Wait-PathUpdatedAfter($Path, $Time, $Timeout) {
+    $testScript = {
+        if (!(Test-Path -Path $Path)) {
+            return $False
+        }
+        (Get-Item -Path $Path).LastWriteTime -ge $Time
+    }
+    $timeoutScript = { Write-Error "Timed out waiting $Timeout seconds for '$Path' to be updated after '$($Time.ToString("yyyy-MM-ddTHH:MM:ssZ"))'" }
+    Wait-True -TestScript $testScript -TimeoutScript $timeoutScript -Timeout $Timeout
+}
+
+function Wait-Supervisor($Timeout = 1) {
+    Write-Host "Waiting up to $Timeout seconds for Supervisor to start..."
+    $testScript = { Test-Connection -ComputerName 127.0.0.1 -TCPPort 9631 }
+    $timeoutScript = { Write-Error "Timed out waiting $Timeout seconds for Supervisor to start" }
+    Wait-True -TestScript $testScript -TimeoutScript $timeoutScript -Timeout $Timeout
+    Write-Host "Supervisor is now running."
+}
+
+function Start-Supervisor($Timeout = 1) {
+    hab sup run --no-color &
+    Wait-Supervisor -Timeout $Timeout
+}
+
+function Wait-SupervisorService($ServiceName, $Timeout = 1) {
+    Write-Host "Waiting up to $Timeout seconds for Supervisor to start $ServiceName ..."
+    $testScript = { 
+        try {
+            $status = (Invoke-WebRequest "http://localhost:9631/services/$ServiceName/default" |
+                ConvertFrom-Json).process.state
+            $status -eq "up"
+        }
+        catch { } # We ignore 404s and other unsuccesful codes
+    }
+    $timeoutScript = { Write-Error "Timed out waiting $Timeout seconds for Supervisor to start $ServiceName" }
+    Wait-True -TestScript $testScript -TimeoutScript $timeoutScript -Timeout $Timeout
+    Write-Host "$ServiceName is now up."
+}
+  
+function Load-SupervisorService($PackageName, $Timeout = 1) {
+    $svcName = ($PackageName -split "/")[1]
+    $_ = Invoke-NativeCommand hab svc load $PackageName --url="$BuilderUrl"
+    Wait-SupervisorService $svcName -Timeout $Timeout
+    $svcName
+}
+  
+function Wait-SupervisorServiceUnload($ServiceName, $Timeout = 1) {
+    Write-Host "Waiting up to $Timeout seconds for Supervisor to unload $ServiceName ..."
+    $testScript = { 
+        try {
+            $_ = Invoke-WebRequest "http://localhost:9631/services/$ServiceName/default"
+            $false
+        }
+        catch {
+            $statusCode = $_.Exception.Response.StatusCode.value__
+            $statusCode -eq 404
+        }
+    }
+    $timeoutScript = { Write-Error "Timed out waiting $Timeout seconds for Supervisor to unload $ServiceName" }
+    Wait-True -TestScript $testScript -TimeoutScript $timeoutScript -Timeout $Timeout
+    Write-Host "$ServiceName is now unloaded."
+}
+
+function Unload-SupervisorService($PackageName, $Timeout = 1) {
+    $svcName = ($PackageName -split "/")[1]
+    $_ = Invoke-NativeCommand hab svc unload $PackageName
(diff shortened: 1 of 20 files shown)


.expeditor/scripts/shared.ps1
```diff
@@ -25,16 +35,15 @@ function Get-Toolchain {
...
 function New-PathString([string]$StartingPath, [string]$Path) {
-  if (-not [string]::IsNullOrEmpty($path)) {
+  if (-not [string]::IsNullOrEmpty($Path)) {
       if (-not [string]::IsNullOrEmpty($StartingPath)) {
-          [string[]]$PathCollection = "$path;$StartingPath" -split ';'
+          [string[]]$PathCollection = "$Path$env:PathSeparator$StartingPath" -split $env:PathSeparator
           $Path = ($PathCollection |
               Select-Object -Unique |
-              Where-Object {-not [string]::IsNullOrEmpty($_.trim())} |
-              Where-Object {test-path "$_"}
```
davidMcneil: Do we need to check if the path exists? This was throwing off adding `/hab/bin` to the `PATH`.

mwrock: no. we really don't need it

.expeditor/scripts/shared.ps1
```diff
@@ -1,3 +1,13 @@
+$env:PathSeparator = if ($IsWindows) {
```
mwrock: We want to make sure this works on "Windows Powershell" as well. So this should be `if ($IsWindows -Or !$IsCoreCLR)`

.expeditor/scripts/shared.ps1
```diff
@@ -1,3 +1,13 @@
...
+  ";"
+} else {
+  ":"
+}
+
+if (!$IsWindows) {
```
mwrock: I'd make this `if ($IsLinux -Or $IsMacOS)`

.expeditor/scripts/shared.ps1
```diff
@@ -72,15 +81,17 @@ function Install-RustToolchain($Toolchain) {
 function Wait-Supervisor($Timeout=1) {
   Write-Host "Waiting up to $Timeout seconds for Supervisor to start..."
   $startTime = [DateTime]::Now
-  $success = (Test-NetConnection -ComputerName localhost -Port 9631).TcpTestSucceeded
+  # Test-NetConnection could be used here, but it is not implemented in PowerShell core
+  # See https://github.com/PowerShell/PowerShell/issues/6932
+  $success = (Test-Connection -ComputerName localhost -TCPPort 9631).TcpTestSucceeded
```
mwrock: Unfortunately there is a `Test-Connection` in Windows powershell but it behaves slightly different. You could do something like:
```
If (Get-Command Test-NetConnection -ErrorAction SilentlyContinue) {
   Test-NetConnection ...
} else {
    Test-Connection ...
}
```

.expeditor/scripts/shared.ps1
```diff
@@ -72,15 +81,17 @@ function Install-RustToolchain($Toolchain) {
...
   while(!$success) {
       Start-Sleep -Seconds 1
       $timeTaken = [DateTime]::Now.Subtract($startTime)
       if($timeTaken.TotalSeconds -ge $Timeout) {
           Write-Error "Timed out waiting $Timeout seconds for Supervisor to start"
           break
       }
-      $success = (Test-NetConnection -ComputerName localhost -Port 9631).TcpTestSucceeded
+      $success = (Test-Connection -ComputerName localhost -TCPPort 9631).TcpTestSucceeded
```
mwrock: See above. We might want to extract the Test-Connection business to a separate function.

.expeditor/scripts/shared.ps1
davidMcneil: Because these scripts are only used for testing, do we gain much by supporting "Windows Powershell"? This specific instance is easy to fix, but we lose out on other nice to haves. For instance, `Join-Path`s ability to take multiple arguments, the `Test-Connection` discrepancy below, and using the `&` operator. I am sure there are others. We could certainly do without these, but if we could standardize on powershell core it would make writing these scripts cross platform easier.

mwrock: Both [build.ps1](https://github.com/habitat-sh/habitat/blob/master/build.ps1) and [e2e_local.ps1](https://github.com/habitat-sh/habitat/blob/master/e2e_local.ps1) which source shared.ps1 are part of a local windows developer workflow where one will just use the native ps console. I use them all the time.

mwrock: build.ps1 is basically a powershell make file

.expeditor/scripts/shared.ps1
```diff
@@ -69,22 +84,35 @@ function Install-RustToolchain($Toolchain) {
...
           break
       }
-      $success = (Test-NetConnection -ComputerName localhost -Port 9631).TcpTestSucceeded
+      $success = Test-Port -ComputerName localhost -Port 9631
   }
   if($success) { Write-Host "Supervisor is now running." }
 }
 
+function Start-Supervisor($Timeout=1) {
+  hab sup run --no-color &
```
mwrock: The tailing `&` really makes the windows tests unhappy. You might try something like the pattern in https://github.com/habitat-sh/habitat/blob/master/test/end-to-end/test_supervisor_load_service.ps1:

Start using:
```
$job = start-job { hab sup run }
```

Then:
```
    AfterAll {
        hab svc unload core/nginx
        Stop-Job -Job $job
        Remove-Job -Job $job
    }
```

.expeditor/scripts/shared.ps1
davidMcneil: I see. Maybe we can have a different "shared" script for the e2e tests or probably it should just be part of `run_e2e_test.ps1`? Would we be ok running the e2e test solely on powershell core? `shared.ps1` would stay completely "Windows Powershell" compatible, but developers of the e2e tests would not need to be concerned with the nuances.

.expeditor/scripts/shared.ps1
davidMcneil: Yeah, if we want our e2e tests to support "Windows Powershell" there will be a few places of cleanup, but I am hoping we can make the e2e tests solely target powershell core. See this [comment](https://github.com/habitat-sh/habitat/pull/7148#discussion_r342723106).

.expeditor/scripts/shared.ps1
mwrock: yeah that sounds like a good idea. We may need to have `run_e2e_test.ps1` load the powershell core package and launch pwsh (powershell core) because right now all of our tests call `run_e2e_test.ps1` from `powershell` (windows powershell).

.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
```diff
@@ -0,0 +1,126 @@
...
+        if (!(Test-Path -Path $Path)) {
+            return $False
+        }
+        (Get-Item -Path $Path).LastWriteTime -ge $Time
+    }
+    $timeoutScript = { Write-Error "Timed out waiting $Timeout seconds for '$Path' to be updated after '$($Time.ToString("yyyy-MM-ddTHH:MM:ssZ"))'" }
+    Wait-True -TestScript $testScript -TimeoutScript $timeoutScript -Timeout $Timeout
+}
+
+function Test-Port([string]$ComputerName, [int]$Port) {
```
mwrock: we should be able to do away with this abstraction now that we are all core

.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
```diff
@@ -0,0 +1,126 @@
...
+    $_ = hab svc load $PackageName
+    Wait-SupervisorService $svcName -Timeout $Timeout
+    $svcName
+}
+  
+function Wait-SupervisorServiceUnload($ServiceName, $Timeout = 1) {
+    Write-Host "Waiting up to $Timeout seconds for Supervisor to unload $ServiceName ..."
+    $testScript = { 
+        try {
+            Invoke-WebRequest "http://localhost:9631/services/$ServiceName/default"
```
mwrock: I think you will want to add a `$false` after the `Invoke-WebRequest` because if it succeeds it will return the response.

.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
```diff
@@ -0,0 +1,126 @@
...
+            $statusCode = $_.Exception.Response.StatusCode.value__
+            $statusCode -eq 404
+        }
+    }
+    $timeoutScript = { Write-Error "Timed out waiting $Timeout seconds for Supervisor to unload $ServiceName" }
+    Wait-True -TestScript $testScript -TimeoutScript $timeoutScript -Timeout $Timeout
+    Write-Host "$ServiceName is now unloaded."
+}
+
+function Uninstall-SupervisorService($PackageName, $Timeout = 1) {
```
mwrock: `unload` instead of `uninstall` since we are not actually uninstalling.

.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
davidMcneil: Oddly enough, `Test-Connection` does not seem to work on windows even in powershell core (it always returns false), but `Test-NetConnection` seems to work fine? Any suggestions?

.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
davidMcneil: Yeah, I was torn between using powershells accepted verbs and useing the habitat vocabulary. I will go ahead and switch it to match habitat.

mwrock: Glad to see you are trying to use the PS verbs! If we wanted to be anal about that here we could actually uninstall too but up to you.

.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
mwrock: weird. I just did some local testing and it does fail with `localhost` but passes with `127.0.0.1`

PR #7157, opened 2019-11-05
[CI] Switch from Acceptance to Production Builder

christophermaier:
This switches our new release pipeline from using our Acceptance Builder to using Production Builder. 

This basically makes this new process "real" for all intents and purposes.

We can't merge it as-is because we're waiting on one bit of new credentials in Vault, though. That'll happen real soon, though.

Among the things that could use particularly close review are the configurations of version bumping and changelog management. In particular, the version is now going to start at 0.90.0, and I've tried to consolidate all the "fake" changelog entries into the real changelog, with all the associated markup that Expeditor needs to manage it, but I would appreciate a second (and third, and fourth...) pair of eyes on all that.


.expeditor/end_to_end.pipeline.yml
```diff
@@ -391,7 +391,7 @@ steps:
         docker:
           privileged: true
           environment:
-            - ACCEPTANCE_HAB_AUTH_TOKEN
+            - PIPELINE_HAB_AUTH_TOKEN
             - HAB_BLDR_URL=https://bldr.acceptance.habitat.sh
```
smacfarlane: There are a few instances still referencing the acceptance environment in this pipeline definition

christophermaier: Yikes! :grimacing: 

Will fix!

smacfarlane:
@christophermaier Another use of `bldr.acceptance.habitat.sh` in `test/end-to-end/test_pkg_bulkupload.sh`,

.expeditor/end_to_end.pipeline.yml
```diff
@@ -380,19 +390,26 @@ steps:
...
+          path: account/static/habitat/chef-ci
+          field: scotthain-sig-key
       executor:
         docker:
           privileged: true
           environment:
-            - ACCEPTANCE_HAB_AUTH_TOKEN
-            - HAB_BLDR_URL=https://bldr.acceptance.habitat.sh
+            - PIPELINE_HAB_AUTH_TOKEN
+            - PIPELINE_HAB_BLDR_URL=http://bldr.acceptance.habitat.sh
```
jeremymv2: http or https?

christophermaier: Ugh... good catch on the `https`. Will fix!

And yes, I'm going to make sure the proper token is wired in for Acceptance before this is merged.

Issue #7150, opened 2019-11-04
[Emergent work] Customer seeing census data that does not match the state of running services


mwrock:
census data claims that service is on an older version than what it is actually running. Also in a `leader` topology they are occasionally seeing `no quorum` messages even when they have 2 services up.

mwrock:
Here are some oddities I am seeing with leader topology:

Working scenario:
- Start sup1 with svc1 and it waits for quorum
- Start sup2 with svc1 peered to sup1
- Both sups start svc1

Broken scenario 1:
- Start sup1 with svc1 on stand-alone topology
- svc1 starts
- Start sup2 with svc2 peered to sup1 using a leader topology
- svc2 starts
I expect svc2 to wait for a second svc2 to have quorum

Interesting (but not broken) scenario:
- Start sup1 with svc1 on stand-alone topology
- svc1 starts
- Start sup2 with svc2 peered to sup1 using a leader topology
- svc2 starts
- Start sup3 with svc2 peered to sup1 using a leader topology
- svc2 starts
- Blow away sup2 and 3 and note that the census sees them as NOT alive
- Start sup4 with svc2 peered to sup1 using a leader topology
- svc2 waits for quorum
- Start sup5 with svc2 peered to sup1 using a leader topology
- svc2 starts on sup4 and 5

PR #7159, opened 2019-11-06
filter out dead members when comparing a rolling leader version to followers

mwrock:
In a `rolling` update consider this scenario:

- You have a working and updating ring of supervisors.
- You eventually kill off all group members or some followers (Note: things go wrong in different ways when a leader dies and followers remain to be covered in a [different issue](https://github.com/habitat-sh/habitat/issues/7160)).
- You add new members or continue to maintain a working quorum
- The next rolling update will never complete because the leaser looks to see if all followers (alive or dead) have updated.

This changes the final point above to only compare live members.

This arose from #7150 


mwrock:
I'm closing this in favor of #7167

Issue #5325, opened 2018-07-12
Gossip when a service is stopped / unloaded


christophermaier:
When you shut down a service, whether with `hab svc unload` or `hab svc stop`, we need to gossip the fact that the Supervisor is no longer running that service.

We don't do this now, and so the only way for a service to "disappear" from the census is if its _Supervisor_ dies, because we have a bit of a conflation of Supervisor presence and Service presence. If you're running on "disposable infrastructure", this may be less of a problem, but can be very surprising / puzzling / dangerous otherwise.

We should confirm that once you `hab svc start` a stopped service re-load a previously unloaded service that we still send out the appropriate "service added" rumors.

jamessewell:
I just got stung by this - I'd been working on the assumption that when I stop a service it will be removed from my templated config files!

Any progress?

christophermaier:
@jamessewell Not yet, but we hope to get to this soon.

username-is-already-taken2:
Hi there, I came across this issue today, where my nginx config has stale members in it config.  How are people working around this at the moment? what do I have to do to refresh my census data after I unload a svc to reflect the current state of play

PR #7167, opened 2019-11-06
Fix rolling updates after the death of group members

### LINKED ISSUE #7160: Rolling updates likely break after a leader is removed from the group
Rolling updates will likely break after a leader is removed. From my observations, removing a `leader` topology leader will have one of the following effects upon a rolling update:

1. Everything is fine. This assumes #7159 is merged
1. If there were only 2 members and you lose quorum after killing the current leader AND a newly added member becomes the leader AND that member has an older version of the service on disk (likely if a docker exported image) THEN the follower will try to "update" itself to the leader's version which may likely be less than its current version. This results in a continuous looping of updating to the lower version, starting as the higher version and then "updating" to the previous version again and again and again and again.
1. If an existing follower becomes the new leader, even though according to the census data it appears to reflect its new leadership role, its internal update `state` continues to reflect `UpdaterState::Rolling(RollingState::Follower(ref mut state))` and thus acts like a follower so everyone is following and no one is leading and therefore no one is updating.

This arose from #7150

mwrock:
fixes #7160

Note that sadly this fix only addresses when supervisors leave/restart and not when a service is simply unloaded. That is addressed in #5325.


test/end-to-end/multi-supervisor/testcases/rolling_update/run_test.sh
```diff
@@ -0,0 +1,51 @@
...
+
+hab pkg promote ${test_ident} "${test_channel}"
+
+hab svc load habitat-testing/nginx --topology leader --strategy rolling --channel "${test_channel}" --remote-sup=alpha.habitat.dev
+hab svc load habitat-testing/nginx --topology leader --strategy rolling --channel "${test_channel}" --remote-sup=beta.habitat.dev
+
+cleanup () {
+    hab bldr channel destroy "${test_channel}" --origin habitat-testing
+}
+
```
christophermaier: We should set up a signal trap that ensures that we run this whether the test passes or not, e.g., 

```sh
trap cleanup INT TERM EXIT
```
(You could then remove the calls to `cleanup` when you test to see if the correct services are running, and at the very bottom of the script.)

test/end-to-end/multi-supervisor/testcases/rolling_update_remove_follower_keep_quorum/run_test.sh
```diff
@@ -0,0 +1,93 @@
...
+hab pkg promote ${test_ident} "${test_channel}"
+
+for server in alpha beta gamma; do
+    hab svc load habitat-testing/nginx --topology leader --strategy rolling --channel "${test_channel}" --remote-sup=${server}.habitat.dev
+done
+
+cleanup () {
+    hab bldr channel destroy "${test_channel}" --origin habitat-testing
+}
+
```
christophermaier: Same `trap` suggestion applies here, too.

test/end-to-end/multi-supervisor/testcases/rolling_update_remove_follower_keep_quorum/run_test.sh
```diff
@@ -0,0 +1,93 @@
...
+        if [[ "${current_ident}" != "${test_ident}" ]]; then
+            echo "Expected nginx ident ${test_ident} on ${server}; got ${current_ident} instead!"
+            cleanup
+            exit 1
+        fi
+    fi
+done
+
+# update again
+test_ident=habitat-testing/nginx/1.17.4/20191115185900
```
christophermaier: Since we've got 3 different packages, naming them like `test_ident_v1`, `test_ident_v2`, etc. might be better (and more general) than my previous suggestion of `initial_release` and `updated_release`.

test/end-to-end/multi-supervisor/testcases/rolling_update_remove_leader_keep_quorum/run_test.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+# This tests that removing the leader from a functioning leader topology
+# service group that has enough nodes to maintain quorum after the leader is
+# lost, it will continue to perform a succesful roaming update after a new
```
christophermaier: "roaming" -> "rolling"

test/end-to-end/multi-supervisor/testcases/rolling_update_remove_leader_lose_quorum/run_test.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# This tests that removing the leader from a functioning 2 node leader topology
+# service group will continue to perform a succesful roaming update after a new
```
christophermaier: "roaming" -> "rolling"

test/end-to-end/multi-supervisor/testcases/rolling_update_remove_leader_lose_quorum/run_test.sh
```diff
@@ -0,0 +1,83 @@
...
+        cleanup
+        exit 1
+    fi
+done
+
+body=$(curl -s "bastion.habitat.dev:9631/census")
+leader_id=$(echo "${body}" | jq -r ".census_groups.\"nginx.default\".leader_id")
+leader_name=$(echo "${body}" | jq -r ".census_groups.\"nginx.default\".population.\"${leader_id}\".sys.hostname")
+docker exec "${COMPOSE_PROJECT_NAME}_${leader_name}_1" hab sup term
+docker exec "${COMPOSE_PROJECT_NAME}_gamma_1" hab pkg install habitat-testing/nginx/1.17.4/20191115184838
```
christophermaier: Naming the identifiers symbolically would really come in handy here.

test/end-to-end/multi-supervisor/testcases/rolling_update_remove_leader_lose_quorum/run_test.sh
```diff
@@ -0,0 +1,83 @@
...
+
+test_ident=habitat-testing/nginx/1.17.4/20191115185900
+hab pkg promote ${test_ident} "${test_channel}"
+sleep 15
+
+for server in alpha beta gamma; do
+    if [[ "${server}" != "${leader_name}" ]]; then
+        current_ident=$(curl -s ${server}.habitat.dev:9631/services/nginx/default | jq -r '.pkg.ident')
+        if [[ "${current_ident}" != "${test_ident}" ]]; then
+            echo "Expected nginx ident ${test_ident} on ${server}; got ${current_ident} instead!"
```
christophermaier: For these "failed expectation" messages, it could aid debugging to make them all unique (e.g., to incorporate a bit more context about the behavior we're trying to assert at each point). 

That way if we get a failure, we can more quickly pinpoint where things are going wrong, both in the test and (hopefully) in the underlying code.

sirajrauff:
Just wanted to comment on some tests I ran against this PR using `core/hab-sup` from the rolling channel, specifically version `0.90.50/20191217193828`:
```shell script
$ systemctl status hab-sup
● hab-sup.service - Habitat Supervisor
   Loaded: loaded (/etc/systemd/system/hab-sup.service; disabled; vendor preset: disabled)
   Active: active (running) since Wed 2019-12-18 19:55:49 UTC; 3min 30s ago
     Docs: https://habitat.sh
 Main PID: 374 (hab-launch)
   CGroup: /docker/2cd1e104143ff2e7fdf43e4883876da78337b2bde1697485904928ff0c0cb1a7/system.slice/hab-sup.service
           ├─374 /hab/pkgs/core/hab-launcher/12605/20191112144831/bin/hab-launch run
           └─379 /hab/pkgs/core/hab-sup/0.90.50/20191217193828/bin/hab-sup run

Dec 18 19:55:49 2cd1e104143f systemd[1]: Started Habitat Supervisor.
...
```
I ran a few tests and it seems to have solved the problems I spoke of in issue [7160](https://github.com/habitat-sh/habitat/issues/7160#issuecomment-566253714), with a few nuances.

1. When a single Supervisor in a ring has a service loaded with strategy `rolling`, the service will update. This is effectively `at-once`, but is an important case because the service will always be loaded _first_ on a single server.

    When the service is then loaded on a second supervisor, the vote does not seem to occur again (according to `/butterfly`'s `last_election_update`), and the first node remains the update leader and both services will update. From here we have two scenarios;

    1. If the second node is now shut down, the vote does _not_ re-occur and the initial node (still the update leader) will continue to update.
    2. If the update leader is shut down, the vote re-occurs with both nodes voting for the _current_ update leader. In this scenario the remaining node is not an update leader and will never update.

    This can be reproduced with the following steps:
    1. Start three Supervisors, one a peer (i'll ignore this) and Supervisor A and Supervisor B.
    2. Load a service on Supervisor A with the `rolling` strategy. This will appear as the update leader in census and butterfly. Try promoting a package; this service will update.
    3. Load the same service on Supervisor B with the `rolling` strategy. Check census/butterfly; no election occurs. Try promoting a package; both service instances update.
    4. Stop Supervisor B. Push an update; Supervisor A's service instance will update.
    5. Start Supervisor B, the service instance will update to the match Supervisor A's service version. Checking census/butterfly no election has occurred.
    6. Stop Supervisor A. Check census/butterfly, an election has occurred with two votes for Supervisor A. Push an update; Supervisor B will not update.

2. Sometimes `/census` & `/butterfly` don't seem to correctly update. i.e. the update leader leaves & is re-assigned (or at least appears so, as updates still work) but the previous, departed update-leader is still considered the leader according to `/census` & `/butterfly`.

    Note this might have to do something with Quorum as I can't reproduce this when going from 4 nodes to 3, but I can when going from 3 to 2.

    Reproducing:
    1. Start 5 nodes, first used just to peer and then A-D. Load a service with `rolling` strategy on A through D:
    ```shell script
    $ curl -s localhost:9631/census | jq '.census_groups["sample-node-app.default"].population[] | { member_id: .member_id, version: ([.pkg.version, .pkg.release] | join("/")), ip: .sys.ip }'
    {
      "member_id": "3c636da889bb44d89a41ecece2c96ee7",
      "version": "0.1.0/20191216195328"
    }
    {
      "member_id": "6730593de5b24f62a5a693ee56cf4eba",
      "version": "0.1.0/20191216195328"
    }
    {
      "member_id": "c7c993780aec49ce898bee94a72d31d0",
      "version": "0.1.0/20191216195328"
    }
    {
      "member_id": "dc2822b15cc940d1a207b0261209d7c7",
      "version": "0.1.0/20191216195328"
    }

    $ curl -s localhost:9631/butterfly | jq .latest_election_update
    {
      "sample-node-app.default": {
        "member_id": "c7c993780aec49ce898bee94a72d31d0",
        "service_group": "sample-node-app.default",
        "term": 0,
        "suitability": 0,
        "status": "Finished",
        "votes": [
          "c7c993780aec49ce898bee94a72d31d0"
        ]
      }
    }

    $ curl -s localhost:9631/butterfly | jq .member.health
    {
      "3c636da889bb44d89a41ecece2c96ee7": "Alive",
      "5522f2438a234f2393e4b4ec99329662": "Alive",
      "6730593de5b24f62a5a693ee56cf4eba": "Alive",
      "c7c993780aec49ce898bee94a72d31d0": "Alive",
      "dc2822b15cc940d1a207b0261209d7c7": "Alive"
    }
    ```

    2. Stop the current Update leader:
    ```shell script
    $ more /hab/sup/default/MEMBER_ID
    c7c993780aec49ce898bee94a72d31d0

    $ systemctl stop hab-sup
    ```
    Which causes a new election:
    ```shell script
    $ curl -s localhost:9631/butterfly | jq .member.health
    {
      "6730593de5b24f62a5a693ee56cf4eba": "Alive",
      "5522f2438a234f2393e4b4ec99329662": "Alive",
      "dc2822b15cc940d1a207b0261209d7c7": "Alive",
      "3c636da889bb44d89a41ecece2c96ee7": "Alive",
      "c7c993780aec49ce898bee94a72d31d0": "Departed"
    }

    $ curl -s localhost:9631/butterfly | jq .latest_election_update
    {
      "sample-node-app.default": {
        "member_id": "dc2822b15cc940d1a207b0261209d7c7",
        "service_group": "sample-node-app.default",
        "term": 1,
        "suitability": 0,
        "status": "Finished",
        "votes": [
          "dc2822b15cc940d1a207b0261209d7c7",
          "c7c993780aec49ce898bee94a72d31d0",
          "3c636da889bb44d89a41ecece2c96ee7",
          "6730593de5b24f62a5a693ee56cf4eba"
        ]
      }
    }
    ```

    3. Depart the new update leader:
    ```shell script
    $ more /hab/sup/default/MEMBER_ID
    dc2822b15cc940d1a207b0261209d7c7

    $ systemctl stop hab-sup
    ```
    Which does not reflect in `/census` or `/butterfly`:
    ```shell script
    $ curl -s localhost:9631/butterfly | jq .latest_election_update
    {
      "sample-node-app.default": {
        "member_id": "dc2822b15cc940d1a207b0261209d7c7",
        "service_group": "sample-node-app.default",
        "term": 1,
        "suitability": 0,
        "status": "Finished",
        "votes": [
          "dc2822b15cc940d1a207b0261209d7c7",
          "c7c993780aec49ce898bee94a72d31d0",
          "3c636da889bb44d89a41ecece2c96ee7",
          "6730593de5b24f62a5a693ee56cf4eba"
        ]
      }
    }

    $ curl -s localhost:9631/butterfly | jq .member.health
    {
      "c7c993780aec49ce898bee94a72d31d0": "Departed",
      "6730593de5b24f62a5a693ee56cf4eba": "Alive",
      "5522f2438a234f2393e4b4ec99329662": "Alive",
      "3c636da889bb44d89a41ecece2c96ee7": "Alive",
      "dc2822b15cc940d1a207b0261209d7c7": "Departed"
    }
    ```
    however updates still work:
    ```shell script
    $ hab pkg promote sirajrauff-public/sample-node-app/0.1.1/20191218174803 stable
    » Promoting sirajrauff-public/sample-node-app/0.1.1/20191218174803 (x86_64-linux) to channel 'stable'
    ✓ Promoted sirajrauff-public/sample-node-app/0.1.1/20191218174803 (x86_64-linux)
    ```
    ```shell script
    $ curl -s localhost:9631/census | jq '.census_groups["sample-node-app.default"].population[] | { member_id: .member_id, version: ([.pkg.version, .pkg.release] | join("/"))}'
    {
      "member_id": "3c636da889bb44d89a41ecece2c96ee7",
      "version": "0.1.1/20191218174803"
    }
    {
      "member_id": "6730593de5b24f62a5a693ee56cf4eba",
      "version": "0.1.1/20191218174803"
    }
    {
      "member_id": "c7c993780aec49ce898bee94a72d31d0",
      "version": "0.1.0/20191216195328"
    }
    {
      "member_id": "dc2822b15cc940d1a207b0261209d7c7",
      "version": "0.1.0/20191216195328"
    }
    ```
Sorry for the wall of text.

mwrock:
Addressed all your feedback @cm. Thanks so much. I agree that the tests could use some DRYing up but was going to wait until I converted them to powershell before doing that.

mwrock:
I just converted the e2e tests to powershell and made them a bit more reusable.

PR #7168, opened 2019-11-07
[CI] Temporarily skip the bulkupload e2e test pending some refactoring

christophermaier:
This is solely to get the pipeline flowing again; a fix will be in later today.


christophermaier:
@smacfarlane :thinking: While I like the "angry red" aspect as a reminder, I was thinking of skip because it allows us to add metadata about why we're skipping. Also, I wasn't really thinking of this as a potentially flaky test, since this failure is really an aspect of how the test is integrated into the pipeline, rather than something in the logic of the tests _per se_.

PR #7172, opened 2019-11-07
[CI] Purge Fastly cache on promotion to stable only

christophermaier:


christophermaier:
@smacfarlane :thinking: Must have missed that somehow... let me ponder this a bit.

Issue #7162, opened 2019-11-06
Implement new hab pkg download file format


markan:
In #7057 we seem to have come to a consensus around the file format (see below)

Implement it

```
# if not specified, defaults to latest version understood by habitat.
format_version = 1
# to be rendered during execution and in potential manifests
file_descriptor = "user specified data" 

# Target grouping. Channel defaults to stable
[["x86_64-linux"]]
channel = "stable"
packages = ["core/foo", "core/bar/3.2", "core/baz/2.71/20190910181446"]

[["x86_64-linux"]]
channel = "unstable"
packages = ["core/crazypants", "core/cliffs_of_insanity/0.0.1/20191010181446"]

[["x86_64-windows"]]
channel = "stable"
packages = ["effortless/audit-baseline"]
```

PR #7176, opened 2019-11-08
Implement toml based file format

markan:
This is a first cut of the toml file format proposed in issue #7162

Ideally this would be implemented via TOML deserialization into a rust
data structure, but I couldn't get that to work, because of issues
with the keys being 'PackageTarget' structures. I'll make another go
at it, perhaps with an intermediate structure that matches the file
structure better, but I thought this might be worthwhile for UX review
and the like.

```
format_version = 1
file_descriptor = "user specified data"

[["x86_64-linux"]]
channel = "stable"
packages = ["core/foo", "core/bar/3.2", "core/baz/2.71/20190910181446"]

[["x86_64-linux"]]
channel = "unstable"
packages = ["core/crazypants", "core/cliffs_of_insanity/0.0.1/20191010181446"]

[["x86_64-windows"]]
channel = "stable"
packages = ["effortless/audit-baseline"]
```

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -5,6 +5,7 @@ use clap::{App,
            Arg,
            ArgMatches};
 use habitat_common::{cli::{file_into_idents,
+                           is_toml_file,
                            BINLINK_DIR_ENVVAR,
                            DEFAULT_BINLINK_DIR,
                            PACKAGE_TARGET_ENVVAR,
@@ -940,14 +941,14 @@ fn sub_pkg_download() -> App<'static, 'static> {
         "Specify an alternate Builder endpoint. If not specified, the value will \
          be taken from the HAB_BLDR_URL environment variable if defined.")
     (@arg CHANNEL: --channel -c +takes_value default_value[stable] env(ChannelIdent::ENVVAR)
-        "Download from the specified release channel")
+        "Download from the specified release channel. Overridden if channel is specified in toml file.")
     (@arg DOWNLOAD_DIRECTORY: --("download-directory") +takes_value "The path to store downloaded artifacts")
-    (@arg PKG_IDENT_FILE: --file +takes_value +multiple {valid_ident_file}
-        "File with newline separated package identifiers")
+    (@arg PKG_IDENT_FILE: --file +takes_value +multiple {valid_ident_or_toml_file}
+        "File with newline separated package identifiers, or TOML file (ending with .toml extension)")
     (@arg PKG_IDENT: +multiple {valid_ident}
             "One or more Habitat package identifiers (ex: acme/redis)")
     (@arg PKG_TARGET: --target -t +takes_value {valid_target}
-            "Target architecture to fetch. E.g. x86_64-linux")
+            "Target architecture to fetch. E.g. x86_64-linux. Overridden if architecture is specified in toml file.")
     (@arg VERIFY: --verify
             "Verify package integrity after download (Warning: this can be slow)")
     );
@@ -1455,6 +1456,17 @@ fn valid_ident(val: String) -> result::Result<(), String> {
                          form origin/name[/version[/release]]",
                         &val))
         }
+    }
+}
+
+#[allow(clippy::needless_pass_by_value)] // Signature required by CLAP
+fn valid_ident_or_toml_file(val: String) -> result::Result<(), String> {
+    if is_toml_file(&val) {
+        // We could do some more validation (parse the whole toml file and check it) but that seems
+        // excessive.
+        Ok(())
+    } else {
+        valid_ident_file(val)
     }
 }
 
--- a/components/hab/src/command/pkg/download.rs
+++ b/components/hab/src/command/pkg/download.rs
@@ -27,7 +27,8 @@
 //! * Verify it is un-altered
 //! * Fetch the signing keys
 
-use std::{collections::HashSet,
+use std::{collections::{HashMap,
+                        HashSet},
           fs::DirBuilder,
           path::{Path,
                  PathBuf},
@@ -50,6 +51,10 @@ use crate::{api_client::{self,
                     ChannelIdent,
                     Error as CoreError}};
 
+use habitat_common::ui::{Glyph,
+                         Status,
+                         UIWriter};
+
 use reqwest::StatusCode;
 use retry::{delay,
             retry};
@@ -57,12 +62,33 @@ use retry::{delay,
 use crate::error::{Error,
                    Result};
 
-use habitat_common::ui::{Glyph,
-                         Status,
-                         UIWriter};
-
 pub const RETRIES: usize = 5;
 pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
+
+#[derive(Debug, Deserialize)]
+pub struct PackageSetFile {
+    pub format_version:  Option<u8>,
+    pub file_descriptor: Option<String>,
+
+    // Note: flatten cause *all* keys not explicitly listed as a struct member to be treated as a
+    // target. If someone gives a bad key, the error message will be incorrect as to the cause.
+    // This is a limitation of the target as keys choice for the TOML file format.
+    #[serde(flatten)]
+    pub targets: HashMap<PackageTarget, Vec<PackageSetValue>>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct PackageSetValue {
+    pub channel:  ChannelIdent,
+    pub packages: Vec<String>,
+}
+
+#[derive(Debug)]
+pub struct PackageSet {
+    pub target:  PackageTarget,
+    pub channel: ChannelIdent,
+    pub idents:  Vec<PackageIdent>,
+}
 
 /// Download a Habitat package.
 ///
@@ -84,35 +110,37 @@ pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
 #[allow(clippy::too_many_arguments)]
 pub fn start<U>(ui: &mut U,
                 url: &str,
-                channel: &ChannelIdent,
                 product: &str,
                 version: &str,
-                idents: Vec<PackageIdent>,
-                target: PackageTarget,
+                package_sets: &[PackageSet], // clippy suggestion
                 download_path: Option<&PathBuf>,
                 token: Option<&str>,
                 verify: bool)
                 -> Result<()>
     where U: UIWriter
 {
-    debug!("Starting download with url: {}, channel: {}, product: {}, version: {}, target: {}, \
-            download_path: {:?}, token: {:?}, verify: {}, ident_count: {}",
+    debug!(
+           "Starting download with url: {}, product: {}, version: {}, 
+         download_path: {:?}, token: {:?}, verify: {}, set_count: {}",
            url,
-           channel,
            product,
            version,
-           target,
            download_path,
            token,
            verify,
-           idents.len());
+           package_sets.len()
+    );
 
     let download_path_default = &cache_root_path::<PathBuf>(None); // Satisfy E0716
     let download_path_expanded = download_path.unwrap_or(download_path_default).as_ref();
     debug!("Using download_path {:?} expanded to {:?}",
            download_path, download_path_expanded);
 
-    if idents.is_empty() {
+    debug!("{} {}",
+           package_sets.is_empty(),
+           package_sets[0].idents.is_empty());
+
+    if package_sets.is_empty() {
         ui.fatal("No package identifers provided. Specify identifiers on the command line, or \
                   via a input file")?;
         return Err(CommonError::MissingCLIInputError(String::from("No package identifiers \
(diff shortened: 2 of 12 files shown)


components/hab/src/cli.rs
```diff
@@ -941,6 +941,8 @@ fn sub_pkg_download() -> App<'static, 'static> {
     (@arg DOWNLOAD_DIRECTORY: --("download-directory") +takes_value "The path to store downloaded artifacts")
     (@arg PKG_IDENT_FILE: --file +takes_value +multiple {valid_ident_file}
         "File with newline separated package identifiers")
+    (@arg PKG_TOML_FILE: --toml +takes_value +multiple // {valid_toml_file}
```
chefsalim: It might be better to just have a single --file param... try a TOML parse first, if that fails, fall back to try a plain text format parse?  That would avoid having an extra param and also retain backward compat.

components/hab/src/cli.rs
```diff
@@ -1261,74 +1263,114 @@ fn sub_svc_unload() -> App<'static, 'static> {
 fn add_event_stream_options(app: App<'static, 'static>) -> App<'static, 'static> {
     // Create shorter alias so formating works correctly
     type ConnectMethod = EventStreamConnectMethod;
-    app.arg(Arg::with_name("EVENT_STREAM_APPLICATION").help("The name of the application for \
```
chefsalim: The event stream stuff seems unrelated - maybe redact that out of this PR?

components/hab/src/command/pkg/download.rs
```diff
@@ -84,50 +92,47 @@ pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
...
     debug!("Using download_path {:?} expanded to {:?}",
            download_path, download_path_expanded);
 
-    if idents.is_empty() {
+    if package_sets.is_empty() {
         ui.fatal("No package identifers provided. Specify identifiers on the command line, or \
                   via a input file")?;
         return Err(CommonError::MissingCLIInputError(String::from("No package identifiers \
-                                                                   found")).into());
+                                                                   found found")).into());
```
chefsalim: Typo?

components/hab/src/command/pkg/download.rs
```diff
@@ -245,9 +254,10 @@ impl<'a> DownloadTask<'a> {
 
         // We rely on the fact that all packages are always added to the unstable channel as well as
         // any other channel specified.
-        let mut channel = &ChannelIdent::unstable();
-        if !ident.fully_qualified() {
-            channel = self.channel
+        let channel = if ident.fully_qualified() {
+            ChannelIdent::unstable()
+        } else {
+            (*channel).clone()
```
chefsalim: I wonder if we can avoid doing a clone..

components/hab/src/error.rs
```diff
@@ -157,6 +158,9 @@ impl fmt::Display for Error {
                 format!("Package archive was unreadable or contained unexpected contents: {:?}",
                         e)
             }
+            Error::PackageSetParseError(ref e) => {
+                format!("Package set file was ill formed: {:?}", e)
```
chefsalim: Maybe rephrase to say "Package set file could not be parsed"?

components/hab/src/main.rs
```diff
@@ -12,6 +12,8 @@ use clap::{ArgMatches,
 use env_logger;
 use futures::prelude::*;
 use glob::glob_with;
+use hab::command::pkg::download::PackageSet;
+// use hab::error::Error::PackageSetParseError;
```
chefsalim: Remove comment

components/hab/src/main.rs
```diff
@@ -1090,7 +1102,7 @@ fn sub_svc_set(m: &ArgMatches<'_>) -> Result<()> {
     // SrvClient from a for_each iterator so we can chain upon a successful stream but I don't
     // know if it's possible with this version of futures.
     SrvClient::connect(&listen_ctl_addr, &secret_key).and_then(|conn| {
-                                                         conn.call(set).for_each(|reply| {
```
chefsalim: Redact the unrelated formatting

components/hab/src/main.rs
```diff
@@ -1676,6 +1686,109 @@ fn idents_from_file_matches(matches: &ArgMatches<'_>) -> Result<Vec<PackageIdent
     Ok(sources)
 }
 
+fn idents_from_toml_matches(matches: &ArgMatches<'_>) -> Result<Vec<PackageSet>> {
```
chefsalim: Somewhat concerned that we have to do some much manual construction of the package set from the TOML...

christophermaier: I haven't dug into this PR yet (this is more of a drive-by comment at this point), but I notice that there isn't an implementation of `Deserialize` for `PackageSet`. Having that would probably simplify a fair bit of this logic, if not remove it altogether.

davidMcneil: Second, simplifying with `Deserialize`. Here is a stab at the types:

```rust
#[derive(Debug, Deserialize)]
pub struct PackageSetFile {
    pub format_version:  u8,
    pub file_descriptor: String,
    pub package_sets:    HashMap<PackageTarget, Vec<PackageSetValue>>,
}

#[derive(Debug, Deserialize)]
pub struct PackageSetValue {
    pub channel: ChannelIdent,
    pub idents:  Vec<PackageIdent>,
}

#[derive(Debug, Deserialize)]
pub struct PackageSet {
    pub target:  PackageTarget,
    pub channel: ChannelIdent,
    pub idents:  Vec<PackageIdent>,
}
```

This will lead to needing a `package_sets` key so the file format will need to use something like `[["package_sets.x86_64-linux"]]`. The other option would be to use some macro magic to add a key for each `PackageTarget` variant

markan: @davidMcneil I did try something like the above, but hit some snags around PackageTarget being a key, and the already defined serializer for PackageIdent. Working on some ideas now.

davidMcneil: @markan serde's field attribute [`#[serde(deserialize_with = "path")]`](https://serde.rs/field-attrs.html#deserialize_with) might help if we need a custom deserializer?

apriofrost:
Had a discussion with @markan , here are my suggestions summarized:

1. I agree that we should try to make the TOML format work with the `--file` option we have now without adding a `--toml` option, for the reason that we are just adding another file format in addition to the previous newline-separated file format that serves the same purpose: to give instructions on what packages to download. 
2. I'd suggest renaming the `format_version` to be `file_format_version`, so that it becomes clearer that the format is for this file.
3. We found that the current `--target` option is optional, and the default value is the target of the machine that initiated the download if users do not use the --file option with the TOML format that specifies the targets. If users use the previous newline-separated file format, they can use the --target option to specify the target that is different from the current host used for download. There fore, we should update the description of the `--target` option so that it is clearer, for example:
`
-t, --target <PKG_TARGET>         Target architecture to fetch. E.g. x86_64-linux. (default: the target of the current host if no target is specified in <PKG_IDENT_FILE> using the --file option) 
`


@kagarmoe do you have suggestions for how we describe the option above? I'd really appreciate your thoughts on it!

chefsalim:
One other comment - when this is ready for re-review it would be good to squash the (WIP) commits, and also add unit/end-to-end tests for the new format as well.

components/hab/src/command/pkg/download.rs
```diff
@@ -234,8 +264,9 @@ impl<'a> DownloadTask<'a> {
 
     fn determine_latest_from_ident<T>(&self,
                                       ui: &mut T,
-                                      ident: &PackageIdent,
-                                      target: PackageTarget)
+                                      _channel: &ChannelIdent,
```
chefsalim: Minor - underscore is a pattern for an unused variable, so would be good to change to use a non-underscore channel var

components/hab/src/main.rs
```diff
@@ -1645,19 +1656,62 @@ fn idents_from_matches(matches: &ArgMatches<'_>) -> Result<Vec<PackageIdent>> {
...
     Ok(sources)
 }
 
+fn idents_from_toml_file(filename: &str) -> Result<Vec<PackageSet>> {
+    let mut sources: Vec<PackageSet> = Vec::new();
+
+    let file_data = std::fs::read_to_string(filename)?;
+    let toml_data: PackageSetFile =
+        toml::from_str(&file_data).map_err(habitat_common::Error::TomlParser)?;
+
```
chefsalim: Should we be checking format_version here?

components/hab/src/command/pkg/download.rs
```diff
@@ -50,20 +51,45 @@ use crate::{api_client::{self,
...
-                         Status,
-                         UIWriter};
-
 pub const RETRIES: usize = 5;
 pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
 
+#[derive(Debug, Deserialize)]
+pub struct PackageSetFile {
+    pub format_version:  Option<u8>,
+    pub file_descriptor: Option<String>,
```
chefsalim: Not sure what the purpose of the file_descriptor field is as we are not checking it?

markan: It was suggested as part of making the flow of package inclusion traceable when generating the manifest. We've not started that work, but I left the optional field in to save a version rev when we do it.

chefsalim: Ok that seems reasonable as long as we're planning to do that work and the design is worked out - if it's just a suggestion that may or may not come to fruition in current form I'd lean toward taking it out for now.

components/hab/src/command/pkg/download.rs
markan: Thanks, renamed variable.

components/hab/src/main.rs
markan: Implemented

components/hab/src/command/pkg/download.rs
markan: Fixed

components/hab/src/cli.rs
```diff
@@ -1458,6 +1459,17 @@ fn valid_ident(val: String) -> result::Result<(), String> {
     }
 }
 
+#[allow(clippy::needless_pass_by_value)] // Signature required by CLAP
+fn valid_ident_or_toml_file(val: String) -> result::Result<(), String> {
+    if is_toml_file(&val) {
+        // We could do some more validation (parse the whole toml file and check it) but that seems
+        // excessive.
```
davidMcneil: I like this comment in `is_toml_file`, but I dont think it is necessary to duplicate it here.

components/hab/src/command/pkg/download.rs
```diff
@@ -84,35 +110,37 @@ pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
 #[allow(clippy::too_many_arguments)]
 pub fn start<U>(ui: &mut U,
                 url: &str,
-                channel: &ChannelIdent,
                 product: &str,
                 version: &str,
-                idents: Vec<PackageIdent>,
-                target: PackageTarget,
+                package_sets: &[PackageSet], // clippy suggestion
```
davidMcneil: Taking a slice is definitely better, I would maybe remove the clippy comment.

components/hab/src/main.rs
```diff
@@ -11,10 +11,13 @@ use clap::{ArgMatches,
            Shell};
 use env_logger;
 use futures::prelude::*;
+
```
davidMcneil: if we keep our imports together rustfmt will automatically sort them

components/hab/src/main.rs
```diff
@@ -27,6 +30,7 @@ use hab::{cli::{self,
           ORIGIN_ENVVAR,
           PRODUCT,
           VERSION};
+
```
davidMcneil: if we keep our imports together rustfmt will automatically sort them

components/hab/src/main.rs
```diff
@@ -1645,19 +1656,73 @@ fn idents_from_matches(matches: &ArgMatches<'_>) -> Result<Vec<PackageIdent>> {
...
+        return Err(Error::PackageSetParseError(format!(
+            "format_version invalid, only version 1 allowed ({} provided",
+            toml_data.format_version.unwrap()
+        )));
+    }
+
+    ui.status(Status::Using,
+              format!("File {}, '{}'",
+                      filename,
+                      toml_data.file_descriptor.unwrap_or_else(|| "".to_string())))?;
```
davidMcneil: We could use `unwrap_or_default` here

components/hab/src/main.rs
```diff
@@ -1645,19 +1656,73 @@ fn idents_from_matches(matches: &ArgMatches<'_>) -> Result<Vec<PackageIdent>> {
...
 }
 
-fn idents_from_file_matches(matches: &ArgMatches<'_>) -> Result<Vec<PackageIdent>> {
-    let mut sources: Vec<PackageIdent> = Vec::new();
+fn idents_from_file_matches(ui: &mut UI,
+                            matches: &ArgMatches<'_>,
+                            cli_channel: &ChannelIdent,
+                            cli_target: PackageTarget)
+                            -> Result<Vec<PackageSet>> {
+    let mut sources: Vec<PackageSet> = Vec::new();
```
davidMcneil: Generally, it is more idomatic to leave off type definitions if possible.

Issue #7058, opened 2019-10-17
Clean up Makefile and Contributing / Local Development Instructions


christophermaier:
There's a lot of historical cruft in our `Makefile` and contributor / local development instructions that make it very difficult to get started with this codebase.

We need to revisit all of it, clean it up, and make sure that it can actually be used by someone that's not already a core team member.

Bonus points for lots of automation as well as continuous integration tests that ensure it stays sane.

christophermaier:
Also interesting: `make test` doesn't actually run the tests of `components/core`. `cargo test` does, of course

PR #7179, opened 2019-11-10
make: fix fmt target

stevendanna:
Without this the `fmt` target uses whatever toolchain happens to be
active, which might yield substantially different formatting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,9 +213,10 @@ clean-$1: image ## cleans the $1 component's project tree
 endef
 $(foreach component,$(ALL),$(eval $(call CLEAN,$(component))))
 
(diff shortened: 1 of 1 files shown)


stevendanna:
> It should be noted that this does not format all rust code only the components defined as a BIN or LIB. Because of this, I use cargo +<current-fmt-version> fmt in the top-level habitat folder which will format all components in the workspace.

@davidMcneil Feel free to push such a change to this branch if that reflects more of what we want people running.

davidMcneil:
> @davidMcneil Feel free to push such a change to this branch if that reflects more of what we want people running.

I think this is the right change for now. As part of #7058, I could see making further changes. Thanks for fixing this!

PR #7180, opened 2019-11-10
Fixes Sass compiling

kagarmoe:
The SCSS hasn't been compiling correctly.  Middleman 4 switched to asset-pipelines, although Middleman is supposed to support SCSS out-of-the-box. 
I took a kitchen-sink approach to making this run. The working solution involves several outdated and unsupported dependencies. Let's hope this is the wrong answer and we cand find something better. The next steps are obviously subtractive.

Consider using webpack: https://rossta.net/blog/using-webpack-with-middleman.html


kagarmoe:
Wow. These tests should not have passed. We need to look into that.

PR #7204, opened 2019-11-14
[CI] Get the correct auth token into the bulkupload test

christophermaier:
I think the previous setup wasn't actually layering things the way I
was thinking.


christophermaier:
Closing in favor of #7203

PR #7206, opened 2019-11-15
Update CLI docs for 0.90.6

christophermaier:

### DIFF
--- a/www/source/docs/habitat-cli.html.md
+++ b/www/source/docs/habitat-cli.html.md
@@ -10,7 +10,7 @@ The commands for the Chef Habitat CLI (`hab`) are listed below.
 
 | Applies to Version | Last Updated |
 | ------- | ------------ |
-| hab 0.88.0/20191009204957 (linux) | 15 Oct 2019 |
+| hab 0.90.6/20191112141314 (linux) | 15 Nov 2019 |
 
 ## hab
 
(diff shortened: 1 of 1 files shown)


christophermaier:
@kagarmoe Those brackets come from the underlying CLI library we're using. I think the parentheses are coming from manually-maintained help text; we should tweak the CLI code to take advantage of the library code for those.

That's a good idea, but bigger than this particular PR.

www/source/docs/habitat-cli.html.md
```diff
@@ -1334,7 +1334,7 @@ hab pkg binlink [FLAGS] [OPTIONS] <PKG_IDENT> [BINARY]
 **OPTIONS**
 
 ```
--d, --dest <DEST_DIR>    Sets the destination directory [env: HAB_BINLINK_DIR=/hab/bin]  [default: /bin]
+-d, --dest <DEST_DIR>    Sets the destination directory [env: HAB_BINLINK_DIR=]  [default: /bin]
```
smacfarlane: Do we know why this changed?   Not a blocker, but I'm curious if this is something unique to the executing environment that generates the docs, or if there was a code change I missed.  

I suspect the former, and if so, what could we do to make this consistent across doc generation? (Automation is the correct answer :joy_cat: )

PR #7211, opened 2019-11-15
catch unsuccessful chocolatey commands

robbkidd:
An unsuccessful execution of a bare executable (as opposed to a PowerShell cmdlet) in a PowerShell script won't fail the script. Every execution of a binary needs to check the error level of the previous command. So, this adds non-zero error level checks that will throw a relatively informative message and will fail-exit the script so that the release pipeline correctly reflects the success/fail status of the package publishing.

Maybe BONUS: remove the duplication of the pack and publish commands in the DO-IT and the NOT-REALLY branches.

### DIFF
--- a/.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
+++ b/.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
@@ -1,4 +1,4 @@
-$ErrorActionPreference="stop" 
+$ErrorActionPreference="stop"
 
 Write-Host "--- Fetching manifest"
 $baseUrl = "https://packages.chef.io/files/stable/habitat/latest"
@@ -37,13 +37,20 @@ Write-Host "--- Publishing package"
 
 $valid_build_creator="Chef Expeditor"
 
-if($env:BUILDKITE_BUILD_CREATOR -eq $valid_build_creator) {
-  choco pack $tempDir/habitat.nuspec --version $Version
-  choco push habitat.$Version.nupkg -k $env:CHOCO_API_KEY --timeout 600
-} else {
-  Write-Host "--- NOT PUBLISHING: Build triggered by $env:BUILDKITE_BUILD_CREATOR and not $valid_build_creator"
-  Write-Host "choco pack $tempDir/habitat.nuspec --version $Version"
-  Write-Host "choco push habitat.$Version.nupkg -k <elided ChocoApiKey> --timeout 600"
-}
+$pack_cmd = "choco pack $tempDir/habitat.nuspec --version $Version"
+$publish_cmd = "choco push habitat.$Version.nupkg --timeout 600"
 
-Remove-Item $tempDir -Recurse -Force
+try {
+    if($env:BUILDKITE_BUILD_CREATOR -eq $valid_build_creator) {
+        Invoke-Expression $pack_cmd
+        if (-not $?) { throw "unable to choco pack" }
+        Invoke-Expression $publish_cmd " --key " $env:CHOCO_API_KEY
+        if (-not $?) { throw "unable to publish Chocolatey package" }
+    } else {
+        Write-Host "--- NOT PUBLISHING: Build triggered by $env:BUILDKITE_BUILD_CREATOR and not $valid_build_creator"
+        Write-Host $pack_cmd
+        Write-Host $publish_cmd " --key <elided ChocoApiKey>"
+    }
+} finally {
+    Remove-Item $tempDir -Recurse -Force
+}


.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
```diff
@@ -37,13 +37,18 @@ Write-Host "--- Publishing package"
 
 $valid_build_creator="Chef Expeditor"
 
+$pack_cmd = "choco pack $tempDir/habitat.nuspec --version $Version"
+$publish_cmd = "choco push habitat.$Version.nupkg -k $env:CHOCO_API_KEY --timeout 600"
+
 if($env:BUILDKITE_BUILD_CREATOR -eq $valid_build_creator) {
```
mwrock: I'd recomend putting this `if` block in a `try` and the below `Remove-Item` in a `finally`

robbkidd: Ooooh. I am happy to learn more better patterns.

.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
```diff
@@ -37,13 +37,20 @@ Write-Host "--- Publishing package"
...
+try {
+    if($env:BUILDKITE_BUILD_CREATOR -eq $valid_build_creator) {
+        Invoke-Expression $pack_cmd
+        if (-not $?) { throw "unable to choco pack" }
+        Invoke-Expression $publish_cmd
+        if (-not $?) { throw "unable to publish Chocolatey package" }
+    } else {
+        Write-Host "--- NOT PUBLISHING: Build triggered by $env:BUILDKITE_BUILD_CREATOR and not $valid_build_creator"
+        Write-Host "$pack_cmd"
+        Write-Host "$publish_cmd"
```
christophermaier: This will print the API key out; before, this was explicitly elided.

.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
```diff
@@ -37,13 +37,20 @@ Write-Host "--- Publishing package"
...
+try {
+    if($env:BUILDKITE_BUILD_CREATOR -eq $valid_build_creator) {
+        Invoke-Expression $pack_cmd
+        if (-not $?) { throw "unable to choco pack" }
+        Invoke-Expression $publish_cmd " --key " $env:CHOCO_API_KEY
+        if (-not $?) { throw "unable to publish Chocolatey package" }
+    } else {
+        Write-Host "--- NOT PUBLISHING: Build triggered by $env:BUILDKITE_BUILD_CREATOR and not $valid_build_creator"
+        Write-Host $pack_cmd
+        Write-Host $publish_cmd " --key <elided ChocoApiKey>"
```
robbkidd: @mwrock @christophermaier I think I'm doing PowerShell string concatenation correctly here ...? 😬

.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
robbkidd: Updated to use string concatenation to use a different key string for the command execution and the NOOP output.

.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
```diff
@@ -37,13 +37,20 @@ Write-Host "--- Publishing package"
...
-} else {
-  Write-Host "--- NOT PUBLISHING: Build triggered by $env:BUILDKITE_BUILD_CREATOR and not $valid_build_creator"
-  Write-Host "choco pack $tempDir/habitat.nuspec --version $Version"
-  Write-Host "choco push habitat.$Version.nupkg -k <elided ChocoApiKey> --timeout 600"
+$pack_cmd = "choco pack $tempDir/habitat.nuspec --version $Version"
+$publish_cmd = "choco push habitat.$Version.nupkg --timeout 600"
+
+try {
+    if($env:BUILDKITE_BUILD_CREATOR -eq $valid_build_creator) {
+        Invoke-Expression $pack_cmd
```
markan: I don't think the error behavior of Invoke-Expression is what you expect;
If you run 
```
$test_fail = "test 1 -ne 1"
try {
    Invoke-Expression $test_fail
    if (-not $?) { throw "bad command" }
    Write-Host "Invoke succeeded"
} finally {
    Write-Host "Finally"
}
```
You get 
```
Invoke succeeded
Finally
```
It is apparently not like bash, where: 
```
% test 1 -ne 1; echo $?
1
```

markan: https://stackoverflow.com/questions/32348794/how-to-get-status-of-invoke-expression-successful-or-failed

mwrock: @markan is correct. What you want here is `if($LASTEXITCODE -ne 0)`. While `$?` will always evaluate to `$true` here, The choco exit code will persist in the `$LASTEXITCODE` variable.

.expeditor/scripts/finish_release/publish_to_chocolatey.ps1
christophermaier: :thinking: When I run 

```powershell
Write-Host "foo" "bar"
```
I get
```
foo bar
```
Seems like the spaces between the strings are incorporated into the final output.

mwrock:
Closing in favor of #7330 which just adds a commit to @robbkidd 's commit transforming the `$?` checks to `$LASTEXITCODE` checks.

PR #7212, opened 2019-11-16
H2 anchors stay in view

kagarmoe:
When you select menu sub-item in the left-nav, you'll land on the H2 for that item instead of 120px below it.

The fix only works when applied to the anchor element, not to the H2 element.


www/source/stylesheets/_global.scss
```diff
@@ -399,9 +399,17 @@ code {
...
+    content: "";
+    height: 220px;
+    margin: -220px 0 0;
+  }
+
 h2:not(:first-child) {
-  margin-top: rem-calc(40);
-}
+   margin-top: rem-calc(40);
+ }
```
davidMcneil: Looks like accidentally added whitespace?

www/source/stylesheets/_global.scss
```diff
@@ -399,9 +399,17 @@ code {
...
 }
 
+h2
+  a::before {
+    display: block;
+    content: "";
+    height: 220px;
+    margin: -220px 0 0;
+  }
+
```
davidMcneil: Adding brackets around the `h2`'s contents would keep it consistent with the rest of the document. Adding a `visibility: hidden;` and `pointer-events: none;` could help indicate that this element is solely for layout purposes. Adding a comment about why this block is needed would also be helpful.

www/source/stylesheets/_global.scss
kagarmoe: Thanks!

www/source/stylesheets/_global.scss
kagarmoe: Thanks @davidMcneil! I tried using a height variable in the new code.

www/source/stylesheets/_code.scss
```diff
@@ -44,4 +44,4 @@ pre {
 
     }
   }
-}
\ No newline at end of file
+}
```
kagarmoe: This was pulled in by an auto-fix when I closed the file.

PR #7214, opened 2019-11-18
Add PIDS_FROM_LAUNCHER feature to eliminate service PID files

### LINKED ISSUE #7021: Remove PID Files for Services
## What problem are we solving? 
The fact that we are storing service PIDs in files on disk has led to several instances of instability when the contents of the files diverges from the current actual state of the system. Additionally, files with PIDs of "0" have been reported, with bad consequences (services are not started appropriately, supervisors can shut down when shutting down a single service, etc.)

Since the Launcher process is a definitive source of truth for service PIDs, since it is the parent process of all services. Rather than storing and reading PIDs to files, the Supervisor can instead query the Launcher. This should give us more confidence by "cutting out the middle man", and should make the system more robust in the face of PID files becoming corrupted or out of sync with reality.

## Objective and key results. 
A new Supervisor that can ask a new Launcher what the PID of a supervised service is. For the time being, this functionality should be gated behind a feature flag. Additionally, if the feature flag is present, but the Launcher being used is old and thus not capable of providing PIDs to the Supervisor, the Supervisor should continue to use PID files (i.e., behave as though the feature flag were not set).

## Acceptance criteria. 

To manually validate:
* Start a Supervisor that does not use PID files, and load one or more services. 
* Observe the output of `hab svc status` to discover the PIDs of these services.
* Restart the Supervisor with a `SIGHUP` to simulate a Supervisor upgrade. PID files are only read to allow the Supervisor to reattach to running services following an upgrade. (run `sudo pkill --signal=HUP hab-launch`)
* Observe the logs of the Supervisor to confirm reattachment to the services.
* Run `hab svc status` again and observe the same PIDs for each service as before.

Documentation of how to enable the new feature.

Automated tests proving the desired behavior, as well as confirming the old behavior if the feature is not enabled.

## Dependencies and blockers. 
None

## Out of scope. 
None

## Open questions. 
None

christophermaier:
By setting `HAB_FEAT_PIDS_FROM_LAUNCHER=1` in the environment, _and_ using a Launcher built from this code or later, the Supervisor will now ask the Launcher for the PIDs of services it is supervising, rather than relying on reading that information from files on disk.

Eventually, this will be the default behavior, but it is feature-flagged at the moment as we begin rolling it out.

If users are using an older Launcher, the Supervisor will continue to use PID files, regardless of whether the feature is enabled or not.

Read individual commit messages for further details.

Fixes #7021

### DIFF
--- a/components/launcher-client/src/client.rs
+++ b/components/launcher-client/src/client.rs
@@ -155,6 +155,18 @@ impl LauncherCli {
         Ok(reply.pid as Pid)
     }
 
+    /// Query the launcher for the PID of the named service. If the
+    /// Launcher is aware of it, you'll get `Ok(Some(Pid))`
+    pub fn pid_of(&self, service_name: &str) -> Result<Option<Pid>> {
+        let msg = protocol::PidOf { service_name: service_name.to_string(), };
+        Self::send(&self.tx, &msg)?;
+        let reply = Self::recv::<protocol::PidIs>(&self.rx)?;
+        match reply.pid {
+            Some(pid) => Ok(Some(pid as Pid)),
+            None => Ok(None),
+        }
+    }
+
     pub fn terminate(&self, pid: Pid) -> Result<i32> {
         let msg = protocol::Terminate { pid: pid.into() };
         Self::send(&self.tx, &msg)?;
--- a/components/launcher-client/src/client.rs
+++ b/components/launcher-client/src/client.rs
@@ -9,11 +9,24 @@ use ipc_channel::ipc::{IpcOneShotServer,
                        IpcSender};
 use std::{collections::BTreeMap,
           io,
-          path::Path};
+          path::Path,
+          thread,
+          time::{Duration,
+                 Instant}};
 
 type Env = BTreeMap<String, String>;
 type IpcServer = IpcOneShotServer<Vec<u8>>;
 
+// Defines how long to wait to receive a reply from the Launcher.
+//
+// Initially used for calls to get the PID from a service as a way to
+// deal with older Launchers that don't know about that protocol
+// message, and don't reply back on unknown messages.
+//
+// Don't override this value unless you know what you're doing.
+habitat_core::env_config_duration!(LauncherInteractionTimeout,
+                                   HAB_LAUNCHER_INTERACTION_TIMEOUT_MS => from_millis,
+                                   Duration::from_millis(1000));
 pub struct LauncherCli {
     tx: IpcSender<Vec<u8>>,
     rx: IpcReceiver<Vec<u8>>,
@@ -23,6 +36,9 @@ pub struct LauncherCli {
     // wraps the pipe in a WinHandle whose drop impl calls CloseHandle.
     #[cfg(not(windows))]
     pipe: String,
+
+    /// Maximum wait time for interactions that can timeout.
+    timeout: Duration,
 }
 
 #[cfg(not(windows))]
@@ -46,10 +62,14 @@ impl LauncherCli {
         Self::send(&tx, &cmd)?;
         let (rx, raw) = ipc_srv.accept().map_err(|_| Error::AcceptConn)?;
         Self::read::<protocol::NetOk>(&raw)?;
+
+        let timeout = LauncherInteractionTimeout::configured_value().into();
+
         Ok(LauncherCli { tx,
                          rx,
                          #[cfg(not(windows))]
-                         pipe: pipe_to_sup })
+                         pipe: pipe_to_sup,
+                         timeout })
     }
 
     /// Read a launcher protocol message from a byte array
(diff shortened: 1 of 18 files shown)


components/sup/src/manager/service/supervisor.rs
```diff
@@ -35,26 +36,53 @@ static LOGKEY: &str = "SV";
...
+    /// Create a new instance for `service_group`.
+    ///
+    /// The `pid_source` governs how we determine the PID of the
+    /// supervised process. Once the PIDS_FROM_LAUNCHER feature flag
+    /// goes away, this can be removed.
+    pub fn new(service_group: &ServiceGroup, pid_source: ServicePidSource) -> Supervisor {
+        let pid_file = match pid_source {
+            ServicePidSource::Files => Some(fs::svc_pid_file(service_group.service())),
+            ServicePidSource::Launcher => None,
+        };
```
stevendanna: I wouldn't necessarily change this since we think that the feature is going to be removed eventually.  But, if we think this is going to live in the code base a while, I'm wondering it it would make sense to move the behavior differences between the pid sources out of this code and out into a trait.

components/launcher-client/src/client.rs
```diff
@@ -75,6 +95,46 @@ impl LauncherCli {
...
+    fn recv_timeout<T>(rx: &IpcReceiver<Vec<u8>>, timeout: Duration) -> Result<T>
+        where T: protocol::LauncherMessage
+    {
+        // If ipc_channel implemented this directly, we wouldn't have
+        // to do this :(
+        let timeout = Instant::now() + timeout;
+        loop {
+            match rx.try_recv().map_err(|e| Error::from(*e)) {
+                Ok(bytes) => return Self::read(&bytes),
+                Err(Error::IPCIO(io::ErrorKind::WouldBlock)) => {
```
davidMcneil: This is cool! It is exactly what Rust's async IO implementations do but instead of simply sleeping, it registers a `Waker` that uses underlying syscalls to determine when more data is available.

components/launcher-client/src/client.rs
```diff
@@ -75,6 +95,46 @@ impl LauncherCli {
...
+    ///
+    /// This function is an attempt to get around this situation, but
+    /// ONLY for the specific scenarios where it is appropriate. I am
+    /// hesitant to introduce a global timeout at this point, because
+    /// it's difficult to know how that will affect the overall
+    /// interactions between the Supervisor and the Launcher (it
+    /// *should* be fine, but I can't guarantee that right now).
+    ///
+    /// As such, use this with caution and intention.
+    fn recv_timeout<T>(rx: &IpcReceiver<Vec<u8>>, timeout: Duration) -> Result<T>
```
davidMcneil: Just thinking out loud... If we had a way to check the version of the launcher, would this be needed? Unfortunately, if this was accomplished through an API call to the launcher, it would not help in this specific scenario (because the old version of the launcher will not have the API), but maybe it would be worth adding for future situations like this?

components/launcher-protocol/src/types.rs
```diff
@@ -276,3 +276,47 @@ impl LauncherMessage for Shutdown {
...
+    const MESSAGE_ID: &'static str = "PidIs";
+
+    fn from_proto(proto: generated::PidIs) -> Result<Self> {
+        // TODO (CM): ensure that the Pid is never Some(0)
+        Ok(PidIs { pid: proto.pid })
+    }
+}
+
+impl From<PidIs> for generated::PidIs {
+    // TODO (CM): I would need to ensure that PidIs can never contain
```
davidMcneil: Could you use [`std::num::NonZeroU32`](https://doc.rust-lang.org/std/num/struct.NonZeroU32.html)?

christophermaier: Yeah, that's kinda what I'm thinking. Hopefully that's not a huge refactor, but something I'm thinking about for a follow-on PR.

components/launcher-client/src/client.rs
christophermaier: Neat :smile: That sure would be nicer than sleeping.

components/launcher-client/src/client.rs
christophermaier: That would certainly have helped, but unfortunately we don't have such an API right now :frowning:

This particular situation was complicated by the fact that if you sent a message to a Launcher that didn't know how to process that message, the Launcher would just ignore it. However, the Supervisor was always using a blocking `recv` call, so it would hang waiting for a reply that was never going to come. The end result is that you can basically get yourself into a mutual deadlock situation between both processes.

Given all that, this seemed the most straightforward way forward (this PR _does_ fix that "ignore unknown messages" issue, though).

Interestingly, the Launcher _does_ have the ability to see what version of the Supervisor it's running. Adding that ability to the Supervisor seems like a good idea.

components/sup/src/manager/service/supervisor.rs
christophermaier: If it was intended as a long-standing variation, then yeah, adding a bit more structure around this would be good.

I think that since this is more of a transition state, though, the amount of structure we've got now is probably good enough.

We _should_ draw up a concrete timeline for when that switchover should take place, though.

Issue #6795, opened 2019-08-06
Service updater not working as expected


abhij22:
Current Setup:
On prem builder
Habitat supervisor running as daemon

Issue:

OS: RHEL 7.6

eeyun:
Hello @abhij22 thanks for opening this issue!  We're looking into this but we could use some more context to help us get the ball rolling. For clarification what you're seeing specifically is that the environment variables getting set in your systemd unit file are not getting propagated to the supervisor, is that correct?

As a related question, have you by any chance tried this on any other OS? Have you tried setting the envvars exported to a profile or any other wider net than in an individual shell?

fifthecho:
@abhij22 can correct me if I'm wrong (and I work with him, so I am probably wrong) but the core of the behavior we're seeing is that if `HAB_BLDR_URL` is specified in the environment and you `hab svc load` the service will load but never update, but if you specify `hab svc load --url <PRIVATE_BLDR_URL>` then the service will update correctly.

abhij22:
Hey @eeyun,

What I am seeing is that the env variables set in the systemd are not propagated to the service updater process. It seems that the service updater is inheriting environment variables from my shell rather than systemd. If I have the `HAB_BLDR_URL` and `HAB_AUTH_TOKEN` set in my shell when I do a `hab svc load`, the service updater works as expected. If they are not set in my shell when I do a `hab svc load` I see the Unauthorized error in the supervisor logs.

Issue #7033, opened 2019-10-14
Design spike: Config file for supervisor


krasnow:
Acceptance criteria:
* Define breakdown of epic

davidMcneil:
One approach would be to transition to using the [`structopt`](https://github.com/TeXitoi/structopt) crate. `structopt` uses `clap` under the hood, but handles argument parsing by defining a single struct and then making heavy use of macros to define how the structs fields map to subcommands, arguments, flags ...

How does this help with a config file? We could use `serde` to define `Serialize` and `Deserialize` on the struct and we would be able to read and write this struct to file, network, or stdout (#7225) nearly for free. It would also make supporting a variety of formats (toml, json, yaml ...) easy (for better or worse).

I think this approach would have another pretty cool benefit. Wrapping the singelton instance of the struct in an `Arc<RwLock<T>>` would give us a safe and easy way to allow dynamically updating the config while the supervisor is running. This has been requested for parts of the config (#6864) This would no longer require a special case. The entire config could be dynamic.

A big negative of using `structopt` would be if their macro language did not allow us to do everything we wanted to do and would be possible by using `clap` directly (I have used `structopt` for pretty complicated interfaces but am not sure what the exact restriction are). There are [plans](https://clap.rs/2019/03/08/clap-v3-update-structopt/) to make `structopt` a native part of `clap`.

It would require a lot of work to port the existing interface, but an incremental transition to `structopt` should be possible. Honestly, I think any addition of a config file will require significant changes to the way the cli code works.

christophermaier:
@davidMcneil That sounds like a really fruitful avenue to pursue. I've long wanted a way to be able to dynamically update a number of aspects of how the Supervisor runs that shouldn't require restarts (e.g., where Supervisor updates come from, if the Supervisor should auto-update at all, etc.)

If this gets us closer to that world, then I think it's definitely worthwhile!

davidMcneil:
See more design work [here](https://github.com/davidMcneil/unified-dynamic-rust-app-config).

PR #7215, opened 2019-11-19
ENVVAR honoring in ServiceSpec impl for bldr_url and channel

jeremymv2:
Closes https://github.com/habitat-sh/habitat/issues/6795

This change adds ENVVAR honoring to the ServiceSpec impl for bldr_url and
channel fields such that they are populated with respect to the
following priority (highest to lowest) when a service is loaded:

1. client cli args
2. client env
3. sup env
4. defaults

This resolves some frustating end user confusion most often when utilizing an OnPrem Depot.
In that scenario, the user sets the `HAB_BLDR_URL` for their private depot in the context of the Supervisor service expecting that it will be honored on a client `hab svc load ..`. However, to their surprise, they also have set the ENVVAR in their client shell context, or use the `--url` option for it to take.

Todo:
- [x] Write e2e test


components/sup/src/manager/service_updater.rs
```diff
@@ -158,7 +158,8 @@ impl ServiceUpdater {
                                                  // census group for our service.
                                                  census_ring: &CensusRing)
                                                  -> Option<PackageIdent> {
-        debug!("Checking for updated package!");
+        debug!("Checking {} for updated {} package in the {} channel",
+               service.bldr_url, service.spec_ident, service.channel);
```
stevendanna: 🎉

christophermaier:
@jeremymv2 I've just started looking at this, and I have a minor nitpick... would you mind reworking the commit message so it has a proper "title" (a good one would be the current title of this PR!), leaving the explanation for the commit body?

While the PR title goes into our changelog, keeping the commit message itself tidy helps when interacting with the code on the command line, other git clients, etc.

Thanks!

components/core/src/lib.rs
```diff
@@ -47,3 +47,9 @@ impl ChannelIdent {
 impl fmt::Display for ChannelIdent {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result { write!(f, "{}", self.0) }
 }
+
+pub fn channel_from_env() -> Option<ChannelIdent> {
+    env::var(ChannelIdent::ENVVAR).ok().map(ChannelIdent::from)
+}
+
+pub fn default_channel() -> ChannelIdent { channel_from_env().unwrap_or_default() }
```
christophermaier: You don't actually need `default_channel` or `channel_from_env`, since the underlying `env_config!` infrastructure provides this functionality for you via `ChannelIdent::configured_value()` (indeed, this is the precise usecase that macro was originally written to serve!). Let me know if you want to talk more about that :smile: 

As an aside, and assuming that macro wasn't present, it would have been good to associate these functions with the type directly, allowing you to call them like `ChannelIdent::channel_from_env()`. That would be even better as `ChannelIdent::from_env()`, since you can let the type "do some of the talking" for you. That same reasoning would also have led to `ChannelIdent::default()`, which in turn would have led to some further thought about whether that should have been an implementation of `Default`, rather than a function in the `ChannelIdent` impl itself (In this case, it would _not_ have been correct to make this a `Default` implementation, because `Default` implementations that depend on the environment are a very sneaky kind of evil... and that's why `env_config!` creates a function called `configured_value` to do this.)

Also, it's a good idea to not make things public that don't need to be. `channel_from_env` was only ever called from `default_channel`, so it could have just been private (if not completely inlined with `default_channel`'s implementation). The Rust compiler is really good at alerting you to unused code, but you have to make sure your visibility settings are correct. If you say an item is public, the compiler takes you at your word, and assumes that external clients of your library could call them. On the other hand, if they're not public, and the compiler sees there are no callers, it'll tell you and you get the joy of deleting code :tada: Don't deprive yourself of that joy :rofl:

components/sup/src/manager/service_updater.rs
christophermaier: This is such a nice touch :+1:

components/hab/src/cli.rs
```diff
@@ -1209,8 +1209,8 @@ fn sub_svc_load() -> App<'static, 'static> {
             "Application name; [default: not set].")
         (@arg ENVIRONMENT: --environment -e +takes_value requires[APPLICATION]
             "Environment name; [default: not set].")
-        (@arg CHANNEL: --channel +takes_value default_value[stable]
-            "Receive package updates from the specified release channel")
+        (@arg CHANNEL: --channel +takes_value
+            "Receive package updates from the specified release channel [default: stable]")
```
christophermaier: Since we've got a type that encapsulates both the default value and the environment variable the data responds to, we can take advantage of that in our CLI definition.

```rust
(@arg CHANNEL: --channel +takes_value env(ChannelIdent::ENVVAR) default_value(ChannelIdent::default_as_str())
 "Receive package updates from the specified release channel")
```

This also requires adding this over in `components/core/src/lib.rs`:
```rust
default_as_str!(ChannelIdent);
```
That last macro is something we wrote to help bridge the existing gap in functionality of Clap to expose such typed default values in the CLI.

The resulting help text looks like this:
```sh
...
--channel <CHANNEL>
     Receive package updates from the specified release channel [env: HAB_BLDR_CHANNEL=]  [default: stable]
...
```
Or this, if you have `HAB_BLDR_CHANNEL` set:
```sh
...
--channel <CHANNEL>
     Receive package updates from the specified release channel [env: HAB_BLDR_CHANNEL=foo]  [default: stable]
...
```
(Since this could respond to an environment variable, it's important to document that fact.)

(Also note that this pattern isn't uniformly followed throughout the CLI, as it's only been recently "discovered". If you see instances where defaults are hardcoded in documentation strings, then that's an indicator that some more refactoring could take place.)

christophermaier: One thing that I'm not quite sure the best way to address is the fact that when you start up a Supervisor, it has a `--channel` and `--url` option that apply to where the Supervisor pulls updates from, if it is auto-updating. That channel _does not_ respond to `HAB_BLDR_CHANNEL` (I'm not sure if that was intentional, or not), but the URL _does_ respond to `HAB_BLDR_URL`. 

This PR would add a "hidden" meaning for `HAB_BLDR_CHANNEL`, and a _double_ meaning for `HAB_BLDR_URL`. The overall layering and precedence of these, as you laid out in the PR message, make sense, and provide ample opportunity for overriding, and the modified log message you include also provides visibility, which is fantastic. However, it seems like there's perhaps room for additional feedback or further documentation / clarification about how these environment variables interact, and which portions of code are sensitive to them and which aren't.

This also brings to mind #7033... I wonder if that might be a more appropriate place to define these kind of "global default" values that makes it easier to reason about? (genuinely asking; I'm not sure what the final answer should be just yet).

@krasnow, @apriofrost: do you have any thoughts on this?

components/core/src/lib.rs
jeremymv2: Thank you! I updated to using the macro and removing `default_channel` and `channel_from_env` in https://github.com/habitat-sh/habitat/pull/7215/commits/1b608543359688820c9848f19e924a778f3f7f86

components/hab/src/cli.rs
jeremymv2: > Since we've got a type that encapsulates both the default value and the environment variable the data responds to, we can take advantage of that in our CLI definition.
> 
> ```rust
> (@arg CHANNEL: --channel +takes_value env(ChannelIdent::ENVVAR) default_value(ChannelIdent::default_as_str())
>  "Receive package updates from the specified release channel")
> ```

@christophermaier I really appreciate that you pointed out the `env()` clap macro! 🎉  That definitely helps provide more UX context when looking at the help text. I've incorporated that in this [commit](https://github.com/habitat-sh/habitat/pull/7215/commits/382bd6a1ac2f3519f38e1afd8c2672ce91a59d27) for both BLDR_URL and CHANNEL args.

I did omit the `default_value(ChannelIdent::default_as_str()` though considering I wanted the defaults to be applied from the Supervisor context, when it instantiates the ServiceSpec. If we omit the default at the client cli level, it will pass [`Option<None>`](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/main.rs#L1959-L1960) for both in the ctl_message. The `None` for both won't then override the defaults when the Service Manager [merges](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/spec.rs#L242-L247) in the values passed via the ctl_message. Otherwise, if we set a default at the client context, it will override any defaults from the Supervisor context and we lose the intent of this PR since the client will _always_ pass a `Some` value.

I'm curious what your thoughts are on that idea 😃

jeremymv2: > One thing that I'm not quite sure the best way to address is the fact that when you start up a Supervisor, it has a `--channel` and `--url` option that apply to where the Supervisor pulls updates from, if it is auto-updating. That channel _does not_ respond to `HAB_BLDR_CHANNEL` (I'm not sure if that was intentional, or not), but the URL _does_ respond to `HAB_BLDR_URL`.
> 
> This PR would add a "hidden" meaning for `HAB_BLDR_CHANNEL`, and a _double_ meaning for `HAB_BLDR_URL`.

My instincts are telling me that what we need is consistency to make this less confusing and more predictable. If we do these two things for `hab sup run`, I think we keep true to that ideal:

* make `CHANNEL` respond to `HAB_BLDR_CHANNEL`
* use the clap `env()` macro for _both_ `BLDR_URL` and `CHANNEL` args

The result is that existing `HAB_BDLR_URL` and `HAB_BLDR_CHANNEL` envvars in the Supervisor or cli context act with a global context and apply to _both_ Supervisor and Services. The values from the ENVVARs would be apparent to the user via the help text due to the `env()` clap macro. If the user wishes to set them differently for different contexts, they can be more specific by utilizing the `--url` and `--channel` arguments.

I believe that will still marry up with #7033 if we make the same arguments from the Supervisor config file the _most_ specific with the highest precedence level or we could even create distinguishing configuration for Supervisor vs Service. 
 
 @christophermaier, @krasnow, @apriofrost: thoughts?

christophermaier: :thinking: Making a change like this to cause the Builder URL and channel apply to both the updating supervisor _and_ services might not great. One thing that seems a bit weird is that the environment variables would apply to both, but the corresponding CLI options only apply to the Supervisor.

I wonder if it might be a good idea to go ahead and separate those things?

jeremymv2: > Ah, that's an excellent point! That should most definitely be documented with a comment in the CLI code, then, because if I didn't know that background, that would be one of the first things I would change in this code, thus breaking everything!

Ahh yes, 1000% agree with this! I've added comments in https://github.com/habitat-sh/habitat/pull/7215/commits/66bbfc007deb5ad35d90192cf43b1955e10c14d3 Additionally the (to-be-added) e2e tests will help cover the code, providing some safety.

> 🤔 Making a change like this to cause the Builder URL and channel apply to both the updating supervisor and services might not great. One thing that seems a bit weird is that the environment variables would apply to both, but the corresponding CLI options only apply to the Supervisor.
> I wonder if it might be a good idea to go ahead and separate those things?

Great point! I do see the `HAB_BLDR_URL` and `HAB_BLDR_CHANNEL` CLI combo options for Supervisor and Services are distinct already:

* The Supervisor's options for `hab sup run` are defined [here](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/cli.rs#L1081-L1086) which apply to Supervisor updates.
* The ServiceSpec options from `hab svc load` are defined [here](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/cli.rs#L1212-L1219) which apply to the Service updates.

If we move forward with this idea, it would mean the Environment variables apply to both, either from the cli or the Supervisor context. Distinction can still be asserted by specifying the CLI arguments (which are a higher precedence) as needed for either the Supervisor (`hab sup run`) or the Service (`hab svc load`).

jeremymv2: @christophermaier I'm curious what your thoughts are on the latest couple of commits and the included tests. I do think the work here solves the user problem from the issue, however I can understand if we want to hold off on it or pivot to a different strategy. Of course, I would just like to move forward in the best way possible for the project. Thanks!

PR #7256, opened 2019-12-03
Update futures ecosystem

### LINKED ISSUE #7209: Update futures dependencies
With the stabilization of Rust's `async`/`await` syntax, the entire futures ecosystem has been updated. We need to understand the changes we need to make to our codebase to consume the new libraries.

This is solely about being able to update our dependencies at this point; updating our asynchronous Rust code to actually use the `async` and `await` syntax is _explicitly_ out of scope for this work.

davidMcneil:
Resolves #7209 

**[Update 12/12]**
This issue is unblocked thanks to this [fix](https://github.com/tokio-rs/bytes/issues/340) in the `bytes` crate.

**[Update 12/09]**
I am moving this PR out of its draft status. All the changes I intended to make are complete and it should be ready for final review. I am still running into memory allocation errors when using jemalloc. The system allocator seems to be working fine. This can be tested by commenting out the line [here](https://github.com/habitat-sh/habitat/blob/b7582dca0f7a923bddea4ba2fd451087e8d953de/components/sup/src/main.rs#L80). I am adding the `BLOCKED` and `DO-NOT-MERGE` labels until this issue is resolved.

**[Update 12/06]**
I finished updating the supervisor code. The good news it was not as bad as I anticipated. The bad news it does not work.

When trying to issue commands to the supervisor, it frequently dies with a `SIGABORT` `memory allocation of 5 bytes failed`. It does not happen every time (I have been able to successfully load and run a service). And the number of bytes it tries to allocate seems impacted by the log level. At trace level it is 3.

**[Initial 12/03]**
The main objective of this PR is to upgrade [`futures`](https://crates.io/crates/futures) to version `0.3` and [`tokio`](https://crates.io/crates/tokio) to version `0.2`. This PR currently upgrades the `hab` cli and its dependencies `habitat-sup-client` and `habitat-sup-protocol`. It does not upgrade `sup` which is the only outstanding crate left to upgrade. However, `sup` contains a significant amount of async code. `sup` will not build off of this branch because `habitat-sup-protocol` changes leak into `sup`.

**Side Note on Updating `tokio`**: The version of [`bytes`](https://crates.io/crates/bytes) `tokio` and `prost` use are not compatible. See this [issue](https://github.com/danburkert/prost/issues/242). This requires using a forked version of `prost`.

We can use these changes to best asses how to proceed in refactoring `sup`. The changes tend towards being a code bomb, but I think this is somewhat necessary. To step back for a second. The changes to the `Future` trait itself are substantial ([old](https://docs.rs/futures/0.1.29/futures/future/trait.Future.html), [new](https://doc.rust-lang.org/std/future/trait.Future.html)) not to mention the significant changes to the surrounding ecosystem. This sometimes makes a one-to-one mapping from old to new impossible.

The worst offender in this PR was the need to rewrite the [`SrvClient`](https://github.com/habitat-sh/habitat/blob/3872b49b3f320c7cc3cf7dc05c04bc46147017d7/components/sup-client/src/lib.rs#L133). The `SrvClient` API  was... odd. You would make a request with [`call`](https://github.com/habitat-sh/habitat/blob/3872b49b3f320c7cc3cf7dc05c04bc46147017d7/components/sup-client/src/lib.rs#L187) which would consume the `SrvClient` and return a `SrvReply`. `SrvReply` was a stream of responses. After all responses where drained `SrvReply` could be converted back to a `SrvClient` to be used again. This was fine in concept, but the old `Stream` combinators consumed the `SrvReply` making it impossible to convert it back to a `SrvClient`. Unless you knew beforehand how many items were in the stream then you could pull them out one at a time. 

This sort of odd API was required by the old async ecosystem and is the very reason the new ecosystem exists. That being said it is very difficult to mimic the old behavior with the new ecosystem. This is what lead to the rewrite of `SrvClient`. It was more expedient to implement it in the new ecosystem than directly port it.

**Side Note on the `SrvClient` Rewrite**: The new version supports multiple concurrent requests over a single TCP connection. This was something I thought was needed, but it turns out the server closes the TCP connection after every request. Meaning the client really only needs to support one request per connection. This simplification would greatly simplify the code.

I think this pattern will apply when porting the supervisor. The overall shape of the async code will be the same, but where we make significant use of the old combinators or are manually implementing the `Future` trait will require more substantial rework.

So how is it best to proceed? Are we ok with the level of rework needed to port `hab`? I think it ultimately comes down to testing. Are we comfortable with our end-to-end tests to catch regressions in `hab`? If not, can we become comfortable with these changes by adding more end-to-end tests? Similar questions will apply to `sup`.


christophermaier:
@davidMcneil So far, this looks pretty good (as good as it can, given the task :sweat_smile: ). The new async / futures API is really nice!

The `SrvClient` code is much easier to understand now, but (as you suggest) it would be interesting to see what it would look like if it didn't have to maintain state for multiple requests.

As for testing, I think more tests are probably a good thing here, but I don't think we'll necessarily need a lot more. All the existing end-to-end tests use `hab` and call many of the operations touched by this PR; I would expect that anything that broke these commands would be surfaced pretty quickly. A survey of existing testing scenarios should reveal which commands are not exercised already.

I'm excited to see how the rest of this ends up! :tada:

components/butterfly/Cargo.toml
```diff
@@ -21,8 +21,10 @@ log = "*"
 lazy_static = "*"
 prometheus = "*"
 parking_lot = "*"
-prost = "*"
-prost-derive = "*"
+# Must use fork due to this issue https://github.com/danburkert/prost/issues/242
```
christophermaier: It'd be a good idea to create an issue to track this fork situation.

components/sup/Cargo.toml
```diff
@@ -15,12 +15,14 @@ path = "src/main.rs"
...
 [dependencies]
-bytes = "*"
+bytes = "0.5.0"
 actix-web = { version = "*", default-features = false, features = [ "rust-tls" ] }
 byteorder = "*"
 clap = { version = "*", features = [ "suggestions", "color", "unstable" ] }
 cpu-time = "*"
-futures = "*"
+futures = { version = "0.3.1" }
+# futures 0.1 is needed because actix-web has not yet upgraded to futures 0.3
```
christophermaier: An issue for tracking this work would also be nice.

components/sup/src/ctl_gateway/server.rs
```diff
@@ -141,91 +142,63 @@ impl CtlCommand {
 
 /// Server's client representation. Each new connection will allocate a new Client.
 struct Client {
-    handle: reactor::Handle,
-    state:  Rc<RefCell<SrvState>>,
+    state: Arc<Mutex<SrvState>>,
```
christophermaier: Oh, did we finally get this code off of its own single-thread reactor? If so, :tada:

davidMcneil:
> The `SrvClient` code is much easier to understand now, but (as you suggest) it would be interesting to see what it would look like if it didn't have to maintain state for multiple requests.

@christophermaier [9d79ed](https://github.com/habitat-sh/habitat/pull/7256/commits/9d79ed4f4acbc148ecae625292f79a0065a1e071) simplifies the `SrvClient`. If we ever need the performance of making multiple requests per TCP connection we could use the more sophisticated approach, but simple is hard to beat!

davidMcneil:
> The trick about this is making sure we can kill this future when we need to shut down the Supervisor; otherwise we'll be back in the same position we were in with the health checks back in #6717.

It is actually going to be a problem the other way. There is now way to tell the runtime to keep run until idle. See [here in the code](https://github.com/habitat-sh/habitat/blob/91539dced24d2ec05ee5d90292ba96046221d85e/components/sup/src/manager.rs#L1198) and a discussion [here](https://github.com/tokio-rs/tokio/issues/1156). They changed it to specificlally support this kind of use case, but now I have to figure out what the ramifications of that are if we wanted the runtime to block.

components/sup-client/src/lib.rs
```diff
@@ -140,56 +126,48 @@ impl From<termcolor::ParseColorError> for SrvClientError {
...
+        let mut handshake = protocol::ctl::Handshake::default();
+        handshake.secret_key = Some(String::from(secret_key));
+        let mut message = SrvMessage::from(handshake);
+        message.set_transaction(current_transaction);
+        socket.send(message).await?;
+
+        // Verify the handshake response
+        socket.next()
+              .await
+              .ok_or(SrvClientError::ConnectionClosed)??;
```
christophermaier: Do we need some kind of timeout for this, or is that effectively handled on the server-side?

components/sup/src/manager/service/hook_runner.rs
```diff
@@ -67,32 +66,20 @@ impl<H> HookRunner<H> where H: Hook + Sync + 'static
...
-                                          .run(&self.service_group, &self.pkg, self.passwd.as_ref())
-                                          .ok();
-                                  let run_time = start.elapsed();
-                                  tx.send((exit_value, run_time))
-                                    .expect("Couldn't send oneshot signal from HookRunner: \
-                                             receiver went away");
-                              })?;
-
-        Ok(rx.await.expect("to receive oneshot signal from HookRunner"))
+        task::spawn_blocking(move || {
```
christophermaier: Whoa! This is great!

components/sup-client/src/lib.rs
```diff
@@ -198,121 +176,4 @@ impl SrvClient {
...
-
-impl Drop for SyncSrvClient {
-    fn drop(&mut self) {
-        // This will break out of the task receiving messages
-        self.disconnect_tx
-            .take()
-            .expect("SvcClient disconnect_tx to be Some")
-            .send(())
-            .ok();
-    }
```
christophermaier: Removing all this is a really nice simplification :+1:

components/butterfly/Cargo.toml
davidMcneil: Added issue [here](https://github.com/habitat-sh/habitat/issues/7270)

components/sup/Cargo.toml
davidMcneil: Added issue [here](https://github.com/habitat-sh/habitat/issues/7271)

components/sup/src/ctl_gateway/server.rs
```diff
@@ -18,7 +18,6 @@ use futures::{channel::mpsc,
               ready,
               task::{Context,
                      Poll}};
-use habitat_common::liveliness_checker;
```
christophermaier: This makes sense, as this is moving into the main multi-threaded reactor. It no longer holds to the assumption the liveness checker is based on, which is that the code in question is always executing in the same thread.

davidMcneil: Slightly related. I wanted to use our liveness checker in the threads spawned by tokio. I was close to getting it to work using the [`on_thread_start`](https://docs.rs/tokio/0.2.4/tokio/runtime/struct.Builder.html#method.on_thread_start) and [`on_thread_stop`](https://docs.rs/tokio/0.2.4/tokio/runtime/struct.Builder.html#method.on_thread_stop) methods of the `Builder`. I wanted to spawn a never ending task that would periodically call `mark_thread_alive` using [`LocalSet`](https://docs.rs/tokio/0.2.4/tokio/task/struct.LocalSet.html). Unfortunately, the way `LocalSet` works you can only spawn task by blocking the runtime so this ultimately did not work. Maybe it will be possible one day?

components/sup/src/ctl_gateway/server.rs
```diff
@@ -141,91 +142,63 @@ impl CtlCommand {
...
+            debug!("No handshake");
+            return Err(HandlerError::from(io::Error::from(io::ErrorKind::ConnectionAborted)));
+        } else if !message.is_transaction() {
+            return Err(HandlerError::from(io::Error::from(io::ErrorKind::ConnectionAborted)));
+        } else {
+            match message.parse::<protocol::ctl::Handshake>() {
+                Ok(decoded) => {
+                    trace!("Received handshake, {:?}", decoded);
+                    let decoded_key = decoded.secret_key.unwrap_or_default();
+                    crypto::secure_eq(decoded_key, secret_key)
```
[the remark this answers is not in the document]
davidMcneil: Yep, that is much better. Addressed in [e9305e](https://github.com/habitat-sh/habitat/pull/7256/commits/e9305ebf5782036ede72505e862e4ea3525669f9).

components/sup/src/ctl_gateway/server.rs
```diff
@@ -473,34 +460,32 @@ struct SrvState {
...
+          let server = async {
+              let mut listner =
+                  TcpListener::bind(&listen_addr).await
+                                                 .expect("Could not bind ctl gateway listen \
+                                                          address!");
+              let mut incoming = listner.incoming();
+              while let Some(tcp_stream) = incoming.next().await {
+                  match tcp_stream {
+                      Ok(tcp_stream) => {
+                          let addr = tcp_stream.peer_addr().expect("Couldn't get peer address!");
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [e9305e](https://github.com/habitat-sh/habitat/pull/7256/commits/e9305ebf5782036ede72505e862e4ea3525669f9).

components/sup-client/src/lib.rs
davidMcneil: The server should handle it, but it would probably be good to add. Addressed in [e9305e](https://github.com/habitat-sh/habitat/pull/7256/commits/e9305ebf5782036ede72505e862e4ea3525669f9).

PR #7259, opened 2019-12-04
Fix setting SSL_CERT_FILE in studio when the cert isn't cached

### LINKED ISSUE #7219: `hab studio enter` sets SSL_CERT_FILE environment variable to an invalid value
On hab 0.90.6, `hab studio enter` will set the `SSL_CERT_FILE` environment variable inside the studio despite it not being set in the calling user's environment.  The value of the variable may be set to point at a file that does not exist inside the studio, resulting in various failures.

This happens under the following conditions:

1) User does NOT have `SSL_CERT_FILE` already set
2) User is starting a non-docker studio
3) User is starting the studio for the first time, requiring the installation of core/hab-studio
4) The host filesystem contains one of a number preconfigured cert file paths present in the openssl-probe crate.

Here is what I believe is likely happening:

1. In the non-docker branch we install the hab-studio package if it is not already installed:

https://github.com/habitat-sh/habitat/blob/0.90.6/components/hab/src/command/studio/enter.rs#L167
https://github.com/habitat-sh/habitat/blob/0.90.6/components/hab/src/exec.rs#L73

2. In order to make the HTTPS request, this code uses our http client which currently uses the native-tls crate: https://github.com/habitat-sh/habitat/blob/0.90.6/Cargo.lock#L1248-L1262

3. The native-tls crate calls an initialization function which sets `SSL_CERT_FILE` if it is not already set and if one of its pre-defined cert locations is found:

https://github.com/sfackler/rust-native-tls/blob/v0.2.3/src/imp/openssl.rs#L92
https://github.com/alexcrichton/openssl-probe/blob/0.1.2/src/lib.rs#L50

4. The paths found in (3) refer to paths on the studio host which might not be available inside the studio. Further note that (3) will happen _after_ we've attempted to copy the file from the host into the studio, since the library only modifies the environment on creation of the first TLSConnector

https://github.com/habitat-sh/habitat/blob/0.90.6/components/hab/src/command/studio/enter.rs#L121-L126
https://github.com/sfackler/rust-native-tls/blob/v0.2.3/src/imp/openssl.rs#L251

5. Finally, even if that path happened to be mounted into the studio, we munge the SSL_CERT_FILE again to point at the /hab/cache/ssl dir under the assumption that we have copied the file there: 

https://github.com/habitat-sh/habitat/blob/0.90.6/components/studio/bin/hab-studio.sh#L887

smacfarlane:
Closes #7219 

Thanks to @stevendanna for the extremely detailed write up in that issue.  

The original intent behind the munging and passing of `SSL_CERT_FILE` into the studio was for the self-signed-cert case, where a user may have it set in order to interact with internal services such as  vcs, or artifact repositories. The assumption being that if a user has it set external to the studio, they will likely need it set inside the studio as well.  

This change attempts to preserve that intent, by only setting `SSL_CERT_FILE` if the the file was cached.  Windows doesn't appear to have the same issue Linux does due to how the ssl-probe crate works, however it's a good check to make anyways.

Testing this is somewhat of a pain, due to the nature of how it is triggered.  I validated this by standing up a local builder instance and building and uploading a package with this change to that local instance.

### DIFF
--- a/test/end-to-end/test_studio_when_ssl_cert_file_isnt_cached.sh
+++ b/test/end-to-end/test_studio_when_ssl_cert_file_isnt_cached.sh
@@ -19,7 +19,9 @@ rm -f ~/.hab/cache/ssl/*
 # the auto-installation of the studio on first-run triggers
 # the behavior we're attempting to test
 echo "--- Uninstalling any existing studio packages"
-hab pkg uninstall core/hab-studio || true
+while hab pkg uninstall core/hab-studio; do 
+  :; # no-op
+done
 
 echo "--- Test SSL_CERT_FILE remains unset inside the studio"
 hab studio run "echo \$SSL_CERT_FILE && test ! -v SSL_CERT_FILE"
--- a/test/end-to-end/test_studio_when_ssl_cert_file_isnt_cached.sh
+++ b/test/end-to-end/test_studio_when_ssl_cert_file_isnt_cached.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Test that when SSL_CERT_FILE isn't cached, the variable isn't 
+# set on the inside of the studio.  
+# https://github.com/habitat-sh/habitat/issues/7219
+
+set -euo pipefail
+
+source .expeditor/scripts/end_to_end/shared_end_to_end.sh
+ 
+echo "--- Generating a signing key"
+hab origin key generate "$HAB_ORIGIN"
+
+echo "--- Removing any existing cached certificates"
+rm -f /hab/cache/ssl/*
+rm -f ~/.hab/cache/ssl/*
+
+# This must be run immediatly before `studio run`, as 
+# the auto-installation of the studio on first-run triggers
+# the behavior we're attempting to test
+echo "--- Uninstalling any existing studio packages"
+hab pkg uninstall core/hab-studio || true
+
+echo "--- Test SSL_CERT_FILE remains unset inside the studio"
+hab studio run "echo \$SSL_CERT_FILE && test ! -v SSL_CERT_FILE"
+
--- a/test/end-to-end/test_studio_with_ssl_cert_file_envvar_set.ps1
+++ b/test/end-to-end/test_studio_with_ssl_cert_file_envvar_set.ps1
@@ -125,12 +125,15 @@ Context "SSL_CERT_FILE isn't set" {
         Cleanup-CachedCertificate
         # Ensure SSL_CERT_FILE isn't set
         Remove-Item Env:\SSL_CERT_FILE
-    }
-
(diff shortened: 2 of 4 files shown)


test/end-to-end/test_studio_with_ssl_cert_file_envvar_set.ps1
```diff
@@ -21,7 +21,7 @@ function New-TemporaryDirectory {
 }
 
 Write-Host "--- Generating a signing key"
-hab origin key generate "$env:HAB_ORIGIN"
+hab origin key generate "$HAB_ORIGIN"
```
mwrock: Do we really want to treat this as a local variable and not an environment variable? I don't see where this variable is bwing set.

smacfarlane: Corrected in 97cb2e787f6ed2015fc99886dcf941b219d0e18d

test/end-to-end/test_studio_when_ssl_cert_file_isnt_cached.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "--- Removing any existing cached certificates"
+rm -f /hab/cache/ssl/*
+rm -f ~/.hab/cache/ssl/*
+
+# This must be run immediatly before `studio run`, as 
+# the auto-installation of the studio on first-run triggers
+# the behavior we're attempting to test
+echo "--- Uninstalling any existing studio packages"
+hab pkg uninstall core/hab-studio || true
```
christophermaier: Is there any danger of there being multiple `core/hab-studio` packages on disk? `hab pkg uninstall` only removes the most recent one, if it's there.

(:thinking: I wonder if a new `hab pkg uninstall core/hab-studio --all` flag is warranted? :thinking:)

smacfarlane: I thought about this and convinced myself that, in CI at least, it should be ok. Worst case remaining studios would be from an older version which _shouldn't_ be used.    

If we're feeling extra paranoid, probably a good thing when testing the studio, I can wrap this in a while loop to remove all studios, or just `rm -rf /hab/pkgs/core/hab-studio` and call it a day.

christophermaier: Wrapping it in a while loop seems like a good way forward, as that stays true to the "don't rely on internal implementation details" spirit behind `hab pkg uninstall`. It also makes it very easy to modify in the future if / when we ever do get an `--all` option.

smacfarlane: Resolved in f1788cdb27b832a29929efb07192f36a6350ce9e

PR #7266, opened 2019-12-05
Add health check interval to event stream

### LINKED ISSUE #7226: Add health check interval to event stream message
It would be useful if the [event stream](https://github.com/habitat-sh/habitat/blob/d8c97148ed75b7cfc69f1de6b1f1855e424ba6e6/components/sup/protocols/event.proto#L88) message included the health check interval.

nellshamrell:
Fixes #7226.

### DIFF
--- a/components/sup/protocols/event.proto
+++ b/components/sup/protocols/event.proto
@@ -76,7 +76,7 @@ message ServiceStoppedEvent {
 message ServiceUpdateStartedEvent {
   EventMetadata event_metadata = 1;
   ServiceMetadata service_metadata = 2;
-  // Details of the service pkg to which 
+  // Details of the service pkg to which
   // we're attempting to update.This is
   // implemented as a string for now,
   // but at some point we'll may want
@@ -98,4 +98,6 @@ message HealthCheckEvent {
   google.protobuf.StringValue stdout = 6;
   // The health check hook stderr output
   google.protobuf.StringValue stderr = 7;
+  // The heath check interval
+  google.protobuf.Duration interval = 8;
 }
--- a/components/sup/src/event.rs
+++ b/components/sup/src/event.rs
@@ -16,6 +16,7 @@ mod stream;
 mod types;
 
 pub(crate) use self::types::ServiceMetadata;
+use prost_types::Duration as ProstDuration;
 use self::types::{EventMessage,
                   EventMetadata,
                   HealthCheckEvent,
@@ -37,7 +38,8 @@ use habitat_common::types::{AutomateAuthToken,
                             EventStreamConnectMethod,
                             EventStreamMetadata,
                             EventStreamServerCertificate};
-use habitat_core::package::ident::PackageIdent;
+use habitat_core::{package::ident::PackageIdent,
+                   service::HealthCheckInterval};
 use parking_lot::Mutex;
 use state::Container;
 use std::{net::SocketAddr,
@@ -223,7 +225,8 @@ pub fn service_update_started(service: &Service, update: &PackageIdent) {
 // currently works. Revisit when async/await + Pin is all stabilized.
 pub fn health_check(metadata: ServiceMetadata,
                     health_check_result: HealthCheckResult,
-                    health_check_hook_status: HealthCheckHookStatus) {
+                    health_check_hook_status: HealthCheckHookStatus,
+                    health_check_interval: HealthCheckInterval) {
     if stream_initialized() {
         let health_check_result: types::HealthCheckResult = health_check_result.into();
         let maybe_duration = health_check_hook_status.maybe_duration();
@@ -233,6 +236,9 @@ pub fn health_check(metadata: ServiceMetadata,
         let StandardStreams { stdout, stderr } =
             maybe_process_output.map(ProcessOutput::standard_streams)
                                 .unwrap_or_default();
+
+        let prost_interval = ProstDuration::from(Duration::from(health_check_interval));
+
         publish(HEALTHCHECK_SUBJECT,
                 HealthCheckEvent { service_metadata: Some(metadata),
                                    event_metadata: None,
@@ -240,7 +246,8 @@ pub fn health_check(metadata: ServiceMetadata,
                                    execution: maybe_duration.map(Duration::into),
                                    exit_status,
                                    stdout,
-                                   stderr });
+                                   stderr,
+                                   interval: Some(prost_interval) });
     }
 }
 
@@ -323,6 +330,7 @@ mod tests {
                   sync::mpsc as futures_mpsc};
     #[cfg(windows)]
     use habitat_core::os::process::windows_child::ExitStatus;
+    use habitat_core::service::HealthCheckInterval;
     #[cfg(unix)]
     use std::{os::unix::process::ExitStatusExt,
               process::ExitStatus};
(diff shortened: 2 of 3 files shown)


components/sup/protocols/event.proto
```diff
@@ -98,4 +98,5 @@ message HealthCheckEvent {
   google.protobuf.StringValue stdout = 6;
   // The health check hook stderr output
   google.protobuf.StringValue stderr = 7;
+  google.protobuf.UInt64Value interval = 8;
```
davidMcneil: Because the healthcheck will always have an interval associated with it, I think it would be better to use the primitive `uint64` type here. This will avoid the need to wrap the value in an `Option` in the rust code. It also might be worth adding a comment stating the units of the interval value.

components/sup/src/event.rs
```diff
@@ -233,14 +235,17 @@ pub fn health_check(metadata: ServiceMetadata,
         let StandardStreams { stdout, stderr } =
             maybe_process_output.map(ProcessOutput::standard_streams)
                                 .unwrap_or_default();
+        let interval: u64 = health_check_interval.into();
```
davidMcneil: It is generally preferred to leave off explicit type definitions in Rust. The compiler should be able to infer what type we want to convert into here.

nellshamrell: Done!

components/sup/src/event.rs
```diff
@@ -378,6 +389,10 @@ mod tests {
         assert_eq!(event.stdout, None);
         assert_eq!(event.stderr, None);
 
+        let interval_in_seconds: u64 = HealthCheckInterval::default().into();
```
nellshamrell: If I do not specify the type here, the compiler throws a type annotations error and suggests giving the "interval in seconds" variable a type, so I have done so.

components/sup/protocols/event.proto
nellshamrell: Done!

components/sup/src/event.rs
nellshamrell: Done!

nellshamrell:
Thanks for the review, @davidMcneil - I believe I have addressed all of your requests for changes :)

components/sup/src/event.rs
```diff
@@ -233,14 +235,17 @@ pub fn health_check(metadata: ServiceMetadata,
...
         publish(HEALTHCHECK_SUBJECT,
                 HealthCheckEvent { service_metadata: Some(metadata),
                                    event_metadata: None,
                                    result: i32::from(health_check_result),
                                    execution: maybe_duration.map(Duration::into),
                                    exit_status,
                                    stdout,
-                                   stderr });
+                                   stderr,
+                                   interval: interval });
```
chefsalim: Minor - I think you can just directly specify the interval as a param without the "interval:" prefix..

raskchanky: Yeah, similar to `exit_status`, `stdout`, and `stderr`.

https://doc.rust-lang.org/book/ch05-01-defining-structs.html#using-the-field-init-shorthand-when-variables-and-fields-have-the-same-name

components/sup/protocols/event.proto
```diff
@@ -98,4 +98,6 @@ message HealthCheckEvent {
   google.protobuf.StringValue stdout = 6;
   // The health check hook stderr output
   google.protobuf.StringValue stderr = 7;
+  // The heath check interval in seconds
+  uint64 interval = 8;
```
christophermaier: This might be better as a `google.protobuf.Duration`, since that removes any ambiguity around the units (we also use this above for the `execution` field).

components/sup/src/event.rs
christophermaier: As an FYI, running `clippy` locally (using `make lint` at the top level of the repo) will catch things like this.

components/sup/src/event.rs
christophermaier: Changing the type of the interval field in the message could also affect this.

components/sup/src/event.rs
nellshamrell: ty!

components/sup/protocols/event.proto
nellshamrell: Good point - will get that in there :)

components/core/src/service.rs
```diff
@@ -386,6 +386,10 @@ impl Into<u64> for HealthCheckInterval {
     fn into(self) -> u64 { self.0.as_secs() }
 }
 
+impl Into<Option<Duration>> for HealthCheckInterval {
```
davidMcneil: I think this can be removed in favor of the existing conversion [here](https://github.com/habitat-sh/habitat/blob/c1e104258c73ddbace9c567deda347f46583dfac/components/core/src/service.rs#L405).

components/sup/src/event.rs
```diff
@@ -233,14 +238,19 @@ pub fn health_check(metadata: ServiceMetadata,
         let StandardStreams { stdout, stderr } =
             maybe_process_output.map(ProcessOutput::standard_streams)
                                 .unwrap_or_default();
+
+        let time_interval: Option<Duration> = health_check_interval.into();
+        let prost_interval = prost_duration::from(time_interval.unwrap());
```
davidMcneil: Using the [`From` trait here](https://github.com/habitat-sh/habitat/blob/c1e104258c73ddbace9c567deda347f46583dfac/components/core/src/service.rs#L405), this could be better expressed as:

```rust
let prost_interval = prost_duration::from(Duration::from(health_check_interval));
```

components/sup/src/event.rs
```diff
@@ -378,6 +394,12 @@ mod tests {
         assert_eq!(event.stdout, None);
         assert_eq!(event.stderr, None);
 
+        let default_interval: Option<Duration> = HealthCheckInterval::default().into();
+        let prost_interval = prost_duration::from(default_interval.unwrap());
```
davidMcneil: Same as above.

components/sup/protocols/event.proto
```diff
@@ -98,4 +98,6 @@ message HealthCheckEvent {
   google.protobuf.StringValue stdout = 6;
   // The health check hook stderr output
   google.protobuf.StringValue stderr = 7;
+  // The heath check interval in seconds
```
davidMcneil: Minor - You could remove the "in seconds" now since `protobuf.Duration` has its own concept of units. I think that goes down to nanoseconds.

components/sup/src/event.rs
```diff
@@ -16,6 +16,8 @@ mod stream;
 mod types;
 
 pub(crate) use self::types::ServiceMetadata;
+use prost_types::Duration as prost_duration;
```
davidMcneil: Rust types are generally represented in pascal case. I think `ProstDuration` would be better.

components/sup/src/event.rs
```diff
@@ -16,6 +16,8 @@ mod stream;
...
+
```
davidMcneil: Extra newline. If we keep our imports together `rustfmt` will automatically sort them.

components/sup/src/event.rs
```diff
@@ -37,7 +39,9 @@ use habitat_common::types::{AutomateAuthToken,
                             EventStreamConnectMethod,
                             EventStreamMetadata,
                             EventStreamServerCertificate};
-use habitat_core::package::ident::PackageIdent;
+use habitat_core::{package::ident::PackageIdent,
+                   service::HealthCheckInterval};
+
```
davidMcneil: Extra newline.

components/sup/src/event.rs
```diff
@@ -321,6 +331,8 @@ mod tests {
     use futures::{future::Future,
                   stream::Stream,
                   sync::mpsc as futures_mpsc};
+    use habitat_core::service::HealthCheckInterval;
+
```
davidMcneil: Extra newline.

nellshamrell: I'm not sure if it's a Rust formatter issue - but when I do not include a new line after that line it tells me it should go under #[cfg(windows)] - which feels incorrect

Screen Shot 2019-12-18 at 11 24 02 AM

davidMcneil: Agreed it looks odd, but it is actually ok. `#[cfg(windows)]` only applies to the immediately followed expression (in this case the line `use habitat_core::os::process::windows_child::ExitStatus;`).

nellshamrell: Ah! Ty!

nellshamrell:
@davidMcneil - believe I got all of your suggestions incorporated in. Ty!

PR #7274, opened 2019-12-09
mirror RUSTFMT_VERSION in RUST_NIGHTLY_VERSION

jeremymv2:
This PR leverages last week's [work](https://github.com/habitat-sh/habitat/pull/7258) which auto bumps the version in `RUSTFMT_VERSION` following a new release and mirrors the same version in `RUST_NIGHTLY_VERSION`.

For all intents and purposes, `RUST_NIGHTLY_VERSION` holds the toolchain version used by CI for running our Unit Test suites via `cargo test`.  Since our automation already selects a nightly toollchain version usable for running `rustfmt` we opt to use that same version for unit tests in order to keep complexity low.

Keeping the two files distinctly separate still allows us the flexibility to decouple the toolchain used by either `rustfmt` or `cargo test` should that need ever arise.

### DIFF
--- a/.expeditor/scripts/verify/run_cargo_test.ps1
+++ b/.expeditor/scripts/verify/run_cargo_test.ps1
@@ -28,6 +28,10 @@ If($Features) {
 
 # Set cargo test invocation
 Install-Rustup $toolchain
+# In this context we are installing a nightly toolchain and do not
+# require any additional components, we'll set a minimal profile.
+# https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles
+rustup set profile minimal
 Install-RustToolchain $toolchain
 $CargoTestCommand = "cargo +$toolchain test $FeatureString -- $TestOptions"
 
--- a/.expeditor/scripts/verify/run_cargo_test.ps1
+++ b/.expeditor/scripts/verify/run_cargo_test.ps1
@@ -28,10 +28,6 @@ If($Features) {
 
 # Set cargo test invocation
 Install-Rustup $toolchain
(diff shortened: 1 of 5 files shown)


.expeditor/scripts/verify/run_cargo_test.ps1
```diff
@@ -28,6 +28,10 @@ If($Features) {
 
 # Set cargo test invocation
 Install-Rustup $toolchain
+# In this context we are installing a nightly toolchain and do not
+# require any additional components, we'll set a minimal profile.
+# https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles
+rustup set profile minimal
```
mwrock: We already set the profile to minimal in both the bash and powershell `Install-Rustup` and `install_rustup` functions

mwrock: If rustup was already installed then those functions leave the profile as is

jeremymv2: Great callout @mwrock, thanks!
Handled in https://github.com/habitat-sh/habitat/pull/7274/commits/abf3174a07c0aa6b13d6600db867d53044b4be92

mwrock: we should remove it from the bash script too since we do the same thing there

jeremymv2: Agreed! https://github.com/habitat-sh/habitat/pull/7274/commits/c45d279a11c63f239b58dbf903139fb2705cc55e takes care of it over there.

PR #7283, opened 2019-12-12
Add a "test-probe" service

christophermaier:
This is a small webapp intended to help test the Supervisor. Right
now, it exposes the rendering context it is provided over an HTTP
endpoint, allowing a view into the Supervisor's inner workings from
outside.

It also has a handful of lifecycle hook implementations, some of which
take a long time to run. This has been handy for verifying async service
behavior.

This service could be augmented in the future to do other things, as
well.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -573,6 +573,19 @@ steps:
 # still build. - Linux
 #######################################################################
 
+  - label: "[build] :linux: test-probe"
+    env:
+      HAB_LICENSE: "accept-no-persist"
+    command:
+      - hab pkg build test-services/test-probe
+    expeditor:
+      executor:
+        docker:
+          privileged: true
+    retry:
+      automatic:
+        limit: 1
+
   - label: "[build] :linux: backline"
     env:
       HAB_LICENSE: "accept-no-persist"
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -577,10 +577,10 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - hab pkg build test-services/test-probe
+      - .expeditor/scripts/verify/build_package.sh test-services/test-probe
     expeditor:
       executor:
-        docker:
+        linux:
           privileged: true
     retry:
       automatic:
@@ -590,7 +590,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh backline
+      - .expeditor/scripts/verify/build_package.sh components/backline
     expeditor:
       executor:
         linux:
@@ -603,7 +603,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh hab
+      - .expeditor/scripts/verify/build_package.sh components/hab
     expeditor:
       executor:
         linux:
@@ -616,7 +616,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh launcher
+      - .expeditor/scripts/verify/build_package.sh components/launcher
     expeditor:
       executor:
         linux:
@@ -629,7 +629,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh pkg-aci
+      - .expeditor/scripts/verify/build_package.sh components/pkg-aci
     expeditor:
       executor:
         linux:
@@ -642,7 +642,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh pkg-cfize
+      - .expeditor/scripts/verify/build_package.sh components/pkg-cfize
     expeditor:
       executor:
         linux:
@@ -655,7 +655,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh pkg-dockerize
+      - .expeditor/scripts/verify/build_package.sh components/pkg-dockerize
     expeditor:
       executor:
         linux:
@@ -668,7 +668,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh pkg-export-docker
+      - .expeditor/scripts/verify/build_package.sh components/pkg-export-docker
     expeditor:
       executor:
         linux:
@@ -681,7 +681,7 @@ steps:
     env:
       HAB_LICENSE: "accept-no-persist"
     command:
-      - .expeditor/scripts/verify/build_component.sh pkg-export-helm
+      - .expeditor/scripts/verify/build_package.sh components/pkg-export-helm
     expeditor:
       executor:
         linux:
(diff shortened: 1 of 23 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -573,6 +573,18 @@ steps:
 # still build. - Linux
 #######################################################################
 
+  - label: "[build] :linux: test-probe"
+    env:
+      HAB_LICENSE: "accept-no-persist"
+    command:
+      - hab pkg build -D test-services/test-probe
```
smacfarlane: This will put you in a docker-in-docker situation.  I'd recommend removing the `-D` here and using 
```
expeditor:
  executor:
    docker:
      privileged: true
```

Or, you can use a linux executor, but it's probably a better idea to use the docker executors unless otherwise needed.

test-services/test-probe/src/main.rs
```diff
@@ -0,0 +1,151 @@
...
+            .route("context", web::get().to(context::show))
+            .route("context/{path:.*}", web::get().to(context::show))
+    }).workers(1)
+        .bind(&bind_addr)?;
+
+    println!("Starting test-probe on {}", bind_addr);
+    server.run()
+}
+
+// fn main() {
```
smacfarlane: If this commented code was left here on purpose,  we should probably leave a note as to why. Otherwise it might be good to remove or un-comment it.

test-services/test-probe/habitat/plan.sh
```diff
@@ -0,0 +1,46 @@
...
+        cargo build --target="$rustc_target" --verbose
+    )
+}
+
+do_install() {
+  install -v -D "$CARGO_TARGET_DIR/$rustc_target/debug/$bin" \
+    "$pkg_prefix/bin/$bin"
+}
+
+do_strip() {
```
smacfarlane: Convention from core-plans is to `return 0` if we need to skip a step.  I know this is equivalent, but I read this as `do_strip: true` the first time, which made me pause `(:thinking: Isn't that the default behavior?)`

.expeditor/verify.pipeline.yml
christophermaier: :man_facepalming: Good call.

test-services/test-probe/src/main.rs
christophermaier: Dang, I though I cleaned up after myself... will fix.

test-services/test-probe/habitat/plan.sh
christophermaier: Ah, I hadn't considered that possible reading; I'll change it back.

test-services/test-probe/src/config.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::{fs,
+          path::PathBuf};
+use toml;
+
+pub const DEFAULT_JSON_SOURCE: &str = "/hab/svc/test-probe/config/render_context_file.json";
+pub const DEFAULT_HOST: &str = "0.0.0.0";
+pub const DEFAULT_PORT: u16 = 8000;
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Config {
```
davidMcneil: If you wanted an excuse to play with `structopt` it would simplify some of this boilerplate, but certainly not necessary.

test-services/test-probe/habitat/hooks/init
```diff
@@ -0,0 +1,15 @@
+#!{{pkgPathFor "core/bash"}}/bin/bash
+
+echo "Initializing package {{pkg.ident}} (PID: $$, PPID: $PPID, PGID: $(ps h -o pgid -p $$))"
+sleep 1
+echo "... reticulating splines ..."
+echo "... integrating curves ..."
+echo "... relaxing splines ..."
+echo "... calculating inverse probability matrices ..."
```
davidMcneil: Nice! :laughing:

test-services/test-probe/src/context.rs
```diff
@@ -0,0 +1,72 @@
...
+fn read_json<P>(path: P) -> Result<serde_json::Value>
+    where P: AsRef<Path>
+{
+    let contents = fs::read_to_string(path)?;
+    serde_json::from_str(&contents).map_err(|e| e.into())
+}
+
+/// Recursively descend through a JSON structure, based on the path
+/// provided in `keys`
+fn lookup(data: serde_json::Value, keys: Components) -> Result<Option<serde_json::Value>> {
```
davidMcneil: This is clever! Could we use [`pointer`](https://docs.serde.rs/serde_json/enum.Value.html#method.pointer)?

test-services/test-probe/src/context.rs
```diff
@@ -0,0 +1,72 @@
...
+                HttpRequest,
+                HttpResponse};
+use serde_json;
+use std::{fs,
+          path::{Components,
+                 Path,
+                 PathBuf}};
+
+/// Show render context rooted at `path`
+pub fn show(config: web::Data<config::Config>,
```
davidMcneil: I think this could be simplified like:
```rust
pub fn show(config: web::Data<config::Config>,
            path: web::Path<PathBuf>)
            -> actix_web::Result<HttpResponse> {
    let json =
        extract_data(&config.render_context_file, path.into_inner()).map_err(actix_web::error::ErrorBadRequest)?;
    ...
}
```

test-services/test-probe/src/context.rs
christophermaier: Whoa, nice... I didn't realize that was a thing. A pointer definitely seems more in the spirit of what I was going for.

I'll check that out!

christophermaier: _Much_ better: https://github.com/habitat-sh/habitat/pull/7283/commits/29c091c026512a46fdce46a1dde75d51895f9f11

test-services/test-probe/src/context.rs
christophermaier: Ah, neat! I think with the pointer-related changes in 29c091c, I'm not sure if this helps, but definitely worth remembering for the future.

davidMcneil: I think it would still be useful. You could pull out the pointer in the function signature with `pointer: web::Path<String>` this would avoid the need to unwrap the `Option`. It also has better static guarentees as `actix-web` will not even route a request here if it does not match the function signature.

christophermaier: Hrm... right now, I think I _do_ want it to route here if there's no path provided, in order to deal with the "show the entire context" use case.

davidMcneil: That is a good point. If you want to allow showing the full context for both `/context` and `/context/` you would have to wrap the path in an `Option`. It does work but is not an obvious improvement.

```rust
/// Show render context rooted at `path`
#[allow(clippy::needless_pass_by_value)] // Required by actix-web API
pub fn show(config: web::Data<config::Config>,
            pointer: Option<web::Path<String>>)
            -> actix_web::Result<HttpResponse> {
    // A JSON pointer of "" means "everything". Pointers starting with
    // "/" describe paths into the structure. We'll chop off any
    // trailing "/" since that will end up looking for an item with a
    // key of "", which we're just not going to deal with.
    let pointer = pointer.map(|p| p.into_inner());
    let pointer = match pointer.as_ref().map(String::as_str) {
        None | Some("") => Cow::Borrowed(""),
        Some(path) => Cow::Owned(format!("/{}", path.trim_end_matches("/"))),
    };
    let json = read_json(&config.render_context_file).map_err(actix_web::error::ErrorBadRequest)?;
    if let Some(value) = json.pointer(&pointer) {
        Ok(HttpResponse::Ok().json(value))
    } else {
        Ok(HttpResponse::NotFound().finish())
    }
}
```

PR #7292, opened 2019-12-17
Remove TOCs

kagarmoe:
Removing them at the beginning of the reorg, rather than at the end.


kagarmoe:
That's really strange.

PR #7297, opened 2019-12-18
Add hab RUST_LOG examples

kagarmoe:


www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,45 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
 ---
 ## <a name="sup-log-configuration" id="sup-log-configuration" data-magellan-target="sup-log-configuration">Supervisor Log Configuration Reference</a>
 
-With the 0.83.0 release of the Chef Habitat Supervisor, it is possible to have greater control over logging output, including the ability to dynamically adjust the logging configuration of a running Supervisor. There are two main ways of configuring logging, each of which has their own strengths and weaknesses: using environment variables, and using a configuration file.
+With the 0.83.0 release of the Chef Habitat Supervisor, it is possible to have greater control over logging output, including the ability to dynamically adjust the logging configuration of a running Supervisor. There are three ways of configuring logging, each of which has their own strengths and weaknesses: as part of the `hab` command, by using environment variables, and by using a configuration file.
 
-### Environment Variable Configuration
+### RUST_LOG with a command
+
+Use `RUST_LOG=debug RUST_BACKTRACE=1` to see an individual command's debug and backtrace.
```
mwrock: Might be worth a note that this ONLY works on mac/linux and not on Windows. However, I am actually in favor (not vehemently) of omitting this section altogether. This is really about setting command scoped environment variables which simply comes with the OS and is not Habitat specific.

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,45 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
...
+
+### RUST_LOG as an Environment Variable Configuration
+
+Use `RUST_LOG=debug RUST_BACKTRACE=1` to see an individual command's debug and backtrace.
+
+    ```bash
+    # Linux/MacOS
+    # replace "hab sup run" with your command
+    RUST_LOG=debug RUST_BACKTRACE=1 hab sup run
+    ```
```
mwrock: This seems like the content above was inadvertently duplicated?

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,43 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
...
+    $env:RUST_LOG="debug"; $env:RUST_BACKTRACE=1; hab sup run
+    ```
+
+### RUST_LOG as an Environment Variable Configuration
 
 It is still possible to configure logging via the `RUST_LOG` environment variable. This approach is often useful for quickly re-configuring logging (requiring a restart), or for easily configuring logging in container-based deployments.
 
-The configuration scheme is essentially the same as that described in the documentation of the Rust [env_logger](https://docs.rs/env_logger/0.6.1/env_logger/#enabling-logging) crate, with the exception that we do not allow additional regular expression-based filtering. The configuration values the Supervisor recognizes are described below:
+```bash
+export HAB_FEAT_EVENT_STREAM = “1”
```
smacfarlane: If this section is strictly related to logging, we probably don't need to reference the event stream feature here.

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,43 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
...
+export RUST_LOG=“debug”
```
smacfarlane: To @mwrock point above, the difference between `VAR=foo command` and `export VAR=foo; command` is an OS concern and not Habitat specific. We can probably omit one of the formats to simplify this page. I do think having one of them is useful for people to copy/paste as this is a common question. 

---

When you run a command in Bash (and I believe this is true for all shells, but I'm not 100% certain on that), the command is run in a process with its own environment. `export` allows the specified variable and its value set in the current shell to be inherited by the spawned process.

By way of example, given the following shell script `var.sh`
```
#!/bin/bash

echo "VAR is $VAR"
```
You get the following outputs based on how `VAR` is set:
```
$ echo $VAR

$ ./var.sh
VAR is
$ VAR="hello"
$ ./var.sh
VAR is
$ echo $VAR
hello
$ VAR="hello" ./var.sh
VAR is hello
$ export VAR="hi"
$ ./var.sh
VAR is hi
$ VAR="hello" ./var.sh
VAR is hello
```

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,43 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
...
+export RUST_BACKTRACE=1
+```
+
+```powershell
+$env:HAB_FEAT_EVENT_STREAM = “1”
```
smacfarlane: I'll defer to @mwrock on this, but I believe that 
```
$env:RUST_LOG="debug"; hab <command>
```
and 
```
$env:RUST_LOG="debug"
hab <command>
``` 

are functionally identical, which is a difference between Bash and Powershell.

www/source/docs/reference/reference.html.md.erb
mwrock: Note that `$env:RUST_LOG="debug"; $env:RUST_BACKTRACE=1; hab sup run` will indeed works but has some very different behavior from `RUST_LOG=debug RUST_BACKTRACE=1 hab sup run` in bash. The bash command will set those variables scoped to the hab process and will not leak them into the current shell. The powershell line will set those vars in the current shell and then the `hab` command will inherit them. the `;` s are just allowing you to run multiple single line commands on the same line. I believe the bash equivilent is `&&`. 

As mentioned above. I am in favor of removing instructions on how to set the vars on a single line because I think that could add more confusion than clarity. I think the important thing to communicate here is that `RUST_LOG` is an environment variable and setting it to `debug` will get you debug output. It is valuable to have both bash and powershell examples but I'm in favor of letting the users leverage their own OS knowledge to set those variables with different scopes.

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,43 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
 ---
 ## <a name="sup-log-configuration" id="sup-log-configuration" data-magellan-target="sup-log-configuration">Supervisor Log Configuration Reference</a>
 
-With the 0.83.0 release of the Chef Habitat Supervisor, it is possible to have greater control over logging output, including the ability to dynamically adjust the logging configuration of a running Supervisor. There are two main ways of configuring logging, each of which has their own strengths and weaknesses: using environment variables, and using a configuration file.
+With the 0.83.0 release of the Chef Habitat Supervisor, it is possible to have greater control over logging output, including the ability to dynamically adjust the logging configuration of a running Supervisor. There are three ways of configuring logging, each of which has their own strengths and weaknesses: as part of the `hab` command, by using environment variables, and by using a configuration file.
```
christophermaier: It's still only 2 ways. The "by using the hab command" way mentioned here is using environment variables.

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,43 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
...
+    RUST_LOG=debug RUST_BACKTRACE=1 hab sup run
+    ```
+
+    ```powershell
+    # Windows
+    # replace "hab sup run" with your command
+    $env:RUST_LOG="debug"; $env:RUST_BACKTRACE=1; hab sup run
+    ```
+
+### RUST_LOG as an Environment Variable Configuration
```
christophermaier: This title is misleading, since `RUST_LOG` is _always_ an environment variable.

www/source/docs/reference/reference.html.md.erb
christophermaier: Strongly agree with this :point_up:

www/source/docs/reference/reference.html.md.erb
```diff
@@ -1260,13 +1260,43 @@ Packages are installed in the `/hab/pkgs/` directory, and then further organized
...
+export RUST_BACKTRACE=1
+```
+
+```powershell
+$env:HAB_FEAT_EVENT_STREAM = “1”
+$env:RUST_LOG=“debug”
+$env:RUST_BACKTRACE=1
+```
+
+### RUST_LOG in a Configuration File
```
christophermaier: This reorganization has mixed up the original documentation.  `RUST_LOG` is never set in a configuration file; it is _always_ an environment variable. 

The original documentation laid out how to configure logging using either `RUST_LOG` _or_ the newer YAML configuration file. Internally, the value of `RUST_LOG` is essentially translated into what you could write in the config file, and is provided as a way to maintain backwards compatibility with the logging library we used before.

What follows here is explaining how you should structure the _content_ of the `RUST_LOG` variable. Below that is the explanation on how you should structure the content of the configuration file, should you choose that route.

PR #7309, opened 2019-12-20
Install Habitat Reorg

kagarmoe:


mwrock:
fixes the missing install-habitat page but has the expanded toc which I'm not sure is correct.

PR #7310, opened 2019-12-20
Migrates e2e bash tests to powershell

### LINKED ISSUE #7156: Update existing e2e tests to use powershell/pester

mwrock:
closes #7156


mwrock:
I was wondering about adding windows tests for all. I decided against it because many tests are pure rust/logic tests and adding another windows test for everything could bloat this pipeline to run for quite some time. I did add a couple for cases where there are strong windows/linux nuances like building packages in a studio. There are some that would not work on windows at all like the one that test a hup signal.

Of course sometimes it is not obvious at all if there is some bit of windows specific functionality that falls through the cracks by failing to create a windows test. So I'm totally open to revisiting this.

davidMcneil:
@mwrock Only adding the Windows variant when we know there is Windows specific functionality sounds like a good approach.

PR #7311, opened 2019-12-20
Reference Reorg

kagarmoe:


mwrock:
Hmm. Tha package-contents look good but now the side bar nave is auto expanded which seems unintentional?

PR #7321, opened 2020-01-02
fix erroneous client output when API response is HTTP 200

jeremymv2:
This change corrects misleading hab client output on a successful deletion of an origin secret.

```
$ hab origin secret delete -o jmtest foo
☒ Deleting secret foo.
✗✗✗
✗✗✗ [200 OK]
✗✗✗
```

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -463,12 +463,14 @@ impl BuilderAPIProvider for BuilderAPIClient {
 
         let path = format!("depot/origins/{}/secret/{}", origin, key_name);
 
-        // We expect NO_CONTENT, because the origin must be empty to delete
+        // Originally, we only returned an Ok result if the response was StatusCode::NO_CONTENT
+        // (HTTP 204). However the Bldr API appears to always have returned HTTP 200. We'll accept
+        // either as indicators of a successful operation moving forward.
         self.0
             .delete(&path)
             .bearer_auth(token)
             .send()?
-            .ok_if(&[StatusCode::OK])
+            .ok_if(&[StatusCode::NO_CONTENT, StatusCode::OK])
     }
 
     /// Check an origin exists
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -468,7 +468,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .delete(&path)
             .bearer_auth(token)
             .send()?
-            .ok_if(&[StatusCode::NO_CONTENT])
+            .ok_if(&[StatusCode::OK])
     }
 
     /// Check an origin exists


mwrock:
Does a `200` get returned only if the origin is not empty? After the `200` is the origin deleted?

jeremymv2:
> Does a 200 get returned only if the origin is not empty? After the 200 is the origin deleted?

@mwrock  The origin itself isn't deleted in this case. This PR just deals with origin secrets 😃 Regardless, the 200 only gets returned when the secret was [successfully](https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/src/server/resources/origins.rs#L665) deleted from the [origin_secrets::table](https://github.com/habitat-sh/builder/blob/master/components/builder-db/src/models/secrets.rs#L59-L67).

components/builder-api-client/src/builder.rs
```diff
@@ -463,12 +463,14 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
-        // We expect NO_CONTENT, because the origin must be empty to delete
+        // Originally, we only returned an Ok result if the response was StatusCode::NO_CONTENT
+        // (HTTP 204). However the Bldr API appears to always have returned HTTP 200. We'll accept
+        // either as indicators of a successful operation moving forward.
         self.0
             .delete(&path)
             .bearer_auth(token)
             .send()?
-            .ok_if(&[StatusCode::NO_CONTENT])
+            .ok_if(&[StatusCode::NO_CONTENT, StatusCode::OK])
```
christophermaier: Did it ever return `NO_CONTENT`? If not, would there be any downside to just accepting `StatusCode::OK`?

It may be better to leverage [is_success()](https://docs.rs/hyper/0.13.1/hyper/struct.StatusCode.html#method.is_success) instead and cover all our bases in a general way.

jeremymv2: The history is hard telling exactly since the Bldr API code has been greatly refactored several times. Also, the exact type of success response that is appropriate here is somewhat debatable. Many, if not most of the Bldr API endpoint responses are 204, yet this one is 200 (but with an empty body via `HttpResponse::Ok().finish()`. This ambiguity between success codes for this endpoint is probably how this "issue" came to be in the first place.

While the `is_success()` idea could potentially alleviate this type of problem from cropping up again in the future, I think introducing `is_success()` should likely be applied to _all_ the client endpoint functions uniformly since it's a introduction of a new pattern in response handling and a departure from our [ok_if](https://github.com/habitat-sh/habitat/blob/718f58a19dd6c0a2e349b1b626b02fe02ca79742/components/builder-api-client/src/response.rs#L22-L27) trait. That being the case, I think we can consider that as a future improvement in another PR.

PR #7323, opened 2020-01-03
fix broken find_command unit test

mwrock:
Somehow I missed this broken test before merging #7322 caused by me allowing the `.bat` extension in the `PATHEXT` setup.


mwrock:
I ended up changing this up and using `env_locked_var` to lock down the `PATHEXT` variable. #7322 falsely went green because this was not locked. Note that I had to move the macro from common to core.

PR #7325, opened 2020-01-13
Update to 1.40.0

davidMcneil:
Update to rust 1.40.0. There are several new clippy lints. The main ones seem to be:

[comparison_chain](https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain)
[redundant_clone](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone)
[needless_doctest_main](https://rust-lang.github.io/rust-clippy/master/index.html#needless_doctest_main)
[inefficient_to_string](https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string)


components/sup/src/manager/service/hooks.rs
```diff
@@ -102,6 +102,7 @@ impl Hook for HealthCheckHook {
         let err_path = hooks::stderr_log_path::<Self>(package_name);
         #[cfg(windows)]
         let path = pair.path.to_path_buf();
+        #[allow(clippy::redundant_clone)]
```
christophermaier: Why this exception? Is it because of the conditional compilation? If so, a brief note to that effect as a comment would be helpful.

components/pkg-export-tar/src/build.rs
```diff
@@ -102,7 +102,7 @@ impl<'a> BuildSpec<'a> {
         Ok((workdir, created_ident))
     }
 
-    fn prepare_rootfs(&self, ui: &mut UI, rootfs: &Path) -> Result<(PackageIdent)> {
```
christophermaier: Ditto

components/builder-api-client/src/builder.rs
```diff
@@ -287,7 +287,7 @@ impl BuilderAPIProvider for BuilderAPIClient {
                     (ident, target): (&PackageIdent, PackageTarget),
                     package_only: bool,
                     token: &str)
-                    -> Result<(String)> {
+                    -> Result<String> {
```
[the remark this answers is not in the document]
davidMcneil: Types can have parentheses. Most commonly seen with traits (ie `dyn (Foo + Bar + 'static)`). They happen to be meaningless in this case. To distinguish this from a one element tuple you would need to do `Result<(String,)>`.

This is actually a new compiler warning and not a clippy lint. https://github.com/rust-lang/rust/issues/64169

components/sup/src/manager/service/hooks.rs
davidMcneil: Yes, it was due to conditional compilation. I should have just switched the order the variables are used. Fixed in, [4a2616](https://github.com/habitat-sh/habitat/pull/7325/commits/4a26167ab43e0ae9c61f7e2830252a6f345e7747)

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
christophermaier: Ah, subtle! :+1:

components/sup/src/manager/service/hooks.rs
christophermaier: Nice :+1:

PR #7335, opened 2020-01-15
Remove the event stream feature flag

### LINKED ISSUE #7294: Remove the HAB_FEAT_EVENT_STREAM feature flag
This will also require adding a cli option to enable the event stream.

davidMcneil:
Resolves #7294 #7336 

- Removes the `--application` and `--environment` flags from `hab svc load` and `hab sup run` commands (#7336)
- Removes the `HAB_FEAT_EVENT_STREAM` feature flag (#7294).

The cli flags related to the event stream are:
```
--event-stream-application <EVENT_STREAM_APPLICATION>
    The name of the application for event stream purposes. This will be attached to all events generated by this
    Supervisor.
--event-stream-connect-timeout <EVENT_STREAM_CONNECT_TIMEOUT>
    How long in seconds to wait for an event stream connection before exiting the supervisor. Set to '0' to immediately start the supervisor and continue running regardless of the event stream status. [env:
    HAB_EVENT_STREAM_CONNECT_TIMEOUT=]  [default: 0]
--event-stream-environment <EVENT_STREAM_ENVIRONMENT>
    The name of the environment for event stream purposes. This will be attached to all events generated by this
    Supervisor.
--event-meta <EVENT_STREAM_METADATA>...
    An arbitrary key-value pair to add to each event generated by this Supervisor
--event-stream-server-certificate <EVENT_STREAM_SERVER_CERTIFICATE>
    The path to the event stream server's certificate in PEM format used to establish a TLS connection
--event-stream-site <EVENT_STREAM_SITE>
    The name of the site where this Supervisor is running. It is used for event stream purposes.
--event-stream-token <EVENT_STREAM_TOKEN>
    An authentication token for streaming events to an messaging server. [env: HAB_AUTOMATE_AUTH_TOKEN=]
--event-stream-url <EVENT_STREAM_URL>
    The event stream connection string (host:port) used by this Supervisor to send events to a messaging server.
```
By default no args are required meaning the event stream is disabled. If the user wishes to enable the event stream they must set the `--event-stream-url` arg. Setting this arg requires that the `--event-stream-application`, `--event-stream-environment`, and `--event-stream-token` args also be set.

**Questions**
- Should we rename `--event-stream-application` and `--event-stream-environment` to simply `--application` and `--environment` now that these flag names are available?
- Any other UX improvements?


krasnow:
@davidMcneil 
> Should we rename --event-stream-application and --event-stream-environment to simply --application and --environment now that these flag names are available?

If we drop `event-stream` from the app and environment flags, I'd want to be consistent with all the flags and drop it across the board. 

> Any other UX improvements?
It would be good to mark which of these are required to make the event stream work, and which are optional. Also, rather than saying "messaging server" we should call out Chef Automate specifically.

davidMcneil:
@krasnow Thanks for the suggestions. Here is the updated cli

```
--event-stream-application <EVENT_STREAM_APPLICATION>
    The name of the application for event stream purposes. This will be attached to all events generated by this
    Supervisor.
--event-stream-connect-timeout <EVENT_STREAM_CONNECT_TIMEOUT>
    How long in seconds to wait for an event stream connection before exiting the Supervisor. Set to '0' to
    immediately start the Supervisor and continue running regardless of the initial connection status. [env:
    HAB_EVENT_STREAM_CONNECT_TIMEOUT=]  [default: 0]
--event-stream-environment <EVENT_STREAM_ENVIRONMENT>
    The name of the environment for event stream purposes. This will be attached to all events generated by this
    Supervisor.
--event-meta <EVENT_STREAM_METADATA>...
    An arbitrary key-value pair to add to each event generated by this Supervisor.
--event-stream-server-certificate <EVENT_STREAM_SERVER_CERTIFICATE>
    The path to Chef Automate's event stream certificate in PEM format used to establish a TLS connection.
--event-stream-site <EVENT_STREAM_SITE>
    The name of the site where this Supervisor is running for event stream purposes.
--event-stream-token <EVENT_STREAM_TOKEN>
    The authentication token for connecting the event stream to Chef Automate. [env: HAB_AUTOMATE_AUTH_TOKEN=]
--event-stream-url <EVENT_STREAM_URL>
    The event stream connection string (host:port) used by this Supervisor to send events to Chef Automate. This
    enables the event stream and requires --event-stream-application, --event-stream-environment, and --event-
    stream-token also be set.
```
I did not explicitly call out which flags are optional. I think this is generally implied for cli options and is consistent with what we do elsewhere. If an option was actually required the user should get a helpful message at run time. Happy to revisit this if we think it should be handled otherwise.

I did call out `--event-stream-url` as the way to actually enable the event stream and listed the options that become required when it is enabled. The user will get a useful runtime error if any of these options are not set.

components/hab/src/cli.rs
```diff
@@ -1106,6 +1106,10 @@ pub fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
                                                                        the Supervisor startup specified by a package identifier \
                                                                        (ex: core/redis) or filepath to a Habitat Artifact \
                                                                        (ex: /home/core-redis-3.0.7-21120102031201-x86_64-linux.hart).")
+                            // TODO (DM): These flags can eventually be removed.
+                            // See https://github.com/habitat-sh/habitat/issues/7339
+                            (@arg APPLICATION: --application -a +hidden +takes_value requires[ENVIRONMENT] "DEPRECATED")
+                            (@arg ENVIRONMENT: --environment -e +hidden +takes_value requires[APPLICATION] "DEPRECATED")
```
christophermaier: Since we're effectively ignoring these now, I don't think they need to require each other anymore.

Along those lines... what happens if they're now provided, but aren't given a value?

Basically, my ideal situation would just be to loosen our constraints as much as possible, accept them if they're there, in any way, and then just ignore them.

components/hab/src/main.rs
```diff
@@ -1966,6 +1966,11 @@ fn ui() -> UI {
 /// Set all fields for an `SvcLoad` message that we can from the given opts. This function
 /// populates all *shared* options between `run` and `load`.
 fn svc_load_from_input(m: &ArgMatches) -> Result<sup_proto::ctl::SvcLoad> {
+    // TODO (DM): This check can eventually be removed.
+    // See https://github.com/habitat-sh/habitat/issues/7339
+    if m.is_present("APPLICATION") || m.is_present("ENVIRONMENT") {
+        ui().warn("--application and --environment flags are deprecated.")?;
```
christophermaier: Also mentioning that they will be completely ignored would be a nice addition.

davidMcneil: Addressed in [20d99a](https://github.com/habitat-sh/habitat/pull/7335/commits/20d99aa36b4dfb14b5f615107863a756b9e18d41)

components/hab/src/cli.rs
davidMcneil: Good thought. It looks like making the arg be multiple achieves the behavior we want.

Addressed in [20d99a](https://github.com/habitat-sh/habitat/pull/7335/commits/20d99aa36b4dfb14b5f615107863a756b9e18d41)

christophermaier: Perfect

PR #7342, opened 2020-01-16
Update to Windows Server 2012

kagarmoe:
MS Windows Server 2008 r2 is EOL. This changes our minimum version to Windows Server 2012 (no revision, because we support both 2012 and 2012r2)


mwrock:
Your latest commits looks good but I just noticed that https://github.com/habitat-sh/habitat/blob/master/www/source/partials/global/_package_platform_support.html.md.erb also needs changing.

kagarmoe:
I think it is in the last commit?

Issue #7340, opened 2020-01-16
Fix https://www.habitat.sh/docs/examples/kubernetes_hab_bastion/


kagarmoe:

kagarmoe:
Fixed in https://github.com/habitat-sh/habitat/issues/7340

PR #7344, opened 2020-01-16
Habitat Pattern Library

kagarmoe:
This PR introduces a pattern library for Chef Habitat examples. I'm ambivalent about supporting code examples because they can be high-maintenance. I also anticipate the feedback of, "This doesn't work for my situation, how do I make it work?" However, this is something that users have asked for frequently. Ultimately, the benefits to our users may, in the long run, outweigh the burden of collecting, updating, and managing a public pattern library.


kagarmoe:
See issue: https://github.com/habitat-sh/habitat/issues/7340

www/data/docs_sidebar.yml
```diff
@@ -204,5 +204,14 @@ sidebar_links:
       link: "/docs/internals/internals/#crypto-internals"
     - title: Bootstrapping Chef Habitat
       link: "/docs/internals/internals/#bootstrap-internals"
+  - title: Habitat Pattern Library
+    link: "/docs/pattern-library/pattern-library"
+    sub_links:
+    - title: Plan.sh Pattern
```
smacfarlane: I think it might be worth grouping these under a Kubernetes heading to indicate what the purpose is.  "plan.sh pattern" on its own is pretty generic and could apply to a number of things.  Perhaps: `Kubernetes bastion service plan.sh`?

krasnow:
This looks great @kagarmoe! Thanks for putting this together. It sounds like the next Pattern might be "Testing Habitat Packages."

bixu:
❤️ this

PR #7349, opened 2020-01-17
hab origin invitations subcommand

jeremymv2:
This implements Habitat Builder origin invitation management via the hab CLI

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -17,8 +17,10 @@ use crate::{error::{Error,
             OriginKeyIdent,
             OriginSecret,
             Package,
+            PendingOriginInvitationsResponse,
             ReverseDependencies,
-            SchedulerResponse};
+            SchedulerResponse,
+            UserOriginInvitationsResponse};
 use broadcast::BroadcastWriter;
 use percent_encoding::{percent_encode,
                        AsciiSet,
@@ -544,6 +546,198 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .bearer_auth(token)
             .send()?
             .ok_if(&[StatusCode::NO_CONTENT])
+    }
+
+    /// Accepts an origin member invitation
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
+    ///    * https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/doc/api.raml#L883
+    ///
+    ///  # Arguments
+    ///    * Origin: &str - origin name where membership invitation exists
+    ///    * Token: &str - bearer token for authentication/authorization
+    ///    * Invitation Id: u64 - id of invitation to accept
+    ///
+    ///  # Expected API response on success
+    ///    * HTTP 204
+    ///
+    ///  # Return
+    ///    * Result<()>
+    fn accept_origin_invitation(&self,
+                                origin: &str,
+                                token: &str,
+                                invitation_id: u64)
+                                -> Result<()> {
+        debug!("Accepting invitation id {} in origin {}",
+               invitation_id, origin);
+
+        let path = format!("depot/origins/{}/invitations/{}", origin, invitation_id);
+
+        self.0
+            .put(&path)
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::NO_CONTENT])
+    }
+
+    /// Marks an origin member invitation ignored
+    ///
+    ///  After ignoring, the user will no longer see the invitation
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
+    ///    * https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/doc/api.raml#L903
+    ///
+    ///  # Arguments
+    ///    * Origin: &str - origin name where membership invitation exists
+    ///    * Token: &str - bearer token for authentication/authorization
+    ///    * Invitation Id: u64 - id of invitation to accept
+    ///
+    ///  # Expected API response on success
+    ///    * HTTP 204
+    ///
+    ///  # Return
+    ///    * Result<()>
+    fn ignore_origin_invitation(&self,
+                                origin: &str,
+                                token: &str,
+                                invitation_id: u64)
+                                -> Result<()> {
+        debug!("Marking invitation {} in origin {} ignored",
+               invitation_id, origin);
+
+        let path = format!("depot/origins/{}/invitations/{}/ignore",
+                           origin, invitation_id);
+
+        self.0
+            .put(&path)
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::NO_CONTENT])
+    }
+
+    /// Retrieves origin member invitations for current user
+    ///
+    ///  The usecase is for any user to discover any origin invitations sent to their
+    ///  account without having to know the origin.
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
+    ///    * https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/doc/api.raml#L614
+    ///
+    ///  # Arguments
+    ///    * Token: &str - bearer token for authentication/authorization
+    ///
+    ///  # Expected API response on success
+    ///    * HTTP 200
+    ///
+    ///  # Return
+    ///    * Result<UserOriginInvitationsResponse>
+    fn list_user_invitations(&self, token: &str) -> Result<UserOriginInvitationsResponse> {
+        let path = "user/invitations";
+
+        let mut resp = self.0.get(&path).bearer_auth(token).send()?;
+        resp.ok_if(&[StatusCode::OK])?;
+
+        Ok(resp.json()?)
+    }
+
+    /// Retrieves pending origin member invitations
+    ///
+    ///  The usecase here is for an origin owner (or simply origin member) to see the outstanding
+    ///  invitations for a given origin.
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
+    ///    * https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/doc/api.raml#L881
+    ///
+    ///  # Arguments
+    ///    * Origin: &str - origin name where membership invitation exists
+    ///    * Token: &str - bearer token for authentication/authorization
+    ///
+    ///  # Expected API response on success
+    ///    * HTTP 200
+    ///
+    ///  # Return
+    ///    * Result<PendingOriginInvitationsResponse>
+    fn list_pending_origin_invitations(&self,
+                                       origin: &str,
+                                       token: &str)
+                                       -> Result<PendingOriginInvitationsResponse> {
+        debug!("Retrieving pending invitations in origin {}", origin);
+        let path = format!("depot/origins/{}/invitations", origin);
+
+        let mut resp = self.0.get(&path).bearer_auth(token).send()?;
+        resp.ok_if(&[StatusCode::OK])?;
+
+        Ok(resp.json()?)
+    }
+
+    /// Rescind an origin member invitation
+    ///
+    ///  Rescind an invitation that hasn't already been ignored. The invitation will be deleted.
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
+    ///    * https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/doc/api.raml#L893
+    ///
+    ///  # Arguments
+    ///    * Origin: &str - origin name where membership invitation exists
+    ///    * Token: &str - bearer token for authentication/authorization
+    ///    * Invitation Id: u64 - id of invitation to accept
+    ///
+    ///  # Expected API response on success
+    ///    * HTTP 204
+    ///
+    ///  # Return
+    ///    * Result<()>
+    fn rescind_origin_invitation(&self,
+                                 origin: &str,
+                                 token: &str,
+                                 invitation_id: u64)
+                                 -> Result<()> {
+        debug!("Rescinding invitation {} in origin {}",
+               invitation_id, origin);
+
+        let path = format!("depot/origins/{}/invitations/{}", origin, invitation_id);
+
+        self.0
+            .delete(&path)
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::NO_CONTENT])
+    }
+
+    /// Send an origin member invitation
+    ///
+    ///  Invite a user into an origin. The invite will show up under pending origin invitations
+    ///  and listed under a user's direct invitations.
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
+    ///    * https://github.com/habitat-sh/builder/blob/da72e9fb86e24d9076268b6b1c913b7531c83ed9/components/builder-api/doc/api.raml#L812
+    ///
+    ///  # Arguments
+    ///    * Origin: &str - origin name where membership invitation exists
+    ///    * Token: &str - bearer token for authentication/authorization
+    ///    * Invitee Account: &str - account name to invite into the origin
+    ///
+    ///  # Expected API response on success
+    ///    * HTTP 201
+    ///
+    ///  # Return
+    ///    * Result<()>
+    fn send_origin_invitation(&self,
+                              origin: &str,
+                              token: &str,
+                              invitee_account: &str)
+                              -> Result<()> {
+        debug!("Sending an invitation to {} for origin {}",
+               invitee_account, origin);
+
+        let path = format!("depot/origins/{}/users/{}/invitations",
+                           origin, invitee_account);
+
+        self.0
+            .post(&path)
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::CREATED])
     }
 
     /// List all secrets keys for an origin
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -268,8 +268,8 @@ pub trait Tabular {
 impl Tabular for UserOriginInvitationsResponse {
     fn as_tabbed(&self) -> Result<String> {
         let tw = tabw().padding(2).minwidth(5);
-        let mut body = Vec::new();
         if !self.0.is_empty() {
+            let mut body = Vec::new();
             body.push(String::from("Invitation Id\tOrigin Name\tAccount Name\tCreation \
                                     Date\tIgnored"));
             for invitation in self.0.iter() {
@@ -280,16 +280,18 @@ impl Tabular for UserOriginInvitationsResponse {
                                   invitation.created_at,
                                   invitation.ignored));
             }
-        };
-        tabify(tw, &body.join("\n"))
+            tabify(tw, &body.join("\n"))
+        } else {
+            Ok(String::from(""))
+        }
     }
 }
 
 impl Tabular for PendingOriginInvitationsResponse {
     fn as_tabbed(&self) -> Result<String> {
         let tw = tabw().padding(2).minwidth(5);
-        let mut body = Vec::new();
         if !self.invitations.is_empty() {
+            let mut body = Vec::new();
             body.push(String::from("Invitation Id\tAccount Name\tCreation Date\tIgnored"));
             for invitation in self.invitations.iter() {
                 body.push(format!("{}\t{}\t{}\t{}",
@@ -298,8 +300,10 @@ impl Tabular for PendingOriginInvitationsResponse {
                                   invitation.created_at,
                                   invitation.ignored));
             }
-        };
-        tabify(tw, &body.join("\n"))
+            tabify(tw, &body.join("\n"))
+        } else {
+            Ok(String::from(""))
+        }
     }
 }
 
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -24,8 +24,11 @@ use std::{fmt,
           path::{Path,
                  PathBuf}};
 
-use chrono::DateTime;
+use chrono::{DateTime,
+             Utc};
 use reqwest::IntoUrl;
+
+use tabwriter::TabWriter;
 
 pub use crate::error::{Error,
                        Result};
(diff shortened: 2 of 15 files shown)


components/hab/src/command/origin/invitations/rescind.rs
```diff
@@ -0,0 +1,50 @@
...
+                      origin.")?;
+            Err(Error::APIClient(err))
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::UNAUTHORIZED, _)) => {
+            ui.fatal("Failed to rescind the invitation!")?;
+            Err(Error::APIClient(err))
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::NOT_FOUND, _)) => {
+            ui.fatal("Failed to rescind the invitation!")?;
+            ui.fatal("The origin or invitation id (or both) does not exist.")?;
```
christophermaier: _<review_comment mode="pedantic">_
I don't think we need the "or both" bit, since we're never going to be in a situation where we would check both and have both not exist.
_</review_comment>_

components/hab/src/command/origin/invitations/send.rs
```diff
@@ -0,0 +1,50 @@
...
+                      origin.")?;
+            Err(Error::APIClient(err))
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::UNAUTHORIZED, _)) => {
+            ui.fatal("Failed to send the invitation!")?;
+            Err(Error::APIClient(err))
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::NOT_FOUND, _)) => {
+            ui.fatal("Failed to send the invitation!")?;
+            ui.fatal("The origin or the member being invitationd (or both) does not exist.")?;
```
christophermaier: "invitationd" -> "invitation"

The same "or both" observation applies here, as well.

components/hab/src/main.rs
```diff
@@ -486,6 +497,61 @@ fn sub_origin_transfer_ownership(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()>
     command::origin::transfer::start(ui, &url, &token, &origin, &account)
 }
 
+fn sub_accept_origin_invitation(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
+    let origin = m.value_of("ORIGIN").expect("required ORIGIN");
+    let invitation_id: i64 = m.value_of("INVITATION_ID")
+                              .expect("required INVITATION_ID")
+                              .parse()
+                              .expect("INVITATION_ID PARSE");
```
christophermaier: If you passed "foobar" as an invitation ID, you'd end up with a nasty crash... we should handle this error in the standard way we handle other errors.

The earlier `expect` is fine because clap won't let us even get down here if that's not set.

components/hab/src/main.rs
```diff
@@ -486,6 +497,61 @@ fn sub_origin_transfer_ownership(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()>
...
+    let token = auth_token_param_or_env(&m)?;
+    command::origin::invitations::accept::start(ui, &url, &token, &origin, invitation_id)
+}
+
+fn sub_ignore_origin_invitation(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
+    let origin = m.value_of("ORIGIN").expect("required ORIGIN");
+    let invitation_id: i64 = m.value_of("INVITATION_ID")
+                              .expect("required INVITATION_ID")
+                              .parse()
+                              .expect("INVITATION_ID PARSE");
```
christophermaier: Same observation here, and elsewhere where appropriate.

components/hab/src/command/origin/invitations/send.rs
jeremymv2: Thanks for spotting this! I corrected in https://github.com/habitat-sh/habitat/pull/7349/commits/cb0ca5bef10de78cad9190a7a17991a21cb5b7d7

components/hab/src/command/origin/invitations/rescind.rs
jeremymv2: Pedantic mode appreciated. "or both" removed in https://github.com/habitat-sh/habitat/pull/7349/commits/cb0ca5bef10de78cad9190a7a17991a21cb5b7d7

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin does not exist
+    ///  * Invitation does not exist
+    fn ignore_origin_invitation(&self,
+                                origin: &str,
+                                token: &str,
+                                invitation_id: i64)
+                                -> Result<()> {
+        debug!("Makrkig invitation {} in origin {} ignored",
```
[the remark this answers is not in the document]
jeremymv2: Oh wow, thanks for catching! I corrected in https://github.com/habitat-sh/habitat/pull/7349/commits/cb0ca5bef10de78cad9190a7a17991a21cb5b7d7

components/hab/src/main.rs
jeremymv2: I'm confused by your concern here since clap is validating the command line input via `{valid_numeric::<i64>}`. If the users passes a value that cannot be coerced from String->i64 it fails like so:

```
error: Invalid value for '<INVITATION_ID>': 'foobar' is not a valid number
```

components/builder-api-client/src/lib.rs
```diff
@@ -292,6 +310,31 @@ pub trait BuilderAPIProvider: Sync + Send {
 
     fn transfer_origin_ownership(&self, origin: &str, token: &str, account: &str) -> Result<()>;
 
+    fn accept_origin_invitation(&self, origin: &str, token: &str, invitation_id: i64)
```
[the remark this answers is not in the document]
jeremymv2: I chose a signed integer for a couple of reasons:

* That is the upstream [implementation](https://github.com/habitat-sh/builder/blob/87a4bb970dea81b7e20c1f0aae66ebdb1a4240bf/components/builder-db/src/models/invitations.rs#L19) in Builder
* I'd rather have it be a primitive type that doesn't require an allocation from the heap. As an i64 it stays on the stack and doesn't incur any avoidable overhead such as it might if were a String.

components/hab/src/command/origin/invitations/list_user.rs
```diff
@@ -0,0 +1,43 @@
...
+            PRODUCT,
+            VERSION};
+use std::io::Write;
+use tabwriter::TabWriter;
+
+pub fn start(ui: &mut UI, bldr_url: &str, token: &str) -> Result<()> {
+    let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
+
+    ui.status(Status::Discovering,
+              "member invitation sent to your account".to_string())?;
```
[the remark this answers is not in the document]
jeremymv2: Nice catch! Thank you. I fixed up in https://github.com/habitat-sh/habitat/pull/7349/commits/3568f83b05411bc9794ced2b4d3c5f4a2423b99e

components/hab/src/main.rs
jeremymv2: Response in https://github.com/habitat-sh/habitat/pull/7349#discussion_r368139365

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
             .ok_if(&[StatusCode::NO_CONTENT])
     }
 
+    /// Accepts an origin member invitation
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Invitation does not exist
```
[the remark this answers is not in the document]
jeremymv2: After pondering your concern here for a bit, I'm reluctant to enumerate all the possible failure scenarios because, frankly, they're not quantifiable.

I have a concern that the lines under `# Failures` for each client API function really aren't providing any measurable value since there are often an untold number of ways functions could fail - ie. Builder disk loss. I think it spirit, they were meant to document the API ❤️  I believe they would provide greater value simply by expressing more verbosely the signature of the function. For example: What does the function expect for arguments? What does the function return in a happy path scenario?

Thoughts?

components/hab/src/main.rs
christophermaier: Ah, I missed that it was validated elsewhere! That definitely addresses my concern.

I think I just got triggered by the presence of the `expect`. While we've kind of standardized on the `expect("required X")` pattern for the presence of required arguments, I don't think we have anything corresponding for things with validation functions. 

Perhaps something like `expect("INVITATION_ID should be valid at this point")` or something more pithy?

components/hab/src/command/origin/invitations/accept.rs
```diff
@@ -0,0 +1,38 @@
...
+    ui.status(Status::Accepting,
+              format!("invitation id {} in origin {}", invitation_id, origin))?;
+
+    match api_client.accept_origin_invitation(origin, token, invitation_id) {
+        Ok(_) => {
+            ui.status(Status::Accepted, "the invitation successfully!".to_string())
+              .or(Ok(()))
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::UNAUTHORIZED, _)) => {
+            ui.fatal("Failed to accept the invitation!")?;
```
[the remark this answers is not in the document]
jeremymv2: No, it was a result of boilerplate copy. I removed in https://github.com/habitat-sh/habitat/pull/7349/commits/086d004335a4032a282ecd8fc796864540a8bfec

components/hab/src/command/origin/invitations/ignore.rs
```diff
@@ -0,0 +1,38 @@
...
+    ui.status(Status::Ignoring,
+              format!("invitation id {} in origin {}", invitation_id, origin))?;
+
+    match api_client.ignore_origin_invitation(origin, token, invitation_id) {
+        Ok(_) => {
+            ui.status(Status::Ignored, "the invitation successfully!".to_string())
+              .or(Ok(()))
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::UNAUTHORIZED, _)) => {
+            ui.fatal("Failed to accept the invitation!")?;
```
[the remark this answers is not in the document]
jeremymv2: A result of boilerplate copy. I removed in https://github.com/habitat-sh/habitat/pull/7349/commits/086d004335a4032a282ecd8fc796864540a8bfec

components/builder-api-client/src/lib.rs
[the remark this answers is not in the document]
jeremymv2: Happy to convert to u64.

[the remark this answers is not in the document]
christophermaier: :thinking: I suspect that implementation type is due to the underlying database field and/or the ORM library Builder is using, which doesn't _necessarily_ need to dictate how we represent it at the edges of the system. However, the input is being properly validated and errors are returned to the user in as friendly a manner as possible (i.e., no panic-induced stack traces), this is fine. Anytime I see a signed numeric type when there's no apparent reason for it to be signed, it always raises an eyebrow :sweat_smile: 

And now, if you'll pardon a bit of a digression....

When I say a "real type", there are ways that you can create new types in Rust that don't carry any extra overhead. One pattern you will often see is called the "newtype" pattern, after the similar pattern from Haskell. Here's a nice example:

https://github.com/habitat-sh/habitat/blob/a4d55f980cdad74d91d56344a91d069d816399f4/components/core/src/os/process.rs#L34-L73

Under the hood, that's going just going be a `u32`, but one that has a lot of very unique and constrained behavior that actually makes it much easier to deal with this particular information. We actually have a lot of other examples in the codebase, though many are "hiding" under our [env_config! macro](https://github.com/habitat-sh/habitat/blob/b83cfd8b7e6ece6dc1017f10d54cc783c1db534e/components/core/src/env.rs). 

So creating a new type doesn't inherently mean that you're going to be consuming heap memory. This can be a very good way of efficiently modeling a particular domain, and brings a lot of type safety with it. The canonical example of this would be something like creating a `Pound` and `Kilogram` type. Both could basically be floats underneath, but a function that is expecting `Kilogram`s would never be able to take a `Pound` argument (and vice versa), thus saving you from an entire class of nasty errors. Same amount of memory as a bare float, but with lots of added information that your code can take advantage of.

(Also, this is an API client library, and isn't likely to be, say, used in super tight loops where something like repeated string allocation might end up being a true performance issue. Unless we know for sure that something is going to be an issue, I generally favor getting a good domain model in place, which can be refined later. But, as I said earlier, with Rust you can often get both! :smile:)

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+    fn list_user_invitations(&self, token: &str) -> Result<Vec<OriginInvitation>> {
+        let path = "user/invitations";
+
+        let mut resp = self.0.get(&path).bearer_auth(token).send()?;
+        resp.ok_if(&[StatusCode::OK])?;
+
+        Ok(resp.json()?)
+    }
+
+    /// Retrieves pending origin member invitations
```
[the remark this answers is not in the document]
eeyun: Agreed on minor clarification. I believe the purpose here is for an individual user identity to check any outstanding invitations available for themselves

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
jeremymv2: I have refactored the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/builder-api-client/src/builder.rs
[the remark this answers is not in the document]
jeremymv2: I have refactored the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+            .ok_if(&[StatusCode::NO_CONTENT])
+    }
+
+    /// Marks an origin member invitation ignored
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin does not exist
+    ///  * Invitation does not exist
```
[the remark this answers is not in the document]
jeremymv2: I have refactored the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+
+        Ok(resp.json()?)
+    }
+
+    /// Retrieves pending origin member invitations
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin does not exist
```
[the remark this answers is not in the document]
jeremymv2: I have refactored the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+        Ok(resp.json()?)
+    }
+
+    /// Rescind an origin member invitation
+    ///
+    ///  # Failures
+    ///
+    ///  * Remote builder is not available
+    ///  * Origin does not exist
+    ///  * Invitation does not exist
```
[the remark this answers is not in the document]
jeremymv2: I have refactored the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+               invitee_account, origin);
+
+        let path = format!("depot/origins/{}/users/{}/invitations",
+                           origin, invitee_account);
+
+        self.0
+            .post(&path)
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::NO_CONTENT, StatusCode::CREATED])
```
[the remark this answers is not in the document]
jeremymv2: I have refactored the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/hab/src/command/origin/invitations/list_pending_origin.rs
```diff
@@ -0,0 +1,49 @@
...
+            for invitation in resp.invitations.iter() {
+                writeln!(&mut tw,
+                         "{}\t{}\t{}\t{}",
+                         invitation.id,
+                         invitation.created_at,
+                         invitation.account_name,
+                         invitation.ignored).unwrap();
+            }
+            tw.flush().unwrap();
+            let written = String::from_utf8(tw.into_inner().unwrap()).unwrap();
```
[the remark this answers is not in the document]
jeremymv2: I've refactored and removed the tabwrite code completely in https://github.com/habitat-sh/habitat/pull/7349/commits/19b27a45acc529f5a8481b9f6f62500b6dd9f75f

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,136 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+        debug!("Rescinding invitation {} in origin {}",
+               invitation_id, origin);
+
+        let path = format!("depot/origins/{}/invitations/{}", origin, invitation_id);
+
+        self.0
+            .put(&path)
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::NO_CONTENT, StatusCode::CREATED])
```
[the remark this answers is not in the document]
jeremymv2: I've improved the doc comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/hab/src/cli.rs
```diff
@@ -311,6 +311,66 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                 (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
                 (@arg NEW_OWNER_ACCOUNT: +required +takes_value {non_empty} "The account name of the new origin owner")
             )
+            (@subcommand invitations =>
+                (about: "Manage origin member invitations")
+                (@subcommand accept =>
+                     (about: "Accept an origin member invitation")
+                     (@arg ORIGIN: +required {valid_origin} "The origin name")
+                     (@arg INVITATION_ID: +required +takes_value {valid_numeric::<i64>} "The id of the invitation to accept")
```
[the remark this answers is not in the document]
jeremymv2: I've improved the cli comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/hab/src/cli.rs
```diff
@@ -311,6 +311,66 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
                 (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
                 (@arg NEW_OWNER_ACCOUNT: +required +takes_value {non_empty} "The account name of the new origin owner")
             )
+            (@subcommand invitations =>
+                (about: "Manage origin member invitations")
+                (@subcommand accept =>
+                     (about: "Accept an origin member invitation")
+                     (@arg ORIGIN: +required {valid_origin} "The origin name")
```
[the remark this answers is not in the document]
jeremymv2: I've improved the cli menu comments in https://github.com/habitat-sh/habitat/pull/7349/commits/473183cd1251e9ef96b11d8c1a6cccfdac6e39d0

components/builder-api-client/src/lib.rs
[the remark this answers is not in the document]
jeremymv2: I've significantly refactored in https://github.com/habitat-sh/habitat/pull/7349/commits/19b27a45acc529f5a8481b9f6f62500b6dd9f75f I'm going to mark this resolved as the code now merits a fresh look.

components/builder-api-client/src/lib.rs
```diff
@@ -197,6 +197,24 @@ pub struct OriginChannelIdent {
...
+pub struct OriginInvitation {
+    pub id:           String,
+    pub account_id:   String,
+    pub account_name: String,
+    pub created_at:   String,
+    pub ignored:      bool,
+    pub origin:       String,
+    pub owner_id:     String,
+    pub updated_at:   String,
+}
```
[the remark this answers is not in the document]
jeremymv2: Agreed. I've added your suggestions in https://github.com/habitat-sh/habitat/pull/7349/commits/19b27a45acc529f5a8481b9f6f62500b6dd9f75f

components/builder-api-client/src/builder.rs
```diff
@@ -531,6 +533,198 @@ impl BuilderAPIProvider for BuilderAPIClient {
...
+            .bearer_auth(token)
+            .send()?
+            .ok_if(&[StatusCode::NO_CONTENT])
+    }
+
+    /// Marks an origin member invitation ignored
+    ///
+    ///  After ignoring, the user will no longer see the invitation
+    ///
+    ///  # Builder API endpiont (api.raml permalink)
```
eeyun: You're a wonderful human being. Thanks for putting this in here.

components/hab/src/command/origin/invitations/list_pending_origin.rs
```diff
@@ -0,0 +1,49 @@
...
+            println!("\n{}", written);
+            Ok(())
+        }
+        Err(err @ api_client::Error::APIError(StatusCode::UNAUTHORIZED, _)) => {
+            ui.fatal("Failed to list pending invitations!")?;
+            ui.fatal("This situation could arise, if for example, you are not a member of the \
+                      origin.")?;
```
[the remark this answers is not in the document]
jeremymv2: I believe I handled all the situations where FORBIDDEN is returned when a member is not in the origin in this commit https://github.com/habitat-sh/habitat/pull/7349/commits/11e5c2b79ed4a796109e5a6601d74ad7b227a0af

components/hab/src/command/origin/invitations/send.rs
```diff
@@ -38,7 +38,7 @@ pub fn start(ui: &mut UI,
         }
         Err(err @ api_client::Error::APIError(StatusCode::NOT_FOUND, _)) => {
             ui.fatal("Failed to send the invitation!")?;
-            ui.fatal("The origin or the member being invitationd (or both) does not exist.")?;
+            ui.fatal("The origin or the member being invitated does not exist.")?;
```
christophermaier: "invitated" -> "invited"

components/builder-api-client/src/lib.rs
```diff
@@ -197,6 +199,86 @@ pub struct OriginChannelIdent {
...
+#[derive(Clone, Deserialize)]
+pub struct UserOriginInvitationsResponse(pub Vec<OriginInvitation>);
+
+#[derive(Clone, Deserialize)]
+pub struct PendingOriginInvitationsResponse {
+    pub origin:      String,
+    pub invitations: Vec<OriginInvitation>,
+}
+
+macro_rules! invitation_format {
```
eeyun: This is pretty clever. I had to sit here for a moment and think about whether or not i liked the decision to use a macro here but ultimately its pretty clean and I'm fairly ambivalent on whether its the right or wrong thing. I'm curious how you might solve the problem without a macro but no obligation to do so from me.

components/builder-api-client/src/lib.rs
```diff
@@ -199,22 +201,84 @@ pub struct OriginChannelIdent {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut output = Vec::new();
+        let created_at =
+            match NaiveDateTime::parse_from_str(&self.created_at, "%Y-%m-%dT%H:%M:%S%.f") {
+                Ok(date) => format!("{}", date.format("%Y-%m-%d %H:%M:%S")),
+                Err(_) => format!("{}", Utc::now()),
+            };
+        output.push(format!(invite_format!(),
+                            self.id, self.origin, self.account_name, created_at, self.ignored));
+        write!(f, "{}", output.join("\n"))
```
christophermaier: I have some misgivings about removing `TabWriter` completely in favor of custom `Display` implementations on these types.

From my perspective, this is taking a very specific output need (i.e., creating a table) and elevating it to the default output rendering. While you _might_ be able to make an argument for that from the point of view of the various response structs below, I think it's more difficult to do that here. For instance, the printing of an `OriginInvitation` is now implicitly assumed to be taking place _inside a table_, to the point that you're also embedding a newline at the end of formatted output (the embedded newline by itself is a bit of a red flag to me, actually).

While this is indeed the only way that we're outputting an `OriginInvitation` now, there's nothing to say that's the way it will be forever. As soon as another output need is uncovered, it will become rather awkward to work with this.

The previous way of doing this with TabWriter was OK, in that it was basically encoding a new, situation-specific rendering behavior for a group of items; I think it was just repeated more than necessary.

If you wanted to go all the way down the rabbit hole, you could envision some kind of `Tabular` trait that you could implement that would allow you to define a `header()` and `as_table_row(&self)` function for a given type. Then when you're dealing with collections of a `Tabular` implementation, you could generically call the `header()` function first, then iterate through the collection and call `as_table_row` for each item. The nice thing about this is that you can encapsulate all that behavior in a single place, and then easily get tabular output for anything by implementing two simple functions. The generic implementation could remain free to use `TabWriter` as an implementation detail.

That makes it very straightforward to see what kind of behavior is being used in each situation, and doesn't assume too much in the general case.

(You don't necessarily need to do all that here, but it could be used as a foundation on which to squeeze more standardization out of our CLI in the future.)

components/builder-api-client/src/lib.rs
```diff
@@ -199,22 +201,84 @@ pub struct OriginChannelIdent {
...
+#[derive(Clone, Deserialize)]
+pub struct UserOriginInvitationsResponse(pub Vec<OriginInvitation>);
+
 #[derive(Clone, Deserialize)]
 pub struct PendingOriginInvitationsResponse {
     pub origin:      String,
     pub invitations: Vec<OriginInvitation>,
 }
 
+macro_rules! invite_format {
```
christophermaier: Why is this a macro? Seems like it could just be a straightforward constant.

components/builder-api-client/src/lib.rs
christophermaier: As an aside, the use of a `Vec` in this implementation is unnecessary, since you're only ever creating a single string. You could just as easily add the newline into the `write!` formatting string and get the same outcome (or, probably, just use `writeln!`).

(However, I think embedding the newline in a `Display` implementation is a bit suspect to begin with!)

components/hab/src/command/origin/invitations/send.rs
jeremymv2: Fixed in a previous commit.

components/builder-api-client/src/lib.rs
jeremymv2: I can re-visit that strategy. When I first tried with a constant I encountered some issues.

components/builder-api-client/src/lib.rs
jeremymv2: I'll follow up on your suggestions.

components/builder-api-client/src/lib.rs
jeremymv2: This is no longer relevant since I've removed the macro and introduced a Tabular trait in https://github.com/habitat-sh/habitat/pull/7349/commits/a65aebfafec7355e188e527768485ba9957fa404

components/builder-api-client/src/lib.rs
jeremymv2: Yes, the macro might have been clever but, alas, merited a more robust implementation. I've removed the macro and introduced a Tabular trait in https://github.com/habitat-sh/habitat/pull/7349/commits/a65aebfafec7355e188e527768485ba9957fa404

components/builder-api-client/src/lib.rs
jeremymv2: @christophermaier I've introduced a Tabular trait in https://github.com/habitat-sh/habitat/pull/7349/commits/a65aebfafec7355e188e527768485ba9957fa404

Since this thread references `impl fmt::Display for OriginInvitation` which is no longer in the code base (the implementation has been moved to a trait), I'm going to mark this resolved and ask for a fresh review 😄

components/builder-api-client/src/lib.rs
```diff
@@ -197,6 +200,115 @@ pub struct OriginChannelIdent {
...
+                                  invitation.id,
+                                  invitation.origin,
+                                  invitation.account_name,
+                                  invitation.created_at,
+                                  invitation.ignored));
+            }
+        };
+        match tabify(tw, &body.join("\n")) {
+            Ok(s) => s,
+            Err(e) => e.to_string(),
```
christophermaier: It's a bit surprising to see this function return a `String`, rather than a `Result<String>`, given that this can return an error. What's the rationale for that?

jeremymv2: I concur that changing the function return type to `Result<String>` adheres to the principle of bubbling the errors all the way up the stack to be handled in `main.rs` and it therefore is a better design. I've made that change in https://github.com/habitat-sh/habitat/pull/7349/commits/70530a5e37a1000b31604847d7a03bd800b71a9a

jeremymv2:
@christophermaier, I am curious if my changes have addressed your original concerns and if there is anything outstanding?

christophermaier:
@jeremymv2 Yeah, your changes are fine... I still think there's a useful abstraction lurking in there, but I'm having a bit of trouble formulating it. We can always dig that out later, though.

Issue #7353, opened 2020-01-21
Upgrade to time 0.2


davidMcneil:
The `time` crate has substantially changed between 0.1 and 0.2. This [issue](https://github.com/time-rs/time/issues/197) describes some of the changes. The change that most impacts us is the removal of [`TimeSpec`](https://docs.rs/time/0.1.42/time/struct.Timespec.html). This is a fairly simple struct so we should be able to replicate its functionality. 

See this [issue](https://github.com/time-rs/time/issues/201) for help in converting format strings.

christophermaier:
![](https://media.giphy.com/media/3iUMl1Fh6HRew/giphy.gif)

Issue #7352, opened 2020-01-21
Simplify the actix-web server startup


davidMcneil:
The way we start actix-web is overly complicated and not as efficient as it could be. The way it works today we have to start the http server in its own thread and play some tricks with a condvar. With the migration to the new futures ecosystem, this is no longer necessary. However, integrating the actix runtime with an existing tokio runtime is not as straightforward as one would hope. There is the [`run_in_tokio`](https://docs.rs/actix/0.9.0/actix/prelude/struct.System.html#method.run_in_tokio) function (and this [issue](https://github.com/actix/actix-web/issues/1283) describing how to use it), but I could not find a way to use it without blocking the current thread. I am hopeful as the futures ecosystem stabilizes an established pattern will emerge.

davidMcneil:
This is still rather difficult given the current `actix-web` api. I am moving it to backlog until the api improves.

Issue #7271, opened 2019-12-09
Remove futures 0.1 when actix-web upgrades


davidMcneil:
`actix-web` requires `futures` 0.1 when `actix-web` 2.0 comes out we can remove the `futures` 0.1 dependency.

chefsalim:
Note - it will be good to make the same upgrades across both hab and Builder when the time comes

PR #7354, opened 2020-01-21
Update dependencies

### LINKED ISSUE #7270: Upgrade prost off of its fork
See issue [here](https://github.com/danburkert/prost/issues/242).

davidMcneil:
Resolves #7270 #7271

The update process spawned two new issues #7352 and #7353. I felt these issues were outside the scope of a dependency update.

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -43,7 +43,7 @@ const X_FILENAME: &str = "x-filename";
 const DEFAULT_API_PATH: &str = "/v1";
 
 // The characters in this set are copied from
-// https://docs.rs/percent-encoding/1.0.1/percent_encoding/struct.SIMPLE_ENCODE_SET.html
+// https://docs.rs/percent-encoding/1.0.1/percent_encoding/struct.PATH_SEGMENT_ENCODE_SET.html
 const PATH_SEGMENT_ENCODE_SET: &AsciiSet = &CONTROLS.add(b' ')
                                                     .add(b'"')
                                                     .add(b'#')
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -20,10 +20,13 @@ use crate::{error::{Error,
             ReverseDependencies,
             SchedulerResponse};
 use broadcast::BroadcastWriter;
-use reqwest::{header::CONTENT_LENGTH,
-              Body,
+use percent_encoding::{percent_encode,
+                       AsciiSet,
+                       CONTROLS};
+use reqwest::{blocking::{Body,
+                         RequestBuilder},
+              header::CONTENT_LENGTH,
               IntoUrl,
-              RequestBuilder,
               StatusCode};
 use std::{fs::{self,
                File},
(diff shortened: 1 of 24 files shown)


components/builder-api-client/src/builder.rs
```diff
@@ -33,14 +36,26 @@ use std::{fs::{self,
...
-          Url};
+use url::Url;
 
 const X_FILENAME: &str = "x-filename";
 
 const DEFAULT_API_PATH: &str = "/v1";
 
+// The characters in this set are copied from
+// https://docs.rs/percent-encoding/1.0.1/percent_encoding/struct.SIMPLE_ENCODE_SET.html
+const PATH_SEGMENT_ENCODE_SET: &AsciiSet = &CONTROLS.add(b' ')
```
eeyun: Forgive my lack of knowledge on this - Why are we defining this ourselves - these all seem to already exist in `percent_encoding::CONTROLS`?

davidMcneil: No worries. I am lacking knowledge here as well and was surprised we had to manually define the encode set so if there is a better way to do this it would be great.

We were very behind in versions. So far in fact that it has now become two distinct crates `url` and `percent-encoding`.

To me it does not look like the `CONTROLS` variable has those characters. The docs point to this [description](https://url.spec.whatwg.org/#c0-control-percent-encode-set). I believe what we want is described just below that and is the [path percent encode set](https://url.spec.whatwg.org/#path-percent-encode-set).

What version has the `PATH_SEGMENT` variable? That sounds like what I want but I could not find it in the most recent version.

eeyun: Well real talk I could be sending you on a wild goose chase so don't spend too much time on it :joy:. I'm not certain which version, I just found a this change in percent-encoding that looks like it might be the source of `PATH_SEGMENT_ENCODE_SET` being removed https://github.com/servo/rust-url/pull/519/files

christophermaier: It looks like what you have is correct, based on my reading of the links you posted, but we might not need '%' and '/'... seems like those are in another set.

That being said, this seems like the kind of very low-level thing that you'd expect a library to handle for you, so it feels like there's something I'm still missing about why we would need this.

:thinking:

davidMcneil: Whoops, I have the wrong lik in the code comment. It was supposed to be pointing [here](https://docs.rs/percent-encoding/1.0.1/percent_encoding/struct.PATH_SEGMENT_ENCODE_SET.html). That is the documentation for the old constant which includes '%' and '/' and I was trying to preserve the old behavior. Happy to remove them if we decide they are not needed.

I agree it is very odd we have to do this by hand. Especially considering old versions of the library defined the necessary encoding set. However, the [last paragraph](https://docs.rs/percent-encoding/2.1.0/percent_encoding/?search=) of the crates documentation states, "This crate deliberately does not provide many different sets".

christophermaier: > However, the [last paragraph](https://docs.rs/percent-encoding/2.1.0/percent_encoding/) of the crates documentation states, "This crate deliberately does not provide many different sets".

Well, that seems pretty straightforward. Guess we get to define our own set. :man_shrugging: 

And it looks like this _is_ replicating the old behavior, so that all looks good :+1:

PR #7365, opened 2020-01-23
Update manual validation docs

christophermaier:
Also tweak the Kernel2 VM `Vagrantfile` to make it easier to evaluate
candidates from the `staging` channel.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -18,6 +18,8 @@ new packages coming in and invalidating your efforts.
 
 ## Validate the Release
 
+
+### Installation / Update
 For each platform
 ([darwin](https://packages.chef.io/files/staging/habitat/latest/hab-x86_64-darwin.zip),
 [linux](https://packages.chef.io/files/staging/habitat/latest/hab-x86_64-linux.tar.gz),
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -27,41 +29,158 @@ download the latest release candidate CLI from `packages.chef.io`. You
...
+#### Linux, Kernel 2
+
+Run either of the following commands:
+
+``` sh
+curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh \
+    | sudo bash -s -- -c staging -t x86_64-linux-kernel2
+```
+(Note the addition of the target option on the "curlbash" command! Without this you will end up with
+the Kernel 3 version, and things won't work properly!)
```
smacfarlane: Nit: There is no Kernel 3 for hab :smile:    `x86_64-linux`, `modern linux cli`, `default linux cli` would be more accurate but are a bit of a mouthful.

christophermaier: Fixed in https://github.com/habitat-sh/habitat/pull/7365/commits/08e67508dea3202bda9db9a8d324ebc89f728559

PR #7368, opened 2020-01-24
Add token and warning to rdeps usage

chefsalim:
Related to https://github.com/habitat-sh/builder/pull/1278/files

This change adds the auth token to the rdeps call made during `hab bldr job start`

### DIFF
--- a/components/hab/src/command/bldr/job/start.rs
+++ b/components/hab/src/command/bldr/job/start.rs
@@ -20,7 +20,7 @@ pub fn start(ui: &mut UI,
     let api_client = Client::new(bldr_url, PRODUCT, VERSION, None).map_err(Error::APIClient)?;
 
     if group {
-        let rdeps = api_client.fetch_rdeps((ident, target))
+        let rdeps = api_client.fetch_rdeps((ident, target), token)
                               .map_err(Error::APIClient)?;
         if !rdeps.is_empty() {
             ui.warn("Found the following reverse dependencies:")?;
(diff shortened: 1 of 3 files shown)


components/hab/src/command/bldr/job/start.rs
```diff
@@ -29,6 +29,8 @@ pub fn start(ui: &mut UI,
                 ui.warn(rdep.to_string())?;
             }
 
+            ui.warn("(Note: dependencies from private origins may not be shown)")
```
jeremymv2: I believe that the phrase "may not be shown" is going to be too vague here and leave the user wondering why they wouldn't have been shown and in what circumstance(s).

Perhaps something like the following expresses the idea more explicitly:
`Note: dependencies from private origins are omitted for non-members.`

jeremymv2: Small but important: the `ui.warn` line is missing a semi-colon at the end 😄

chefsalim: Sounds reasonable, thanks!

components/hab/src/command/bldr/job/start.rs
```diff
@@ -29,6 +29,8 @@ pub fn start(ui: &mut UI,
                 ui.warn(rdep.to_string())?;
             }
 
+            ui.warn("Note: dependencies from private origins are omitted for non-members.");
+
             let question = "If you choose to start a group build for this package, all of the \
```
kagarmoe: Do you mean:
```suggestion
            let question = "Starting a group build for this package will \ also build all of the reverse dependencies listed above. Is this what you want?";
```
You could also add the reason they might not want to do it, which I assume is because it is time-consuming. Does it have any other impact?

chefsalim: Right, builds can take time and tie up resources that a user might not want. So having a basic confirm message seems appropriate.

PR #7387, opened 2020-01-27
Use packages.chef.io for installation of all versions

### LINKED ISSUE #7312: Migrate old releases out of bintray
In order to provide a consistent download location and simplify our install.sh/ps1 scripts, we should migrate our existing bintray releases to packages.chef.io

- [x] Download all existing releases from the stable bintray channel
- [x] Validate checksums and signatures and note any abnormalities
- [x] Migrate package naming and directory layout to format expected by packages.chef.io
- [x] Regenerate checksums for new filenames
- [ ] Re-sign archives with Chef gpg key.
- [ ] Decide how to handle duplicate and -dev versions that made it to the stable channel
- [x] Decide how to handle contents of and generate manifest for each release
- [x] Determine strategy for legacy studio container images. (See #7384)

smacfarlane:
Closes #7312 

All releases of legacy versions of Habitat have been migrated to packages.chef.io. This PR drops support for installation of those packages from their previous host. 

### DIFF
--- a/components/hab/install.sh
+++ b/components/hab/install.sh
@@ -42,7 +42,7 @@ main() {
   create_workdir
   get_platform
   validate_target
-  download_packages_chef_io_archive "$version" "$channel" "$target"
+  download_archive "$version" "$channel" "$target"
   verify_archive
   extract_archive
   install_hab
@@ -163,7 +163,7 @@ validate_target() {
   fi
 }
 
-download_packages_chef_io_archive() {
+download_archive() {
   need_cmd mv
   
   local _version="${1:-latest}"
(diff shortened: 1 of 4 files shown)


components/hab/install.sh
```diff
@@ -44,12 +42,7 @@ main() {
   create_workdir
   get_platform
   validate_target
-  if use_packages_chef_io "$version"; then
-    download_packages_chef_io_archive "$version" "$channel" "$target"
-  else
-    get_bintray_version
-    download_bintray_archive
-  fi
+  download_packages_chef_io_archive "$version" "$channel" "$target"
```
christophermaier: Since there's no reason to make a distinction anymore, you could just call this function `download_packages` now.

PR #7391, opened 2020-01-29
[CI] Update Github Release automation

christophermaier:
Previously, we were using `hub` installed as a Ruby gem. However,
modern `hub` is written in Go, and is available through standard
package managers.

Additionally, we no longer need a `--commitish` argument, since we
know the pre-existing tag we want to use for the release.

### DIFF
--- a/.expeditor/scripts/finish_release/create_github_release.sh
+++ b/.expeditor/scripts/finish_release/create_github_release.sh
@@ -7,12 +7,15 @@ set -euo pipefail
 echo "--- :lock: Importing GPG keys"
 import_gpg_keys
 
-echo "--- :ruby: Installing 'hub' gem"
-gem install hub
+echo "--- Installing 'hub' CLI tool"
+sudo apt-get install hub
 
-echo "--- :thinking_face: Determining sha and version to release"
+echo "--- :thinking_face: Determining the version to release"
 get_manifest_for_environment "stable"
-read -r version gitsha <<< "$(jq -r '.version + " " + .sha' manifest.json)"
+version="$(jq -r '.version' manifest.json)"
+
+# It is our convention to have a tag for every version.
+tag="${version}"
 
 echo "--- :github: Creating release"
-maybe_run hub release create --message "$version" --commitish "$gitsha" "$version"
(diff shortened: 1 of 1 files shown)


.expeditor/scripts/finish_release/create_github_release.sh
```diff
@@ -7,12 +7,15 @@ set -euo pipefail
 echo "--- :lock: Importing GPG keys"
 import_gpg_keys
 
-echo "--- :ruby: Installing 'hub' gem"
-gem install hub
+echo "--- Installing 'hub' CLI tool"
+sudo apt-get install hub
```
jeremymv2: Unless I'm mistaken, this isn't going to work since our `chefes/buildkite` docker image is Ubuntu 18.04 based and that would require utilizing [Snap](https://github.com/github/hub#installation)

That being the case, why not just install with our shared bash function [install_hub](https://github.com/habitat-sh/habitat/blob/1f454ed987f6be875fdbaf4b5ae048ae73af1155/.expeditor/scripts/shared.sh#L313) 😄

jeremymv2:
```
10:06 $ docker run -it --rm --init --privileged --volume ${PWD}:${PWD} --workdir ${PWD} chefes/buildkite bash -c "source .expeditor/scripts/shared.sh && install_hub && hub --version"
--- :github: Installing hub version v2.14.1 to /bin/hub from https://github.com/github/hub/releases/download/v2.14.1/hub-linux-amd64-2.14.1.tgz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   614    0   614    0     0   2379      0 --:--:-- --:--:-- --:--:--  2389
100 4994k  100 4994k    0     0  6002k      0 --:--:-- --:--:-- --:--:-- 16.7M
git version 2.17.1
hub version 2.14.1
✔ ~/Projects/habitat [master L|…2⚑ 12]
10:06 $
```

christophermaier: Good call... I missed that earlier.

PR #7392, opened 2020-01-29
[CI] Fix nightly bump automation

christophermaier:


.expeditor/scripts/finish_release/bump_rustfmt.sh
```diff
@@ -73,16 +73,26 @@ run_fmt_open_pr() {
...
+      --message "\"Bump nightly toolchain to ${FOUND_VERSION}\""
 
   # This script runs from a pipeline step, thus we do not
   # have access to expeditor's open_pull_request bash
   # action helper function.
   install_hub
 
   echo "--- :github: Creating PR"
-  maybe_run hub pull-request --push --no-edit
+  hub pull-request --push --no-edit
```
smacfarlane: This won't work* until the github token provided to us by expeditor can push to non-chef organizations.  For now, perhaps we should just skip this step and revisit in the future?

*(May have changed since I last spoke to @tduffield in December)

christophermaier:
There is still some releng-related stuff that needs to happen before this will work. Closing for now.

PR #7408, opened 2020-02-01
Upgrades to Foundation 6.5.3

kagarmoe:

Trying to upgrade some of our tools. My local build looks OK, but from past attempts, we should see if it works in acceptance.


kagarmoe:
I've rethought my approach. Will address in an issue for discussion.

PR #7416, opened 2020-02-04
Adding a Learn Chef Rally link to habitat.sh/learn

shaunyap:
I'm not super familiar with Middleman but did what I could.


mwrock:
@kagarmoe might want to have a look at this. I fired this up locally and the link doesn't look quite right. It's almost hidden due to the color and there seems to be too little space surrounding it. I'm sure it is a CSS thing.

PR #7431, opened 2020-02-08
close atomic writer temp file before rename

mwrock:
I noticed that the supervisor was not running in a Linux caontainer studio on a windows host which led me to find that the rename in `hab pkg install` was failing. The failuew was `Resource busy (os error 16)`. I noticed that `mv` worked just fine on the same file as well as a very simple rust binary that merely called `fs::rename`.

Changing the AtomicWriter to call `into_temp_path`, which closes the file without deleting it, before the call to `rename` appears to fix this error.

On the one hand this makes sense because the rename was in a mounted directory in the windows filesystem which does not allow a rename of an open file.

What I am baffled by is how this EVER worked. I went back as far as 0.71 and observed this same behavior. Also why does this not fail on pure windows? Might need some more investigation and validation.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -723,10 +723,9 @@ impl AtomicWriter {
         debug!("Renaming {} to {}",
                self.tempfile.path().to_string_lossy(),
                &self.dest.to_string_lossy());
-        fs::rename(self.tempfile.path(), &self.dest)?;
-
+        fs::rename(self.tempfile.into_temp_path(), &self.dest)?;
         #[cfg(unix)]
-        self.sync_parent()?;
+        AtomicWriter::sync_parent(&self.dest)?;
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


mwrock:
Its possible this new buggy behavior is due to a Docker for Windows Desktop upgrade that appears to [change the way shared folders work](https://www.docker.com/blog/new-filesharing-implementation-in-docker-desktop-windows/). Will experiment with downgrading.

stevendanna:
>Its possible this new buggy behavior is due to a Docker for Windows Desktop upgrade that appears to change the way shared folders work. Will experiment with downgrading.

For clarity though, do we still think this change makes sense regardless of the docker changes?  It seems to make sense to me.  

The temporary file is open with `FILE_SHARE_DELETE` which I thought should be enough to allow the rename even when the file is open. But given the current behavior, clearly there is more to the story.

christophermaier:
:thinking: I find myself in @stevendanna's shoes... it seems logical as is, but I'm not quite sure why it would be needed, given the file was open as `FILE_SHARE_DELETE` :thinking:

mwrock:
While the file is onened with `FILE_SHARE_DELETE` in the linux container. Running a `procmon` trace on the windows side reveals that its `ShareMode` is `Read, Write`. Running the same trace running `hab pkg install` locally shows the temp file has a `ShareMode` of `Read, Write, Delete`.

I'm guessing this is a nuance of how the newer Docker for Windows is dealing with mounting.

I think we want to move forward with this PR.

PR #7433, opened 2020-02-10
Use 1.0.8 of mac bootstrapper

christophermaier:
This should help get us past some macOS-specific build pipeline
issues.


christophermaier:
Merging, as this allows us to build on mac again.

PR #7442, opened 2020-02-11
Add gather-logs alias to align with other products

moutons:
Chef Infra Server and Automate use `gather-logs` when generating a support bundle, let's use this for Habitat as well please!

Signed-off-by: Shaun Mouton <smouton@chef.io>


christophermaier:
@moutons Unfortunately, `supportbundle` does not include logs right now, so I'm afraid this would cause more confusion :frowning_face:

christophermaier:
@moutons aaaaaand I see you've filed https://github.com/habitat-sh/habitat/issues/7444 already; nice.

Until _that_ is addressed, however, I think adding a `gather-logs` alias would just be confusing, so I'm going to close this for now. Happy to add it back once the extra functionality is added, though!

PR #7445, opened 2020-02-11
adds hab origin info subcommand

jeremymv2:
This provides `hab origin info <ORIGIN>` such that users can determine public origin metadata; for example, the origin's owner. Determining the owner is now possible due to https://github.com/habitat-sh/builder/pull/1290.

Who-Owns-It jpg

It provides a `--json` flag and underlying Trait implementation that can easily be applied to other API responses, where appropriate.

```
$ ./target/debug/hab origin info jeremymv2
☁ Discovering origin metadata
Origin [jeremymv2]:
Owner Id            Owner Account  Private Key               Package Visibility
835646971166441582  jeremymv2      jeremymv2-20171024221121  public

$ ./target/debug/hab origin info jeremymv2 --json
{"default_package_visibility":"public","name":"jeremymv2","owner_account":"jeremymv2","owner_id":"835646971166441582","private_key_name":"jeremymv2-20171024221121"}
$
```

### DIFF
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -254,7 +254,7 @@ mod json_date_format {
     }
 }
 
-pub fn convert_to_json<T>(src: &T) -> Result<Json>
+fn convert_to_json<T>(src: &T) -> Result<Json>
     where T: Serialize
 {
     serde_json::to_value(src).map_err(|e| habitat_core::Error::RenderContextSerialization(e).into())
--- a/components/builder-api-client/src/lib.rs
+++ b/components/builder-api-client/src/lib.rs
@@ -27,6 +27,8 @@ use std::{fmt,
 use chrono::{DateTime,
              Utc};
 use reqwest::IntoUrl;
+use serde::Serialize;
+use serde_json::Value as Json;
 use tabwriter::TabWriter;
 
 use crate::hab_core::package::PackageIdent;
@@ -194,6 +196,16 @@ pub struct OriginChannelIdent {
     pub name: String,
 }
 
+#[derive(Clone, Serialize, Deserialize)]
+pub struct OriginInfoResponse {
+    pub default_package_visibility: String,
+    pub name: String,
+    #[serde(with = "json_u64")]
+    pub owner_id: u64,
+    pub owner_account: String,
+    pub private_key_name: String,
+}
+
 #[derive(Clone, Deserialize)]
 pub struct OriginInvitation {
     #[serde(with = "json_u64")]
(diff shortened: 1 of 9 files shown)


components/builder-api-client/src/lib.rs
```diff
@@ -242,6 +254,12 @@ mod json_date_format {
     }
 }
 
+pub fn convert_to_json<T>(src: &T) -> Result<Json>
```
davidMcneil: Remove `pub`?

components/builder-api-client/src/lib.rs
```diff
@@ -301,6 +323,24 @@ impl Tabular for PendingOriginInvitationsResponse {
...
+        body.push(format!("{}\t{}\t{}\t{}",
+                          self.owner_id,
+                          self.owner_account,
+                          self.private_key_name,
+                          self.default_package_visibility));
+        tabify(tw, &body.join("\n"))
+    }
+}
+
+impl PortableText for OriginInfoResponse {
```
davidMcneil: Having `json` output is awesome! Far outside the scope of this PR, and just thinking outloud. It would be awesome to have all CLI output be represented as structs that implements `Serialize`. We could then easily support all kinds of output. We could even implement (or maybe one exists) a `Tabular` format for `serde`.

components/builder-api-client/src/lib.rs
```diff
@@ -194,6 +196,16 @@ pub struct OriginChannelIdent {
     pub name: String,
 }
 
+#[derive(Clone, Serialize, Deserialize)]
+pub struct OriginInfoResponse {
+    pub default_package_visibility: String,
+    pub name: String,
+    #[serde(with = "json_u64")]
```
davidMcneil: This is probably outside the scope of this PR, but I think we can remove `json_u64` in favor of the more general [`serde_string`](https://github.com/habitat-sh/habitat/blob/41d4cb42f080866fff8b57c63c6d433f52daf667/components/core/src/util/serde_string.rs#L1).

components/builder-api-client/src/lib.rs
jeremymv2: Whoops, that was vestigial. I removed in https://github.com/habitat-sh/habitat/pull/7445/commits/ce4f105d78ba8b4ac483215297dde0617606fff2

components/builder-api-client/src/lib.rs
jeremymv2: Hmm, it probably might make sense to investigate changing it in **all** the structs in this file:

```
17:15 $ grep "\"json_u64" components/builder-api-client/src/lib.rs | grep -v ///
    #[serde(with = "json_u64")]
    #[serde(with = "json_u64")]
    #[serde(with = "json_u64")]
    #[serde(with = "json_u64")]
    #[serde(with = "json_u64")]
    #[serde(with = "json_u64")]
    #[serde(with = "json_u64")]
✔ ~/Projects/habitat [jeremymv2/origin_info_subcommand L|…4⚑ 13]
17:16 $
```

components/builder-api-client/src/lib.rs
christophermaier: That's a fantastic idea!

PR #7474, opened 2020-02-19
add NO_NAMED_PIPE_HEALTH_CHECK feature flag to suppress named pipe on windows

### LINKED ISSUE #7464: Introduce feature flag to toggle named-pipe based health checks on windows
This is related to research conducted in #7439. In an environment with many health checks (~15 services), having a `pwsh` process running for the lifetime of the service increases total memory overhead by 35MB per service which is prohibitive in a  large constrained environment.

For now we want such users to be able to run on the older model where a health check spawns a `pwsh` process only for the duration of the health check. While this has proven to result in CPU spikes, such spikes can be mitigated by providing different health check intervals per service.

mwrock:
resolves #7464

### DIFF
--- a/components/common/src/command/package/install.rs
+++ b/components/common/src/command/package/install.rs
@@ -65,7 +65,8 @@ use crate::{error::{Error,
                                  InstallHook},
                          package::Pkg},
             ui::{Status,
-                 UIWriter}};
+                 UIWriter},
+            FeatureFlag};
 
 pub const RETRIES: usize = 5;
 pub const RETRY_WAIT: Duration = Duration::from_millis(3000);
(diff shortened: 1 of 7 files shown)


mwrock:
Its the only hook that actually uses feature flags but the tricky parts was feeding the feature flags to the hook's `new` function. That function gets called from the trait function `load` which gets called from the `HookTable`'s `load`. I briefly debated having the `HealthCheck` hook implement its own `load` but that function does quite a lot and it seemed beneficial for all hooks to have access to the feature flags if we ever needed to implement feature specific functionality for another hook.

That said maybe I'm going about this in more of a harebrained manner than I need to and am totally open for alternatives!

Issue #7361, opened 2020-01-23
Default options for `hab setup` lead to multiple dead-ends for first-time users


themightychris:
A first-time user will not have on hand, and doesn't need for their first build, several tokens that `hab setup` will prompt them for if they accept default options:

- `Connect to an on-premises bldr instance? [**Yes**/no/quit]`
  - Explained well enough
  - Not needed to start though, and you can come back when you want to switch to one
  - Not practical that a user could set this up during `hab setup`
  - Should default to **no**
- `Set up a default Habitat personal access token? [**Yes**/no/quit]`
  - Explained well enough
  - Not needed to start though, and you can come back when you want to switch to one
  - Enough information is provided that a user could set up a token while at the prompt
  - Defaulting to **no** would provide the least first-run friction, but defaulting to **yes** might provide the least first-publish friction. Seems like a priority call for product leaders
- `Set up a default Habitat Supervisor CtlGateway secret? [Yes/no/quit]`
  - No explanation offered
  - Not needed to start though, and you can come back when you want to switch to one
  - Not practical that a user could set this up during `hab setup`
  - Should default to **no**
  - Should have a clear explanation

In each of these cases, accepting the default of **yes** leads to a prompt the user may not be able to answer. Forcing them to decide to brave terminating `hab setup` and starting over.

This provides a bad experience for first-time users, and is a regression from previous releases since the additions of the on-premises bldr and CtlGateway prompts. In my experience referring users to try Habitat, I've found them getting frequently hung up here and needing to come back for guidance on how to get through it (if not simply deferring the entire effort)

mwrock:
I have to agree. I find my self almost always specifying "no" to most of these when testing/demoing habitat.

eeyun:
Same agreement here. I'm sure it's useful to a subset of users but "default" behavior should always look like the common public configuration.

PR #7480, opened 2020-02-19
modification to hab setup defaults and language

jeremymv2:
Closes: https://github.com/habitat-sh/habitat/issues/7361

### DIFF
--- a/components/hab/src/command/cli/setup.rs
+++ b/components/hab/src/command/cli/setup.rs
@@ -57,7 +57,7 @@ pub fn start(ui: &mut UI, cache_path: &Path) -> Result<()> {
         ui.br()?;
         ui.para("Enter the url of your builder instance. The default is \
                  https://bldr.habitat.sh. The configured endpoint can be overridden any time \
-                 with a `HAB_BLDR_URL` envvar or a --url flag on the cli.")?;
+                 with a `HAB_BLDR_URL` environment variable or a --url flag on the cli.")?;
         let mut url = prompt_url(ui)?;
 
         while valid_url(&url).is_err() {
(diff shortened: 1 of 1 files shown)


components/hab/src/command/cli/setup.rs
```diff
@@ -148,15 +149,24 @@ pub fn start(ui: &mut UI, cache_path: &Path) -> Result<()> {
...
+        ui.para("Okay, maybe another time. You can also set a `HAB_AUTH_TOKEN` envvar or pass \
+                 the `--auth` flag to the cli!")?;
     }
+    ui.heading("Supervisor Control Gateway Secret")?;
+    ui.para("The Supervisor control gateway secret is used to authenticate hab client commands \
+             to a Supervisor. When a new Supervisor is created, a unique secret is generated \
+             for it automatically and stored locally in a file: `/hab/sup/default/CTL_SECRET`. \
+             When issuing commands to a local Supervisor, this file is read for the secret. \
+             Typically, a default Supervisor control gateway secret would only be needed if you \
+             with to send commands to a remote supervisor.")?;
```
christophermaier: The tricky bit here is that if you want to set a value here, it really needs to be the secret being used by another Supervisor that you want to communicate with remotely.

I wish we had a better way to manage multiple Supervisors here :frowning:

components/hab/src/command/cli/setup.rs
```diff
@@ -126,7 +126,8 @@ pub fn start(ui: &mut UI, cache_path: &Path) -> Result<()> {
             }
         }
     } else {
-        ui.para("Okay, maybe another time.")?;
+        ui.para("Okay, maybe another time. You can also set a `HAB_ORIGIN` envvar when using \
```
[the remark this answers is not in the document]
jeremymv2: I agree! I went ahead and changed all the other, pre-existing instances of this 😄

components/hab/src/command/cli/setup.rs
jeremymv2: I agree with on both counts. Adjusted.

PR #7487, opened 2020-02-19
add a remotesigned execution policy to pwsh spawns

### LINKED ISSUE #7421: hooks should run pwsh with RemoteSigned execution policy
In the event that an organization sets their machine execution policy to AllSigned, any unsigned script will fail to run. This will not affect the studio since it launches powershell with `-ExecutionPolicy Bypass`. However the supervisor hooks are not run with this (the health check pipe service may also need this).

mwrock:
resolves #7421 

This also consolidates the creation of `pwsh` args into one place.

### DIFF
--- a/components/core/src/util.rs
+++ b/components/core/src/util.rs
@@ -8,6 +8,24 @@ pub mod win_perm;
 
 use std::mem;
 
+/// returns the common arguments to pass to pwsh.exe when spawning a powershell instance.
+/// These arguments are optimized for a background powershell process running hooks.
+/// The NonInteractive flag specifies that the console is not intended to interact with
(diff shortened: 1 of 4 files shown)


components/core/src/util.rs
```diff
@@ -8,6 +8,15 @@ pub mod win_perm;
 
 use std::mem;
 
+#[cfg(windows)]
+pub fn pwsh_args(command: &str) -> Vec<&str> {
```
christophermaier: Nice consolidation... it would be nice to have some documentation comments about why this is important, though.

mwrock: good call. comment added!

PR #7489, opened 2020-02-20
sup: set PID file mode to 0644 

### LINKED ISSUE #7351: PID file permissions are wrong
I can't find any code related to set PID file permissions. So if service is running under the `hab` user I unable to read `PID` from `reconfigure` hook.

```
-rw-------  1 root root   5 Jan 21 11:30 PID
```

```
my-elasticsearch.default hook[reconfigure]:(HK): cat: can't open '/hab/svc/my-elasticsearch/PID': Permission denied
```

stevendanna:
Users currently expect to be able to read the PID file from their
runtime hooks. In 5734cce we started using atomic writer for the PID
file to avoid corruption, but this meant that the default mode for the
file became 0600.

Note: I've restricted this change to non-Windows systems. If anyone
knows of what our Window's users might expect, I'm happy to extend
this PR to include a change for Windows as well.

Fixes #7351

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -6,6 +6,14 @@ use crate::{env as henv,
             package::{Identifiable,
                       PackageIdent,
                       PackageInstall}};
+
+#[cfg(not(windows))]
+use crate::util::posix_perm::{self,
+                              set_permissions};
+#[cfg(windows)]
+use crate::util::win_perm::{self,
+                            set_permissions};
+
 use dirs;
 use std::{env,
           fs,
@@ -418,7 +426,7 @@ impl<'a> SvcDir<'a> {
         }
 
         Self::create_dir_all(&path)?;
-        self.set_permissions(&path)
+        self.set_permissions_and_ownership(&path)
     }
 
     fn create_dir_all<P: AsRef<Path>>(path: P) -> Result<()> {
@@ -433,9 +441,7 @@ impl<'a> SvcDir<'a> {
     }
 
     #[cfg(not(windows))]
-    fn set_permissions<T: AsRef<Path>>(&self, path: T) -> Result<()> {
-        use crate::util::posix_perm;
-
+    fn set_permissions_and_ownership<T: AsRef<Path>>(&self, path: T) -> Result<()> {
         if users::can_run_services_as_svc_user() {
             posix_perm::set_owner(path.as_ref(), &self.svc_user, &self.svc_group)?;
         }
@@ -443,9 +449,7 @@ impl<'a> SvcDir<'a> {
     }
 
     #[cfg(windows)]
-    fn set_permissions<T: AsRef<Path>>(&self, path: T) -> Result<()> {
-        use crate::util::win_perm;
-
+    fn set_permissions_and_ownership<T: AsRef<Path>>(&self, path: T) -> Result<()> {
         win_perm::harden_path(path.as_ref()).map_err(From::from)
     }
 }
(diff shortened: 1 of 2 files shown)


components/core/src/fs.rs
```diff
@@ -719,6 +749,11 @@ impl AtomicWriter {
     /// temporary file to ensure all data is flushed to disk and then
     /// renaming the file into place.
     fn finish(self) -> io::Result<()> {
+        if let Some(permission) = self.permission {
+            set_permissions(&self.tempfile.path(), permission).map_err(|e| {
+                io::Error::new(io::ErrorKind::Other, e.to_string())
+            })?;
```
stevendanna: I don't love this.  We made a decision when creating the API to return an `io::Result<()>` here to make it easier on callers.  I don't have enough Rust experience to know if there is a better way to achieve the same ease of use so I'm happy to take whatever alternative y'all have (or just leave this bit of grossness).

components/core/src/fs.rs
```diff
@@ -694,16 +709,34 @@ fn parent(p: &Path) -> io::Result<&Path> {
...
 /// Assumes that the parent directory of dest_path exists.
 pub struct AtomicWriter {
-    dest:     PathBuf,
-    tempfile: tempfile::NamedTempFile,
+    dest:       PathBuf,
+    tempfile:   tempfile::NamedTempFile,
+    #[cfg(not(windows))]
+    permission: Option<u32>,
+    #[cfg(windows)]
+    permission: Option<&[win_perm::PermissionEntry]>,
```
stevendanna: This doesn't actually work, it was just my first swing at what we might want to do, but for this to work I think we'll need some lifetime annotations.  If a reviewer things this is a useful feature for windows, I'll put some time into getting a cross-compilation setup going so I can develop the window's portions, otherwise, I'll make this PR posix-only for now since it is posix-only at the call-site.

components/core/src/fs.rs
davidMcneil: You could implement `From<Error>` for `io::Error`

```rust
impl From<Error> for io::Error {
    fn from(e: Error) -> Self {
        io::Error::new(io::ErrorKind::Other, e.to_string())
    }
}
```

and then you would just need 
```rust
if let Some(permission) = self.permission {
    set_permissions(&self.tempfile.path(), permission)?;
}
```

You could even match on the `Error` type and try and convert it to the corresponding `io::Error` (`Error::PermissionFailed` to `io::Error::PermissionDenied`). But that just leads to the question, why does `set_permission` not use `io::Result`?

With all that said, the "correct" answer probably involves refactoring outside the scope of this PR so I am totally fine leaving it as is.

components/core/src/fs.rs
stevendanna: I think a Vec<PermissionEntry> is probably the way to go here if we proceed with the windows stuff, thanks.

components/core/src/fs.rs
```diff
@@ -443,13 +447,24 @@ impl<'a> SvcDir<'a> {
...
+    posix_perm::set_permissions(path.as_ref(), mode)
+}
+
+#[cfg(windows)]
+pub fn set_permissions<T: AsRef<Path>>(path: T,
+                                       entries: &[win_perm::PermissionEntry])
+                                       -> Result<()> {
+    win_perm::set_permissions(path.as_ref(), entries)
+}
+
```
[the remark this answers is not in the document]
stevendanna: 👍 I think this was the result of some iterative development on my part, originally these functions were doing a bit more but eventually they reduced to this and I didn't think to remove them.

components/core/src/fs.rs
christophermaier: I think the "correct" answer lies in https://github.com/habitat-sh/habitat/issues/6557, but that is most assuredly outside the scope of this PR :smile: 

I'm OK with leaving this as-is for the time being.

mwrock:
I think providing no windows implementation in the supervisor here is actually the right thing to do. By doing nothing, windows will simply allow the file to inherit the perms of the parent folder. We don't "harden" the `svc` folders (only some of their child folders so any authenticated user should have read access to the PID file on windows.

We should add a comment to this effect.

stevendanna:
@mwrock Awesome, thanks!  I'll get this updated sometime this week.

stevendanna:
I left the windows implementation of with_permissions in place just for symmetry, even though we aren't gonna use it for the PID file for now.

PR #7491, opened 2020-02-20
[CI] Restrict certain release pipeline steps to "real" runs only

christophermaier:
Sometimes we need to run a pipeline manually, for testing purposes,
etc. In this case, the build creator will _not_ be "Chef Expeditor",
so we'll skip several steps in that case.

Read the comments for further details around all this.


christophermaier:
@smacfarlane That's a good call... I'll add that.

PR #7499, opened 2020-02-24
Remove time crate from almost everything

### LINKED ISSUE #7353: Upgrade to time 0.2

christophermaier:
Rather than updating everything to use the new 0.2 line of the `time` crate, it turns out that we can basically remove it from everything with a bit of refactoring. Juggling multiple time representations was a bit of a pain internally, and made things more confusing than necessary.

This PR performs the (mostly) mechanical translation from `time` to `std::time`, with interesting diversions pointed out along the way. A bit of internal refactoring and cleanup helped with this as well.

Fixes #7353

### DIFF
--- a/components/core/src/crypto/keys.rs
+++ b/components/core/src/crypto/keys.rs
@@ -10,6 +10,7 @@ use super::{PUBLIC_BOX_KEY_VERSION,
 use crate::error::{Error,
                    Result};
 use base64;
+use chrono::Utc;
 use regex::Regex;
 use std::{collections::HashSet,
           fmt,
@@ -22,7 +23,6 @@ use std::{collections::HashSet,
                  PathBuf},
           result,
           str::FromStr};
-use time;
 
 lazy_static::lazy_static! {
     static ref NAME_WITH_REV_RE: Regex = Regex::new(r"\A(?P<name>.+)-(?P<rev>\d{14})\z").unwrap();
@@ -310,7 +310,7 @@ fn mk_key_filename<P, S1, S2>(path: P, keyname: S1, suffix: S2) -> PathBuf
 /// `{year}{month}{day}{hour24}{minute}{second}`
 /// Timestamps are in UTC time.
 // TODO (CM): Return a String, not a Result
-fn mk_revision_string() -> Result<String> { Ok(time::OffsetDateTime::now().format("%Y%m%d%H%M%S")) }
+fn mk_revision_string() -> Result<String> { Ok(Utc::now().format("%Y%m%d%H%M%S").to_string()) }
 
 pub fn parse_name_with_rev<T>(name_with_rev: T) -> Result<(String, String)>
     where T: AsRef<str>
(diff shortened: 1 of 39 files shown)


davidMcneil:
It looks like there still might be dependencies on [`time 0.1.42`](https://github.com/habitat-sh/habitat/blob/09333bcf5361c217e33dcc62fbc5d954870b83e0/Cargo.lock#L1536). Did only the `launcher` update?

christophermaier:
@davidMcneil :thinking: It might be due to transitive dependencies on the `time` crate... I'll look into it.

components/core/src/crypto/keys.rs
```diff
@@ -309,15 +309,8 @@ fn mk_key_filename<P, S1, S2>(path: P, keyname: S1, suffix: S2) -> PathBuf
...
-    let now = time::now_utc();
-    // https://github.com/rust-lang-deprecated/time/blob/master/src/display.rs
-    // http://man7.org/linux/man-pages/man3/strftime.3.html
-    match now.strftime("%Y%m%d%H%M%S") {
-        Ok(result) => Ok(result.to_string()),
-        Err(_) => Err(Error::CryptoError("Can't parse system time".to_string())),
-    }
-}
+// TODO (CM): Return a String, not a Result
+fn mk_revision_string() -> Result<String> { Ok(time::OffsetDateTime::now().format("%Y%m%d%H%M%S")) }
```
davidMcneil: Would it be possible to use `chrono`'s [`format`](https://docs.rs/chrono/0.4.10/chrono/struct.DateTime.html#method.format) here and then we could be completely rid of `time`?

components/sup-protocol/protocols/types.proto
```diff
@@ -55,7 +55,7 @@ message PackageIdent {
 }
 
 message ProcessStatus {
-  optional int64 elapsed = 1;
+  optional uint64 elapsed = 1;
```
davidMcneil: Do we have to worry about breaking protocol compatability here?

components/core/src/crypto/keys.rs
christophermaier: :thinking: Hrmm, that might be worth looking into. `chrono` actually wraps `time`, but maybe we could standardize on `chrono` to help simplify things; we do use `chrono` elsewhere, after all.

components/sup-protocol/protocols/types.proto
christophermaier: I mentioned this in the commit message, but yeah, I think it should be fine:

https://developers.google.com/protocol-buffers/docs/proto#updating

components/core/src/crypto/keys.rs
christophermaier: Actually, yes. We use `chrono` in just a small handful of places in the rest of the code, but it's always around formatting dates. I think it makes a lot of sense to standardize on `chrono` for that kind of stuff (Or standardize on `time`, for that matter!)

christophermaier: OK, we're on chrono, now!

PR #7503, opened 2020-02-25
Better test for comparing structopt to clap_app

davidMcneil:
In doing some more testing of the config file, I discovered that simply comparing the help output between the `structopt` cli and `clap_app` cli was not sufficient to catch all the possible nuances of the arguments. 

It turns out, [`clap::Parser`](https://github.com/clap-rs/clap/blob/fa45af5caddf891b6e827a97b4969366c46dc76f/src/app/parser.rs#L52) has several `pub` fields. It is kinda hacky, but we can use these fields to check almost every field of the `clap::App`'s.

This PR uses this more rigorous testing method and fixes the differences it uncovered.


components/hab/src/cli.rs
```diff
@@ -1113,28 +1116,26 @@ fn sub_pkg_install(feature_flags: FeatureFlag) -> App<'static, 'static> {
         (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
         (@arg IGNORE_INSTALL_HOOK: --("ignore-install-hook") "Do not run any install hooks")
     );
-    if feature_flags.contains(FeatureFlag::OFFLINE_INSTALL) {
-        sub = sub.arg(Arg::with_name("OFFLINE").help("Install packages in offline mode")
+    sub = sub.arg(Arg::with_name("OFFLINE").help("Install packages in offline mode")
+                                               .hidden(!feature_flags.contains(FeatureFlag::OFFLINE_INSTALL))
```
davidMcneil: It is not possible to conditionally add an arg with `structopt` at runtime. The best we can do is conditionally hide the arg.

PR #7507, opened 2020-02-27
Make FullyQualifiedPackageIdent public and use it in Pkg

davidMcneil:
There is much more work to be done, but this is the start of using a `FullyQualifiedPackageIdent` type. This PR does two things. Cleanup and add to the existing `FullyQualifiedPackageIdent`, and use `FullyQualifiedPackageIdent` in `Pkg`.

I removed the `Cow` in `FullyQualifiedPackageIdent`. I do not think it was buying us much, but I will certainly add it back if that is what reviewers would prefer.

### DIFF
--- a/components/common/src/templating/package.rs
+++ b/components/common/src/templating/package.rs
@@ -118,8 +118,8 @@ impl Pkg {
                         path: package.installed_path.clone(),
                         origin: package.ident.origin.clone(),
                         name: package.ident.name.clone(),
-                        version: String::from(ident.guaranteed_version()),
-                        release: String::from(ident.guaranteed_release()),
+                        version: String::from(ident.version()),
+                        release: String::from(ident.release()),
                         shutdown_signal: package.shutdown_signal()?.unwrap_or_default(),
                         shutdown_timeout: package.shutdown_timeout()?.unwrap_or_default(),
                         ident };
--- a/components/common/src/templating/package.rs
+++ b/components/common/src/templating/package.rs
@@ -4,7 +4,8 @@ use crate::{error::{Error,
                     os::{process::{ShutdownSignal,
                                    ShutdownTimeout},
                          users},
-                    package::{PackageIdent,
+                    package::{FullyQualifiedPackageIdent,
+                              PackageIdent,
                               PackageInstall},
                     util::serde_string},
             util::path};
@@ -13,6 +14,7 @@ use serde::{ser::SerializeStruct,
             Serializer};
 use std::{collections::{BTreeMap,
                         HashMap},
+          convert::TryFrom,
           env,
           iter::FromIterator,
           ops::Deref,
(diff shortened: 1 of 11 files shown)


components/common/src/templating/package.rs
```diff
@@ -68,7 +70,7 @@ impl Env {
 #[derive(Clone, Debug, Deserialize)]
 pub struct Pkg {
     #[serde(with = "serde_string")]
-    pub ident:                   PackageIdent,
+    pub ident:                   FullyQualifiedPackageIdent,
```
christophermaier: This might be a change for a later PR, but it might be nice to remove the `oritgin`, `name`, `version`, and `release` fields from this struct and replace them with accessor functions that pull the desired part from the `ident` field.

components/core/src/package/ident.rs
```diff
@@ -304,12 +299,40 @@ impl<'a> From<&'a PackageIdent> for Cow<'a, PackageIdent> {
...
+    pub fn guaranteed_version(&self) -> &str {
+        self.version()
+            .unwrap_or_else(|| panic!("PackageIdent {} should be fully qualified", self.0))
+    }
+
+    pub fn guaranteed_release(&self) -> &str {
+        self.release()
+            .unwrap_or_else(|| panic!("PackageIdent {} should be fully qualified", self.0))
+    }
+}
```
christophermaier: This API seems a little "off" to me... I would expect that these would be guaranteed by virtue of being methods on a `FullyQualifiedPackageIdent`, so I would expect just `version()` and `release()`. Is there a reason that I'm missing that makes the `guaranteed_` prefix more useful?

I assume that the fact that we keep a `PackageIdent` inside the struct is to help inherit the various comparison trait implementations, but I can imagine a future where we just keep the individual components as fields within this struct, which would allow us to get rid of the `unwrap_or_else` here. I'm cool with keeping this for now, though, since this is transitional, and there's a lot more work to do.

markan: Possibly dumb related question: Are the guaranteed_*  methods needed because we implement Identifiable, which returns Option<string> for the release and version methods? If we implemented release() -> string that would have a different type sig, and I don't fully understand rust's type rules.

christophermaier: Strictly speaking, no, but I think issues arise from using `Identifiable#version` in this implementation.

(This could also be a suggestion we postpone until later, since there will be more work on this.)

components/common/src/templating/package.rs
davidMcneil: Completely agree, but I think I will hold off for now. The methods actually exist with the `Identifiable` trait. I think a future PR could do a lot of cleanup around this trait.

components/core/src/package/ident.rs
davidMcneil: > This API seems a little "off" to me

Agreed. Addressed in [ad924b](https://github.com/habitat-sh/habitat/pull/7507/commits/ad924bf0e2062c5f15a01877d3ff4de22abcf13e). This was me just trying to get something to compile and then forgetting to come back and clean it up. Thanks!

> I assume that the fact that we keep a PackageIdent inside the struct is to help inherit the various comparison trait implementations, but I can imagine a future where we just keep the individual components as fields within this struct,

Right... Agreed, that sounds like a much better future.

PR #7513, opened 2020-03-02
Add service `--update-condition` enabling service rollback

### LINKED ISSUE #7428: Implement service rollback
See #7378 for planning details

Depends on #7427

davidMcneil:
Resolves #7428

This PR introduces various service update conditions. The main purpose for adding these conditions is to allow service rollback. This PR adds a new `update-condition` cli option when starting the supervisor with a package and when loading a package. The `update-condition` help text is as follows:
```text
--update-condition <UPDATE_CONDITION>
        The condition dictating when this service should update 
        latest: Runs the latest package that can be found in the
                   configured channel and local packages.
        track-channel: Always run what is at the head of a given channel. 
                   This enables service rollback where demoting a package from
                   a channel will cause the package to rollback to an older version
                   of the package. A ramification of enabling this condition is packages
                   newer than the package at the head of the channel will be
                   automatically uninstalled during a service rollback.
[default:latest]  [possible values: latest, track-channel]
``` 
@apriofrost UI input on the `update-condition` would be great. I am curious what people think about the option describing what the supervisor is doing under the hood instead of calling out specific features such as "rollback".

There is a third `update-condition` that naturally presents itself when examining the other two. It would essentially be a `track-local` option which only looks at local packages and not a channel. This PR does not implement this option, but it is just something to consider.

There are parts of this PR that feel a bit hacky. I do not see a clear way to clean them up without further refactoring. I would be particularly interested in getting reviewers opinions on these parts.   


components/hab/src/command/pkg/uninstall.rs
```diff
@@ -33,21 +39,24 @@ use std::{fs,
...
-pub fn start(ui: &mut UI,
-             ident: &PackageIdent,
-             fs_root_path: &Path,
-             execution_strategy: ExecutionStrategy,
-             scope: Scope,
-             excludes: &[PackageIdent],
-             services: &[PackageIdent])
-             -> Result<()> {
+/// `even_if_running` is a flag indictating that the package should be uninstalled even if it is
+/// running.
```
davidMcneil: This option feels hacky. See here for the [explanantion](https://github.com/habitat-sh/habitat/blob/7240f505a9f5b31f2938ea193df92b507724c73f/components/sup/src/util/pkg.rs#L130) of why it is necessary.

components/sup/src/manager.rs
```diff
@@ -1399,11 +1406,12 @@ impl Manager {
 
     /// Create a future for stopping a Service removing it from the manager. The Service is assumed
     /// to have been removed from the internal list of active services already (see, e.g.,
-    /// take_services_with_updates and remove_service_from_state).
+    /// restart_services_rsw_mlr_rhw_msw and remove_service_from_state).
     /// # Locking for the returned Future (see locking.md)
     /// * `GatewayState::inner` (write)
     fn stop_service_future_gsw(&self,
                                mut service: Service,
+                               latest_desired_on_restart: Option<PackageIdent>,
```
davidMcneil: This feels a bit odd, but I did not see an obviously better alternative.

components/sup/src/manager/service_updater/package_update_worker.rs
```diff
@@ -103,8 +110,11 @@ impl PackageUpdateWorker {
         }
     }
 
-    /// Use the service spec's package ident to search for packages.
-    pub async fn update(&self) -> PackageIdent { self.update_to(self.ident.clone()).await }
+    /// Use the service spec's package ident to search for packages taking into account the update
+    /// condition.
+    pub async fn update(&self, condition: UpdateCondition) -> PackageIdent {
```
christophermaier: Should the `condition` be part of the update worker's internal state? It and the ident both come from the service spec, for instance. It's also part of the `RollingUpdateWorker`.

(Same goes for the `update_to` method.)

components/sup/src/manager.rs
```diff
@@ -1201,39 +1193,54 @@ impl Manager {
...
-        let mut idents_to_restart = Vec::new();
-        for (current_ident, service) in state_services.iter() {
-            if service.needs_restart {
-                idents_to_restart.push(current_ident.clone());
-            } else if let Some(new_ident) = service_updater.has_update(&service.service_group) {
-                outputln!("Updating from {} to {}", current_ident, new_ident);
+        let mut idents_to_restart_and_latest_desired_on_restart = Vec::new();
+        for (ident, service) in state_services.iter() {
+            if let Some(new_ident) = service_updater.has_update(&service.service_group) {
+                outputln!("Updating from {} to {}", ident, new_ident);
```
christophermaier: It would be useful to indicate here whether we are actually rolling back to an older package, since we can compare the identifiers. Something as simple as changing the wording to "Rolling back to X from Y" would be all that's needed.

components/sup/src/manager.rs
```diff
@@ -1421,6 +1429,24 @@ impl Manager {
...
+                        if let Err(e) = pkg::uninstall(&latest_ident).await {
+                            error!("Failed to uninstall '{}' during rollback. On restart, \
+                                    service will start with the wrong package. err: {}",
+                                   latest_ident, e);
+                        }
+                    } else {
+                        break;
+                    }
+                }
+            }
```
christophermaier: This feels like a chunk of logic that could be extracted into a standalone function, rather than being embedded within the "stop service" function.

I'm also wondering if this logic should ultimately be invoked from the service updater threads themselves. They are the ones that are responsible for a) figuring out what the next version of the service to run should be and b) getting that version on disk in the first place. If you squint your eyes, "getting that version on disk in the first place" could straightforwardly become "ensuring that version is the _latest_ on disk", given the appropriate update condition. I think that would also prevent having to make a bunch of changes outside of the updaters themselves.

components/sup/src/manager/service_updater/package_update_worker.rs
```diff
@@ -60,38 +61,44 @@ impl From<&Service> for PackageUpdateWorker {
...
-            // is what the old updater used, but its exact behavior should be tested/documented.
-            match util::pkg::install_no_ui(&self.builder_url, &install_source, &self.channel).await
-            {
+            let package_result = match condition {
+                UpdateCondition::Latest => {
+                    let install_source = ident.clone().into();
+                    util::pkg::install_no_ui(&self.builder_url, &install_source, &self.channel).await
+                }
+                UpdateCondition::TrackChannel => {
+                    util::pkg::install_channel_head(&self.builder_url, &ident, &self.channel).await
```
christophermaier: I think you might be able to dispense with this `install_channel_head` function by taking advantage of the `LocalPackageUsage::Ignore` enum member.

(It still may be a good idea to have a separate function that hides the usage of that, to keep the code clear and intention-revealing, but that may keep the amount of extra code written to a minimum.)

components/hab/src/cli/hab/svc.rs
```diff
@@ -41,6 +42,19 @@ pub enum Svc {
...
+        /// track-channel: Always run what is at the head of a given channel. This enables service
+        /// rollback where demoting a package from a channel will cause the package to rollback to
+        /// an older version of the package. A ramification of enabling this condition is packages
+        /// newer than the package at the head of the channel will be automatically uninstalled
+        /// during a service rollback.
+        #[structopt(name = "UPDATE_CONDITION",
+                    long = "update-condition",
+                    default_value = UpdateCondition::Latest.as_str(),
+                    possible_values = UpdateCondition::VARIANTS)]
+        update_condition:      UpdateCondition,
```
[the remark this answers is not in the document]
davidMcneil: Yes, the reason it is not at the moment is because the macro used to autogenerate the config file support is not able to support the nesting.

components/hab/src/command/pkg/uninstall.rs
```diff
@@ -141,19 +152,65 @@ pub fn start(ui: &mut UI,
...
+                return Err(SrvClientError::from(err).into());
+            }
+            _ => {
+                warn!("Unexpected status message, {:?}", reply);
+            }
+        }
+    }
+    Ok(out)
+}
+
```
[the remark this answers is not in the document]
davidMcneil: I dont like how this is laid out as well. The problem with moving them to `habitat_core` is that we are then missing several dependencies like `habitat_sup_protocol`, `habitat_common`, and `habitat_sup_client`. It does feel like we should tighten up where we put code, but I am not sure how to accomplish that without significant refactoring :disappointed:.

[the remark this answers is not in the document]
christophermaier: Maybe we could make that distinction within this crate, or between this crate and the supervisor crate... pull the common "library" code into a new module, and have facades for the CLI usage and the Supervisor usage. That might help a little bit, and might help prepare for an eventual future where the code can go where we want it to go.

components/sup/src/manager/service_updater/package_update_worker.rs
davidMcneil: I tried this and it did not work. It looks like ignoring local packages only applies in [`determine_latest_from_ident`](https://github.com/habitat-sh/habitat/blob/b652179e772ddfd5a7ee324d4e5b24e4582ebbe1/components/common/src/command/package/install.rs#L463), but then right after that we check if the package is [installed](https://github.com/habitat-sh/habitat/blob/b652179e772ddfd5a7ee324d4e5b24e4582ebbe1/components/common/src/command/package/install.rs#L466) and use the local copy if it exists.

Maybe this is a bug and should be fixed?

components/sup/src/manager/service_updater/package_update_worker.rs
davidMcneil: This is much better, thanks! Addressed in [e31171](https://github.com/habitat-sh/habitat/pull/7513/commits/e31171b0761c41c07923152087f0ca83409dcbc0)

components/sup/src/manager.rs
davidMcneil: Addressed in [79221](https://github.com/habitat-sh/habitat/pull/7513/commits/79221e07f7987043219a834e2b1f510f7848c0b9).

components/sup/src/manager.rs
davidMcneil: > This feels like a chunk of logic that could be extracted into a standalone function

Addressed in [ff708a](https://github.com/habitat-sh/habitat/pull/7513/commits/ff708a4d80cdda8257d25a8e052fe6d64d3cef4c)

> I'm also wondering if this logic should ultimately be invoked from the service updater threads themselves.

Agreed, that is a much more intuitive way to think about it and this was the original approach I was taking. I ran into a dead-end though when the package had to be not "running" inorder to uninstall it. So I moved the logic here after the service is supposed to be not "running". Turns out "running" is only determined by if a spec file exists. So the service is always running and we cannot remove the spec file because we rely on the spec file in order to restart the service.

So we could move the uninstall logic back into the update workers, but then the uninstall will happen before these lines

```rust
service.stop_gsw(shutdown_config).await;
event::service_stopped(&service);
user_config_watcher.remove(&service);
service_updater.lock().remove(&service.service_group);
```

Seems like that might have the potential to cause some problems. Thoughts?

Ultimately, when updating and restarting is a core part of the service lifecycle this should be much cleaner.

components/sup/src/manager/service_updater/package_update_worker.rs
christophermaier: Hrmm... I don't think it's a bug, so much as it's an instance of the long-standing assumption that we always run the latest package on disk and never roll back coming into conflict with the goal of this PR :sweat_smile: 

I think there is probably a way to adjust the overall installation logic to cope with this, but that's likely a big lift (or at least, too big for this particular PR). Since the `LocalPackageUsage` flag is a bit of a work-around itself, this is probably a sign that such an adjustment is due. For the purposes of the current feature, though, I think the approach you've taken will be fine. 

Would you mind filing an issue to consolidate the installation logic and more holistically account for the local package usage as well as this rollback-related behavior?

components/sup/src/manager.rs
christophermaier: Yeah, deleting the package while the service is still running could cause headaches. In most cases it should probably be OK, since the binary is already in memory, but I'm sure there would be some package out there somewhere that does something "unconventional" that would trigger all kinds of weird behavior if we deleted before stopping.

Let's be conservative for now, and keep an eye out for future cleanups as we continue to formalize the service lifecycle.

components/hab/src/command/pkg/uninstall.rs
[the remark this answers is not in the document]
davidMcneil: Addressed in [b790e8](https://github.com/habitat-sh/habitat/pull/7513/commits/b790e8f43f619d2efa9bd7c70491e98b0d467726).

components/sup/src/manager/service_updater/package_update_worker.rs
davidMcneil: Issue #7531 for package installation consolidation.

PR #7516, opened 2020-03-03
ensure that studio image builder runs the windows-service install hook

mwrock:
This fixes a bug where we were not running the windows-service install hook in the docker studio build and therefore not configuring the services config file causing the supervisor to fail in the studio. We used to suppress the install hook intentional and instead ran a script that installs the service in the docker build phase. We still suppress this hook when we install the windows-service package locally so that it does not install an acrual windows-service in the build environment. However we now call `hab pkg install` on the package during the docker build phase which will see that it has the latest version and then invoke the install hook.

### DIFF
--- a/.expeditor/scripts/release_habitat/build_docker_image.ps1
+++ b/.expeditor/scripts/release_habitat/build_docker_image.ps1
@@ -18,8 +18,8 @@ param (
 # Note that changes here should be mirrored in components/core/src/util/docker.rs
 function Get-DefaultTagForHost {
     if((docker info --format='{{.Isolation}}') -eq 'hyperv') {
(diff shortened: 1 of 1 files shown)


.expeditor/scripts/release_habitat/build_docker_image.ps1
```diff
@@ -19,7 +19,7 @@ param (
 function Get-DefaultTagForHost {
     if((docker info --format='{{.Isolation}}') -eq 'hyperv') {
         # hyperv isolation can build any version so we will default to 2016
```
smacfarlane: We should correct this comment

mwrock: doh!

PR #7526, opened 2020-03-04
Always write out service PID files

### LINKED ISSUE #7508: Retain service PID files for use by hooks only
Removing PID files for services and getting service PIDs directly from the Launcher (the true parent process of the services, and thus the source of truth for this information) is good for overall stability, but it broke an important usecase: users of `reconfigure` hooks need a way to access the PID of a service so they can, for example, send it a signal to reload its configuration files.

While the Supervisor should not go back to using PID files to determine what processes it is monitoring, it should continue to write them out for the benefit of hooks that need them.

See https://github.com/habitat-sh/habitat/issues/6076#issuecomment-588867077

christophermaier:
Previously in #7214 and then #7341, the Launcher and Supervisor were
modified so that the Launcher itself became the source of truth for
the PIDs of supervised services, rather than the contents of a file on
disk.

While this improves overall Supervisor reliability, a blanket removal
the PID files altogether can break things like `reconfigure` hooks,
which need a way to access a service's PID in order to do things like
send the service signals.

This PR re-introduces PID files for services, but _only_ as a means to
communicate with hooks. The Launcher is still the ultimate source of
the PID itself.

Fixes #7508

### DIFF
--- a/components/sup/src/manager/service/supervisor.rs
+++ b/components/sup/src/manager/service/supervisor.rs
@@ -109,7 +109,7 @@ impl Supervisor {
             self.change_state(ProcessState::Up);
         } else {
             self.change_state(ProcessState::Down);
-            Self::cleanup_pidfile(self.pid_file.clone());
+            Self::cleanup_pidfile(&self.pid_file);
         }
 
         self.pid.is_some()
@@ -221,7 +221,6 @@ impl Supervisor {
         let service_group = self.service_group.clone();
 
         if let Some(pid) = self.pid {
-            let pid_file = self.pid_file.clone();
             if pid == 0 {
                 warn!(target: "pidfile_tracing", "Cowardly refusing to stop {}, because we think it has a PID of 0, which makes no sense",
                       service_group);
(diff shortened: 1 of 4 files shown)


components/sup/src/manager/service/supervisor.rs
```diff
@@ -101,9 +109,7 @@ impl Supervisor {
             self.change_state(ProcessState::Up);
         } else {
             self.change_state(ProcessState::Down);
-            if let Some(ref pid_file) = self.pid_file {
-                Self::cleanup_pidfile(pid_file.clone());
-            }
+            Self::cleanup_pidfile(self.pid_file.clone());
```
davidMcneil: Very minor - we could change the signature of `cleanup_pidfile` to `fn cleanup_pidfile(pid_file: impl AsRef<Path>)` and avoid the `clone`. `Self::cleanup_pidfile(&self.pid_file);`

christophermaier: @davidMcneil Good catch; addressed in https://github.com/habitat-sh/habitat/pull/7526/commits/41b26e2e738f031a477fb940fc3e43474c1c7c85

Issue #7530, opened 2020-03-06
1.5.50 may not be gossiping to all peers when creating a new supervisor network


mwrock:
I have a set of terraform scripts that stand up a SQL HA cluster, Active Directory controllwe and web app (4 nodes) and I have been able to reliably have all 4 nodes peer together. As of yesterday when invoking `terraform apply` I am consistently seeing either no services gossiped in the census or only some when all are running on their individual nodes. This coincides with the 1.5.50 release. I canonly fix the network by stopping all supervisors, deleting their `sup` folders and then restarting and reloading services.

mwrock:
It should be noted that all members are in each server's butterfly endpoint.

mwrock:
I am seeing this pop up here and there in the logs:

```
thread 'push-bf664c3c28c54059980eaccbeceb1bc8' panicked at 'supplied instant is later than self', src\libcore\option.rs:1188:5
```

I'm pretty certain this will point us to the root cause

davidMcneil:
@mwrock sounds like it might have to do with timing? #7499 might be worth looking into.

mwrock:
While I feel confident this is isolated to 1.5.50 and it's time refactorings, I should document here that I did stand up several 1.5.29 instances and each of those clusters peered correctly.

PR #7533, opened 2020-03-09
Make better use of `Instant::elapsed()` for safer timing operations

christophermaier:
The use of unchecked subtraction operations with `std::time::Instant` instances in some of our time calculations can panic in some cases if the result is a negative duration.

Here, we make better use of `Instant::elapsed()`, as well as using checked subtraction operations, for equivalent and easier-to-understand code, without the possibility of panic.

Additionally, some dead code was cleaned up!

Fixes https://github.com/habitat-sh/habitat/issues/7530


components/butterfly/src/server/timing.rs
```diff
@@ -16,11 +15,11 @@ const DEPARTURE_TIMEOUT_DEFAULT_MS: u64 = 259_200_000;
 /// The timing of the outbound threads.
 #[derive(Debug, Clone)]
 pub struct Timing {
-    pub ping_ms: u64,
-    pub pingreq_ms: u64,
-    pub gossip_period_ms: u64,
-    pub suspicion_timeout_protocol_periods: u64,
-    pub departure_timeout_ms: u64,
```
davidMcneil: Nice! :tada:

components/butterfly/src/server/push.rs
```diff
@@ -107,18 +109,17 @@ fn run_loop(server: &Server, timing: &Timing) -> ! {
...
-            let wait_time = (long_wait - Instant::now()).as_millis();
-            if wait_time > 0 {
-                thread::sleep(Duration::from_millis(wait_time as u64));
-            }
+
+        // If we've still got any time left in the gossip period, sleep
+        // for that long.
+        if let Some(wait_time) = gossip_period.checked_sub(fanout_loop_start_time.elapsed()) {
+            thread::sleep(wait_time);
         }
```
davidMcneil: This pattern is used in a few other places. Would it be worth factoring it out into a function with a descriptive name (`sleep_from_start_time`?)? It might make the code easier to parse at a glance and remove the need for comments.

christophermaier: Yes! This is one of the additional cleanups I'm working on. I wanted to get the main fix out first, though, but the change you ask for will be coming soon (just maybe not in this PR) :smile:

mwrock:
I ran this a couple times in my terraform environment that was failing with 1.5.50 and both runs were successful.

PR #7545, opened 2020-03-10
Readable Cached Artifacts

### LINKED ISSUE #6298: 'hab pkg upload' fails with permission denied when uploading from /hab/cache/artifacts directory
It is a fairly useful scenario to be able to upload packages to a depot from a cached artifact directory (eg /hab/cache/artifacts).  As of 0.78 this fails with a 'permission denied' error because we are writing the downloaded artifacts with a root-only read access.  So a user has to do a 'sudo' in this case.  Also we don't give any indication of which file the permission denied is happening on, so its not super obvious to debug from a users perspective.

We should consider changing the privileges back to what they were prior to 0.78 so we don't require sudo in such cases.  The installed packages themselves are fully world readable, just the artifact is locked down.

This is a side effect of this change - https://github.com/habitat-sh/habitat/pull/5803#issuecomment-465790809

christophermaier:
This addresses #6298, building off the work @stevendanna did in #7489. In particular, it ensures that installing packages yields cached artifacts that are world readable, rather than only readable by root, on Unix systems.

It introduces `habitat_core::fs::Permissions`, a new `Option`-like type abstraction over platform-specific permission representations and makes the setting of these permissions explicit in function type signatures. Though the `download` function in our API client is used to download multiple things from Builder, this change only affects package artifacts; permissions for public and private origin keys, as well as public origin encryption keys, are _not_ manipulated at all. 

Note that these permission changes apply to _all_ artifacts that are downloaded, regardless of where they are downloaded to, or by whom.

This new `Permissions` type is now used by `AtomicWriter`, and (by extension) everything that calls it. It may prove useful for other permission-related operations elsewhere in the codebase, but that work is not attempted here.

Fixes #6298

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -16,10 +16,10 @@ use crate::{allow_std_io::AllowStdIo,
 use broadcast::BroadcastWriter;
 use bytes::BytesMut;
 use futures::stream::TryStreamExt;
-#[cfg(not(windows))]
-use habitat_core::fs::DEFAULT_CACHED_ARTIFACT_PERMISSIONS;
 use habitat_core::{crypto::keys::box_key_pair::WrappedSealedBox,
-                   fs::AtomicWriter,
+                   fs::{AtomicWriter,
+                        Permissions,
+                        DEFAULT_CACHED_ARTIFACT_PERMISSIONS},
                    package::{Identifiable,
                              PackageArchive,
                              PackageIdent,
@@ -90,24 +90,6 @@ mod json_u64 {
         let s = String::deserialize(deserializer)?;
         s.parse::<u64>().map_err(serde::de::Error::custom)
     }
-}
-
-#[cfg(windows)]
-use crate::util::win_perm;
-
-/// Abstraction over platform-specific ways to model file permissions.
-enum Permissions {
-    /// Don't take any special action to set permissions beyond what
-    /// they are "normally" set to when they are created. Here,
-    /// "normal" denotes the low-level programming library sense,
-    /// rather than any particular domain-specific sense.
-    Standard,
-    /// Indicates that a file should be created with very specific
-    /// permissions.
-    #[cfg(windows)]
-    Explicit(Vec<win_perm::PermissionEntry>),
-    #[cfg(not(windows))]
-    Explicit(u32),
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
@@ -209,9 +191,7 @@ impl BuilderAPIClient {
         let dst_file_path = dst_path.join(file_name);
         let w = {
             let mut w = AtomicWriter::new(&dst_file_path)?;
-            if let Permissions::Explicit(permissions) = permissions {
-                w.with_permissions(permissions);
-            }
+            w.with_permissions(permissions);
             w
         };
         let content_length = response::get_header(&resp, CONTENT_LENGTH);
@@ -1036,13 +1016,11 @@ impl BuilderAPIClient {
         let req_builder = self.0.get_with_custom_url(&package_download(ident), |u| {
                                     u.set_query(Some(&format!("target={}", target)))
                                 });
-
-        let permissions = if cfg!(not(windows)) {
-            Permissions::Explicit(DEFAULT_CACHED_ARTIFACT_PERMISSIONS)
-        } else {
-            Permissions::Standard
-        };
-        self.download(req_builder, dst_path.as_ref(), token, permissions, progress)
+        self.download(req_builder,
+                      dst_path.as_ref(),
+                      token,
+                      DEFAULT_CACHED_ARTIFACT_PERMISSIONS,
+                      progress)
             .await
             .map(PackageArchive::new)
     }
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -1,13 +1,6 @@
 use crate::{allow_std_io::AllowStdIo,
             error::{Error,
                     Result},
-            hab_core::{crypto::keys::box_key_pair::WrappedSealedBox,
-                       fs::AtomicWriter,
-                       package::{Identifiable,
-                                 PackageArchive,
-                                 PackageIdent,
-                                 PackageTarget},
-                       ChannelIdent},
             hab_http::ApiClient,
             response,
             BuildOnUpload,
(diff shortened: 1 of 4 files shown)


components/builder-api-client/src/builder.rs
```diff
@@ -90,6 +92,24 @@ mod json_u64 {
...
+    /// they are "normally" set to when they are created. Here,
+    /// "normal" denotes the low-level programming library sense,
+    /// rather than any particular domain-specific sense.
+    Standard,
+    /// Indicates that a file should be created with very specific
+    /// permissions.
+    #[cfg(windows)]
+    Explicit(Vec<win_perm::PermissionEntry>),
+    #[cfg(not(windows))]
+    Explicit(u32),
```
davidMcneil: Wonder if it is worth using [`std::fs::Permissions`](https://doc.rust-lang.org/stable/std/fs/struct.Permissions.html) here? It looks like it is really just a thin wrapper around a `u32` so maybe it does not buy us much. Probably fine either way.

components/builder-api-client/src/builder.rs
```diff
@@ -186,7 +207,13 @@ impl BuilderAPIClient {
         fs::create_dir_all(&dst_path)?;
         let file_name = response::get_header(&resp, X_FILENAME)?;
         let dst_file_path = dst_path.join(file_name);
-        let w = AtomicWriter::new(&dst_file_path)?;
+        let w = {
+            let mut w = AtomicWriter::new(&dst_file_path)?;
+            if let Permissions::Explicit(permissions) = permissions {
```
davidMcneil: > To make further distinctions would be tricky and a bit more complicated.

Is this where you would make that distinction? Essentially, only add the explicit permissions if it is the `/hab/cache/artifacts` directory.

mwrock:
This seems like a sound approach to me. I really don't see a problem with making artifacts universally world readable. Perhaps if a user ever became concerned with that they could `chmod` it after download.

components/builder-api-client/src/builder.rs
christophermaier: It's not that it's impossible, it's just that it would be a bit messy and would end up distributing filesystem awareness in various parts of the codebase.

components/builder-api-client/src/builder.rs
christophermaier: Yeah, I don't think that really buys us much at all; we'd still need some additional abstraction on top of that to deal with Windows stuff the way we would want to.

The main thing this type really does is provide a richer, permission-specific `Option` for us to use (as well as a facade behind which to hide all the platform-specific differences). I felt it was much better to have function signatures like 
```
do_something(..., Permissions::Standard, ...)
``` 

than

```
do_something(..., None, ...)
```

davidMcneil: I was not very clear. I definitely like the `Option` like wrapper. I was saying the enum could become:
```rust
pub enum Permissions {
    Standard,
    #[cfg(windows)]
    Explicit(Vec<win_perm::PermissionEntry>),
    #[cfg(not(windows))]
    Explicit(std::fs::Permissions),
}
```

components/builder-api-client/src/builder.rs
```diff
@@ -186,7 +189,11 @@ impl BuilderAPIClient {
         fs::create_dir_all(&dst_path)?;
         let file_name = response::get_header(&resp, X_FILENAME)?;
         let dst_file_path = dst_path.join(file_name);
-        let w = AtomicWriter::new(&dst_file_path)?;
+        let w = {
+            let mut w = AtomicWriter::new(&dst_file_path)?;
+            w.with_permissions(permissions);
+            w
+        };
```
davidMcneil: This could be slightly shortned to:
```rust
let mut w = AtomicWriter::new(&dst_file_path)?;
w.with_permissions(permissions);
let w = w;
```

Should `with_permissions` be a constructor? Which avoids the need for `w` ever being mutable.

components/core/src/fs.rs
```diff
@@ -712,21 +746,14 @@ impl AtomicWriter {
...
-    pub fn with_permissions(&mut self, mode: u32) { self.permission = Some(mode); }
-
-    #[cfg(windows)]
-    /// with_permission sets given permissions on the destination path.
-    pub fn with_permissions(&mut self, entries: Vec<win_perm::PermissionEntry>) {
-        self.permission = Some(entries);
-    }
+    /// Note: On Unix platforms, permissions are set explicitly and
+    /// not subject to the process umask.
+    pub fn with_permissions(&mut self, permissions: Permissions) { self.permissions = permissions; }
```
davidMcneil: What would you think about making this a constructor?

components/builder-api-client/src/builder.rs
christophermaier: I kinda want `AtomicWriter` to be created with a Builder pattern, but that's a whole bunch of extra code to write.

A constructor might be a nice compromise. I'll play with it.

stevendanna: I agree that a builder pattern might be nice.  I held off just because it felt silly with a single method, but if we ever get to 2 methods, then I think moving it to the builder pattern would be nice.

components/core/src/fs.rs
```diff
@@ -57,6 +57,43 @@ pub const USER_CONFIG_FILE: &str = "user.toml";
...
+/// Permissions applied to artifacts that are downloaded and/or
+/// cached. On Windows, we don't need do to anything particularly
+/// special, since artifacts will generally inherit the permissions of
+/// their containing directory.
+#[cfg(windows)]
+pub const DEFAULT_CACHED_ARTIFACT_PERMISSIONS: Permissions = Permissions::Standard;
+
+/// An `Option`-like abstraction over platform-specific ways to model
+/// file permissions.
+pub enum Permissions {
```
stevendanna: A+ 💯

components/builder-api-client/src/builder.rs
christophermaier: ah, I see what you meant :smile:

components/sup/src/manager/service/supervisor.rs
```diff
@@ -266,7 +266,7 @@ impl Supervisor {
                 // already allow broad reading of the PID file.
                 const PIDFILE_PERMISSIONS: u32 = 0o644;
                 let mut w = fs::AtomicWriter::new(pid_file)?;
-                w.with_permissions(PIDFILE_PERMISSIONS);
+                w.with_permissions(fs::Permissions::Explicit(PIDFILE_PERMISSIONS));
```
stevendanna: If we wanted we could move the platform conditionality just around the definition of the pidfile permissions now that we have your nice abstraction.

components/core/src/fs.rs
christophermaier: Yup! Handled in https://github.com/habitat-sh/habitat/pull/7545/commits/a63b8a0577fcc27dea88b67befe917c51dfd790c

components/builder-api-client/src/builder.rs
christophermaier: Worked out pretty nicely: https://github.com/habitat-sh/habitat/pull/7545/commits/a63b8a0577fcc27dea88b67befe917c51dfd790c

components/sup/src/manager/service/supervisor.rs
christophermaier: Good call! Addressed in https://github.com/habitat-sh/habitat/pull/7545/commits/4379e7d7b350fdf216d2670dd71a569b1ca42c40

PR #7548, opened 2020-03-11
allow plan context to be rooted in target folder

### LINKED ISSUE #1952: Need target/shell specific hooks
Currently the supervisor will run hooks using the interpreter appropriate to the package target. While it is straightforward to writye bash and powershell plans for the same service, it is not for hooks because they currently have no extension.

We should allow hooks to have an optional `.sh` or `.ps1` extension in order to allow a single service to have different hooks for different targets.

mwrock:
closes #1952


components/plan-build/bin/hab-plan-build.sh
```diff
@@ -2401,38 +2401,39 @@ THIS_PROGRAM=$(abspath "$0")
...
+if [[ ${#places[@]} -eq 0 ]]; then
+  for place in "${candidatePlaces[@]}"; do
+    if [[ -f $place ]]; then
+      places+=("$place")
+      PLAN_CONTEXT="$(dirname "$place")"
+    fi
+  done
+fi
+if [[ ${#places[@]} -gt 1 ]]; then
+    exit_with "A Plan file was found in the following places: $(join_by ',' "${places[@]}"). Only one is allowed at a time" 42
```
smacfarlane: Nit: `s/places/paths/` for consistency with the language below

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1532,51 +1532,43 @@ if (-Not (Test-Path "$Context")) {
...
+if($places.Count -eq 0) {
+    foreach($place in $candidatePlaces) {
+        if(Test-Path $place) {
+            $places += $place
+            $script:PLAN_CONTEXT = (Split-Path $place -Parent)
+        }
+    }
+}
+if($places.Count -gt 1) {
+    throw "A Plan file was found in the following places: $($places -Join ', '). Only one is allowed at a time"
```
smacfarlane: Nit: `s/places/paths/` for consistency with the language below

mwrock:
Just fixed the nit and also changed the camel casing to snake casing in the bash.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -2401,38 +2401,39 @@ THIS_PROGRAM=$(abspath "$0")
...
+      PLAN_CONTEXT="$(dirname "$path")"
+    fi
+  done
+fi
+if [[ ${#paths[@]} -gt 1 ]]; then
+    exit_with "A Plan file was found in the following paths: $(join_by ',' "${paths[@]}"). Only one is allowed at a time" 42
+fi
+if [[ ${#paths[@]} -eq 0 ]]; then
+    all_paths=( "${candidate_paths[@]}" "${candidate_target_paths[@]}" )
+    exit_with "Plan file not found in any of these paths: $(join_by ',' "${all_paths[@]}")" 42
```
christophermaier: My suggestions for the Powershell implementation apply here, as well.

www/source/docs/developing-packages.html.md.erb
```diff
@@ -87,6 +87,47 @@ All plans must have a `plan.sh` or `plan.ps1` at the root of the plan context. T
...
+    └── hooks/
+            run
+```
+
+The build script will look for the base of your plan in the following locations:
+
+- `<app_root>/`
+- `<app_root>/habitat/`
+- `<app_root>/<target>/`
+- `<app_root>/habitat/<target>/`
```
christophermaier: Again, I feel like it's a bit more natural and clear to list the locations in order of preference. At least, that's the order I would expect to read them in.

The fact that `<app_root>/habitat/` and `<app_root>/<target>` aren't swapped here is a little odd to me, in that it kind of turns the natural hierarchical nature of the filesystem inside-out.

www/source/docs/developing-packages.html.md.erb
```diff
@@ -87,6 +87,47 @@ All plans must have a `plan.sh` or `plan.ps1` at the root of the plan context. T
...
+
+If it finds a plan both inside as well as outside of a target folder, the target specific plan will be used as long as that is the target currently being built.
```
christophermaier: This seems like something we'd want to warn about (at the very least); I think it might justify stopping the build, though (assuming that doesn't break a bunch of existing plans).

Having a mix of locations for a single package seems like a recipe for confusion and disaster.

Is there a reason we allow `<app_root>/<target>` in the first place? It seems that once you have to specify a target, you've got more than one, by definition, at which point _requiring_ them to all be contained in `<app_root>/habitat` makes a lot of sense. I suppose if you're just dealing with a pure plan repository, and not code + plans, target directories in the root makes sense. Perhaps drawing that distinction here in the documentation would be good; we should be encouraging people towards the patterns they should be using, and saying why you'd want to use one over the other, rather than just saying "you can do X, Y, or Z; have fun!"

smacfarlane: >Is there a reason we allow <app_root>/<target> 
>I suppose if you're just dealing with a pure plan repository

That's essentially it, core plans don't use the `habitat` directory, so for the time being we'd need to support that case.  We could make that change, but I think that warrants a larger discussion than this pr around how we manage changes to Plans (the entire structure of plan.sh/ps1, hooks, variants, etc.) 

:+1: on encouraging people to use the appropriate pattern for their use case though.

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1532,51 +1532,43 @@ if (-Not (Test-Path "$Context")) {
...
+    }
+}
+if($paths.Count -eq 0) {
+    foreach($place in $candidatePaths) {
+        if(Test-Path $place) {
+            $paths += $place
+            $script:PLAN_CONTEXT = (Split-Path $place -Parent)
+        }
+    }
+}
```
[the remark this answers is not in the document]
mwrock: I'll add some clarifying comments to the code but the answer is that we want to be able to see if the user has more than one candidate plan later triggering the `if($paths.Count -gt 1)` error condition. We do need to track the target and non-target candidates separately because it is OK to have a target plan and a non-target candidate plan. In such a case the target plan takes precedence.

www/source/docs/developing-packages.html.md.erb
mwrock: > This seems like something we'd want to warn about (at the very least); I think it might justify stopping the build, though (assuming that doesn't break a bunch of existing plans).

> Having a mix of locations for a single package seems like a recipe for confusion and disaster.

I agree a warning would be good here and will add that. We don't fail because that would break some core plans (like glibc as one example). However after the refresh is complete, it may make sense to correct those plans and then adjust here.

mwrock:
OK I have addressed all comments and retested @christophermaier and @smacfarlane

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1550,20 +1555,32 @@ $candidateTargetPaths = @(
...
     }
 }
+
+if($paths.Count -gt 0 -and ($targetPaths.Count -gt 0)) {
+    Write-Warning "There is a plan.ps1 inside $pkg_target and outside as well. Using the plan in $pkg_target."
+    Write-Warning "It is advisable to either remove the plan that is outside $pkg_target"
+    Write-Warning "or move that plan to its own target folder if it is intended for a different target."
+}
+if($targetPaths.Count -gt 0) {
+    $paths = $targetPaths
```
christophermaier: This operation coupled with the `if / throw` statements below is more confusing and error prone than it should be. Perhaps creating a `$finalPaths` variable would be good, since that makes things explicit, and keeps the `$paths` and `$targetPaths` variables single-purpose.

components/plan-build-ps1/bin/hab-plan-build.ps1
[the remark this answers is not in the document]
christophermaier: I don't think this logic works as described.

Here's the overall order in which things are checked (targets first, then "plain" plans):
- "$PLAN_CONTEXT\$pkg_target\plan.ps1",
- "$PLAN_CONTEXT\habitat\$pkg_target\plan.ps1"
- "$PLAN_CONTEXT\plan.ps1",
- "$PLAN_CONTEXT\habitat\plan.ps1"

If you have a plan at `$pkg_target/plan.ps1` and `plan.ps1`, the logic below will claim to use `$pkg_target/plan.ps1`, but the `$script:PLAN_CONTEXT` variable would have been set for the `plan.ps1` plan. This is because the value of `$script:PLAN_CONTEXT` is set to the last existing path that we check, regardless of what came before.

I understand the need to track targeted plans separately from non-targeted plans, but I think this could be simplified by gathering just the path information first, checking to see if you've got the right number (issuing any relevant warnings in the process), and then once you are certain you have one path to operate on, determine what the value of `$script:PLAN_CONTEXT` is. In other words, only do that once, rather than multiple times in a loop. That decouples the two logical operations that we are doing and makes it much more clear what is going on.

The same concerns apply for the Bash implementation.

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1550,20 +1555,32 @@ $candidateTargetPaths = @(
     "$PLAN_CONTEXT\$pkg_target\plan.ps1",
     "$PLAN_CONTEXT\habitat\$pkg_target\plan.ps1"
 )
-foreach($place in $candidateTargetPaths) {
-    if(Test-Path $place) {
-        $paths += $place
-        $script:PLAN_CONTEXT = (Split-Path $place -Parent)
+foreach($path in $candidateTargetPaths) {
```
christophermaier: As a point of clarity, if we're checking the `$candidateTargetPaths` first, it would make the code a bit easier to read and understand by declaring that array before the `$candidatePaths` one, just to maintain parallel structure.

components/plan-build-ps1/bin/hab-plan-build.ps1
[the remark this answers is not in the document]
christophermaier: I'm also wondering if using the variable name `$PLAN_CONTEXT` here is appropriate. We're actually trying to figure out what the ultimate `$PLAN_CONTEXT` is, right? So `$PLAN_CONTEXT` at _this_ point is really just the initial or root directory that we're searching in, correct? We don't know what the plan is, so we _can't_ know what that plan's context is.

Not overloading variable names (particularly in this corner of the code :sweat_smile:) is going to make it much easier to maintain this code.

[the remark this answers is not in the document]
mwrock: > If you have a plan at $pkg_target/plan.ps1 and plan.ps1, the logic below will claim to use $pkg_target/plan.ps1, but the $script:PLAN_CONTEXT variable would have been set for the plan.ps1 plan. This is because the value of $script:PLAN_CONTEXT is set to the last existing path that we check, regardless of what came before.

Son of a...@$$$#!

PR #7557, opened 2020-03-16
 fix failing test where windows and linux expect different error codes

mwrock:
Linux and Windows retrurn different exit codes on failure. This fixes the e2e test on linux and also works on windows by simply checking for a non 0 value.

One can argue that we should return the same exit codes on both platforms but I would like to defer that to a separate PR.


mwrock:
Filed #7558 to address the exit codes.

PR #7560, opened 2020-03-17
only stop the studio container at the end of the docker studio supervisor test

### LINKED ISSUE #7538: Investigate 127 errors in end to end pipeline on windows builds
We have been seeing alot of these errors occuring in Windows builds in our e2e pipeline:

```
🚨 Error: The command exited with status 127
```

It is not clear where these are coming from since they seem to occur in random points of execution.

mwrock:
fixes #7538 

We need to forcefully stop a studio container in one of the docker host tests. It ends up that this same host is either hosting other test containers or connects to a docker daemon that is testing the other container tests. Therefore we were stopping all of the tests causing them to exit with `127`. This filters to only stop a studio container.

### DIFF
--- a/test/end-to-end/test_studio_supervisor.ps1
+++ b/test/end-to-end/test_studio_supervisor.ps1
@@ -2,17 +2,8 @@ $cliVersion = ((hab --version) -split " ")[1]
 $env:HAB_STUDIO_SECRET_HAB_INTERNAL_BLDR_CHANNEL="dev"
 hab origin key generate $env:HAB_ORIGIN
 
-# Note we kill all containers before and after this test
-# At some point we can remove the BeforeAll but neet it now
-# for straggler test VMs
 
 Describe "Studio supervisor" {
-    BeforeAll {
-        if($env:DOCKER_STUDIO_TEST) {
-            docker ps -q | ForEach-Object { docker stop $_ }
-        }
-    }
-
     It "version should match hab cli" {
         $result = (Invoke-StudioRun "hab sup --version")
         $result[-1] | Should -match "sup $(($cliVersion -split '/')[0])/*"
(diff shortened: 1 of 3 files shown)


test/end-to-end/test_studio_supervisor.ps1
```diff
@@ -2,17 +2,7 @@ $cliVersion = ((hab --version) -split " ")[1]
 $env:HAB_STUDIO_SECRET_HAB_INTERNAL_BLDR_CHANNEL="dev"
 hab origin key generate $env:HAB_ORIGIN
 
-# Note we kill all containers before and after this test
-# At some point we can remove the BeforeAll but neet it now
-# for straggler test VMs
-
 Describe "Studio supervisor" {
-    BeforeAll {
```
markan: Would there still be a need to cleanup things prior just to make sure we don't have any zombies from prior runs? Wondering whether we might still need that.

test/end-to-end/test_studio_supervisor.ps1
```diff
@@ -37,7 +27,7 @@ Describe "Studio supervisor" {
 
     AfterAll {
         if($env:DOCKER_STUDIO_TEST) {
-            docker ps -q | ForEach-Object { docker stop $_ }
+            docker ps -q --filter "ancestor=habitat/win-studio-x86_64-windows:ltsc2019-$(($cliVersion -split '/')[0])" | ForEach-Object { docker stop $_ }
```
smacfarlane: Definitely an improvement over before.  This reminded me that we are on a shared host (I think), so docker stop of everything may have been catching containers that were not ours.  This _should_ be safe from that, as at this point we should be the only ones running this container version.   Ad-hoc runs of this pipeline should be safe as well, as those are typically done off of feature branches which will have an unreleased version.

smacfarlane: For example, this is what the automate team does: https://github.com/chef/automate/blob/master/.buildkite/hooks/pre-exit#L10

mwrock: I made some changes to take a similar approach.

test/end-to-end/test_studio_supervisor.ps1
mwrock: I made some changes so that we are more guaranteed to only stop the container that the current build spun up. While this means that concurrent e2e tests will not trample on each other it also opens the possibility that a cancelled build cancelled at the right time could leave an orphaned container. Given the nature of the e2e pipeline, canceled build are less likely to occur and I don't forsee this to be a problem but it is something we should be aware of.

PR #7564, opened 2020-03-18
Correct some e2e behavior

smacfarlane:
This PR addresses a few issues discovered while attempting to test packages built against the upcoming refresh. 

A number of tests will timeout if the dependencies for the packages aren't already on disk. This can occur if the base image `chefes/buildkite` doesn't already have the packages cached.  Since these tests aren't specifically testing install logic, it should be safe to pre-install them using the specified $HAB_BLDR_CHANNEL. An alternative approach would be to increase the timeout values, but they'd have to be made arbitrarily large to minimize the risk of timeout. 

This also disables the promotion step via Buildkite, if the creator isn't Chef Expeditor. This gives us an extra level of safety beyond `maybe_run` from an accidental promotion. 

The final change is to remove hard-coded `HAB_BLDR_CHANNEL` in tests, in favor of letting the environment drive that. This moves us close to being able to run these e2e tests against an arbitrary channel in Builder, which will be useful for validation of pre-release changes. 

### DIFF
--- a/test/end-to-end/test_launcher_checks_supervisor_version.ps1
+++ b/test/end-to-end/test_launcher_checks_supervisor_version.ps1
@@ -41,7 +41,7 @@ function Test-LauncherFailure($version) {
 }
 
 # These tests will timeout unless the launcher and all of its dependencies
-# are already on disk. 
+# are already on disk.
 hab pkg install core/hab-launcher --channel $HAB_BLDR_CHANNEL
 
 Describe "Launcher version check" {
--- a/test/end-to-end/test_launcher_checks_supervisor_version.ps1
+++ b/test/end-to-end/test_launcher_checks_supervisor_version.ps1
@@ -40,6 +40,10 @@ function Test-LauncherFailure($version) {
     return ($supPid.ExitCode -ne 0)
 }
 
+# These tests will timeout unless the launcher and all of its dependencies
+# are already on disk. 
+hab pkg install core/hab-launcher --channel $HAB_BLDR_CHANNEL
+
 Describe "Launcher version check" {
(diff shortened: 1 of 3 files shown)


smacfarlane:
Passing run based off of this branch: https://buildkite.com/chef/habitat-sh-habitat-master-end-to-end/builds/513#2d85d1c9-1416-4852-836c-92901e29235d

test/end-to-end/test_supervisor_term.ps1
```diff
@@ -1,5 +1,3 @@
-$env:HAB_INTERNAL_BLDR_CHANNEL = "dev"
-$env:HAB_BLDR_CHANNEL = "dev"
```
mwrock: OMG

test/end-to-end/test_launcher_checks_supervisor_version.ps1
```diff
@@ -40,6 +40,10 @@ function Test-LauncherFailure($version) {
     return ($supPid.ExitCode -ne 0)
 }
 
+# These tests will timeout unless the launcher and all of its dependencies
+# are already on disk. 
+hab pkg install core/hab-launcher --channel $HAB_BLDR_CHANNEL
+
```
mwrock: This is good but it makes me wonder if we should be adding the sup and launcher to the `setup_environment.ps1` and `.sh` e2e scripts.

smacfarlane: I thought about that, but thought it best to avoid installing _most_ components in the shared scripts for now in favor of being explicit about what we expect on the system.

PR #7565, opened 2020-03-18
Add testing steps for core-plans refresh

smacfarlane:
This adds some documentation around the process used to test against a core-plans refresh.  Ideally in the future we'd be able to leverage buildkite to do this work for us, but our current pipelines are designed around releasing and need additional work.  For now, it's captured here if we need to go through this process again before those improvements are made. 

### DIFF
--- a/END_TO_END_TESTING.md
+++ b/END_TO_END_TESTING.md
@@ -25,7 +25,7 @@ Also see the [end-to-end pipeline definition](./.expeditor/end_to_end.pipeline.y
 ## Testing changes before they're built
 
 We currently don't have a good mechanism for building a set of packages off of a branch. The release pipeline 
-can be used for this purpose, however there is always the risk of change causing an unindended release. Additionally, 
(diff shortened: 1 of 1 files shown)


END_TO_END_TESTING.md
```diff
@@ -8,14 +8,42 @@ the result.
...
 Ideally, test files will take take no arguments, but can respond to
 environment variables. The variables used for local testing are
 currently found in [e2e_env](./e2e_env).
 
 Also see the [end-to-end pipeline definition](./.expeditor/end_to_end.pipeline.yml) for additional background.
+
+## Testing changes before they're built
+
+We currently don't have a good mechanism for building a set of packages off of a branch. The release pipeline 
+can be used for this purpose, however there is always the risk of change causing an unindended release. Additionally,
```
mwrock: unindended -> unintended

smacfarlane: :man_facepalming:  Fixed.

PR #1230, opened 2016-09-14
Add Habitat Community Summit text to global message

magwalk:
Response to Kayleigh's request to add the banner ad to habitat.sh (https://chefio.atlassian.net/browse/CIA-1076)

screen shot 2016-09-14 at 10 16 04 am

PR #7577, opened 2020-03-21
Move builder-worker post release step higher in the release readme

mwrock:
Currently one of the first post-release tasks mentioned in the `RELEASE.md` is to verify that acceptance is using the new backline. However it will not use the new backline until the new builder-workers have been promoted which is mentioned later in the readme.

This moves the notes about the builder-worker higher up in the document. Then it removes the section about verifying backline because verifying backline is part of the builder-worker promotion and everything one needs to do to verify it is covered in the builder worker notes.

One point up for discusion here is the section [Update the Acceptance environment with the new hab-backline](https://github.com/habitat-sh/habitat/blob/master/RELEASE.md#update-the-acceptance-environment-with-the-new-hab-backline). I am confused about this section and have the folllowing questions:

* Why do we need to upload the current unstable backline packages to acceptance?
* Is the actual intent of this section to upload the `stable` versions to acceptance because they are missing.

My guess is that this section is outdated and should be updated to cover syncing the acceptance hab packages with what is in stable. Does that sound correct?

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -205,10 +205,17 @@ making them "officially" available to the world, and thus "released".
 # Post-Release Tasks
 The Buildkite release is fairly-well automated at this point, but once it is complete, there are still a few remaining manual tasks to perform. In time, these will be automated as well.
 
-## Verify the Acceptance environment is using the new hab-backline
+## The Builder Worker
 
-Running [`update-hab-backline.sh`](https://github.com/habitat-sh/habitat/blob/master/update-hab-backline.sh)
-is handled by buildkite. If it is necessary to do manually, you can find instructions in [a previous release of this file.](https://github.com/habitat-sh/habitat/blob/bebf0fdfb738e1304ea201717fb6054733b17939/RELEASE.md#update-the-acceptance-environment-with-the-new-hab-backline)
+New `habitat/builder-worker` packages will automatically be built by
+Builder's `post_habitat_release` pipeline ([definition](https://github.com/habitat-sh/builder/blob/master/.expeditor/post_habitat_release.pipeline.yml)),
+[Buildkite
+pipeline](https://buildkite.com/chef/habitat-sh-builder-master-post-habitat-release))
+when a Habitat release completes. Please follow along in that pipeline
+to ensure that the new packages are validated and properly promoted to
+the stable channel.
+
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -205,10 +205,17 @@ making them "officially" available to the world, and thus "released".
...
-is handled by buildkite. If it is necessary to do manually, you can find instructions in [a previous release of this file.](https://github.com/habitat-sh/habitat/blob/bebf0fdfb738e1304ea201717fb6054733b17939/RELEASE.md#update-the-acceptance-environment-with-the-new-hab-backline)
+New `habitat/builder-worker` packages will automatically be built by
+Builder's `post_habitat_release` pipeline ([definition](https://github.com/habitat-sh/builder/blob/master/.expeditor/post_habitat_release.pipeline.yml)),
+[Buildkite
+pipeline](https://buildkite.com/chef/habitat-sh-builder-master-post-habitat-release))
+when a Habitat release completes. Please follow along in that pipeline
+to ensure that the new packages are validated and properly promoted to
+the stable channel.
+
+After building the workers and promoting them to acceptance, you will be prompted to build packages in acceptance to ensure that they are behaving propperly. You should build both a Windows and a Linux package. The key things you are looking for are that the builds succeed and that their build output indicates they are using the new stable versions of `hab`, `hab-plan-build` and `hab-studio`.
```
markan: Where is this prompt coming from? 

Also nit: 'properly'

davidMcneil: I actually address this in this in #7579. I will open up a PR to this branch so we can address all release changes at once.

mwrock:
after discussing further with @smacfarlane, it sounds like my assumption above is correct and much of the content in [Update the Acceptance environment with the new hab-backline](https://github.com/habitat-sh/habitat/blob/master/RELEASE.md#update-the-acceptance-environment-with-the-new-hab-backline) is outdated.

I have changed those instructions in this PR to cover updating the backline packages in the acceptance bldr depot with the new released stable packages.

mwrock:
As I have been editing `Update the Acceptance environment with the new hab-backline` I now remember why it updated prod/unstable backline to acceptance/stable. We used to expect calls to acceptance builder to give us the latest `-dev` packages. So I am just deleting that entire section here to eliminate confusion and we can let #1230 take care of getting the packages synced.

PR #7580, opened 2020-03-23
Add to release instructions

davidMcneil:

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -88,8 +88,8 @@ hab pkg install core/windows-service --channel=staging
 ### What to Test
 
 There may be special behavior related to this release that you will
-want to validate but at the very least, you should try running some
-services, exporting some services, and exercising the Studio.
+want to validate but at the very least, you should validate the version,
+try running some services, exporting some services, and exercising the Studio.
 
 To ensure everything is working properly, you will need to have the
 following environment variables set:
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -208,14 +230,16 @@ The Buildkite release is fairly-well automated at this point, but once it is com
...
-the stable channel.
+when a Habitat release completes. The instructions below describe how to
+follow along in that pipeline to ensure that the new packages are validated and
+properly promoted to the stable channel.
+
+After the workers are built and promoted to acceptance by the pipeline, you will be prompted to "Evaluate habitat/builder-worker in Acceptance" in the pipeline. Use these [instructions](https://github.com/habitat-sh/builder/pull/1233) to connect to the `acceptance-worker` nodes. After connecting to the node, validate the version of the Supervisor and the version of `habitat/builder-worker`. An easy way to check the version is with `hab sup status`, the version should match the version output by the pipeline. Next, build a package in acceptance to ensure that the workers are behaving properly. This can be done through the [acceptance builder ui](https://bldr.acceptance.habitat.sh/). The key things you are looking for are that the builds succeed and that their build output indicates they are using the new stable versions of `hab`, `hab-plan-build` and `hab-studio`. You should follow these steps for both a Windows and a Linux worker.
```
mwrock: After thinking about this last week, I don't think it is necessary to ssh/rdp to the workers. I think one just needs to initiate a linux/windows/linux2 build and validate the versions of hab-studio, hab and hab-plan-build being used by looking at the build log. Brcause the workers take a dependency on hab-studio, if the new versions show up in the log, it would have to be a new worker. Because an older worker would have been built with a previous srudio that would have included the older hab and hab-plan-build.

As for confitming the version of the Supervisor, I think that is overkill here and is actually a new concept entirely we are introducing to the process. I know I was the one who mentioned we should check this during the last post-release steps but I think I suggested that out of ignorance.

I think if we could eliminate the need to remote to the workers that would make the release alot smoother. However I am going to add an issue to the automation epic suggesting we add worker and supervisor versions to the top of the build log.

davidMcneil: Agreed, that makes sense. Addressed in [28bbaa](https://github.com/habitat-sh/habitat/pull/7580/commits/28bbaaa60686aa86fee58624b60607751fe353a0)

PR #7582, opened 2020-03-26
Automatic package cleanup with the `--keep-latest-packages` flag

### LINKED ISSUE #7552: Allow Supervisor to purge old versions of a service
As the Supervisor updates its services (and itself), we can accumulate a large number of old versions of the package on disk. Over time on long-lived systems, this can result in excessive disk usage. This is exacerbated with larger numbers of services, as well as with frequently-updating services.

Since we have (almost?) all the necessary logic for removing packages and dependencies present in the `hab pkg uninstall` command, we should be able to utilize this in the Supervisor directly. For instance, imagine a mode in which the Supervisor keeps only the most recent 5 releases of services it is running (configurable, of course). As it updates for the 6th time, it would delete the oldest version on disk, as well as any dependencies of that release that weren't being used by any other service that was running.

When a service is unloaded, we could remove all releases of that service on disk.

I'm not sure if it makes more sense to have this capability enabled Supervisor-wide, or on a service-by-service basis. The former seems to be more straightforward at the moment, but arguments could be made for the latter.

davidMcneil:
Resolves #7552 

This PR adds the `--keep-latest-packages` flag to the `hab sup run` command.

```
--keep-latest-packages <KEEP_LATEST_PACKAGES>
    Automatically cleanup old packages.
    
    If this flag is enabled, service startup will initiate an uninstall of all previous versions of the
    associated package. This also applies when a service is restarted due to an update. If a number is passed to
    this argument, that number of latest versions will be kept. The same logic applies to the Supervisor
    package. [env: HAB_KEEP_LATEST_PACKAGES=]  [default: 1]

```

I went with `--keep-latest-packages` argument name instead of just `--keep-latest` to try and better indicate the purpose of the argument. Which name do reviewers prefer?

### DIFF
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -1277,29 +1277,23 @@ fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
                              `127.0.0.1`")
     );
 
-    // The clap_app macro does not allow numeric default values
+    // clap_app macro does not allow setting short and long help seperately
     let sub =
         sub.arg(Arg::with_name("KEEP_LATEST_PACKAGES").long("keep-latest-packages")
                                                       .takes_value(true)
-                                                      .min_values(0)
-                                                      .max_values(2)
-                                                      .default_value("1")
                                                       .validator(valid_numeric::<usize>)
                                                       .env("HAB_KEEP_LATEST_PACKAGES")
                                                       .help("Automatically cleanup old packages")
                                                       .long_help("Automatically cleanup old \
-                                                                  packages.\n\nIf this flag is \
-                                                                  enabled, service startup will \
-                                                                  initiate an uninstall of all \
-                                                                  previous versions of the \
-                                                                  associated package. This also \
-                                                                  applies when a service is \
-                                                                  restarted due to an update. \
-                                                                  If a number is passed to this \
-                                                                  argument, that number of \
-                                                                  latest versions will be kept. \
-                                                                  The same logic applies to the \
-                                                                  Supervisor package."));
+                                                                  packages.\n\nIf enabled, \
+                                                                  service startup will initiate \
+                                                                  an uninstall of all packages \
+                                                                  except for the \
+                                                                  `KEEP_LATEST_PACKAGES` most \
+                                                                  recent packages. The same \
+                                                                  logic applies to the \
+                                                                  `core/hab-sup` package on \
+                                                                  Supervisor startup."));
 
     // The clap_app macro does not allow "-" in possible values
     let sub = sub.arg(Arg::with_name("UPDATE_CONDITION").long("update-condition")
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -1277,6 +1277,33 @@ fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
                              `127.0.0.1`")
     );
 
+    // The clap_app macro does not allow numeric default values
+    let sub =
+        sub.arg(Arg::with_name("KEEP_LATEST_PACKAGES").long("keep-latest-packages")
+                                                      .takes_value(true)
+                                                      .min_values(0)
+                                                      .max_values(2)
+                                                      .default_value("1")
+                                                      .validator(valid_numeric::<usize>)
+                                                      .env("HAB_KEEP_LATEST_PACKAGES")
+                                                      .help("Automatically cleanup old package \
+                                                             versions")
+                                                      .long_help("Automatically cleanup old \
+                                                                  package versions.\n\nIf this \
+                                                                  flag is enabled, service \
+                                                                  startup will initiate an \
+                                                                  uninstall of all previous \
+                                                                  versions of the associated \
+                                                                  package. This also applies \
+                                                                  when a service is restarted \
+                                                                  due to an update. If a number \
+                                                                  is passed to this argument, \
+                                                                  that number of latest \
+                                                                  versions will be kept. The \
+                                                                  same logic applies to the \
+                                                                  Supervisor package if self \
+                                                                  updates are enabled."));
+
     // The clap_app macro does not allow "-" in possible values
     let sub = sub.arg(Arg::with_name("UPDATE_CONDITION").long("update-condition")
                                                         .takes_value(true)
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -1279,21 +1279,29 @@ fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
 
     // clap_app macro does not allow setting short and long help seperately
     let sub =
-        sub.arg(Arg::with_name("KEEP_LATEST_PACKAGES").long("keep-latest-packages")
-                                                      .takes_value(true)
-                                                      .validator(valid_numeric::<usize>)
-                                                      .env("HAB_KEEP_LATEST_PACKAGES")
-                                                      .help("Automatically cleanup old packages")
-                                                      .long_help("Automatically cleanup old \
-                                                                  packages.\n\nIf enabled, \
-                                                                  service startup will initiate \
-                                                                  an uninstall of all packages \
-                                                                  except for the \
-                                                                  `KEEP_LATEST_PACKAGES` most \
-                                                                  recent packages. The same \
-                                                                  logic applies to the \
-                                                                  `core/hab-sup` package on \
-                                                                  Supervisor startup."));
+        sub.arg(Arg::with_name("NUM_LATEST_PACKAGES_TO_KEEP").long("keep-latest-packages")
+                                                             .takes_value(true)
+                                                             .validator(valid_numeric::<usize>)
+                                                             .env("HAB_KEEP_LATEST_PACKAGES")
+                                                             .help("Automatically cleanup old \
+                                                                    packages")
+                                                             .long_help("Automatically cleanup \
+                                                                         old packages.\n\nIf \
+                                                                         enabled, service \
+                                                                         startup will initiate \
+                                                                         an uninstall of all \
+                                                                         packages except for \
+                                                                         the `NUM_LATEST_PACKAGES_TO_KEEP` most recent \
+                                                                         packages. This also \
+                                                                         applies when a service \
+                                                                         is restarted due to an \
+                                                                         update. The same logic \
+                                                                         applies to the \
+                                                                         `core/hab-sup` package \
+                                                                         on Supervisor startup. \
+                                                                         If this argument is \
+                                                                         not set no package \
+                                                                         cleanup is performed."));
 
     // The clap_app macro does not allow "-" in possible values
     let sub = sub.arg(Arg::with_name("UPDATE_CONDITION").long("update-condition")
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -1286,21 +1286,15 @@ fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
                                                              .help("Automatically cleanup old \
                                                                     packages")
                                                              .long_help("Automatically cleanup \
-                                                                         old packages.\n\nIf \
-                                                                         enabled, service \
-                                                                         startup will initiate \
-                                                                         an uninstall of all \
-                                                                         packages except for \
-                                                                         the `NUM_LATEST_PACKAGES_TO_KEEP` most recent \
-                                                                         packages. This also \
-                                                                         applies when a service \
-                                                                         is restarted due to an \
-                                                                         update. The same logic \
-                                                                         applies to the \
-                                                                         `core/hab-sup` package \
-                                                                         on Supervisor startup. \
-                                                                         If this argument is \
-                                                                         not set no package \
+                                                                         old packages.\n\nThe \
+                                                                         Supervisor will \
+                                                                         automatically cleanup \
+                                                                         old packages only \
+                                                                         keeping the `NUM_LATEST_PACKAGES_TO_KEEP` latest \
+                                                                         packages. If this \
+                                                                         argument is not \
+                                                                         specified, no \
+                                                                         automatic package \
                                                                          cleanup is performed."));
 
     // The clap_app macro does not allow "-" in possible values
--- a/components/hab/src/cli.rs
+++ b/components/hab/src/cli.rs
@@ -1286,23 +1286,20 @@ fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
                                                       .default_value("1")
                                                       .validator(valid_numeric::<usize>)
                                                       .env("HAB_KEEP_LATEST_PACKAGES")
-                                                      .help("Automatically cleanup old package \
-                                                             versions")
+                                                      .help("Automatically cleanup old packages")
                                                       .long_help("Automatically cleanup old \
-                                                                  package versions.\n\nIf this \
-                                                                  flag is enabled, service \
-                                                                  startup will initiate an \
-                                                                  uninstall of all previous \
-                                                                  versions of the associated \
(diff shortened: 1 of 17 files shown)


components/hab/src/command/pkg/uninstall_impl.rs
```diff
@@ -223,6 +225,22 @@ async fn supervisor_services() -> Result<Vec<PackageIdent>> {
     Ok(out)
 }
 
+#[derive(Clone, Copy)]
+enum UninstallIfRunning<'a> {
+    EvenIfRunning,
+    SkipIfRunning(&'a [PackageIdent]),
+}
```
christophermaier: The naming of this is a little off, in that you can get statements like `UninstallIfLoaded::SkipIfLoaded(...)`, which seem contradictory.

Perhaps something like `UninstallMode::SkipIfLoaded`, etc.? Admittedly, I'm not in love with that name, but it feels like it's moving in a good direction (IMO, at least :sweat_smile: )

components/hab/src/cli/hab/sup.rs
```diff
@@ -338,6 +338,21 @@ pub struct SupRun {
...
+    /// If this flag is enabled, service startup will initiate an uninstall of all previous
+    /// versions of the associated package. This also applies when a service is restarted due to an
+    /// update. If a number is passed to this argument, that number of latest versions will be
+    /// kept. The same logic applies to the Supervisor package if self updates are enabled.
+    // TODO (DM): We want this to have a default value of 1 if set otherwise it should be none. Does
+    // Option<Option<T>> achieve this
+    #[structopt(name = "KEEP_LATEST_PACKAGES",
+                long = "keep-latest-packages",
+                default_value = "1",
+                max_values = 2, // This is needed to use an env var
```
christophermaier: :thinking: That's unexpected :thinking:

components/hab/src/cli/hab/sup.rs
```diff
@@ -338,6 +338,21 @@ pub struct SupRun {
...
+                env = "HAB_KEEP_LATEST_PACKAGES")]
+    #[allow(clippy::option_option)]
+    keep_latest_packages: Option<Option<usize>>,
```
christophermaier: I think this is the first time I've ever seen an optional Option :rofl:

Since this is weird enough to have its own Clippy lint, I wonder if we could create a custom type that better encapsulates the behavior we're after here... :thinking: 

(On further reading, I'm not 100% clear what the nested `Option` indicates... it seems like we either have `None`, or we have `Some(non-zero number)`. Is it required in this form to play nicely with Clap somehow?)

components/sup/src/util/pkg.rs
```diff
@@ -125,11 +126,46 @@ pub async fn install_channel_head(url: &str,
...
+/// Returns the number of packages that were uninstalled
+pub async fn uninstall_all_but_latest(ident: impl AsRef<PackageIdent>,
+                                      number_latest_to_keep: usize)
+                                      -> HabResult<usize> {
+    let ident = ident.as_ref().clone();
+    let mut idents = list::package_list(&ident.into())?;
+    if number_latest_to_keep >= idents.len() {
+        return Ok(0);
+    }
+    idents.reverse(); // The packages to keep now occur first in the list
```
christophermaier: It would be safer to explicitly order these how we want them, rather than implicitly rely on them being given to us in a particular order already.

I'd hate for `list::package_list` to change at some point in the future and then we suddenly start rolling people back to the earliest service on their machine and prevent them from ever updating it :grimacing: 

(Alternatively, make it painfully clear in the documentation for `list::package_list` that the contract for that function includes the order the results are returned in, along with a test that validates that behavior.)

components/sup/src/manager.rs
```diff
@@ -961,7 +961,19 @@ impl Manager {
...
 
-            self.cleanup_packages(&SUP_PKG_IDENT.parse().unwrap()).await;
+            let fully_qualified_ident =
+                PackageIdent::from_str(&format!("{}/{}", SUP_PKG_IDENT, VERSION)).unwrap();
+            let ident = SUP_PKG_IDENT.parse().unwrap();
+            // Check if we are running the most recent version of the Supervisor. If we are, cleanup
+            // the Supervisor packages. If we are not, it does not make sense to cleanup because we
+            // might uninstall the package that is running.
+            if let Some(package_install) = pkg::installed(&ident) {
+                if fully_qualified_ident.fully_qualified()
```
christophermaier: Shouldn't `fully_qualified_ident` always be fully-qualified, by definition?

Actually, `VERSION` only includes the version, not version+release, so I think this will _never_ be fully-qualified.

components/sup/src/manager.rs
```diff
@@ -961,7 +961,19 @@ impl Manager {
...
                                       pair.clone());
 
-            self.cleanup_packages(&SUP_PKG_IDENT.parse().unwrap()).await;
+            let fully_qualified_ident =
+                PackageIdent::from_str(&format!("{}/{}", SUP_PKG_IDENT, VERSION)).unwrap();
+            let ident = SUP_PKG_IDENT.parse().unwrap();
+            // Check if we are running the most recent version of the Supervisor. If we are, cleanup
+            // the Supervisor packages. If we are not, it does not make sense to cleanup because we
+            // might uninstall the package that is running.
+            if let Some(package_install) = pkg::installed(&ident) {
```
christophermaier: I think we could use an `expect` here (though it might break some tests, or developer workflows, now that I think about it :thinking: )

At the very least, documenting how this could ever be `None` would be good.

e2e_local.sh
```diff
@@ -2,15 +2,32 @@
 
 set -euo pipefail
 
-test_name=${1:?You must specify a test name}
-channel=${2:-dev}
+channel=${1:?You must specify a channel value}
+# If no `$test_name` is specified, after the test setup you will be dropped into an interactive bash
+# prompt. From there you can run `pwsh .expeditor/scripts/end_to_end/run_e2e_test_core.ps1 $test_name`
+# to quickly iterate on tests.
+test_name=${2:-}
```
christophermaier: Nice!

test/end-to-end/test_self_keep_latest_packages.ps1
```diff
@@ -0,0 +1,40 @@
...
+
+    Context "start the Supervisor with package cleanup" {
+        $supLog = New-TemporaryFile
+        Start-Supervisor -LogFile $supLog -Timeout 45 -SupArgs @( `
+                "--keep-latest-packages=2"
+        ) | Out-Null
+
+        It "removes old Supervisor packages" {
+            Wait-CommandLinesOfOutput "hab pkg list $pkg" 2
+            hab pkg list $pkg | Select-Object -Index 0 | Should -BeLike "$pkg/1.5.60*"
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [0a1327](https://github.com/habitat-sh/habitat/pull/7582/commits/0a1327528e1f769f85ffd459e07ecbd57aa55c8e)

components/sup/src/manager.rs
```diff
@@ -943,6 +961,20 @@ impl Manager {
...
 
+            let fully_qualified_ident =
+                PackageIdent::from_str(&format!("{}/{}", SUP_PKG_IDENT, VERSION)).unwrap();
+            let ident = SUP_PKG_IDENT.parse().unwrap();
+            // Check if we are running the most recent version of the Supervisor. If we are, cleanup
+            // the Supervisor packages. If we are not, it does not make sense to cleanup because we
+            // might uninstall the package that is running.
+            if let Some(package_install) = pkg::installed(&ident) {
+                if fully_qualified_ident.fully_qualified()
+                   && fully_qualified_ident == package_install.ident
```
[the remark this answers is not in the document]
davidMcneil: Yeah, that logic was pretty confusing. We want to make sure we are running the most recent local version. It should be clearer in [bb9151](https://github.com/habitat-sh/habitat/pull/7582/commits/bb91519cbbf3e1d4b5a73e2ec01242cc21333e0d).

components/sup/src/manager.rs
davidMcneil: Right, it would be a rare case, but we could be running a `cargo` built supervisor and have no actual `core/hab-sup` package installed. Addressed in [bb9151](https://github.com/habitat-sh/habitat/pull/7582/commits/bb91519cbbf3e1d4b5a73e2ec01242cc21333e0d)

components/sup/src/manager.rs
davidMcneil: Yeah, that was a bad name. Addressed in [bb9151](https://github.com/habitat-sh/habitat/pull/7582/commits/bb91519cbbf3e1d4b5a73e2ec01242cc21333e0d).

Depending on if the supervisor was built using `hab pkg build` or `cargo build` `VERSION` may or may not produce a fully qualified ident. See [here](https://github.com/habitat-sh/habitat/blob/20243c481238afa46851931ffc8029502a7969b2/components/libbuild.rs#L13). There is probably some cleanup that could happen with how we populate that `VERSION` file.

components/sup/src/util/pkg.rs
davidMcneil: Explicitly sorting is much safer, thanks. Addressed in [a0ce28](https://github.com/habitat-sh/habitat/pull/7582/commits/a0ce289bcaa98503aaba027293e72427e64ef570)

components/hab/src/command/pkg/uninstall_impl.rs
davidMcneil: That is pretty comical. I think it went through a few too many refactorings :sweat_smile: Addressed in [22b87b](https://github.com/habitat-sh/habitat/pull/7582/commits/22b87b3417cbd7943bea168a868cc1e2afa1c91b)

components/hab/src/cli/hab/sup.rs
davidMcneil: Yeah, that was a mess. That is some `structopt` [type magic](https://docs.rs/structopt/0.3.12/structopt/index.html#type-magic) to try and indicate that an optional argument can optionally take a value. In this case, I wanted`--keep-latest-packages` to default to `1` if no explicit value was specified.

There are a few issues that keep this from working well ([here](https://github.com/TeXitoi/structopt/issues/366) and [here](https://github.com/TeXitoi/structopt/issues/368)). In [1036bc](https://github.com/habitat-sh/habitat/pull/7582/commits/1036bc25017ea5df6a92a7bd6e8537e4ce175b03) I switched to using a normal `Option` which requires the user explicitly set `--keep-latest-packages=1` instead of just `--keep-latest-packages`.

components/hab/src/cli/hab/sup.rs
davidMcneil: This was a hack. Fixing the issue discussed [here](https://github.com/habitat-sh/habitat/pull/7582#discussion_r401095494). Addressed in [1036bc](https://github.com/habitat-sh/habitat/pull/7582/commits/1036bc25017ea5df6a92a7bd6e8537e4ce175b03).

components/hab/src/cli.rs
```diff
@@ -1277,6 +1277,30 @@ fn sub_sup_run(_feature_flags: FeatureFlag) -> App<'static, 'static> {
                              `127.0.0.1`")
     );
 
+    // The clap_app macro does not allow numeric default values
+    let sub =
+        sub.arg(Arg::with_name("KEEP_LATEST_PACKAGES").long("keep-latest-packages")
+                                                      .takes_value(true)
+                                                      .min_values(0)
+                                                      .max_values(2)
```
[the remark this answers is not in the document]
davidMcneil: This was a hack. Fixing the issue discussed [here](https://github.com/habitat-sh/habitat/pull/7582#discussion_r401095494). Addressed in [1036bc](https://github.com/habitat-sh/habitat/pull/7582/commits/1036bc25017ea5df6a92a7bd6e8537e4ce175b03).

components/sup/src/manager.rs
```diff
@@ -142,6 +142,9 @@ lazy_static! {
     static ref CPU_TIME: IntGauge = register_int_gauge!("hab_sup_cpu_time_nanoseconds",
                                                         "CPU time of the supervisor process in \
                                                          nanoseconds").unwrap();
+    /// Depending on the value of `VERSION` this ident may or may not be fully qualified.
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [9bb242](https://github.com/habitat-sh/habitat/pull/7582/commits/9bb2429077ca713ede1551f1d9b6831e94e5dc62)

components/sup/src/manager.rs
```diff
@@ -961,16 +963,11 @@ impl Manager {
                                       self.feature_flags,
                                       pair.clone());
 
-            let fully_qualified_ident =
-                PackageIdent::from_str(&format!("{}/{}", SUP_PKG_IDENT, VERSION)).unwrap();
+            // Only cleanup supervisor packages if we are running the latest installed version. It
+            // is possible to have no versions installed if a development build is being run.
             let ident = SUP_PKG_IDENT.parse().unwrap();
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [9bb242](https://github.com/habitat-sh/habitat/pull/7582/commits/9bb2429077ca713ede1551f1d9b6831e94e5dc62)

components/sup/src/manager.rs
```diff
@@ -961,16 +963,11 @@ impl Manager {
...
-            // Check if we are running the most recent version of the Supervisor. If we are, cleanup
-            // the Supervisor packages. If we are not, it does not make sense to cleanup because we
-            // might uninstall the package that is running.
-            if let Some(package_install) = pkg::installed(&ident) {
-                if fully_qualified_ident.fully_qualified()
-                   && fully_qualified_ident == package_install.ident
-                {
+            if let Some(latest) = pkg::installed(&ident) {
+                if *VERSIONED_IDENT == latest.ident {
                     self.cleanup_packages(&ident).await;
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [9bb242](https://github.com/habitat-sh/habitat/pull/7582/commits/9bb2429077ca713ede1551f1d9b6831e94e5dc62)

components/hab/src/cli/hab/sup.rs
```diff
@@ -341,9 +341,10 @@ pub struct SupRun {
     /// Automatically cleanup old packages.
     ///
     /// If enabled, service startup will initiate an uninstall of all packages except for the
-    /// `KEEP_LATEST_PACKAGES` most recent packages. The same logic applies to the `core/hab-sup`
-    /// package on Supervisor startup.
-    #[structopt(name = "KEEP_LATEST_PACKAGES",
+    /// `NUM_LATEST_PACKAGES_TO_KEEP` most recent packages. This also applies when a service is
+    /// restarted due to an update. The same logic applies to the `core/hab-sup` package on
+    /// Supervisor startup. If this argument is not set no package cleanup is performed.
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [9bb242](https://github.com/habitat-sh/habitat/pull/7582/commits/9bb2429077ca713ede1551f1d9b6831e94e5dc62)

components/sup/src/manager.rs
```diff
@@ -253,23 +256,24 @@ impl FsCfg {
...
+    pub http_disable:         bool,
+    pub gossip_peers:         Vec<SocketAddr>,
+    pub gossip_permanent:     bool,
+    pub ring_key:             Option<SymKey>,
+    pub organization:         Option<String>,
+    pub watch_peer_file:      Option<String>,
+    pub tls_config:           Option<TLSConfig>,
+    pub feature_flags:        FeatureFlag,
+    pub event_stream_config:  Option<EventStreamConfig>,
+    pub keep_latest_packages: Option<usize>,
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [9bb242](https://github.com/habitat-sh/habitat/pull/7582/commits/9bb2429077ca713ede1551f1d9b6831e94e5dc62)

PR #7585, opened 2020-03-27
Error if using `-d` when entering Windows studio

### LINKED ISSUE #7578: Create alias for hab studio enter -D when users use hab studio enter -d
This is a quality of life request to make the Hab cli friendlier.

If a user enters `hab studio enter -D` a Docker studio is created, if they mistakenly type `hab studio enter -d` the flag misses and a non-docker studio is created. 

I purpose we create an alias so both the lowercase/capital flags `-d` and `-D` create a Docker studio. 

We could also provide users an error message for `-d` rather than enter a non-docker studio.

davidMcneil:
Resolves #7578

This PR adds the ability to enter a docker studio with `-d` on windows. I would have rather thrown an error, but I could not come up with a reasonable workaround for the lack of case-sensitive variables in powershell (@mwrock do you have any suggestions?). This PR also adds logic causing the studio to throw an error if an unknown argument is passed in.

### DIFF
--- a/components/studio/bin/hab-studio.ps1
+++ b/components/studio/bin/hab-studio.ps1
@@ -17,7 +17,6 @@ param (
     [switch]$q,
     [switch]$v,
     [switch]$R,
-    [switch]$D,
     [string]$command,
     [string]$commandVal,
     [string]$k,
@@ -568,17 +567,22 @@ if(!(Test-InContainer)) {
 }
 
 try {
-    switch ($command) {
-        "new" { New-Studio }
-        "run" { Invoke-StudioRun $commandVal }
-        "rm" { Remove-Studio }
-        "enter" { Enter-Studio }
-        "build" { Invoke-StudioBuild $commandVal $R }
-        "version" { Write-Host "$program $version" }
-        "help" { Write-Help }
-        default {
-            Write-Help
-            Write-Error "Invalid Argument $command"
+    if ($args.Count -gt 0) {
+        Write-Help
+        Write-Error "Invalid Argument '$args'"
+    } else {
+        switch ($command) {
+            "new" { New-Studio }
+            "run" { Invoke-StudioRun $commandVal }
+            "rm" { Remove-Studio }
+            "enter" { Enter-Studio }
+            "build" { Invoke-StudioBuild $commandVal $R }
+            "version" { Write-Host "$program $version" }
+            "help" { Write-Help }
+            default {
+                Write-Help
+                Write-Error "Invalid Command '$command'"
+            }
         }
     }
 } finally { $VerbosePreference = $currentVerbose }


components/studio/bin/hab-studio.ps1
```diff
@@ -22,7 +22,11 @@ param (
     [string]$commandVal,
     [string]$k,
     [string]$o,
-    [string]$s
+    [string]$s,
+
+    [Parameter(ValueFromRemainingArguments = $true, DontShow = $true)]
+    [Object[]]
+    $extraParameters = @()
```
mwrock: you should not need this. See note below.

components/studio/bin/hab-studio.ps1
```diff
@@ -568,17 +572,22 @@ if(!(Test-InContainer)) {
...
-        "run" { Invoke-StudioRun $commandVal }
-        "rm" { Remove-Studio }
-        "enter" { Enter-Studio }
-        "build" { Invoke-StudioBuild $commandVal $R }
-        "version" { Write-Host "$program $version" }
-        "help" { Write-Help }
-        default {
-            Write-Help
-            Write-Error "Invalid Argument $command"
+    if ($extraParameters.Count -gt 0) {
```
mwrock: this can instead be `if ($args.count -gt 0) {`

In powershell you have 2 automatic variables for arguments `$args` and `$PSBoundPaeameters`. `$args` is the array of "unbound" arguments. So in this case `-d` would qualify.

components/hab/src/command/studio/enter.rs
```diff
@@ -300,7 +300,7 @@ mod inner {
 
         for arg in args.iter() {
             let str_arg = arg.to_string_lossy();
-            if str_arg == "-D" {
+            if str_arg == "-D" || str_arg == "-d" {
```
mwrock: If we leave this as it was, the `-d` should pass into the native studio which will then be treated as an error with the changes below.

davidMcneil: I initially thought this as well. Unfortunately, because powershell variables are case insensitive `-d` does not show up in `$args` instead it sets the `$D` variable. Or am I missing something?

mwrock: I think we can actually ommit `$D` since once you get to the powershell we don't do anything with it. We just need to have the rust strip it off.

components/studio/bin/hab-studio.ps1
davidMcneil: This is much better thanks. Addressed in [27e558](https://github.com/habitat-sh/habitat/pull/7585/commits/27e558c4c733781eefa76f70a428cf0d92d493fe)

components/hab/src/command/studio/enter.rs
davidMcneil: This is much better thanks. Addressed in [27e558](https://github.com/habitat-sh/habitat/pull/7585/commits/27e558c4c733781eefa76f70a428cf0d92d493fe)

PR #7588, opened 2020-03-30
Prepare e2e tests for powershell 7.0

mwrock:
This makes a few tweaks to the e2e tests in preparation for powershell 7. It is mainly extending the default timeout for waiting for supervisor services to start and removing the overrides. While I was testing, locally the current timeout were not a problem but I was having lots of buildkite timeout issues that this cleared up.

This also allows for some slightly different behavior of `Get-Process` on Linux with powershell 7. Powershell 7 will now behave similar to the native `ps` command and include not just the binary name in the process name but any additional args that is also rendered by `ps`. For example `redis-server` would be the redis process name in powershell versions prior to 7.0. However in version 7 it may be `redis-server *:6379 `.

### DIFF
--- a/.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
+++ b/.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
@@ -23,8 +23,9 @@ if ($IsLinux -Or $IsMacOS) {
 
 # It may take a while to download a given service package, and
 # services may have long init scripts. Let's be generous in how long
-# we're willing to wait.
-$DefaultServiceTimeout = 40
+# we're willing to wait. Also note that tests may take considerably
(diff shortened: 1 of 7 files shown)


.expeditor/scripts/end_to_end/run_e2e_test_core.ps1
```diff
@@ -24,7 +24,7 @@ if ($IsLinux -Or $IsMacOS) {
 # It may take a while to download a given service package, and
 # services may have long init scripts. Let's be generous in how long
 # we're willing to wait.
-$DefaultServiceTimeout = 40
+$DefaultServiceTimeout = 90
```
christophermaier, line 27: Might be worth mentioning that the size of this timeout takes into account apparent Buildkite slowness, too.

PR #7595, opened 2020-04-02
Add `--keep-latest` to `hab pkg uninstall`

### LINKED ISSUE #7591: Add `--keep-latest` to `hab pkg uninstall`
Continue on the work started in #7582 and add a `--keep-latest` flag to `hab pkg uninstall`.

davidMcneil:
Resolves #7591 

Add the `--keep-latest` flag to `hab pkg uninstall`

### DIFF
--- a/components/sup/src/util/pkg.rs
+++ b/components/sup/src/util/pkg.rs
@@ -3,7 +3,8 @@ use crate::{error::{Error,
             PRODUCT,
             VERSION};
 use hab::{command::pkg::{self,
-                         uninstall_impl},
+                         uninstall_impl::{self,
+                                          UninstallSafety}},
           error::Result as HabResult};
 use habitat_api_client::BuilderAPIClient;
 use habitat_common::{cli::FS_ROOT,
@@ -135,7 +136,7 @@ pub async fn uninstall_all_but_latest(ident: impl AsRef<PackageIdent>,
                                              pkg::ExecutionStrategy::Run,
                                              pkg::Scope::PackageAndDependencies,
                                              &[],
-                                             false).await
+                                             UninstallSafety::Safe).await
 }
 
 /// Uninstall multiple packages.
@@ -146,7 +147,7 @@ pub async fn uninstall_many(idents: &[impl AsRef<PackageIdent>]) -> HabResult<()
                                    pkg::ExecutionStrategy::Run,
                                    pkg::Scope::PackageAndDependencies,
                                    &[],
-                                   false).await
+                                   UninstallSafety::Safe).await
 }
 
 /// Uninstall a package given a package identifier.
@@ -161,5 +162,5 @@ pub async fn uninstall_even_if_loaded(ident: impl AsRef<PackageIdent>) -> HabRes
                               pkg::ExecutionStrategy::Run,
                               pkg::Scope::PackageAndDependencies,
                               &[],
-                              true).await
+                              UninstallSafety::Force).await
 }
--- a/components/sup/src/util/pkg.rs
+++ b/components/sup/src/util/pkg.rs
@@ -3,7 +3,6 @@ use crate::{error::{Error,
             PRODUCT,
             VERSION};
 use hab::{command::pkg::{self,
-                         list,
                          uninstall_impl},
           error::Result as HabResult};
 use habitat_api_client::BuilderAPIClient;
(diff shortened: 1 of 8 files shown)


test/end-to-end/test_pkg_uninstall.ps1
```diff
@@ -0,0 +1,57 @@
...
+    It "uninstall all but the three latest" {
+        hab pkg uninstall --keep-latest=3 "$pkg"
+        hab pkg list "$pkg" | Should -BeExactly $pkgs[6..8]
+    }
+
+    It "uninstall does nothing if keeping all" {
+        hab pkg uninstall --keep-latest=10 "$pkg"
+        hab pkg list "$pkg" | Should -BeExactly $pkgs[6..8]
+    }
+
```
christophermaier, line 48: An interesting variant of this test case to add would be something like

``` sh
hab pkg uninstall --keep-latest=2 core/redis/3.2.3/20161102201135
```
That is, providing a fully-qualified identifier along with `keep-latest` should never delete anything, since there's only one "latest" incarnation of a fully-qualified package, by definition.

I was debating whether that should be an error, since that combination of inputs is inherently contradictory. If read as "keep at most 2 releases of this thing", however, it still works, because that post-condition still holds. Thus, it's not an error condition.

test/end-to-end/test_pkg_uninstall.ps1
```diff
@@ -0,0 +1,57 @@
...
+    It "uninstall all" {
+        hab pkg uninstall --keep-latest=0 "$pkg"
```
christophermaier, line 50: I thought `keep-latest` was meant to be a non-zero integer?

I'm also somewhat surprised we don't have `hab pkg uninstall --all $PKG`.

Perhaps somewhat at odds with my argumentation above, I wonder if we _should_ treat `--keep-latest=0` as an error, and instead implement `--all` as an explicit indication of "yes, I really mean to get rid of _everything_". Perhaps I'm overthinking, but I wonder about scripted use of this where the number given for `--keep-latest` is a variable that, due to some script bug, ends up accidentally becoming 0. In that situation, you just accidentally blew away all your installed instances of that package.

I'd love some of your thoughts on this UX, @apriofrost :smile:

test/end-to-end/test_pkg_uninstall.ps1
davidMcneil: Agreed, this would be a good test to add. An interesting UX question that arises is what happens if `--keep-latest=0`? If we allow that case it should probably uninstall the package.

test/end-to-end/test_pkg_uninstall.ps1
mwrock: I'm assuming `--all` or `--keep-latest=0` is the default here if you do not provide a `--keep-latest`. Personally I am fine with allowing 0. I don't find it confusing or surprising. Just my 2 cents.

davidMcneil: No, it is not currently the default. If a partial ident is specified with no `--keep-latest` flag only the newest package is removed. Truthfully, I find this behavior a little surprising. If it was only going to remove one package I would have thought it would be the oldest.

I could see that it would make sense for `--keep-latest` to default to `0` and then rely on the user to use full or partial idents to specify exactly what they want to remove. For example, if someone ran `hab pkg uninstall core/redis`, I don't think it would be all that crazy to assume they want to remove all `core/redis` packages. But this would be a breaking change and a pretty dangerous one at that.

davidMcneil: @krasnow I am curious what your thoughts are here?

krasnow: `--keep-latest=0` seems like a good option to have. It follows that if I wanted to delete all packages, this would be a valid way to do so.

As David says above, there's a separate usability problem with the default behavior of `hab pkg unstall` in that I would expect that to delete all versions of a package if I don't specify an ident. We should split that out as a separate issue, but fixing that would make the `keep-latest` option more intuitive.

components/sup/src/util/pkg.rs
```diff
@@ -126,28 +125,17 @@ pub async fn install_channel_head(url: &str,
...
-                                   false).await?;
-    Ok(to_uninstall.len())
+    uninstall_impl::uninstall_all_but_latest(&mut NullUi::new(),
+                                             ident,
+                                             number_latest_to_keep,
+                                             &*FS_ROOT,
+                                             pkg::ExecutionStrategy::Run,
+                                             pkg::Scope::PackageAndDependencies,
+                                             &[],
+                                             false).await
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [5c8830](https://github.com/habitat-sh/habitat/pull/7595/commits/5c883050b9494065ac903e296520b02e8bc3cc5b)

test/end-to-end/test_pkg_uninstall.ps1
davidMcneil: Addressed in [a31fc9](https://github.com/habitat-sh/habitat/pull/7595/commits/a31fc99bcb4820737ff6cf98750b66b3145ae7d0)

test/end-to-end/test_pkg_uninstall.ps1
davidMcneil: #7604 is the separate issue for addressing `hab pkg uninstall` UX.

Issue #7412, opened 2020-02-03
Use daemon-less container creation in our Docker exporter


christophermaier:
Our current Docker exporter uses Docker to create images (:tada: _surprise!_ :tada:). This requires a running Docker daemon, which introduces a host of security concerns when trying to build within a container. If we could pivot to using one of the new crop of daemon-less container creation tools in our exporter, we could eliminate these security concerns, allowing things like `habitat/builder-worker` to run in unprivileged containers. 

Some possible tools include [img](https://github.com/genuinetools/img) and [buildah](https://github.com/containers/buildah), among others. While it currently appears that building Windows containers with such tools ranges from extremely difficult to impossible, nothing prevents us from implementing this first for Linux container export; the benefits would still be significant.

jeremymv2:
Here's a great in-depth article covering some of the best in breed tools on this topic along with their security trade-offs: https://www.tngtech.com/fileadmin/Public/Images/BigTechday/BTD12/Folien/Slides_MartinHoefling_PatrickHarboeck_compressed.pdf

chefsalim:
Agreed, the ability to run builder-worker unprivileged could even be a/the main use case to start with.  So the unprevileged Docker export could be a means to an end in addition to being a "good thing".  Going after that use case, we'd want to look at all of the exporters.

christophermaier:
Closed with https://github.com/habitat-sh/habitat/pull/7685

PR #7596, opened 2020-04-02
Layered images from `hab pkg export docker`

### LINKED ISSUE #7411: Create exported Docker images with layers
Our current way of exporting Docker containers creates a container with effectively a single layer (i.e., all code dependencies are added in a single layer). This makes layer caching basically useless, as even a single byte change in your exported container will require you do download the entire container. 

If we could create containers that were properly layered, we could make life easier for our users by allowing them to take advantage of layers they already have, decreasing the time it takes to get new containers distributed, and taking up less network bandwidth in the process.

A possible implementation idea is to do a topological sort of the entire transitive dependency tree of the package being exported (dependencies at the same layer of the tree could be lexicographically sorted for stability), and to add the dependencies one layer at a time, adding the base dependencies first, and moving up the stack until you reach the package being exported. Theoretically, base dependencies should change less frequently than direct dependencies, which may allow for more reuse. It may not end up being perfect, but could end up being a welcome improvement over the current all-in-one-layer approach.

christophermaier:
Historically, our exported container images have essentially been a single layer, containing all the Habitat packages needed to run the Supervisor and the exported service.

While this works, it does not sufficiently leverage image layer caching. In the normal case when a user is iterating on a service, very often that single package is the only one that will be changing. Ideally, a container image for a new version of that service should be able to re-use the layers containing the Habitat Supervisor and all its dependencies, since none of them changed. Since everything is exported in a single layer, though, all of this is utterly opaque to the container storage engine.

Here, we introduce the ability to make multi-layer images. We do bit of pre-processing of the collection of packages to create a heuristically optimized ordering in which to add them to the container image. We generate a topological sort for all the "top-level" packages being added to the container (basically, `core/hab`, `core/hab-sup`, `core/hab-launcher`, and whatever the exported service is, but also `core/busybox-static` and `core/cacerts` on Linux), reverse them so that the dependencies come first, and then union all the lists together, with the service package last.

Then, armed with this list, we add each package to the image one at a time, resulting in a layer per package. While this is not necessarily guaranteed to give maximum reuse in absolutely all cases, it should yield a lot more reuse than before (i.e., none!). If you're iterating on your service package, you should basically be able to reuse everything below it in the stack; that is, your difference will be just the single layer containing that service's package.

To build such a multi-layer image, pass `--multi-layer` to `hab pkg export docker`. Leaving this option off will continue to use the "all in one" approach we have used until now (this mode is also useful when the number of packages in an image exceeds the amount of layers Docker allows in an image).

A fair bit of cleanup and simplification of the exporter code is also included in this PR; see individual commits for further details.

Fixes #7411


components/pkg-export-docker/src/graph.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// dependency order.
+    ///
+    /// Note that this is essentially a consistent union of the
+    /// reverse topological sorts of all the "top level" packages that
+    /// are added to a container image.
+    ///
+    /// User packages will be last. Ideally, as users are iterating on
+    /// their packages and creating images, this should mean that all
+    /// the dependencies are already available as cached layers.
+    pub fn reverse_topological_sort(&self) -> Vec<PackageIdent> {
```
davidMcneil, line 51: Minor - Can the result of this calculation ever change after the graph has been created? Could we do the calculation as part of `from_packages` and just store the `Vec<PackageIdent>`?

components/pkg-export-docker/defaults/Dockerfile.hbs
```diff
@@ -1,6 +1,15 @@
 FROM {{base_image}}
 ENV PATH {{path}}
-ADD {{rootfs}} /
+
+ADD {{rootfs}}/root /root
+ADD {{rootfs}}/tmp /tmp
+ADD {{rootfs}}/var /var
+ADD {{rootfs}}/etc /etc
```
davidMcneil, line 7: Just for my own learning, why do we create these directories now? Just convenient for the user?

components/pkg-export-docker/src/docker.rs
```diff
@@ -448,6 +446,7 @@ impl DockerBuildRoot {
         });
         let image_name = match naming.custom_image_name {
                              Some(ref custom) => {
+                                 // TODO (CM): why is this handlebars???
```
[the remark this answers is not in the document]
christophermaier: damn it... I thought I took that out :joy:

components/pkg-export-docker/src/docker.rs
```diff
@@ -70,10 +70,7 @@ impl<'a> DockerBuilder<'a> {
     /// * If building the Docker image fails
     pub fn build(self) -> Result<DockerImage> {
         let mut cmd = docker_cmd();
-        cmd.current_dir(self.workdir)
-           .arg("build")
-           .arg("--force-rm")
-           .arg("--no-cache");
```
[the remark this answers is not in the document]
christophermaier: I did a lot of spelunking to determine why, and it appears to have been there since the beginning, with no comment.

Based on https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache, however, I believe that we are safe in not using `--no-cache`. If we had commands like `RUN apt-get -y update` we'd need it, but all the commands we use seem to do the right thing.

Of course, I'm open to second opinions on that! :sweat_smile:

components/pkg-export-docker/src/docker.rs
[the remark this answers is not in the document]
davidMcneil: I think it is a valid question. There are a few rust [Dockerfile generators](https://crates.io/search?q=dockerfile), not sure if any would make it worth switching.

components/pkg-export-docker/src/graph.rs
```diff
@@ -0,0 +1,214 @@
...
+                // likely to be any sort of bottlneck in the creation
+                // of a Docker image.
+
+                // You have to check first before inserting;
+                // otherwise, it increments the insertion order
+                // each time, which will give us an incorrect
+                // overall ordering.
+                if !acc.contains_key(&ident) {
+                    // Treat this map like a set
+                    acc.insert(ident, true);
```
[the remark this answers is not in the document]
christophermaier: Ah, that's a great point! Whenever I've had to do this in other languages, I have always used `true`, and I just reflexively did that here, even though `()` would be the obvious Rust idiom.

Thanks for catching that!

components/pkg-export-docker/src/graph.rs
```diff
@@ -0,0 +1,214 @@
...
+        Ok(Graph { g, base, user })
+    }
+
+    /// Helper function to create a Vec of our base idents in a
+    /// sensible order (roughly in order of package volatility).
+    ///
+    /// The idea is that we'll add packages to the image in this
+    /// order, one layer at a time, in order to try and maximize
+    /// layer caching.
+    fn idents_from_base(&self) -> Vec<PackageIdent> {
```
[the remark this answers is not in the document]
christophermaier: :thinking: Maybe? Hard to say one is better than the other at this point, I think.

components/pkg-export-docker/defaults/Dockerfile.hbs
christophermaier: We start from a `scratch` image, which has absolutely nothing in it. We've always created these directories (as well as several items inside `/etc`) to ensure things work properly.

This is just breaking it out explicitly.

components/pkg-export-docker/src/graph.rs
christophermaier: No, it wouldn't change, so we could technically store it at creation time.

With the current usage, it's basically a "six of one, half-a-dozen of the other" kind of situation, I think.

components/common/src/package_graph.rs
```diff
@@ -68,7 +68,10 @@ impl PackageGraph {
     }
 
     /// Extend a graph by adding in dependencies for a package
-    fn extend(&mut self, package: &PackageIdent, deps: &[PackageIdent]) -> (usize, usize) {
+    // NOTE: I suppose in an ideal world, `deps` would be something
+    // like a slice of &PackageIdent to cut down on needless cloning,
+    // but it's fine for now.
```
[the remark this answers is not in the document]
christophermaier: Hrm... indeed. That might be a good way to go.

christophermaier:
> It looks like any new exporter we add (#7412) could easily leverage all of this work. Do you think that is accurate @christophermaier?

Yup, that's the intention.

After some further testing, however, I find I'm running into an unintended change to file ownership modes in this new code that breaks the ability to run the containers as a non-root user (see https://www.habitat.sh/docs/best-practices/running-habitat-linux-containers/ for background on this).

I'm researching this further today; it may be that #7412 provides a solution.

components/pkg-export-docker/defaults/Dockerfile.hbs
```diff
@@ -1,6 +1,15 @@
...
+{{~ #each packages as |pkg|}}
+ADD {{../rootfs}}/hab/pkgs/{{pkg}} /hab/pkgs/{{pkg}}
```
jsirex, line 9: I see each package is going to its own layer, but I'm not sure is it good idea from performance point of view to have such many layers? 
From docker documentation:
> The overlay2 driver natively supports up to 128 lower OverlayFS layers

Is it possible to package have such number of TDEPS? Probably yes

Issue: https://github.com/moby/moby/issues/26380

jsirex: UPD2: After googling and thinking looks like nothing can be done here, probable just show a warning when there are more than 100 packages in TDEPS

christophermaier: @jsirex Agreed. I'll have an updated version up soon that will let you choose whether to use a package-per-layer, or use the existing everything-in-one-layer behavior, which remains the default. That way, if you want reuse, you can have it, but if that results in too many layers, there's still a way to build a container.

test/end-to-end/test_docker_exporter.ps1
```diff
@@ -0,0 +1,135 @@
...
+        [Parameter(Mandatory=$false)][string]$extra_args
+    )
+    $name="e2e-docker-export-container"
+    Write-Host (docker run -d --name=$name --rm --env=HAB_LICENSE=accept-no-persist $extra_args $image | Out-String)
+    "$name"
+}
+
+# If we can set and get a value from Redis running in the container,
+# then we know we created a container that can actually run.
+function Confirm-ContainerBehavior() {
```
mwrock, line 40: I think if we tested with nginx instead of redis and then changed this function to an http check of the default home page, we would have a test that would work on both linux and windows. We would just need to run the non-root describes inside of an `if($IsLinux)`

christophermaier: Agreed... with luck I'll get around to this today.

Issue #6410, opened 2019-04-11
supervisor not considering FS_ROOT when looking for pwsh.exe


mwrock:
When running in a local studio, the supervisoe looks for `pwsh.exe` in `/hab/pkgs/core/powershell` and not prepended with the `FS_ROOT`.

mwrock:
Just did some digging and it looks like `interpreter_paths()` has never rooted the path of the interpreter. Ideally it should do something similar to `environment_for_command(&self)` that roots each path declared in a packages runtime.

This has just never surfaced because only a local windows studio is not rooted in `c:\` and prior to 0.79 the studio's vendored powershell was always on the path. which led to the fix in 0.79.1.

PR #7601, opened 2020-04-06
cleanup adding interpreter paths and make it configurable

mwrock:
This PR accomplishes four things:

1. Fixes a bug where the supervisor will use the most recent version of powershell or busybox-static installed instead of the one packaged. This is fixed with [this line](https://github.com/habitat-sh/habitat/pull/7601/files#diff-9e4f1f1544b6c3ae7b8a78d289f54773R68). My guess is that we used to store the `hab-sup` binary in the root of the package and not in `bin`.
1. Fixes [another bug](https://github.com/habitat-sh/habitat/issues/6410) where the interpreter is not propperly rooted in a local windows studio.
1. No longer allow an interpreter binary that is found on the path but not in a package to be used. I do not think this is a breaking change and do not know why we would allow this.
1. Allows a user to configure the interpreter ident via a `HAB_INTERPRETER_IDENT` env variable or config setting.

### DIFF
--- a/components/common/src/util/path.rs
+++ b/components/common/src/util/path.rs
@@ -94,24 +94,24 @@ async fn interpreter_paths() -> Result<Vec<PathBuf>> {
                 // Nope, no packages of the interpreter installed. Now we're going to see if the
                 // interpreter command is present on `PATH`.
                 Err(_) => {
-                    if install::type_erased_start(&mut ui::NullUi::new(),
-                                                &default_bldr_url(),
-                                                &ChannelIdent::stable(),
-                                                &(ident.clone(),
-                                                    PackageTarget::active_target())
-                                                                                .into(),
-                                                &*PROGRAM_NAME,
-                                                VERSION,
-                                                FS_ROOT_PATH.as_path(),
-                                                &cache_artifact_path(None::<String>),
-                                                None,
-                                                &InstallMode::default(),
-                                                &LocalPackageUsage::default(),
-                                                InstallHookMode::default()).await.is_err() {
-                                                    return Err(Error::InterpreterNotFound(ident));
-                                                }
-                    let pkg_install = PackageInstall::load(&ident, Some(FS_ROOT_PATH.as_ref()))?;
-                    pkg_install.paths()?
+                    match install::type_erased_start(&mut ui::NullUi::new(),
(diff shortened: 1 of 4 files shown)


components/common/src/util/path.rs
```diff
@@ -92,58 +87,47 @@ async fn interpreter_paths() -> Result<Vec<PathBuf>> {
...
+                                                VERSION,
+                                                FS_ROOT_PATH.as_path(),
+                                                &cache_artifact_path(None::<String>),
+                                                None,
+                                                &InstallMode::default(),
+                                                &LocalPackageUsage::default(),
+                                                InstallHookMode::default()).await.is_err() {
+                                                    return Err(Error::InterpreterNotFound(ident));
+                                                }
+                    let pkg_install = PackageInstall::load(&ident, Some(FS_ROOT_PATH.as_ref()))?;
```
davidMcneil, line 113: Can we use the `PackageInstall` returned from `install::typed_erased_start` or is it important to try and `load` the package?

components/common/src/util/path.rs
```diff
@@ -92,58 +87,47 @@ async fn interpreter_paths() -> Result<Vec<PathBuf>> {
...
+                                                                                .into(),
+                                                &*PROGRAM_NAME,
+                                                VERSION,
+                                                FS_ROOT_PATH.as_path(),
+                                                &cache_artifact_path(None::<String>),
+                                                None,
+                                                &InstallMode::default(),
+                                                &LocalPackageUsage::default(),
+                                                InstallHookMode::default()).await.is_err() {
+                                                    return Err(Error::InterpreterNotFound(ident));
```
davidMcneil, line 111: We end up losing the underlying reason _why_ the interpreter could not be found. Maybe this is not a big deal. I have never seen it done before but I wonder if we should do something like `InterpreterNotFound(PackageIdent, Box<Self>)`?

mwrock: good call!

components/common/src/util/path.rs
mwrock: No that is a great point and I just did not notice that the previous code was not taking advantage of that

mwrock:
I implemented both of your suggestions @davidMcneil

PR #7615, opened 2020-04-11
append existing path to package path in hab pkg exec

### LINKED ISSUE #7135: Existing PATH handled differently by supervisor and `pkg exec`
**OS:** The Linuxes

# The Problem

I cannot `hab pkg exec` a command that shells out to a utility provided by the host OS.

"Of course not, Robb," you might say. "A Habitat package can only run commands from within itself or its dependencies."

Except, I _can_ run that command in a supervised process.

* Supervised processes get [a PATH that includes the PATH from the Supervisor's start up environment](https://github.com/habitat-sh/habitat/blob/c20916cdb2694f58fb7aa26be29d347404e5bc1b/components/sup/src/command/shell.rs#L31).
* Commands run with `hab pkg exec` do not.

This has made testing difficult for a package of command line tools that expect to be able to call commands provided by the OS because the patterns for testing commands in a hab package are to `hab pkg exec` them and verify their exit code and output.

## Some Related Issues

I have searched through the issues for discussions about `hab pkg exec` and the symlinks/binstubs generated by `binlink`. These two seem the closest, but are focued on binlinks.

* #6657 - Windows binlinks are PowerShell wrapper scripts for a package's executables that include the Windows system paths. (I [commented on that issue](https://github.com/habitat-sh/habitat/issues/6657#issuecomment-533168366) when I first noticed _this_ issue.)
* #6812 - A design proposal for changing generated binlinks on not-Windows and how that may be ~fraught with peril~ tricky.

Changing the behavior of binlinks appears to be complicated, particularly hard if opt-in/out for the behavior is required.

# Idea №1

🤔💭 { `hab pkg exec --preserve-path pumpkin/pie eat_it_up` }

Add an option (of some name to-be-determined) to `pkg exec` to do what the supervisor does with including the existing `PATH`.

While this would violate the purity of "Habitat provides all your dependencies," the supervisor appears to already do so. This would allow users to have `pkg exec` behavior that is consistent with the supervisor while also making it an opt-in behavior.

# Idea №2

Similar to №1, but modeled after `sudo`'s options (thanks, @jsirex):

```
  -E, --preserve-env            preserve user environment when running command
      --preserve-env=list       preserve specific environment variables
```

✋ **I'd be satisfied with any option that allowed me to include the calling context's `PATH`.** Similarity with `sudo`'s options seems interesting, but I totally acknowledge that would over-reach the goal of `PATH` behavior consistent with the supervisor.

# Idea №3

Update the supervisor to _not_ include the "system" `PATH` in the supervised process' runtime environment. If an application requires commands from outside the directories provided by itself and its dependencies, the plan authors must write scripts to wrap their executables to include other directories. Pseudo-codey example (again, thanks @jsirex):

```
# some where in $pkg_path/bin/wrapper-executable

PKG_PATH="$PATH"

# load host system's profile
unset PATH
source /etc/profile

export PATH=$PKG_PATH:$PATH

what_i_really_want_to_do
```

mwrock:
fixes #7135

This appends the existing PATH to the path used by `hab pkg exec` to achieve PATH parity with the supervisor. The only key difference between this and the supervisor PATH for a service is that the supervisor also appends the interpreter path (powershell or busybox) because those are always needed to run the hook script which does not apply to `exec`.

### DIFF
--- a/components/hab/src/command/pkg/exec.rs
+++ b/components/hab/src/command/pkg/exec.rs
@@ -1,5 +1,6 @@
 use std::{env,
           ffi::OsString,
+          io,
           path::PathBuf};
 
 use crate::hcore::{fs::{find_command,
@@ -11,12 +12,29 @@ use crate::hcore::{fs::{find_command,
 use crate::error::{Error,
                    Result};
 
+const PATH_KEY: &str = "PATH";
+
 pub fn start<T>(ident: &PackageIdent, command: T, args: &[OsString]) -> Result<()>
     where T: Into<PathBuf>
 {
     let command = command.into();
     let pkg_install = PackageInstall::load(&ident, Some(&*FS_ROOT_PATH))?;
-    let cmd_env = pkg_install.environment_for_command()?;
+    let mut cmd_env = pkg_install.environment_for_command()?;
+
+    if let Some(path) = cmd_env.get(PATH_KEY) {
+        if let Some(val) = env::var_os(PATH_KEY) {
+            let mut paths: Vec<PathBuf> = env::split_paths(&path).collect();
+            let mut os_paths = env::split_paths(&val).collect();
+            paths.append(&mut os_paths);
+            let joined = env::join_paths(paths)?;
+            let path_str =
+                joined.into_string()
+                      .map_err(|s| {
+                          io::Error::new(io::ErrorKind::InvalidData, s.to_string_lossy())
+                      })?;
+            cmd_env.insert(PATH_KEY.to_string(), path_str);
+        }
+    }
 
     for (key, value) in cmd_env.into_iter() {
         debug!("Setting: {}='{}'", key, value);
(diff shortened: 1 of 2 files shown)


components/hab/src/command/pkg/exec.rs
```diff
@@ -11,12 +11,26 @@ use crate::hcore::{fs::{find_command,
...
+    let mut cmd_env = pkg_install.environment_for_command()?;
+
+    if let Some(path) = cmd_env.get(PATH_KEY) {
+        if let Some(val) = env::var_os(PATH_KEY) {
+            let mut paths: Vec<PathBuf> = env::split_paths(&path).collect();
+            let mut os_paths = env::split_paths(&val).collect();
+            paths.append(&mut os_paths);
+            let joined = env::join_paths(paths)?;
+            let path_str = joined.into_string()
+                                 .expect("Unable to convert OsStr path to string!");
```
davidMcneil, line 30: Could we do something like `.map_err(|s| io::Error::new(io::ErrorKind::InvalidData, s.to_string_lossy()))?;` instead of `expect`?

PR #7620, opened 2020-04-17
core: move from users to nix crate

### LINKED ISSUE #7298: hab-sup erroneously reports missing group, fails to load service
On hab 0.90.6 (and possibly older), hab-sup can fail to start a service with the following error:

```
hab-sup(MR): Starting chef/deployment-service (chef/deployment-service/0.1.0/20191211163131)
hab-sup(MR): Can't create directory /hab/svc/deployment-service: Package requires group hab to exist, but it doesn't
hab-sup(MR): If this service is running as non-root, you'll need to create /hab/svc/deployment-service and give the curre...ccess to it
```

even when the `hab` group exists. This is because neither our code or the underlying library we use to get the group by name surfaces errors from `getgrnam_r`.

Our code always reports this error whenever the library returns None:

https://github.com/habitat-sh/habitat/blob/master/components/core/src/os/users/linux.rs#L77-L89

but the underlying library returns None in the case of an error:

https://github.com/ogham/rust-users/blob/master/src/base.rs#L484-L488

We've recently seen such an error occur because the user had an /etc/groups entry that was too large for the hard-coded buffer value of 2048 bytes. (I can provide a reproduction if you'd like to see the problem locally. Note that because of the way glibc's `getgrname` works, _any_ entry that precedes the entry you are looking for that requires a larger buffer will cause a failure.):

https://github.com/ogham/rust-users/blob/master/src/base.rs#L474

In our own code, we cannot simply check last_os_error() because the library does not give us access to the error code from the underlying system call. Since `errno` is not reset to 0 on success so we have no way to distinguish an error in this system call from an error in a previous system call on this thread.

While we could fix part of this upstream, completely fixing the problems with this call would require a breaking change to their library.  Namely, this api should have a return value of
`Result<Option<Group>>` rather than `Option<Group>`.

Alternatively, I think we should consider making the `nix` crate our default system-interface dependency for linux in any case where we don't want to call libc directly.  For instances, the nix version of this function already has the correct return value:

https://github.com/nix-rust/nix/blob/5cb526d55df223735602d2f50cae4c9b43120d9c/src/unistd.rs#L2724

and already handles buffer resizing:

https://github.com/nix-rust/nix/blob/5cb526d55df223735602d2f50cae4c9b43120d9c/src/unistd.rs#L2564-L2566

stevendanna:
The nix crate is a bit more robust than the users create with respect
to its error handling. This moves us directly to the nix create
without changing any of the APIs in core::users.

This likely resolves #7298 for many users as nix correctly handles
ERANGE errors from getgrnam_r. However, the nix still enforces a
maximum buffer size of 16KB, so there is still a chance for a failure
caused by a large group. We could avoid this by calling libc directly
or working with upstream to make the buffer size configurable.

Review Notes:

- We should follow up with a change to change the API in the users module from returning `Option<T>`s to `Result<Option<T>`.  I started on this and the changes are pretty straightforward, but I need to look at each call site to decide whether it makes sense to ignore the errors or not.

- I still don't love the 16KB limit.  I'm thinking the best bet might be to work with upstream to make this configurable.

### DIFF
--- a/components/core/Cargo.toml
+++ b/components/core/Cargo.toml
@@ -36,7 +36,8 @@ typemap = "*"
 url = "*"
 
 [target.'cfg(not(windows))'.dependencies]
-users = "*"
+# Pulling from git until https://github.com/nix-rust/nix/commit/3c2107bdc221a90b02d3f7118dd96f3496762cd0 is released
+nix = { git = "https://github.com/nix-rust/nix.git" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 caps = "*"
--- a/components/core/Cargo.toml
+++ b/components/core/Cargo.toml
@@ -36,9 +36,10 @@ typemap = "*"
 url = "*"
 
 [target.'cfg(not(windows))'.dependencies]
-# Pulling from git until https://github.com/nix-rust/nix/commit/3c2107bdc221a90b02d3f7118dd96f3496762cd0 is released
-nix = { git = "https://github.com/nix-rust/nix.git" }
-
+# Pulling from git until
+# https://github.com/nix-rust/nix/commit/3c2107bdc221a90b02d3f7118dd96f3496762cd0
+# is released, which will be the next release after v0.17.0
+nix = {  git = "https://github.com/nix-rust/nix.git", rev = "3c2107bdc221a90b02d3f7118dd96f3496762cd0" }
 [target.'cfg(target_os = "linux")'.dependencies]
 caps = "*"
 
--- a/components/core/src/os/users/linux.rs
+++ b/components/core/src/os/users/linux.rs
@@ -1,10 +1,9 @@
-use std::path::PathBuf;
-
-use crate::error::{Error,
-                   Result};
-
+use crate::{error::{Error,
+                    Result},
+            ok_log};
 use nix::unistd::{Group,
                   User};
+use std::path::PathBuf;
 
 /// This is currently the "master check" for whether the Supervisor
 /// can behave "as root".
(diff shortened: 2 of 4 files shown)


components/core/Cargo.toml
```diff
@@ -37,7 +37,8 @@ typemap = "*"
 url = "*"
 
 [target.'cfg(not(windows))'.dependencies]
-users = "*"
+# Pulling from git until https://github.com/nix-rust/nix/commit/3c2107bdc221a90b02d3f7118dd96f3496762cd0 is released
+nix = { git = "https://github.com/nix-rust/nix.git" }
```
christophermaier, line 41: Might be worth specifying the exact commit here to be explicit:

```toml
nix = {  git = "https://github.com/nix-rust/nix.git", rev = "3c2107bdc221a90b02d3f7118dd96f3496762cd0" }
```

components/core/src/os/users/linux.rs
```diff
@@ -26,46 +26,48 @@ pub fn can_run_services_as_svc_user() -> bool {
 pub fn can_run_services_as_svc_user() -> bool { true }
 
 pub fn get_uid_by_name(owner: &str) -> Option<u32> {
-    users::get_user_by_name(owner).map(|u| u.uid())
+    User::from_name(owner).ok()
+                          .flatten()
```
christophermaier, line 30: The use of `flatten` here (and elsewhere in this PR) surprised me a bit, but it's pretty clever.

However, I'm not sure if it's a great idea to swallow this error like this. This might be a good place to use the ["lognore"](https://github.com/habitat-sh/habitat/pull/6794#discussion_r311768892) pattern that has been discussed a few times, but not yet implemented.

stevendanna: @christophermaier I have a follow up change that changes these functions to all return Result<Option<T>> so that we don't have to ignore the change.

But, now that you mention it, I think I like logging on error here better because the story around errors out of the underlying functions is a bit more complicated than I would like.

components/core/Cargo.toml
christophermaier: Addressed in https://github.com/habitat-sh/habitat/pull/7620/commits/ab95f2a3384489c9333127f010bcb2d0688dcc09

components/core/src/os/users/linux.rs
christophermaier: Addressed in https://github.com/habitat-sh/habitat/pull/7620/commits/4b068a3f34e9ba5fcecacbefdf0d84dfa11e9f36

Issue #6477, opened 2019-04-24
Allow multiple scaffolding or something like hab-plan-build plugin


jsirex:
When packaging software we often see some basic patterns which can be extracted and re-used. In context of project it is good to use `.studiorc` or something like `base-plan.sh` with all stuff defined.

When I want to share number of helpers across the projects or community I can create a scaffolding package which enhances habitat experience. However I can specify only one scaffolding. I'm sure that scaffolding must be only one and this ok.

It would be really useful if I can specify plugin-packages which extend my `hab-plan-build`.
For example, Debian has set of build helpers to make packages directly from git repositories.

I see good potential of this feature and easy implementation.

I create package with `plugin/main.sh` (or `lib/plugin.sh`, or something like that. need to think about naming).
That script adds set of methods typical for an use-case.

Example:
```bash
: "${pkg_git_branch:=master}"
# I implement git-source plugin
# in file lib/plugin.sh
pkg_version() {
  pkg_git_version()
}

pkg_git_version() {
    git rev-list master --count
}

do_download() {
  do_git_download
}

do_git_download() {
    if [ -d "$CACHE_PATH/.git" ]; then
        build_line "Git repository exists. Attempt to reuse"
        pushd "$CACHE_PATH" > /dev/null
        # shellcheck disable=SC2154
        git remote set-url origin "$pkg_source"
        git fetch --depth 1 origin "$pkg_git_branch"
        popd > /dev/null
    else

        build_line "Git download! Doing shallow clone from $pkg_source to $CACHE_PATH"
        git clone --depth 1 --no-checkout --branch "$pkg_git_branch" "$pkg_source" "$CACHE_PATH"
    fi
}

do_verify() {
  do_git_verify
}

do_git_verify() {
    build_line "Skipping checksum verification on the repository we just created."
    return 0
}

go_clean() {
  do_git_clean
}

do_git_clean() {
    build_line "Clean up repository using git"
    pushd "$CACHE_PATH" > /dev/null
    git reset --hard HEAD
    git clean -fdx
    popd > /dev/null
}

do_unpack() {
  do_git_unpack
}

do_git_unpack() {
    build_line "Force checkout $pkg_git_branch"
    pushd "$CACHE_PATH" > /dev/null
    git checkout -f "origin/$pkg_git_branch"
    popd > /dev/null
}
```

Now, in my package I can use scaffolding and multiple plugins like following:
```bash
# in plan.sh

pkg_name=my_package
[...]

PR #7641, opened 2020-04-24
Add pkg_plugins feature

jsirex:
This feature is very similar to scaffolding but allows you to provided
multiple plugins instead of single one. Multiple plugins can bring you
small pieces of handy functionality. Consider the forum thread[1]
about plugins.

[1] https://forums.habitat.sh/t/habitat-plugins/1305

Resolves: https://github.com/habitat-sh/habitat/issues/6477
Signed-off-by: Yauhen Artsiukhou <jsirex@gmail.com>

### DIFF
--- a/components/hab/static/template_plan.sh
+++ b/components/hab/static/template_plan.sh
@@ -71,6 +71,12 @@ pkg_scaffolding="{{ scaffolding_ident }}"
 # pkg_scaffolding="some/scaffolding"
 {{~ /if}}
 
+{{~ #unless minimal}}
+
+# Optional.
+# The plugins for this plan.
+pkg_plugins=()
+{{~ /unless}}
 
 {{~ #unless minimal}}
 
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -375,6 +375,8 @@ pkg_target='@@pkg_target@@'
 pkg_origin=""
 # Each release is a timestamp - `YYYYMMDDhhmmss`
 pkg_release=$(date -u +%Y%m%d%H%M%S)
+# The default plugins setting - an empty array
+pkg_plugins=()
 # The default build deps setting - an empty array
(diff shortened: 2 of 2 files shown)


christophermaier:
@jsirex This looks really nice, and addresses a number of scaffolding issues in a clean and flexible way.

@smacfarlane @markan I'm curious what your thoughts on this are!

christophermaier:
@jsirex the build failures are due to something that has been fixed on master. If you could rebase on current master, that will fix things. Thanks!

smacfarlane:
@jsirex  Thanks for this PR,  this looks pretty awesome.   
 
We're currently working on a fairly large revamp to how we track package relationships for build dispatch in Builder and are discovering a number of edge cases we'll need account for. This means that there may be some churn in and around build and metadata, possibly changing the shape of how this feature would come together.  

In particular, because build dependencies can form [cycles](https://en.wikipedia.org/wiki/Cycle_(graph_theory)) between packages, we need to be very careful to account for how this feature would interact with other packages in the graph.   An example of this is `gcc` and `gcc-libs`.   Since we are in a cycle when evaluating build order between the two,  we need to rely on runtime edges to force some ordering.  The dependency between the two is a build edge from `gcc-libs` of `gcc`, so when evaluating build order which one goes first comes down to the order you walk the dependency graph. In this case though, `gcc` must go first, since `gcc-libs` depends on the same version of `gcc`, which as it stands today, can't be known until we build.  This is a known problem spot in the graph, so we are able to provide hints internally in order to resolve the conflict and force the correct ordering.  

At this time, I don't have a good feel for how this feature would interact with the cycle resolution and build ordering. I'd like to hold off on accepting this for now, and revisit it once the Builder work has landed. This is something I think we would like to accept and we really appreciate the effort you've put into this, it just a matter of timing :smile:

rahulgoel1:
leave a message to community member and tell him that we want to evaluate this feature. After the comment , we can close this issue.

PR #7642, opened 2020-04-27
Cleanup `FS_ROOT`

davidMcneil:
This PR unifies the `FS_ROOT_PATH` implementations and removes the
`TESTING_FS_ROOT` env var in favor of only using `FS_ROOT`.


mwrock:
looks like a windows specific `use`  of `fs` got dropped somewhere and the compiler is not happy.

components/core/src/fs.rs
```diff
@@ -663,51 +678,6 @@ fn find_command_with_pathext(candidate: &PathBuf) -> Option<PathBuf> {
...
-/// in a program at CLI parsing time and explicitly passed to inner module functions.
-///
-/// There is one exception to this rule which is supported for testing only--primarily exercising
-/// the Supervisor behavior. It allows setting a testing-only environment variable to influence the
-/// file system root for the duration of a running program.  Note that when using such an
-/// environment varible, any existing/actual Habitat packages may not run correctly due to the
-/// presence of absolute paths in package binaries and libraries. The environment variable will not
-/// be referenced, exported, or consumed anywhere else in the system to ensure that it is *only*
-/// used internally in test suites.
-///
```
christophermaier, line 684: It would be good to preserve this knowledge in the updated code. Basically, FS_ROOT should only be changed via environment variable if you're doing things like limited testing or creating new self-contained root filesystems for things like tarballs or containers.

test/end-to-end/test_launcher_exits_on_supervisor_startup_failure.ps1
```diff
@@ -5,20 +5,20 @@
...
-# Installing the launcher here because TESTING_FS_ROOT is an imperfect
+# Installing the launcher here because FS_ROOT is an imperfect
 # abstraction that needs to be removed. It turns out that even if
-# TESTING_FS_ROOT is in place, we still look for the launcher in
+# FS_ROOT is in place, we still look for the launcher in
 # `/hab` when we start up.
 #
-# Once we remove TESTING_FS_ROOT completely, we'll need to rethink how
+# Once we remove FS_ROOT completely, we'll need to rethink how
 # this test works, since we can't really make `/` read-only
```
christophermaier, line 17: I'm not sure that this big chunk of exposition makes sense anymore; it should probably be removed... if this test still functions, that is :thinking:

components/core/src/fs.rs
davidMcneil: I thought most of this sentiment was summed up in the comment on [`FS_ROOT_ENVVAR`](https://github.com/habitat-sh/habitat/blob/9df697676d01954ad9d214460e2c8b3289f933ee/components/core/src/fs.rs#L52). Do you think it is necessary to duplicate it on `FS_ROOT_PATH`?

test/end-to-end/test_launcher_exits_on_supervisor_startup_failure.ps1
davidMcneil: Fixed in (23ee7e3)[https://github.com/habitat-sh/habitat/pull/7642/commits/23ee7e3c29c1be30062bd429cbf8ca12892f873d]

components/core/src/fs.rs
christophermaier: I liked this deleted comment because it goes more into _why_ it's important to not change it (rather than the more vague sense of "don't change it"). I think it would also be good to augment that with information about what situations you might legitimately need to change it.

It probably makes sense to have the comment be on `FS_ROOT_ENVVAR`, since that is the avenue by which you would be making such a change. Perhaps a "See FS_ROOT_ENVVAR documentation, also" note here would be nice, too.

davidMcneil: That makes sense, thanks! Addressed in (92592d)[https://github.com/habitat-sh/habitat/pull/7642/commits/92592d6b5b81b3ede057115635b383e9de9d259e).

PR #7649, opened 2020-04-29
[CI] Re-enable skipped verify builds affected by core-plans refresh 

christophermaier:
A new release has gone out now, so this is no longer necessary.


christophermaier:
After the next release goes out (as of May 1, 2020), this can become a real PR and can be merged.

PR #7655, opened 2020-04-30
[REFACTOR] Push Windows Docker Engine check earlier

christophermaier:
When exporting a container on Windows, the Docker daemon must be in
Windows mode. Previously, this check was buried deep in the logic,
after we had already done a lot of work to pull packages down. Now, we
perform this check first, thereby short circuiting a whole bunch of
unnecessary work, and cleaning up a bunch of logic.

Here, I'm trying to be a bit more intentional in how platform-specific
logic is organized in this crate, in order to make it more clear what
the general interfaces are, as well as what the *exact* platform
differences are, without littering conditional compilation directives
everywhere.

Additionally, in an effort to keep related code together, as well as
to start experimenting with better error organization, the error for a
failed platform check on Windows is actually kept with the platform
checking code itself, and converted into the generic Error we use
across the crate.

### DIFF
--- a/components/pkg-export-docker/src/os/check/windows.rs
+++ b/components/pkg-export-docker/src/os/check/windows.rs
@@ -1,0 +1,61 @@
+use crate::{error::Result,
+            util};
+
+/// Currently when exporting containers on Windows, the Docker daemon
+/// *must* be in Windows mode (i.e., only Windows containers can be
+/// exported on Windows machines).
+///
+/// If the daemon is in Linux mode, we return an error and should stop
+/// the export process.
+pub(crate) fn ensure_proper_docker_platform() -> Result<()> {
+    match DockerOS::current() {
+        DockerOS::Windows => Ok(()),
+        other => Err(Error::DockerNotInWindowsMode(other).into()),
+    }
+}
+
+#[derive(Debug, Fail)]
+enum Error {
+    #[fail(display = "Only Windows container export is supported; please set your Docker daemon \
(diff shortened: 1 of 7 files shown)


components/pkg-export-docker/src/os/check/windows.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(crate) fn ensure_proper_docker_platform() -> Result<()> {
+    match DockerOS::current() {
+        DockerOS::Windows => Ok(()),
+        other => Err(Error::DockerNotInWindowsMode(other).into()),
+    }
+}
+
+#[derive(Debug, Fail)]
+enum Error {
+    #[fail(display = "Only Windows container export is supported; please set your Docker daemon \
```
davidMcneil, line 19: Very excited to standardize our error handling! Rust's error handling has undergone a lot of churn. I think currently my preferred approach is to use [`thiserror`](https://github.com/dtolnay/thiserror) in library code and [`anyhow`](https://github.com/dtolnay/anyhow) in application code. For my motivation on when to use each see [here](https://github.com/dtolnay/thiserror#comparison-to-anyhow). For my motivation for not using `failure` see [here](https://github.com/dtolnay/anyhow#comparison-to-failure). What are your thoughts?

christophermaier: I haven't seen `thiserror` or `anyhow` yet. I am aware that `failure` seems to not the the way to go, though, and would love to move this particular crate away from it.

I'll check these out!

PR #7658, opened 2020-04-30
Config file support for `hab sup run`

davidMcneil:
This PR allows users to generate a config file for `hab sup run` with the `--generate-config` flag. A config file field can be used to set every argument of `hab sup run`. The order of precedence for `hab sup run` configuration is as follows:

1. Fixed config file location (`/hab/sup/default/config/sup.toml`)
2. Config files passed in with `hab sup run --config-files /config/file1.toml /config/file2.toml`
    * Multiple files can be specified
    * Files that occur later in the list have precedence over previous files
3. Environment variable if one exists
4. Command-line interface

This is a large change with a high potential for breakage. In order to combat this, there are several "stages" to the PR with several tests added along the way.

### PR Stages

1. Add tests for setting every `hab sup run` CLI argument (a7dd1c9a7 - f847f0453)
    * There are three main functions that need to be tested: [mgrcfg_from_sup_run_matches](https://github.com/habitat-sh/habitat/blob/404f558d158ed57054843add431d28befd8eca01/components/sup/src/main.rs#L278), [svc_load_from_input](https://github.com/habitat-sh/habitat/blob/404f558d158ed57054843add431d28befd8eca01/components/sup/src/main.rs#L502), and [set_supervisor_logging_options](https://github.com/habitat-sh/habitat/blob/404f558d158ed57054843add431d28befd8eca01/components/sup/src/main.rs#L471)
    * These can be thought of as the main CLI entry points where we convert the passed in arguments into internal data structures.
    * The tests start [here](https://github.com/habitat-sh/habitat/blob/538894de8f1e25d45f7843cb7491f796905df9eb/components/sup/src/main.rs#L743) 
2. Cleanup the CLI (02f4e4816 - 5cf19dec4)
    * Our CLI definition had gotten a bit crufty. There were several places where we were not using the full power of clap (eg: not using default_value, not using env, ...)
    * While this cleanup was happening the `clap_app` and `structopt` version of the CLI were kept in sync with this [test](https://github.com/habitat-sh/habitat/blob/538894de8f1e25d45f7843cb7491f796905df9eb/components/hab/src/cli/hab/tests.rs#L396).
3. Remove the `clap_app` version of `hab sup run` (d3ac5efe8)
    * `structopt` provides a [method](https://github.com/habitat-sh/habitat/blob/538894de8f1e25d45f7843cb7491f796905df9eb/components/hab/src/cli.rs#L1134) to easily get the clap app from a `structopt` type
    * We can use our CLI tests to ensure everthing is still working correctly
    * We are using the `structopt` definition of `hab sup run` but we are basically using it as a fancy way to define a `clap::App`. We are not using its full power to deserialize args into the `structopt` type.
4. Use the `structopt` type instead of `ArgMatches` for `hab sup run` (b1670b367)
    * This is the main [entry point](https://github.com/habitat-sh/habitat/blob/538894de8f1e25d45f7843cb7491f796905df9eb/components/sup/src/main.rs#L169) of the transistion
    * This significantly reduced the need for manual parsing logic. We can instead simply rely on the types.
    * `mgrcfg_from_sup_run_matches` and `svc_load_from_input` where combined into [`split_apart_sup_run`](https://github.com/habitat-sh/habitat/blob/538894de8f1e25d45f7843cb7491f796905df9eb/components/sup/src/main.rs#L268)
5. Add tests for setting every `hab sup run` CLI argument from a config file (a831a94e2)
    * The tests start [here](https://github.com/habitat-sh/habitat/blob/538894de8f1e25d45f7843cb7491f796905df9eb/components/sup/src/main.rs#L1073)
    * As part of adding these tests there was some work to fix some deserialization


components/sup/src/main.rs
```diff
@@ -801,17 +801,18 @@ RCFaO84j41GmrzWddxMdsXpGdn3iuIy7Mw3xYrjPLsE="#,
                                                  application_environment: None,
                                                  binds:                   None,
                                                  specified_binds:         None,
-                                                 binding_mode:            None,
+                                                 binding_mode:            Some(1),
```
christophermaier, line 786: Are we really required to refer to numbers here in the protobuf types?

components/common/src/types.rs
```diff
@@ -40,81 +40,57 @@ pub struct UserInfo {
...
+}
+
+impl From<Vec<EventStreamMetaPair>> for EventStreamMetadata {
+    fn from(pairs: Vec<EventStreamMetaPair>) -> Self {
+        let mut h = HashMap::new();
+        for pair in pairs {
+            h.insert(pair.0, pair.1);
         }
-        Ok(())
+        EventStreamMetadata(h)
```
christophermaier, line 101: A more functional formulation could be:  

```rust
pairs.into_iter()
     .map(|p| (p.0, p.1))
     .collect::<HashMap<_, _>>()
     .into()
```

components/common/src/command/package/install.rs
```diff
@@ -164,6 +165,12 @@ impl FromStr for InstallSource {
     }
 }
 
+impl std::convert::TryFrom<&str> for InstallSource {
+    type Error = habitat_core::Error;
+
+    fn try_from(s: &str) -> StdResult<Self, Self::Error> { InstallSource::from_str(s) }
```
christophermaier, line 171: Similarly here... is there benefit of having a `TryFrom` if we already have `FromStr`?

components/hab/src/cli/hab/svc.rs
```diff
@@ -147,8 +146,12 @@ pub struct SharedLoad {
                 possible_values = &["strict", "relaxed"])]
     pub binding_mode:          habitat_sup_protocol::types::BindingMode,
     /// The interval in seconds on which to run health checks
+    // We would prefer to use `HealthCheckInterval`. However, `HealthCheckInterval` uses a map based
+    // serialization format. We want to allow the user to simply specify a `u64` to be consistent
+    // with the CLI, but we cannot change the serialization because the spec file depends on the map
+    // based format.
```
christophermaier, line 152: We really should fix this :frowning:

components/common/src/types.rs
```diff
@@ -241,6 +245,15 @@ impl fmt::Debug for EventStreamServerCertificate {
     }
 }
 
+impl PartialEq<EventStreamServerCertificate> for EventStreamServerCertificate {
+    fn eq(&self, other: &EventStreamServerCertificate) -> bool {
+        match (self.0.to_der(), other.0.to_der()) {
```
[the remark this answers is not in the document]
davidMcneil: Maybe because `to_der` is a failable operation which makes the logic a little awkward. Should two errors be equal? Since we only use it for testing I think this implementation is ok.

components/common/src/types.rs
```diff
@@ -230,24 +265,44 @@ impl FromStr for EventStreamServerCertificate {
...
+        let contents = fs::read(&path)?;
+        Ok(EventStreamServerCertificate { path,
+                                          certificate: Certificate::from_pem(&contents)? })
     }
 }
 
+impl std::convert::TryFrom<&str> for EventStreamServerCertificate {
+    type Error = Error;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> { EventStreamServerCertificate::from_str(s) }
```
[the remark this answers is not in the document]
davidMcneil: Yeah, this is unfortunate. `serde` uses is it as a trait bound [here](https://github.com/habitat-sh/habitat/blob/b22c4a07a9509bd628438adc443b0367332242ab/components/common/src/types.rs#L230). Once rust gets specialization there should probably be a blanket `impl` for `From<&str>` for all types that implement `Parse`.

components/hab/src/cli/hab/svc.rs
davidMcneil: Agreed. How do we do it without breaking backwards compatibility? I guess support both formats when `Deserialize`ing a spec file.

components/common/src/types.rs
davidMcneil: Much better, thanks!

components/sup/src/main.rs
```diff
@@ -519,6 +521,14 @@ fn svc_load_from_input(m: &ArgMatches) -> Result<sup_proto::ctl::SvcLoad> {
     Ok(msg)
 }
 
+fn get_health_check_interval_from_input(m: &ArgMatches<'_>)
+                                        -> Option<sup_proto::types::HealthCheckInterval> {
+    // Value will have already been validated by `cli::valid_health_check_interval`
+    m.value_of("HEALTH_CHECK_INTERVAL")
+     .and_then(|s| HealthCheckInterval::from_str(s).ok())
+     .map(HealthCheckInterval::into)
```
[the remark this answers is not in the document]
davidMcneil: Nice, much better! This is also good to know about as we transition more of the cli to `structopt`.

components/sup/src/main.rs
davidMcneil: No, I just needed to use `into` on the enum. Fixed in 26e8c1bcf29de69db2a9576043da8c18eb29b9d1 along with your other suggestions.

components/common/src/types.rs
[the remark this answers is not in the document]
christophermaier: Could you add a note that it's only for testing?

I don't suppose we can flag this with a `#[test]` attribute because of the crate it's in , right?

components/common/src/cli.rs
```diff
@@ -47,8 +48,7 @@ pub const DEFAULT_BINLINK_DIR: &str = "/bin";
 pub const DEFAULT_BINLINK_DIR: &str = "/usr/local/bin";
 
 pub fn cache_key_path_from_matches(matches: &ArgMatches<'_>) -> PathBuf {
-    let path = matches.value_of("CACHE_KEY_PATH")
-                      .expect("CACHE_KEY_PATH required");
+    let path = clap::value_t!(matches, "CACHE_KEY_PATH", PathBuf).expect("CACHE_KEY_PATH required");
     PathBuf::from(path)
```
christophermaier: You don't need `PathBuf::from(path)` anymore.

components/hab/src/cli/hab/svc.rs
christophermaier: Yup, exactly.

davidMcneil:
> I don't suppose we can flag this with a #[test] attribute because of the crate it's in , right?

Right :disappointed:. At some point I think it would be interesting to do some high level refactorings on what types/functions are contained in what crates.

components/common/src/types.rs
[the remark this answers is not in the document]
davidMcneil: Yes, that would be good.

Right :disappointed:. At some point I think it would be interesting to do some high level refactorings on what types/functions are contained in what crates.

PR #7660, opened 2020-05-01
[REFACTOR] Remove some lifetimes to simplify Docker exporter

christophermaier:
These lifetime parameters were added to allow us to hold onto `str`
references from our Clap `ArgMatches`. However, this ends up
implicitly coupling all of our types to choices made by our CLI
library. Also, there's no real benefit to not allocating a few extra
Strings in a short-lived CLI application like this, so there's no
reason to carry around the extra cognitive overhead.

### DIFF
--- a/components/pkg-export-docker/src/build.rs
+++ b/components/pkg-export-docker/src/build.rs
@@ -84,27 +84,27 @@ fn default_docker_base_image() -> Result<String> {
 /// When a `BuildSpec` is created, a `BuildRoot` is returned which can be used to produce exported
 /// images, archives, etc.
 #[derive(Debug)]
-pub struct BuildSpec<'a> {
+pub struct BuildSpec {
     /// A string representation of a Habitat Package Identifer for the Habitat CLI package.
-    pub hab:                &'a str,
+    pub hab:                String,
     /// A string representation of a Habitat Package Identifer for the Habitat Launcher package.
-    pub hab_launcher:       &'a str,
+    pub hab_launcher:       String,
     /// A string representation of a Habitat Package Identifer for the Habitat Supervisor package.
-    pub hab_sup:            &'a str,
+    pub hab_sup:            String,
     /// The Builder URL which is used to install all service and extra Habitat packages.
-    pub url:                &'a str,
+    pub url:                String,
     /// The Habitat release channel which is used to install all service and extra Habitat
     /// packages.
     pub channel:            ChannelIdent,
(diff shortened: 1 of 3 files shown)


components/pkg-export-docker/src/build.rs
```diff
@@ -328,10 +339,10 @@ impl<'a> BuildSpec<'a> {
                               -> Result<FullyQualifiedPackageIdent> {
         self.install(ui,
                      ident_or_archive,
-                     self.url,
+                     &self.url,
                      &self.channel,
                      fs_root_path,
-                     self.auth)
+                     self.auth.as_ref().map(String::as_str)) //eww
```
davidMcneil, line 345: Using [`as_deref`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.as_deref) makes this cleaner.

components/pkg-export-docker/src/build.rs
```diff
@@ -84,52 +84,62 @@ fn default_docker_base_image() -> Result<String> {
...
+                                            .to_string(),
                        hab_launcher:       m.value_of("HAB_LAUNCHER_PKG")
-                                            .unwrap_or(DEFAULT_LAUNCHER_IDENT),
-                       hab_sup:            m.value_of("HAB_SUP_PKG").unwrap_or(DEFAULT_SUP_IDENT),
-                       url:                m.value_of("BLDR_URL").unwrap_or(&default_url),
+                                            .unwrap_or(DEFAULT_LAUNCHER_IDENT)
+                                            .to_string(),
+                       hab_sup:            m.value_of("HAB_SUP_PKG")
+                                            .unwrap_or(DEFAULT_SUP_IDENT)
+                                            .to_string(),
```
davidMcneil, line 126: I dont know if the effort is worth it, but if we switched this to using `structopt` this duplication could go away. It would also remove duplication in how we define the CLI.

christophermaier: Yeah, I'd eventually like to get all our CLI tooling using structopt.

components/pkg-export-docker/src/build.rs
christophermaier: Ah, yes :+1:

christophermaier: Fixed in https://github.com/habitat-sh/habitat/pull/7660/commits/da8926ab7034998abd66c172fd0e66a062a0f4d3

PR #7663, opened 2020-05-01
Update MacOS Bootstrapper version

smacfarlane:
This updates to the latest MacOS bootstrapper version, as well as adding a build of the MacOS artifact to the verify pipeline. 

The update to the bootstrapper version is to update the version of libarchive to match with the version used in the habitat `core/libarchive` package.


.expeditor/scripts/shared.sh
```diff
@@ -339,3 +339,68 @@ push_current_branch() {
...
+    sudo hdiutil attach "${bootstrap_package_name}.dmg"
+    sudo installer -verbose -pkg "/Volumes/Habitat macOS Bootstrapper/${bootstrap_package_name}.pkg" -target /
+    sudo hdiutil detach "/Volumes/Habitat macOS Bootstrapper"
+    brew install wget
+    export PATH=/opt/mac-bootstrapper/embedded/bin:/usr/local/bin:$PATH
+    
+}
+
+macos_use_cert_file_from_linux_cacerts_package() {
+    # BEGIN CERTIFICATE SHENANIGANS
```
christophermaier, line 357: With all the SSL changes in recent months, I wonder if all this is still necessary thinking

(This doesn't need to be addressed for this PR, though.)

.expeditor/scripts/verify/build_mac_package.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash
+#!/usr/local/bin/bash
```
christophermaier, line 1: What are the pros and cons of this versus `#!/usr/bin/env bash`?

smacfarlane: The pro is that I can be reasonably certain `/usr/local/bin/bash` is the Homebrew installed bash.  `/usr/bin/env` bash would probably also resolve to that, but would be dependent on how the PATH was constructed.    

I personally prefer the failure case of `/usr/local/bin/bash` to  `/usr/bin/env bash` resolving to the wrong bash; `/usr/local/bin/bash command not found` vs `declare -g invalid option` (if you got system bash),  if something were to go sideways on us.

PR #7673, opened 2020-05-06
Remove duplicated serialization logic in SvcMember

davidMcneil:
Both the [`SvcMember`](https://github.com/habitat-sh/habitat/blob/4c10302b0c3f8aeb04fb2739580472c257e03ee7/components/sup/src/manager/service/context.rs#L376) and [`CensusMemberProxy`](https://github.com/habitat-sh/habitat/blob/4c10302b0c3f8aeb04fb2739580472c257e03ee7/components/sup/src/census.rs#L671) acted as a serialization proxy for [`CensusMember`](https://github.com/habitat-sh/habitat/blob/4c10302b0c3f8aeb04fb2739580472c257e03ee7/components/sup/src/census.rs#L563). This lead to significant duplication of code. This PR makes `SvcMember` an alias of `CensusMemberProxy` removing the duplication. The one difference between the two serialization methods was the [`package`](https://github.com/habitat-sh/habitat/blob/2d4d374c7eb19c5c01d9adadfa3d4d7bedaf50f0/components/sup/doc/render_context_schema.json#L103) field. I think it is ok to add this field. There was even a [todo comment](https://github.com/habitat-sh/habitat/blob/4c10302b0c3f8aeb04fb2739580472c257e03ee7/components/sup/src/manager/service/context.rs#L444) to add it. This PR also moves the `Cow` into the `CensusMemberProxy` type reducing the need to clutter `Cow`s everywhere and increasing encapsulation.

### DIFF
--- a/components/sup/src/census.rs
+++ b/components/sup/src/census.rs
@@ -578,16 +578,15 @@ pub struct CensusMember {
     pub update_election_is_no_quorum: bool,
     pub update_election_is_finished: bool,
     pub sys: SysInfo,
-
-    alive:     bool,
-    suspect:   bool,
-    confirmed: bool,
-    departed:  bool,
+    pub alive: bool,
+    pub suspect: bool,
+    pub confirmed: bool,
+    pub departed: bool,
     // Maps must be represented last in a serializable struct for the current version of the toml
     // crate. Additionally, this deserialization method is required to correct any ordering issues
     // with the table being serialized - https://docs.rs/toml/0.4.0/toml/ser/fn.tables_last.html
     #[serde(serialize_with = "toml::ser::tables_last")]
-    pub cfg:   toml::value::Table,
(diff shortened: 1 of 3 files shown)


components/sup/src/census.rs
```diff
@@ -668,44 +667,57 @@ impl CensusMember {
...
+    pub fn new_owned(c: CensusMember) -> Self { CensusMemberProxy(Cow::Owned(c)) }
+
+    #[cfg(test)]
+    pub fn to_mut(&mut self) -> &mut CensusMember { self.0.to_mut() }
+}
+
+impl std::ops::Deref for CensusMemberProxy<'_> {
+    type Target = CensusMember;
+
+    fn deref(&self) -> &Self::Target { &(self.0) }
```
christophermaier, line 684: This is a nice addition :+1:

PR #7674, opened 2020-05-06
`hab pkg export docker` is now `hab pkg export container`

christophermaier:
This PR renames the `hab pkg export docker` command (and all associated code) to be more tool-agnostic, as `hab pkg export container`. 

This particular change is mostly cosmetic in nature, but paves the way for upcoming changes where users will have a choice of container-generating backends, so it will no longer be just Docker in play.

`hab pkg export docker` remains as an alias for backwards compatibility.


christophermaier:
@smacfarlane We actually do have a recently-added suite of e2e tests for this exporter. It's currently done in terms of `hab pkg export docker`, though, so I need to go tweak that :sweat_smile:

smacfarlane:
> @smacfarlane We actually do have a recently-added suite of e2e tests for this exporter. It's currently done in terms of `hab pkg export docker`, though, so I need to go tweak that sweat_smile

TIL :joy_cat:

PR #7677, opened 2020-05-08
[REFACTOR] Introduce "Engine" abstraction to `hab pkg export container`

christophermaier:
This is yet another refactor in preparation for more changes in the container exporter. Here, we push all the shelling out to `docker` that we do behind an "Engine" abstraction. No outward-facing changes are present here, but it will make it easier for us to swap something else in for `docker` in the future with minimal disruption to the fundamental logic of the crate.

Please read the commit messages for further detail and background.

### DIFF
--- a/components/pkg-export-container/src/engine.rs
+++ b/components/pkg-export-container/src/engine.rs
@@ -2,8 +2,7 @@
 //! dealing with container images, while remaining unconcerned about
 //! which underlying tool is actually performing those tasks.
 
-use crate::error::{Error,
-                   Result};
+use crate::error::Result;
 use habitat_core::fs::find_command;
 use std::{path::{Path,
                  PathBuf},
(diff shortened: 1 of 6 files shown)


components/pkg-export-container/src/engine.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[fail(display = "Could not determine container image ID for: {}", _0)]
+    ImageIdNotFound(String),
+    #[fail(display = "Removing local container images failed with exit code: {}",
+           _0)]
+    RemoveFailed(ExitStatus),
+    #[fail(display = "Container image push failed with exit code: {}", _0)]
+    PushFailed(ExitStatus),
+}
+
+pub struct Engine {
```
davidMcneil, line 28: Will this ultimately be a trait?

christophermaier: Yup!

PR #7682, opened 2020-05-12
REFACTOR: Remove ImageBuilder struct

christophermaier:

### DIFF
--- a/components/pkg-export-container/src/container.rs
+++ b/components/pkg-export-container/src/container.rs
@@ -1,7 +1,8 @@
 use crate::{build::BuildRoot,
             engine::Engine,
             error::Result,
-            naming::Naming,
+            naming::{ImageIdentifiers,
+                     Naming},
             util};
 use failure::SyncFailure;
 use habitat_common::ui::{Status,
@@ -262,7 +263,10 @@ impl BuildContext {
 
         // TODO (CM): Ideally, we'd toss this error much earlier,
         // since this error would be based on user input errors
-        let (name, tags, expanded_identifiers) = naming.image_identifiers(&ident, &channel)?;
+        let ImageIdentifiers { name,
+                               tags,
+                               expanded_identifiers, } =
+            naming.image_identifiers(&ident, &channel)?;
 
         let id = engine.build(self.0.workdir(), &expanded_identifiers, memory)?;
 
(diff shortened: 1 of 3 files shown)


components/pkg-export-container/src/container.rs
```diff
@@ -144,18 +42,21 @@ pub struct ContainerImage {
...
-    fn tags(&self) -> Vec<String> { self.tags.clone() }
+    /// All the identifiers for this image (in {name}:{tag} format)
+    expanded_identifiers: Vec<String>,
 }
 
 impl ContainerImage {
     // TODO (CM): temporary; we shouldn't use this at all
     pub fn workdir(&self) -> &Path { self.workdir.as_path() }
 
+    pub fn expanded_identifiers(&self) -> &Vec<String> { &self.expanded_identifiers }
```
davidMcneil, line 54: Can we make the return type `&[String]`?

components/pkg-export-container/src/naming.rs
```diff
@@ -97,11 +97,12 @@ impl Naming {
...
 
     /// Return the image name, along with a (possibly empty) vector of
-    /// additional tags.
+    /// additional bare tags, and a vector containing "name:tag"
+    /// identifiers (or just "name" if there are no tags).
     pub fn image_identifiers(&self,
                              ident: &FullyQualifiedPackageIdent,
                              channel: &ChannelIdent)
-                             -> Result<(String, Vec<String>)> {
+                             -> Result<(String, Vec<String>, Vec<String>)> {
```
davidMcneil, line 105: Would it be worth having a type alias or maybe even a new type for the different types of tags?

components/pkg-export-container/src/container.rs
davidMcneil, line 54: Can we make the return type `&[String]`?

components/pkg-export-container/src/naming.rs
christophermaier: Maybe... writing this return type killed a little bit of my soul.

I think that with some additional future refactorings this would get a lot simpler, but it may be a good idea to contain the gnarliness until then.

components/pkg-export-container/src/container.rs
christophermaier: Addressed in https://github.com/habitat-sh/habitat/pull/7682/commits/decc37827a8b30249603803fcba5f9f7d5922bbe

components/pkg-export-container/src/naming.rs
christophermaier: Addressed in https://github.com/habitat-sh/habitat/pull/7682/commits/f057ddb7a2c74ef41ca2717d995e051020b23400; much better :smile:

PR #7685, opened 2020-05-14
Add Bulidah as an engine for unprivileged container builds

christophermaier:
Buildah (https://buildah.io) is another tool that can be used to
create container images. While it has several interesting
capabilities (like an alternative imperative, non-Dockerfile-driven
build process), the main one that interests us is the ability to build
images without having to use a Docker daemon.

This allows ordinary unprivileged users to export Habitat container
images.

As Buildah only works on Linux, this ability is thus necessarily
limited to Linux platforms. To enable this functionality, use the new
`--engine` option:

```sh
hab pkg export container core/redis --engine=buildah
```

The default engine remains Docker. While the `--engine` option *is*
technically available on Windows, `docker` is the only allowed value.

One important thing to note is that Docker and Buildah store their
local images in different locations. This means that, despite the
tools creating compatible images, you will not be able to (directly)
run a Buildah-built container under Docker, in contrast to what your
typical "export a container, run that container" development
workflow. There are two options, though.

First, use `podman` (https://podman.io). Podman and Buildah are sister
projects in an overall effort to split apart the features of the
Docker monolith. While Buildah handles construction of images, Podman
handles the running of those images as containers (that is, `buildah`
is to `docker build` as `podman` is to `docker run`). As such, Podman
looks for images in the same place that Buildah creates them. Like
Buildah, Podman also does not require a privileged account to
run. And, as a bonus, you can essentially use `podman` as an alias for
`docker`.

The second alternative is to use `buildah` to export your image to
your Docker daemon. This effectively copies your image from "Buildah
space" to "Docker space".

```sh
hab pkg exec core/buildah buildah push core/redis:TAG docker-daemon:core/redis:TAG
```

Since this command does interact with a Docker daemon, you will
require suitable privileges to execute this operation.

It should be noted that these approaches are not required if you wish
to export a container and push the generated images to a remote
repository as part of the same `hab pkg export container`
invocation. The Buildah engine implementation knows how to push to
remote repositories without any additional effort on the part of the
user.

As an implementation note, the previous `Engine` struct was converted
into a trait, with an implementation each for Docker and Buildah. The
trait defines both the commands needed to perform the various
operations of an Engine, as well as the methods to execute those
commands. Implementors of the trait only have to define the base
commands; execution is provided in the trait definition itself.

Then, an instance is created based on user input and
used throughout. The broader exporter code knows nothing of any
concrete implementation (indeed, those are left intentionally private
within the `engine` module), instead operating only in terms of the
`Engine` abstraction.

### DIFF
--- a/components/pkg-export-container/habitat/plan.sh
+++ b/components/pkg-export-container/habitat/plan.sh
@@ -5,6 +5,7 @@ pkg_origin=core
 pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_license=('Apache-2.0')
 pkg_deps=(
+    core/buildah
     core/docker
 )
 pkg_build_deps=(
--- a/components/pkg-export-container/habitat/plan.sh
+++ b/components/pkg-export-container/habitat/plan.sh
@@ -4,20 +4,22 @@ _pkg_distname=$pkg_name
 pkg_origin=core
 pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_license=('Apache-2.0')
-# The result is a portable, static binary. However, we shell out to the
-# Docker command which we need at runtime.
-pkg_deps=(core/docker)
-pkg_build_deps=(core/musl
-                core/zlib-musl
-                core/xz-musl
-                core/bzip2-musl
-                core/libarchive-musl
-                core/openssl-musl
-                core/coreutils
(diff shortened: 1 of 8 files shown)


christophermaier:
> Should we add some buildah specific end-to-end tests?

:man_facepalming: Yes. I thought I had done that earlier, but I think I was getting that confused with the general container export tests I'd added a few PRs back.

Will add.

PR #7686, opened 2020-05-15
Remove libarchive dependency

davidMcneil:
This PR replaces the `libarchive` crate with `xz2` and `tar`. The `tar` crate is pure rust and the `xz2` crate automatically vendors its dependencies easing the build process.

This PR also removes the `libarchive` habitat package from the build process as well as its dependencies (`bzip2`, `xz`, and `zlib`).

As part of removing `libarchive`, `PackageArchive` was cleaned up. [`metadata`](https://github.com/habitat-sh/habitat/blob/eed8fa4d09970284b1658fc03875c646ef441862/components/core/src/package/archive.rs#L176) used to be an `Option` that was lazily loaded. This overly complicated the logic and api. `metadata` is now loaded up front and `PackageArchive::new` returns a `Result`.

When developing this PR it was discovered there was a test `hart` package that was not xz compressed (bb9ce9037). This PR removes support for uncompressed `hart`s. ~~This PR is marked with `DO-NOT-MERGE` until it can be verified that uncompressed `hart`s do not exist in the wild.~~ I tested 1,000's of packages all of which were compressed. An uncompressed package would be very odd as the build process explicitly includes a compression step.


mwrock:
Looks like you are hitting some permissions related errors on windows from `unpack`, which is surfacing is a somewhat convoluted manner when it tries to access powershell (this errors because of the tar error trying to unpack the powershell artifact). I'm not sure what the root cause is but it may not be a bad idea to add some unit tests around unpack to more directly catch failures.

davidMcneil:
Thanks for the heads up @mwrock. It was not clear to me exactly what was going on. Looks like this [line](https://github.com/habitat-sh/habitat/blob/67fcdebae9071ef43ba45d5e6b2922cbe8b5d307/components/core/src/package/archive.rs#L345) is not working or is at least behaving differently than it used to. I will have to dig into this more.

components/hab/habitat/plan.sh
```diff
@@ -60,15 +56,12 @@ do_prepare() {
   export rustc_target="x86_64-unknown-linux-musl"
   build_line "Setting rustc_target=$rustc_target"
 
+  # TODO (DM): `core/openssl-musl` includes the `core/zlib-musl` package even though
```
smacfarlane, line 59: This is likely due to the kernel2 package not having been rebuilt  since the `no-zlib` change was introduced.  This _may_ cause us some problems between `x86_64-linux` and `x86_86-linux-kernel2`.   We should work with the core-plans maintainers to get a build of openssl-musl for `x86_86-linux-kernel2` to correct this, and ensure we're using the latest version of openssl.

davidMcneil: Good catch, thanks! This looks like exactly what is happening and does cause issues between the platforms. I opened a new PR (#7710) that builds on top of this PR, but it also vendors openssl fixing this problem. If #7710 gets approved I can merge that PR and close this one.

PR #7702, opened 2020-05-19
Gossip service health

davidMcneil:
Resolves #5236

This is still a work in progress. The way we generate rumors seems good, but the way we consume them seems very wrong.

### DIFF
--- a/components/butterfly/src/rumor/service_health.rs
+++ b/components/butterfly/src/rumor/service_health.rs
@@ -1,0 +1,131 @@
+use crate::{error::{Error,
+                    Result},
+            protocol::{self,
+                       newscast::{self,
+                                  service_health::Health,
+                                  Rumor as ProtoRumor,
+                                  RumorPayload},
+                       FromProto},
+            rumor::{Rumor,
+                    RumorType}};
+use habitat_core::service::ServiceGroup;
+use std::{cmp::Ordering,
+          fmt,
+          mem,
+          str::FromStr};
+
+/// Convert numeric values from protobuf into `Health` statuses.
+impl From<i32> for Health {
+    fn from(value: i32) -> Self {
+        match value {
+            0 => Health::Ok,
+            1 => Health::Warning,
+            2 => Health::Critical,
+            3 => Health::Unknown,
+            _ => Health::Unknown,
+        }
+    }
+}
+
+impl fmt::Display for Health {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let msg = match *self {
+            Health::Ok => "OK",
+            Health::Warning => "WARNING",
+            Health::Critical => "CRITICAL",
+            Health::Unknown => "UNKNOWN",
+        };
+        write!(f, "{}", msg)
+    }
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq)]
+pub struct ServiceHealth {
+    pub member_id:     String,
+    pub service_group: ServiceGroup,
+    pub incarnation:   u64,
+    pub health:        Health,
+}
+
+impl ServiceHealth {
+    /// Creates a new ServiceFile.
+    pub fn new<S>(member_id: S, service_group: ServiceGroup, health: Health) -> Self
+        where S: Into<String>
+    {
+        ServiceHealth { member_id: member_id.into(),
+                        service_group,
+                        incarnation: 0,
+                        health }
+    }
+}
+
+impl fmt::Display for ServiceHealth {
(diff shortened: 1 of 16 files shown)


components/sup/src/census.rs
```diff
@@ -85,6 +90,7 @@ impl CensusRing {
            || (election_update_rumors.get_update_counter() > self.last_election_update_counter)
            || (service_config_rumors.get_update_counter() > self.last_service_config_counter)
            || (service_file_rumors.get_update_counter() > self.last_service_file_counter)
+           || (service_health_rumors.get_update_counter() > self.last_service_health_counter)
```
davidMcneil, line 93: Given the potential frequency and volume of health checks, I do not think it is feasible to use the same logic for service health rumors.

components/sup/src/census.rs
```diff
@@ -715,6 +754,9 @@ impl<'a> Serialize for CensusMemberProxy<'a> {
         strukt.serialize_field("suspect", &self.suspect)?;
         strukt.serialize_field("confirmed", &self.confirmed)?;
         strukt.serialize_field("departed", &self.departed)?;
+        strukt.serialize_field("service_health", &self.service_health)?;
+        strukt.serialize_field("service_health_incarnation",
```
davidMcneil, line 758: I do not think this is a good approach. The need to serialize the incarnation feels very wrong. I think it is possible we do not want an incarnation for service health. Instead, we should just always update the status, but inorder for this to be possible we need to be sure to garbage-collect old rumors.

components/butterfly/src/rumor/service_health.rs
```diff
@@ -0,0 +1,131 @@
...
+            mem::swap(self, &mut other);
+            true
+        }
+    }
+
+    fn kind(&self) -> RumorType { RumorType::ServiceHealth }
+
+    fn id(&self) -> &str { &self.member_id }
+
+    fn key(&self) -> &str { &self.service_group }
```
christophermaier, line 131: :thinking: Since health of a service is, thus far, a Supervisor-specific thing (Service X on Supervisor A can be unhealthy, while Service X on Supervisor B can be just fine), this starts to feel like we'll need to have logic to coalesce the health of each service instance into an overall "service health", which could get really thorny :grimacing:

components/sup/src/manager/service.rs
```diff
@@ -352,11 +356,13 @@ impl Service {
                    .join("hooks")
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub async fn new(sys: Arc<Sys>,
                      spec: ServiceSpec,
                      manager_fs_cfg: Arc<FsCfg>,
                      organization: Option<&str>,
                      gateway_state: Arc<GatewayState>,
+                     butterfly: ButterflyServer,
```
christophermaier, line 365: I know this is a draft and all, but further threading a Butterfly server down here squicks me out a little bit :grimacing: 

(I don't yet have anything more constructive than that, sorry :sweat_smile:)

components/sup/src/census.rs
christophermaier: What are your specific concerns? (Not doubting it, just asking for clarity)

components/sup/src/census.rs
christophermaier: I don't think incarnation is wrong here, necessarily (it's a good and simple way to signal when information changes). Given the rumor merging logic, a new health will overwrite an old one, which is what we want. Garbage collection is (according to my perhaps fuzzy recollection) more concerned with knowing when it's OK to stop holding on to rumors that we shouldn't care about anymore; e.g., supervisor A stopped running service X 10 days ago, so the fact that we're still holding onto the fact that it was healthy back then (and potentially telling others that fact _now_) is silly.

components/sup/src/manager/service.rs
```diff
@@ -844,6 +856,29 @@ impl Service {
...
+    fn gossip_latest_health_generator_rumor_rsw_mlw_rhw(&self) -> impl Fn(HealthCheckResult) {
+        let butterfly = self.butterfly.clone();
+        let member_id = self.sys.member_id.clone();
+        let service_group = self.service_group.clone();
+        move |health| {
+            let incarnation = butterfly.service_health_store
+                                       .lock_rsr()
+                                       .service_group(&service_group)
+                                       .map_rumor(&member_id, |rumor| rumor.incarnation + 1)
+                                       .unwrap_or(1);
```
christophermaier, line 873: Would there ever be a case where we _don't_ have an incarnation here?

davidMcneil:
Closing for now because the gossip layer probably requires additional work before this change can be made.

PR #7707, opened 2020-05-21
Use docker executor for kernel2 verify

smacfarlane:
The kernel2 verify steps mutate the CI host by installing the kernel2  hab cli, so we should use disposable infrastructure for these steps.
 
### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -571,7 +571,7 @@ steps:
       - .expeditor/scripts/verify/build_package.sh test-services/test-probe
     expeditor:
       executor:
-        linux:
+        docker:
           privileged: true
     retry:
       automatic:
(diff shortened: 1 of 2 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -571,7 +571,7 @@ steps:
       - .expeditor/scripts/verify/build_package.sh test-services/test-probe
     expeditor:
       executor:
-        linux:
+        docker:
```
christophermaier, line 574: I think the `env` block needs to be pulled down to an `environment` block under `docker`

smacfarlane: Hmm, good point.  It does appear to be working though :thinking: 
https://buildkite.com/chef/habitat-sh-habitat-master-verify/builds/5629#3a29f9a0-1008-41ed-a1fd-daaaea8c9cbd/211-603
https://buildkite.com/chef/habitat-sh-habitat-master-verify/builds/5629#46d73875-e5b3-42f4-8278-76c2686e710a/181-524

PR #7710, opened 2020-05-26
Remove libarchive and Vendor openssl

davidMcneil:
This PR vendors the openssl dependency making the setup to build habitat easier and making the versions built with `hab` and those built with `cargo` more consistent.

It builds on top of the #7686, but fixes the issue discussed [here](https://github.com/habitat-sh/habitat/pull/7686#discussion_r430640833) by vendering openssl.


Cargo.lock
```diff
@@ -2365,6 +2357,15 @@ version = "0.1.2"
...
+name = "openssl-src"
+version = "111.9.0+1.1.1g"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2dbe10ddd1eb335aba3780eb2eaa13e1b7b441d2562fd962398740927f39ec4"
+dependencies = [
+ "cc",
+]
+
 [[package]]
 name = "openssl-sys"
```
sdelano, line 2370: Is `openssl-sys` still required with this change?

davidMcneil: Yes, `openssl-sys` is Rust bindings to `openssl`. We still require `openssl`, but `openssl-sys` will automatically build `openssl` as opposed to requiring it to be installed on the OS.

components/core/src/package/archive.rs
```diff
@@ -262,71 +248,53 @@ impl PackageArchive {
...
-                                          .filter_map(|port| port.parse::<u16>().ok())
-                                          .collect();
-                Ok(ports)
-            }
-            Ok(None) => Ok(vec![]),
-            Err(e) => Err(e),
+    pub fn exposes(&mut self) -> Vec<u16> {
+        if let Some(data) = self.read_metadata(MetaFile::Exposes) {
+            data.split_whitespace()
+                .filter_map(|port| port.parse().ok())
```
christophermaier, line 254: Even though the previous implementation did the same thing, it seems that it would be useful to at least log an error in case the metadata file existed but contained bogus contents that could not be parsed as a port.

Better still, perhaps this should still return a `Result`.

davidMcneil: Thanks, that is much better. Addressed in [521641](https://github.com/habitat-sh/habitat/pull/7710/commits/5216411dae740b5184c0e9efbb6aab96d8d50b67)

PR #7711, opened 2020-05-26
Add a signature policy file for Buildah exports

christophermaier:
Buildah normally requires access to `/etc/containers/policy.json`, a
configuration file that describes signature policies for deciding
whether to use a particular image during a build. (See
https://www.mankier.com/5/containers-policy.json for additional
background.) This is not a path that we have control over in a Habitat
context, though.

To get around this, we invoke `buildah` with the `--signature-policy`
option. Concretely, the engine creates and manages a temporary file
that contains a permissive default policy, and our build command uses
that file for policy decisions. This is the same policy you would get
if you installed `buildah` or `podman` from an RPM or `deb` file.

This PR also continues the efforts at error modularization within the
container exporter by confining the new Buildah-specific errors to the
buildah module, and explicitly mapping general IO errors to a more
context-carrying `SignaturePolicyError` error instance (rather than
implementing a blanket `From<std::io::Error>`). On the other hand, we
now have a general `EngineError::EngineSpecificError` variant that
simply serves to integrate any engine-specific errors into the overall
error structure of the crate. Here, a `From<BuildahError> for
EngineError` implementation makes a bit more sense.

Also fixes the end-to-end test that should have caught the fact that
the Buildah engine wasn't working in the absence of
`/etc/containers/policy.json`. That, plus the fact that I had the file
on my workstation by virtue of my previous vetting of `buildah` and
subsequent implementation of the feature, is why this snuck through
undetected until now. Mea culpa.

### DIFF
--- a/test/end-to-end/test_container_exporter.ps1
+++ b/test/end-to-end/test_container_exporter.ps1
@@ -167,8 +167,8 @@ if ($IsLinux) {
     Describe "hab pkg export container --engine=buildah" {
         It "Runs successfully" {
             $tag = New-CustomTag
-            hab pkg export container core/nginx --engine=buildah --tag-custom="$tag"
-            hab pkg exec core/buildah buildah rmi "core/nginx:$tag"
+            Invoke-NativeCommand hab pkg export container core/nginx --engine=buildah --tag-custom="$tag"
+            Invoke-NativeCommand hab pkg exec core/buildah buildah rmi "core/nginx:$tag"
         }
     }
 }
--- a/components/pkg-export-container/defaults/containers-policy.json
+++ b/components/pkg-export-container/defaults/containers-policy.json
@@ -1,0 +1,7 @@
+{
+    "default": [
+        {
+            "type": "insecureAcceptAnything"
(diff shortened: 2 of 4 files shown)


test/end-to-end/test_container_exporter.ps1
```diff
@@ -167,8 +167,8 @@ if ($IsLinux) {
     Describe "hab pkg export container --engine=buildah" {
         It "Runs successfully" {
             $tag = New-CustomTag
-            hab pkg export container core/nginx --engine=buildah --tag-custom="$tag"
-            hab pkg exec core/buildah buildah rmi "core/nginx:$tag"
+            Invoke-NativeCommand hab pkg export container core/nginx --engine=buildah --tag-custom="$tag"
+            Invoke-NativeCommand pkg exec core/buildah buildah rmi "core/nginx:$tag"
```
jsirex, line 171: Typo?

christophermaier: ooh, good catch!

PR #7717, opened 2020-05-28
adding documentation on the HAB_SUP_UPDATE_MS environment variable

jeremymv2:

### DIFF
--- a/components/habitat-chef-io/content/habitat/environment-variables.md
+++ b/components/habitat-chef-io/content/habitat/environment-variables.md
@@ -41,7 +41,8 @@ This is a list of all environment variables that can be used to modify the opera
 | `HAB_STUDIO_NOSTUDIORC` | build system | no default | When set to a non-empty value, a `.studiorc` will not be sourced when entering an interactive Studio via `hab studio enter`. |
 | `HAB_STUDIO_SUP` | build system | no default | Used to customize the arguments passed to an automatically launched Supervisor, or to disable the automatic launching by setting it to `false`, `no`, or `0`. |
 | `HAB_GLYPH_STYLE` | build system | `full` (`limited` on Windows) | Used to customize the rendering of unicode glyphs in UI messages. Valid values are `full`, `limited`, or `ascii`. |
-| `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency of milliseconds to check for updates when running with an [update strategy](/docs/using-habitat#using-updates) |
+| `HAB_SUP_UPDATE_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for Supervisor updates when running with an [update strategy](/docs/using-habitat#using-updates) |
+| `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for Service updates when running with an [update strategy](/docs/using-habitat#using-updates) |
 | `HAB_USER` | Supervisor | no default | User key to use when running with [service group encryption](/docs/using-habitat#using-encryption) |
 | `http_proxy` | build system, Supervisor | no default | A URL for a local HTTP proxy server optionally supporting basic authentication |
 | `https_proxy` | build system, Supervisor | no default | A URL for a local HTTPS proxy server optionally supporting basic authentication |
--- a/components/habitat-chef-io/content/habitat/environment-variables.md
+++ b/components/habitat-chef-io/content/habitat/environment-variables.md
@@ -43,6 +43,7 @@ This is a list of all environment variables that can be used to modify the opera
 | `HAB_GLYPH_STYLE` | build system | `full` (`limited` on Windows) | Used to customize the rendering of unicode glyphs in UI messages. Valid values are `full`, `limited`, or `ascii`. |
 | `HAB_SUP_UPDATE_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for Supervisor updates when running with the [--auto-update](/docs/habitat-cli/#hab-sup-run) flag. |
 | `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for service updates when running with an [update strategy](/docs/using-habitat#using-updates) |
+| `HAB_UPDATE_STRATEGY_FREQUENCY_BYPASS_CHECK` | Supervisor | no default | If set to `1`, will enable the Supervisor to utilize the `HAB_UPDATE_STRATEGY_FREQUENCY_MS` variable value from the line above for service updates. |
 | `HAB_USER` | Supervisor | no default | User key to use when running with [service group encryption](/docs/using-habitat#using-encryption) |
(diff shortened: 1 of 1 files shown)


components/habitat-chef-io/content/habitat/environment-variables.md
```diff
@@ -41,7 +41,8 @@ This is a list of all environment variables that can be used to modify the opera
 | `HAB_STUDIO_NOSTUDIORC` | build system | no default | When set to a non-empty value, a `.studiorc` will not be sourced when entering an interactive Studio via `hab studio enter`. |
 | `HAB_STUDIO_SUP` | build system | no default | Used to customize the arguments passed to an automatically launched Supervisor, or to disable the automatic launching by setting it to `false`, `no`, or `0`. |
 | `HAB_GLYPH_STYLE` | build system | `full` (`limited` on Windows) | Used to customize the rendering of unicode glyphs in UI messages. Valid values are `full`, `limited`, or `ascii`. |
-| `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency of milliseconds to check for updates when running with an [update strategy](/docs/using-habitat#using-updates) |
+| `HAB_SUP_UPDATE_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for Supervisor updates when running with an [update strategy](/docs/using-habitat#using-updates) |
+| `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for Service updates when running with an [update strategy](/docs/using-habitat#using-updates) |
```
kagarmoe, line 45:
```suggestion
| `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for service updates when running with an [update strategy](/docs/using-habitat#using-updates) |
```

components/habitat-chef-io/content/habitat/environment-variables.md
```diff
@@ -41,7 +41,8 @@ This is a list of all environment variables that can be used to modify the opera
 | `HAB_STUDIO_NOSTUDIORC` | build system | no default | When set to a non-empty value, a `.studiorc` will not be sourced when entering an interactive Studio via `hab studio enter`. |
 | `HAB_STUDIO_SUP` | build system | no default | Used to customize the arguments passed to an automatically launched Supervisor, or to disable the automatic launching by setting it to `false`, `no`, or `0`. |
 | `HAB_GLYPH_STYLE` | build system | `full` (`limited` on Windows) | Used to customize the rendering of unicode glyphs in UI messages. Valid values are `full`, `limited`, or `ascii`. |
-| `HAB_UPDATE_STRATEGY_FREQUENCY_MS` | Supervisor | 60000 | Frequency of milliseconds to check for updates when running with an [update strategy](/docs/using-habitat#using-updates) |
+| `HAB_SUP_UPDATE_MS` | Supervisor | 60000 | Frequency in milliseconds governing how often to check for Supervisor updates when running with an [update strategy](/docs/using-habitat#using-updates) |
```
davidMcneil, line 44: > "when running with an update strategy"

I dont believe this setting is connected to an update strategy instead it is connected to running with `auto-update`

christophermaier: Yikes, yes, well-spotted

davidMcneil:
I am currently working on making these settings configurable through the CLI (#7727). I think the naming and units of them may change.

components/habitat-chef-io/content/habitat/environment-variables.md
jeremymv2: @davidMcneil @christophermaier Well spotted indeed. I handled in https://github.com/habitat-sh/habitat/pull/7717/commits/a4ca1bcde50925aaa02fbd8f2fecf8a0bd5624d4

jeremymv2:
@davidMcneil 

> I am currently working on making these settings configurable through the CLI (#7727). I think the naming and units of them may change.

If it's ok with you, could we still get this one in first please? That way I can close out the card from the previous sprint :D

davidMcneil:
> If it's ok with you, could we still get this one in first please?

@jeremymv2 yes, absolutely! Just wanted to give a heads up to future work.

components/habitat-chef-io/content/habitat/environment-variables.md
jeremymv2: Thanks @kagarmoe. I've handled in https://github.com/habitat-sh/habitat/commit/a4ca1bcde50925aaa02fbd8f2fecf8a0bd5624d4

christophermaier:
Good catch @markan... at some point I was referring to this as "period" rather than "frequency" (in my head, at least :sweat_smile:)... "interval" is cool, too. "Frequency" is definitely wrong, though. Addressing that in the documentation first is probably the easiest first step to fixing it overall.

davidMcneil:
Just wanted to give a heads up that the environment variables that this is documenting were removed in #7733. Do we want environment variables for these settings?

jeremymv2:
> Do we want environment variables for these settings?

I think there is a potential that the removal will surprise some users whom are reliant on the variables to stabilize their environments and whom update the Supervisor not knowing that the environment variables will no longer be used. Those users will then have to start troubleshooting any behavioral changes that may result.

davidMcneil:
> I think there is a potential that the removal will surprise some users

Yes, this could be and if that is the case they should be added back in.

Because those environment variables were explicitly undocumented, I was under the impression they were not intended to be set by users. But this could be a completely wrong assumption and users could certainly still be relying on the settings.

I will put together a PR to add the environment variables back in.

jeremymv2:
> Because those environment variables were explicitly undocumented, 

Interestingly, only `HAB_UPDATE_STRATEGY_FREQUENCY_MS` was documented, which by itself, without `HAB_UPDATE_STRATEGY_FREQUENCY_BYPASS_CHECK` wouldn't be utilized 🤦  I do know, however, that these variables have been around in slack channels and therefore are being used by at least one very large on-premisis deployment.

davidMcneil:
>  I do know, however, that these variables have been around in slack channels and therefore are being used by at least one very large on-premisis deployment.

Bummer, seems like we need to address this in some way then.

#7748 adds the environment variables back in, but unfortunately, it just introduces more questions. The units for the env vars were milliseconds, but the cli and config file settings take seconds. Should we change the cli and config units? The env vars are poorly named can we change them? Obviously this would break backward compatibility, but if we made sure to call it out in the changelog would that be acceptable?

davidMcneil:
Sorry to keep adding to this, but could we also add that these environment variables are deprecated and users should prefer using `--service-update-period` and `--auto-update-period `?

jeremymv2:
@davidMcneil https://github.com/habitat-sh/habitat/pull/7717/commits/5f30823f890257a0766c366baee52875b49c6918 has a deprecation notice.

jeremymv2:
Hi @christophermaier. Github is indicating that merging of this PR is blocked until some requested changes are addressed. I think it's referring to a few changes which you requested that have since been addressed 😄

christophermaier:
@jeremymv2 Sorry! Feel free to merge whenever you like.

PR #7724, opened 2020-05-29
read hook stdout/err allowing for lossy utf8 strings

### LINKED ISSUE #7705: launcher panics on non-UTF8 characters
We had a customer report (and I am pretty sure I ran into this recently somewhere else):

```
2020-05-20 12:45:24,194 - thread 'package.develop-out' panicked at 'called `Result::unwrap()` on an `Err` value: Custom {
 kind: InvalidData, error: "stream did not contain valid UTF-8" }', src\libcore\result.rs:1188:5
2020-05-20 12:45:24,194 - stack backtrace:
2020-05-20 12:45:24,195 -    0: <unknown>
2020-05-20 12:45:24,195 -    1: <unknown>
2020-05-20 12:45:24,195 -    2: <unknown>
```

Seems like both the stdout and stderr [handlers](https://github.com/habitat-sh/habitat/blob/c8a128162feed82ea8859f7ade4a7ecdaa9707ce/components/launcher/src/service.rs#L77) in the launcher should not be using `unwrap` here.

mwrock:
fixes #7705

### DIFF
--- a/components/core/src/util.rs
+++ b/components/core/src/util.rs
@@ -6,7 +6,9 @@ pub mod sys;
 #[cfg(windows)]
 pub mod win_perm;
 
-use std::mem;
+use std::{io::{self,
+               BufRead},
+          mem};
 
 /// Same as `Result::ok()`, but logs the error case. Useful for
 /// ignoring error cases, while still leaving a paper trail.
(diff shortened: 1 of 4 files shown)


components/core/src/util.rs
```diff
@@ -86,6 +90,30 @@ pub fn pwsh_args(command: &str) -> Vec<&str> {
...
 
+/// This allows you to read a line from a buffer. It is similar to BufRead's
+/// new_line except instead of erroring on a non utf-8 bytes, it will replace
+/// those characters with U+FFFD REPLACEMENT CHARACTER, which looks like this: �
+/// using string's from_utf8_lossy. This uses BufReader's read_until which will
+/// read the buffer until reaching a new line char or an EOF. If succesful, this
+/// returns a utf8 lossy String with trailing line ending and carriage return
+/// stripped. If there are 0 bytes left, it returns None. If read_until throws
+/// an error, this will bubble that same exception inside a CannotReadLine error.
+pub fn read_lossy_line<T>(reader: &mut T) -> Result<Option<String>>
```
davidMcneil, line 101: Would it be worth making a `BufReadLossy` trait so we could essentially keep the same api as `BufRead`?

mwrock: Forgive my rust "rustyness" but if I create the trait, who would implement it. Is it possible to make a struct that you do not "own" implement the trait? In this case `ChildStdout` and `ChildStdout` would need to implement it.

davidMcneil: Yeah, I think you could do a blanket `impl` for all types that `impl` `BufRead`. Something like:

```
use std::io::{self, BufRead, BufReader};

// This is copied from [here](https://github.com/rust-lang/rust/blob/d3cba254e464303a6495942f3a831c2bbd7f1768/src/libstd/io/mod.rs#L2495),
// but converted into a "lossy" version
#[derive(Debug)]
pub struct LossyLines<B> {
    buf: B,
}

impl<B: BufReadLossy> Iterator for LossyLines<B> {
    type Item = io::Result<String>;

    fn next(&mut self) -> Option<io::Result<String>> {
        let mut buf = String::new();
        match self.buf.read_line_lossy(&mut buf) {
            Ok(0) => None,
            Ok(_n) => {
                if buf.ends_with('\n') {
                    buf.pop();
                    if buf.ends_with('\r') {
                        buf.pop();
                    }
                }
                Some(Ok(buf))
            }
            Err(e) => Some(Err(e)),
        }
    }
}

// A lossy way to read lines
pub trait BufReadLossy: BufRead {
    fn read_line_lossy(&mut self, buf: &mut String) -> io::Result<usize> {
        let mut buffer = Vec::new();
        let size = self.read_until(b'\n', &mut buffer)?;
        let s = String::from_utf8_lossy(&buffer);
        buf.push_str(&s);
        Ok(size)
    }

    fn lines_lossy(self) -> LossyLines<Self>
    where
        Self: Sized,
    {
        LossyLines { buf: self }
    }
}

// Implement `BufReadLossy` for all types that implement `BufRead`
impl<T: BufRead> BufReadLossy for T {}

fn main() {
    let mut buf = "line 1\r\nline « 2\r\nline 3\r\n".as_bytes().to_vec();
        // the char « is [194, 171] in utf8
        // we will swap in 171 which is it's pure ASCII value
        buf.push(174);
        let idx1 = buf.iter().position(|&x| x == 194).expect("no 194");
        buf.remove(idx1);
        let idx2 = buf.iter().position(|&x| x == 171).expect("no 171");
        buf.swap_remove(idx2);

        let reader = BufReader::new(buf.as_slice());
        let mut str_vec = Vec::new();
        for line in reader.lines_lossy() {
            let line = line.unwrap();
            str_vec.push(line);
        }

        assert_eq!(str_vec.len(), 3);
        assert_eq!(str_vec[0], "line 1");
        assert_eq!(str_vec[1], "line � 2");
        assert_eq!(str_vec[2], "line 3");
}
```

I dont know if this is worth it, but it gives `read_line_lossy` and `lines_lossy` methods with the same api as there non lossy versions (ie we can keep using a `for` loop and dont need to `match`).

mwrock:
ah `impl<T: BufRead> BufReadLossy for T {}` was the answer to my ignorance. This is MUCH better thanks!

PR #7728, opened 2020-06-01
Update rust to 1.43.1

davidMcneil:


davidMcneil:
@christophermaier sorry, I should have called this out in the comments. There is a new Clippy lint for redundant use statements. Everything added to a `Cargo.toml` is available in the top level namespace so there is no need to `use` it again.

christophermaier:
@davidMcneil Cool, thanks for the explanation 👍

PR #7733, opened 2020-06-03
Supervisor update config

### LINKED ISSUE #7723: Smooth out Supervisor checkins with splay
In order to allow users to more easily scale our their Habitat deployments with an OnPrem Builder, it would probably be useful to introduce a new Supervisor configuration that allowed the user to smooth out Supervisor checkins to the Builder API. The idea is very similar to the `splay` configurable for the Chef Infra Client that has proven to be very beneficial to temper HTTP checkin storms.

The `splay` would theoretically function as a random delay after either `HAB_SUP_UPDATE_MS` and/or `HAB_UPDATE_STRATEGY_FREQUENCY_MS` had expired. There may need to be an individual `splay` for each of those.

davidMcneil:
Resolves #7723 

This PR adds `--auto-update-period` ~~and `--auto-update-splay`~~ options to give control over the Supervisor update behavior. It also remove the `HAB_SUP_UPDATE_MS` environment variable which is a breaking change, but I think it is acceptable in this case because it is an undocumented internal envar.

The trickiest part of this PR was settling on the ui/ux. There are a lot of possibilities. Some possible changes:

* Currently, the units are in seconds with the value represented as a ~~`f64` allowing fractional seconds~~ `u64`. Should we allow fractional seconds?
* We could have a single flag `--auto-update` that takes the period as an optional value, but this leads to complications with the config file
* I completely removed the ability to set this value with an environment variable. Was this the right choice? How do we decide which options should have an environment variable?
* ~~We could have a special syntax for specifying period and splay in a single option. Something like `60+5` (ie 60 second period with 5 seconds of splay). This seemed too obscure.~~
* ~~Represent the splay as a percentage of the period~~
* Allow the user to specify splay

I tried to favor simplicity and explicitness in the implementation I chose but would love to hear other thoughts.

**[Update]** Thanks to [this comment](https://github.com/habitat-sh/habitat/pull/7733#pullrequestreview-423874201) this PR has been drastically simplified. Instead of allowing user-defined splay with `--auto-update-splay`, we add an offset to the start of the update checks by waiting a random amount of time between 0 and the configured `--auto-update-period`.

**[Update]** This PR adds the `--service-update-period` option to give control over the Supervisor update behavior. The `HAB_UPDATE_STRATEGY_FREQUENCY_BYPASS_CHECK` and `HAB_UPDATE_STRATEGY_FREQUENCY_MS` have been removed. Random splay between 0 and the configured `--service-update-period` has been added to the package update worker startup.

### DIFF
--- a/components/sup/src/manager/self_updater.rs
+++ b/components/sup/src/manager/self_updater.rs
@@ -77,11 +77,10 @@ impl SelfUpdater {
                      update_url,
                      update_channel,
                      period, } = runner;
-        let splay =
-            Duration::from_secs_f64(rand::thread_rng().gen_range(0.0, period.as_secs_f64()));
+        let splay = Duration::from_secs(rand::thread_rng().gen_range(0, period.as_secs()));
         debug!("Starting self updater with current package {} in {}s",
                current,
-               splay.as_secs_f64());
+               splay.as_secs());
         tokiotime::delay_for(splay).await;
         loop {
             match util::pkg::install_no_ui(&update_url, &install_source, &update_channel).await {
@@ -99,7 +98,7 @@ impl SelfUpdater {
                     warn!("Self updater failed to get latest, {}", err);
                 }
             }
-            trace!("Self updater delaying for {}s", period.as_secs_f64());
+            trace!("Self updater delaying for {}s", period.as_secs());
             tokiotime::delay_for(period).await;
         }
     }
--- a/components/sup/src/manager/self_updater.rs
+++ b/components/sup/src/manager/self_updater.rs
@@ -5,9 +5,10 @@ use crate::util;
 use habitat_common::command::package::install::InstallSource;
 use habitat_core::{package::{PackageIdent,
                              PackageInstall},
-                   util::DurationWithSplay,
                    ChannelIdent};
-use std::borrow::Borrow;
+use rand::Rng;
+use std::{borrow::Borrow,
+          time::Duration};
 use tokio::{self,
             sync::oneshot::{self,
                             error::TryRecvError,
@@ -22,7 +23,7 @@ pub struct SelfUpdater {
     current:        PackageIdent,
     update_url:     String,
     update_channel: ChannelIdent,
-    period:         DurationWithSplay,
+    period:         Duration,
 }
 
 /// The subset of data from `SelfUpdater` needed to spawn the updater task.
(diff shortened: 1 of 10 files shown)


components/sup/src/manager/self_updater.rs
```diff
@@ -15,52 +16,67 @@ use tokio::{self,
...
+    pub fn new(current: &PackageIdent,
+               update_url: String,
+               update_channel: ChannelIdent,
+               period: DurationWithSplay)
+               -> Self {
+        let runner = Runner { current: current.clone(),
+                              update_url: update_url.clone(),
+                              update_channel: update_channel.clone(),
+                              period };
+        let rx = Self::init(runner);
```
jeremymv2, line 56: I wonder if, now that we're exposing the auto-update-period tunable, we should incorporate a minimum allowed value of the current default 60s? Similar to how the [service_updater](https://github.com/habitat-sh/habitat/blob/119b59269008109138545eec32b22f1abe2e2a4c/components/sup/src/manager/service_updater/package_update_worker.rs#L28) has a minimum. Otherwise, it could potentially be somewhat misused and I don't think I've seen any demand for decreasing the value from 60.

davidMcneil: The use cases I have heard sited for very short periods is demos and testing. @ericcalabretta has sited those as motivation for not having a minimum. While I agree there is probably not a "real" use case for periods shorter than 60s, is it worth adding another flag that allows overriding the minimum?

davidMcneil:
> Having the splay specified as a percentage of the update time seems less error prone

@markan yeah, I was torn between allowing precise control and error proneness. You've convinced me that the percentage is better. Thanks!

 > One way to implement this is to have a one time offset of up to the full period applied at startup; then little or no splay is needed. However that might be a pain to retrofit.

This sounds like the better solution and should be straightforward to implement. :thinking: Maybe splay should not even be a public setting. Instead we just always use 25% (or whatever percentage) of the period.

components/sup/src/manager/self_updater.rs
jeremymv2: I understand the desire for a shorter period in testing and demos and therefore I retract my reservations. The additional override flag seems like just another variable to have to account for and document and won't really change any behaviors if a user is intent on having a lower interval across their fleet. We'll just have to adapt if this ends up impacting the req/s rate coming into the public Builder API.

Issue #7737, opened 2020-06-05
Docker studio breaks key ownership


themightychris:
For a new user trying to get straight into studio, you can end up with commands crashing because your user can't access keys downloaded by studios.

Two issues surfaced here:

- `hab origin key export` crashing because it can't read everything it scans, even though it can read the key I'm asking for
- the studio leaves challenging permissions on the keys it downloads. Leaving .pubs globally readable and not writable at all seems like it would be an improvement

```console
$ ls -l ~/.hab
ls: cannot access '/home/chris/.hab': No such file or directory

$ hab --version
hab 1.6.39/20200527162613

$ hab setup

Habitat CLI Setup
=================
Connect to an on-premises Builder instance? No 
Set up a default origin? Yes
Default origin name: chris
Create an origin key for `chris'? Yes
» Generating origin key for chris
★ Generated origin key pair chris-20200605013300.
Set up a default Builder personal access token? No 
Set up a default Habitat Supervisor control gateway secret? No

$ ls -l ~/.hab/cache/keys/
total 8
-r-------- 1 chris chris  76 Jun  4 21:33 chris-20200605013300.pub
-r-------- 1 chris chris 120 Jun  4 21:33 chris-20200605013300.sig.key

$ hab origin key export --type secret chris
SIG-SEC-1
chris-20200605013411

ltGEJs4evpscXXYBjwD7fuv6tAESasr4Ku0cc1Lv9GIQBTinGJn4Z8wFoyCYWBM4LvvqURHCeQXoto9gNFzE0A==chris-desktop:~

$ HAB_DOCKER_OPTS="-d" hab studio enter -D
3913d908eca269cc076f2b422723b9320010f38c5b3f4214d0d52b626f7ffc3f

$ ls -l ~/.hab/cache/keys/
total 16
-r-------- 1 chris chris  76 Jun  4 21:33 chris-20200605013300.pub
-r-------- 1 chris chris 120 Jun  4 21:33 chris-20200605013300.sig.key
-rw------- 1 root  root   75 Jun  4 21:33 core-20180119235000.pub
-rw------- 1 root  root   75 Jun  4 21:33 core-20200219002627.pub

$ hab origin key export --type secret chris
✗✗✗
✗✗✗ Permission denied (os error 13)
✗✗✗
```

stevendanna:
Related: https://github.com/habitat-sh/habitat/issues/7269

stevendanna:
[...]

PR #7739, opened 2020-06-05
 Explicit permissions on keys 

stevendanna:
I don't think this will solve all of #7737, but I don't think it can
hurt.

I haven't touched the Windows side of the equation, but am happy to
update it if anyone has opinions on what we should do there.

### DIFF
--- a/components/builder-api-client/src/builder.rs
+++ b/components/builder-api-client/src/builder.rs
@@ -19,7 +19,9 @@ use futures::stream::TryStreamExt;
 use habitat_core::{crypto::keys::box_key_pair::WrappedSealedBox,
                    fs::{AtomicWriter,
                         Permissions,
-                        DEFAULT_CACHED_ARTIFACT_PERMISSIONS},
+                        DEFAULT_CACHED_ARTIFACT_PERMISSIONS,
+                        DEFAULT_PUBLIC_KEY_PERMISSIONS,
+                        DEFAULT_SECRET_KEY_PERMISSIONS},
                    package::{Identifiable,
                              PackageArchive,
                              PackageIdent,
@@ -464,7 +466,7 @@ impl BuilderAPIClient {
                           .get(&format!("depot/origins/{}/encryption_key", origin)),
                       dst_path.as_ref(),
                       Some(token),
-                      Permissions::Standard,
+                      DEFAULT_PUBLIC_KEY_PERMISSIONS,
                       progress)
             .await
     }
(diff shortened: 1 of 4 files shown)


components/builder-api-client/src/builder.rs
```diff
@@ -464,7 +466,7 @@ impl BuilderAPIClient {
                           .get(&format!("depot/origins/{}/encryption_key", origin)),
                       dst_path.as_ref(),
                       Some(token),
-                      Permissions::Standard,
+                      DEFAULT_PUBLIC_KEY_PERMISSIONS,
```
stevendanna, line 469: Is this one right?

christophermaier: It seems like this would be fine. Previously, the permissions would have been set to whatever the low-level calls set them to by default, but now they'd be `444` (on Linux, anyway; unchanged on Windows).

stevendanna: Yeah, so my main question here really is: What is encryption_key? Is it a private or public bit of information?  If private, we likely don't want read for other.

components/core/src/fs.rs
```diff
@@ -66,6 +66,20 @@ pub const DEFAULT_CACHED_ARTIFACT_PERMISSIONS: Permissions = Permissions::Explic
...
+/// Permissions applied to downloaded secret keys.
+#[cfg(not(windows))]
+pub const DEFAULT_SECRET_KEY_PERMISSIONS: Permissions = Permissions::Explicit(0o400);
+/// Permissions applied to downloaded secret keys.
+#[cfg(windows)]
+pub const DEFAULT_SECRET_KEY_PERMISSIONS: Permissions = Permissions::Standard;
+
 /// An `Option`-like abstraction over platform-specific ways to model
 /// file permissions.
 pub enum Permissions {
```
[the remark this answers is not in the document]
stevendanna: This seems like a reasonable refactoring to me but it is unlikely I'll be able to get to it this week.

components/builder-api-client/src/builder.rs
christophermaier: Based on the function's name, it looks like public information, so I think we're good here.

components/core/src/fs.rs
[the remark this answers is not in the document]
christophermaier: That would be a cool addition at some point.

christophermaier:
@stevendanna is there anything else we're waiting on with this PR?

stevendanna:
@christophermaier I think we are good to go.

christophermaier:
@stevendanna Nice... thanks again for this :+1:

PR #7740, opened 2020-06-05
allow hook extentions in plan

### LINKED ISSUE #7535: Allow hook file extensions
It would be nice to allow for `.sh` or '.ps1` or whatever extension is appropriate to the hook script. This would allow editors to more easily and automatically provide syntax highlighting.

In terms of implementation, this could be a build-time only change and hook files would be added to packages without the extension.

mwrock:
closes #7535 

This allows one to have file extensions for hook files in their plan. The file is copied to the package without the extension so supervisor behavior remains unchanged. This could potentially get tricky if one has multiple extentions for a hook. For example, a `run.ps1` amd a `run.sh`. The motivation might be to handle building for different platforms from the same plan. The most explicit and best way to handle that scenario would be with package target directories. However for convenience we use this logic when there are multiple extensions for a single target:

Windows:
`.ps1` files have the highest precedence followed by an extensionless file. Any other extension is ignored since powershell is the only possible interpreter one can use on windows.

Linux:
`.sh` takes precedence.  If there is no .sh hook then it will just use whatever the last file exists for the given hook. This becomes unreliable if there were a `.ps1` and a `.py` extension for the same hook. However one can argue that this is where one should be using target folders.

### DIFF
--- a/components/plan-build-ps1/bin/hab-plan-build.ps1
+++ b/components/plan-build-ps1/bin/hab-plan-build.ps1
@@ -1211,7 +1211,18 @@ function Invoke-DefaultBuildConfig {
     }
     if (Test-Path "$PLAN_CONTEXT/hooks") {
         Write-BuildLine "Writing hooks"
-        Copy-Item "$PLAN_CONTEXT/hooks" $pkg_prefix -Recurse
+        # The supervisor does not recognize extensions so all hooks are
+        # copied without extensions
+        $hooksDir = Join-Path $pkg_prefix hooks
+        New-Item $hooksDir -ItemType Directory
+        foreach ($file in Get-ChildItem "$PLAN_CONTEXT/hooks") {
+            $singleHook = Get-ChildItem "$PLAN_CONTEXT/hooks" | Where-Object { $_.BaseName -eq $file.BaseName } | ForEach-Object { $_.Name }
+            if($singleHook -is [Array]) {
+                throw "No more than one hook file permitted per lifecycle hook. Found '$($singleHook -Join ', ')'"
+            } else {
+                Copy-Item $file (Join-Path $hooksDir $file.BaseName)
+            }
+        }
     }
     if (Test-Path "$PLAN_CONTEXT/default.toml") {
         Write-BuildLine "Writing default.toml"
@@ -1236,7 +1247,7 @@ function Invoke-BuildService {
 # Default implementation of the `Invoke-BuildService` phase.
 function Invoke-DefaultBuildService {
     Write-BuildLine "Writing service management scripts"
-    if (Test-Path "${PLAN_CONTEXT}/hooks/run") {
+    if ((Test-Path "${PLAN_CONTEXT}/hooks/run") -Or (Test-Path "${PLAN_CONTEXT}/hooks/run.ps1")) {
         Write-BuildLine "Using run hook $PLAN_CONTEXT/hooks/run"
     } else {
         if ($pkg_svc_run -ne "") {
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -1977,7 +1977,19 @@ do_default_build_config() {
   _do_copy_templates "config"
   _do_copy_templates "config_install"
   if [[ -d "$PLAN_CONTEXT/hooks" ]]; then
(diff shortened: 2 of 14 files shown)


components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1977,7 +1977,18 @@ do_default_build_config() {
...
   if [[ -d "$PLAN_CONTEXT/hooks" ]]; then
-    cp -r "$PLAN_CONTEXT/hooks" "$pkg_prefix"
+    mkdir -p "$pkg_prefix"/hooks
+    for file in "$PLAN_CONTEXT"/hooks/*
+    do
+      # If there are multiple files for the same hook but different extensions
+      # give precedense to .sh
+      # The supervisor does not recognize extensions so all hooks are
+      # copied without extensions
+      target="$pkg_prefix"/hooks/"$(basename "${file%.*}")"
```
smacfarlane, line 1987: How does the supervisor handle files in the `hooks` directory that are not one of the available lifecycle hooks? 

Ex: I could see a user wanting a `shared_functions.sh` file to share among their hooks, and the natural place to put that would be in the hooks directory.  It  looks like the supervisor only copies known hooks to the `/hab/svc/<svc_name>/hooks` directory, so we shouldn't have this case out in the wild, but it would be good to confirm that.

mwrock: Thats absolutely true. The Supervisor only copies files to the `/hab/svc/hooks` directory as part of its hooks compilation process. Only hooks defined in the `HookTable` would be eligible to be compiled.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1977,7 +1977,18 @@ do_default_build_config() {
   _do_copy_templates "config"
   _do_copy_templates "config_install"
   if [[ -d "$PLAN_CONTEXT/hooks" ]]; then
-    cp -r "$PLAN_CONTEXT/hooks" "$pkg_prefix"
+    mkdir -p "$pkg_prefix"/hooks
```
smacfarlane, line 1980: We should declare `local target file` in this function to keep those variables in the scope of this function.

components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -1211,7 +1211,19 @@ function Invoke-DefaultBuildConfig {
...
+        # The supervisor does not recognize extensions so all hooks are
+        # copied without extensions
+        $hooksDir = Join-Path $pkg_prefix hooks
+        New-Item $hooksDir -ItemType Directory
+        foreach ($file in Get-ChildItem "$PLAN_CONTEXT/hooks") {
+            if($file.Extension -eq ".ps1") {
+                Copy-Item $file (Join-Path $hooksDir $file.BaseName)
+            } elseif (!(Test-Path (Join-Path $hooksDir $file.BaseName)) -and ($file.Extension -eq "")) {
+                Copy-Item $file (Join-Path $hooksDir $file.BaseName)
+            }
```
davidMcneil, line 1225: These could be combined in a single `if` statement with an `or`, but maybe you intentionally separated them in order to make the logic easier to follow?

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -1977,7 +1977,18 @@ do_default_build_config() {
...
+    for file in "$PLAN_CONTEXT"/hooks/*
+    do
+      # If there are multiple files for the same hook but different extensions
```
smacfarlane, line 1983: Rather than guess, I wonder if we want to bail if a user specifies more than one hook?  I don't have a good idea on how to do that on the Linux side though since your hook language can be arbitrary (including .ps1).

davidMcneil: I like the idea of erroring when encountering multiple hooks. If any files in the hooks directory have the same filename sans the extension it would be an error.

This would also prevent a minor regression. If we prefer an extensioned hook (eg `ps1` on windows, and `sh` on linux) if a user had a `<hook>` and `<hook>.<sh|ps1>` files this PR would now use the extensioned version whereas the current behavior would use the non-extensioned version.

mwrock: ok this should be good to go now @smacfarlane and @davidMcneil

PR #7742, opened 2020-06-12
fix e2e failures

mwrock:
This should fix all e2e failures in the pipeline right now. There are 3 fixes included:

* Powershell 7 is appending the port to the redis-server process name similar to how `ps` behaves. I had previously patched a couple usages but this catches 2 that got away.
* There is no windows openssl package in the `dev` channel (not sure if there ever was?). This installs that package from `stable`
* It appears that the `tar-rs` crate fails when unpacking a path over 260 characters on windows. Libarchive would prepend paths with `\\?\` which forces windows APIs to treat the path as an extended-length path that can be 32k. See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file. This canonicalizes the temp_package_directory which will perform the same prepending. I have also submitted https://github.com/alexcrichton/tar-rs/pull/230 which would allow the `tar-rs` crate to do it.

### DIFF
--- a/test/end-to-end/test_service_pids_written_to_file_using_old_launcher.ps1
+++ b/test/end-to-end/test_service_pids_written_to_file_using_old_launcher.ps1
@@ -17,11 +17,11 @@ Describe "Using a Launcher that cannot provide service PIDs" {
 
     Context "Supervisor is restarted" {
         $supProc = Get-Process hab-sup
-        $redisProc = Get-Process redis-server*
+        $redisProc = Get-Process "redis-server *:6379"
         Restart-Supervisor
         Wait-Process redis-server -Timeout 10
         $newSupProc = Get-Process hab-sup
-        $newRedisProc = Get-Process redis-server*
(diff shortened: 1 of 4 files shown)


test/end-to-end/test_service_pids_written_to_file_using_old_launcher.ps1
```diff
@@ -21,7 +21,7 @@ Describe "Using a Launcher that cannot provide service PIDs" {
         Restart-Supervisor
         Wait-Process redis-server -Timeout 10
         $newSupProc = Get-Process hab-sup
-        $newRedisProc = Get-Process redis-server
+        $newRedisProc = Get-Process redis-server*
```
davidMcneil, line 24: Do we know what port redis is going to start on? Should we hardcode the value we expect?

mwrock: good call. Just added that.

PR #7748, opened 2020-06-16
Add environment variables for setting the update periods

davidMcneil:
This adds back the `HAB_SUP_UPDATE_MS` and `HAB_UPDATE_STRATEGY_FREQUENCY_MS`. However, the units are currently wrong (seconds instead of milliseconds). Should we change the cli and config file units as well? If we were not worried about backward compatibility I think these env vars would be removed? The naming of them is also unfortunate can we change the names?


davidMcneil:
Addressed in backwards compatable way by #7756

Issue #7635, opened 2020-04-22
Ability to change channel and update strategy without reloading service


mwrock:
We have some customers with an extremely large number of Supervisors where they need to roll out updates in an extremely controlled manner. Rolling updates does not really accommodate this scenario well because it can only update one node at a time and is intended to update a single clustered service and not thousands of applications installations. It makes more sense to leverage channels for phasing in updates at this scale. For instance, you may have several channels: prod-a, prod-b, etc. Then promote packages to these channels one by one where each channel has a larger number of subscribed services.

Some feedback we have gotten is that there is a need to move some services from one channel to another during different deployment cycles. For instance server group A who was in the `prod-a` channel during the last deployment and was the first to deploy, may wand to move to channel `prod-y` in the next deployment. Further, there may be a batch of very "sensitive" applications that require reboots and want to temporarily change their update strategy to `none`.

Unfortunately today changing either channel or update strategy requires a service reload which can be disruptive. What would be great is something like a `hab svc update` where one could edit some service metadata without needing to unload and reload services. Certainly some metadata would be very difficult to update while a service is running like binds or topology. However channel, update strategy and some other properties may be more feasible to update live.

davidMcneil:
Just calling this out as a specific instance of #7243. It would be nice if any solution took into account the possibility of a general approach to dynamic configuration.

christophermaier:
I was coming here to make that exact same suggestion @davidMcneil but you beat me to it :joy:

PR #7752, opened 2020-06-17
[REFACTOR] Store service spec in Service struct

christophermaier:
This is a preliminary refactoring in service of https://github.com/habitat-sh/habitat/issues/7635.

Rather than storing every individual component of a `ServiceSpec` inside a `Service` instance, we'll simply store a copy of the whole spec itself.

Not only will this help clean up and streamline the `Service` struct, it should make it easier to update service configuration on-the-fly in the future by swapping in a new `ServiceSpec`.

Each commit deals with a small isolated aspect of this change (e.g., on field per commit) in order to ease the review burden.

### DIFF
--- a/components/sup/src/event/types.rs
+++ b/components/sup/src/event/types.rs
@@ -53,7 +53,7 @@ impl Service {
         };
 
         Some(UpdateConfig { strategy: strategy.into(),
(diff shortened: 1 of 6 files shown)


davidMcneil:
Much improved!

PR #7761, opened 2020-06-22
[CI] Simplify PKG_CONFIG_PATH for cargo_update.sh

christophermaier:
With recent changes, the only thing needed in `PKG_CONFIG_PATH` is the
ZeroMQ entry.

(As it was, this was failing because `PKG_CONFIG_PATH` was unset in
the environment. Since we don't need to concatenate anything, we'll
just set the value directly and move on.)

### DIFF
--- a/.expeditor/scripts/finish_release/cargo_update.sh
+++ b/.expeditor/scripts/finish_release/cargo_update.sh
@@ -18,8 +18,8 @@ install_hub
 echo "--- :habicat: Installing and configuring build dependencies"
 hab pkg install core/zeromq
 
-PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$(< "$(hab pkg path core/zeromq)"/PKG_CONFIG_PATH)"
-export PKG_CONFIG_PATH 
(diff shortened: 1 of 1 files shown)


.expeditor/scripts/finish_release/cargo_update.sh
```diff
@@ -18,8 +18,8 @@ install_hub
 echo "--- :habicat: Installing and configuring build dependencies"
 hab pkg install core/zeromq
 
-PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$(< "$(hab pkg path core/zeromq)"/PKG_CONFIG_PATH)"
-export PKG_CONFIG_PATH 
+PKG_CONFIG_PATH="$(hab pkg path core/zeromq)/PKG_CONFIG_PATH"
```
smacfarlane, line 21: I  think you wan to keep the  `$(< `, otherwise the path will be set to the path of that file, rather than the contents.

christophermaier: :man_facepalming:

PR #7764, opened 2020-06-23
add uninstall hook

### LINKED ISSUE #6481: Add an uninstall hook
Now that we have `install` hooks for performing one-time machine wide installation tasks AND an `uninstall` command. It makes sense to have an `uninstall` hook that would be invoked at uninstall time to uninstall any machine state that was installed in the `install` hook. Foe example, `core/sqlserver` would uninstall its sql server instance from the OS.

mwrock:
resolves #6481 

This adds an uninstall hook to be run when uninstalling packages. The hook should be run in the event that the cli or the supervisor uninstalls a package. There are some considerations to take note of here:

* The hook is ignored is `--ignore-uninstall-hook` is passed to `hab pkg uninstall`
* The hook should ONLY fire if there are no other revisions installed for the package (more on this below)
* When uninstalling multiple packages we now ensure packages are uninstalled from oldest to newest. This way the newest uninstall hook is executed.

The uninstall hook is intended to be the inverse of the install hook and is where one would undo anything performed in the install hook. Note that install hooks are executed only once when a package is initially installed and are usually used to perform installation or setup that affects a machine's global state. For example it might be used to enable a "windows feature" or invoke a complicated software installer (like sql server). Over time one might update the package with better logic or adding a software patch. We only want to run the uninstaller if all revisions are being removed from the machine. Otherwise if an uninstall hook is run when installing an old version while a newer version remains installed, that hook could potentially corrupt the state of the existing package.


components/hab/src/command/pkg/uninstall_impl.rs
```diff
@@ -330,12 +362,58 @@ fn maybe_delete<U>(ui: &mut U,
         }
         ExecutionStrategy::Run => {
             ui.status(Status::Deleting, &ident)?;
+            if uninstall_hook_mode != UninstallHookMode::Ignore {
+                run_uninstall_hook(ui, &install).await?;
+            }
```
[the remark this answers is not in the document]
davidMcneil: I think the logic is [here](https://github.com/habitat-sh/habitat/blob/d985cf65aaaba02d52ec68cfec38cca574b5e158/components/hab/src/command/pkg/uninstall_impl.rs#L389)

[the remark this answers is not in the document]
mwrock: yeah its about 20 lines below. look for `if installed.len() > 1`

components/hab/src/command/pkg/uninstall_impl.rs
```diff
@@ -330,12 +362,58 @@ fn maybe_delete<U>(ui: &mut U,
...
             let pkg_dir = install.installed_path();
             do_clean_delete(&pkg_root_path, &pkg_dir)
         }
     }
 }
 
+async fn run_uninstall_hook<T>(ui: &mut T, package: &PackageInstall) -> Result<()>
```
davidMcneil, line 374: Maybe change the name to indicate that this *might* run the hook? Or factor out the number of versions check?

components/hab/src/command/pkg/uninstall_impl.rs
```diff
@@ -330,12 +362,58 @@ fn maybe_delete<U>(ui: &mut U,
...
+        let mut pkg = Pkg::from_install(package).await?;
+        // Only windows uses svc_password
+        if cfg!(target_os = "windows") {
+            // Install hooks do not have access to svc_passwords so
+            // we execute them under the current user account.
+            if let Some(user) = users::get_current_username() {
+                pkg.svc_user = user;
+            }
+        }
+        if !hook.run(&ident.name, &pkg, None::<&str>).unwrap_or(false) {
```
davidMcneil, line 410: Can we use `map_err` here?

components/hab/src/main.rs
```diff
@@ -15,7 +15,8 @@ use hab::{cli::{self,
           command::{self,
                     pkg::{download::{PackageSet,
                                      PackageSetFile},
-                          list::ListingType}},
+                          list::ListingType,
+                          uninstall_impl::UninstallHookMode}},
```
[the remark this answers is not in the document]
davidMcneil: Yeah, `uninstall_impl` should probably be a module of `uninstall` instead of them residing at the same level like they do now. And `uninstall` should probably `pub use` `UninstallHookMode` and `UninstallSafety`.

mwrock:
I believe I have addressed everything @christophermaier and @davidMcneil

components/hab/src/command/pkg/uninstall/uninstall_impl.rs
```diff
@@ -61,28 +84,30 @@ pub async fn uninstall_all_but_latest<U>(ui: &mut U,
...
+        ui.status(Status::Skipping, format!("Only {} packages installed", len))?;
         ui.end(format!("Uninstall of {} complete", ident))?;
         return Ok(0);
     }
-    // Reverse sort the idents so the latest occur first in the list
-    idents.sort_unstable_by(|a, b| b.by_parts_cmp(a));
-    let to_uninstall = &idents[number_latest_to_keep..];
+    // sort the idents so the latest occur last in the list
+    idents.sort_unstable_by(|a, b| a.by_parts_cmp(b));
+    let to_uninstall = &mut idents[..(len - number_latest_to_keep)];
```
christophermaier, line 103: This is functionally identical, in that the same identifiers to install are targeted, but their ordering is now oldest to most recent... is that significant? It would be nice to add a bit to the comment to say _why_ that's important.

Additionally, if the ordering is important, it would be good for `uninstall_many` to actually enforce that requirement.

[LATER EDIT]: Reading further through the code, my guess is that by ordering them in this way, the last package to be deleted is the most recent, and if that is also the last instance of that package on disk, then that means that the "most recent uninstall hook" is the one that runs. The assumption here is that the latest version of the hook should be able to "undo" anything that previous versions may have done. That makes sense, but it's a pretty subtle point that is really easy to miss in this code. Capturing that requirement both documentation-wise as well as structurally within the code (as well as in a test!) would be very helpful.

components/hab/src/command/pkg/uninstall/uninstall_impl.rs
```diff
@@ -330,12 +360,60 @@ fn maybe_delete<U>(ui: &mut U,
...
+{
+    let ident = package.ident();
+    let feature_flags = FeatureFlag::from_env(ui);
+    if let Some(ref hook) = UninstallHook::load(&ident.name,
+                                                &svc_hooks_path(ident.name.clone()),
+                                                &package.installed_path.join("hooks"),
+                                                feature_flags)
+    {
+        // We only want to run the uninstall hook if this is there are no other
+        // revisions of this package
```
christophermaier, line 383: Seems like this should actually be the documentation comment for this function, perhaps with a bit more concrete detail about why it's important.

test/end-to-end/test_pkg_uninstall_hook.ps1
```diff
@@ -0,0 +1,40 @@
...
+        hab pkg uninstall --keep-latest=0 $env:HAB_ORIGIN/uninstall-hook
+        $LASTEXITCODE | Should -Be 1
+        $log | Should -FileContentMatchExactly "uninstalling 0.3.0"
+    }
+
+    AfterEach {
+        hab pkg uninstall --keep-latest=0 $env:HAB_ORIGIN/uninstall-hook
+        Remove-Item $log -ErrorAction SilentlyContinue
+    }
+}
```
christophermaier, line 40: One potentially interesting test case would be uninstalling package X, which has a dependency on package Y. Package Y has an uninstall hook, but also multiple versions on disk: Y/1, Y/2, Y/3. The version of package X we're deleting depends concretely on Y/1. We would expect Y/1 to be deleted, but _not_ to run its uninstall hook, since there are other versions of Y present.

components/hab/src/command/pkg/uninstall/uninstall_impl.rs
mwrock: In terms of the sorting implementation. It seems like it would be best to revert the above bit to what is way and then have `uninstall_many` do a second sort. That does introduce additional cost of a second sort but in this case I don't anticipate that cost being burdensome.

mwrock:
@christophermaier your feedback should now be included in the last couple commits.

components/hab/src/command/pkg/uninstall/uninstall_impl.rs
```diff
@@ -98,9 +98,16 @@ pub async fn uninstall_all_but_latest<U>(ui: &mut U,
...
+
+    // Reverse sort the idents so the latest occur first in the list. Because we
+    // execute the uninstall hook only when the last revision of a package is
+    // being removed AND we want the most recent uninstall hook to be the one to
+    // run. Performing this sort ensures that the last package to be removed is the
+    // most recent package. This is important because a package's uninstall logic may
+    // include undoing any installation of previous packages.
+    idents.sort_unstable_by(|a, b| b.by_parts_cmp(a));
+
+    let to_uninstall = &mut idents[number_latest_to_keep..];
```
christophermaier, line 110: I think we can leave the logic that was here before in place, right? I like having the "belt and suspenders" sort in `uninstall_many`, because that really is the code that is most responsible for enforcing this uninstall behavior. This sorting here is still needed to determine what packages we need to deal with, and leaving it as it was means less churn (also, sorting an already sorted list later on in `uninstall_many` should be pretty quick).

This comment would also be best in `uninstall_many` (again, because it is the most responsible for this behavior).

mwrock: sorry @christophermaier  I'm a little confused about what you are asking. This commit does leave the logic that was originally here the same as it was. And it performs a second sort in `uninstall_many`. Are you just asking to move the comment?

christophermaier: Ugh, I'm a dork... I was looking at the diff of your commit, and not the overall PR diff. You're right. Yeah, I think just moving the comment is the only thing to do.

Sorry for the confusion! :bow: :man_facepalming:

mwrock: Moved the comment. Yeah that was totally the wrong spot for the comment. I think I left it there by accident.

components/hab/src/command/pkg/uninstall/uninstall_impl.rs
```diff
@@ -330,12 +369,70 @@ fn maybe_delete<U>(ui: &mut U,
...
+        templating::compile_for_package_install(package, feature_flags).await?;
+        let mut pkg = Pkg::from_install(package).await?;
+        // Only windows uses svc_password
+        if cfg!(target_os = "windows") {
+            // Install hooks do not have access to svc_passwords so
+            // we execute them under the current user account.
+            if let Some(user) = users::get_current_username() {
+                pkg.svc_user = user;
+            }
+        }
```
davidMcneil, line 426: I am sure this ultimately gets compiled into the same thing but we could use a conditional attribute instead of a conditional expression.

```rust
        #[cfg(target_os = "linux")]
        // Install hooks do not have access to svc_passwords so
        // we execute them under the current user account.
        if let Some(user) = users::get_current_username() {
            pkg.svc_user = user;
        }
```

Issue #7667, opened 2020-05-04
Bootstrapping services is easy to manage


davidMcneil:
Customers have provided us feedback that bootstrapping their habitat deployments is difficult.

This epic contains features that will make it easier to stand up habitat services on customer systems.

Once this epic is completed, it will be easier for folks to use similar methods to deploying Chef and InSpec clients to get Habitat deployed and running services on their machines.

Features:

Add ability to load multiple services in a single command

Load service on supervisor start up using 'hab svc load' config

Add config file support to `hab svc load`

Aha! Link: https://chef.aha.io/features/SH-104

sdelano:
Comment added by Lisa Stidham in Aha! - [View](https://chef.aha.io/comments/6923237518464720514)

[Issue 7667](https://github.com/habitat-sh/habitat/issues/7667)  
noting issue number for move to App Delivery workspace

Issue #7622, opened 2020-04-17
Add ability to load multiple services in a single command


davidMcneil:
This would be very useful for bootstrapping multiple services. Now that we have a deserializable type representing the fields for the `hab svc load` command this should be straightforward to implement. The biggest question is what the UI/UX should look like. Do we load config files from a directory, do we pass a list of files, do we have a single file with multiple service configurations?

PR #7765, opened 2020-06-24
Service bootstrap

### LINKED ISSUE #7627: Add config file support to `hab svc load` subcommand
Per the instructions outlined in #7621 verify the `hab svc load` subcommand.

davidMcneil:
Resolves #7627 #7622 #7653 #7667

This PR adds the following functionality to make bootstrapping easier:

* `hab svc load` is now configurable with a config file (ie it has `--generate-config` and `--config-files` arguments)
    * Its default config file location is `/hab/sup/default/config/svc.toml`
    * This file will be used to patch any config files specified with `--config-files`
* `hab svc bulkload` is a new command
> hab-svc-bulkload 1.6.60
> Load services using the service config files from the specified paths
> 
> The service config files are in the format generated by `hab svc load --generate-config`. The specified paths will be
> searched recursively for all files with a `.toml` extension. Service config files will be patched with the default
> values from `/hab/sup/default/ config/svc.toml`.
> 
> USAGE:
>     hab svc bulkload [OPTIONS]
> 
> FLAGS:
>     -h, --help       
>             Prints help information
> 
>     -V, --version    
>             Prints version information
> 
> 
> OPTIONS:
>         --svc-config-paths <SVC_CONFIG_PATHS>...
>             Paths to files or directories of service config files [default: /hab/sup/default/config/svc]
> 
 * The `--svc-config-paths` argument is also added to `hab sup run`. This behaves very similarly to running `hab svc bulkload` immediatly after `hab sup run`.
    * Any service config files will automatically be run on Supervisor startup
    * The default location is `/hab/sup/default/config/svc`

Any feedback or suggestions on the UI/UX of the feature is particularly valuable.

This PR needs testing which will be added when the basic design of the feature is approved.


components/hab/src/main.rs
```diff
@@ -1261,22 +1258,8 @@ async fn sub_svc_load(svc_load: SvcLoad) -> Result<()> {
...
-                        let mut svc_load = configopt::from_toml_file(path)?;
-                        // Patch the svc load with values from the default svc load
-                        default_svc_load.clone().patch_for(&mut svc_load);
-                        sub_svc_load(svc_load).await?;
-                    }
-                }
-            }
-        }
+    for svc_load in svc::svc_loads_from_paths(&svc_bulk_load.svc_config_paths)? {
+        sub_svc_load(svc_load).await?;
```
christophermaier, line 1263: If one of these loads fails for whatever reason, you could end up with some number of services that also didn't load, only because the first one didn't. Since this is really just the "send the Supervisor the command" operation, and not a synchronous "wait for the service to actually be running" operation, that's _probably_ OK, but it might be nice to indicate to the user which services' load commands got through, which one failed, and which ones didn't even get sent.

components/sup/src/main.rs
```diff
@@ -202,9 +203,17 @@ async fn sub_run_rsr_imlw_mlw_gsw_smw_rhw_msw(sup_run: SupRun,
                                               -> Result<()> {
     set_supervisor_logging_options(&sup_run);
 
+    let mut svc_load_msgs =
+        svc::svc_loads_from_paths(&sup_run.svc_config_paths)?.into_iter()
+                                                             .map(|svc_load| svc_load.into())
+                                                             .collect::<Vec<_>>();
```
christophermaier: This will work nicely the first time a supervisor is run on a machine, but every time thereafter, I think we're going to run into potential issues. Once a service has been loaded, a spec file will exist, which is what the Supervisor reads to know which services it's responsible for. Those spec files may also differ from what's in these config files (e.g., I start up a service based on a config file, then change how its configured using `hab svc load --force`).

Additionally, if you `hab svc unload`ed a previously-loaded service, but didn't also remember to remove the config file, that service would come back the next time you started the Supervisor.

I believe all this would also be in play when the Supervisor restarted after an update or `SIGHUP`, too.

I worry about the potential for conflicts, "spooky action", and impacts to running services. Having two different files to represent how a given service is supposed to run leaves a lot of space for weird behavior. Ultimately, I think I'd like to come up with a solution that unifies this config file with the spec file, formally exposes the spec file as "the way" to manage services, or some other approach that leaves us with a single source of truth for this information.

davidMcneil: Thanks! These are very useful cases to consider.

> This will work nicely the first time a supervisor is run on a machine, but every time thereafter, I think we're going to run into potential issues.

As long as the `force` option is set, I believe it will "just work".

> Once a service has been loaded, a spec file will exist, which is what the Supervisor reads to know which services it's responsible for. Those spec files may also differ from what's in these config files (e.g., I start up a service based on a config file, then change how its configured using hab svc load --force).

True. In this case we will revert back to what is in the config file on restart. I think the answer might be simply "don't do that" :sweat_smile:. If you are going to use config files you should probably always use config files and run `hab svc bulkload" to apply new changes instead of `hab svc load --force`.

> Additionally, if you hab svc unloaded a previously-loaded service, but didn't also remember to remove the config file, that service would come back the next time you started the Supervisor.

This is a tough one. Maybe we could have an option on Supervisor startup to unload any services that are running and were not explicitly found in the configuration?

> I believe all this would also be in play when the Supervisor restarted after an update or SIGHUP, too.

I think this just works as long as the `force` option is set.

I think of the config files as duplicating the functionality of `hab svc load` and not duplicating the functionality of spec files. The `hab svc load` and service config files act as two APIs for instructing the Supervisor what data to write to the spec file. The reason for two APIs existing is because there are two separate modes in which you want to interact with the Supervisor. One is very interactive and iterative (ie the cli). One is set it once and forget it (ie the service config files). I agree you can get a lot of odd behavior mixing these two modes, but maybe we can simply say "dont do that". I am certainly open to a better answer, but I am not sure what it is yet.

components/sup/src/main.rs
```diff
@@ -537,13 +482,18 @@ mod test {
...
             let sup_run = sup_run_from_cmd_str(cmd);
-            split_apart_sup_run(sup_run, no_feature_flags()).expect("Could not get split apart \
+            executor::block_on(split_apart_sup_run(sup_run, no_feature_flags()))
+                                                            .expect("Could not get split apart \
                                                                      SupRun")
                                                             .1
         }
 
+        fn service_load_from_cmd_str(cmd: &str) -> sup_proto::ctl::SvcLoad {
+            maybe_service_load_from_cmd_str(cmd).unwrap()
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [87c005](https://github.com/habitat-sh/habitat/pull/7765/commits/87c005215f3c0ac3a1d18a57a31fad3ded81b1c9)

components/hab/src/cli/hab/svc.rs
```diff
@@ -91,15 +108,23 @@ pub enum Key {
 
 lazy_static::lazy_static! {
     static ref CHANNEL_IDENT_DEFAULT: String = String::from(ChannelIdent::default().as_str());
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [87c005](https://github.com/habitat-sh/habitat/pull/7765/commits/87c005215f3c0ac3a1d18a57a31fad3ded81b1c9)

components/hab/src/main.rs
davidMcneil: Addressed in [87c005](https://github.com/habitat-sh/habitat/pull/7765/commits/87c005215f3c0ac3a1d18a57a31fad3ded81b1c9). It now tries to load all services and then presents an error message for each failed load. Thanks for pointing this out!

components/sup/src/main.rs
christophermaier: To the extent possible, I'd like to avoid having a lot of "don't do that" functionality. It's a bad user experience to have something fail or behave contrary to what you're expecting and have the response be "don't do that" (a rational response to that would be "well, why did you _let_ me do that?"). If we know that something is going to be potentially confusing in some way, we should build it in a way that's not confusing, or anticipate potential conflicts and make them impossible.

For instance, in this case, one potential avenue might be to have the Supervisor know which mode its supposed to be operating in ("config file mode" or "spec file mode") and behave appropriately. This could extend to having `hab svc load --force` calls fail with an explanatory message if the Supervisor is in "config file mode". (This is just an example for argument's sake; I'm skeptical that this concrete design would be an overall win from a UX perspective.)

components/hab/src/main.rs
```diff
@@ -1259,10 +1259,17 @@ async fn sub_svc_load(svc_load: SvcLoad) -> Result<()> {
 }
 
 async fn sub_svc_bulk_load(svc_bulk_load: SvcBulkLoad) -> Result<()> {
+    let mut errors = Vec::new();
     for svc_load in svc::svc_loads_from_paths(&svc_bulk_load.svc_config_paths)? {
-        sub_svc_load(svc_load).await?;
+        if let Err(e) = sub_svc_load(svc_load).await {
+            errors.push(e);
```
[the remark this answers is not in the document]
davidMcneil: It depends on the error. The ones I originally tested with had the idents, but you are right some do not. [c4bdbd](https://github.com/habitat-sh/habitat/pull/7765/commits/c4bdbda9961c66d34ac4554c0cab502bb67a04f2) adds the package idents to the error messages.

davidMcneil:
This feature is now behind the `HAB_FEAT_SERVICE_CONFIG_FILES` feature flag. `structopt` does not have a way to conditionally add arguments at runtime. Instead, I hide the arguments if the feature flag is not set, and ignore the values of the arguments. While not perfect, this gets us most of the guarantees we want.

components/hab/src/cli/hab/tests.rs
```diff
@@ -5,10 +5,15 @@ use clap::{App,
 use habitat_common::FeatureFlag;
 use std::str;
 
-fn no_feature_flags() -> FeatureFlag { FeatureFlag::empty() }
+fn feature_flags() -> FeatureFlag {
+    let mut f = FeatureFlag::empty();
+    f.insert(FeatureFlag::SERVICE_CONFIG_FILES);
```
christophermaier, line 10: This looks like `SERVICE_CONFIG_FILES` is somehow a "default" flag... is that intended?

davidMcneil: This is just for the cli test that compares the `structopt` implementation to the `clap` implementation. The `structopt` version always adds the `bulkload` command regardless of the feature flag. The `clap` version conditionally adds it based on the feature flag with the `maybe_subcommand` function. The `clap` version is the one that is actually used.

This is a limitation of `structopt` it is hard conditionally add subcommands or arguments at runtime because all the code is generated at compile time. The best I have come up with so far is simply hiding the arguments.

components/hab/src/cli.rs
```diff
@@ -900,6 +900,14 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
     )
 }
 
+fn maybe_subcommand<T: StructOpt>(condition: bool) -> App<'static, 'static> {
+    if condition {
+        T::clap()
+    } else {
+        App::new("")
```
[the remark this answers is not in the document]
davidMcneil: This is determined at the call site. For [example](https://github.com/habitat-sh/habitat/blob/57430a72c543bfa83c1149c2a613b1d65b628db4/components/hab/src/cli.rs#L834), this will add the `SvcBulkLoad` subcommand if the `SERVICE_CONFIG_FILES` feature flag is set. Otherwise, it adds a completely empty app. This is a bit hacky, but it was the best I could come up with. The compile-time nature of `structopt` makes it difficult to conditionally add subcommands or arguments at runtime.

components/common/src/lib.rs
```diff
@@ -86,6 +86,7 @@ bitflags::bitflags! {
         const TRIGGER_ELECTION           = 0b0010_0000_0000;
         const STRUCTOPT_CLI              = 0b0100_0000_0000;
         const NO_NAMED_PIPE_HEALTH_CHECK = 0b1000_0000_0000;
+        const SERVICE_CONFIG_FILES       = 0b0000_0001_0000;
```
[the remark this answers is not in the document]
davidMcneil: Addressed in [54dd96](https://github.com/habitat-sh/habitat/pull/7765/commits/54dd96e1ac149da9848db144664a968cf501b337)

components/hab/src/cli/hab/tests.rs
christophermaier: I meant that the function name `no_feature_flags` before clearly implied there were no feature flags being set, whereas the renamed `feature_flags` is a little less clear. The fact that there's a specific feature flag set in the implementation seems to carry a bit of a "default" flavor.

I think a more intention-revealing name (perhaps along with some documentation comments conveying what you said above) would fix things up nicely.

davidMcneil: I see, thanks! Addressed in [635ad4](https://github.com/habitat-sh/habitat/pull/7765/commits/635ad4c7934171b337862c0fb808732d561035ad ).

christophermaier: Perfect!

christophermaier:
Nice work @davidMcneil :+1:

PR #7766, opened 2020-06-24
[REFACTOR] Clean up and streamline spec reconciliation

christophermaier:
This is preliminary work to implementing #7635. It's pure refactoring, with no new functionality, so I'm submitting this "early" to ease review burden.

### DIFF
--- a/components/sup/src/manager/service/spec.rs
+++ b/components/sup/src/manager/service/spec.rs
@@ -287,23 +287,58 @@ impl ServiceSpec {
                             -> Option<ServiceOperation> {
         // TODO (CM): Invariant:
         // old.is_some() && new.is_some() => old.ident == new.ident
-        match (new, old) {
-            (Some(disk_spec
-                  @
-                  ServiceSpec { desired_state: DesiredState::Up,
-                                .. }),
-             None) => {
-                debug!("Reconciliation: '{}' queued for start", disk_spec.ident);
-                Some(ServiceOperation::Start(disk_spec))
+
+        // We need to compare the old spec to the new spec, taking
+        // into consideration the desired state of each. While we can
+        // do that via pattern matching directly, it gets a little
+        // hairy. Instead, we'll just extract the data we need into
+        // one unified match statement and go from there.
+        use DesiredState::{Down,
+                           Up};
+        match (old.as_ref().map(|o| o.desired_state),
+               old,
+               new.as_ref().map(|n| n.desired_state),
+               new)
+        {
+            // theoretically shouldn't happen, but no harm if it does.
+            (_, None, _, None)
+            // Somebody manually added a spec file that for some
+            // reason stated the service should be down. Weird, but
+            // okay....
+            | (_, None, Some(Down), Some(_))
+            // A stopped service's spec file was removed
+            | (Some(Down), Some(_), _, None)
+            // A stopped service's spec file was changed, but it is
+            // still supposed to be down. This would also likely
+            // require manual intervention.
+            | (Some(Down), Some(_), Some(Down), Some(_)) => {
+                // None of these situations require us to do anything
+                // in the way of starting, stopping, restarting, or
+                // modifying a service.
+                None
             }
 
-            (Some(disk_spec
-                  @
-                  ServiceSpec { desired_state: DesiredState::Up,
-                                .. }),
-             Some(running_spec)) => {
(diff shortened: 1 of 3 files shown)


components/sup/src/manager/service/spec.rs
```diff
@@ -266,6 +266,113 @@ impl ServiceSpec {
...
+    /// are running a service, but wish to stop it.
+    ///
+    /// Currently, it is *assumed* that both specs (when present)
+    /// refer to the same service.
+    ///
+    /// Returning `None` indicates that no operation is required.
+    pub(crate) fn reconcile(old: Option<ServiceSpec>,
+                            new: Option<ServiceSpec>)
+                            -> Option<ServiceOperation> {
+        // TODO (CM): Invariant:
```
davidMcneil, line 288: Do you still want this here?

components/sup/src/manager/service/spec.rs
```diff
@@ -266,6 +266,113 @@ impl ServiceSpec {
...
+        // old.is_some() && new.is_some() => old.ident == new.ident
+
+        // We need to compare the old spec to the new spec, taking
+        // into consideration the desired state of each. While we can
+        // do that via pattern matching directly, it gets a little
+        // hairy. Instead, we'll just extract the data we need into
+        // one unified match statement and go from there.
+        use DesiredState::{Down,
+                           Up};
+        match (old.as_ref().map(|o| o.desired_state),
```
davidMcneil, line 298: This is awesome! Trying to express this sort of logic without rust's pattern matching gives me nightmares.

components/sup/src/manager/service/spec.rs
```diff
@@ -266,6 +266,113 @@ impl ServiceSpec {
...
+            (Some(_), None, ..)
+            | (_, _, Some(_), None)
+            | (None, Some(_), ..)
+            | (_, _, None, Some(_)) => unreachable!(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[allow(clippy::large_enum_variant)]
```
davidMcneil, line 368: Just out of curiousity, do you have a guiding principle on when to allow this lint vs `Box`ing the offending variants?

components/sup/src/manager/service/spec.rs
```diff
@@ -266,6 +266,113 @@ impl ServiceSpec {
...
+                    // in-memory representation in line without having
+                    // to restart the entire service.
+                    debug!("Reconciliation: '{}' queued for restart",
+                           running_spec.ident);
+                    Some(ServiceOperation::Restart { to_stop:  running_spec,
+                                                     to_start: disk_spec, })
+                }
+            }
+
+            // These are unreachable because a None spec can't yield a
```
davidMcneil, line 357: Instead of matching on `(Option<DesiredState>, Option<ServiceSpec>, Option<DesiredState>, Option<ServiceSpec>)` could you match on `(Option<(DesiredState, ServiceSpec)>, Option<DesiredState, ServiceSpec>)` to remove thes invalid states?

christophermaier: Good call... that's a much clearer formulation! :+1:

components/sup/src/manager/service/spec.rs
christophermaier: I don't think I did back when I initially made this enum :facepalm:, but reading a bit at https://rust-lang.github.io/rust-clippy/master/#large_enum_variant shows a nice rule of thumb relating to the distribution of variants in your program. While I haven't measured, my suspicion is that the `Restart` variant (i.e., the big one) is the more common one. 

Also, these things aren't getting created all the time, so my guess is the memory concerns here are pretty narrow anyway.

components/sup/src/manager/service/spec.rs
christophermaier: No, it's not necessary, particularly with the upcoming PR I have to actually make the configuration dynamic. The way this function is called, the spec identifiers should be the same, but if they aren't, we'll still do the right thing (i.e., restart).

This was more a note to myself while working through the logic.

components/sup/src/manager/service/spec.rs
christophermaier: This was one of my favorite tricks I picked up from Erlang :smile:

christophermaier:
@davidMcneil I believe I've addressed your feedback... let me know what you think!

PR #7768, opened 2020-06-25
always promote package on upload if a user specifies a channel

### LINKED ISSUE #7704: pkg upload and pkg bulk-upload should promote all packages listed
### Describe the Enhancement
pkg upload and pkg bulk-upload only promote the packages they actually upload. Other packages listed do not get promoted. The upload command should (perhaps as an option, perhaps as the default) promote all packages named.

### Describe the need
The current behavior becomes a problem when trying to upload/promote groups of packages built outside of builder. For example, if an error occurs midway through, a rerun of the command only promotes the packages uploaded during that attempt.

There are a couple of ways this could be addressed.

1) a --promote-all flag could be added to the CLI for upload (and bulk-upload); this would directly enable the workflow.

2) There may also need to be some refinement around promotion. The API only allows promotion of one package at time. Entire channels can be promoted atomically, but not subsets of packages outside of a channel. We might want to add a process to atomically promote a list of packages.

That would change the upload workflow above. Instead, the user would upload their packages somewhere (the channel doesn't matter, could just be unstable), and no promotion would occur. Then as a separate event those packages could be promoted as an atomic group.

3) alternatively we could rethink the workflow around channels as short lived entities, (maybe 'sandbox ' channels as first class entities, that would be intended as staging areas, not long lived entities)

In any event, we should enhance this workflow, or propose an alternative method to achieve this.

jeremymv2:
closes #7704 

On an upload, we will ensure that a channel promotion always occurs for the package and its tdeps (from the archive path) if they exist in the target and if the user specified `--channel xxx` on the cli.

### DIFF
--- a/components/hab/src/command/pkg/upload.rs
+++ b/components/hab/src/command/pkg/upload.rs
@@ -73,10 +73,8 @@ pub async fn start(ui: &mut UI,
             ui.status(Status::Using,
                       format!("existing {} already on target", &ident))?;
             // Always promote to additional_release_channel if specified
-            if additional_release_channel.is_some() {
-                if let Some(channel) = additional_release_channel.clone() {
-                    promote_to_channel(ui, &api_client, (&ident, target), channel, token).await?
-                }
+            if let Some(channel) = additional_release_channel.clone() {
+                promote_to_channel(ui, &api_client, (&ident, target), channel, token).await?
             }
             Ok(())
         }
@@ -183,7 +181,7 @@ async fn upload_into_depot(ui: &mut UI,
     ui.status(Status::Uploaded, ident)?;
 
     // Promote to additional_release_channel if specified
-    if package_exists_in_target && additional_release_channel.is_some() {
+    if package_exists_in_target {
         if let Some(channel) = additional_release_channel.clone() {
             promote_to_channel(ui, api_client, (ident, target), channel, token).await?
         }
--- a/components/hab/src/command/pkg/upload.rs
+++ b/components/hab/src/command/pkg/upload.rs
@@ -74,10 +74,8 @@ pub async fn start(ui: &mut UI,
                       format!("existing {} already on target", &ident))?;
             // Always promote to additional_release_channel if specified
             if additional_release_channel.is_some() {
-                let channel = additional_release_channel.clone().unwrap();
-                match promote_to_channel(ui, &api_client, (&ident, target), channel, token).await {
-                    Ok(_) => (),
-                    Err(e) => return Err(e),
+                if let Some(channel) = additional_release_channel.clone() {
+                    promote_to_channel(ui, &api_client, (&ident, target), channel, token).await?
                 }
             }
             Ok(())
@@ -186,10 +184,8 @@ async fn upload_into_depot(ui: &mut UI,
 
     // Promote to additional_release_channel if specified
     if package_exists_in_target && additional_release_channel.is_some() {
-        let channel = additional_release_channel.clone().unwrap();
-        match promote_to_channel(ui, api_client, (ident, target), channel, token).await {
-            Ok(_) => (),
-            Err(e) => return Err(e),
+        if let Some(channel) = additional_release_channel.clone() {
+            promote_to_channel(ui, api_client, (ident, target), channel, token).await?
         }
     }
 
@@ -214,12 +210,8 @@ async fn promote_to_channel(ui: &mut UI,
         };
     }
 
-    match api_client.promote_package((ident, target), &channel, token)
-                    .await
-    {
-        Ok(_) => (),
-        Err(e) => return Err(Error::from(e)),
-    };
+    api_client.promote_package((ident, target), &channel, token)
+              .await?;
     ui.status(Status::Promoted, ident)?;
 
     Ok(())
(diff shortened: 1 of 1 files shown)


components/hab/src/command/pkg/upload.rs
```diff
@@ -173,29 +185,43 @@ async fn upload_into_depot(ui: &mut UI,
...
-            {
-                Ok(_) => (),
-                Err(api_client::Error::APIError(StatusCode::CONFLICT, _)) => (),
-                Err(e) => return Err(Error::from(e)),
-            };
+        match promote_to_channel(ui, api_client, (ident, target), channel, token).await {
+            Ok(_) => (),
+            Err(e) => return Err(e),
         }
+    }
```
davidMcneil, line 194: We can use the same simplifications from above to make this:
```rust
    if package_exists_in_target {
        if let Some(channel) = additional_release_channel.clone() {
            promote_to_channel(ui, api_client, (ident, target), channel, token).await?;
        }
    }
```

components/hab/src/command/pkg/upload.rs
```diff
@@ -173,29 +185,43 @@ async fn upload_into_depot(ui: &mut UI,
...
         };
-        ui.status(Status::Promoted, ident)?;
     }
 
+    match api_client.promote_package((ident, target), &channel, token)
+                    .await
+    {
+        Ok(_) => (),
+        Err(e) => return Err(Error::from(e)),
+    };
```
davidMcneil, line 222: `?` operator?

components/hab/src/command/pkg/upload.rs
```diff
@@ -70,13 +70,25 @@ pub async fn start(ui: &mut UI,
                     .await
     {
         Ok(_) if !force_upload => {
-            ui.status(Status::Using, format!("existing {}", &ident))?;
+            ui.status(Status::Using,
+                      format!("existing {} already on target", &ident))?;
+            // Always promote to additional_release_channel if specified
+            if additional_release_channel.is_some() {
+                let channel = additional_release_channel.clone().unwrap();
```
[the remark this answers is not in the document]
jeremymv2: 👍 handled in https://github.com/habitat-sh/habitat/pull/7768/commits/9f75668f9a777642aae2b820bab1cd280fd00898

components/hab/src/command/pkg/upload.rs
```diff
@@ -70,13 +70,25 @@ pub async fn start(ui: &mut UI,
...
+                match promote_to_channel(ui, &api_client, (&ident, target), channel, token).await {
+                    Ok(_) => (),
+                    Err(e) => return Err(e),
+                }
```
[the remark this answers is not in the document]
jeremymv2: Definitely! Addressed in https://github.com/habitat-sh/habitat/pull/7768/commits/9f75668f9a777642aae2b820bab1cd280fd00898

components/hab/src/command/pkg/upload.rs
jeremymv2: 💯 Changed in https://github.com/habitat-sh/habitat/pull/7768/commits/9f75668f9a777642aae2b820bab1cd280fd00898

components/hab/src/command/pkg/upload.rs
jeremymv2: Let's do it! Addressed in https://github.com/habitat-sh/habitat/pull/7768/commits/9f75668f9a777642aae2b820bab1cd280fd00898

christophermaier:
> On an upload, we will ensure that a channel promotion always occurs for the package and its deps if they exist in the target and if the user specified --channel xxx on the cli.

Why are a package's _dependencies_ being promoted into a channel when the package is promoted? From the package's perspective, the channels of any dependency are irrelevant since the package has its fully-qualified identifier. Promoting a dependency package could have big impacts on how other packages build, however, since *their* dependencies might inadvertently be affected.

components/hab/src/command/pkg/upload.rs
```diff
@@ -70,13 +70,23 @@ pub async fn start(ui: &mut UI,
                     .await
     {
         Ok(_) if !force_upload => {
-            ui.status(Status::Using, format!("existing {}", &ident))?;
+            ui.status(Status::Using,
+                      format!("existing {} already on target", &ident))?;
+            // Always promote to additional_release_channel if specified
+            if additional_release_channel.is_some() {
```
davidMcneil, line 76: We dont need this check the `if let Some` check below will handle this.

components/hab/src/command/pkg/upload.rs
```diff
@@ -173,29 +183,37 @@ async fn upload_into_depot(ui: &mut UI,
...
-
-        if channel != ChannelIdent::stable() && channel != ChannelIdent::unstable() {
-            match api_client.create_channel(&ident.origin, &channel, token)
-                            .await
-            {
-                Ok(_) => (),
-                Err(api_client::Error::APIError(StatusCode::CONFLICT, _)) => (),
-                Err(e) => return Err(Error::from(e)),
-            };
+    if package_exists_in_target && additional_release_channel.is_some() {
```
davidMcneil, line 186: We dont need the `&& additional_release_channel.is_some()` part

PR #7770, opened 2020-06-26
Allow dynamic changes to updater-related config for services 

### LINKED ISSUE #7635: Ability to change channel and update strategy without reloading service

christophermaier:
Previously, if you wanted to change the channel your service was pulling from for its updates, or what builder instance it was pointed to, or even whether it was updating or not, you would have to endure a complete service restart. Technically, this should not be the case, since service updates are managed completely outside of the running service process itself; it is something the Supervisor does _around_ the service, not _to_ the service.

Now, changes to the `bldr_url`, `channel`, `update_strategy`, or `update_condition` can be made to a service without incurring this restart penalty. Changing other service parameters will still result in a restart of a running service, but more of these may be made dynamic in the future.

To take advantage of this now, you will need to use `hab svc load --force`. However, be aware that this is not a "patchy" operation. If you only wish to change the channel, you must supply all other non-default service options in your load command; otherwise, default values will be used, which will most likely result in a service restart (as well as a potentially different configuration). We are looking into providing a dedicated command for modifying the configuration of running services that *will* be a patchy operation.

Fixes #7635


components/sup/src/manager/service/spec.rs
```diff
@@ -337,27 +337,94 @@ impl ServiceSpec {
...
+                        // need to be investigated more deeply.
+                        || running_spec.binds != disk_spec.binds
+                        || running_spec.binding_mode != disk_spec.binding_mode
+                        || running_spec.config_from != disk_spec.config_from
+                        // TODO (CM): This probably doesn't need to be here
+                        || running_spec.shutdown_timeout != disk_spec.shutdown_timeout
+                        || running_spec.svc_encrypted_password != disk_spec.svc_encrypted_password
+                        // TODO (CM): This probably doesn't need to be here, either
+                        || running_spec.health_check_interval != disk_spec.health_check_interval
+                    {
```
davidMcneil, line 371: Could we use pattern matching to pull out the fields on one (or both?) of the `ServiceSpec`s here? If we explicitly list every field in the pattern match, if we add a new field to `ServiceSpec` the compiler will automatically pull us to this section of the code. This will force us to evaluate how the update behavior of the field should work.

components/sup/src/manager/service/spec.rs
```diff
@@ -337,27 +337,94 @@ impl ServiceSpec {
...
+                            || running_spec.update_condition != disk_spec.update_condition
+                        {
+                            ops.insert(RefreshOperation::RestartUpdater);
+                        }
+
+                        // We should have *something* to do down
+                        // here, but if we don't, let's be explicit
+                        // about it.
+                        if ops.is_empty() {
+                            None
```
davidMcneil, line 390: Would it be worth adding a `debug_assert` and error log here? If we ever reach this it seems like something has to be wrong with our logic.

components/sup/src/manager.rs
```diff
@@ -1595,6 +1596,40 @@ impl Manager {
...
+                    if let Some(mut s) = services.remove(&ident) {
+                        s.set_spec(spec);
+
+                        for op in ops {
+                            match op {
+                                RefreshOperation::RestartUpdater => {
+                                    self.service_updater.lock().register(&s);
+                                }
+                            }
+                        }
```
davidMcneil, line 1620: Not for this PR. But I could see extracting this into an `update` function. Then on service startup, we would pass in a list of all the various updates. This would prevent us from having duplicated logic between service startup and update.

components/sup/src/manager/service/spec.rs
christophermaier: Do you have a concrete solution for that? I agree that it's a bummer that this isn't as compiler-driven as I would like, for the reasons you mention. I couldn't come up with a way to do that, though. If you've got a way to do it that I overlooked, that would be great to add!

components/sup/src/manager.rs
christophermaier: Yup, I think that's the natural next step.

components/sup/src/manager/service/spec.rs
davidMcneil: It is not perfect, but just pattern marching on all the fields of `ServiceSpec` will ensure we examine the update logic should a new field be added.

```rust
                    // Explicitly destructure every field here. This will force us to examine the
                    // update logic for any new fields added to `ServiceSpec`.
                    let ServiceSpec {
                        ident,
                        group,
                        bldr_url,
                        channel,
                        topology,
                        update_strategy,
                        update_condition,
                        binds,
                        binding_mode,
                        config_from,
                        desired_state: _,
                        shutdown_timeout,
                        svc_encrypted_password,
                        health_check_interval,
                    } = &running_spec;

                    if ident != &disk_spec.ident ...
```

christophermaier: Ah, I see what you mean... yes, we can do that for sure :+1:

components/sup/src/manager.rs
```diff
@@ -1595,6 +1596,40 @@ impl Manager {
...
+                    // Relies on spec.ident not having changed, which
+                    // ServiceSpec#reconcile must guarantee.
+                    let ident = spec.ident.clone();
+
+                    // TODO (CM): While it is technically OK to remove
+                    // the service, modify it, and then re-insert it,
+                    // this is conceptually a modify-in-place
+                    // operation. Ideally, the code would reflect
+                    // this.
+                    if let Some(mut s) = services.remove(&ident) {
```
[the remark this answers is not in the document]
christophermaier: Yup!

christophermaier:
@davidMcneil Great feedback; thanks! I think the code is much improved as a result :+1:

PR #7776, opened 2020-06-30
ensure tdeps file is built with lower level deps at the bottom of the list

### LINKED ISSUE #7749: hooks are not run in the correct order if a tdep is included as a dep in a plan
If pkg A depends on pkg B which depends on pkg C, pkg A should only need to depend on pkg B and then A, B and C install hooks should run C then B then A..  This works on windows today (broken on linux). However if pkg B and pkg C are included in pkg A's dep list, and pkg B is last in the list, pkg B's install hook will run first and fail if it requires pkg C's install hook to have ran.

mwrock:
fixes #7749 

As the TDEPS metadata file is build, lowest level dependencies should be at the bottom of the list because install hooks are reun from bottom to top. As long as each top level dep is not also included as a tdep, this works today on windows (broken on linux). However if a top level dep is also a tdep of a previously listed top level dep, the top level dep remains at the top of the list. What needs to happen instead is for the top level dep (that is also a tdep) to be removed from the top and then appended to the bottom.

Linux adds an additional bug by sorting the TDEP (and all other dependency metadata files) so all deps are listed in ident order regardless of where they fall in the dependency tree. We do not perform this sort on windows and from what I can tell, the sort has no value.


stevendanna:
Minor typo in the commit message:

```
 ensure tdeps file is built with de[ps at the bottom of the list 
```

If you get a chance, I typically also find it helpful for the detail that you have in this PR message to also be in the commit message.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -870,6 +870,19 @@ _return_or_append_to_set() {
   return 0
 }
 
+# **Internal** Appends an entry to the given array and removes any entries
+# already in the array that match the passed entry.
```
stevendanna, line 874: This function is fine based on the current call sites, but I think we might want to add more detail to the comment or change the same to indicate how to use it safely since it has some confusing behavior when used on more arbitrary input.

The opportunity for misuse is produced by

```
   dedupe_set=("${deps[@]/$1}")
```

This is doing string replacement on each member of the array, replacing any matched substring with the empty string.  This produces two problems:

1) For exact matches, it produces an empty element in the array, thus this function ends up extending the length of the array by 1 empyt element.

```
> _dedupe_and_append_to_set_debug bar bar bat baz
            deps: bar bat baz
        deps len: 3
      dedupe_set:  bat baz
  dedupe_set len: 3
    appended_set:  bat baz bar
appended_set len: 4
```

at the callsites this is OK because of our use of `read` to rebuild the array.

2) If someone where to use this on a set with the possibility of non-exact matches, you'd get very surprising results:

```
> _dedupe_and_append_to_set_debug b bar bat baz
            deps: bar bat baz
        deps len: 3
      dedupe_set: ar at az
  dedupe_set len: 3
    appended_set: ar at az b
appended_set len: 4
```

At our callsites this is currently OK because we are using this on idents that should never have substring matches.

mwrock: thanks for this clarification @stevendanna ! I was a bit nervous about this function since bash is a second shell language for me. I renamed the function to be specific to our use case and not a generic array utility.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -870,6 +870,22 @@ _return_or_append_to_set() {
...
+_add_dep_to_tdep_list() {
+  local appended_set
+  local dedupe_set
+  local deps
+  deps=("${@:2}")
+  dedupe_set=("${deps[@]/$1}")
+  appended_set=("${dedupe_set[@]}" "$1")
+  echo "${appended_set[@]}"
+  return 0
+}
```
christophermaier, line 887: In light of the subtleties that @stevendanna raised, it may be better to be more explicit here:

```bash
_add_dep_to_tdep_list() {
  local to_add="${1}"
  local deps=("${@:2}")
  local result=()

  # Explicitly filter out any instances of the to-be-added dependency
  # that we may have already seen.
  for d in "${deps[@]}"; do
      if [[ "$d" != "${to_add}" ]]; then
          result=( "${result[@]}" "$d" )
      fi
  done
  # Append the dependency to the end of the list
  result=( "${result[@]}" "$to_add" )

  echo "${result[@]}"
  return 0
}
```
It's wordier, and perhaps less efficient (though I'd be surprised to discover this being a bottleneck in the overall process of building a Habitat package). However, it's _much_ more obvious what is going on (I _always_ have to look up these variable manipulation Bash-isms), and isn't subject to the partial replacement issues the other implementation is. Additionally, if we ever get around to rewriting this in Rust, it will be much easier to translate this explicit implementation than the subtle "idiomatic" other one.

christophermaier: Incidentally, as a point of Bash style, I personally find it much easier to reason about functions when the arguments are immediately bound to descriptive variable names; `$to_add` is better than `$1`.

mwrock: I like it

mwrock: I have made this change @christophermaier and rebuilt everything and reran locally against the e2e tests which passed.

PR #7779, opened 2020-07-01
compile windows-service package version into binary and log version at start of log

### LINKED ISSUE #7736: Make it easier to find the windows-service version currently running
Currently in order to find the version of the windows-service running, one needs to find the highest `version/release` folder in `hab\pkgs\core\windows-service`. There is also no guarantee that is the actual one running if they have multiple. We should provide a `-version` flag to `HabService.exe` or log the version to the sup log. We should also make sure that the binary version stamp in the file's "Product info" is marked with this version.

mwrock:
fixes #7736 

### DIFF
--- a/components/windows-service/HabService.cs
+++ b/components/windows-service/HabService.cs
@@ -82,6 +82,8 @@ namespace HabService
                 // but not here because there is no actual console. To prevent a larger refactor
                 // and research excercise, we will just emit our glyphs in ascii.
                 Environment.SetEnvironmentVariable("HAB_GLYPH_STYLE", "ascii");
+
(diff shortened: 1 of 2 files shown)


components/windows-service/HabService.cs
```diff
@@ -82,6 +82,8 @@ protected override void OnStart(string[] args)
                 // but not here because there is no actual console. To prevent a larger refactor
                 // and research excercise, we will just emit our glyphs in ascii.
                 Environment.SetEnvironmentVariable("HAB_GLYPH_STYLE", "ascii");
+
+                log.Info(String.Format("Chef Habitat windows service {0}", Assembly.GetExecutingAssembly().GetName().Version));
```
christophermaier, line 86: Should it be "Windows Service", instead (i.e., all capitalized)?

PR #7782, opened 2020-07-07
Propagate install and uninstall hook exit codes

### LINKED ISSUE #6964: Failed install hook should bubble up exit code to hab pkg install
Today if an install hook fails with, for example, code `50000`. The cli command `hab pkg install` will exit with `1`. It would be helpful if `hab pkg install` exited with the same failed code as the install hook.

One case where this could be particularly helpful is code `3010` which an MSI exits with when a install is "successful" but a reboot is required to guarantee that running it will work. If an automated process is running `hab pkg install` it could perhaps reboot the machine if it detects `3010`.

davidMcneil:
Resolves #6964

This PR adds the logic to propagate the exit code of the `install` and `uninstall` hook to the `hab` command.


components/common/src/templating/hooks.rs
```diff
@@ -61,7 +61,7 @@ impl Default for ExitCode {
 pub trait Hook: fmt::Debug + Sized + Send {
     type ExitValue: fmt::Debug + Send;
 
-    fn file_name() -> &'static str;
+    const FILE_NAME: &'static str;
```
christophermaier, line 69: Nice :+1:

components/common/src/templating/hooks.rs
```diff
@@ -315,6 +320,61 @@ pub trait Hook: fmt::Debug + Sized + Send {
     fn stderr_log_path(&self) -> &Path;
 }
 
+#[async_trait::async_trait]
+pub trait HookExt: Hook<ExitValue = ExitStatus> + Sync {
```
christophermaier, line 324: The naming and typing of this trait doesn't clearly capture that it's for install and uninstall hooks only. It'd be nice to make that more obvious (though I admittedly am not sure what a better name for the trait and function would be).

davidMcneil: Agreed. I struggled around the naming of this. There is no reason this has to apply _only_ to the `install` and `uninstall` hooks so I left the naming very general.

The method and frequency of when the `install` and `uninstall` hooks are run may provide an axis of differentiation. What if I called the trait one of the following: `OneOffHookExt`, `AuxiliaryHookExt`, `PackageMaintenceHookExt`...?

Ultimately, I think the `Hook` trait could use some cleanup. There is a lot of duplicated functionality when calling the hooks that should be moved into the trait itself and the `handle_exit` functionality seems unnecessarily general. There is really only a few different kinds of exit types and this would probably be better handled with different run functions.

christophermaier: I agree that the `Hook` trait needs a lot of love :frowning_face: 

The idea of using method/frequency of hook execution as an axis of differentiation is a great one. Of the suggestions given, I lean toward `PackageMaintenanceHookExt` as it's more descriptive and seems to capture something core to the whole idea of having `install` and `uninstall` hooks. Making that name change, along with a nice documentation comment, would be a nice addition :+1:

davidMcneil: Thanks! Addressed in [19aa2c](https://github.com/habitat-sh/habitat/pull/7782/commits/19aa2c96080261d223409866295e7bb56427046a)

PR #7783, opened 2020-07-07
increase rumor file atomicity and durability

jeremymv2:
please read the commit for summary of analysis and reason for changes.

### DIFF
--- a/components/butterfly/src/rumor/dat_file.rs
+++ b/components/butterfly/src/rumor/dat_file.rs
@@ -30,7 +30,8 @@ use std::{collections::HashMap,
           path::{Path,
                  PathBuf}};
 
-const HEADER_VERSION: u8 = 2;
+const CURRENT_HEADER_VERSION: u8 = 2;
+const OLDEST_HEADER_VERSION: u8 = 1;
 
 // And now for a riveting discussion on version 1 vs version 2 headers in this magical file. The
 // version 1 header was a struct consisting of 6 u64 fields. It did not contain any information on
@@ -45,6 +46,15 @@ const HEADER_VERSION_2_NUM_FIELDS: usize = 7;
 const HEADER_VERSION_1_SIZE: usize = SIZE_OF_HEADER_FIELD * HEADER_VERSION_1_NUM_FIELDS;
 const HEADER_VERSION_2_SIZE: usize =
     (SIZE_OF_HEADER_FIELD * HEADER_VERSION_2_NUM_FIELDS) + SIZE_OF_HEADER_FIELD;
+
+trait WriteExt: Write {
+    fn write_all_with_size(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.write_all(buf)?;
+        Ok(buf.len())
+    }
+}
+
+impl<W: Write> WriteExt for W {}
 
 /// A versioned binary file containing rumors exchanged by the butterfly server which have
 /// been periodically persisted to disk.
@@ -210,9 +220,12 @@ impl DatFileWriter {
         w.with_writer(|mut f| {
              let mut writer = BufWriter::new(&mut f);
              let header_reserve = vec![0; HEADER_VERSION_2_SIZE];
-             writer.write(&[HEADER_VERSION])
-                   .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?;
-             writer.write(&header_reserve)
+             writer.write_all(&[CURRENT_HEADER_VERSION]).map_err(|err| {
+                                                             Error::DatFileIO(self.path()
+                                                                                  .to_path_buf(),
+                                                                              err)
+                                                         })?;
+             writer.write_all(&header_reserve)
                    .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?;
              header.insert_member_offset(self.write_member_list_mlr(&mut writer, member_list)?);
              header.insert_offset_for_rumor(Service::MESSAGE_ID,
@@ -249,8 +262,9 @@ impl DatFileWriter {
         where W: Write
     {
         let bytes = header.write_to_bytes();
-        let total = writer.write(&bytes)
-                          .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?;
+        let total =
+            writer.write_all_with_size(&bytes)
+                  .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?;
         Ok(total)
     }
 
@@ -274,10 +288,11 @@ impl DatFileWriter {
         let mut len_buf = [0; 8];
         let bytes = membership.clone().write_to_bytes().unwrap();
         LittleEndian::write_u64(&mut len_buf, bytes.len() as u64);
-        total += writer.write(&len_buf)
-                       .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?
-                 as u64;
-        total += writer.write(&bytes)
+        total +=
+            writer.write_all_with_size(&len_buf)
+                  .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?
+            as u64;
+        total += writer.write_all_with_size(&bytes)
                        .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?
                  as u64;
         Ok(total)
@@ -304,10 +319,11 @@ impl DatFileWriter {
         let mut rumor_len = [0; 8];
         let bytes = rumor.write_to_bytes().unwrap();
         LittleEndian::write_u64(&mut rumor_len, bytes.len() as u64);
-        total += writer.write(&rumor_len)
-                       .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?
-                 as u64;
-        total += writer.write(&bytes)
+        total +=
+            writer.write_all_with_size(&rumor_len)
+                  .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?
+            as u64;
+        total += writer.write_all_with_size(&bytes)
                        .map_err(|err| Error::DatFileIO(self.path().to_path_buf(), err))?
                  as u64;
         Ok(total)
@@ -322,10 +338,11 @@ impl DatFile {
               .map_err(|err| Error::DatFileIO(path.to_path_buf(), err))?;
         debug!("Header Version: {}", version[0]);
 
-        // If this has happened, it's likely that the file is corrupt
-        if version[0] > HEADER_VERSION {
-            let msg = format!("Unable to read Dat File {}: corrupt file header.",
-                              path.display());
+        // If this has happened, it's likely that the file was mutated in some unexpected way.
+        if version[0] < OLDEST_HEADER_VERSION || version[0] > CURRENT_HEADER_VERSION {
+            let msg = format!("Unable to read Dat File {}: invalid header version: {}",
+                              path.display(),
+                              version[0]);
             let err = io::Error::new(io::ErrorKind::InvalidData, msg);
             return Err(Error::DatFileIO(path.to_path_buf(), err));
         }
@@ -517,7 +534,7 @@ mod tests {
         original.insert_offset_for_rumor(Departure::MESSAGE_ID, rand::random::<u64>());
 
         let bytes = original.write_to_bytes();
-        let restored = Header::from_bytes(&bytes, HEADER_VERSION);
+        let restored = Header::from_bytes(&bytes, CURRENT_HEADER_VERSION);
         assert_eq!(bytes.len() as u64, restored.size);
         assert_eq!(original.offsets, restored.offsets);
         assert_eq!(original.version, restored.version);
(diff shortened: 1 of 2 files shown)


mwrock:
I know this is in draft still but definitely think adding the `MOVEFILE_WRITE_THROUGH` flag is the right thing to do here!

.expeditor/verify.pipeline.yml
```diff
@@ -937,4 +937,4 @@ steps:
           os-version: "10.13"
           inherit-environment-vars: true
       buildkite:
-        timeout_in_minutes: 45
+        timeout_in_minutes: 60
```
christophermaier, line 940: I had to make this exact change in on of my PRs yesterday.

I'm going to make a standalone PR for this now, and then we can all benefit :smile:

christophermaier: Aaaand here we are: #7785

jeremymv2: Thanks @christophermaier!  I've removed this change.

components/core/src/fs.rs
```diff
@@ -773,7 +790,29 @@ impl AtomicWriter {
...
+        unsafe {
+            if MoveFileExW(AtomicWriter::windows_u16s(self.tempfile.path()).as_ptr(),
+                           AtomicWriter::windows_u16s(&self.dest).as_ptr(),
+                           winapi::um::winbase::MOVEFILE_WRITE_THROUGH
+                           | winapi::um::winbase::MOVEFILE_REPLACE_EXISTING)
+               == 0
+            {
+                return Err(io::Error::last_os_error());
+            }
+        }
```
davidMcneil, line 811: This is great! I think a cross platform atomic rename could be useful in general. Would you mind extracting this code to a function? We could then export a function (maybe `atomic_rename`?) with the same signature for both windows and linux making it easy to use in the future.

Not for this PR but we at some point should probably check all the places we call `fs::rename`.

:thinking: I wonder if the rust team would accept a PR adding `MOVEFILE_WRITE_THROUGH` to their rename implementation?

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -210,9 +210,9 @@ impl DatFileWriter {
         w.with_writer(|mut f| {
              let mut writer = BufWriter::new(&mut f);
              let header_reserve = vec![0; HEADER_VERSION_2_SIZE];
-             writer.write(&[HEADER_VERSION])
+             writer.write_all(&[HEADER_VERSION])
```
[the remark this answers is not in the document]
christophermaier: Probably so, given that we currently make no attempt to recover from incomplete writes.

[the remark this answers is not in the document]
jeremymv2: @davidMcneil I had this same thought myself until I realized why it wouldn't work without a much larger overhaul. Not that it isn't a good idea, but perhaps a bigger change that we could take on in another discrete unit of work.

The current implementation relies on the `write` function's return of `Ok(usize)` variant to determine the size of data written and therefore the offset of where to begin reading the rumor on the next read iteration.  

It certainly appears to leave the door wide open for incomplete rumors written to disk though... 😱

https://github.com/habitat-sh/habitat/blob/edde4584add1d9dd5fc552f747c44bc95ffafb0f/components/butterfly/src/rumor/dat_file.rs#L157-L166



We do know the size of the rumor to be written in advance and could leverage the more reliable `write_all` to just simply write out the rumor and we store the offset as the detected rumor byte size, but like I said, that's a lot more surface area to change. I think we could break that work up.

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -322,8 +322,8 @@ impl DatFile {
               .map_err(|err| Error::DatFileIO(path.to_path_buf(), err))?;
         debug!("Header Version: {}", version[0]);
 
-        // If this has happened, it's likely that the file is corrupt
-        if version[0] > HEADER_VERSION {
+        // If this has happened, it's likely that the file was mutated in some unexpected way.
+        if version[0] == 0 || version[0] > HEADER_VERSION {
```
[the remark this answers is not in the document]
christophermaier: The answer is almost assuredly "no"... I did a bit of spelunking, and version 2 was introduced back in version 0.27.0, which is 3 years ago. If anybody is still actively using 0.27.0, they have much bigger problems than the format of this file :joy:

However, if we're going to commit to removing support for version 1, we should do that in a dedicated manner in its own PR.

In the meantime, it may be more clarifying to create a `OLDEST_VERSION = 1` constant and reframe this conditional as
```rust
if version[0] < OLDEST_VERSION || version[0] > HEADER_VERSION
```
(Changing `HEADER_VERSION` to `CURRENT_VERSION` might be a further improvement.)

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -544,7 +544,14 @@ mod tests {
...
         assert!(file_path.is_file());
+        let dat_path = file_path.clone();
         let dat_file_length = fs::metadata(file_path).map(|md| md.len());
         assert_ne!(dat_file_length.unwrap(), 0);
+
+        // Now that the dat file content was written, re-read the content back in
+        // to ensure underlying filesystem operations occurred successfully.
+        let content = DatFileReader::read(dat_path).unwrap();
+        assert_eq!(content.header.version, 2);
+        assert_eq!(content.header.size, 64);
```
christophermaier, line 555: Great addition! :+1:

components/core/src/fs.rs
```diff
@@ -4,6 +4,16 @@ use crate::util::posix_perm::{self,
...
+
+#[cfg(windows)]
+use std::os::windows::ffi::OsStrExt;
+
+#[cfg(windows)]
+use winapi::um::winbase::MoveFileExW;
+
+#[cfg(windows)]
+use std::iter;
+
```
christophermaier, line 16: It would be nice to remove all the newlines from between these use statements.

components/butterfly/src/rumor/dat_file.rs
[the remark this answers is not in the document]
jeremymv2: I also thought about doing this but then realized it might be a trap we're setting for our future selves.  If we only support the _current_ version, then how can we safely upgrade? Given that the Dat file implementation is perhaps a little brittle and not subject to change very often unless we completely abandon it for something else, I think it might be best to support all known versions (currently just 1, 2) so that we can always ensure an upgrade path exists.

components/core/src/fs.rs
jeremymv2: @davidMcneil I actually had this implementation then reverted the commit because I realized that a `atomic_rename` is effectively what we already have in the `AtomicWriter::finish` function, where I've added the platform specific bits.

https://github.com/habitat-sh/habitat/blob/edde4584add1d9dd5fc552f747c44bc95ffafb0f/components/core/src/fs.rs#L747-L781

It should generally be called via `atomic_write`:

https://github.com/habitat-sh/habitat/blob/edde4584add1d9dd5fc552f747c44bc95ffafb0f/components/core/src/fs.rs#L809-L814

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -322,8 +322,8 @@ impl DatFile {
               .map_err(|err| Error::DatFileIO(path.to_path_buf(), err))?;
         debug!("Header Version: {}", version[0]);
 
-        // If this has happened, it's likely that the file is corrupt
-        if version[0] > HEADER_VERSION {
+        // If this has happened, it's likely that the file was mutated in some unexpected way.
+        if version[0] == 0 || version[0] > HEADER_VERSION {
             let msg = format!("Unable to read Dat File {}: corrupt file header.",
                               path.display());
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/7783/commits/f355b8355962b63a29d3b51ccff0d4451ae710d9

components/core/src/fs.rs
jeremymv2: > I wonder if the rust team would accept a PR adding MOVEFILE_WRITE_THROUGH to their rename implementation?

👍 I'm definitely curious too. I'll open one up.

components/butterfly/src/rumor/dat_file.rs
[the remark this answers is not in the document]
jeremymv2: I went with

```
if version[0] < OLDEST_HEADER_VERSION || version[0] > CURRENT_HEADER_VERSION {
```

components/core/src/fs.rs
jeremymv2: Agreed. I took care of this.

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -552,7 +556,7 @@ mod tests {
         // Now that the dat file content was written, re-read the content back in
         // to ensure underlying filesystem operations occurred successfully.
         let content = DatFileReader::read(dat_path).unwrap();
-        assert_eq!(content.header.version, 2);
-        assert_eq!(content.header.size, 64);
+        assert_eq!(content.header.version, CURRENT_HEADER_VERSION);
+        assert_eq!(content.header.size, HEADER_VERSION_2_SIZE);
```
christophermaier: @jeremymv2 With the constant name changes, it might be more consistent to call this `CURRENT_HEADER_VERSION_SIZE`.

components/butterfly/src/rumor/dat_file.rs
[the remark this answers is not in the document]
davidMcneil: @jeremymv2 If we added a trait like below and replaced the calls to `write` with `write_all_with_size` would that work? 

```rust
trait WriteExt: Write {
    fn write_all_with_size(&mut self, buf: &[u8]) -> io::Result<usize> {
        self.write_all(buf)?;
        Ok(buf.len())
    }
}

impl<W: Write> WriteExt for W {}
```

components/core/src/fs.rs
davidMcneil: > It should generally be called via atomic_write:

:thinking: I think that might be slightly different? To use that to rename a file I would write something like:

```rust
pub fn rename(src: &Path, dst: &Path) -> io::Result<()> {
    let contents = fs::read_to_string(src)?;
    atomic_write(dst, contents)?;
    fs::remove_file(src)
}
```

Between each of the three steps (reading, writing, and deleting) I think there are potentials for race conditions that an `atomic_rename` would guard against.

jeremymv2: @davidMcneil most definitely a brilliant idea!

I've added a commit for it: https://github.com/habitat-sh/habitat/pull/7783/commits/04145e4d68e1af2624af4fbb68f6c7c4c66daef1

davidMcneil: @jeremymv2 sorry, I think my last comment was confusing :sweat_smile: ? I was trying to say the example `rename` functionality was **NOT** atomic because there could be race conditions between each of those expressions. The true atomic rename I think should look something like:

```rust
// `fs::rename` calls `MoveFileExW` on Windows, however the underlying implementation only
// utilizes the `MOVEFILE_REPLACE_EXISTING` flag which allows for file overwrite but no
// guarantee on durability. For this, we additionally pass `MOVEFILE_WRITE_THROUGH`.
// This causes the function to not return until the file is actually moved on the disk.
// Setting this value guarantees that a move performed as a copy and delete operation is
// flushed to disk before the function returns. The flush occurs at the end of the copy
// operation.
#[cfg(windows)]
pub fn rename_windows<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> io::Result<()> {
    // Helper function to transform a Path to a `LPCWSTR` Windows string data type. It is
    // essentially a null-terminated string of 16-bit Unicode characters.
    fn windows_u16s(s: &Path) -> Vec<u16> {
        s.as_os_str().encode_wide().chain(iter::once(0)).collect()
    }
    unsafe {
        if MoveFileExW(windows_u16s(from.as_ref()).as_ptr(),
                       windows_u16s(to.as_ref()).as_ptr(),
                       winapi::um::winbase::MOVEFILE_WRITE_THROUGH
                       | winapi::um::winbase::MOVEFILE_REPLACE_EXISTING)
           == 0
        {
            Err(io::Error::last_os_error());
        } else {
            Ok(())
        }
    }
}

/// atomic_rename is a cross platform  helper function for renaming a file atomically with
/// durability guarantees. It leverages the atomicity and durability features of atomic_write.
pub fn atomic_rename<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> io::Result<()> {
    #[cfg(windows)]
    return fs::rename_windows(from, to);
    #[cfg(not(windows))]
    return fs::rename(from, to);
}
```

components/butterfly/src/rumor/dat_file.rs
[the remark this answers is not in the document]
jeremymv2: That's a fantastic suggestion @davidMcneil. Handled in https://github.com/habitat-sh/habitat/pull/7783/commits/406a125fb7fb87eb46de16fba18df77f40525a2e

components/core/src/fs.rs
jeremymv2: It turned out to be a little tricky getting the trait bounds defined correctly because we also needed to call `AtomicWriter::sync_parent()` for `rename_unix()` so that the file changes were durable. However, I believe I've got it in https://github.com/habitat-sh/habitat/commit/406a125fb7fb87eb46de16fba18df77f40525a2e

jeremymv2:
The https://github.com/habitat-sh/habitat/pull/7783/commits/fd367566cc0e1fba19919c82030ffeac374ebc01 is just a squash.

PR #7799, opened 2020-07-13
 Write all service files when a service is loaded 

### LINKED ISSUE #4877: Adding new service group member does not drop (gossip) service files
Hi there, we are running a small Habitat ring on Ubuntu linux on AWS and upload a few files to the ring (certificates, key files) for one of our service groups. It seems that when a new node comes online and joins the service group, the configuration is generated properly but none of the service files are dropped in place until we re-run `hab file upload` to re-upload them to the ring. It seems like the desired behavior is that the service files (like the configuration files) should be generated when a service group is joined by a new node. I've confirmed that the service files are available in the ring state (via `/census`) at the time the node joins.

christophermaier:
Previously, because of how service files are tracked, you could start
up a service that did not have access to all its relevant service
files.

For instance, assume Supervisor A and Supervisor B are peered, and
that Supervisor A is running Service X, but Supervisor B is not. A
user uploads a file for Service X using `hab file upload`. The file is
written out to disk on Supervisor A, but not on Supervisor B, as
expected.

Then, the user loads Service X onto Supervisor B. The expectation
would be that the file is also written to disk there, but it is
not. Files are only written out at the time the file rumor is
received, and only if it is a "new" rumor. Since Supervisor B wasn't
running Service X when the file was uploaded, there was no managed
service present to cause the file to be written.

Uploading the file again with a new incarnation number would result in
Supervisor B writing out the file.

Here, we fix this by having the Supervisor write out any files for the
service that it knows about when the service is loaded, regardless of
whether or not they are "new" to the Supervisor.

Fixes #4877

### DIFF
--- a/components/sup/src/census.rs
+++ b/components/sup/src/census.rs
@@ -24,6 +24,7 @@ use std::{borrow::Cow,
                         HashMap,
                         HashSet},
           fmt,
+          iter::IntoIterator,
           path::Path,
           result,
           str::FromStr};
@@ -320,7 +321,7 @@ pub struct CensusGroup {
     local_member_id:       MemberId,
     population:            BTreeMap<MemberId, CensusMember>,
     update_leader_id:      Option<MemberId>,
-    changed_service_files: Vec<String>,
+    changed_service_files: HashSet<String>,
     service_files:         HashMap<String, ServiceFile>,
 }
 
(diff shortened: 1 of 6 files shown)


christophermaier:
@davidMcneil I think rumor garbage collection could have an impact here, depending on how it was implemented. There may need to be some distinction drawn between rumors concerning higher-level network constructs (like "service groups", and thus service config and service file rumors), and lower-level Supervisor-specific constructs (such as the presence or absence of a specific Supervisor).

PR #7807, opened 2020-07-14
Fix e2e failures

davidMcneil:
Add the following fixes to the e2e pipeline: 

* ignore errors when trying to stop the `hab-lauch` process
* increase the timeout for waiting to stop the supervisor
* use the full path to the binlinked `hab` binary, this avoids us trying to run the version installed with chocolatey

### DIFF
--- a/.expeditor/scripts/end_to_end/SupervisorRunner.cs
+++ b/.expeditor/scripts/end_to_end/SupervisorRunner.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 
 public class SupervisorRunner {
     private String logPath = null;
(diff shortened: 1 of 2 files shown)


.expeditor/scripts/end_to_end/SupervisorRunner.cs
```diff
@@ -17,6 +18,10 @@ public class SupervisorRunner {
         proc.StartInfo.RedirectStandardOutput = true;
         proc.StartInfo.RedirectStandardError = true;
         proc.StartInfo.FileName = "hab";
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
```
christophermaier, line 21: Do we need / have the equivalent of `$IsLinux`?

davidMcneil: Yeah, it would be more clear if this had an `else`. Added in [0edbb5](https://github.com/habitat-sh/habitat/pull/7807/commits/0edbb5009c4d62a814b3e90fd901a1af8ceb2e36)

christophermaier: This is a good change (and I hadn't even noticed it at first :sweat_smile: ), but I was thinking this conditional would be something we could formally set up to use in the same way that we use `$IsLinux`. That is, perhaps we should have an `$IsWindows` to complement it.

davidMcneil: Gotcha, this is declaring a type in `C#` which we pull into powershell [here](https://github.com/habitat-sh/habitat/blob/0edbb5009c4d62a814b3e90fd901a1af8ceb2e36/.expeditor/scripts/end_to_end/run_e2e_test_core.ps1#L95). `$IsLinux` and `$IsWindows` are both powershellisms. Maybe @mwrock knows a better way to express this?

christophermaier: Ooooh, gotcha. Maybe not a big issue, then.

Feel free to ignore this :sweat_smile:

PR #7808, opened 2020-07-14
fix ctrlc in local windows studio

### LINKED ISSUE #7794: ctrl+c breaks native windows studio
This has been a fairly long standing and painful issue. When invoking ctrl+c from an interactive local windows studio, stdin/out go haywire and the only effective remedy is to completely kill the console. This is not a problem in the docker studio.

I'm pretty sure this is related to nesting powerershells (outer windows powershell is running an inner powershell core) and I'm not entirely sure there is a good fix for this. I'd like to at least dig in and see if I can better understand the problem and if it can be resolved.

mwrock:
fixes #7794 

On windows, when a user enters ctrl-c, that sends a signal to every process running inside the console and not just the process running in the foreground. An application can register a handler for these signals or register `null` to disable the signal handling alltogether. Otherwise by default, a ctrlc will result in a call to `ExitProcess`. This PR registers an empty handler effectively turning off ctrl+c behavior for the hab cli process. This allows a user to stop builds and other processes inside of a studio with ctrlc without jacking up the stdin/out streams hosted by the parent hab process.

There are a couple associated fixes here to smooth this all out:

1. Removing the `habitat-studio.bat` wrapper. In order to provide cross-platform support for the `hab-studio` command. We mad a bat wrapper because windows will scan for files matching the command and that end in .`bat` and run those files against the `cmd` interpreter. Windows will not automatically run `.ps1` files with `pwsh`. Unfortunately `cmd` will prompt the user upon entering ctrl-c with `Terminate batch job (Y/N)`. The only way to suppress this is to send stdin to `NUL` which takes the interactive out of an interactive studio! So we are just jettisoning the wrapper here and calling the ps1 directly via pwsh.exe.

2. We are registering the local studio supervisor shutdown with `Register-EngineEvent` instead of just shutting down after the interactive prompt returns because the ctrl+c will exit the outer pwsh before it can execute any more code.

### DIFF
--- a/components/hab/src/command/studio/enter.rs
+++ b/components/hab/src/command/studio/enter.rs
@@ -23,8 +23,6 @@ pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 pub const CERT_PATH_ENVVAR: &str = "CERT_PATH";
 pub const SSL_CERT_FILE_ENVVAR: &str = "SSL_CERT_FILE";
 
-const STUDIO_CMD: &str = "hab-studio";
-const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 const STUDIO_PACKAGE_IDENT: &str = "core/hab-studio";
 
 #[derive(Clone, Copy)]
@@ -150,13 +148,15 @@ mod inner {
               str::FromStr};
 
     const SUDO_CMD: &str = "sudo";
+    const STUDIO_CMD: &str = "hab-studio";
+    const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
 
     pub async fn start(ui: &mut UI, args: &[OsString]) -> Result<()> {
         rerun_with_sudo_if_needed(ui, &args)?;
         if is_docker_studio(&args) {
             docker::start_docker_studio(ui, args)
         } else {
-            let command = match henv::var(super::STUDIO_CMD_ENVVAR) {
+            let command = match henv::var(STUDIO_CMD_ENVVAR) {
                 Ok(command) => PathBuf::from(command),
                 Err(_) => {
                     init()?;
@@ -164,7 +164,7 @@ mod inner {
                     let ident = PackageIdent::from_str(&format!("{}/{}",
                                                                 super::STUDIO_PACKAGE_IDENT,
                                                                 version[0]))?;
-                    let command = exec::command_from_min_pkg(ui, super::STUDIO_CMD, &ident).await?;
+                    let command = exec::command_from_min_pkg(ui, STUDIO_CMD, &ident).await?;
                     // This is a duplicate of the code in `hab pkg exec` and
                     // should be refactored as part of or after:
                     // https://github.com/habitat-sh/habitat/issues/6633
@@ -176,7 +176,7 @@ mod inner {
                         env::set_var(key, value);
                     }
 
-                    let mut display_args = super::STUDIO_CMD.to_string();
+                    let mut display_args = STUDIO_CMD.to_string();
                     for arg in args {
                         display_args.push(' ');
                         display_args.push_str(arg.to_string_lossy().as_ref());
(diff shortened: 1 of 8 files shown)


components/hab/src/command/studio/enter.rs
```diff
@@ -150,21 +148,23 @@ mod inner {
               str::FromStr};
 
     const SUDO_CMD: &str = "sudo";
+    const STUDIO_CMD: &str = "hab-studio";
+    const STUDIO_CMD_ENVVAR: &str = "HAB_STUDIO_BINARY";
```
christophermaier, line 152: These were moved into this sub-module because they're no longer used on Windows, correct?

christophermaier: Random aside that I'll mention here, since I don't have a better way to do it... while you're in here, it would be good to remove this code: https://github.com/habitat-sh/habitat/blob/0f3e785046e3eb9dca8cbd3b324d69762e392e59/components/hab/src/command/studio/enter.rs#L200-L202

That's in a module that's only compiled on Linux, so it'll never be called.

components/hab/src/command/studio/enter.rs
```diff
@@ -273,22 +272,30 @@ mod inner {
...
+            PackageIdent::from_str(&format!("{}/{}", super::STUDIO_PACKAGE_IDENT, version[0]))?;
+        let pwsh_command = exec::command_from_min_pkg(ui, "pwsh.exe", &ident).await?;
+        let studio_command = exec::command_from_min_pkg(ui, "hab-studio.ps1", &ident).await?;
+
+        let mut cmd_args: Vec<OsString> = vec!["-NoProfile".into(),
+                                               "-ExecutionPolicy".into(),
+                                               "bypass".into(),
+                                               "-NoLogo".into(),
+                                               "-File".into(),
+                                               studio_command.to_string_lossy().as_ref().into()];
```
christophermaier, line 287: You could use `PathBuf::into_os_string()` here to simplify (or really just `.into()`).

Going further, I think you could probably do 
```
vec!["-NoProfile", 
     "-ExecutionPolicy", 
     "bypass", 
     "-NoLogo",
     "-File", 
     studio_command]
.into_iter()
.map(Into::into)
.collect();
```

components/hab/src/command/studio/enter.rs
```diff
@@ -273,22 +272,30 @@ mod inner {
...
+        cmd_args.extend_from_slice(args);
+
+        if let Some(cmd) = find_command(pwsh_command.to_string_lossy().as_ref()) {
```
christophermaier, line 290: I don't think you need the `to_string_lossy().as_ref()` business here; `pwsh_command` should be a `PathBuf`, and `find_command` takes an `AsRef<Path>`, so everything should already be compatible.

components/studio/test/test.ps1
```diff
@@ -24,7 +24,7 @@ Copy-Item "$(hab pkg path core/7zip)/bin/*" "bin/7zip"
 Copy-Item "$(hab pkg path core/hab-plan-build-ps1)/bin/*" "bin/"
 
 try {
-    & bin/hab-studio.bat new
+    & bin/powershell/pwsh.exe -NoProfile -ExecutionPolicy bypass -NoLogo -File "bin/hab-studio.ps1" new
```
christophermaier, line 27: What's the context in which this test file is run? Is there any way to share this with the same invocation that encoded in the Rust code now?

At the very least, it would be good to add a comment here referencing the Rust code (and vice versa) to ensure that future changes are made in both places.

mwrock: this is a "unit" test at least in the sense that the test can run directly on the studio code without requiring a compiled hab cli. This basically just replaces what was in the .bat file and is a pretty boilerplate/ubiquitous way to call into a powershell script in an embedded manner.

components/hab/src/command/studio/enter.rs
mwrock: correct. deleting this - good call!

components/hab/src/command/studio/enter.rs
mwrock: thanks! I knew there had to be a better way.

mwrock:
suggestions applied @christophermaier

PR #7823, opened 2020-07-20
Add supplemental groups to Linux hook processes

### LINKED ISSUE #7731: Processes started by hab-sup/launcher do not have supplementary group list set
# Summary and Example

Processes started with the Habitat supervisor/launcher may not be able to access the files that a user expects because Habitat does not set the process's supplementary group IDs.

For example, consider the following situation:

```
 > ls -al /tmp/testfile
-rw-rw----. 1 vagrant foo 30 Jun  3 09:03 /tmp/testfile
> getent group foo
foo:x:1005:hab,vagrant
> sudo -u hab cat /tmp/testfile
This is the testfile content!
> sudo -u chrony cat /tmp/testfile
cat: /tmp/testfile: Permission denied
```

However, when we run a service with the following run hook:

```
#!/bin/bash

echo "Attempting to access /tmp/testfile as hab svc (PID: $$, EIUD: $EUID)"

cat /tmp/testfile
sleep 300
```

we get:

```
test-hab-service.default(SV): Starting service as user=hab, group=hab
test-hab-service.default(O): Attempting to access /tmp/testfile as hab svc (PID: 13326, EIUD: 1001)
test-hab-service.default(E): cat: can't open '/tmp/testfile': Permission denied
```

This difference is because when a process is launched via `sudo` we get our supplementary group list set, but when it is launched under Habitat it does not:

```
> sudo grep 'Groups' /proc/13326/status 
Groups:	 
> sudo -u hab bash -c 'grep "Groups" /proc/$$/status'
Groups:	1001 1005 
```

# Background

On Linux and other unix systems, one part of a process's credentials is a list of _Supplementary group IDs_.  From [credentials(7)](https://www.man7.org/linux/man-pages/man7/credentials.7.html):

>  This is a set of additional group IDs that are used for permission checks when accessing files and other shared resources... A process can obtain its set of supplementary group IDs using getgroups(2), and can modify the set using setgroups(2).

This lists of groups is inherited from a processes parent and can be explicitly modified with the [initgroups](https://www.man7.org/linux/man-pages/man3/initgroups.3.html) and [setgroups](https://www.man7.org/linux/man-pages/man2/setgroups.2.html) system calls.

When the launcher starts our run hook, we use `uid` and `gid` from `std::os::unix::process::CommandExt` to set the effective UID and effective GID of the process:

https://github.com/habitat-sh/habitat/blob/master/components/launcher/src/sys/unix/service.rs#L97-L98

But these do not set the supplementary group IDs list.  There is currently a pull request open to add such a feature here: https://github.com/rust-lang/rust/pull/72160

For comparison, systemd sets the supplementary group IDs automatically:

> If the User= setting is used the supplementary group list is initialized from the specified user's default group list, as defined in the system's user and group database. Additional groups may be configured through the SupplementaryGroups= setting (see below).

(https://www.freedesktop.org/software/systemd/man/systemd.exec.html)

# Potential Path for a Patch

Until the standard library supports setting supplementary groups directly, we could use https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.pre_exec to call initgroups or setgroups ourselves before the process is exec'd.

christophermaier:
Whenever the Supervisor or Launcher spawn a lifecycle hook process, we'll now add supplemental group information to the new process, ensuring that it has access to everything the service user needs.

See individual commits for further details, as well as several accessory refactorings.

Fixes #7731


components/common/Cargo.toml
```diff
@@ -48,6 +48,13 @@ toml = { version = "*", features = ["preserve_order"] }
 uuid = { version = "*", features = ["v4"] }
 valico = "*"
 
+[target.'cfg(not(windows))'.dependencies]
+# Pulling from git until
+# https://github.com/nix-rust/nix/commit/3c2107bdc221a90b02d3f7118dd96f3496762cd0
+# is released, which will be the next release after v0.17.0
+# (A similar pin is in components/core and components/launcher as well).
+nix = {  git = "https://github.com/nix-rust/nix.git", rev = "3c2107bdc221a90b02d3f7118dd96f3496762cd0" }
```
davidMcneil, line 56: Not necessary to change here, but in the future, we could utilize [`[patch.crates-io]`](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html) to patch a dependency at the workspace level. Although I am not sure if it works with the `'cfg(not(windows))'` condition.

.expeditor/verify.pipeline.yml
```diff
@@ -577,6 +577,20 @@ steps:
       automatic:
         limit: 1
 
+  - label: "[build] :linux: supplemental-group-tester"
```
davidMcneil, line 580: I missed it; why do we build this here?

christophermaier: :thinking: I guess we don't really need to... it makes more sense with the `test-probe` package, since there's actually code to compile there. This is just a dumb service.

I'll remove it; in retrospect, I don't think it really buys us anything.

components/common/Cargo.toml
christophermaier: Ah, good suggestion! I'll give that a try.

christophermaier: Addressed in https://github.com/habitat-sh/habitat/pull/7823/commits/71c82566aad09d69bda334c5229992470c863410

.expeditor/verify.pipeline.yml
christophermaier: Addressed in https://github.com/habitat-sh/habitat/pull/7823/commits/691da5a033e6221cb442e52f27a5ca880a983bb2

PR #7829, opened 2020-07-22
delete old objects in S3 bucket that do not exist in local website build

jeremymv2:
We would like to ensure that the sync of content from each new website build removes files that exist in the S3 bucket but do not exist in the local build. 

From https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html#options

> --delete (boolean) Files that exist in the destination but not in the source are deleted during sync.

### DIFF
--- a/.expeditor/scripts/shared.sh
+++ b/.expeditor/scripts/shared.sh
@@ -122,11 +122,6 @@ s3_channel_url_root() {
     echo "s3://${s3_bucket_name}/${channel}/latest/habitat"
 }
 
-s3_sync() {
-    # ex. s3_sync "src" "dst" "option1" "option2 <value>" ...
-    aws s3 sync "$@"
-}
-
 purge_fastly_cache() {
 	local service_key="${1}"
 	local channel="${2:-}"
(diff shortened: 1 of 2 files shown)


.expeditor/scripts/shared.sh
```diff
@@ -123,9 +123,8 @@ s3_channel_url_root() {
 }
 
 s3_sync() {
-	local src="${1}"
-	local dst="${2}"
-	aws s3 sync "${src}" "${dst}"
+    # ex. s3_sync "src" "dst" "option1" "option2 <value>" ...
+    aws s3 sync "$@"
```
christophermaier, line 127: :thinking: It appears that the website deploy is the only consumer of `s3_sync`. As such, I don't really see that this abstraction really buys us anything. I think you could just remove this function and call `aws s3 sync . "s3://$AWS_BUCKET" --delete` directly in `website-deploy.sh` and call it a day.

jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/7829/commits/316dbb662531496c918c61c95a7e8c40403bf2bc

PR #7834, opened 2020-07-23
Allow version check without license

### LINKED ISSUE #7751: exempt `hab --version` from license check
I have [STRTA bootstrap](https://github.com/github/scripts-to-rule-them-all#scriptbootstrap) scripts that check developer workstations for dependencies, and recently added a check of `hab --version` to notify anyone running a dated version

I'm handling walking them through license acceptance separately (#6978 would help a lot there), and the license prompt potentially happening while running `hab --version` breaks my version parsing. I want to get their `hab` CLI current before they modify the state of their workstation with it. It is common for someone who has checked out one of my projects previously to have an ancient pre-0.80 hab binary on their workstation already.

It was surprising to me that `hab --version` could fail due to environmental state, I think it's reasonable to expect it to _only_ report the version of the binary and be a reliable way to check that an executable and current binary is present. Consider exempting `hab --version` from license checking so introspection tools can measure the state of the hab binary separately from environment setup

davidMcneil:
Resolves #7751 


test/end-to-end/test_version_check_without_license.ps1
```diff
@@ -0,0 +1,28 @@
...
+    It "non-version check commands do not work without license" {
+        $process = Start-Process "hab" -ArgumentList "--help" -PassThru
+        {
+            $process | Wait-Process -Timeout 1 -ErrorAction Stop
+        } | Should -Throw "time-out"
+
+        $process | Stop-Process -Force
+    }
+
+    hab license accept
```
mwrock, line 19: I would put this line in the below `It` block. I forget some of the pester nuances here, but I think it may call ^^^ as it is building its list of `It` tests and therefore it could impact the top 2 tests.

davidMcneil: Thanks! Addressed in [2ff021](https://github.com/habitat-sh/habitat/pull/7834/commits/2ff0217633fa6e637645ba7e3c7f100191ae8052)

components/hab/src/main.rs
```diff
@@ -117,10 +117,13 @@ async fn main() {
...
 async fn start(ui: &mut UI, feature_flags: FeatureFlag) -> Result<()> {
-    if std::env::args().skip(1).collect::<Vec<_>>() == vec!["license", "accept"] {
+    let args = std::env::args().skip(1).collect::<Vec<_>>();
+    if args == vec!["license", "accept"] {
         license::accept_license(ui)?;
         return Ok(());
-    } else {
+    }
+    // If we are just checking the version do not require accepting the license
+    else if args != vec!["--version"] {
```
stevendanna, line 126: I think we also accept `-V` for version.

Might it also be worth it to check for `--help`?

davidMcneil: > I think we also accept -V for version.

Agreed, it would be good to add `-V`.

> Might it also be worth it to check for `--help`?

Yes, this would be great. If we do, I would like to allow `--help` for all subcommands. This is actually pretty straightforward with clap, but because of our... "unique"... argument parsing logic it becomes slightly complicated. Also because this change had a legal element I defaulted to a more simplistic approach. From a purely technical standpoint I would prefer the implementation to look like [c8fed7](https://github.com/habitat-sh/habitat/commit/c8fed7e7b700f2e91b4e80505bf1c3806100f879). I could switch to this implementation and add more tests. Which would people prefer? @christophermaier @mwrock

christophermaier: If there's a nice way to segment license-requiring vs. non-license-requiring commands, then I'm all for it :+1:

test/end-to-end/test_license.ps1
```diff
@@ -0,0 +1,82 @@
+Remove-Item -Recurse -Force -ErrorAction SilentlyContinue "$HOME/.hab/accepted-licenses"
+Remove-Item -Recurse -Force -ErrorAction SilentlyContinue "/hab/accepted-licenses"
+
+Describe "license" {
+    It "version check without license works" {
+        ./target/debug/hab --version
```
mwrock, line 6: I think all of the `./target/debug/` instances were accedentally left here during local testing.

components/hab/src/license.rs
```diff
@@ -152,19 +172,21 @@ fn writeable_license_path() -> PathBuf {
...
     match env::var(LICENSE_ACCEPT_ENVVAR) {
         Ok(val) => {
             if &val == "accept" {
                 write_license_file()?;
-                Ok(true)
+                Ok(LicenseAcceptance::Accepted)
             } else if &val == "accept-no-persist" {
-                Ok(true)
+                Ok(LicenseAcceptance::Accepted)
+            } else if &val == "deny" {
```
mwrock, line 183: Do we need to add the concept of `deny`?

test/end-to-end/test_license.ps1
davidMcneil: Yep :sweat_smile: , thanks! Addressed in [fdf8c8](https://github.com/habitat-sh/habitat/pull/7834/commits/fdf8c8518285352a45d44e5bb0f71578c40f349b)

components/hab/src/license.rs
davidMcneil: > Do we need to add the concept of `deny`?

No, we do not. It makes testing slightly more robust because we do not need to rely on timing out the license acceptance prompt. I also thought about a `no-prompt` option. Do we think `deny` or `no-prompt` could ever be useful outside of our own testing? If not I will remove that logic.

mwrock: Oh I had a feeling it may have been for testing but the connection was not super clear to me. Now I see what you are doing with that. It not be a bad idea to add a comment on the motivation behind deny, but it seems legit to me.

davidMcneil: Agreed. Addressed in [72e3ba](https://github.com/habitat-sh/habitat/pull/7834/commits/72e3ba8a860d9c78dedf679540b814909bce1672)

components/hab/src/license.rs
```diff
@@ -152,19 +174,21 @@ fn writeable_license_path() -> PathBuf {
     license_path(&root_dir)
 }
 
-fn env_var_present() -> Result<bool> {
+fn env_var_present() -> Result<LicenseAcceptance> {
```
christophermaier, line 177: Might be better to name this something like `acceptance_from_env_var` now.

davidMcneil: Agreed, thanks! Addressed in [2c27f4](https://github.com/habitat-sh/habitat/pull/7834/commits/2c27f45f18ef19704aaf768198df3517c2e0e273)

PR #7840, opened 2020-07-27
use win32 api and not env var for determining current user name

### LINKED ISSUE #6457: Use `GetUserNameW` to get the username of the current Windows user
We currently use the `USERNAME` environment variable when we check what the current username is running on Windows. This pretty much always works but technically, using `GetUserNameW` is more reliable.

mwrock:
fixes #6457

### DIFF
--- a/components/launcher-protocol/protocols/launcher.proto
+++ b/components/launcher-protocol/protocols/launcher.proto
@@ -59,3 +59,12 @@ message PidIs {
   // way to evolve that?
   optional uint32 pid = 1;
 }
+
+// Query the Launcher to determine the launcher's version
+message Version {}
+
+// The response that corresponds to `Version` and contains the
+// version number.
+message VersionNumber {
+  optional uint32 version = 1;
+}
--- a/components/launcher/src/lib.rs
+++ b/components/launcher/src/lib.rs
@@ -12,3 +12,4 @@ mod sys;
 
 pub const SUP_CMD: &str = "hab-sup";
 pub const SUP_PACKAGE_IDENT: &str = "core/hab-sup";
+pub const VERSION: Option<&str> = option_env!("PLAN_VERSION");
--- a/components/launcher/src/server/handlers/version.rs
+++ b/components/launcher/src/server/handlers/version.rs
@@ -1,0 +1,28 @@
+use super::{HandleResult,
+            Handler};
+use crate::{protocol,
+            server::ServiceTable,
+            VERSION};
+
+pub struct VersionHandler;
+
+impl Handler for VersionHandler {
+    type Message = protocol::Version;
+    type Reply = protocol::VersionNumber;
+
+    fn handle(_: Self::Message, _: &mut ServiceTable) -> HandleResult<Self::Reply> {
+        // VERSION will be none if this is a cargo built binary as opposed to
+        // being built by our hab plan. So in that case we will fallback to u32::MAX.
+        match VERSION.unwrap_or(&u32::MAX.to_string()).parse::<u32>() {
+            Ok(version) => {
+                let reply = protocol::VersionNumber { version };
+                Ok(reply)
+            }
+            Err(_) => {
+                let mut reply = protocol::NetErr::default();
+                reply.code = protocol::ErrCode::InvalidVersionNumber;
+                Err(reply)
+            }
+        }
+    }
+}
--- a/components/launcher/src/sys/windows/service.rs
+++ b/components/launcher/src/sys/windows/service.rs
@@ -3,12 +3,14 @@ use crate::{error::{Error,
             protocol::{self,
                        ShutdownMethod},
             service::Service};
-use core::{os::process::{handle_from_pid,
-                         windows_child::{Child,
-                                         ExitStatus,
-                                         Handle}},
+use core::{os::{process::{handle_from_pid,
+                          windows_child::{Child,
+                                          ExitStatus,
+                                          Handle}},
+                users::get_current_username},
            util};
 use std::{collections::HashMap,
+          env,
           io,
           iter::FromIterator,
           mem,
@@ -140,7 +142,27 @@ fn spawn_pwsh(ps_binary_name: &str, msg: protocol::Spawn) -> Result<Service> {
     let password = msg.svc_password.clone();
 
     let user = match msg.svc_user.as_ref() {
-        Some(u) => u.to_string(),
+        Some(u) => {
+            // In the case where we are spawning on behalf of an older Supervisor
+            // we will need to revert to older 'get_current_username' behavior. When
+            // running as the Local System account, the former behavior would return
+            // the host name followed by a dollar sign. The new behavior returns
+            // 'system'. Both the supervisor and the launcher behavior must match.
+            // Otherwise if we are running under system, we will interpret the user
+            // to spawn as different from ourselves and thus attempt to logon as ourselves
+            // which will fail since you cannot simply logon as system. One day we can
+            // remove this when we are confident everyone is on a recent supervisor
+            // and launcher.
+            let mut username = u.to_string();
+            if get_current_username() == Some("system".to_string()) {
+                if let Ok(cn) = env::var("COMPUTERNAME") {
+                    if u == &(cn.to_lowercase() + "$") {
+                        username = "system".to_string();
+                    }
+                }
+            }
+            username
+        }
         None => {
             return Err(Error::UserNotFound(String::from("")));
         }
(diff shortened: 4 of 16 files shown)


components/launcher-protocol/protocols/launcher.proto
```diff
@@ -59,3 +59,12 @@ message PidIs {
...
   optional uint32 pid = 1;
 }
+
+// Query the Launcher to determine the launcher's version
+message Version {}
+
+// The response that corresponds to `Version` and contains the
+// version number.
+message VersionNumber {
+  optional uint32 version = 1;
```
davidMcneil, line 69: Could this be `required`?

components/sup/src/manager/service/supervisor.rs
```diff
@@ -167,14 +169,47 @@ impl Supervisor {
...
+        //
+        // So we will check the version of the launcher. This version check is added
+        // in the same set of commits that changed the current username inspection
+        // implementation. So if the version handler is not supported, which it would
+        // not be in an older launcher, we will convert the user name to the old style
+        // system user using the host name.
+        let user = {
+            if pkg.svc_user == "system" {
+                match launcher.version() {
+                    Ok(_) => pkg.svc_user.clone(),
```
davidMcneil, line 195: Could we put a match guard here checking the version is greater than what we expect? I know current user change was added at the same time as the launcher version call, but it seems like it would be good to be explicit and would help make the code more self documenting.

components/launcher/src/lib.rs
```diff
@@ -12,3 +12,4 @@ mod sys;
 
 pub const SUP_CMD: &str = "hab-sup";
 pub const SUP_PACKAGE_IDENT: &str = "core/hab-sup";
+pub const VERSION: Option<&str> = option_env!("PLAN_VERSION");
```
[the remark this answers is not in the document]
mwrock: The launcher actually uses a different versioning scheme than most of our other components. It does not use the version in the `VERSION` file but instead does a count of git commits at build time in the habitat plan. So in the launcher's case, a plain local cargo build will not have a version.

components/launcher-protocol/protocols/launcher.proto
mwrock: maybe? I was trying to conform to the convention used for all the other messages which all specify `optional`. I just assume there is a reason for that but I'm not certain what it is.

components/sup/src/manager/service/supervisor.rs
mwrock: I had thought about that too. The trouble is I was not sure what version number to use. However now that I think of this, it really doesn't matter since there will be no version at all until this is in dev so I guess I can just use the latest unstable and it will be fine.

components/launcher/src/server/handlers/version.rs
```diff
@@ -0,0 +1,20 @@
...
+pub struct VersionHandler;
+
+impl Handler for VersionHandler {
+    type Message = protocol::Version;
+    type Reply = protocol::VersionNumber;
+
+    fn handle(_: Self::Message, _: &mut ServiceTable) -> HandleResult<Self::Reply> {
+        let version = VERSION.unwrap_or_default()
+                             .parse::<u32>()
+                             .unwrap_or_default();
```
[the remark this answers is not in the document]
mwrock: I think `0` would be the expected version of a locally built launcher ala cargo. However it makes good sense to me to err if the version is not numeric.

components/launcher/src/lib.rs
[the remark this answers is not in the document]
davidMcneil: Ah, that's right... I wonder if it would make sense to make this be `u32::max` or something in the case that `PLAN_VERSION` is not set. It seems probable that we will have launcher version checks that check for some minimum version. Having local builds always be 0 could cause features to not work that should.

components/launcher-protocol/protocols/launcher.proto
stevendanna: I'm not sure on the original intent, but in proto-land there is a strong bias against required fields. So far so that they were removed in proto3.  The reason that is typically given is that `required` fields make it hard to make non-breaking changes to the message down the road.

components/sup/src/manager/service/supervisor.rs
mwrock: good call

mwrock:
@davidMcneil I made the `u32::MAX` change and also added some sad logic to accomodate the scenario where you have a new launcher and older supervisor.

components/launcher/src/sys/windows/service.rs
```diff
@@ -140,7 +142,27 @@ fn spawn_pwsh(ps_binary_name: &str, msg: protocol::Spawn) -> Result<Service> {
     let password = msg.svc_password.clone();
 
     let user = match msg.svc_user.as_ref() {
-        Some(u) => u.to_string(),
+        Some(u) => {
+            // In the case where we are spawning on behalf of an older Supervisor
+            // we will need to revert to older 'get_current_username' behavior. When
+            // eunning as the Local System account, the former behavior would return
```
stevendanna, line 148:
```suggestion
            // running as the Local System account, the former behavior would return
```

components/launcher/src/sys/windows/service.rs
```diff
@@ -140,7 +142,27 @@ fn spawn_pwsh(ps_binary_name: &str, msg: protocol::Spawn) -> Result<Service> {
...
+            // the host name followed by a dollar sign. The new behavior returns
+            // 'system'. Both the supervisor and the launcher behavior must match.
+            // Otherwise if we are running under system, we will interpret the user
+            // to spawn as different from ourselves and thus attempt to logon as ourselves
+            // which will fail since you cannot simply logon as system. One day we can
+            // remove this when we are confidentr everyone is on a recent supervisor
```
stevendanna, line 154:
```suggestion
            // remove this when we are confident everyone is on a recent supervisor
```

components/launcher/src/server/handlers/version.rs
```diff
@@ -0,0 +1,26 @@
...
+            VERSION};
+
+pub struct VersionHandler;
+
+impl Handler for VersionHandler {
+    type Message = protocol::Version;
+    type Reply = protocol::VersionNumber;
+
+    fn handle(_: Self::Message, _: &mut ServiceTable) -> HandleResult<Self::Reply> {
+        match VERSION.unwrap_or(&u32::MAX.to_string()).parse::<u32>() {
```
davidMcneil, line 14: It might be worth adding a comment explaining the why of using `u32::MAX`.

components/sup/src/manager/service/supervisor.rs
```diff
@@ -167,14 +170,56 @@ impl Supervisor {
...
+                    Ok(_) => legacy_user,
+                    Err(err) => {
+                        match err {
+                            Timeout => {
+                                error!("Error getting version from launcher: {:?}", err);
+                                legacy_user
+                            }
+                            _ => return Err(Error::Launcher(err)),
+                        }
+                    }
```
davidMcneil, line 209: You could write this as:

```rust
Err(Timeout) => {
    error!("Error getting version from launcher: {:?}", err);
    legacy_user
}
Err(err) => {
    return Err(Error::Launcher(err));
}
```

to avoid an extra match statement.

components/sup/src/error.rs
```diff
@@ -40,6 +40,7 @@ pub enum Error {
     CtlSecretIo(PathBuf, io::Error),
     APIClient(habitat_api_client::Error),
     EnvJoinPathsError(env::JoinPathsError),
+    EnvVarError(env::VarError),
```
davidMcneil, line 43: It does not look like this is used anywhere.

mwrock: Well I needed the `From` trait implementation because of `?` at https://github.com/habitat-sh/habitat/pull/7840/files#diff-b7cc1dcc0589de0582a1250393299db1R195

components/sup/src/manager/service/supervisor.rs
mwrock: I actually struggled a bit with this and ended up with the seemingly verbose implementation I used. Your suggestion fails because `err` in `error!("Error getting version from launcher: {:?}", err);` is not in scope. Is there a way to specify the `Timeout` type and also get its value?

davidMcneil: Yep there is; sorry I missed that. A [binding](https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/binding.html) should work. Something like:

```rust
Err(e @ Timeout) => {
    error!("Error getting version from launcher: {:?}", err);
    legacy_user
}
Err(err) => {
    return Err(Error::Launcher(err));
}
```

mwrock: ahh sweet. thanks!

components/sup/src/error.rs
davidMcneil: May bad :sweat_smile:; I was testing on linux.

Issue #7599, opened 2020-04-06
hab cli subcommand for builder user role management


jeremymv2:
With the advent of lightweight RBAC in builder (https://github.com/habitat-sh/builder/pull/1378), we need to add a CLI subcommand for users to manage roles within an origin.

jamesma100:
Hey @jeremymv2  I'm interested in this issue, could I work on it? Since this would be my first contribution it would be great if I could have some guidance!

jeremymv2:
@jamesma100 thanks for looking into potential contributions to Habitat!  There's always plenty to be done 😄 

I've begun work on this feature and have a branch that I pushed up:
https://github.com/habitat-sh/habitat/tree/jeremymv2/rbac_subcommand 

I haven't pushed to move forward with the work since I'm waiting on two issues to harmonize first:

- https://github.com/habitat-sh/builder/issues/1367#issuecomment-628802308
- https://github.com/habitat-sh/builder/issues/1403

Finalizing RBAC is definitely on the short list of action items though! cc: @eeyun, @markan, @smacfarlane

jamesma100:
@jeremymv2 Thanks for getting back to me. I'd be happy to help out with RBAC. I see we have the owner, admin, and maintainer roles left to implement. Thanks.

PR #7841, opened 2020-07-28
builder origin rbac subcommand

jeremymv2:
Closes https://github.com/habitat-sh/habitat/issues/7599
Companion to https://github.com/habitat-sh/builder/pull/1431

Although there was a lot more potential structopt work to do, I had to stop myself and stash that effort as the blast radius became very broad (ie. AuthToken, Origin and BldrURL touch 90% of CLI commands).
To handle the remaining structopt work that will allow us to be more streamlined, I created https://github.com/habitat-sh/habitat/issues/7866


eeyun:
Code looks good, but it does appear to have broken some unit tests. Otherwise its :+1: from me.

jeremymv2:
@eeyun the unit tests are fixed up now.

components/builder-api-client/src/lib.rs
```diff
@@ -380,6 +381,53 @@ impl Client {
...
+            "maintainer" => Ok(OriginMemberRole::Maintainer),
+            "administrator" => Ok(OriginMemberRole::Administrator),
+            "owner" => Ok(OriginMemberRole::Owner),
+            _ => Err(Error::BadOriginMemberRole(value.to_string())),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct OriginMemberRoleResponse {
```
davidMcneil, line 423: Do we need this type? Could we implement `PortableText` directly on `OriginMemberRole`?

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +238,60 @@ pub struct UploadGroup {
     public_file: Option<PathBuf>,
 }
 
+#[derive(ConfigOpt, StructOpt)]
+#[structopt(no_version)]
```
davidMcneil, line 242: You could make this `#[structopt(no_version, rename_all = "screamingsnake")]` to avoid the need to explicitly type `#[structopt(name = "<NAME">` on the fields below.

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +238,60 @@ pub struct UploadGroup {
...
+    /// Display an origin member's current role
+    Show {
+        /// The account name of the role to display
+        #[structopt(name = "MEMBER_ACCOUNT")]
+        member_account: String,
+        /// The name of the origin for the origin member association
+        #[structopt(name = "ORIGIN",
+                short = "o",
+                long = "origin",
+                required = true,
```
davidMcneil, line 254: I dont believe this should be needed. Fields are automatically required if they are not wrapped in `Option`.

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +238,60 @@ pub struct UploadGroup {
...
+                validator = valid_origin)]
```
davidMcneil, line 255: :thinking: should we have an `Origin` type. It could help make the code more self document and we would not need to explicitly state this validator. We could just have `origin: Origin`.

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +238,60 @@ pub struct UploadGroup {
...
+    /// Change an origin member's role
+    Set {
+        /// The account name whose role will be changed
+        #[structopt(name = "MEMBER_ACCOUNT")]
+        member_account: String,
+        /// The name of the origin for the origin member association
+        #[structopt(name = "ORIGIN",
+                short = "o",
+                long = "origin",
+                required = true,
```
davidMcneil, line 274: Should not be needed

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +238,60 @@ pub struct UploadGroup {
...
+                validator = valid_origin)]
+        origin:         String,
+        /// The role name to enforce for the member account [values: member, maintainer,
+        /// administrator]
+        #[structopt(name = "ROLE",
+                short = "r",
+                long = "role",
+                required = true,
```
davidMcneil, line 282: Should not be needed

components/hab/src/cli.rs
```diff
@@ -403,6 +404,35 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
...
+                     (@arg ROLE: --role -r +takes_value +required {valid_role} "The role name to enforce for the member account \
+                          [values: member, maintainer, administrator]")
+                     (@arg BLDR_URL: -u --url +takes_value {valid_url}
+                          "Specify an alternate Builder endpoint. If not specified, the value will \
+                          be taken from the HAB_BLDR_URL environment variable if defined. (default: \
+                          https://bldr.habitat.sh)")
+                     (@arg AUTH_TOKEN: -z --auth +takes_value "Authentication token for Builder")
+                     (@arg NO_PROMPT: -n --("no-prompt") "Do not prompt for confirmation")
+                )
+            )
```
davidMcneil, line 435: Because this is a new command I think we should go all in on `structopt` and use it instead of the explicit `clap` version. To keep the `structopt` and `clap` versions in sync we still want to add the subcommand, but we can do that with `(subcommand Rbac::clap())`. See [here](https://github.com/habitat-sh/habitat/blob/9472ea6f36ff1d2aaba6be16a736a60ab8d4ad1e/components/hab/src/cli.rs#L881) for an example.

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +238,60 @@ pub struct UploadGroup {
...
+                validator = valid_role)]
+        role:           String,
```
davidMcneil, line 284: Should this be `OriginMemberRole`?

components/hab/src/main.rs
```diff
@@ -242,6 +242,13 @@ async fn start(ui: &mut UI, feature_flags: FeatureFlag) -> Result<()> {
                         _ => unreachable!(),
                     }
                 }
+                ("rbac", Some(m)) => {
+                    match m.subcommand() {
+                        ("set", Some(sc)) => sub_origin_member_role_set(ui, sc).await?,
+                        ("show", Some(sc)) => sub_origin_member_role_show(ui, sc).await?,
+                        _ => unreachable!(),
+                    }
+                }
```
davidMcneil, line 251: Instead of parsing this here using `clap` methods, could we move it to part of this [match statement](https://github.com/habitat-sh/habitat/blob/9472ea6f36ff1d2aaba6be16a736a60ab8d4ad1e/components/hab/src/main.rs#L150). This would allow us to use the `Rbac` type instead of `get_matches`.

davidMcneil:
I know this was not added in the PR, but could we add a default implementation for [`PortableText::as_json`](https://github.com/habitat-sh/habitat/blob/9472ea6f36ff1d2aaba6be16a736a60ab8d4ad1e/components/builder-api-client/src/lib.rs#L251). Something like:

```rust
pub trait PortableText: Serialize {
    fn as_json(&self) -> Result<Json> {
        serde_json::to_value(self).map_err(|e| habitat_core::Error::RenderContextSerialization(e).into())
    }
}
```

This should avoid the need for multiple `[convert_to_json](https://github.com/habitat-sh/habitat/blob/9472ea6f36ff1d2aaba6be16a736a60ab8d4ad1e/components/builder-api-client/src/lib.rs#L232)`s

components/builder-api-client/src/lib.rs
jeremymv2: Yep, we need this one. The response that comes from the API endpoint is not a `OriginMemberRole` but a struct that effectively wraps it, hence `OriginMemberRoleResponse`.

components/hab/src/cli/hab/origin.rs
jeremymv2: I have done so.

components/builder-api-client/src/error.rs
```diff
@@ -62,6 +63,9 @@ impl fmt::Display for Error {
                 format!("Failed to read package artifact, {}, {}", p.display(), e)
             }
             Error::ParseIntError(ref err) => format!("{}", err),
+            Error::RenderContextSerialization(ref e) => {
+                format!("Unable to serialize rendering context, {}", e)
+            }
```
jeremymv2, line 68: 🤔 I'm going to try and remove this as it's now in the `core` crate.

jeremymv2: removed

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
+    match bldr_auth_token_from_args_env_or_load(opt) {
+        Ok(v) => Some(v),
+        Err(_) => None,
+    }
 }
 
 lazy_static! {
     pub static ref CACHE_KEY_PATH_DEFAULT: String =
         hab_core_fs::CACHE_KEY_PATH.to_string_lossy().to_string();
+    pub static ref HAB_BLDR_URL_DEFAULT: String = bldr_url_from_env_load_or_default();
```
davidMcneil, line 122: It does not look like this is currently being used? Could we add it back in a subsequent PR when we know precisly how we are going to handle it?

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
+            }
+        }
+    }
+}
+
+pub fn maybe_bldr_auth_token_from_args_or_load(opt: Option<String>) -> Option<String> {
+    match bldr_auth_token_from_args_env_or_load(opt) {
+        Ok(v) => Some(v),
+        Err(_) => None,
+    }
```
davidMcneil, line 116: We could use [`ok()`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.ok)

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
+    bldr_url_from_env().unwrap_or_else(|| {
+                           match config::load() {
+                               Ok(config) => {
+                                   match config.bldr_url {
+                                       Some(v) => v,
+                                       None => DEFAULT_BLDR_URL.to_string(),
+                                   }
+                               }
+                               Err(e) => {
+                                   println!("Found a cli.toml but unable to load it. Resorting \
```
davidMcneil, line 73: Maybe use `error!` instead of `println!`

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
+    /// The Builder origin name to target
+    #[structopt(name = "ORIGIN", short = "o", long = "origin", validator = valid_origin)]
+    pub name: Origin,
+}
+
+pub fn bldr_url_from_env_load_or_default() -> String {
+    bldr_url_from_env().unwrap_or_else(|| {
+                           match config::load() {
+                               Ok(config) => {
+                                   match config.bldr_url {
```
davidMcneil, line 67: We could use `unwrap_or`

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
+                                   println!("Found a cli.toml but unable to load it. Resorting \
+                                             to default BLDR_URL: {}",
+                                            e);
+                                   DEFAULT_BLDR_URL.to_string()
+                               }
+                           }
+                       })
+}
+
+pub fn bldr_url_from_args_env_load_or_default(opt: Option<Url>) -> String {
```
davidMcneil, line 82: It is a bummer we lose type information here (ie `Url` to `String`). Could we make this return a `Url`?

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
+                match config::load()?.auth_token {
+                    Some(v) => Ok(v),
+                    None => {
+                        Err(Error::ArgumentError("No auth token specified. Please \
+                                                  check that you have specified a \
+                                                  valid Personal Access Token with:  \
+                                                  -z, --auth <AUTH_TOKEN>"
+                                                                          .into()))
+                    }
+                }
```
davidMcneil, line 106: We could use `ok_or`

components/hab/src/cli/hab/util.rs
davidMcneil: The config file should probably require `bldr_url` be a `Url`, correct?

components/core/src/util/text_render.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub trait TabularText {
+    fn as_tabbed(&self) -> Result<String>;
+}
+
+pub trait PortableText: Serialize {
+    fn as_json(&self) -> Result<Json> {
+        serde_json::to_value(self).map_err(Error::RenderContextSerialization)
+    }
+}
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/core/src/origin.rs
```diff
@@ -0,0 +1,86 @@
...
+    type Err = Error;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        if s.is_empty() {
+            Err(Error::InvalidOrigin(s.to_string()))
+        } else {
+            Ok(Origin(s.to_string()))
+        }
+    }
+}
```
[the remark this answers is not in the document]
jeremymv2: I added your suggestion in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/core/src/origin.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl fmt::Display for OriginMemberRole {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let value = match *self {
+            OriginMemberRole::ReadonlyMember => "readonly_member",
+            OriginMemberRole::Member => "member",
+            OriginMemberRole::Maintainer => "maintainer",
+            OriginMemberRole::Administrator => "administrator",
+            OriginMemberRole::Owner => "owner",
```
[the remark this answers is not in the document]
jeremymv2: I added the constants in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/core/Cargo.toml
```diff
@@ -13,6 +13,7 @@ cc = "*"
 [dependencies]
 base64 = "*"
 chrono = "*"
+configopt = { git = "https://github.com/davidMcneil/configopt.git" }
```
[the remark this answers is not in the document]
jeremymv2: Nope. Removed in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/hab/src/cli/hab/util.rs
jeremymv2: Indeed. Added in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/hab/src/cli/hab/util.rs
jeremymv2: Modified as such in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/hab/src/cli/hab/util.rs
jeremymv2: True. Added in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/hab/src/cli/hab/util.rs
jeremymv2: Modified as such in https://github.com/habitat-sh/habitat/pull/7841/commits/633e75036f3864b3ac7d2cfbf02a8a4704ce4120

components/builder-api-client/src/lib.rs
```diff
@@ -228,35 +230,8 @@ mod json_date_format {
...
-
-// Returns a library object that implements elastic tabstops
-fn tabw() -> TabWriter<Vec<u8>> { TabWriter::new(Vec::new()) }
-
-// Given a TabWriter object and a str slice, return a Result
-// where the Ok() variant comprises a String with nicely tab aligned columns
-fn tabify(mut tw: TabWriter<Vec<u8>>, s: &str) -> Result<String> {
-    write!(&mut tw, "{}", s)?;
-    tw.flush()?;
-    String::from_utf8(tw.into_inner().expect("TABWRITER into_inner")).map_err(|e| {
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/7841/commits/6efd8ad8c2ced3e02a72cc66e0e80c633e933f36

components/hab/src/cli/hab/util.rs
jeremymv2: The function now returns a `Url` type instead of `String`. Take a look and see if what's there now looks agreeable.

components/core/src/origin.rs
```diff
@@ -0,0 +1,89 @@
...
+                   Serialize};
+use std::{fmt,
+          result,
+          str::FromStr};
+
+pub const ORIGIN_MEMBER_ROLE_READONLY_MEMBER: &str = "readonly_member";
+pub const ORIGIN_MEMBER_ROLE_MEMBER: &str = "member";
+pub const ORIGIN_MEMBER_ROLE_MAINTAINER: &str = "maintainer";
+pub const ORIGIN_MEMBER_ROLE_ADMINISTRATOR: &str = "administrator";
+pub const ORIGIN_MEMBER_ROLE_OWNER: &str = "owner";
```
davidMcneil, line 13: No need to change anything, but I am personally a fan of asciating constants like this with the type. So something like:

```rust
impl OriginMemberRole {
    const READONLY_MEMBER: &'static str = "readonly_member";
    const MEMBER: &'static str = "member";
    const MAINTAINER: &'static str = "maintainer";
    const ADMINISTRATOR: &'static str = "administrator";
    const OWNER: &'static str = "owner";
}
```

You can then access them with `OriginMemberRole::READONLY_MEMBER`

components/hab/src/cli/hab/util.rs
```diff
@@ -18,29 +27,99 @@ use structopt::StructOpt;
...
 }
 
 #[derive(ConfigOpt, StructOpt, Deserialize)]
 #[configopt(derive(Serialize))]
 #[structopt(no_version)]
 #[allow(dead_code)]
 pub struct BldrUrl {
-    /// Specify an alternate Builder endpoint. If not specified, the value will be taken from
-    /// the HAB_BLDR_URL environment variable if defined. (default: https://bldr.habitat.sh)
-    // TODO (DM): This should probably use `env` and `default_value`
```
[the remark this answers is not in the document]
jeremymv2: Sure! Added back in https://github.com/habitat-sh/habitat/pull/7841/commits/80dffb74528b538f9c81acc0118e7413594defca

components/core/src/origin.rs
jeremymv2: Totally agree. Adjusted in https://github.com/habitat-sh/habitat/pull/7841/commits/9aee65a4697d2a584ada86c503987c6b5abd8120

jeremymv2:
@davidMcneil I believe I've taken care of all the feedback items thus far. Care to sanity check me? ❤️

components/core/src/origin.rs
davidMcneil: I like your change. Although, it was not quite what I had in mind :smile:. My intention was to scope the constants within the `OriginMemberRole` so they should be defined in an `impl OriginMemberRole { }` block and accessed with `OriginMemberRole::READONLY_MEMBER`. But I dont think there is any need to change anything for this case.

components/hab/src/cli/hab/origin.rs
```diff
@@ -235,6 +240,51 @@ pub struct UploadGroup {
...
+
+#[derive(ConfigOpt, StructOpt)]
+#[structopt(no_version, rename_all = "screamingsnake")]
+pub struct RbacSet {
+    #[structopt(flatten)]
+    pub origin:         BldrOrigin,
+    /// The account name whose role will be changed
+    pub member_account: String,
+    /// The role name to enforce for the member account [values: readonly_member, member,
+    /// maintainer, administrator, owner]
```
davidMcneil, line 267: We could use `#[structopt(possible_values = &[OriginMemberRole::READONLY_MEMBER ...])]` which will automatically display the possible values and has better traceability should we ever add an OriginMemberRole` variant.

components/core/src/util/text_render.rs
```diff
@@ -0,0 +1,37 @@
...
+    write!(&mut tw, "{}", s)?;
+    tw.flush()?;
+    let res = tw.into_inner();
+    if res.is_err() {
+        return Err(Error::TabWriterIntoInnerFailed("Unable to flush \
+                                                    tabwriter buffer to \
+                                                    inner."
+                                                           .to_string()));
+    }
+    let inner = res.unwrap();
```
davidMcneil, line 23: I dont think we should have a custom `TabWriterIntoInnerFailed` type. Instead we could use the `io::Error` that is part of [`tabwriter::IntoInnerError`](https://docs.rs/tabwriter/1.2.1/tabwriter/struct.IntoInnerError.html#method.error) 

We could use:

```rust
    let inner = tw.into_inner().map_err(|e| {
                                    IoError::new(e.error().kind(),
                                                 "Unable to flush tabwriter buffer to inner.")
                                })?;
```

This preserves the `io::Error` kind, prevents us from having to create a new type, and avoids an unwrap.

components/core/src/error.rs
```diff
@@ -340,6 +347,7 @@ impl fmt::Display for Error {
             Error::GetExitCodeProcessFailed(ref e) => e.to_string(),
             Error::CreateToolhelp32SnapshotFailed(ref e) => e.to_string(),
             Error::WaitForSingleObjectFailed(ref e) => e.to_string(),
+            Error::TabWriterIntoInnerFailed(ref e) => e.to_string(),
```
davidMcneil, line 350: See other comments. I dont think this type should be needed.

components/hab/src/error.rs
```diff
@@ -48,6 +48,7 @@ pub enum Error {
     HabitatCore(hcore::Error),
     // Boxed due to clippy::large_enum_variant
     HandlebarsRenderError(Box<handlebars::TemplateRenderError>),
+    InvalidUrl(String),
```
davidMcneil, line 51: I dont think we should add this variant. See other comments.

components/hab/src/cli/hab/util.rs
```diff
@@ -18,24 +25,84 @@ use structopt::StructOpt;
...
+                       })
+}
+
+pub fn bldr_url_from_args_env_load_or_default(opt: Option<Url>) -> Result<Url, Error> {
+    if let Some(url) = opt {
+        Ok(url)
+    } else {
+        Url::parse(&bldr_url_from_env_load_or_default()).map_err(|e| {
+                                                            Error::InvalidUrl(e.to_string())
+                                                        })
```
davidMcneil, line 80: Instead of mapping the error here, because this is the only possible error the function can return can we change the `Error` type the function returns?

components/hab/src/cli/hab/util.rs
```diff
@@ -18,24 +25,84 @@ use structopt::StructOpt;
...
+#[derive(ConfigOpt, StructOpt, Deserialize, Serialize)]
+#[structopt(no_version)]
+#[configopt(derive(Serialize))]
+pub struct BldrOrigin {
+    /// The Builder origin name to target
+    #[structopt(name = "ORIGIN", short = "o", long = "origin")]
+    pub inner: Origin,
+}
+
+pub fn bldr_url_from_env_load_or_default() -> String {
```
davidMcneil, line 57: This could be private.

components/hab/src/cli/hab/util.rs
```diff
@@ -18,24 +25,84 @@ use structopt::StructOpt;
...
+            Err(_) => {
+                config::load()?.auth_token.ok_or_else(|| {
+                                              Error::ArgumentError("No auth token specified. \
+                                                                    Please check that you have \
+                                                                    specified a valid Personal \
+                                                                    Access Token with:  -z, \
+                                                                    --auth <AUTH_TOKEN>"
+                                                                                        .into())
+                                          })
+            }
```
davidMcneil, line 99: I think this could use `or_else`

components/hab/src/command/origin/rbac/show_role.rs
```diff
@@ -0,0 +1,64 @@
...
+                        Ok(())
+                    }
+                    Err(e) => {
+                        ui.fatal(format!("Failed to deserialize into json! {:?}.", e))?;
+                        Err(Error::from(e))
+                    }
+                }
+            } else {
+                ui.status(Status::Discovering, "origin member role".to_string())?;
+                println!("Member {} has the '{}' role in the {} origin.",
```
davidMcneil, line 41: Should this be a `println`? Maybe a `ui.status` or `debug!` instead?

components/core/src/origin.rs
jeremymv2: ahh yes. I'll make it so.

components/hab/src/command/origin/rbac/show_role.rs
jeremymv2: Added `ui.status` in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/hab/src/cli/hab/util.rs
jeremymv2: Changed to private in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/hab/src/cli/hab/util.rs
jeremymv2: removed the error mapping in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/hab/src/error.rs
jeremymv2: Variant removed in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/core/src/error.rs
jeremymv2: Removed in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/core/src/util/text_render.rs
jeremymv2: Modified as such in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/hab/src/cli/hab/origin.rs
jeremymv2: This is a nice capability. Added in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/hab/src/cli/hab/util.rs
jeremymv2: I believe this has to be `ok_or_else` since `config::load` is returning an option for `auth_token`. If I change to `or_else` clippy says:

```
error[E0308]: mismatched types
  --> components/hab/src/cli/hab/util.rs:90:17
   |
83 |   pub fn bldr_auth_token_from_args_env_or_load(opt: Option<String>) -> Result<String, Error> {
   |                                                                        --------------------- expected `std::result::Result<std::string::String, error::Error>` because of return type
...
90 | /                 config::load()?.auth_token.or_else(|| {
91 | |                                               Error::ArgumentError("No auth token specified. \
92 | |                                                                     Please check that you have \
93 | |                                                                     specified a valid Personal \
...  |
96 | |                                                                                         .into())
97 | |                                           })
   | |____________________________________________^ expected enum `std::result::Result`, found enum `std::option::Option`
   |
   = note: expected enum `std::result::Result<std::string::String, error::Error>`
              found enum `std::option::Option<std::string::String>`
```

I need to return an result so that the CLI marshaling can error if no auth token is found and one is required. We need to figure out how to handle this natively in the future (loading from args, ENV, OR cli.toml. See: https://github.com/habitat-sh/habitat/issues/7866

For now we need to return a Result so that we can error appropriately before calling the `start` function of the command:
```let auth_token = bldr_auth_token_from_args_env_or_load(r.auth_token.value)?;```

components/core/src/origin.rs
jeremymv2: I've modified in https://github.com/habitat-sh/habitat/pull/7841/commits/78f9d5e6a2b76e641bb6f2a2971c62744b5e762b

components/hab/src/cli/hab/util.rs
davidMcneil: Yeah, sorry I was not clear. I meant this part `match henv::var(AUTH_TOKEN_ENVVAR)` we could probably avoid the need to `match` on the `Ok` just to wrap it back up in `Ok`, but it would not really buy us that much.

PR #7844, opened 2020-07-29
refactor account/sid/win_perms to not need to contact a domian controller

### LINKED ISSUE #7836: panic at 'local administrator account'
A customer reported a Supervisor crash with this log line:

```
2020-07-14 21:12:46,560 - thread 'main' panicked at 'local administrator account', src\libcore\option.rs:1185:5
```

[This is where](https://github.com/habitat-sh/habitat/blob/bca095d26759996ff73f51a7a39b77f45aecbf54/components/core/src/util/win_perm.rs#L89) we get the built in administrator account to add the ACL to one of the service directories and call [built_in_administrators](https://github.com/habitat-sh/habitat/blob/bca095d26759996ff73f51a7a39b77f45aecbf54/components/win-users/src/account.rs#L113) to get the account.

It's very odd that this would fail because we are retrieving the account of a well known SID and unfortunately given the use of `expect` we lose the underlying error. My hunch is that the call to `LookupAccountSidW` resultedt in an OS error 1789 (The trust relationship between this workstation and the primary domain failed). We have [seen that before](https://github.com/habitat-sh/habitat/issues/5258) with `LookupAccountNameW`. Still, I hold this hunch loosely.

I think we could have `from_sid` return a `Result<Account>` and bubble up the underlying error. I also think in this particular scenario, we could get away with not calling `LookupAccountSidW` at all. To set the perm, we just need the SID so we could add a way to refactor `PermissionEntry` ro just encapsulate the SID instead of the `Account`.

mwrock:
fixes #7836

In some scenarios where a supervisor runs on a node that is domain joined, some account lookup functions may attempt to reach out to the domain controller for account information. In a scenario where a node is being provisioned or perhaps shortly after a
reboot, this may fail. However for the accounts that we typically deal with when setting folder ACLs, the information we need should not require ant DC interaction. We only need the SID and these are for "well known" users or the currently logged in user.

This PR simply retrieves the SIDs. There was also some refactoring of `Sid::from_current_user` which was previously plain wrong and fortunately not used, but it is used now. It was using the "logon SID" and not the user SID. The comments go into the details on what is a logon sid.

### DIFF
--- a/components/core/src/os/process/windows_child.rs
+++ b/components/core/src/os/process/windows_child.rs
@@ -931,7 +931,7 @@ fn create_process_as_user(credential: &ServiceCredential,
                                                         io::Error::last_os_error())));
         }
 
-        let sid = Sid::from_token(token)?;
+        let sid = Sid::logon_sid_from_token(token)?;
         sid.add_to_user_object(station as HANDLE,
                                sid::CONTAINER_INHERIT_ACE
                                | sid::INHERIT_ONLY_ACE
--- a/components/core/src/util/win_perm.rs
+++ b/components/core/src/util/win_perm.rs
@@ -1,6 +1,6 @@
 use crate::error::{Error,
                    Result};
-use habitat_win_users::account::Account;
+use habitat_win_users::sid::Sid;
 use std::path::Path;
 use widestring::WideCString;
 use winapi::{shared::{minwindef::DWORD,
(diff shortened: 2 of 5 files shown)


davidMcneil:
I wonder if we should cache the value of [`to_string`](https://github.com/habitat-sh/habitat/blob/38a04ce4e362ffab0955aca9eda454e615ce8db4/components/win-users/src/sid.rs#L224) in the `Sid` struct. We could then return an error if we can not convert the `Sid` to a `String` on `Sid` construction. We could then implement `Display` for `Sid`.

davidMcneil:
> I wonder if we should cache the value of to_string in the Sid struct.

Maybe not it looks like we only call `to_string()?` in one place

PR #7848, opened 2020-07-30
Add channel promotion datadog event

### LINKED ISSUE #7842: Add datadog event on supervisor release
Add a datadog event when the Supervisor is released to `dev`, `acceptance`, and `stable`. See example [here]( https://github.com/habitat-sh/builder/blob/master/.expeditor/scripts/post_datadog_merge_event.sh).

davidMcneil:
Resolves #7842

Adds logic to add a datadog event when a Supervisor version is promoted to a new Builder channel.

### DIFF
--- a/.expeditor/scripts/shared.sh
+++ b/.expeditor/scripts/shared.sh
@@ -224,6 +224,8 @@ promote_packages_to_builder_channel() {
 
     local manifest_json
     manifest_json=$(cat "${manifest}")
+    local version
+    version="$(echo "${manifest_json}" | jq -r '.version')"
 
     mapfile -t targets < <(echo "${manifest_json}" | jq -r ".packages | keys | .[]")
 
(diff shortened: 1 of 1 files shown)


.expeditor/scripts/shared.sh
```diff
@@ -239,7 +241,40 @@ promote_packages_to_builder_channel() {
         done
     done
 
+    send_channel_promotion_datadog_event "${version}" "${destination_channel}"
+}
 
+# Create a datadog event for the promotion of a Supervisor version to a builder channel. 
+# 
+# e.g. send_channel_promotion_datadog_event 0.88.0 dev
+send_channel_promotion_datadog_event() {
```
smacfarlane, line 250: Question:  Unless I've missed something, a failure here would fail the promotion event.  Is that the behavior we want? Are there other actions afterwords that could not be triggered leaving us in an partially-promoted state?

davidMcneil: Nice catch, thanks! Yes, I believe this would fail the entire pipeline. That is probably not the behavior we want. We can always retroctivly add the event in the case of failure.

[162679](https://github.com/habitat-sh/habitat/pull/7848/commits/162679b9fc266475bdfa41f8c7f645ad3c4cbe7b) should address this. Not sure if it is the best method. Happy to change if anyone has a better pattern.

smacfarlane: I think we'd still end up with failure if curl (or vault key fetch) were to exit non-zero, as the `set -e` would cause the script to exit immediately with that status.  We have a pattern in the cargo update https://github.com/habitat-sh/habitat/blob/master/.expeditor/scripts/finish_release/cargo_update.sh#L33 that isn't great but may be how we need to handle this. 

The vault key fetch _may_ be an error we want to halt on,as it could indicate an issue with our release infra., vs the curl which is external and informational only :thinking:   @christophermaier  Do you have any thoughts here?

davidMcneil: This is the script I was using to test this pattern:

```bash
set -euo pipefail

test_1() {
    echo "starting test_1"

    local EXIT_CODE=0
    test_2 || EXIT_CODE=$?
    if [ $EXIT_CODE -ne 0 ]; then
      echo "Failed to run test_2"
    fi

    echo "ending test_1"
}

test_2() {
    curl "this_is_bad_url"
}

test_1
```

I see what you are saying concerning why it might cause the whole script to exit, but it does not appear to behave that way :thinking: .

smacfarlane: Yeah, I  was using a similar script, but forgot the `||` on the first pass :facepalm: .  I think my only outstanding question is do we care in this context why it failed (vault vs curl) and should they behave differently.

Other than that question, this looks great and will be good data to have :+1:  I'd also be good with shipping this and deferring that question to later if it proves to be worth following up on.

PR #7856, opened 2020-08-05
fix e2e test with older launcher

mwrock:
Fixes failing test because supervisor stops before unloading package.

This also allows supervisor to survive any error from the launcher when checking version. It was failing because of an `UnknownMessage NerErr`. I think breaking it down by timeout, UnknownMessage and other is over kill and we should just write out the error message.

### DIFF
--- a/components/sup/src/manager/service/supervisor.rs
+++ b/components/sup/src/manager/service/supervisor.rs
@@ -21,8 +21,11 @@ use habitat_core::{fs,
                                  Pid},
                    service::ServiceGroup};
 #[cfg(windows)]
-use habitat_launcher_client::Error::Timeout;
+use habitat_launcher_client::Error as launcher_error;
 use habitat_launcher_client::LauncherCli;
+#[cfg(windows)]
+use habitat_launcher_protocol::{self as protocol,
+                                Error as launcher_protocol_error};
 use serde::{ser::SerializeStruct,
             Serialize,
             Serializer};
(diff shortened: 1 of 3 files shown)


components/sup/src/manager/service/supervisor.rs
```diff
@@ -198,13 +195,10 @@ impl Supervisor {
                     // of this comment.
                     Ok(v) if v > 14227 => pkg.svc_user.clone(),
                     Ok(_) => legacy_user,
-                    Err(err @ Timeout) => {
+                    Err(err) => {
                         error!("Error getting version from launcher: {:?}", err);
                         legacy_user
                     }
```
davidMcneil, line 201: On error, this assumes the launcher is "old", correct? This means that if it is really a "new" launcher we will get an error when we try to login as a different user.

It looks like the `UnknownMessage` response was [added somewhat recently](https://github.com/habitat-sh/habitat/blob/b48f8f443dc690e24dbf8adf5725fd84d04f97f1/components/launcher/src/server.rs#L510). This means there are really three kinds of launchers:

1. Only respond with `Timeout` on version check
2. Respond with `UnknownMessage` on version check unless there truly was a timeout
3. Respond with `Version` on version check unless there truly was a timeout or an `InvalidVersionNumber`

I agree it seems like we have to assume we are dealing with a legacy launcher on `UnknownMessage` or `Timeout` and just log an error message, but it seems like some errors (eg `InvalidVersionNumber`) it would be more helpful to return the error. Can we even get an `InvalidVersionNumer` error unless it is a new launcher?

christophermaier: It may also be beneficial to us debugging current installs to capture more information about the nature of the error.

mwrock: I now explicitly handle `Timeout` and `UnknownMessage` and bubble up any other error.

test/end-to-end/test_supervisor_windows_service.ps1
```diff
@@ -35,6 +35,7 @@ Describe "Habitat Windows Service" {
 
         AfterAll {
             hab svc unload core/nginx
+            Start-Sleep -Seconds 3 # tears
```
christophermaier, line 38: :laughing:

PR #7857, opened 2020-08-05
Cleanup `hab pkg export` subcommand

davidMcneil:
This PR cleans up the `hab pkg export` subcommand. It does three main things:

* Standarize `hab pkg export` command
* Parse `hab pkg export` with structopt
* Condition exporter platform support durring CLI parsing

See the commit messages for more details.


components/hab/src/cli/hab/pkg.rs
```diff
@@ -437,3 +425,130 @@ pub enum Pkg {
...
+        }
+    }
+}
+
+/// This struct can only be constructed within this module because it has a private field. All
+/// `ExportFormat` variants contain an instance of this types. This effectively "seals" construction
+/// of `ExportFormat` to this module. Construction should only happen in `FromStr` which uses
+/// conditional compilation to only allow constructing platform supported `ExportFormat`s. By
+/// "sealing" construction of this enum we avoid the need to litter conditional compilation wherever
+/// `ExportFormat` is used.
```
christophermaier, line 490: This is pretty slick; I'm a big fan of leveraging the type system like this to make the code cleaner and safer :+1:

components/hab/src/cli/hab/pkg.rs
```diff
@@ -437,3 +425,130 @@ pub enum Pkg {
...
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            #[cfg(any(target_os = "linux", target_os = "windows"))]
+            Self::CF => Ok(Self::Cf(SealedExportFormat(()))),
+            #[cfg(any(target_os = "linux", target_os = "windows"))]
+            Self::CONTAINER => Ok(Self::Container(SealedExportFormat(()))),
+            #[cfg(any(target_os = "linux", target_os = "windows"))]
+            Self::DOCKER => Ok(Self::Docker(SealedExportFormat(()))),
```
christophermaier, line 546: Since "docker" is an old deprecated alias of "container", could we just not have `Self::Docker` and map both strings to `Self::Container`?

davidMcneil: Great suggestion! Addressed in [d748fba](https://github.com/habitat-sh/habitat/pull/7857/commits/d748fbabef54092b166042e7f704d160e542bb82)

PR #7858, opened 2020-08-05
make bind timeout configurable and add debug output around startup

### LINKED ISSUE #7780: supervisor crashing on startup with Timeout waiting to bind to 0.0.0.0:9631
An excerpt from the supervisor log includes:

```
2020-06-19 19:58:40,433 - Timeout waiting to bind to 0.0.0.0:9631
2020-06-19 19:58:40,435 - [2020-06-20T02:58:40Z ERROR habitat_sup::ctl_gateway::acceptor] Error polling CtlAcceptor shutdown trigger: oneshot canceled
2020-06-19 19:58:40,435 - [2020-06-20T02:58:40Z ERROR habitat_sup::manager::service_updater] Service updater has gone away, yikes!
2020-06-19 19:58:40,435 - [2020-06-20T02:58:40Z ERROR habitat_sup::manager::service_updater] Service updater has gone away, yikes!
2020-06-19 19:58:40,435 - [2020-06-20T02:58:40Z ERROR habitat_sup::manager::service_updater] Service updater has gone away, yikes!
2020-06-19 19:58:40,435 - [2020-06-20T02:58:40Z ERROR habitat_sup::manager::service_updater] Service updater has gone away, yikes!
2020-06-19 19:58:43,600 - [2020-06-20T02:58:43Z ERROR hab_launch] Launcher exiting with code 86
```

It looks as though the supervisor [waits 10 seconds](https://github.com/habitat-sh/habitat/blob/b1786e51366b145a1483533098cdad5c518562da/components/sup/src/manager.rs#L987) for the http gateway to start and then throws a `BindTimeout` it has still not started. While one would assume 10 seconds to be adequate, a low resource instance that is CPU saturated at startup (this occurs upon server reboot) may very well hit this timeout.

We should probably extend this timeout and/or make it configurable.

mwrock:
fixes #7780 

This allows the hard coded 10 second timeout of the http_gateway to be configurable. At this point we want this to be fairly obscure just to allow users affected by #7780 to tune the timeout and determine if it indeed needs more time in their environment or if there might be something truly hanging their server. This also adds some debug messaging so that if it is hanging, we can get a better idea of where that hang might be occurring.

### DIFF
--- a/components/sup/src/manager.rs
+++ b/components/sup/src/manager.rs
@@ -154,6 +154,10 @@ lazy_static! {
         PackageIdent::from_str(&format!("{}/{}", SUP_PKG_IDENT, VERSION)).unwrap();
 }
 
+habitat_core::env_config_duration!( HttpStartupTimeout,
+                                    HAB_HTTP_STARTUP_TIMEOUT_SECS => from_secs,
+                                    Duration::from_secs(10));
+
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 /// Determines whether the new pidfile-less behavior is enabled, or
 /// the old behavior is used.
@@ -994,19 +998,22 @@ impl Manager {
             loop {
                 match *started {
                     http_gateway::ServerStartup::NotStarted => {
-                        started = match cvar.wait_timeout(started, Duration::from_secs(10)) {
-                            Ok((mutex, timeout_result)) => {
-                                if timeout_result.timed_out() {
-                                    return Err(Error::BindTimeout(http_listen_addr.to_string()));
-                                } else {
-                                    mutex
+                        started =
+                            match cvar.wait_timeout(started,
+                                                    HttpStartupTimeout::configured_value().into())
+                            {
+                                Ok((mutex, timeout_result)) => {
+                                    if timeout_result.timed_out() {
+                                        return Err(Error::BindTimeout(http_listen_addr.to_string()));
+                                    } else {
+                                        mutex
+                                    }
                                 }
-                            }
-                            Err(e) => {
-                                error!("Mutex for the HTTP gateway was poisoned. e = {:?}", e);
-                                return Err(Error::LockPoisoned);
-                            }
-                        };
+                                Err(e) => {
+                                    error!("Mutex for the HTTP gateway was poisoned. e = {:?}", e);
+                                    return Err(Error::LockPoisoned);
+                                }
+                            };
                     }
                     http_gateway::ServerStartup::BindFailed => {
                         return Err(Error::BadAddress(http_listen_addr.to_string()));
--- a/components/sup/src/http_gateway.rs
+++ b/components/sup/src/http_gateway.rs
@@ -226,6 +226,7 @@ impl Server {
                feature_flags: FeatureFlag,
                control: Arc<(Mutex<ServerStartup>, Condvar)>) {
         thread::spawn(move || {
+            debug!("Entering http_gateway run thread");
             let &(ref lock, ref cvar) = &*control;
             let thread_count = match henv::var(HTTP_THREADS_ENVVAR) {
                 Ok(val) => {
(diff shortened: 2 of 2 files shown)


components/sup/src/main.rs
```diff
@@ -712,6 +718,7 @@ gpoVMSncu2jMIDZX63IkQII=
                                        ctl_listen:            ListenCtlAddr::default(),
                                        http_listen:           HttpListenAddr::default(),
                                        http_disable:          false,
+                                       http_startup_timeout:  Duration::from_secs(10),
```
christophermaier, line 721: Why not change all these to `HttpStartupTimeout::default()`?

components/sup/src/main.rs
```diff
@@ -288,6 +292,8 @@ async fn split_apart_sup_run(sup_run: SupRun,
                               update_url: bldr_url.clone(),
                               update_channel: shared_load.channel.clone(),
                               http_disable: sup_run.http_disable,
+                              http_startup_timeout:
+                                  HttpStartupTimeout::configured_value().into(),
```
christophermaier, line 296: If you keep `http_startup_timeout` typed as `HttpStartupTimeout`, this could just be `HttpStartupTimeout::configured_value()`; you only really need the `into()` at the point where it's actually time to wait. You don't lose anything, and you actually make it easier to track all the information flows and make sure they're consistent.

mwrock: Nice! Applied in bddeedd7fa6522fa2d2b1592d9518c80e19b6cf6 @christophermaier

components/core/src/env.rs
```diff
@@ -154,9 +154,9 @@ macro_rules! env_config {
 /// ```
 #[macro_export]
 macro_rules! env_config_duration {
-    ($(#[$attr:meta])* $wrapping_type:ident, $env_var:ident => $from_str_fn:ident, $default_value:expr) => {
+    ($(#[$attr:meta])* $vis:vis $wrapping_type:ident, $env_var:ident => $from_str_fn:ident, $default_value:expr) => {
```
christophermaier, line 157: :thinking: Why did you end up having to add this?

components/sup/src/main.rs
```diff
@@ -1480,6 +1493,7 @@ organization = "MY_ORG_FROM_SECOND_CONFG"
                                        http_listen:
                                            HttpListenAddr::from_str("3.3.3.3:3333").unwrap(),
                                        http_disable:          false,
+                                       http_startup_timeout:  HttpStartupTimeout::configured_value(),
```
christophermaier, line 1496: In all these tests, is it the intention that the HTTP startup timeout, _as a fundamental part of the test itself_, be sensitive to the value of its environment variable?

If that is _not_ the case, and the value of the environment variable is not important to the test, these `configured_value()` calls should be replaced with `default()`, as that always returns a stable value. 

`configured_value()` should only be used in tests where you're actually looking at how different values of the environment variable affect the behavior you're testing.

components/core/src/env.rs
mwrock: I needed to declare HttpStartupTimeout as `pub` because the `http_startup_timeout` field is public. Thats what the compiler told me to do.

components/sup/src/main.rs
mwrock: applied in ba5c5ffa74d08bdc53af7733f335fcc78bd893e5 @christophermaier

components/sup/src/manager.rs
```diff
@@ -994,7 +1000,12 @@ impl Manager {
             loop {
                 match *started {
                     http_gateway::ServerStartup::NotStarted => {
-                        started = match cvar.wait_timeout(started, Duration::from_secs(10)) {
+                        started = match cvar.wait_timeout(started,
+                                                          self.state
+                                                              .cfg
+                                                              .http_startup_timeout
+                                                              .into())
```
davidMcneil, line 1007: Strictly speaking we do not need to add `http_startup_timeout` to `ManagerConfig`, correct? We could simply use `HttpStartupTimeout::configured_value()` here. I think it is fine to leave in `ManagerConfig` because it sets us up to easily add `http_startup_timeout` as a cli option in the future, but I thought I would bring it up as a potential simplification.

mwrock: OMG!!! You are so right. I think when I started this I did not know that the `env_config` just creates a type and was not aware it would be accesible to manager.

PR #7865, opened 2020-08-10
Improve parsing of `hab pkg export`

davidMcneil:
This is an improvement and follow-up to #7857. With a closer reading of the `clap` docs I was able to come up with a work around to the problem discussed [here](https://github.com/habitat-sh/habitat/blob/a6eb21bebf086422bc38ee51a5fa7808b4c24b8d/components/hab/src/cli/hab/pkg.rs#L447). This makes it so that exporters can simply be subcommands and all parsing logic can be handled by `clap`/`structopt`. The secret was use of `AppSettings::AllowLeadingHyphen`.


.expeditor/end_to_end.pipeline.yml
```diff
@@ -858,6 +857,36 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
 
+  - label: "[:linux: test_pkg_exporter_help]"
+    command:
+      - bash .expeditor/scripts/end_to_end/run_e2e_test.sh dev test_pkg_exporter_help
```
jeremymv2, line 862: Did you happen to omit the `test_pkg_exporter_help.ps1` file in this PR perhaps?

davidMcneil: No, I actually added [it](https://github.com/habitat-sh/habitat/blob/795cd2fc4a2d59e1d84b19c75e75488bbabafe75/test/end-to-end/test_pkg_exporters_help.ps1#L1) in a previous PR but then forgot to actually add it to the pipeline :sweat_smile: . It should be on master.

jeremymv2: Ahh, the reason I didn't find it on master is that it's named **exporters** : `./test/end-to-end/test_pkg_exporters_help.ps1`. In this PR it's referenced as `test_pkg_exporter_help`.

davidMcneil: Aha, nice catch, thanks! Addressed in [ad86cc](https://github.com/habitat-sh/habitat/pull/7865/commits/ad86ccc5c9828de1aa63c60e3ee65a8fe9e2d0a0)

PR #7871, opened 2020-08-13
Pull packages from the `--channel` argument on `hab sup run`

### LINKED ISSUE #7833: `hab sup run --channel` should pull `hab-sup` and `hab-launch` from the specified channel
Currently, the `--channel` attribute only applies to `hab-sup` updates. It should also apply when first trying to install `hab-sup` and `hab-launch`. The fact that it does not requires explicitly running `hab pkg install --channel <channel> core/hab-sup` before running `hab sup run` when using anything but the `stable` channel.

davidMcneil:
Resolves #7833

This PR has two parts:

1. Make it impossible to run the supervisor on MacOS using conditional compilation on the CLI. Previously the user would get an error when issuing `hab sup run` (as well as a few other commands) on MacOS now that CLI option is not even available.
2. Pull `core/hab-sup` and `core/hab-launch` from the `--channel` argument on `hab sup run`

Future work could add an `--internal-channel` argument to `hab sup run` to avoid the overloading of the `--channel` argument. It is used for both pulling internal packages and the package loaded on Supervisor startup. Note, the `--channel` argument was overloaded before this PR by the `hab-sup` updater.

### DIFF
--- a/components/hab/src/exec.rs
+++ b/components/hab/src/exec.rs
@@ -25,6 +25,21 @@ use std::path::PathBuf;
 const RETRY_LIMIT: usize = 5;
 const INTERNAL_TOOLING_CHANNEL_ENVVAR: &str = "HAB_INTERNAL_BLDR_CHANNEL";
 
+pub async fn command_from_min_pkg(ui: &mut UI,
+                                  command: impl Into<PathBuf>,
+                                  ident: &PackageIdent)
+                                  -> Result<PathBuf> {
+    command_from_min_pkg_with_optional_channel(ui, command, ident, None).await
+}
+
+pub async fn command_from_min_pkg_with_channel(ui: &mut UI,
+                                               command: impl Into<PathBuf>,
+                                               ident: &PackageIdent,
+                                               channel: ChannelIdent)
+                                               -> Result<PathBuf> {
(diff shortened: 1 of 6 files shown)


components/hab/src/exec.rs
```diff
@@ -35,9 +50,12 @@ const INTERNAL_TOOLING_CHANNEL_ENVVAR: &str = "HAB_INTERNAL_BLDR_CHANNEL";
 /// # Notes on Package Installation
 ///
 /// By default, Habitat will install packages from the `stable`
-/// channel. However, if you'd rather use unstable (particularly if
-/// you're developing Habitat), you'll need to set the
-/// `INTERNAL_TOOLING_CHANNEL_ENVVAR` appropriately.
+/// channel or will use a channel specific to a particular context (eg
+/// `hab sup run --internal-channel`) passed with the `channel`
```
christophermaier, line 54: I don't think we have `--internal-channel` yet, right?

davidMcneil: Oops, thanks! [7c093d3](https://github.com/habitat-sh/habitat/pull/7871/commits/7c093d36863454f4f5da4afca50687efefe87d65)

Issue #7619, opened 2020-04-17
`hab pkg install` should not retry HTTP 404 errors


stevendanna:
When attempting to download a non-existent package with a fully-qualified ident, the user will currently see:

```
> hab pkg install core/foobar/1/1
» Installing core/foobar/1/1
↓ Downloading core/foobar/1/1
↓ Downloading core/foobar/1/1
↓ Downloading core/foobar/1/1
↓ Downloading core/foobar/1/1
↓ Downloading core/foobar/1/1
↓ Downloading core/foobar/1/1
✗✗✗
✗✗✗ We tried 5 times but could not download core/foobar/1/1. Last error was: description() is deprecated; use Display
✗✗✗
```

The combination of the unhelpful error message and the fact that hab attempted to retry the request, might indicate to the user that there is some server-side or network problem.  In reality, the download was never going to work:

```
↓ Downloading core/foobar/1/1
[2020-04-17T09:51:29Z DEBUG habitat_http_client::api_client] GET https://bldr.habitat.sh/v1/depot/pkgs/core/foobar/1/1/download?target=x86_64-linux with ApiClient { endpoint: "https://bldr.habitat.sh/v1", inner: Client { gzip: false, brotli: false, proxies: [Proxy(Custom(_))], referer: true, default_headers: {"accept": "*/*", "user-agent": "hab/1.5.86/20200408002109 (x86_64-linux; 5.2.18-100.fc29.x86_64)", "connection": "close"}, timeout: 300s } }
[2020-04-17T09:51:29Z DEBUG habitat_api_client::builder] Downloading file to path: /home/vagrant/.hab/cache/artifacts
[2020-04-17T09:51:29Z DEBUG reqwest::connect] starting new connection: https://bldr.habitat.sh/
[2020-04-17T09:51:29Z DEBUG hyper::client::connect::dns] resolving host="bldr.habitat.sh"
[2020-04-17T09:51:29Z DEBUG hyper::client::connect::http] connecting to 151.101.17.209:443
[2020-04-17T09:51:29Z DEBUG hyper::client::connect::http] connected to 151.101.17.209:443
[2020-04-17T09:51:29Z DEBUG hyper::proto::h1::io] flushed 209 bytes
[2020-04-17T09:51:29Z DEBUG hyper::proto::h1::io] read 449 bytes
[2020-04-17T09:51:29Z DEBUG hyper::proto::h1::io] parsed 17 headers
[...]

Issue #7671, opened 2020-05-05
Improve API retry logic around package installation, and beyond


christophermaier:
Recently, a handful of issues relating in one way or another to API retry logic have popped up:

* https://github.com/habitat-sh/habitat/issues/7639
* https://github.com/habitat-sh/habitat/issues/7619
* https://github.com/habitat-sh/habitat/issues/7670

The first two deal with retries that aren't sufficiently smart (e.g., repeatedly making requests that always return 404). That is, retrying API calls should be limited to situations where a retry _of the same request_ is likely to result in a success, as with request timeouts. If you get a `Forbidden` result from an API call, trying the call four more times with the same credentials is unlikely to result in success.

The third issue deals with the fact that while we have retry logic around the API call to fetch packages, we don't have it around other API calls that are also in the package installation flow. It doesn't make sense for only part of the installation flow to be robust against network timeouts, and any retries we add here would also need to be sufficiently "smart" to be truly useful.

This suggests that we might want to add retriability as part of our API client crate itself. This would ensure that the proper logic is added to where it is most directly needed, and would allow callers to easily adopt the desired amount of retries by calling a new function, or perhaps passing an `Option<uint>` argument to existing functions indicating the number of retries.

sdelano:
From team discussion:

There are some areas where we wrap the low-level HTTP calls w/ retries and we should survey the code so that we don't end up with multiple levels of nested retries.

jeremymv2:
https://github.com/habitat-sh/habitat/pull/7874 establishes a new pattern for HTTP retry attempts, but there is still plenty to do to resolve the entirety of this issue.

For starters:
- [ ] use `retry_builder_api!` throughout the code base on all remaining API calls that are yet not wrapped in retry logic
- [ ] refactor the [pkg upload](https://github.com/habitat-sh/habitat/blob/master/components/hab/src/command/pkg/upload.rs) logic to use `retry_builder_api` but also take a stab a reducing that commands complexity
- [ ] smart resumption of interrupted downloads (although possibly outside the scope defined here as it entails changes to Builder)

PR #7874, opened 2020-08-14
an API retry pattern

jeremymv2:
Fixes https://github.com/habitat-sh/habitat/issues/7619

Behavioral changes:
- We no longer attempt a HTTP retry on HTTP errors that are effectively unrecoverable (400-499)
- Replace the linear back off interval between retry attempts with a new algorithm using a Fibonacci sequence. Previously the client would retry over a total duration of 15 seconds. Now the retries occur over a total duration of 120 seconds, initially closely grouped, then further spread apart.

This is an initial step to more reasonably bounded but more resilient HTTP API retries. There's still plenty more to do as detailed in https://github.com/habitat-sh/habitat/issues/7671#issuecomment-675686030


components/builder-api-client/src/builder.rs
```diff
@@ -57,6 +58,9 @@ const X_FILENAME: &str = "x-filename";
 
 const DEFAULT_API_PATH: &str = "/v1";
 
+pub const API_RETRIES: usize = 5;
+pub const API_RETRY_WAIT: Duration = Duration::from_millis(10000);
```
christophermaier, line 62: Unless it's really important to use the scale of milliseconds here, this would be more readable as `Duration::from_secs(10)`.

PLUS YOU GET TO SAVE THREE ZEROES!!!! :tada:

components/hab/src/command/origin/key/upload.rs
```diff
@@ -31,58 +31,64 @@ pub async fn start(ui: &mut UI,
...
-          .map_err(|_| {
-              Error::from(api_client::Error::UploadFailed(format!("We tried {} times but could \
-                                                                   not upload {}/{} public \
-                                                                   origin key. Giving up.",
-                                                                  RETRIES, &name, &rev)))
-          })?;
+                Ok::<_, habitat_api_client::error::Error>(())
+            }).await
+        {
+            return Err(common::error::Error::DownloadFailed(format!("When suitable, we try \
```
jeremymv2, line 52: This error should be `UploadFailed`

components/hab/src/command/origin/key/upload.rs
```diff
@@ -31,58 +31,64 @@ pub async fn start(ui: &mut UI,
...
-                                                                  RETRIES, &name, &rev)))
-          })?;
+        if let Err(e) =
+            retry_builder_api!(None, async {
+                ui.status(Status::Uploading, secret_keyfile.display())?;
+                api_client.put_origin_secret_key(&name, &rev, secret_keyfile, token, ui.progress())
+                          .await
+            }).await
+        {
+            return Err(common::error::Error::DownloadFailed(format!("When suitable, we try once then \
```
jeremymv2, line 81: `UploadFailed`

components/builder-api-client/src/builder.rs
```diff
@@ -1426,6 +1430,45 @@ impl BuilderAPIClient {
...
+                    Err(e @ api_client::Error::APIError(StatusCode::UNAUTHORIZED, _)) => {
+                        retry::OperationResult::Err(e)
+                    }
+                    Err(e @ api_client::Error::APIError(StatusCode::FORBIDDEN, _)) => {
+                        retry::OperationResult::Err(e)
+                    }
+                    Err(e @ api_client::Error::APIError(StatusCode::UNPROCESSABLE_ENTITY, _)) => {
+                        retry::OperationResult::Err(e)
+                    }
+                    Err(e) => retry::OperationResult::Retry(e),
```
christophermaier, line 1465: Are there going to be specific errors you'll return in the future (when this is out of draft status) for each of these cases? As currently written, everything is basically the same, allowing you to collapse most of these match arms.

components/builder-api-client/src/builder.rs
```diff
@@ -1426,6 +1430,45 @@ impl BuilderAPIClient {
     }
 }
 
+/// Retry API operations within reasonable bounds.
+/// Provide an Option with a custom back-off algorithm from the `retry` crate
+/// or a default Fibonacci algorithm will be used. Unrecoverable HTTP errors
+/// will not be retried.
+#[macro_export]
+macro_rules! retry_builder_api {
+    ($iter_opt:expr, $api_future:expr) => {
```
christophermaier, line 1439: Declarative macros like this can match multiple different cases. As such, you actually don't have to invoke this with an explicit `None` in the first position if you want to use the Fibonacci delay strategy; you could just have that be another match case of the macro.

So, to use the default strategy, you could call it like:
```rust
retry_builder_api!(future);
```
and to call it with the custom iterator, you'd use

```rust
retry_builder_api!(iterator, future)
```
If you want to get fancier, you could tweak the macro for readability, so you could call it as something like:
```rust
retry_builder_api!(future, with_custom_iterator: iterator);
```
(As currently coded, I don't think this has to be a macro, but it's probably a relatively cheap way to get optional parameters as described above without having to create some new types, so that's fine.)

davidMcneil: > As currently coded, I don't think this has to be a macro

I believe `retry_builder_api` has to be a macro because futures can only be `await`ed once. We dont want to resolve the expression that was passed into `retry_builder_api`. That would only give is the ability to try the future once. Instead, we want to preserve the expression used to generate the future so we can call that expression multiple times.

This all stems from a lack of true async closures. If we had async closures we could pass a closure to generate the future instead of resorting to a macro.

components/builder-api-client/src/builder.rs
```diff
@@ -1426,6 +1430,45 @@ impl BuilderAPIClient {
...
+                    }
+                    Err(e @ api_client::Error::APIError(StatusCode::FORBIDDEN, _)) => {
+                        retry::OperationResult::Err(e)
+                    }
+                    Err(e @ api_client::Error::APIError(StatusCode::UNPROCESSABLE_ENTITY, _)) => {
+                        retry::OperationResult::Err(e)
+                    }
+                    Err(e) => retry::OperationResult::Retry(e),
+                }
+            }).await
```
davidMcneil, line 1467: At the end we should probably pull out the last error instead of leaving it wrapped in a `retry::OperationResult`. I think you should be able to just use `.error` here.

components/common/src/command/package/install.rs
```diff
@@ -663,14 +665,8 @@ impl<'a> InstallTask<'a> {
...
-            retry::retry_future!(delay::Fixed::from(RETRY_WAIT).take(RETRIES),
-                                 self.fetch_artifact(ui, (ident, target), token)).await
-        {
-            return Err(Error::DownloadFailed(format!("We tried {} times but \
-                                                      could not download {}. \
-                                                      Last error was: {}",
-                                                     RETRIES, ident, err)));
+        } else if let Err(err) = self.fetch_artifact(ui, (ident, target), token).await {
+            return Err(err);
         }
```
davidMcneil, line 670: I think this `else if let Err` could be removed and we just have `self.fetch_artifact(ui, (ident, target), token).await?`

components/common/src/command/package/install.rs
```diff
@@ -663,14 +665,8 @@ impl<'a> InstallTask<'a> {
                    ident);
         } else if self.is_offline() {
             return Err(Error::OfflineArtifactNotFound(ident.as_ref().clone()));
-        } else if let Err(err) =
-            retry::retry_future!(delay::Fixed::from(RETRY_WAIT).take(RETRIES),
```
davidMcneil, line 667: We lose this retry logic, correct? Are we ok with that here?

components/builder-api-client/src/builder.rs
jeremymv2: @davidMcneil is correct, this is the exact reason I had to go down this macro path.

components/common/src/command/package/install.rs
```diff
@@ -843,22 +839,33 @@ impl<'a> InstallTask<'a> {
...
+                                                      not download {} for {}. \
+                                                      (Some HTTP error \
+                                                      conditions are not \
+                                                      practically worth \
+                                                      retrying) - last error: \
+                                                      {}.",
+                                                     API_RETRIES,
+                                                     ident,
+                                                     target,
+                                                     e)));
```
davidMcneil, line 865: These `DownloadFailed` messages are very similar. Instead of having `DownloadFailed` take a `String` what if it took an `enum` describing what kind of download was attempted (eg `Package`, `OriginKey`, `EncryptionKey ...). Each of those enums could contain the relevant information for the failure (eg `Package` would have the `PackageIdent`). Then these long strings would mostly live in the `fmt::Display` `impl` of `Error`.

components/builder-api-client/src/builder.rs
christophermaier: Ah, cool :+1:

components/builder-api-client/src/builder.rs
jeremymv2: Added in https://github.com/habitat-sh/habitat/pull/7874/commits/2a457e8b6f09ace44d7248a5f5146dcfbd2bb44e

components/builder-api-client/src/builder.rs
jeremymv2: Brilliant! I added in https://github.com/habitat-sh/habitat/pull/7874/commits/2a457e8b6f09ace44d7248a5f5146dcfbd2bb44e

components/common/src/command/package/install.rs
jeremymv2: Added in https://github.com/habitat-sh/habitat/pull/7874/commits/2a457e8b6f09ace44d7248a5f5146dcfbd2bb44e

components/common/src/command/package/install.rs
jeremymv2: This is a superb idea. Added in https://github.com/habitat-sh/habitat/pull/7874/commits/2a457e8b6f09ace44d7248a5f5146dcfbd2bb44e

components/common/src/command/package/install.rs
jeremymv2: It still has the retry logic, I just moved it down to the `fetch_artifact` function that actually calls the Builder API, on line 842 of this file.

components/builder-api-client/src/builder.rs
jeremymv2: These match arms remain are purposeful as such so that they return a `retry::OperationResult::Err(e)` which terminates the retry wrapper.  Every other error type returns a `retry::OperationResult::Retry(e)` which allows the retry wrapper to iterate to the next loop cycle. 

This is because I don't want to needlessly retry these HTTP errors that are effectively unrecoverable (404, 401, 403, 422).

components/common/src/error.rs
```diff
@@ -55,17 +56,76 @@ impl fmt::Display for CommandExecutionError {
...
 }
 
+#[derive(Debug)]
+pub enum APIFailure {
+    DownloadFailed(String),
+    UploadFailed(String),
+    Package(FullyQualifiedPackageIdent, PackageTarget),
+    OriginKeyRev(String, String),
+    LatestOriginKey(String),
+}
```
christophermaier, line 66: I love breaking out these errors in a more granular way like this, but I think there are some improvements we can make to better preserve encapsulation and also make it even easier to manage these errors.

First, as API errors, it feels like this particular enum really belongs in the API client crate. I'd also argue that it should actually implement `Error` directly.

One neat thing about enums is that the individual instances can serve as their own constructors. That is, if `DownloadFailed` were defined as `DownloadFailed(usize, FullyQualifiedPackageIdent, PackageTarget, Error)`, then you wouldn't actually need the `APIFailure::package_download_failed` function; you could just call, say, `DownloadFailed(5, my_ident, my_target, e)` and call it a day. What _package_download_failed_  is doing is actually what `impl std::fmt::Display for APIFailure` should be doing.

What the `Error::BuilderAPITransferError` enum case is actually doing is providing a mapping of the API client error type into the `habitat_common::error::Error` type. This is often done with `From` implementations for errors, which is what the `?` operator uses to coerce errors on your behalf. For instance, you'll often see a bunch of implementations in our various error modules like this:

```rust
impl From<io::Error> for Error {
    fn from(err: io::Error) -> Self { Error::IO(err) }
}
```
That's basically taking a error from some other crate we depend on (here, a standard library I/O error) and mapping it to one of _our_ error instances, which ensures that everything coming out of _our_ crate has a unified error type. That, in turn, makes it easier for consumers of _our_ crate to deal with errors that come from it.

With this implementation in place, we really don't have to worry about explicitly mapping standard library I/O errors into our own errors; any time we have a function that calls a standard library function that could return one of those, simply adding a `?` to that call will automatically invoke this `From` implementation to coerce it into our own error type, free of charge.

This _can_ also end up being a bad thing if it is used indiscriminately; you may not actually want every conceivable I/O error to map to the same error case in your crate. Perhaps problems reading a configuration file should actually be a case of a `ConfigurationError`, for example, which is more meaningful and carries more useful information that "I tried to do an I/O operation somewhere and it failed; sorry! :shrug: " We actually have a lot of this in the codebase right now, to our detriment. However, I think a big reason this is a problem for us is that we tend to have giant catch-all error cases at the top level of our crates that effectively flatten out our error mapping in this bad way. If you had a `ConfigurationError` that was thrown by anything in your `config.rs` module (and submodules!), then you could map all your I/O errors there to `ConfigurationErrors`, and then at the top level of your crate, you could map `ConfigurationError` to whatever your top-level `Error` was. It's a matter of capturing whatever content is important at the site of the error, and then providing the means to bubble that up as appropriate.

I did some of this in the container exporter a little while ago, where I have an `Engine` abstraction that has its own error type, and then Docker and Buildah implementations of that `Engine` that have their _own_ error types, each of which get progressively bubbled up to higher-and-higher levels. (e.g., https://github.com/habitat-sh/habitat/blob/9224683d30f5cc7c276ac179ae0f3d9b8eee02d6/components/pkg-export-container/src/engine/buildah.rs#L33-L41). This helps to keep knowledge localized, but also provides clear paths for information to flow from one place to another without any degradation.

(Just to be clear, I'm absolutely _not_ advising that every single module in a crate necessarily must have their own error type, and that their parent module then has to coerce all errors from them and their sibling modules into a higher-level error type, and on and on, until you have a crate that has 52 different error types. Rather, I'm arguing that distinct logical subdomains within a crate could very well benefit from having their own error type that adequately captures all the necessary nuance of the error conditions that they can produce. This is, in fact, _exactly_ why `std::io::Error` exists in the first place.) 

Viewed from this perspective, let's take a look at this code, picked arbitrarily from `hab/src/command/origin/key/upload.rs`
```rust
 return Err(CommonError::BuilderAPITransferError(APIFailure::key_upload_failed(API_RETRIES, &name, &rev, e.into())).into());
```
Here, we're in the `hab` CLI crate, but we are manually creating an error that is logically from the API crate (using `APIFailure::key_upload_failed`), manually wrapping that in an error from _another_ crate that isn't even in this code flow (`habitat_common`, via `CommonError::BuilderAPITransferError`), and then finally converting that into an instance of the catch-all error case from _our_ crate (using `into()`, which invokes the code from a `From` implementation: https://github.com/habitat-sh/habitat/blob/9224683d30f5cc7c276ac179ae0f3d9b8eee02d6/components/common/src/error.rs#L252-L254).

We're concerning ourselves with too many low-level concerns in what is fairly high-level code here. The root of the error here, the `APIFailure` should be something that is produced completely from within the API crate itself, because that is where all knowledge about that error is located. If we had some additional context up here in the CLI crate that is not present in the API crate, but is nevertheless important to capture, then that would signal that our CLI crate should have some kind of error instance that wraps an API error, but also adds whatever important information it needs to. We shouldn't be creating a `habitat_common` error in the CLI crate, though, because only the `habitat_common` crate should be creating its errors.

If things were set up like I'm arguing for, then we could remove that entire `return` clause and simply replace it with `?`, and everything would just work. Additionally, it would declutter the code so you could focus on what it's actually doing, rather than wading through lots of error handling. We'd basically be taking a lot of repeated logic that lives in `match` and `if` statements, and encoding them into the structure of our type system.

davidMcneil: Completely agree! This seems like the "correct" way to do error handling. Two crates that may be useful are [`thiserror`](https://github.com/dtolnay/thiserror) and [`anyhow`](https://github.com/dtolnay/anyhow). See [here](https://github.com/dtolnay/thiserror#comparison-to-anyhow) for a good explanation of when to use each one. I think `thiserror` will be especially useful in cutting down the boilerplate of having many error types. The great thing about `thiserror` is it solely deals with `std::error::Error` (vs `failure` which introduces new traits).

christophermaier: @davidMcneil Ooh, those both look incredibly useful. I'm all up for some experiments with them :+1:

components/builder-api-client/src/builder.rs
davidMcneil: I like @christophermaier suggestion [here](https://github.com/habitat-sh/habitat/pull/7874#discussion_r470732256) of allowing the `retry_builder_api` macro to be called with or without a retry iterator. I think it would look something like:

```rust
#[macro_export]
macro_rules! retry_builder_api {
    ($api_future:expr) => {
        retry_builder_api!($api_future,
                           retry::delay::Fibonacci::from(API_RETRY_WAIT).take(API_RETRIES))
    };
    ($api_future:expr,iter_opt: expr,) => {
        async {
            ...
        }
    };
}
```

Or we could simply hardcode the retry iterator. It looks like we only ever call the macro with `None`.

components/builder-api-client/src/builder.rs
stevendanna: I wonder if we should skip retry for any error in the 4XX Client Errros range? (Or alternatively only retry HTTP errors in the 5XX Server Errors range?)

jeremymv2: @christophermaier I've been battling this error when I try your suggestion above:
```
error[E0382]: use of moved value
--> components/common/src/command/package/install.rs:843:9
|
843 | /         retry_builder_api!(async {
844 | |             ui.status(Status::Downloading, format!("{} for {}", ident, target))?;
845 | |             self.api_client
846 | |                 .fetch_package((ident.as_ref(), target),
...   |
850 | |                 .await
851 | |         }).await
| |          ^
| |          |
| |__________value moved here
|            value used here after move
|
= note: move occurs because value has type `habitat_api_client::error::Error`, which does not implement the `Copy` trait
= note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```  
:thinking: The Error enum doesn't implement Copy or Clone and doesn't seem to be easily changed due to all the `String` types within.

components/builder-api-client/src/builder.rs
```diff
@@ -57,6 +58,9 @@ const X_FILENAME: &str = "x-filename";
 
 const DEFAULT_API_PATH: &str = "/v1";
 
+pub const API_RETRIES: usize = 5;
+pub const API_RETRY_WAIT: Duration = Duration::from_secs(10);
+
```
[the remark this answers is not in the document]
jeremymv2: Handled in https://github.com/habitat-sh/habitat/pull/7874/commits/26704a92a178597076d034d64aa498acfd33132d

components/builder-api-client/src/builder.rs
jeremymv2: I love this idea @christophermaier @davidMcneil!
I've refactored the macro in https://github.com/habitat-sh/habitat/pull/7874/commits/26704a92a178597076d034d64aa498acfd33132d allowing for it to be called with or without the iterator and including a `with_custom_iterator: ` for readability.

components/common/src/error.rs
jeremymv2: @christophermaier Thanks for the detailed review in this section. After reading carefully, I believe I've implemented the patterns you're suggesting in https://github.com/habitat-sh/habitat/pull/7874/commits/26704a92a178597076d034d64aa498acfd33132d

components/builder-api-client/src/builder.rs
davidMcneil: I am not sure why the error is happening. The error message loses some of its usefulness passing through the macro. It looks like just reconstructing an `ApiError` works though. Something like:

```rust

jeremymv2: Ah yes! the `if` guard on the match arm was just what I was looking for to avoid the move!
Thanks @davidMcneil - added in https://github.com/habitat-sh/habitat/pull/7874/commits/879cbd572b6ac202a85b3d4b2fa201fd6326abe5

davidMcneil (changes requested):
Wrapping `APIClientError`s in `Error::APIClient()` should not be needed. The ? will automatically convert the `APIClientError` to the appropriate type. [Here](https://github.com/habitat-sh/habitat/blob/64cbf5e7116baa5cda057b4dc6c9ae04dde86e00/components/hab/src/command/pkg/download.rs#L370) is an example.

This looks great. I like the new error handling!

davidMcneil (approved):

PR #7878, opened 2020-08-17
Add `hab sup restart` command

### LINKED ISSUE #7651: Add a command to restart the supervisor without service disruption
Add a command (`hab sup restart`?) that allows restarting the supervisor without disrupting the running services. This would be similar to sending the `SIGHUP` signal on linux, but would allow it to work on windows as well.

The motivation is to provide a mechanism to do a manual Supervisor upgrade. The latest Supervisor version on disk will be started on restart.

davidMcneil:
Resolves #7651


components/hab/src/cli/hab/sup.rs
```diff
@@ -64,6 +64,12 @@ pub enum HabSup {
         #[structopt(flatten)]
         remote_sup: RemoteSup,
     },
+    /// Restart a Supervisor without restarting its services
+    #[structopt(no_version)]
+    Restart {
```
davidMcneil, line 69: It pains me that we are adding another command to the `hab sup` family of commands (see comment [here](https://github.com/habitat-sh/habitat/blob/ea661c301f55bc930c3edd0b8ba1122536f5dc0a/components/hab/src/main.rs#L156)). I considered adding a `hab sup-manage` subcommand which would be for subcommands handled by the `hab` binary. However, I was not a fan of the name, and separating it from the other commands felt awkward.

Any thoughts on what good UI/UX would be here without adding another `hab` handled command to `hab sup`?

components/sup/src/manager.rs
```diff
@@ -380,9 +380,10 @@ impl ReconciliationFlag {
 /// state gets shared with all the CtlGateway handlers.
 pub struct ManagerState {
     /// The configuration used to instantiate this Manager instance
-    cfg:           ManagerConfig,
-    services:      Arc<sync::ManagerServices>,
-    gateway_state: Arc<sync::GatewayState>,
+    cfg:            ManagerConfig,
+    services:       Arc<sync::ManagerServices>,
+    gateway_state:  Arc<sync::GatewayState>,
+    should_restart: AtomicBool,
```
davidMcneil, line 386: This does not feel great but seemed to be the simplest solution given the current architecture.

test/end-to-end/test_supervisor_restarts.ps1
```diff
@@ -0,0 +1,63 @@
...
+Describe "Supervisor restarts on 'hab sup restart'" {
+    $launcher_proc = Start-Supervisor -Timeout 45 -LogFile "sup.log"
+    $supervisor_proc = Get-Process hab-sup
+    Invoke-NativeCommand hab sup restart
+    while (!$supervisor_proc.HasExited) {
+        Start-Sleep 1
+    }
+    while (!(Get-Process hab-sup -ErrorAction SilentlyContinue)) {
+        Start-Sleep 1
+    }
```
christophermaier, line 46: Can these while loops be a little smarter with a maximum timeout? If something is broken such that `hab sup restart` no longer works, it seems that we would hang indefinitely here (at least, until the Buildkite job itself timed out).

(Same applies to the above `Describe` block, too)

davidMcneil: Good call! Addressed in https://github.com/habitat-sh/habitat/pull/7878/commits/73474b60635d3e6779961a3af5a5563706a92418

components/sup/src/manager.rs
christophermaier: Yeah, agreed :frowning_face:

components/hab/src/cli/hab/sup.rs
christophermaier: I agree that it's awkward internally, but as far as the desired UX is concerned, this _does_ work. That's not to say there's not another UX we could settle on that make the internal code cleaner. That being said, I'm not sure what that would be right now :frowning_face:

PR #7879, opened 2020-08-17
Report errors when getting user and group

### LINKED ISSUE #7298: hab-sup erroneously reports missing group, fails to load service

davidMcneil:
Resolves #7298 

This improves our APIs used to get users and groups by returning `Result<Option<T>>` instead of simply an `Option<T>`. This means we should now report an error instead of simply assuming there is no group in the case that the buffer is not large (or any other error). Users should then be able to increase the buffer size appropriately with the [`_SC_GETGR_R_SIZE_MAX` setting](https://github.com/nix-rust/nix/blob/96054b6933404b4c1d5dbd49db7ec7c7a26e44de/src/unistd.rs#L2707).

### DIFF
--- a/components/common/src/templating/config.rs
+++ b/components/common/src/templating/config.rs
@@ -613,10 +613,12 @@ mod test {
 
     fn curr_username() -> String {
         users::get_current_username().expect("Can get current username")
+                                     .expect("Can get current username")
     }
 
     fn curr_groupname() -> String {
         users::get_current_groupname().expect("Can get current groupname")
+                                      .expect("Can get current groupname")
     }
 
     fn toml_from_str(content: &str) -> toml::value::Table {
--- a/components/common/src/templating/config.rs
+++ b/components/common/src/templating/config.rs
@@ -612,13 +612,13 @@ mod test {
     use tempfile::TempDir;
 
     fn curr_username() -> String {
-        users::get_current_username().expect("Can get current username")
-                                     .expect("Can get current username")
+        users::get_current_username().expect("Failed to get username")
+                                     .expect("No username found")
     }
 
     fn curr_groupname() -> String {
-        users::get_current_groupname().expect("Can get current groupname")
-                                      .expect("Can get current groupname")
+        users::get_current_groupname().expect("Failed to get groupname")
+                                      .expect("No groupname found")
     }
 
     fn toml_from_str(content: &str) -> toml::value::Table {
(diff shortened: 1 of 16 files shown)


components/common/src/templating/config.rs
```diff
@@ -613,10 +613,12 @@ mod test {
 
     fn curr_username() -> String {
         users::get_current_username().expect("Can get current username")
+                                     .expect("Can get current username")
```
christophermaier, line 616: I know this is a test helper function, but it would be good to make these `expect` messages different to help distinguish which operation failed, if one of them ever does.

I see a handful of other instances of this pattern elsewhere; same suggestion applies.

davidMcneil: Agreed. Addressed in https://github.com/habitat-sh/habitat/pull/7879/commits/1bd6407da4aecc1299a5ac0eef5ddcb0a9dbc46c

stevendanna (commented):
:D You are much braver than I was.  Thanks for changing this API!

PR #7880, opened 2020-08-19
additional validation for PackageIdent fields

jeremymv2:
Closes https://github.com/habitat-sh/builder/issues/1407

This will prevent bad `PackageIdent` version or release fields from
creating an inoperable situation on either the filesystem or via HTTP
calls to Builder.

In both cases, sequences such as '/../' or '/./' or even '/..1/' in the package name
create problems for operations on the CLI or Builder API as those have
special meaning and result in an inoperable or unintended outcome.

For all intents and purposes, `.` should not exist:

- at the start of a  field
- end of field
- or more than once, in a repeated sequence anywhere in a field

We prevent the above patterns in new packages by using `PackageIdent` field validation in the package signing (executed during builds). Validation will also be picked up by Builder [here](https://github.com/habitat-sh/builder/blob/2d1db94bba7ae3b4a1854ca77220134460d9e6c7/components/builder-api/src/server/resources/pkgs.rs#L504-L507) and will prevent such packages from being uploaded.

### DIFF
--- a/components/core/src/package/ident.rs
+++ b/components/core/src/package/ident.rs
@@ -13,7 +13,11 @@ use std::{cmp::{Ordering,
 
 lazy_static::lazy_static! {
     static ref ORIGIN_NAME_RE: Regex =
-        Regex::new(r"\A[a-z0-9][a-z0-9_-]*\z").expect("Unable to compile regex");
+        Regex::new(r"\A[a-z0-9][a-z0-9_-]*\z").expect("Unable to compile ORIGIN_NAME_RE");
+    static ref IDENT_NAME_RE: Regex =
+        Regex::new(r"^[A-Za-z0-9_-]+$").expect("Unable to compile IDENT_NAME_RE");
+    static ref INOPERABLE_SEQUENCE_RE: Regex =
+        Regex::new(r"^\.{2,}|^\.$").expect("Unable to compile INOPERABLE_SEQUENCE_RE");
 }
 
 #[derive(Deserialize, Serialize, Eq, PartialEq, Debug, Clone, Hash)]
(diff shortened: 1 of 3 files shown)


components/hab/src/command/pkg/sign.rs
```diff
@@ -1,21 +1,35 @@
...
+    // Detect if what we're about to sign is in fact a PackageArchive and
+    // if so, we need to do some basic PackageIdent field validation.
+    if let Ok(mut archive) = PackageArchive::new(PathBuf::from(dst)) {
+        let ident = archive.ident()?;
+        ui.status(Status::Verifying,
+                  format!("{} has valid PackageIdent fields", ident))?;
+        if !ident.valid() {
+            ident.validate_fields()?;
+        }
+    }
```
christophermaier, line 24: Rather than have this kind of manual verification, it would be good to see what the blast radius of making `PackageIdent::new` return a `Result` instead, and just do all that verification when you create the type. This could also be incorporated into the `FromStr` implementation, as well... basically, any time you have a `PackageIdent`, you _know_ it's valid, because the compiler made it that way.

components/core/src/package/ident.rs
```diff
@@ -33,8 +37,31 @@ pub trait Identifiable: fmt::Display {
     fn fully_qualified(&self) -> bool { self.version().is_some() && self.release().is_some() }
 
     fn valid(&self) -> bool {
-        let re = Regex::new(r"^[A-Za-z0-9_-]+$").unwrap();
-        re.is_match(self.name())
+        if self.fully_qualified() {
+            IDENT_NAME_RE.is_match(self.name())
+            && !INOPERABLE_SEQUENCE_RE.is_match(self.version().unwrap())
+            && !INOPERABLE_SEQUENCE_RE.is_match(self.release().unwrap())
```
christophermaier, line 43: Doesn't release have to be a timestamp string? Or is that a thing that we've always implicitly assumed, but never enforced, and now we have packages that have weird releases?

PR #7885, opened 2020-08-24
readonly pkg_release and ensure pkg_version is semver

jeremymv2:
Fixes https://github.com/habitat-sh/builder/issues/1407

This should help prevent packages with invalid `pkg_version` or `pkg_release` from being created and uploaded to Builder. It enforces adherance to a semver format that we already require internally in Builder's table views.

### DIFF
--- a/components/plan-build/bin/hab-plan-build.sh
+++ b/components/plan-build/bin/hab-plan-build.sh
@@ -367,8 +367,8 @@ INITIAL_PWD="$(pwd)"
 pkg_target='@@pkg_target@@'
 # The package's origin (i.e. acme)
 pkg_origin=""
-# Each release is a timestamp - `YYYYMMDDhhmmss`
-pkg_release=$(date -u +%Y%m%d%H%M%S)
+# Each release is a timestamp - `YYYYMMDDhhmmss`. Once set, we do not support mutating the pkg_release.
+readonly pkg_release=$(date -u +%Y%m%d%H%M%S)
 # The default build deps setting - an empty array
 pkg_build_deps=()
 # The default runtime deps setting - an empty array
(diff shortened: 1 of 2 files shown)


components/plan-build/bin/hab-plan-build.sh
```diff
@@ -569,6 +569,21 @@ _ensure_origin_key_present() {
...
 }
 
+# **Internal** Ensures that the plan pkg_version adheres to the semver specification.
+# Additionally, it checks to ensure that the pkg_version does not contain any dots at
+# the beginning of the string, at the end, nor multiple sequential dots anywhere within.
+# The checks for dots as just described, prevent inoperable or unexpected results when operating
+# the hab CLI or interacting with Builder HTTP APIs as the dots have special meaning in a URI.
+_validate_pkg_version() {
+  local inoperable_re="^\.+|\.$|\.{2,}"
+  local wellformed_re="^([[[:digit:]]\.]*[[:digit:]]+)(.+)?"
```
christophermaier, line 579: Wouldn't something that matched a "wellformed" regular expression necessarily _not_ match an "inoperable" one, and vice versa? I'm not sure we need two regexes for this. If we do, that seems to indicate that the "wellformed" regular expression isn't strong enough.

Also, I don't think we actually accept all semver version strings, since we don't (as far as I recall) allow for pre-release / build version / etc. "tags" (e.g. "1.0.0-alpha1"). It would be good to be very explicit about that in any documentation. Tests that capture this distinction would also be great.

components/plan-build/bin/hab-plan-build.sh
```diff
@@ -569,6 +569,21 @@ _ensure_origin_key_present() {
...
+
+  if [[ ! $pkg_version =~ $wellformed_re ]] || [[ $pkg_version =~ $inoperable_re ]]; then
+     exit_with "Found pkg_version with an invalid format, aborting! See: https://semver.org/"
```
christophermaier, line 582: If we don't accept _all_ semver strings, then we should be _very_ clear about that in this user-facing error message. Otherwise, that will only lead to confusion and frustration.

components/plan-build/bin/hab-plan-build.sh
jeremymv2: I chose to separate the two regexes out in order to reduce their complexity and preserver readability 😃 
I certainly understand the point you make though that a wellformed version regex should encompass guarantees that it's not also malformed. Be forewarned, the resulting regex could cause seizures though!

We do in fact accept pre-release and build versions such that `1.2.3.4-git-SHA1` is acceptable and will not break Builder. I'll add some clarification to the error message and improve supporting documentation around this.

jeremymv2: Screen Shot 2020-08-25 at 10 45 30 AM

jeremymv2:
We spoke internally as a team and have decided on a more fundamental approach to tackle this. This PR did lay the foundation for what's to come though..

PR #7581, opened 2020-03-23
Update Builder Documentation

kagarmoe:
To build the docs locally:

`cd www` 
`bundle exec middleman serve`

Signed-off-by: kagarmoe <kgarmoe@chef.io>

PR #7892, opened 2020-08-26
Builder on-prem documentation

kagarmoe:
This PR adds the documentation from the [Builder on-prem](https://github.com/habitat-sh/on-prem-builder) into the `www/source/docs` and the `components/habitat-chef-io`  directories. I also added some styling for notes and warnings.

ALSO: My local hab repo was in a strange state & added commits from another branch. I need help remove the commits from this PR. I think they'll vanish once #7581 merges.

Nav:

Contents:


kagarmoe:
We will add the on-pre components through a go.mod instead.

PR #7903, opened 2020-09-01
allow `make lint` to use system libraries 

jeremymv2:
This allows a devloper to set `LD_LIBRARY_PATH` in order to leverage system libraries while clippy runs a build for lint checks.

Prior to this change, on `arch` linux with `glibc 2.32-2`, `make lint` fails with:

```
error: failed to run custom build command for `zmq v0.8.3 (https://github.com/habitat-sh/rust-zmq?branch=v0.8-symlinks-removed#20a73b92)`

Caused by:
process didn't exit successfully: `/home/jmiller/Projects/habitat/target/debug/build/zmq-0b23a8ebf05ec083/build-script-build` (exit code: 127)
--- stderr
/home/jmiller/Projects/habitat/target/debug/build/zmq-0b23a8ebf05ec083/build-script-build: symbol lookup error: /hab/pkgs/core/glibc/2.29/20200305172459/lib/librt.so.1: undefined symbol: __clock_nanosleep, version GLIBC_PRIVATE
```

`cargo build` compiles it just fine though and the latest `zmq` crate does not have this issue.

If I set `$LD_LIBRARY_PATH` to `/usr/lib`, where all my system zero-mq libraries exist, `make lint` is back on the happy path.
Since this is a change that developers can opt into, shouldn't introduce any CI behavior changes, and it's only for `clippy` builds, I thought it would be ok to incorporate this workaround into the `run_clippy.sh` script.

### DIFF
--- a/.expeditor/scripts/verify/run_clippy.sh
+++ b/.expeditor/scripts/verify/run_clippy.sh
@@ -16,7 +16,7 @@ export LIBZMQ_PREFIX
 LIBZMQ_PREFIX=$(hab pkg path core/zeromq)
 # now include zeromq so it exists in the runtime library path when cargo test is run
 export LD_LIBRARY_PATH
-LD_LIBRARY_PATH="$(hab pkg path core/zeromq)/lib"
+LD_LIBRARY_PATH="$(hab pkg path core/zeromq)/lib:${LD_LIBRARY_PATH:-}"
 
 # Install clippy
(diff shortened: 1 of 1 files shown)


jeremymv2:
created a local workaround:
```
alias lint='cp -f run_clippy.works.sh .expeditor/scripts/verify/run_clippy.sh && make lint'
```

PR #7905, opened 2020-09-01
Remove Operators from docs

kagarmoe:


www/source/docs/best-practices.html.md.erb
```diff
@@ -281,216 +281,7 @@ If you would like additional details, the following technical journal articles d
 * [A Robust and Scalable Peer-to-Peer Gossiping Protocol](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.90.8773&rep=rep1&type=pdf) by Spyros Voulgaris, Mark Jelasity, and Maarten van Steen.
 
 ---
-## <a name="container-orchestration" id="container-orchestration" data-magellan-target="container-orchestration">Container Orchestration with Chef Habitat</a>
```
eeyun, line 284: This seems to remove more than just info on the k8s operator from the docs and instead removes all  kubernetes content completely. Is this intentional? Most of these behaviors still exist even if its out of date and an inch from deprecated. Would it be better to keep these docs and add a warning to users that these  behaviors are deprecated and mileage may vary?

eeyun: As an additional note. If we are infact removing all mention of kubernetes from our docs we should probably also remove all references to AKS and OpenShift.

christophermaier: Hrmm.... I may have spoken a bit too quickly in Slack. It's probably still useful to preserve documentation we have around _raw_ Kubernetes usage. People are still going to try and use Habitat in that environment, so any basic information we have on that would still be useful. Anything that mentions the Operator, or the old Kubernetes or Helm exporters, on the other hand, should definitely be removed. I merged a commit a while back (https://github.com/habitat-sh/habitat/pull/7463/commits/a96bb5caa182e7bb5f48107bf8ae7f4991d2e74f) that removed things, but I'm not sure how stuff got back into the documentation.

kagarmoe: Its easy enough to back out of changes, thanks to GitHub! @christophermaier, I think you're looking for the phrase "gaslighting bucket". I suspect that our sync process red bucket content back, but I have no idea how it would work. The point of fact, however, is that your work is missing, as is some of mine. Also, the bucket broke the left-nav, which is no longer stays on the screen.

kagarmoe:
handled in the hugo docs.

PR #7918, opened 2020-09-16
[REFACTOR] Simplify Blake2b hash implementation

christophermaier:
In addition to a number of other smaller cleanups, the main thing this PR brings in is the replacement of a complex, manual, `unsafe` implementation of Blake2b hashing with a pure Rust, safe, implementation from the `blake2b_simd` crate. As a result, we can remove the `libsodium_sys` crate as a dependency.


components/core/src/crypto/hash.rs
```diff
@@ -27,34 +29,57 @@ const HASH_DIGEST_SIZE: usize = libsodium_sys::crypto_generichash_BYTES as usize
...
-                                 Error::CryptoError(format!("Did not get {} bytes for digest",
-                                                            HASH_DIGEST_SIZE))
-                             })?;
-        Ok(Blake2bHash::new(*boxed))
+// We *could* just wrap the `blake2b_simd::Hash` directly in our
+// `Blake2bHash` type, but then we wouldn't be able to parse a
+// `Blake2bHash` from a string, because `blake2b_simd::Hash does not
+// have that ability. So, we'll just grab the bytes and call it a
+// day.
+impl From<blake2b_simd::Hash> for Blake2bHash {
```
davidMcneil, line 77: You would still be able to implement `FromStr` for `Blake2bHash` even if it was wrapping `balke2b_simd::Hash` correct? Or am I misunderstanding this comment.

davidMcneil:
I think we could `#derive` this if we wanted to https://github.com/habitat-sh/habitat/blob/5981b8404834d691613645fb634d18f6ddc8082d/components/core/src/crypto/hash.rs#L99

components/core/src/crypto/hash.rs
christophermaier: That would turn into producing a `blake2b_simd::Hash` instance from the string, but I don't see a clear way to do that.

davidMcneil (approved):

PR #7932, opened 2020-09-28
Update rust to 1.46.0

davidMcneil:


christophermaier (approved):

christophermaier:
Going to close this for now... we're seeing some compilation failures with the following error:

```
error: reached the type-length limit while instantiating `std::thread::LocalKey::<std::cel...:Context]], std::task::Poll<()>>`

= note: consider adding a `#![type_length_limit="1899201"]` attribute to your crate
```

Interestingly this only seems to happen when compiling during the package build of `core/hab`. Others in the community have seen this error lately as well; for now, our approach will be "wait to see if this is fixed in the next version of Rust".

PR #7939, opened 2020-10-13
Make shorter Builder pages

kagarmoe:
Our pages are still unwieldy. This PR shortens the Builder content by creating more short pages and moving some of them into a newly-created top-level heading for `Origins`.

This is an example:

RBAC


components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+All Chef Habitat Builder users with access to the origin can view the origin public key revisions in the origin key tab (Builder > Origin > Keys) and download the origin public key, but only users with the origin 'administrator' or 'owner' roles can view or download the origin private key, or change the origin key pair.
+
+| Keys Actions | Read-Only | Member | Maintainer | Administrator | Owner |
+|---------|-------|-------|-------|-------|-------|
+| View keys | Y | Y | Y | Y | Y |
+| Add/Update/Delete keys | N | N | N | Y | Y |
+
+Chef Habitat uses origin keys:
+
+- When you build an artifact in your local environment, Chef Habitat signs the artifact with a public key
```
christophermaier, line 35: It actually signs with a secret key, not a public one.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+- When you upload an artifact to Chef Habitat Builder or Builder on-prem, Chef Habitat verifies that the artifact was signed with its public key
```
christophermaier, line 36: This may be ambiguous. Is Habitat verifying "that the artifact was signed with its public key", or verifying, using the public key, that the artifact was signed?

(It should be the latter)

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+- When you install an artifact on a Chef Habitat Supervisor, Chef Habitat uses the public origin key to authorize the artifact's installation; Chef Habitat only installs artifacts for which it has the public origin key
```
christophermaier, line 37: This happens _any_ time a package is installed, whether it's the Supervisor doing it or not.

christophermaier: It would be more precise to say that it only installs artifacts that it can verify were properly signed by the package's origin.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+- When you download an artifact to your local environment, Chef Habitat uses the public origin key to verify the artifact's integrity before it starts the installation
```
christophermaier, line 38: This is essentially the same point as the sentence above, but worded a bit differently; of the two, I personally prefer this one.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+
+```hab
+testorigin-20190416223046.pub
+testorigin-20190416223046.sig.key
+```
+
+- "testorigin" is the origin's name
+- "20190416223046" is the date and time of the key's creation, which was 2019-04-16 22:30:46.
+- `.pub` is the file extension for the public key
+- `.sig.key` is the file extension for the private key, which is also called a "signing key"
```
christophermaier, line 57: While I don't think there's really a semantic difference, the code refers to all the "private" keys as "secret". If we could be consistent across code and documentation, I think that would be nice.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+
+## The Keys Tab
+
+When you create an origin, Chef Habitat Builder automatically generates an origin key pair and saves both keys. To view your origin keys on Chef Habitat Builder, navigate to your origin and select the **Keys** tab. (Builder > Origins > Keys) You will always be able to view and download origin public keys, but you will only see the private keys for origins in which you are an "administrator" or "owner".
```
christophermaier, line 61: It has not always been very clear, but there are actually _two_ distinct "origin key pairs" floating around. The first one, and the one described in this section, is the signing key pair. The second one came along much later, and is an encryption key pair, used for safely uploading build secrets to Builder. Our CLI and documentation hasn't ever really been direct about which is which, unfortunately. As such, it would be nice to start being more specific in this documentation.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+### Generate Origin Keys
+
+When you create an origin through the site, Chef Habitat Builder automatically generates an origin key pair.
+
+The Chef Habitat CLI creates origin key pairs through two different commands, for two different uses:
+
+- Use [`hab setup`]({{< relref "install-habitat.md" >}}) to generate your first origin key pair as part of setting up the `hab` CLI
+- Use the `hab origin key generate <ORIGIN>` command to create an key pair for an origin created with the `hab origin create` command
```
christophermaier, line 122: You can actually use `hab origin key generate` to create a new signing key pair for _any_ origin, not just one created via `hab origin create`.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+Upload the origin private key:
+
+```hab
+hab origin key upload --secret <ORIGIN>
+```
+
+Upload both origin keys at the same time:
+
+```hab
+hab origin key upload  --secfile <PATH_TO_PRIVATE_KEY> --pubfile <PATH_TO_PUBLIC_KEY>
```
christophermaier, line 161: This is another private / secret distinction.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+### Import Origin Keys
+
+Use `hab origin key import` to read the key from a standard input stream into Chef Habitat Builder:
+
+```hab
+hab origin key import <enter or paste key>
+hab origin key import <PATH_TO_KEY>
+curl <URL_THAT_RETURNS_KEY> | hab origin key import
```
christophermaier, line 171: Just to keep things simple (and more true-to-life), I'd probably replace this `curl` with something like 

```bash
cat <PATH_TO_KEY> | hab origin key import
```

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -0,0 +1,190 @@
...
+#### Troubleshoot Origin Key Import
+
+On a macOS, you may encounter an upload failure.
+To remediate this failure:
+
+- Check that your `HAB_AUTH_TOKEN` environment variable is properly set and initialized
+- Add your `SSL_CERT_FILE` to the environment variables in your interactive shell configuration file, such as your `.bashrc`.
+
+```bash
+  export SSL_CERT_FILE=/usr/local/etc/openssl/cert.pem
```
christophermaier, line 183: :thinking: I don't know that we still run into this issue... I think we fixed it.

components/docs-chef-io/content/habitat/origin-rbac.md
```diff
@@ -0,0 +1,176 @@
...
+- [Create a Chef Habitat Builder account]({{< relref "builder-account" >}})
+- [Generate a personal access token]({{< relref "builder-profile#create-a-personal-access-token" >}})
+- [Create an origin]({{< relref "origins.md#create-origin" >}}) or accept an [invitation]({{< relref "#manage-origin-membership-with-hab-origin-invitations" >}}) to an existing origin
+- [Get origin keys]({{< relref "origins.md#origin-keys" >}})
+
+## Role-Based Access Control (RBAC) for Chef Habitat Builder (SaaS and on-prem)
+
+**New in: 1.6.140**
+{{< note >}}
+RBAC improves operational safety by letting you enable specific levels of access to each user of an origin. With this upgrade, existing origins and their users are automatically converted to RBAC. The RBAC upgrade assigns the 'Maintainer' role to existing users with a generic 'member' role. The differences between 'member' and 'Maintainer' are detailed below. The origin owner role remains unchanged.
```
christophermaier, line 29: Should "member" be capitalized?

components/docs-chef-io/content/habitat/origin-rbac.md
```diff
@@ -0,0 +1,176 @@
...
+
+Prerequisites:
+
+- [Download the Chef Habitat CLI]({{< relref "install-habitat.md" >}})
+- [Create a Chef Habitat Builder account]({{< relref "builder-account" >}})
+- [Generate a personal access token]({{< relref "builder-profile#create-a-personal-access-token" >}})
+- [Create an origin]({{< relref "origins.md#create-origin" >}}) or accept an [invitation]({{< relref "#manage-origin-membership-with-hab-origin-invitations" >}}) to an existing origin
+- [Get origin keys]({{< relref "origins.md#origin-keys" >}})
+
+## Role-Based Access Control (RBAC) for Chef Habitat Builder (SaaS and on-prem)
```
christophermaier, line 25: Should "on-prem" be capitalized? Should it be formalized to "On Premises" since "on-prem" is (to me, at least) a little slangy / jargony (or is that OK in this case)?

components/docs-chef-io/content/habitat/origin-rbac.md
```diff
@@ -0,0 +1,176 @@
...
+RBAC Origin Member Roles:
+
+Read-Only
+: The default membership role for any user joining an origin. 'Read-Only' users can read an origin's packages, channels, origin membership, jobs, keys, integrations, invitations, roles, settings but cannot add to, change, or delete anything in the origin, including uploading packages and inviting users to the origin.
+
+Member
+: In addition to read-only access, an origin 'Member' can upload and build packages in the 'unstable' channel, but they cannot promote packages to other channels.
+
+Maintainer
+: Existing origin 'members' are now 'Maintainers'. This role has full read and write access to packages, channels, origin membership, jobs, integrations, invitations, settings. However, the 'Maintainer' role is more limited than the past role, in that 'Maintainers' only have read access to packages, channels, origin membership, jobs, keys, integrations, and settings. Origin 'Maintainers' can read origin membership roles and see and send invitations, but they cannot otherwise change origin membership--their own or anybody else's. Finally, 'Maintainers' can neither read nor write origin secrets.
```
christophermaier, line 43: Might be worth calling out that "Existing origin 'members' are now 'Maintainers'." pertains to when the RBAC feature was initially rolled out. As such, it's historical information, and not current.

components/docs-chef-io/content/habitat/origin-settings.md
```diff
@@ -0,0 +1,105 @@
...
+
+Everyone with origin membership can see the _Settings_ tab, but only origin administrators and owners can add, update, or delete settings content.
+
+| Settings Actions | Read-Only | Member | Maintainer | Administrator | Owner |
+|---------|-------|-------|-------|-------|-------|
+| View settings | Y | Y | Y | Y | Y |
+| Add/Update/Delete settings | N | N | N | Y | Y |
+| **Origin Secrets Actions** |
+| View secrets | N | N | Y | Y | Y |
+| Add/Update/Delete secrets | N | N | N | Y | Y |
```
christophermaier, line 30: Is there any magic that allows these ASCII tables to be manually aligned in the source, but still generate the proper tables in the rendered output?

When reviewing, it would make it much easier to visually verify the accuracy of the table without having to do all the row/column bookkeeping mentally.

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: I know that I've made this change before. Frequently. However, this time it will stick in Hugo. 😄

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: First, let me tell you that your origin_keys.rs code is really easy to follow.

It should be the latter.

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: I'm not sure what the original intent of the two was. Can we make a meaningful distinction between uploading packages to the supervisor vs. downloading them into a studio environment?

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: Yes, but the common term is "private key" not "secret key", so I would want to change the code. :fightme:

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: I have a plan for this!!! But not in this PR, because we need a separate PR to adequately represent your key work.

components/docs-chef-io/content/habitat/origin-rbac.md
kagarmoe: It was a product decision of yore.

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: I've made a note to follow up on this.

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: ok

components/docs-chef-io/content/habitat/origin-settings.md
kagarmoe: Sadly, it would need to be done manually for this table, because it is kind of complex.

kagarmoe: HOWEVER, you can preview it on the netlify build

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: I tried to  make the distinction in this change.

components/docs-chef-io/content/habitat/origin-keys.md
christophermaier: Counterpoint: the crypto library we're using calls them secret keys 😂 

https://github.com/habitat-sh/habitat/blob/e3d33d0bcc0d4d0585b25061941a02f489531478/components/core/src/crypto/keys/signing.rs#L15-L16

tenor-114036778

components/docs-chef-io/content/habitat/origin-rbac.md
christophermaier: tenor-148759549

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -32,9 +32,9 @@ All Chef Habitat Builder users with access to the origin can view the origin pub
 
 Chef Habitat uses origin keys:
 
-- When you build an artifact in your local environment, Chef Habitat signs the artifact with a public key
-- When you upload an artifact to Chef Habitat Builder or Builder on-prem, Chef Habitat verifies that the artifact was signed with its public key
-- When you install an artifact on a Chef Habitat Supervisor, Chef Habitat uses the public origin key to authorize the artifact's installation; Chef Habitat only installs artifacts for which it has the public origin key
+- When you build an artifact in your local environment, Chef Habitat signs the artifact with the private origin key key
```
christophermaier, line 35: "key key"

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -32,9 +32,9 @@ All Chef Habitat Builder users with access to the origin can view the origin pub
...
+- When you upload an artifact to Chef Habitat Builder or Builder on-prem, Chef Habitat uses the public origin key to verify that the artifact was signed with the private origin key
+- When you upload any package--including artifacts--onto a Chef Habitat Supervisor, Chef Habitat uses the public origin key to authorize the artifact's installation; Chef Habitat only installs artifacts that it can verify were signed with the private origin key
```
christophermaier, line 37: There isn't a distinction to be drawn between packages and artifacts. Also, I wouldn't describe installing packages for a Supervisor to be an "upload" operation.

Again, I think that the point made here on line 37 is the same as the one made immediately below on line 38, but line 38 is more direct and to-the-point.

components/docs-chef-io/content/habitat/origin-keys.md
```diff
@@ -119,7 +119,7 @@ When you create an origin through the site, Chef Habitat Builder automatically g
 The Chef Habitat CLI creates origin key pairs through two different commands, for two different uses:
 
 - Use [`hab setup`]({{< relref "install-habitat.md" >}}) to generate your first origin key pair as part of setting up the `hab` CLI
-- Use the `hab origin key generate <ORIGIN>` command to create an key pair for an origin created with the `hab origin create` command
+- Use the `hab origin key generate <ORIGIN>` command to create an key pair for an origin
```
christophermaier, line 122: "a key pair", rather than "an key pair"

components/docs-chef-io/content/habitat/origin-rbac.md
```diff
@@ -40,7 +40,7 @@ Member
...
+: Origin 'Members' from Chef Habitat versions before 1.6.140 are now called 'Maintainers'. This role has full read and write access to packages, channels, origin membership, jobs, integrations, invitations, settings. However, the 'Maintainer' role is more limited than the past role, in that 'Maintainers' only have read access to packages, channels, origin membership, jobs, keys, integrations, and settings. Origin 'Maintainers' can read origin membership roles and see and send invitations, but they cannot otherwise change origin membership--their own or anybody else's. Finally, 'Maintainers' can neither read nor write origin secrets.
```
christophermaier, line 43: 👌

christophermaier (changes requested):

components/docs-chef-io/content/habitat/origin-keys.md
kagarmoe: Changed to make symmetrical. I'm keeping the distinction between Supervisor and Studio because it is consistent with the way marketing is talking about the Studio. I get that it isn't a difference that makes a difference for Habitat, but I'm trying to be consistent with the story we're telling our users.

christophermaier (approved):

PR #7957, opened 2020-10-28
1.6.175 Documentation, plus docs generation script update

christophermaier:
The new `hab sup secret generate-tls` subcommand broke our document
generation script, because the script assumed that subcommands would
only contain word characters, and not hyphens.

Now, it just looks for a string of non-space characters, and goes on
with life.

The documentation string for that command was also shortened, to
ensure the output remained on one line.


kagarmoe (approved):
tenor-14389923

PR #7960, opened 2020-10-29
[FIX] Minimal change to enable non-root containers again

christophermaier:
With the changes in #7887 for how key cache directories were handled,
we inadvertently broke non-root containers. The Supervisor creates a
cache directory explicitly when it starts (previously, it was
implicitly created). However, when running as a non-root user, there
isn't a suitable directory present.

Here, we modify the Dockerfile template to create a cache directory
for the non-root Supervisor to use.

Further cleanup and reconsideration of this pattern should be done at
some point.

There's also a tweak to the e2e test to help catch things like this earlier.

### DIFF
--- a/components/pkg-export-container/defaults/Dockerfile.hbs
+++ b/components/pkg-export-container/defaults/Dockerfile.hbs
@@ -42,7 +42,12 @@ RUN \
     # communication pipe.
     mkdir /root && chmod 750 root && \
     mkdir /tmp && chmod 1777 /tmp && \
-    mkdir -p /var/tmp && chmod 1777 /var/tmp
+    mkdir -p /var/tmp && chmod 1777 /var/tmp && \
+
+    # This is currently needed in order for non-root containers to
+    # function... this ends up being their home directory, and we
+    # expect to find key cache directory there right now.
+    mkdir -p /.hab/cache/keys && chmod -R 770 /.hab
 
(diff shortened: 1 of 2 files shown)


jeremymv2 (approved):

components/pkg-export-container/defaults/Dockerfile.hbs
```diff
@@ -42,7 +42,12 @@ RUN \
     # communication pipe.
     mkdir /root && chmod 750 root && \
     mkdir /tmp && chmod 1777 /tmp && \
-    mkdir -p /var/tmp && chmod 1777 /var/tmp
+    mkdir -p /var/tmp && chmod 1777 /var/tmp && \
+
+    # This is currently needed in order for non-root containers to
+    # function... this ends up being their home directory, and we
+    # expect to find key cache directory there right now.
+    mkdir -p /.hab/cache/keys && chmod -R 770 /.hab
```
smacfarlane, line 50: Double checking my understanding here:  This path is the result of `hab` being run as a user, so we use the `$HOME/.hab` directory for a cache, and `$HOME` being unset in the running container,  correct?

smacfarlane (approved):
Looks great, just one small question for my own education.

components/pkg-export-container/defaults/Dockerfile.hbs
christophermaier: yup

PR #7961, opened 2020-10-30
Update page names and page links

IanMadd:

- Replace hyphens in page names with underscores.
- Add `relref` to all links.
- Fix links so they go to the correct pages.
- Add page aliases so we can handle redirects from existing Hab docs site.
   Redirects aren't added for the following pages because they've been split into multiple pages:
   - [Reference](https://www.habitat.sh/docs/reference/)
   - [Diagrams](https://www.habitat.sh/docs/diagrams/)
   - [Glossary](https://www.habitat.sh/docs/glossary/)

This replaces #7952


kagarmoe (approved):
tenor-261761397

mjingle (approved):
tenor-210331183

PR #7967, opened 2020-11-04
Add redirects for api pages

IanMadd:

Add aliases for the two API pages.


kagarmoe (approved):
tenor-81200567

PR #7969, opened 2020-11-04
fix studio failure building outside system drive

mwrock:
Trying to build a package or enter a studio when not in the same drive as hab, will result in failure. This makes sure that the studio command (`hab-studio.ps1`) is correctly rooted in its lettered directory and also makes sure to mount the artifacts from the correct target path on `c:`.


christophermaier (approved):
Looks good... thanks @mwrock !

Issue #6471, opened 2019-04-22
Add a utility function for dealing with required/expected Clap arguments


christophermaier:
Frequently, our Clap argument processing code has many `unwrap()` calls, because the argument/option in question is either required, or has a default value, meaning we can safely assume there will always be a value.

Given the way that Clap code is currently defined in several place, it's easy to imagine a potential refactoring that might change either end of this relationship (i.e., argument definition --> argument processing) in a way that could break things.

Furthermore, using `expect(message)` is more informative than `unwrap()` because the former gives us source code location, as well as a helpful message. 

It would be nice to add a helper function to consolidate this information, and then use it throughout the Clap code to ensure consistency and safety.

```rust
fn required_value_of<'a>(matches: &'a ArgMatches<'a>, name: &str) -> &'a str {
    matches.value_of(name)
           .expect(&format!("{} CLAP reqired arg missing", name))
}
```

See https://github.com/habitat-sh/habitat/pull/6450#discussion_r276858135

PR #7984, opened 2020-11-12
added utility function required_value_of and refactored

sajjaphani:
Added a utility function required_value_of and refactored the code to use this utility function Clap argument parsing.

The ArgMatches is being passed as a reference, it is unnecessary to pass it as a reference when calling other functions. Removed unnecessary re-referencing the ArgMatches ('&m') when passing the argument.

https://github.com/habitat-sh/habitat/issues/6471

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -1258,8 +1258,7 @@ async fn sub_pkg_delete(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
 }
 
 fn sub_pkg_verify(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
-    let src = required_value_of(m, "SOURCE");
-    let src = Path::new(src);
+    let src = Path::new(required_value_of(m, "SOURCE"));
     let key_cache = key_cache_from_matches(m)?;
     init()?;
 
@@ -1267,16 +1266,14 @@ fn sub_pkg_verify(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
 }
 
 fn sub_pkg_header(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
-    let src = required_value_of(m, "SOURCE");
-    let src = Path::new(src);
+    let src = Path::new(required_value_of(m, "SOURCE"));
     init()?;
 
     command::pkg::header::start(ui, &src)
 }
 
 fn sub_pkg_info(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
-    let src = required_value_of(m, "SOURCE");
-    let src = Path::new(src);
+    let src = Path::new(required_value_of(m, "SOURCE"));
     let to_json = m.is_present("TO_JSON");
     init()?;
 
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -1310,8 +1310,7 @@ async fn sub_pkg_channels(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
 async fn sub_svc_set(m: &ArgMatches<'_>) -> Result<()> {
     let remote_sup_addr = remote_sup_from_input(m)?;
     let remote_sup_addr = SrvClient::ctl_addr(remote_sup_addr.as_ref())?;
-    let service_group = required_value_of(m, "SERVICE_GROUP");
-    let service_group = ServiceGroup::from_str(service_group)?;
+    let service_group = required_value_of(m, "SERVICE_GROUP").parse::<ServiceGroup>()?;
     let mut ui = ui::ui();
     let mut validate = sup_proto::ctl::SvcValidateCfg::default();
     validate.service_group = Some(service_group.clone().into());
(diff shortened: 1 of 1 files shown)


jeremymv2:
@sajjaphani one of the things you will need to do on every commit is add your DCO (sign your commit) in order for the CI checks to pass. You can read about how to do that [here](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#dco-sign-off-methods)

components/hab/src/main.rs
```diff
@@ -1309,7 +1317,8 @@ async fn sub_pkg_channels(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
 async fn sub_svc_set(m: &ArgMatches<'_>) -> Result<()> {
     let remote_sup_addr = remote_sup_from_input(m)?;
     let remote_sup_addr = SrvClient::ctl_addr(remote_sup_addr.as_ref())?;
-    let service_group = ServiceGroup::from_str(m.value_of("SERVICE_GROUP").unwrap())?;
+    let service_group = required_value_of(m, "SERVICE_GROUP");
+    let service_group = ServiceGroup::from_str(service_group)?;
```
christophermaier, line 1321: Variable bindings

christophermaier: Additionally, this would be more idiomatically written as `service_group.parse()?`, rather than `ServiceGroup::from_str(service_group)`.

From [the documentation](https://doc.rust-lang.org/std/str/trait.FromStr.html):
> FromStr's from_str method is often used implicitly, through str's parse method. See parse's documentation for examples.

components/hab/src/main.rs
```diff
@@ -1495,12 +1504,14 @@ async fn sub_svc_stop(m: &ArgMatches<'_>) -> Result<()> {
 }
 
 async fn sub_file_put(m: &ArgMatches<'_>) -> Result<()> {
-    let service_group = ServiceGroup::from_str(m.value_of("SERVICE_GROUP").unwrap())?;
+    let service_group = required_value_of(m, "SERVICE_GROUP");
+    let service_group = ServiceGroup::from_str(service_group)?;
```
christophermaier, line 1508: Variable bindings, and `parse`.

components/hab/src/main.rs
```diff
@@ -1495,12 +1504,14 @@ async fn sub_svc_stop(m: &ArgMatches<'_>) -> Result<()> {
...
     let remote_sup_addr = remote_sup_from_input(m)?;
     let remote_sup_addr = SrvClient::ctl_addr(remote_sup_addr.as_ref())?;
     let mut ui = ui::ui();
     let mut msg = sup_proto::ctl::SvcFilePut::default();
-    let file = Path::new(m.value_of("FILE").unwrap());
+    let file = required_value_of(m, "FILE");
+    let file = Path::new(file);
```
christophermaier, line 1514: Variable bindings.

components/hab/src/main.rs
```diff
@@ -1664,17 +1675,18 @@ fn sub_ring_key_import(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
 }
 
 fn sub_service_key_generate(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
-    let org = org_param_or_env(&m)?;
-    let service_group = ServiceGroup::from_str(m.value_of("SERVICE_GROUP").unwrap())?;
-    let key_cache = key_cache_from_matches(&m)?;
+    let org = org_param_or_env(m)?;
+    let service_group = required_value_of(m, "SERVICE_GROUP");
+    let service_group = ServiceGroup::from_str(service_group)?;
```
christophermaier, line 1680: Variable bindings and `parse`.

christophermaier (changes requested):

components/hab/src/main.rs
```diff
@@ -1669,8 +1666,7 @@ fn sub_ring_key_import(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
 
 fn sub_service_key_generate(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
     let org = org_param_or_env(m)?;
-    let service_group = required_value_of(m, "SERVICE_GROUP");
-    let service_group = ServiceGroup::from_str(service_group)?;
+    let service_group = required_value_of(m, "SERVICE_GROUP").parse::<ServiceGroup>()?;
```
christophermaier, line 1669: Unlike the two instances above, with this call, you don't have to supply a type hint (using the `::<>` "turbofish" operator). The compiler has enough information here to know that `parse` is expected to return a `ServiceGroup`. That's because `command::service::key::generate::start` takes a `&ServiceGroup` as an argument. In the other calls, a service group was only being used to create a String, without any intermediate explicit typing that would indicate that we should have parsed to a `ServiceGroup`.

`parse` is one of the common times you'll need to use the `::<>` operator, but if it's not required, it's generally considered good form to leave it off and let the compiler do the work for you.

components/hab/src/main.rs
```diff
@@ -607,21 +607,23 @@ async fn sub_origin_key_upload(ui: &mut UI, m: &ArgMatches<'_>) -> Result<()> {
...
+            let with_secret = m.is_present("WITH_SECRET");
+            command::origin::key::upload_latest::start(ui,
+                                                       &url,
+                                                       &token,
+                                                       &origin,
+                                                       with_secret,
+                                                       &key_cache).await
+        },
+        None => {
+            let keyfile = Path::new(required_value_of(m, "PUBLIC_FILE"));
```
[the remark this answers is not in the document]
christophermaier: Actually, I was mistaken... this is defined in an argument group with `origin`, so once we get down to this part of the code, the value is effectively required. Sorry for the confusion!

christophermaier (approved):

christophermaier (approved):

PR #7993, opened 2020-11-17
[CI] Move documentation generation to release pipeline

christophermaier:
To harmonize with the new company-wide way of managing our
documentation updates, we are changing how we generate our
documentation.

Previously, once a release had been promoted to stable, we generated
two pieces of dynamically-updatable documentation, created a PR for
them, and then merged them back into the repository.

Now, we generate these docs during our release pipeline, for every
potential release candidate. Rather than storing the documentation
back in the repository, we save them in a tarball (specifically,
`generated-documentation.tar.gz`) in our S3 bucket
with the rest of the artifacts we generate. Later on, this artifact
will be used to assemble the final documentation that will actually be
published (that part of the process is owned by release engineering;
this PR simply makes what changes are needed to enable their work).

Since we are no longer committing generated content to the repository,
and in order to prevent future confusion, the content of the existing
documentation files has been replaced with a placeholder comment
explaining the situation.


components/docs-chef-io/content/habitat/habitat_cli.md
```diff
@@ -1,3275 +1,3 @@
...
-
-aliases = ["/habitat/habitat-cli/"]
-
-[menu]
-  [menu.habitat]
-    title = "hab CLI Reference"
-    identifier = "habitat/reference/habitat-cli CLI Reference"
-    parent = "habitat/reference"
-    weight = 10
-+++
```
IanMadd, line 13: We should keep this page metadata. Hugo won't build without those variables. Same with `service_templates.md`. Otherwise it looks good to me.

.expeditor/scripts/release_habitat/update_documentation.sh
```diff
@@ -0,0 +1,68 @@
...
+tempdir="$(mktemp --directory --tmpdir="$(pwd)" -t "docs-XXXX")"
+
+cp components/sup/doc/* "${tempdir}"
+
+npm install json-schema-ref-parser@6.1.0
+node .expeditor/scripts/release_habitat/generate-template-reference.js \
+     "${tempdir}"/render_context_schema.json > "${docs_dir}/service_templates.md"
+
+echo "--- :package: Packaging generated documentation into a tarball"
+docs_tarball="${docs_dir}-${version}.tar.gz"
```
schisamo, line 59: I think we should leave the version out of the tarball name. The version will already be part of the underlying URL (e.g. `https://packages.chef.io/files/habitat/X.Y.Z/generated-documentation.tar.gz` or `https://packages.chef.io/files/stable/habitat/latest/generated-documentation.tar.gz`) and making it part of the filename will just make it more difficult to deal with on the consuming end (e.g. in `chef/chef-web-docs`).

schisamo (changes requested):

components/docs-chef-io/content/habitat/habitat_cli.md
christophermaier: @IanMadd These are the old files that are destined to be overwritten by the generated ones. Are you saying that we need to have the metadata in place in order to build the site for "everyday" work on the site, too?

.expeditor/scripts/release_habitat/update_documentation.sh
christophermaier: Good point; I'll change that now.

christophermaier: Addressed in 9b14823

schisamo (approved):

components/docs-chef-io/content/habitat/habitat_cli.md
IanMadd: @christophermaier Yep, leaving that metadata out will make Hugo fail if we try to do build previews with Netlify or just preview the docs locally with `make serve`. The other option is to just delete the files entirely, but I think  it's better to have them there and then include the note stating that the content is generated.

christophermaier: @IanMadd Gotcha, thanks for catching that!

Addressed in 58b9714

IanMadd (approved):
👍

PR #8017, opened 2020-12-02
Update how we preview documentation

IanMadd:

The big thing is that this deletes the chef-web-docs submodule so we don't have to update it anymore.

In place of the chef-web-docs submodule, the `make serve` and `make preview_netlify` commands will clone down chef-web-docs into `components/docs-chef-io` and then run the usual build commands as before. 

I added `chef-web-docs` to the .gitignore so it doesn't get added to the repo.

I cleaned up the .gitignore in `components/docs-chef-io` a bit. There were some things in there that only matter in chef-web-docs, some things that are related to Sphinx 🤮, and some redundant things that are already covered in the .gitignore at the base of the repo.

I updated the docs makefile to handle this new process.

I also edited the Readme. I removed a bunch of content that should just be in the Style Guide. I updated things to reflect how things are with this configuration, and I also added the new Progress Chef docs email.

I think that's it.


kagarmoe (approved):
tenor-215763070

PR #8023, opened 2020-12-03
update mio crate and explicitly close named pipe handle after use

### LINKED ISSUE #7522: relax version pin on mio ceate in supervisor
As of the date this issue was opened mio is currently at 0.7.0. This includes some breaking changes. Most are fairly straight forward however other crates are brining in 0.6.21 as well and I have not been able to get around:

```
error[E0599]: no function or associated item named `new` found for type `mio::poll::Poll` in the current scope
   --> components\sup\src\manager\service\pipe_hook_client.rs:134:26
    |
134 |         let poll = Poll::new()?;
    |                          ^^^ function or associated item not found in `mio::poll::Poll`
```

mwrock:
fixes #7522

A customer reported that the supervisor was leaking handles on windows. After some investigation it appeared 2 handles were leaking on each health check. One was a file handle (for the named pipe) and the other was a IO completion port handle setup by the mio polling. I had spent quite a bit of time looking at the mio and miow libraries and found a way to close each handle. Later I discovered that we can just update mio and close the one named pipe handle which also closes the IOCP handle (I'm not sure why). So this seems the fastest way to resolve this for us.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,7 +150,7 @@ dependencies = [
  "futures-channel",
  "futures-util",
  "log 0.4.11",
- "mio",
+ "mio 0.6.22",
  "mio-uds",
  "num_cpus",
  "slab",
@@ -1632,8 +1632,7 @@ dependencies = [
  "libc",
  "log 0.4.11",
  "log4rs",
- "mio",
- "mio-named-pipes",
+ "mio 0.7.6",
  "notify",
  "num_cpus",
  "parking_lot",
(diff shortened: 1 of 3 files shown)


christophermaier (changes requested):
I think this also closes #7522, correct?

Looks like this needs a rebase, and another look at the `toml` dependency change, but other than that looks good! Glad it was a relatively easy fix!

christophermaier (approved):

PR #8024, opened 2020-12-03
Remove placeholders for generated pages

IanMadd:

These have to be deleted because of the way Hugo handles vendored files in chef-web-docs.
See chef/chef-web-docs#2780


kagarmoe (approved):
tenor-208529076

Issue #7866, opened 2020-08-11
finish structopt cli implementation and remove old clap parsing


jeremymv2:
Maintaining both structopt and native clap argument validation and marshaling in our CLI code base has become a non-trivial overhead.

An important tech-debt task should entail converting the full CLI code base to structopt and removing the old clap parsing and marshaling functions.

PR #8027, opened 2020-12-04
Migrate cli implementation from clap to structopt

sajjaphani:
https://github.com/habitat-sh/habitat/issues/7866

Removes the clap parsing logic and implements the struct opt cli implementations.


components/hab/src/cli/hab/pkg.rs
```diff
@@ -418,9 +414,17 @@ pub struct PkgExec {
...
+#[derive(ConfigOpt, StructOpt)]
+#[structopt(name = "env", no_version)]
+pub struct PkgEnv {
+    #[structopt(flatten)]
+    pkg_ident: PkgIdent,
+}
+
 /// Generates a blake2b hashsum from a target at any given filepath
 #[derive(ConfigOpt, StructOpt)]
-#[structopt(name = "hash", no_version, settings = &[AppSettings::Hidden])]
```
christophermaier: Why this change?

christophermaier (commented):
Looking good so far!

components/hab/src/cli/hab/cli.rs
```diff
@@ -18,7 +18,6 @@ arg_enum! {
 /// Commands relating to Habitat runtime config
 pub enum Cli {
     /// Sets up the CLI with reasonable defaults
```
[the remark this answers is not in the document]
sajjaphani: Corrected!

components/hab/src/cli/hab/cli.rs
```diff
@@ -4,7 +4,7 @@ use configopt::ConfigOpt;
 use structopt::StructOpt;
 
 arg_enum! {
-    #[derive(Deserialize)]
+    #[derive(Serialize, Deserialize)]
```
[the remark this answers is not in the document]
sajjaphani: Serialize required by the bound in toml::Value::try_from

sajjaphani:
Removed all the **clap_app** macro invocations of clap and converted the options into **structopt** structures. The remaining work related to consuming the Structure returned by **structopt** will be taken into a separate PR.

components/hab/src/cli.rs
```diff
@@ -367,16 +358,17 @@ impl FromStr for KeyType {
...
-    Arg::with_name("CACHE_KEY_PATH").long("cache-key-path")
-                                    .validator(util::non_empty)
-                                    .env(CACHE_KEY_PATH_ENV_VAR)
-                                    .default_value(&*CACHE_KEY_PATH_DEFAULT)
-                                    .help("Cache for creating and searching for encryption keys")
+impl fmt::Display for KeyType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            KeyType::Public => write!(f, "Public"),
+            KeyType::Secret => write!(f, "Secret"),
```
christophermaier, line 455: Unless there's a clear reason otherwise, I think it's good practice for `Display` and `FromStr` to be inverses of each other. That is, the following should hold: 

```rust
assert_eq!(original_string, original_string.parse::<MyType>()?.to_string()?);
```

Here, that's just a long way of saying that `"Public"` and `"Secret"` should be lower-case :sweat_smile:

components/hab/src/cli.rs
```diff
@@ -322,9 +330,7 @@ pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
             (aliases: &["u", "us", "use"])
             (@setting ArgRequiredElseHelp)
             (@setting SubcommandRequiredElseHelp)
-            (subcommand: UserKey::clap().aliases(&["stu", "stud", "studi"])
```
christophermaier: Good catch; I'm not sure how these aliases got here in the first place!

components/hab/src/cli.rs
```diff
@@ -115,232 +216,216 @@ pub enum UserKey {
...
+}
+
+#[derive(StructOpt)]
+#[structopt(name = "key", no_version, aliases = &["k", "ke"], settings = &[AppSettings::ArgRequiredElseHelp, AppSettings::SubcommandRequiredElseHelp])]
+/// Commands relating to Habitat origin key maintenance
+pub enum OriginKey {
+    #[structopt(no_version, aliases = &["d", "do", "dow", "down", "downl", "downlo", "downloa"])]
+    Download(KeyDownload),
+    #[structopt(no_version, aliases = &["e", "ex", "exp", "expo", "expor"])]
+    Export(OritinKeyExport),
```
christophermaier, line 405: Spelling error.

components/hab/src/cli.rs
```diff
@@ -115,232 +216,216 @@ pub enum UserKey {
...
-        (subcommand: alias_apply)
-        (subcommand: alias_install)
-        (subcommand: alias_run)
-        (subcommand: alias_setup)
-        (subcommand: alias_start)
-        (subcommand: alias_stop)
-        (subcommand: alias_term)
-        (after_help: AFTER_HELP)
-    )
+    Options::clap()
```
christophermaier, line 428: Is this not the same as `Hab::clap()` above?

christophermaier (commented):
A few changes to make, but overall this looks good! I think with this we can remove the feature flag for a structopt CLI, since that's now the only choice.

components/hab/src/cli.rs
sajjaphani: Corrected!

components/hab/src/cli.rs
sajjaphani: Yup, here is the updated code.
```
pub fn get(feature_flags: FeatureFlag) -> App<'static, 'static> {
    if feature_flags.contains(FeatureFlag::STRUCTOPT_CLI) {
        return Hab::clap();
    }

    Options::clap()
}
```

components/hab/src/cli.rs
sajjaphani: 😅

components/hab/src/cli.rs
christophermaier: Originally, when we were first rolling out `structopt` to represent parts of the CLI, we hid it behind a feature flag so users could opt in and give feedback, but if anything went wrong, there was an "escape hatch".

Now, however, the CLI is only going to be implemented using `structopt`. In that case, it seems that `Hab::clap()` and `Options::clap()` are going to be the same. Thus, `Options` is basically a duplicate of `Hab`, which means we don't need `Options` at all. It furthermore implies that we can remove the `STRUCTOPT_CLI` feature flag altogether.

Are there meaningful differences between `Hab` and `Options`, or can the latter be removed?

sajjaphani: I see some **aliases** are being explicitly set on certain subcommands, which are not present in **Hab::clap()** part, and also see some settings are being set **settings = &[AppSettings::ArgRequiredElseHelp, AppSettings::SubcommandRequiredElseHelp]**

components/hab/src/cli/hab.rs
```diff
@@ -1,42 +1,154 @@
...
+                  ConfigOptJobCancel,
+                  ConfigOptJobDemote,
+                  ConfigOptJobPromote,
+                  ConfigOptJobStart,
+                  ConfigOptJobStatus,
+                  JobCancel,
+                  JobDemote,
+                  JobPromote,
+                  JobStart,
+                  JobStatus},
```
christophermaier, line 37: While I generally dislike `*` imports as a rule, I think it could be helpful for the `use` statements that we've got in this corner of the code, simply because of the sheer volume of imports, as well as the fact that many of those imports are auto-generated `ConfigOpt` ones, which tends to make things look a bit confusing.

So, this highlighted code could be collapsed to `use self::bldr::*`, with similar replacements for the other module declarations, both in this file and in others that this PR is touching.

What do you think?

christophermaier (approved):
I have one comment, but otherwise I think this looks good. Thanks for tackling this!

components/hab/src/cli/hab.rs
sajjaphani: Agree in this case!

PR #8036, opened 2020-12-10
Improve handling of Supervisor LOCK file

### LINKED ISSUE #7501: supervisor will not start if LOCK file is orphaned with active PID
Related to [this comment](https://github.com/habitat-sh/habitat/issues/5903#issuecomment-523615597), the supervisor fails to start if the launcher and supervisor were forcefully terminated leaving the `LOCK` file and the PID in that file is being used by a different process.

We should probably check if the active process is a launcher process.

christophermaier:
The Supervisor uses a `LOCK` file (`/hab/sup/default/LOCK`, to be
specific) to ensure that at most only one Supervisor process may run
on a host at a time. Most of the time, this works fine, but on
occasion it can backfire. In particular, if the Supervisor exits
without deleting the `LOCK` file (as with a sudden machine shutdown),
the Supervisor can sometimes refuse to come back up again, if the PID
contained in the `LOCK` file is an already-running process (which can
very plausibly happen in such a shutdown scenario).

Here, we refactor the behavior around the `LOCK` file to fix this
loophole. Now, rather than simply relying on the presence or absence
of the file, and whether or not the PID in the file refers to a
running process, we take advantage of file locking primitives. Now, a
running Supervisor holds a file handle to the `LOCK` file open for the
life of the Supervisor, as well as a shared file lock on the file. For
a Supervisor to start, however, it must be able to obtain an
_exclusive_ lock to the `LOCK` file. In this way, a running Supervisor
will prevent any other Supervisor from running. A "stale" `LOCK` file
won't be locked by any process, allowing a new Supervisor process to
start without a problem.

We hold a shared lock in order to allow other processses to read the
`LOCK` file, which is required for the current implementation of `hab
sup term`. Part of this change modifies the behavior of `hab sup term`
in order to be more robust. In particular, it will only terminate the
process whose PID is in the `LOCK` file if the file is currently locked.

The big behavioral change here that users may notice, however, is that
we no longer delete the `LOCK` file at all when the Supervisor shuts
down. Since file lock release and file deletion cannot together be an
atomic operation, attempting to do both operations (in either order)
allows race conditions that can lead to two Supervisors running at
once (which rather defeats the purpose of the file in the first
place!). Since stale `LOCK` files will no longer block a Supervisor
from coming up, though, we can eliminate these races by just not
deleting the file.

Fixes #7501

### DIFF
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -1070,6 +1070,37 @@ steps:
             - BUILDKITE_AGENT_ACCESS_TOKEN
             - PIPELINE_HAB_AUTH_TOKEN
 
+  - label: "[:linux: test_supervisor_lock_file_behavior]"
+    command:
+      - bash .expeditor/scripts/end_to_end/run_e2e_test.sh dev test_supervisor_lock_file_behavior
+    artifact_paths:
+      - "*.log"
+    env:
+      BUILD_PKG_TARGET: x86_64-linux
+    expeditor:
+      executor:
+        docker:
+          privileged: true
+          environment:
+            - BUILD_PKG_TARGET=x86_64-linux
+            - PIPELINE_HAB_AUTH_TOKEN
+
+  - label: "[:windows: test_supervisor_lock_file_behavior]"
+    command:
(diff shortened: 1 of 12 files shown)


mwrock (approved):
I'm sure you will be clearing out the `println!`s later but this looks very sound to me.

PR #2848, opened 2017-07-28
build clap command tree in a separate thread with increased stack size

mwrock:
The hab cli is crashing at runtime with a stack overflow. This occurs in the building of the clap commands. This is a [known issue](https://github.com/kbknapp/clap-rs/issues/86) that can be remedies by building the commands in a separate thread. While the main thread's stack size cannot be altered, we can run a separate thread and increase its size.

Signed-off-by: Matt Wrock <matt@mattwrock.com>

PR #8041, opened 2020-12-15
perform configopt parsing in separate thread to avoid stack overflow in main

mwrock:
Windows debug builds recently (last couple months) started throwing stack overflow errors in main. We ran into something similar with #2848 a few years ago. This applies a similar fix to the configopt parsing which is where the error is occurring now.

### DIFF
--- a/components/hab/src/main.rs
+++ b/components/hab/src/main.rs
@@ -133,7 +133,14 @@ async fn main() {
 
 #[allow(clippy::cognitive_complexity)]
 async fn start(ui: &mut UI, feature_flags: FeatureFlag) -> Result<()> {
-    let hab = Hab::try_from_args_with_configopt();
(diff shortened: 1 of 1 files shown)


components/hab/src/main.rs
```diff
@@ -133,7 +133,14 @@ async fn main() {
 
 #[allow(clippy::cognitive_complexity)]
 async fn start(ui: &mut UI, feature_flags: FeatureFlag) -> Result<()> {
-    let hab = Hab::try_from_args_with_configopt();
+    // We parse arguments with configopt in a separate thread to eliminate
+    // possible stack overflow crashes at runtime. OSX oe a debug Windows build,
```
christophermaier, line 137: "oe" -> "or"?

christophermaier (approved):

PR #8050, opened 2020-12-18
remove unnecessary format! from panic! in fs

mwrock:
The `format!` is unnecesary and also causes the `PanicInfo` returned in a [builder-workers panic hook](https://github.com/habitat-sh/builder/blob/e084bef4556eccf22e039f27fed1d6f5f5a3e1cf/components/builder-worker/src/server.rs#L81) to report a `message` of `None` which makes debugging challenging.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -659,21 +659,21 @@ pub fn resolve_cmd_in_pkg(program: &str, ident_str: &str) -> PathBuf {
             match find_command_in_pkg(program, pkg_install, Path::new(&*FS_ROOT_PATH)) {
                 Ok(Some(p)) => p,
                 Ok(None) => {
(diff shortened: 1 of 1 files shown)


christophermaier:
I agree that the `format!` call is unnecessary, but how did having it result in a panic message of `None`?

christophermaier:
Also, it looks like `PanicInfo::message` is a nightly-only API (which looks like it would always return `None` in our case, since we don't use the `core` crate). Using `PanicInfo::payload` instead looks like it should work fine, whether `format!` is used or not.

christophermaier (approved):

mwrock:
I checked and its a plan build `Using core/rust/1.46.0/20200925181007`. When the panic hook spits out the `{:?}` of the `PanicInfo` struct, its `payload` just prints as `Any`. I'll play with explicitly printing the payload.

PR #8174, opened 2021-04-16
Fix sup content.

kagarmoe:

These files were incorrectly converted. Includes light copy editing.


IanMadd (approved):

components/docs-chef-io/content/habitat/sup_guide.md
```diff
@@ -12,111 +12,134 @@ gh_repo = "habitat"
     weight = 120
 +++
```
mwrock, line 14: I actually think this entire file should be removed. I don't think we have followed this process since 2018 and it is completely unrelated to the supervisor. This is likely only going to confuse and scare people. I'm scared having looked at it.

mwrock (changes requested):

components/docs-chef-io/content/habitat/sup_guide.md
kagarmoe: I AGREE!

kagarmoe: What I saw when I read it:
tenor-253386806

mwrock (approved):

PR #8181, opened 2021-04-21
Update plan_settings.md - "Add a note for pkg_version on semver"

hhodzic:


mwrock (changes requested):
Looks like your commit is missing a DCO. Also it would be great if you could include a link to the semver spec.

mwrock (approved):

components/docs-chef-io/content/habitat/plan_settings.md
```diff
@@ -100,6 +100,10 @@ pkg_version
   {{< /foundation_tabs_panel >}}
 {{< /foundation_tabs_panels >}}
 
+{{< note >}}
+Habitat uses the guidelines of the semantic versioning specification to determine version order (see https://semver.org/). Keep in mind that including a hyphen in your version indicates that the package is a pre-release that will be given lower precedence to other versions.
```
kagarmoe, line 104: I suggest something along the lines: 

"A hyphen in your version , for example [add example], indicates a pre-release package. Pre-release packages have a lower precedence than other versions, which means [ELIF]"

kagarmoe (commented):
Could you add an example and explain what happens if I use a pre-release package?

PR #8249, opened 2021-07-13
remove dependency on rust-crypto

sajjaphani:
We depend on rust-crypto crate (crypto::util::fixed_time_eq(x: &[u8], y: &[u8])) to check the equality of two time references in fixed time. This crate is no longer maintained.
Similar functionality is offered by sodiumoxide with the function sodiumoxide::utils::memcmp(x: &[u8], y: &[u8]). 

This function 
https://docs.rs/sodiumoxide/0.2.7/sodiumoxide/utils/fn.memcmp.html

Signed-off-by: Phani Sajja <psajja@progress.com>


mwrock (approved):
excellent!

PR #8293, opened 2021-08-27
replace rants dependency with nats.io crate

pozsgaic:
Signed-off-by: pozsgaic <chris.pozsgai@progress.com>


components/sup/src/event/nats_message_stream.rs
```diff
@@ -34,74 +83,92 @@ pub struct NatsMessageStream(pub(super) UnboundedSender<NatsMessage>);
...
-                               packet.subject);
-                    } else {
-                        error!("Failed to publish message to subject '{}', err: {}",
-                               packet.subject, e);
-                    }
-                }
+                let subj = format!("{}", packet.subject);
+                println!("Have packet to send for subject {}", subj);
+
+                let res = nats_client.publish(&subj, packet.payload()).await;
```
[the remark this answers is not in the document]
pozsgaic: I am not clear on where connection is created.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -34,74 +83,92 @@ pub struct NatsMessageStream(pub(super) UnboundedSender<NatsMessage>);
...
+            let res = time::timeout(timeout, nats_client.connect())
+                .await
+                .map_err(|_| Error::ConnectNatsServer)?;
+            println!("result from timeout = {:?}", res);
+        }
+        else {
+            println!("About to connect - no timeout specified");
+            let nats_opts: Options = Options::with_token(&token_str) 
+                .add_root_certificate(&cert_path);
+            tokio::spawn(async move { nats_client.connect().await });
```
mwrock, line 118: Not sure why this comment disappeared before!

Basically I think you can do with the `NatsClient` and replace the above if/else with:

```
        let connection = match config.connect_method.into() {
            Some(timeout) => {
                println!("About to connect with timeout");
                time::timeout(timeout, nats_options.connect(&config_url))
                    .await
                    .map_err(|_| Error::ConnectNatsServer)?.unwrap()
            }
            None => {
                println!("About to connect - no timeout specified");
                let nats_opts: Options = Options::with_token(&token_str) 
                    .add_root_certificate(&cert_path);
                nats_options.connect(&config_url).await.unwrap()
            }
        };
```

components/sup/src/event/nats_message_stream.rs
```diff
@@ -40,57 +43,56 @@ impl NatsMessageStream {
...
+        //  Not sure how to add a vector into the nats_options object below (in Options::with_token(...))
+        //  when the interface is add_root_certificate
+        let mut hab_certs: Vec<PathBuf> = Vec::new();
+        for certificate in habitat_core::tls::native_tls_wrapper::certificates_pathbuf(None)? {
+            hab_certs.push(certificate);
         }
-        let tls_connector = tls_connector.build()?;
-        client.set_tls_config(tls_connector).await;
+
+        let nats_options  = match server_certificate {
```
mwrock, line 55: make this `let mut nats_options  = match server_certificate {` so that the `nats_options` are mutable and can be reassigned to the same variable as you add certs.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -40,57 +43,56 @@ impl NatsMessageStream {
...
+                let cert_path: PathBuf = nats_options.into(); 
+                Options::with_token(&token.to_string())
+                    .with_name(&name)
+                    .add_root_certificate(cert_path)
+            }
+            None => {
+                Options::with_token(&token.to_string()) 
+                    .with_name(&name)
+            }
+        };
```
mwrock, line 66: Now you can add:
```
        for certificate in habitat_core::tls::native_tls_wrapper::certificates_pathbuf(None)? {
            nats_options = nats_options.add_root_certificate(certificate);
        }
```

pozsgaic: That worked.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -40,57 +43,56 @@ impl NatsMessageStream {
...
+        let connection : Connection = match connect_method.into() {
+            Some(timeout) => {
+                println!("Timeout used -> {:?}", timeout);
+                time::timeout(timeout,nats_options.connect(&url.to_string()))
+                    .await
+                    .map_err(|_| Error::ConnectNatsServer)??
+            } 
+            None => {
+                println!("Timeout not used");
+                nats_options.connect(&url.to_string())
```
mwrock, line 83: looking over the rants documentation it now becomes clear why the former implementation called this in a separate thread. The rants client will continuously try to reconnect if unable to connect or if it becomes disconnected. It looks like the nats library has similar functionality.

components/sup/Cargo.toml
```diff
@@ -38,6 +38,7 @@ lazy_static = "*"
 libc = "*"
 log = "*"
 log4rs = "*"
+nats = { git="https://github.com/nats-io/nats.rs", branch="main"}
```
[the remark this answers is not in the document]
pozsgaic: Updated

components/sup/src/event/nats_message_stream.rs
pozsgaic: That worked - thanks!

components/hab/src/cli/hab/sup.rs
```diff
@@ -32,7 +32,7 @@ use habitat_core::{env::Config,
                    fs::HAB_CTL_KEYS_CACHE,
                    package::PackageIdent,
                    util as core_util};
-use rants::{error::Error as RantsError,
+use rants::{error::Error as NatsError,
```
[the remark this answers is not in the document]
pozsgaic: Will remove.  Have not phased out Rants::Address yet but will do so and remove all Rants references for next commit

components/sup/Cargo.toml
```diff
@@ -38,6 +39,7 @@ lazy_static = "*"
 libc = "*"
 log = "*"
 log4rs = "*"
+nats = "*"
```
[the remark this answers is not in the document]
pozsgaic: Will do before next commit

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,14 +1,17 @@
...
                    EventStreamConfig,
                    Result};
 use futures::{channel::{mpsc as futures_mpsc,
                         mpsc::UnboundedSender},
               stream::StreamExt};
-use rants::{error::Error as RantsError,
-            native_tls::TlsConnector,
-            Client,
-            Subject};
+use rants::Subject;
```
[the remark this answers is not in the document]
pozsgaic: will replace w/ nats equivalent

components/sup/src/event.rs
```diff
@@ -50,14 +48,14 @@ lazy_static! {
...
-    static ref SERVICE_STARTED_SUBJECT: Subject =
-        "habitat.event.service_started".parse().expect("valid NATS subject");
-    static ref SERVICE_STOPPED_SUBJECT: Subject =
-        "habitat.event.service_stopped".parse().expect("valid NATS subject");
-    static ref SERVICE_UPDATE_STARTED_SUBJECT: Subject =
-        "habitat.event.service_update_started".parse().expect("valid NATS subject");
-    static ref HEALTHCHECK_SUBJECT: Subject =
-        "habitat.event.healthcheck".parse().expect("valid NATS subject");
+    static ref SERVICE_STARTED_SUBJECT: String =
+        "habitat.event.service_started".to_string();
```
mwrock, line 52: since these are just strings, you can now move out of the `lazy_static!` replace with:

```
const SERVICE_STARTED_SUBJECT: &str = "habitat.event.service_started";
```
Same for the other subjects below.

components/sup/src/event.rs
```diff
@@ -207,7 +205,7 @@ fn initialized() -> bool { NATS_MESSAGE_STREAM.try_get().is_some() }
 ///
 /// If `init_stream` has not been called already, this function will
 /// be a no-op.
-fn publish(subject: &'static Subject, mut event: impl EventMessage) {
+fn publish(subject: &'static str, mut event: impl EventMessage) {
     if let Some(stream) = NATS_MESSAGE_STREAM.try_get() {
```
mwrock, line 209: I think you can change `&'static str` to `&str`

components/sup/src/event/error.rs
```diff
@@ -1,9 +1,10 @@
 //! Event subsystem-specific error handling
 
-use rants::{error::Error as RantsError,
-            native_tls};
+use native_tls;
+use tokio::time::error::Elapsed;
 use std::{error,
           fmt,
+          io:: Error as NatsError,
```
mwrock, line 7: Consider changing `NatsError` ro IOError` since an IO error could come from alot of places other than `nats`

components/sup/src/event/error.rs
```diff
@@ -50,8 +54,12 @@ impl From<habitat_core::Error> for Error {
...
 }
 
-impl From<RantsError> for Error {
-    fn from(error: RantsError) -> Self { Error::Rants(error) }
+impl From<NatsError> for Error {
+    fn from(error: NatsError) -> Self { Error::Nats(error) }
+}
+
+impl From<Elapsed> for Error {
+    fn from(error: Elapsed) -> Self { println!("TIME ELAPSED!"); Error::NatsTimeout(error) }
```
mwrock, line 62: make sure to remove the `println!`. I'm assuming that was debug code.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,30 +1,116 @@
...
+    io,
+    path::PathBuf,
+    sync::Arc,
+    time::Duration};
 
 /// The subject and payload of a NATS message.
 #[derive(Debug)]
 pub struct NatsMessage {
-    subject: &'static Subject,
+    subject: &'static str,
```
mwrock, line 18: can just be `&str`

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,30 +1,116 @@
...
     payload: Vec<u8>,
 }
 
 impl NatsMessage {
-    pub fn new(subject: &'static Subject, payload: Vec<u8>) -> Self {
+    pub fn new(subject: &'static str, payload: Vec<u8>) -> Self {
```
mwrock, line 23: `&'static str` to `&str`

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,30 +1,116 @@
...
+                trace!("Timeout not used");
+                tokio::spawn(async move { this.lock().await.connect_impl(supervisor_id, &config).await });
+            }
+        };
+    }
+
+    async fn connect_impl_with_reconnect(&mut self, supervisor_id: String, config: &EventStreamConfig) {        
+        while self.0.is_none() {
+            match Self::options_from_config(&supervisor_id, config)
+                .with_retry_first_connect()
```
mwrock, line 55: I would move these options builder calls to `options_from_config`.

components/sup/src/event/error.rs
```diff
@@ -13,7 +14,8 @@ pub enum Error {
     ConnectNatsServer,
     HabitatCore(habitat_core::Error),
     NativeTls(native_tls::Error),
-    Rants(RantsError),
+    Nats(NatsError),
```
mwrock, line 17: I'd change `Nats` to `IOError` since there is no guarantee than an io::error is going to be a Nats related error.

components/sup/src/event/error.rs
```diff
@@ -13,7 +14,8 @@ pub enum Error {
...
+    NatsTimeout(Elapsed),
```
mwrock, line 18: I'd change to `Timeout` for the same reason as above.

mwrock (changes requested):
Here are some initial review comments that I wanted to get out sooner than later. I'm gonna look at some of the threading/connecting related stuff in closer detail and will send more comments later today.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,30 +1,116 @@
...
     }
 
     pub fn payload(&self) -> &[u8] { self.payload.as_slice() }
 }
 
+#[derive(Clone)]
+pub struct NatsClient(Option<Connection>);
+
+impl NatsClient {
+    pub async fn connect(this: Arc<Mutex<Self>>, supervisor_id: String, config: EventStreamConfig) {
```
mwrock, line 34: will want to return a `Result<()>` to get the exit behavior

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,30 +1,116 @@
...
 
+#[derive(Clone)]
+pub struct NatsClient(Option<Connection>);
+
+impl NatsClient {
+    pub async fn connect(this: Arc<Mutex<Self>>, supervisor_id: String, config: EventStreamConfig) {        
+        match config.connect_method.into() {
+            Some(timeout) => {
+                trace!("Timeout used -> {:?}", timeout);
+                if let Err(e) = time::timeout(timeout, this.lock().await.connect_impl_with_reconnect(supervisor_id, &config))
```
mwrock, line 38: Need to remove the `if let` and the just use:

```
time::timeout(timeout, this.lock().await.connect_impl_with_reconnect(supervisor_id, &config)).await?
```

which will bubble up the error.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,30 +1,116 @@
...
+                    trace!("Error connecting: {}", e);
+                    //  The process does not exit.  We should shut down if timeout specified and we cannot connect.
+                    //std::process::exit(1);
+                }
+            } 
+            None => {
+                trace!("Timeout not used");
+                tokio::spawn(async move { this.lock().await.connect_impl(supervisor_id, &config).await });
+            }
+        };
```
mwrock, line 49: and add an `Ok(())` here

components/sup/src/event/nats_message_stream.rs
```diff
@@ -33,64 +119,22 @@ impl NatsMessage {
...
-        for certificate in habitat_core::tls::native_tls_wrapper::certificates(None)? {
-            tls_connector.add_root_certificate(certificate);
-        }
-        if let Some(certificate) = server_certificate {
-            tls_connector.add_root_certificate(certificate.into());
-        }
-        let tls_connector = tls_connector.build()?;
-        client.set_tls_config(tls_connector).await;
-
+    pub async fn new(supervisor_id: String, config: EventStreamConfig) -> Result<NatsMessageStream> {
```
mwrock, line 122: event is passing a `&str` so I don't think this compiles.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -33,64 +119,22 @@ impl NatsMessage {
...
         // connection, the client will handle reconnecting if necessary.
-        if let Some(timeout) = connect_method.into() {
-            time::timeout(timeout, client.connect()).await
-                                                    .map_err(|_| Error::ConnectNatsServer)?;
-        } else {
-            let client = Client::clone(&client);
-            tokio::spawn(async move { client.connect().await });
-        }
+        let client = Arc::new(Mutex::new(NatsClient(None)));
+        NatsClient::connect(client.clone(), supervisor_id, config).await;
```
mwrock, line 129: add the `?` to get the desired exiting error behavior

mwrock (commented):
Added comments reflecting our discussion earlier today about getting the sup to exit on the timeout. The main thing that confused me with this change is that I don't see how `with_retry_first_connect()` helps. My understanding was that that was going to prevent us from needing to loop and wait for the initial connection  and thus avoid all the mutex/arc shenanigans. Since that code still remains, if you leave out the `with_retry_first_connect()`, I was still seeing the desired behavior. I also found something we must not have noticed before in the nats code which is even with the retry on the first connect, nats will block until the first connect is made which throws off the "no timeout" scenario.

components/sup/src/event.rs
pozsgaic: OK - will do!

components/sup/src/event.rs
pozsgaic: Will do.

components/sup/src/event/nats_message_stream.rs
pozsgaic: Pending as related to comment:
https://github.com/habitat-sh/habitat/pull/8293#discussion_r712062829

components/sup/src/event/nats_message_stream.rs
pozsgaic: Pending as related to comment:
https://github.com/habitat-sh/habitat/pull/8293#discussion_r712062829

components/sup/src/event/nats_message_stream.rs
pozsgaic: My bad.  Missed one part when I reverted some experimental changes before I committed the code.  Will revert on next commit.

components/sup/src/event/nats_message_stream.rs
pozsgaic: Got it

components/sup/src/event/nats_message_stream.rs
pozsgaic: Got it

components/sup/src/event/nats_message_stream.rs
```diff
@@ -4,93 +4,158 @@ use crate::event::{Error,
...
+            None => {
+                trace!("Timeout not used");
+                tokio::spawn(async move { self.client.lock().await.connect(supervisor_id, &config).await });
+            }
+        }
+        Ok(())
+    }
+
+    async fn publish(&mut self, subject: &str, msg: impl AsRef<[u8]>) -> io::Result<()> {
+        self.client.lock().await.publish(subject, msg).await
```
mwrock, line 131: Looks like we still have the potential for thread contention here. If there is no nats server, then the lock in line 124 will remain open possibly indefinitely. And threads attempting to publish will all wait for a lock that may never materialize.

Might consider using a try_lock here.

components/sup/src/event.rs
pozsgaic: Updated

components/sup/src/event/error.rs
pozsgaic: Updaetd

components/sup/src/event/nats_message_stream.rs
pozsgaic: Updated

components/sup/src/event/nats_message_stream.rs
pozsgaic: The client is cloned in line 150 so I think it should be OK.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -4,93 +4,151 @@ use crate::event::{Error,
...
-    pub fn new(subject: &'static Subject, payload: Vec<u8>) -> Self {
+    pub fn new(subject: &'static str, payload: Vec<u8>) -> Self {
         NatsMessage { subject, payload }
     }
 
     pub fn payload(&self) -> &[u8] { self.payload.as_slice() }
 }
 
+#[derive(Clone)]
+struct NatsClientImpl {
```
mwrock, line 33: You can refactor the NatsClient to avoid needing the NatsClientImpl and the extra `Arc<Mutex>`. You can try:
```
#[derive(Clone)]
struct NatsClient {
    connection: Arc<Mutex<Option<Connection>>>,
}

impl NatsClient {
    fn new() -> NatsClient {
        NatsClient { connection: Arc::new(Mutex::new(None)) }
    }

    async fn connect(self, supervisor_id: String, config: EventStreamConfig) -> io::Result<()> {        
        match config.connect_method.into() {
            Some(timeout) => {
                trace!("Timeout used -> {:?}", timeout);
                
                time::timeout(timeout, self.connect_impl(supervisor_id, &config))
                    .await??;
            } 
            None => {
                trace!("Timeout not used");
                tokio::spawn(async move { self.connect_impl(supervisor_id, &config).await });
            }
        }
        Ok(())
    }

    async fn connect_impl(self, supervisor_id: String, config: &EventStreamConfig) -> io::Result<()> {        
        if self.connection.lock().await.is_none() {
            match Self::options_from_config(&supervisor_id, config) 
                .connect(&config.url.to_string())
                .await {
                    Ok(conn) => {
                        *self.connection.lock().await = Some(conn);
                    }
                    Err(e) => {
                        Error::ConnectNatsServer(e);
                    }
                }
        }
        Ok(())
    }

    fn options_from_config(supervisor_id: &str, config: &EventStreamConfig) -> Options {
        let name = format!("hab_client_{}", supervisor_id);
        let ca_certs = match habitat_core::tls::native_tls_wrapper::installed_cacerts(None) {
            Ok(ca_certs) => ca_certs,
            Err(_)         => None
        };
        match config.server_certificate {
            Some(ref nats_options) => {
                let cert_path: PathBuf = nats_options.clone().into(); 
                Options::with_token(&config.token.to_string())
                    .with_name(&name)
                    .add_root_certificate(cert_path)
                    .add_root_certificate(ca_certs.unwrap())
                    .with_retry_on_failed_connect()
            }
            None => {
                Options::with_token(&config.token.to_string()) 
                    .with_name(&name)
                    .add_root_certificate(ca_certs.unwrap())
                    .with_retry_on_failed_connect()
            }
        }
    }

    async fn publish(&self, subject: &str, msg: impl AsRef<[u8]>) -> io::Result<()> {
        if let Some(conn) = &*self.connection.lock().await {
            conn.publish(subject, msg).await
        }
        else {
            Err(io::Error::new(io::ErrorKind::Other, "Not connected to NATS server!"))
        }
    }
}
```

components/sup/src/event/nats_message_stream.rs
```diff
@@ -4,93 +4,151 @@ use crate::event::{Error,
...
         // a future that will resolve when a connection is possible. Once we establish a
         // connection, the client will handle reconnecting if necessary.
-        if let Some(timeout) = connect_method.into() {
-            time::timeout(timeout, client.connect()).await
-                                                    .map_err(|_| Error::ConnectNatsServer)?;
-        } else {
-            let client = Client::clone(&client);
-            tokio::spawn(async move { client.connect().await });
-        }
+        let mut client = NatsClient::new();
```
mwrock, line 142: With the NatsClient consolidation above, this `client` will not need to be mutable.

mwrock:
There is more stuff with the error and certs I will comment on later but wanted to get this main thing out for you to look at.

components/sup/src/event/nats_message_stream.rs
pozsgaic: I agree and like how this is evolving into less code.

mwrock (changes requested):

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+                    .with_retry_on_failed_connect()
+            }
+            None => {
+                Options::with_token(&config.token.to_string()) 
+                    .with_name(&name)
+                    .add_root_certificate(ca_certs.unwrap())
+                    .with_retry_on_failed_connect()
+            }
+        }
+    }
```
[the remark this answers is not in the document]
pozsgaic: Got it.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+                        .await
+                });
+            }
+        }
+        Ok(())
+    }
+
+    async fn connect_impl(self, supervisor_id: String, config: &EventStreamConfig) -> io::Result<()> {
+        if self.connection.lock().await.is_none() {
+            match Self::options_from_config(&supervisor_id, config)
```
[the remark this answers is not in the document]
pozsgaic: Got it

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+    fn new() -> NatsClient {
+        NatsClient { connection: Arc::new(Mutex::new(None)), }
+    }
+
+    async fn connect(self, supervisor_id: String, config: EventStreamConfig) -> io::Result<()> {        
+        match config.connect_method.into() {
+            Some(timeout) => {
+                trace!("Timeout used -> {:?}", timeout);
+                time::timeout(timeout, 
+                              self.connect_impl(supervisor_id, &config)).await??;
```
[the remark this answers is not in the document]
pozsgaic: Got it.  I had this defined at one point and refactored the code.  Now I see that we need this to go with Result instead of io::Result.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+struct NatsClient {
+    connection: Arc<Mutex<Option<Connection>>>,
+}
+
+impl NatsClient {
+    fn new() -> NatsClient {
+        NatsClient { connection: Arc::new(Mutex::new(None)), }
+    }
+
+    async fn connect(self, supervisor_id: String, config: EventStreamConfig) -> io::Result<()> {
```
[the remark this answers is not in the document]
pozsgaic: OK

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
-                        trace!("Failed to publish message to subject '{}' because the client is \
-                                not connected",
-                               packet.subject);
-                    } else {
-                        error!("Failed to publish message to subject '{}', err: {}",
-                               packet.subject, e);
-                    }
-                }
+                let res = client.publish(packet.subject, packet.payload()).await;
+                trace!("publish result: {:?}", res);
```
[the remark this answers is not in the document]
pozsgaic: OK - will be updating this.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
         // a future that will resolve when a connection is possible. Once we establish a
         // connection, the client will handle reconnecting if necessary.
-        if let Some(timeout) = connect_method.into() {
-            time::timeout(timeout, client.connect()).await
-                                                    .map_err(|_| Error::ConnectNatsServer)?;
-        } else {
-            let client = Client::clone(&client);
-            tokio::spawn(async move { client.connect().await });
-        }
+        //let mut client = NatsClient::new();
```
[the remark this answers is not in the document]
pozsgaic: OK - changed the verbage slightly for context.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+
+    async fn connect(self, supervisor_id: String, config: EventStreamConfig) -> io::Result<()> {        
+        match config.connect_method.into() {
+            Some(timeout) => {
+                trace!("Timeout used -> {:?}", timeout);
+                time::timeout(timeout, 
+                              self.connect_impl(supervisor_id, &config)).await??;
+            } 
+            None => {
+                trace!("Timeout not used");
```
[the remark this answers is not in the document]
pozsgaic: OK

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+            }
+        }
+    }
+
+    async fn publish(&self, subject: &str, msg: impl AsRef<[u8]>) -> io::Result<()> {
+        if let Some(conn) = &*self.connection.lock().await {
+            conn.publish(subject, msg).await
+        }
+        else {
+            Err(io::Error::new(io::ErrorKind::Other, "Not connected to NATS server!"))
```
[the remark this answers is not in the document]
pozsgaic: I refactored the ConnectNatsServer enum to NotConnected.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
 
     pub fn payload(&self) -> &[u8] { self.payload.as_slice() }
 }
 
+/// NatsClient is main accessor to connect to NATS Server and
+/// to publish messages to NATS.
+#[derive(Clone)]
+struct NatsClient {
+    connection: Arc<Mutex<Option<Connection>>>,
+}
```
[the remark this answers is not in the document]
pozsgaic: OK

components/sup/src/event/nats_message_stream.rs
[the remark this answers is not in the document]
pozsgaic: Good idea - there was a little redundancy there :)

components/sup/src/event/nats_message_stream.rs
```diff
@@ -1,96 +1,137 @@
...
+    async fn connect_impl(self, supervisor_id: String, config: &EventStreamConfig) -> io::Result<()> {
+        if self.connection.lock().await.is_none() {
+            match Self::options_from_config(&supervisor_id, config) 
+                .connect(&config.url.to_string())
+                .await {
+                    Ok(conn) => {
+                        *self.connection.lock().await = Some(conn);
+                    }
+                    Err(e) => {
+                        return Err(e);
```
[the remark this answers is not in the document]
pozsgaic: Got it

components/sup/src/event/error.rs
```diff
@@ -50,8 +47,8 @@ impl From<habitat_core::Error> for Error {
     fn from(error: habitat_core::Error) -> Self { Error::HabitatCore(error) }
 }
 
-impl From<RantsError> for Error {
-    fn from(error: RantsError) -> Self { Error::Rants(error) }
+impl From<IOError> for Error {
```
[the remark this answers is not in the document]
pozsgaic: OK - removed and (re) added impl for Elapsed.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -4,27 +4,99 @@ use crate::event::{Error,
...
+            let options = Self::options_from_config(&supervisor_id, config)?;
+            let conn = options
+                       .connect(&config.url.to_string())
+                       .await.
+                       map_err(|_| Error::NotConnected)?;
+            *self.0.lock().await = Some(conn);
+            Ok(())
+        } else {
+            Ok(())
+        }
```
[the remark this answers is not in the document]
pozsgaic: Updated

components/sup/src/event/nats_message_stream.rs
```diff
@@ -34,62 +106,19 @@ pub struct NatsMessageStream(pub(super) UnboundedSender<NatsMessage>);
...
-                    // never be sent.
-                    if let RantsError::NotConnected = e {
-                        trace!("Failed to publish message to subject '{}' because the client is \
-                                not connected",
-                               packet.subject);
-                    } else {
-                        error!("Failed to publish message to subject '{}', err: {}",
-                               packet.subject, e);
-                    }
+                    error!("{:?}", e);
```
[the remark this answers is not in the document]
pozsgaic: We are embedding a different client and this client does not emit an error if you have connected once and then
disconnect.  My observation is that this is a no-op and does not emit an error, so we cannot readily duplicate that logic.
We do however see same behavior when there is no connection at all and we try to publish.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -4,27 +4,99 @@ use crate::event::{Error,
...
+                    self.connect_impl(supervisor_id, &config)
+                        .await
+                });
+            }
+        }
+        Ok(())
+    }
+
+    async fn connect_impl(self, supervisor_id: String, config: &EventStreamConfig) -> Result<()> {
+        if self.0.lock().await.is_none() {
```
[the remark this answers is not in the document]
pozsgaic: If we go this route then we are implementing the reconnect, and won't need the PR.  I tested it out and it works.  Do we want to stick with the 1 sec sleep?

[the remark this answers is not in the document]
pozsgaic: Strike previous comment on max_reconnects - we need it

mwrock:
I would squash your commits first.

PR #8363, opened 2021-11-08
Revert "Revert "replace rants dependency with nats.io crate""

pozsgaic:
Reverts habitat-sh/habitat#8353


components/sup/src/event.rs
```diff
@@ -93,15 +84,15 @@ pub struct EventStreamConfig {
     pub site:               Option<String>,
     pub meta:               EventStreamMetadata,
     pub token:              EventStreamToken,
-    pub url:                Address,
+    pub url:                SocketAddr,
```
mwrock, line 87: I think we need to move away from `SocketAddr` and just use a string.  `SocketAddr` will convert domain names to an IP address. If the nats service is answering on a domain name and has a server certificate signed against that CN which in almost all cases it will, then this will fail with a TLS connection error.

components/sup/src/event/nats_message_stream.rs
```diff
@@ -4,27 +4,96 @@ use crate::event::{Error,
...
+
+    async fn connect_impl(self, supervisor_id: String, config: &EventStreamConfig) -> Result<()> {
+        while self.0.lock().await.is_none() {
+            match Self::options_from_config(&supervisor_id, config)?.connect(&config.url
+                                                                                    .to_string())
+                                                                    .await
+            {
+                Ok(conn) => *self.0.lock().await = Some(conn),
+                Err(e) => {
+                    trace!("Failed to connect to NATS server: {}", e);
```
mwrock, line 63: I ended up having issues getting this to render and had to use `println!`. I think this might have something to do with the rust `log` crate marcros being used in a separate thread but will need some research.

mwrock (changes requested):
I ended up having a few issues running this. First `localhost:4222` was getting transformed to `127.0.0.1:4222` and that resultes in TLS issues since the automate cert has a CN of `localhost`.

Next I just hardcoded the nats config to use `localhost:4222` to temporarily work around that issue. Then I got another certificate error: `TLS error: invalid certificate: CAUsedAsEndEntity`.

Just to make sure I was not using a weird cert, I switched to the 1.6.420 supervisor and it managed to connect and register the health check events.

pozsgaic:
We are going to stick with the rants crate - closing.

mwrock:
Just to add some context here. The blocker with nats.rs was that you could not add a self-signed cert to the root certificates.This resultect in `CAUsedAsEndEntity` TLS errors. This comes from nats.rs dep on rustls and its dep on webpki which is what ultimately does the validation. The ranta crate uses native_tls which seems to allow self-signed certs as a root cert.

In production the self-signed cert problems would likely not be a problem. However, Automate generates a self-signed cert as part of its default deployment and we want to continue to support that scenario without breaking changes.

PR #8364, opened 2021-11-09
add e2e test with a real automate instance

mwrock:
This includes a `Dockerfile` that builds a automate image of of the latest release of automate and then runs the supervisor connecting to that container. It then queries the chef-automate CLI to see that the application's health check was recorded.

### DIFF
--- a/.expeditor/end_to_end.pipeline.yml
+++ b/.expeditor/end_to_end.pipeline.yml
@@ -747,7 +747,8 @@ steps:
       - "*.log"
     expeditor:
       executor:
-        docker:
+        linux:
+          single-use: true
           privileged: true
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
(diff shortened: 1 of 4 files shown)


pozsgaic (commented):
This is a nice improvement to run integrated with automate.
I think we still need a way to send a health check or other event
type from https://github.com/habitat-sh/habitat/blob/e2e_automate/components/sup/src/event.rs#L53

mwrock:
The test-probe is just a dummy service that endlessly runs and includes a health check. The test queries automate's `application` command to see if the test-probe service health check has surfaced as OK. Because the applications listed with the `application` command are populated via the NATS communication from the supervisor, we know that if we see the health-check status then the NATS communication was established and the message was received.

pozsgaic (approved):
Looks good.

PR #8378, opened 2021-11-23
bump rust to 1.56.1

pozsgaic:
Signed-off-by: pozsgaic <chris.pozsgai@progress.com>


support/allowed_lints.txt
```diff
@@ -1,3 +1,5 @@
 clippy::module_inception
 clippy::new_ret_no_self
 clippy::new_without_default
+clippy::suspicious_else_formatting
+unused_must_use
```
mwrock, line 5: I think you just need to put these allows in https://github.com/davidMcneil/configopt/blob/master/configopt-derive/src/configopt_type/generate.rs and then fork that. That will isolate the exceptions to only ConfigOpt generated code.

mwrock (changes requested):

support/allowed_lints.txt
pozsgaic: We will need to update our habitat-sh/structopt code as well to fix the clippy::suspicious_else_formatting linter error

components/hab/Cargo.toml
```diff
@@ -17,7 +17,7 @@ base64 = "*"
 bitflags = "*"
 chrono = {version = "*", features = ["serde"]}
 clap = { git = "https://github.com/habitat-sh/clap.git", branch = "v2-master", features = [ "suggestions", "color", "unstable" ] }
-configopt = { git = "https://github.com/davidMcneil/configopt.git" }
+configopt = { git = "https://github.com/pozsgaic/configopt.git" }
```
mwrock, line 20: we should fork this into the habitat-sh org.

mwrock (changes requested):

components/hab/Cargo.toml
pozsgaic: I forked to habitat-sh and renamed master to main in the process

mwrock (changes requested):
We want to remove changes to VERSION and CHANGELOG.md since those should only be auto generated by expeditor.

pozsgaic:
I don't recall adding this but maybe I pulled it in merging from main?  Will fix.

mwrock (approved):

PR #8380, opened 2021-12-01
ensure hook interpreter ps module path is always before windows ps module path

### LINKED ISSUE #8366: Core/powershell package references native Windows libraries causing errors
Customer bug : https://github.com/chef/customer-bugs/issues/492

mwrock:
closes #8366 

Under normal circumstances, entering a powershell 7 shell will put its own `Modules` path before the Windows powershell `Modules` path in the list of PSModulesPath variable. However if for some reason, the powershell 7 Modules path is already in the PSModulesPath after the windows powershell modules path before entering a ps 7 shell, the ps7 Modules path will then remain where it is giving precedence to windows powershell modules.

This PR ensures that the ps7 path is always before the windows powershell path when invoking a powershell hook.

### DIFF
--- a/components/core/src/util.rs
+++ b/components/core/src/util.rs
@@ -8,11 +8,14 @@ pub mod text_render;
 pub mod win_perm;
 
 #[cfg(windows)]
-use crate::{error::Result,
+use crate::{env as henv,
+            error::Result,
             os::process::windows_child::Child};
 
 #[cfg(windows)]
-use std::collections::HashMap;
+use std::{collections::HashMap,
+          env,
+          path::PathBuf};
 
 use std::{io::{self,
                BufRead},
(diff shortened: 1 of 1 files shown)


pozsgaic (commented):
Should we have a check on line 150 to ensure that the candidate path is a
valid hab studio path?  e.g. it should be in C:\hab for example, and not C:\temp.

mwrock:
It should be impoossible for a hook to run with a non-habitat powershell because of https://github.com/habitat-sh/habitat/blob/546fc79ea141e4d239e8cd7724a0f2fb5e5b3b3d/components/common/src/util/path.rs#L131 where it sets up the path to have the habitat interpreter (powershell for windows) at the start of the path. The goal here is mainly to make sure that the PSModulePath at the head matches the `pwsh` running in the hook.

pozsgaic (approved):
looks good

PR #8412, opened 2022-01-21
update habitat api.raml

pozsgaic:

### DIFF
--- a/components/sup/doc/api.raml
+++ b/components/sup/doc/api.raml
@@ -211,6 +211,13 @@ types:
             200:
                 body:
                     application/json:
+/metrics:
+    get:
+        description: metrics output
+        responses:
+            200:
+                body:
+                    text/plain:
 /services:
     get:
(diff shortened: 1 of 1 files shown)


mwrock:
Is there a reason why we are changing `name` to `svc` and `organization` to `org`? I thing `organization` and `name` are more clear. Also I was curious about the `/metrics` end point because I wasn't familiar with it. I just viewed it on a locally running supervisor and it looks like internal metrics only of interest internally. Not a huge deal if we make it public but I'm just wondering if this PR is really necessary.

pozsgaic:
If we don't care about the metrics, then sure, we can close it.

mwrock:
We don't really need or even want the placeholder names to match between code and docs. The code variables might be too succinct. I'd go with whatever is most descriptive.

pozsgaic:
Since the metrics is more internal in nature, we will just close this.

PR #8418, opened 2022-01-24
Update release pipeline to generate API docs

pozsgaic:

Note that this should be merged ONLY after the merging of the .raml file updates for habitat AND builder


.expeditor/scripts/release_habitat/update_api_docs.sh
```diff
@@ -0,0 +1,74 @@
...
+hub_check=$(which hub)
+if [ -z "$hub_check" ]; then
+  install_hub
+fi
+
+# Check if node is installed and install if not
+node_version=$(node -v)
+if [ -z "$node_version" ]; then
+  sudo apt-get install -y nodejs
+fi
```
mwrock, line 22: You could replace lines 19-22 with `hab pkg install core/node --binlink`

.expeditor/scripts/release_habitat/update_api_docs_dev.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
mwrock, line 1: I don't see anything referencing this script. We should remove it.

mwrock (changes requested):
This looks good to me. Just a couple small things. Have you tested this in an ad hoc release_habitat build against this branch? You'd of course need to temporarily remove the `if: build.creator.name == 'Chef Expeditor'`.

mwrock:
This PR is only for habitat and not builder and I don't see a PR for the raml. Should this still be in draft?

.expeditor/release_habitat.pipeline.yml
```diff
@@ -410,6 +410,17 @@ steps:
           environment:
             - BUILD_PKG_TARGET=x86_64-linux
 
+  - label: "Update Habitat API Docs"
```
mwrock, line 413: this should actually go in the `finish_release` pipeline since we only want this to trigger when we promote to stable.

.expeditor/scripts/release_habitat/update_api_docs.sh
pozsgaic: Agreed - will update.

.expeditor/scripts/release_habitat/update_api_docs_dev.sh
pozsgaic: It is a standalone script to run in the dev environment only.  I did not want to make the update_api_docs.sh convoluted with dev logic but still wanted to be able to test things out on my local dev setup.

.expeditor/release_habitat.pipeline.yml
pozsgaic: The supervisor raml was already in 1.0 and we decided not to update as the only difference observed was the addition of the /metrics path, which is internal.

pozsgaic: I will move this to the correct pipeline.

.expeditor/finish_release.pipeline.yml
```diff
@@ -71,3 +71,14 @@ steps:
           path: account/static/habitat/chef-ci
           field: acceptance_auth_token
     soft_fail: true
+
+  - label: "Update Habitat API Docs"
+    command:
+      - .expeditor/scripts/finish_release/update_api_docs.sh
+    if: build.creator.name == 'Chef Expeditor'
```
mwrock, line 78: this conditional should not be necesary

mwrock (changes requested):

mwrock (approved):
This looks good. Please squash the commits before merge.

PR #8429, opened 2022-02-02
Add documentation guide on rotating builder SSL certificates

chef-davin:
Signed-off-by: Davin Taddeo <davin@chef.io>

This creates a new section under habitat called `Guides` and then adds a guide on how to rotate certificates on an on-prem instance of Habitat Builder via updating the ssl-certificate files on the builder-api-proxy service.

### DIFF
--- a/components/docs-chef-io/content/habitat/guide_rotate_builder_ssl_certs.md
+++ b/components/docs-chef-io/content/habitat/guide_rotate_builder_ssl_certs.md
@@ -8,11 +8,11 @@ gh_repo = "habitat"
   [menu.habitat]
     title = "Rotate Builder SSL Certs"
     identifier = "habitat/guides/rotate-builder-certs"
-    parent = "habitat/guids"
+    parent = "habitat/guides"
     weight = 50
 +++
 
-A short guide to SSL certs used by Builder's web interface.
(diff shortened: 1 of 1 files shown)


chef-davin:
This will require an update to the menu.toml in the chef-web-docs repo.  Need to add in a section for `Guides`, after `Habitat Studio` and before `Reference`:

```toml
  [[habitat]]
  title = "Guides"
  identifier = "habitat/guides"
  parent = "habitat"
  weight = 85
```

mwrock:
The content looks good to me. Thanks for adding this!! I'll defer to @kagarmoe for approval.

chef-davin:
Should be coordinated with [Chef-Web-Docs PR #3689](https://github.com/chef/chef-web-docs/pull/3689)

IanMadd:
A couple thoughts on this.

This looks like it's just docs for on-prem Builder, and if only a relatively small percentage Hab users use on-prem, we should keep this content in the on-prem-builder repo. If a significant percentage of users use on-prem, we should set those docs up for deployment on docs.chef.io.

One thing we don't want to do is have customers looking for docs content for on-prem in two places. So I'd keep the on-prem content together wherever that is.

PR #8449, opened 2022-02-17
Add poll watcher support to spec watcher for M1

### LINKED ISSUE #8406: Supervisor crashes in Docker studio on M1 macbooks
When a Docker Habitat studio is launched in amd64 compatibility mode, everything generally seems to work but supervisor crashes and can't be run manually.

## Steps to reproduce

```bash
HAB_DOCKER_OPTS="--platform linux/amd64" hab studio enter -D
hab svc status
```

## Observations

Screen Shot 2021-12-31 at 4 47 31 PM

- Binaries all work
- `uname` shows `x86_64` architecture
- Any attempt to run supervisor crashes with:

    ```console
    Notify error: other os error
    [2021-12-31T21:46:41Z ERROR hab_launch] Launcher exiting with code 86
    ```

## Theory

Docker is confirmed to be emulating amd64, and the supervisor binary is linked and executes alright at the surface level, so Supervisor or some dependency must be making some system calls to examine the OS and finding an unexpected value

pozsgaic:

closes #8406

Note that this is only for SpecWatcher - there is more work for PeerWatcher and UserConfigWatcher


pozsgaic:
The commit [e74d41b] is only for testing.  It needs to perform a runtime check.  Please ignore as this will be replaced soon.

components/sup/src/manager/file_watcher.rs
```diff
@@ -1141,56 +1141,33 @@ pub struct FileWatcher<C: Callbacks, W: Watcher> {
...
-    /// initial "file appeared" event even if the watched file existed
-    /// when the file watcher was created.
-    ///
+    /// "file appeared" event if ignore_initial is Some(false) and the
+    /// watched file existed when the file watcher was created.
     /// Will return `Error::NotifyCreateError` if creating the watcher
     /// fails. In case of watching errors, it returns
     /// `Error::NotifyError`.
-    pub fn create_with_no_initial_event<P>(path: P, callbacks: C) -> Result<Self>
+    pub fn create<P>(path: P, callbacks: C, ignore_initial: Option<bool>) -> Result<Self>
```
atrniv, line 1164: Why is this `Option<bool>` when a `bool` would suffice. There are only 2 possible code paths. Option<bool> implies 3 code paths, `Some(true)`, `Some(false)` and `None`.

components/sup/src/manager/sup_watcher.rs
```diff
@@ -0,0 +1,55 @@
...
+          time::Duration};
+
+pub enum SupWatcher {
+    Native(RecommendedWatcher),
+    Fallback(PollWatcher),
+}
+
+impl Watcher for SupWatcher {
+    fn new_raw(tx: Sender<notify::RawEvent>) -> Result<Self> {
+        if let Ok(arch_type) = env::var("HAB_STUDIO_HOST_ARCH") {
```
atrniv, line 21: This might be a more idiomatic way of doing this given we already have platform architecture detection capabilities in the codebase.
```rust
// The habitat_core::package::target contains constants for all of habitat's
// supported platforms.
use habitat_core::package::target::{PackageTarget, AARCH64_DARWIN};
// This will replace the HAB_STUDIO_HOST_ARCH environment variable with the default current arch
// if the current value is invalid
let target = PackageTarget::from_str(env::var("HAB_STUDIO_HOST_ARCH").unwrap_or_default()).unwrap_or(PackageTarget::active_target());
// If the studio host target is AARCH64_DARWIN 
if target == AARCH64_DARWIN {
    // Use poll watcher
} else {
    // Use recommended watcher
}
```

atrniv (commented):
I've mentioned some changes that could ensure the code is more idiomatic

components/hab/src/command/studio/enter.rs
```diff
@@ -19,6 +19,7 @@ use std::{env,
 pub const ARTIFACT_PATH_ENVVAR: &str = "ARTIFACT_PATH";
 pub const CERT_PATH_ENVVAR: &str = "CERT_PATH";
 pub const SSL_CERT_FILE_ENVVAR: &str = "SSL_CERT_FILE";
+pub const STUDIO_HOST_ARCH_ENVVAR: &str = "HAB_STUDIO_SECRET_HAB_STUDIO_HOST_ARCH";
```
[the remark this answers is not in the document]
pozsgaic: In order to pass this variable into studio, it needs to be prepended with HAB_STUDIO_SECRET_
Otherwise it gets dropped.

components/hab/src/command/studio/enter.rs
```diff
@@ -63,6 +64,32 @@ fn cache_ssl_cert_file(cert_file: &str, cert_cache_dir: &Path) -> Result<()> {
     Ok(())
 }
 
+#[cfg(all(target_arch = "aarch64", target_os = "macos"))]
+fn set_arch_env_var() {
+    env::set_var(STUDIO_HOST_ARCH_ENVVAR, "aarch64-macos".to_string());
+    match env::var(STUDIO_HOST_ARCH_ENVVAR) {
```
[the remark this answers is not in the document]
pozsgaic: Yes I will remove that - it is indeed test code.

components/hab/src/command/studio/enter.rs
```diff
@@ -63,6 +64,32 @@ fn cache_ssl_cert_file(cert_file: &str, cert_cache_dir: &Path) -> Result<()> {
...
+        }
+        Err(_) => {
+            debug!("ERROR {} not set", STUDIO_HOST_ARCH_ENVVAR);
+        }
+    }
+}
+
+#[cfg(not(all(target_arch = "aarch64", target_os = "macos")))]
+fn set_arch_env_var() {
+    env::set_var(STUDIO_HOST_ARCH_ENVVAR, "x86_64".to_string());
```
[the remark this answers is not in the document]
pozsgaic: First comment: Yes I agree that using env::set_var without conditionals is preferable.  Will adjust.

Second comment:  I looked at active_target and I ran the cfg test code directly on the Mac M1 and it yields "aarch64-macos".  This does not agree with active_target()

components/sup/src/manager/file_watcher.rs
pozsgaic: I originally was thinking of optional parameters in other languages and associated default behavior.  That does not apply here - will refactor to use bool.

components/sup/src/manager/sup_watcher.rs
pozsgaic: This won't work because the following yields "aarch64-macos", not "aarch64-darwin":
env::set_var(STUDIO_HOST_ARCH_ENVVAR, format!("{}-{}", std::env::consts::ARCH,std::env::consts::OS)
I thought of adding this to habitat_core::package target.rs
    ("aarch64-macos", AARCH64_Macos, AARCH64_MACOS, "aarch64", "macos");
 But I think it would add confusion.

pozsgaic: The lookup fails however with this approach.
I did try adding this line to the target.rs for PackageTarget:
("aarch64-macos", AARCH64_Darwin, AARCH64_DARWIN, "aarch64", "macos");
and rebuilding.
But I would always get back the default x86_64-linux.  
So I left this code as is until we can talk about it.

pozsgaic: When we start supervisor inside of the studio, won't PackageTarget::active_target() resolve to x86_64-linux, not aarch64-darwin?

mwrock: When setting `HAB_STUDIO_SECRET_HAB_STUDIO_HOST_ARCH` you would need to set it to `*PackageTarget::active_target()`. 

Yes, `PackageTarget::active_target()` will always be x86 in the studio however in the above logic it would only be used as a fallback if there is no `HAB_STUDIO_HOST_ARCH` set.

atrniv (approved):

components/sup/src/manager/file_watcher.rs
```diff
@@ -2882,6 +2915,13 @@ mod tests {
         fn prepend_root(&self, p: &Path) -> PathBuf {
             prepend_root_impl(self.root, p, self.debug_info)
         }
+
+        fn is_poll_watcher(&self) -> bool {
+            match &self.watcher_type {
+                WatcherType::NotifyWatcherType => false,
+                WatcherType::PollWatcherType => true,
+            }
```
mwrock, line 2923: do we need `watcher_type`? Could you use `watcher.get_mut_underlying_watcher().real_watcher` instead?

components/sup/src/manager/file_watcher.rs
```diff
@@ -2685,7 +2687,11 @@ mod tests {
                                       });
             if self.parent_is_watched(&pp) {
                 // One event - create.
-                1
+                if !self.is_poll_watcher() {
+                    1
+                } else {
+                    5
+                }
```
mwrock, line 2694: Here and all the other places where you are returning the number of events feels hacky to handle here because the interface is now misleasing. Seems like you are trying to return an arbitrary number of events for the pollwatcher which does not actually match up with any actual event count. Looks like this is all in service of bumping the iteration count later. Seems like things would be more clear and involve less code changes if you consolidated this in `spin_watcher` and have that multiply the iterations for the poll watcher.

mwrock (changes requested):
I'm not sure these tests are getting run in CI. Looking at both of the linux sup builds, all the tests are filtered out according to the output. The windows tests are eunning the unit tests but these watcher tests are all linux only.

components/sup/src/manager/peer_watcher.rs
```diff
@@ -189,4 +190,36 @@ mod tests {
...
+                               ..Default::default() };
+
+        let expected_members = vec![member1, member2];
+        let mut members = watcher.get_members().unwrap();
+        for mut member in &mut members {
+            member.id = String::new();
+        }
+        assert_eq!(expected_members, members);
+        env::remove_var("HAB_STUDIO_HOST_ARCH");
+    }
```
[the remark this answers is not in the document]
pozsgaic: It seems that these tests need to be run in order otherwise the environmental variable will be incorrect.
By default rust will run cargo tests in multiple threads.
If I use the --test-threads=1 switch then the results are as expected.

components/sup/src/manager/sup_watcher.rs
```diff
@@ -0,0 +1,86 @@
...
+        let delay = Duration::from_millis(1000);
+        let _sup_watcher = SupWatcher::new(sender, delay);
+        let watcher_type = match _sup_watcher {
+            Ok(SupWatcher::Native(_sup_watcher)) => "Native",
+            Ok(SupWatcher::Fallback(_sup_watcher)) => "Fallback",
+            _ => "Error",
+        };
+        env::remove_var(TEST_STUDIO_HOST_ARCH_ENVVAR);
+        assert_eq!(watcher_type, "Fallback");
+    }
```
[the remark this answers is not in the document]
pozsgaic: These functions will need to be run sequentially to work.  If we can run tests for this file in a single thread it should work.

components/sup/src/manager/file_watcher.rs
```diff
@@ -2402,15 +2379,40 @@ mod tests {
...
+        runner.debug_info.add(format!("test case: {}", tc.name));
+        runner.run_init_commands(&tc.init.commands);
+        let setup = runner.prepare_watcher(&tc.init.path);
+        runner.run_steps(setup, &tc.init.initial_file, &tc.steps);
+    }
+
     #[test]
     fn file_watcher() {
+        let lock = lock_env_var();
+        lock.set("");
```
mwrock, line 2393: this could just be `lock.unset()` which will clear out the variable.

pozsgaic: OK - there are a number of other instances as well across our test cases, so will fix all of them.

components/sup/src/manager/file_watcher.rs
```diff
@@ -2974,12 +2999,21 @@ mod tests {
...
-            thread::sleep(Duration::from_secs(3));
+            let is_poll_watcher = match &setup.watcher.get_mut_underlying_watcher().real_watcher {
+                SupWatcher::Native(_watcher) => false,
+                SupWatcher::Fallback(_watcher) => true,
+            };
+
+            let mut iterations = expected_event_count;
+            if is_poll_watcher {
+                thread::sleep(Duration::from_secs(15));
+                iterations *= 5;
```
mwrock, line 3013: can you explain why the iteration count needs to be bumped?

mwrock, line 3013: can you explain why the iteration count needs to be bumped?

components/sup/src/manager/file_watcher.rs
```diff
@@ -2974,12 +2999,21 @@ mod tests {
...
-            // for these tests to pass.
-            thread::sleep(Duration::from_secs(3));
+            let is_poll_watcher = match &setup.watcher.get_mut_underlying_watcher().real_watcher {
+                SupWatcher::Native(_watcher) => false,
+                SupWatcher::Fallback(_watcher) => true,
+            };
+
+            let mut iterations = expected_event_count;
+            if is_poll_watcher {
+                thread::sleep(Duration::from_secs(15));
```
mwrock, line 3012: 15 seconds seems really high. Did we play much with smaller delays?

mwrock, line 3012: 15 seconds seems really high. Did we play much with smaller delays?

components/sup/src/manager/file_watcher.rs
```diff
@@ -3038,6 +3072,24 @@ mod tests {
             }
         }
 
+        //  For the poll watcher tehre are more than one Debounced event received
+        //  and this test will fail.  Instead we can look at the last event and
+        //  ensure that it is correct, as it will be the last entry that will
+        //  determine the final state of the watched.
```
mwrock, line 3078: should we have both tests just look at the last one if that is the only one that matters?

mwrock, line 3078: should we have both tests just look at the last one if that is the only one that matters?

mwrock (changes requested):

components/sup/src/manager/peer_watcher.rs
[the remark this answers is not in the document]
pozsgaic: No.  The original was refactored by introducing the function peer_watcher_member_load_test().  These two tests use different values in for the Member constructors.  We could pass in the raw data into the above function as well, but I did not think that it added value or readability.

components/sup/src/manager/file_watcher.rs
pozsgaic: I originally got things working with a 30 second delay and then reduced it to 15 seconds and did not find it to be a problem.  I can try.  Will reduce if it will continue to pass the tests.

pozsgaic: I originally got things working with a 30 second delay and then reduced it to 15 seconds and did not find it to be a problem. I can try. Will reduce if it will continue to pass the tests.

components/sup/src/manager/file_watcher.rs
pozsgaic: Explanation added as a code comment.

pozsgaic: "Through experimentation it was determined that the PollWatcher is less responsive
and emits more events than the NotifyWatcher.  The initial sleep used in NotifyWatcher
was not adequate to pass the tests and was increased as a result.  Also the number of iterations
required is larger for the PollWatcher case as there were intermediate events observed that would lead to
test case failure with the original iteration count used.  The test cases will fail if the desired events are not emitted, so the iteration count was increased to account for the increased number of events in the PollWatcher."

components/sup/src/manager/file_watcher.rs
pozsgaic: I did not want to change the original test case but will give it a try.

pozsgaic: Duplicate question

components/sup/src/manager/file_watcher.rs
pozsgaic: Update - 5 second sleep failed.  10 second sleep failed.  Reverting to 15 second delay.

components/sup/src/manager/file_watcher.rs
```diff
@@ -2402,15 +2379,39 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn polling_file_watcher() {
+        let lock = lock_env_var();
+        lock.set("aarch64-darwin");
+
+        //  When using the PollWatcher variant of SupWatcher, the
+        //  behavior is different than the NotifyWatcher.  The
```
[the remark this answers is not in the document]
pozsgaic: Description updated in the code:
        //  When using the PollWatcher variant of SupWatcher, the
        //  behavior is different than the NotifyWatcher.  The NotifyWatcher
        //  receives event callbacks in response to a file or directory change,
        //  while the PollWatcher must poll and check for changes.
        //  As such, there were observed differences in the timing and number
        //  of events that the PollWatcher will handle.  It was determined
        //  through analyzing the output that the poll watcher as written is not handling
        //  test cases correctly after the second test case.  It does not receive
        //  the required events to pass the test case regardless of timing.

components/sup/src/manager/file_watcher.rs
```diff
@@ -2937,29 +2947,49 @@ mod tests {
...
                     .add(format!("step {}:\n{}", step_idx, dits!(step)));
-                let iterations = self.execute_step_action(&mut setup, &step.action);
-                self.spin_watcher(&mut setup, iterations);
+                let expected_event_count = self.execute_step_action(&mut setup, &step.action);
+                self.spin_watcher(&mut setup, expected_event_count);
+
+                if is_poll_watcher {
+                    thread::sleep(Duration::from_secs(5));
+                };
                 self.test_dirs(&step.dirs, &setup.watcher.paths.dirs);
```
mwrock, line 2966: based on my buildkite and local tests, we can remove this sleep.

mwrock, line 2966: based on my buildkite and local tests, we can remove this sleep.

components/sup/src/manager/file_watcher.rs
mwrock: we should be able to bring this to 5

components/sup/src/manager/file_watcher.rs
mwrock: we should be able to make this 3 iterations

mwrock: we should be able to make this 3 iterations

components/sup/src/manager/file_watcher.rs
pozsgaic: Updated

components/sup/src/manager/file_watcher.rs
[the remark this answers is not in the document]
pozsgaic: Updated

mwrock (approved):

PR #8454, opened 2022-02-22
Implement advanced behavior for init and run hooks

atrniv:
# Summary

The proposed changes introduce additional behaviors in service startup failure scenarios. 
The existing restart behaviors are preserved unless any of the new configuration parameters 
are set. This is to accommodate situations where you do always want an immediate 
restart on failure, e.g.: A crashing NodeJS server that does need to be restarted immediately. 

# Motivation

We don't want to cause infrastructure outages due to a failing service hook that keeps 
calling external services effectively simulating a Denial-of-Service attack. 

# Behavior Changes

When a service restarts due to:

- A package update, it will restart immediately
- A change to templated application config files or hooks due to config application, it will restart immediately
- A stop command followed by a start command from the hab supervisor, it will restart immediately
- An init hook that returns a non-zero exit code, the service will restart with the current backoff period in effect
- A run hook that terminates early regardless of the exit code returned, the service will restart with the current backoff period in effect

The current backoff period is initially set to the `service-min-backoff-period`

Once a service has started up and the `service-restart-cooldown-period` has elapsed the current backoff period will be reset to the `service-min-backoff-period`.

# API Changes

The following fields have been added to the `Service` object in the Supervisor HTTP Gateway API

| Field | Description | 
|--|--|
| last_process_state | Information about the state of the previous process for this service. Only present in case the service has restarted |
| next_restart_at | The time at which the service will start up again, expressed as seconds since epoch. This will be non-null only when a service is down for a restart due to init / run hook failures |
| restart_count | The number of times the service has restarted due to a init / run hook failure. This gets reset to 0 whenever the service is restarted intentionally |
| restart_config | The parameters associated with the service restart behaviour |

The `last_process_state.termination_reason` can have the following values which give more insight into what caused the restart.

| Value | Description |
|--|--|
| package_updated | The service package was updated to a newer version. This resets `restart_count` to 0 |
| init_hook_failed | The init hook failed to run. This does not reset `restart_count` |
| run_hook_failed | The run hook failed to run, usually because the application crashed or a bug in the script. This does not reset the `restart_count` |
| app_config_updated | An application config file was recompiled due to a change in configuration. This resets the `restart_count` |
| run_hook_updated | The run hook was recompiled due to a change in configuration. This resets the `restart_count` |
| post_run_hook_updated | The post-run hook was recompiled due to a change in configuration. This resets the `restart_count` |

In general, any restart which occurs intentionally due to a user action causes the restart count to reset to 0. Eg: Stopping the service, applying config updates, etc.

Unintentional restarts caused due to init / run hook failures do not reset the restart count. This is useful to monitor your service for any crashes due to init / run hook failure since it was started.

The full Service API Schema can be found [here](https://github.com/habitat-sh/habitat/blob/jj/hook-retries/components/sup/doc/http_gateway_services_schema.json)

# Configuration Parameters

These parameters have to be passed to the `hab sup run` command and apply across all services run by the supervisor.

- `service-min-backoff-period`: The minimum period of time to wait before attempting to restart the service. Defaults to 0 secs to preserve existing behaviour.
- `service-max-backoff-period`: The maximum period of time to wait before attempting to restart the service. Defaults to 0 secs to preserve existing behaviour.
- `service-restart-cooldown-period`: The period of time to wait before concluding that a service in no longer failing after a restart has occurred. Defaults to 300 secs.

### Sample sup.toml config parameters

```bash
### The minimum duration in seconds to wait before restarting a service due to a init / run hook failure. 
### 
### The default value if this config is absent is 0 seconds to preserve legacy supervisor behavior. 

service_min_backoff_period = 10 

### The maximum duration in seconds to wait before restarting a service due to init / run hook failure. 
### 
### The default value if this config is absent is 0 seconds to preserve legacy supervisor behavior. 

service_max_backoff_period = 180 

### The duration of time in seconds to wait before resetting the current backoff duration to the 'service_min_backoff_period'. 
### This is important to ensure the supervisor does not mistake a slow start for a successful startup. 
### 
### The default value if this config is absent is 300 seconds 

service_restart_cooldown_period = 300 
```

# FAQs

## How do I configure a fixed backoff period?
If you wish to have a fixed backoff period simply configure the `service-min-backoff-period` and `service-max-backoff-period` to the same value.

### Example

```bash 

# This will cause a service to restart after 10 seconds regardless of the number of times it has failed 
hab sup run --service-min-backoff-period 10 --service-max-backoff-period 10  my-org/my-service 
 
# T = 0, service starts up 
# T = 5, service crashes, will be restarted after 10 secs 
# T = 15, service is restarted 
# T = 18, service crashes, will be restarted after 10 secs 
# T = 28, service is restarted 
``` 

**Note**  If you supply a `service-max-backoff-period` less than the `service-min-backoff-period` then you will also effectively have a fixed backoff period equal to `service-max-backoff-period` regardless of the value of `service-min-backoff-period`. We do not advise you use this approach to achieve a fixed backoff, it is mentioned only for completeness.

## How do I handle services with slow startup times?

If your service is slow to start, it would be advised to configure the `service-restart-cooldown-period` to at least double your expected startup time. This is to prevent the current backoff period from being reset prematurely. 

#### An incorrect configuration

```bash
# Lets say the service takes 30 seconds after starting up to crash
hab sup run --service-min-backoff-period 5 --service-max-backoff-period 20 --service-restart-cooldown-period 10  my-org/my-slow-service 
 
# T = 0, service starts up 
# T = 30, service crashes, will be restarted after 5 secs 
# T = 35, service is restarted
# T = 45, service backoff period is reset to 5 secs because 10 secs has elapsed since the last restart
# T = 65, service crashes again, it will now be restarted after 5 secs due to the backoff period resetting at T=45
# T = 70, service is restarted again
```

#### A correct configuration

```bash
# Lets say the service takes 30 seconds after starting up to crash
hab sup run --service-min-backoff-period 5 --service-max-backoff-period 20 --service-restart-cooldown-period 60  my-org/my-slow-service 
 
# T = 0, service starts up 
# T = 30, service crashes, will be restarted after 5 secs 
# T = 35, service is restarted
# T = 65, service crashes again, it will now be restarted after a random duration between 5 and 20 secs, let's assume 15.
# T = 80, service is restarted again, notice that the backoff period has not been reset.
# T = 140, service backoff period is reset to 5 secs because 60 secs has elapsed since the last restart
```


components/sup/src/manager/service.rs
```diff
@@ -168,10 +169,110 @@ impl TemplateUpdate {
...
+
+    pub fn service(&self) -> Option<&Service> { self.inner.as_ref() }
+
+    pub fn service_mut(&mut self) -> Option<&mut Service> { self.inner.as_mut() }
+
+    pub fn start(&mut self, service: Service) {
+        self.run_state.restart_state = match self.run_state.restart_state {
+            RestartState::None => RestartState::None,
+            RestartState::NeedsRestart => {
+                panic!("Start called on service which was not ready to be restarted");
```
mwrock, line 217: I think we want to make all of these `panic!` instances `error!` so that the text is emitted to stderr but the supervisor does not crash. We try to be super careful in the supervisor code to make sure that we do not panic.

components/core/src/flowcontrol/backoff.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    // Record the completion of attempting an operation
+    pub fn record_attempt(&mut self) -> Option<Duration> {
+        match &self.last_attempt {
+            Some(last_attempt)
+                if Instant::now().duration_since(last_attempt.attempted_at)
+                   >= last_attempt.sleep_duration =>
+            {
+                let distribution = Uniform::new(self.base_backoff,
```
mwrock, line 46: probably not a huge deal but the [docs](https://docs.rs/rand/0.8.5/rand/distributions/uniform/struct.Uniform.html) on `Uniform` suggest using [`Rng::gen_range`](https://docs.rs/rand/0.8.5/rand/trait.Rng.html#method.gen_range) when only using a single sample to avoid the upfront cost of `Uniform`.

components/sup/src/manager/service.rs
```diff
@@ -168,10 +169,110 @@ impl TemplateUpdate {
...
+        && self.run_state
+               .restart_backoff
+               .duration_until_next_attempt()
+               .is_none()
+    }
+
+    pub fn tick(&mut self, census_ring: &CensusRing, launcher: &LauncherCli) -> bool {
+        match &mut self.inner {
+            Some(ref mut service) => {
+                debug!("Starting service tick with persistent state: {:?}",
```
mwrock, line 267: I'd be inclined to make this a `trace!` since it is called so frequently

components/sup/src/manager/service.rs
```diff
@@ -168,10 +169,110 @@ impl TemplateUpdate {
...
+    Restarted,
+}
+
+impl Default for RestartState {
+    fn default() -> Self { RestartState::None }
+}
+
+#[derive(Debug, Default)]
+struct ServiceRunState {
+    ticks:            u64,
```
mwrock, line 197: doesn't seem like we ever use this. Do we need it?

components/core/src/flowcontrol/backoff.rs
atrniv: Thanks, yes that would be a better function to use here.

components/sup/src/manager/service.rs
atrniv: Agreed

components/sup/src/manager/service.rs
atrniv: Yup, I've removed it

sajjaphani:
I do not know whether this proposal is currently accepted or not. I have a couple of comments regarding this approach.

1. Instead of having min and max values for backoff, can we have an initial seed and then have the number of attempts (configurable) to retry with some exponential backoff approach. With each failure, the time waits between attempts increases.
2. What is the significance of 'service-restart-cooldown-period'. Is it possible to determine the service stability through its internal state?

mwrock (changes requested):
This looks really great Johny and I really like the functional tests too! There is really not much to say. However, if I am reading this right, changing the init hook does not trigger a restart here and it should.

.expeditor/scripts/verify/run_cargo_test.sh
```diff
@@ -37,6 +37,11 @@ FS_ROOT=$(mktemp -d /tmp/testing-fs-root-XXXXXX)
 
 export RUST_BACKTRACE=1
 
+# Build the all the hab binaries so that we can run integration tests
+if [[ "$scope" == "sup" ]]; then
+  cargo build
+fi
+
```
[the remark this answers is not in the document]
atrniv: Yes, this basically ensures that if we are running supervisor tests which includes the integration tests, then all the binaries get built prior to that. Just running the integration tests only compiles the integration test binary. The launcher, supervisor and hab cli binaries don't get compiled.

components/sup/tests/integration.rs
```diff
@@ -15,207 +20,630 @@ lazy_static! {
...
+                                           ProcessTerminationReason::PostRunHookUpdated).await?;
+    Ok(())
 }
 
-#[test]
+#[tokio::test]
 #[cfg_attr(feature = "ignore_integration_tests", ignore)]
-fn config_files_change_but_hooks_do_not_still_restarts() {
-    let hab_root = utils::HabRoot::new("config_files_change_but_hooks_do_not_still_restarts");
+async fn no_restart_for_templated_init_hook() -> Result<()> {
```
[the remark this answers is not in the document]
atrniv: I've made this change in the restart behavior

components/sup/tests/integration.rs
```diff
@@ -15,207 +20,630 @@ lazy_static! {
...
+                               &hab_root).await?;
+
+    test_sup.ensure_service_started(package_name, service_group, Duration::from_secs(10))
+            .await?;
+
+    test_sup.service_stop(origin_name, package_name).await?;
+    test_sup.ensure_service_stopped(package_name, service_group, Duration::from_secs(10))
+            .await?;
+
+    test_sup.apply_config(package_name, service_group, applied_config)
```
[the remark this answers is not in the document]
atrniv: Yea, I had assumed that was the case. This test case no longer needs this though, because the init hook template change will cause a restart anyways now.

components/sup/src/manager/service.rs
atrniv: Any invalid usage of the internal restart state would cause a panic. If it did happen, it means we've essentially made a terrible mistake in our code. These code paths are essentially unreachable.

atrniv:
> This looks really great Johny and I really like the functional tests too! There is really not much to say. However, if I am reading this right, changing the init hook does not trigger a restart here and it should.

This was the current behavior in the code base, however I agree that a templated init hook getting changed due to an update should cause a restart in order to keep consistency between the current runtime state of the service and the hook and config state on disk.

atrniv:
> I do not know whether this proposal is currently accepted or not. I have a couple of comments regarding this approach.
> 
>     1. Instead of having min and max values for backoff, can we have an initial seed and then have the number of attempts (configurable) to retry with some exponential backoff approach. With each failure, the time waits between attempts increases.
> 
>     2. What is the significance of 'service-restart-cooldown-period'. Is it possible to determine the service stability through its internal state?

1. Regarding the algorithm we have used, it's called *Decorrelated Jitter* and its based off some research work by AWS. Simple deterministic backoff policies actually don't do very well in scenarios where multiple service instances will be restarting in similar time frames very close to each other due to outage of a common external service. There is a very good write up about it, comparing various backoff algorithms on their [blog](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) .
We do need to cap our maximum backoff period regardless of the strategy we use to determine the next backoff. Otherwise given a sufficiently long outage, you might get backoff waiting periods that could extend for days or months.

2. The service restart cooldown period is a parameter to ensure that a slow startup does not get confused for a successful restart by the supervisor. The danger of getting it wrong is that every restart would be considered a success and would reset the backoff period to the minimum value, effectively nullifying any backoff behaviour. In general a longer cooldown won't cause a problem, but a cooldown that is too short will be a problem.

mwrock (changes requested):
This all looks good Johny. I think one good thing to add here would be another page of documentation that specifically talks about the new backoff behavior. You could add a `.md` file with this info to https://github.com/habitat-sh/habitat/tree/main/components/docs-chef-io/content/habitat. For example https://docs.chef.io/habitat/sup_remote_control/ is built based on https://github.com/habitat-sh/habitat/blob/main/components/docs-chef-io/content/habitat/sup_remote_control.md. So you would create something like sup_backoff.md and just make sure the headers at the top of the page are adjusted accordingly.

mwrock (approved):

mwrock:
Note I moved docs updates to #8490 so we can move this along.

PR #8460, opened 2022-03-10
comma separate multiple test features

mwrock:
This should get butterfly and sup unit tests running on linux. It looks as though ever since we removed the `+toolchain` from `cargo` and started just using the hab package, the cli parsing changed and it was filtering out all tests when using multiple feature filters. Comma separating the features seems to fix this.


mwrock:
I reverted the fix for butterfly. That is going to require `--nightly` which is tricky here. Will address in a separate PR so we can get the sup tests out the door ASAP.

PR #8464, opened 2022-03-16
Hab-sup-run arguments modified.

dikshagupta1:
Raise usage errors if inappropriate hab svc load options are given in hab sup run.

Signed-off-by: dikshagupta1 <diksha.gupta@progress.com>

### DIFF
--- a/components/hab/src/cli/hab/sup.rs
+++ b/components/hab/src/cli/hab/sup.rs
@@ -1,6 +1,8 @@
-use super::{svc::{ConfigOptSharedLoad,
-                  SharedLoad,
-                  DEFAULT_SVC_CONFIG_DIR},
+use super::{svc::{SharedLoad,
+                  CHANNEL_IDENT_DEFAULT,
+                  DEFAULT_SVC_CONFIG_DIR,
+                  GROUP_DEFAULT,
+                  HEALTH_CHECK_INTERVAL_DEFAULT},
             util::{tls::{CertificateChainCli,
                          PrivateKeyCli,
                          RootCertificateStoreCli},
@@ -30,8 +32,14 @@ use habitat_common::{cli::{RING_ENVVAR,
                      FEATURE_FLAGS};
 use habitat_core::{env::Config,
                    fs::HAB_CTL_KEYS_CACHE,
+                   os::process::ShutdownTimeout,
                    package::PackageIdent,
-                   util as core_util};
+                   service::ServiceBind,
+                   util as core_util,
+                   ChannelIdent};
+use habitat_sup_protocol::types::{BindingMode,
+                                  UpdateCondition,
+                                  UpdateStrategy};
 use rants::{error::Error as RantsError,
             Address as NatsAddress};
 use std::{fmt,
@@ -41,6 +49,7 @@ use std::{fmt,
           str::FromStr};
 use structopt::{clap::AppSettings,
                 StructOpt};
+use url::Url;
 
 // All commands relating to the Supervisor (ie commands handled by both the `hab` and `hab-sup`
 // binary)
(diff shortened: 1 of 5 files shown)


mwrock:
I did a little digging and it looks like the reason `hab sup run` with no args complains about the missing `pkg_ident_or_artifact` arg comes back to the default values. Adding `requires = "pkg_ident_or_artifact"` to all fields with no default values works as expected for me. I think the final solution will likely need to put the `SharedLoad` fields into `SupRun` like you have here. You could eliminate the default values from the `structopt` attributes in `SupRun` and then have your `from` implementation handle injecting the defaults if the fields are `None`.

As for handling the issue with the move compile errors, see my comments in the code for how to deal with that.

components/hab/src/command/launcher.rs
```diff
@@ -22,7 +23,8 @@ const LAUNCH_PKG_IDENT: &str = "core/hab-launcher";
 
 pub async fn start(ui: &mut UI, sup_run: SupRun, args: &[OsString]) -> Result<()> {
     init()?;
-    let channel = sup_run.shared_load.channel;
+    //let channel = sup_run.shared_load.channel;
+    let channel = SharedLoad::from(sup_run).channel;
```
mwrock, line 27: send a reference: `SharedLoad::from(&sup_run).channel`

components/sup/src/main.rs
```diff
@@ -241,8 +241,8 @@ async fn split_apart_sup_run(sup_run: SupRun,
                              feature_flags: FeatureFlag)
                              -> Result<(ManagerConfig, Option<sup_proto::ctl::SvcLoad>)> {
     let ring_key = get_ring_key(&sup_run)?;
-
-    let shared_load = sup_run.shared_load;
+    //let shared_load = sup_run.shared_load;
+    let shared_load = hab::cli::hab::svc::SharedLoad::from(sup_run);
```
mwrock, line 245: send a reference: `SharedLoad::from(&sup_run)`

components/hab/src/cli/hab/sup.rs
```diff
@@ -322,11 +330,111 @@ pub struct SupRun {
...
+    #[serde(skip)]
+    pub environment:           Vec<String>,
+    /// Use the package config from this path rather than the package itself
+    #[structopt(long = "config-from", requires = "pkg_ident_or_artifact")]
+    pub config_from:           Option<PathBuf>,
 }
 
+impl From<SupRun> for SharedLoad {
+    fn from(sup_run: SupRun) -> Self {
+      Self {
```
mwrock, line 411: The definition can be:

```
impl From<&SupRun> for SharedLoad {
    fn from(sup_run: &SupRun) -> Self {
```

components/hab/src/cli/hab/sup.rs
```diff
@@ -322,11 +330,111 @@ pub struct SupRun {
...
+        bind: sup_run.bind,
+        update_condition: sup_run.update_condition,
+        binding_mode: sup_run.binding_mode,
+        health_check_interval: sup_run.health_check_interval,
+        shutdown_timeout: sup_run.shutdown_timeout,
+        #[cfg(target_os = "windows")]
+        password: sup_run.password,
+        environment: sup_run.environment,
+        application: sup_run.application,
+        config_from: sup_run.config_from,
```
mwrock, line 426: Use `clone()` on these fields: `config_from: sup_run.config_from.clone()`

components/hab/src/cli/hab/sup.rs
```diff
@@ -322,9 +330,93 @@ pub struct SupRun {
...
+               group: sup_run.group
+                             .as_ref()
+                             .unwrap_or(&GROUP_DEFAULT::get())
+                             .to_string(),
+               topology: sup_run.topology,
+               strategy: sup_run.strategy.unwrap_or(UpdateStrategy::None),
+               bind: sup_run.bind.clone(),
+               update_condition: sup_run.update_condition.unwrap_or(UpdateCondition::Latest),
+               binding_mode: sup_run.binding_mode.unwrap_or(BindingMode::Strict),
+               health_check_interval: sup_run.health_check_interval.unwrap_or(30),
```
mwrock, line 412: We should be calling `health_check_interval_default` instead of the `30`.

components/hab/src/command/launcher.rs
```diff
@@ -22,7 +23,7 @@ const LAUNCH_PKG_IDENT: &str = "core/hab-launcher";
 
 pub async fn start(ui: &mut UI, sup_run: SupRun, args: &[OsString]) -> Result<()> {
     init()?;
-    let channel = sup_run.shared_load.channel;
+    let channel = SharedLoad::from(&sup_run).channel;
```
mwrock, line 26: I think this could just be `sup_run.channel` now and no need to transform to a `SharedLoad`.

mwrock (changes requested):

components/hab/src/cli/hab/svc.rs
```diff
@@ -30,6 +30,8 @@ use walkdir::WalkDir;
 
 const DEFAULT_SVC_CONFIG_FILE: &str = "/hab/sup/default/config/svc.toml";
 pub const DEFAULT_SVC_CONFIG_DIR: &str = "/hab/sup/default/config/svc";
+pub const HEALTH_CHECK_INTERVAL_DEFAULT: u64 = 30;
+const HEALTH_CHECK_INTERVAL_STR_DEFAULT: &str = "30";
```
mwrock, line 34: I don't think this is necessary. See comment below.

components/hab/src/cli/hab/svc.rs
```diff
@@ -201,7 +203,7 @@ pub struct SharedLoad {
     // serialization format. We want to allow the user to simply specify a `u64` to be consistent
     // with the CLI, but we cannot change the serialization because the spec file depends on the map
     // based format.
-    #[structopt(long = "health-check-interval", short = "i", default_value = "30")]
+    #[structopt(long = "health-check-interval", short = "i", default_value = HEALTH_CHECK_INTERVAL_STR_DEFAULT)]
     #[serde(default = "health_check_interval_default")]
```
mwrock, line 207: could be : `default_value = &HEALTH_CHECK_INTERVAL_DEFAULT.to_string()`

mwrock (changes requested):

mwrock (approved):

PR #8490, opened 2022-04-19
Docs updates reflecting hook retry/backoff changes

mwrock:
This is documentation updates for #8454

### DIFF
--- a/components/docs-chef-io/content/habitat/service_restarts.md
+++ b/components/docs-chef-io/content/habitat/service_restarts.md
@@ -1,6 +1,5 @@
 +++
 title = "Service Restarts"
-description = "When does the supervisor restart a service"
 gh_repo = "habitat"
 
 [menu]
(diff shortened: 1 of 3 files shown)


components/docs-chef-io/content/habitat/service_restarts.md
```diff
@@ -0,0 +1,75 @@
...
+gh_repo = "habitat"
+
+[menu]
+  [menu.habitat]
+    title = "Service Restarts"
+    identifier = "habitat/reference/service-restarts Service Restarts"
+    parent = "habitat/reference"
++++
+
+Since the 1.6.477 Supervisor release, it is possible to configure a service restart backoff period to control how quickly a service is restarted following the
```
IanMadd, line 13: Is this Habitat 1.6.477?

components/docs-chef-io/content/habitat/service_restarts.md
```diff
@@ -0,0 +1,75 @@
...
+failure of an init or run hook. Before this, if there was a failure of the init / run hook the supervisor would attempt to restart the service immediately,
+potentially leading to excessive load on external APIs similar to a Denial of Service attack.
+
+## Overview
+
+The new restart backoff behavior can be enabled by explicitly passing the `service-min-backoff-period`, `service-max-backoff-period` and 
+`service-restart-cooldown-period` parameters to the `sup run` command. These can also be configured in the [supervisor configuration file]({{< relref "sup_config" >}}) as `service_min_backoff_period`, `service_max_backoff_period` and `service_restart_cooldown_period`.
+
+```bash
+$ hab sup run --service-min-backoff-period 5 --service-max-backoff-period 20 --service-restart-cooldown-period 60  core/redis
```
IanMadd, line 23: Are the parameters specifying seconds?

components/docs-chef-io/content/habitat/sup_config.md
```diff
@@ -175,6 +175,22 @@ strategy = "at-once"
...
+### The default value if this config is absent is 0 seconds to preserve legacy supervisor behavior. 
+service_min_backoff_period = 10 
+
+### The maximum duration in seconds to wait before restarting a service due to init / run hook failure. 
+### 
+### The default value if this config is absent is 0 seconds to preserve legacy supervisor behavior. 
+service_max_backoff_period = 180 
+
+### The duration of time in seconds to wait before resetting the current backoff duration to the 'service_min_backoff_period'. 
+### This is important to ensure the supervisor does not mistake a slow start for a successful startup.
```
IanMadd, line 189: Should this be more like: 
```suggestion
### This is important to ensure the supervisor does not mistake a slow start with a service failure. 
```

components/docs-chef-io/content/habitat/service_restarts.md
mwrock: Yeah that is actually the last release (the latest one as of right now). Maybe we should say since 1.6.491 which is the latest version in acceptance and it has this feature.

components/docs-chef-io/content/habitat/sup_config.md
```diff
@@ -175,21 +175,21 @@ strategy = "at-once"
...
+### The default value if this config is absent is 0 seconds to preserve legacy supervisor behavior.
+service_min_backoff_period = 10
+
+### The maximum duration in seconds to wait before restarting a service due to init / run hook failure.
+###
+### The default value if this config is absent is 0 seconds to preserve legacy supervisor behavior.
+service_max_backoff_period = 180
+
+### The duration of time in seconds to wait before resetting the current backoff duration to the 'service_min_backoff_period'.
+### This is important to ensure the supervisor does not mistake a slow start with a service failure.
```
atrniv, line 189: The service considered successfully started once the cooldown has elapsed. This may be a bit confusing but I'll try to list the various cases to make it clear.
- Service Takes 30s to start
  - Cooldown is 15s
    - Service considered success after 15s even though it hasn't started up and could potentially fail later in the startup process. Here a slow start is confused with a successful startup.
  - Cooldown is 60s
    - Service considered success correctly after 60s. Here any failure before 60s is correctly identified as failure, and we do not reset the backoff period.

Namely, there is no scenario where we mistake a slow start for a service failure. We always identify failures correctly. But we mistakenly assume success for cases which are actually failures. 

We need correct identification of success in order to reset the backoff period and nothing else.

Notice that even if the cooldown was 600 secs it would only prevent the backoff from resetting, potentially having the service restarted continuously between the min and the capped max backoff period. All failures would still be detected correctly and cause some backoff to be applied.

atrniv (changes requested):
There's a minor error in the documentation of the restart cooldown period.

atrniv (approved):

atrniv:
@IanMadd , the additional changes makes the use of the cooldown period easier to understand. Thanks !

PR #8559, opened 2022-07-06
SockerAddressProxy type support

dikshagupta1:
This merge fixes the bug - HABC-115 which points towards the parsing issues while using a config.toml file in the cab sup run --config-files option.

### DIFF
--- a/components/hab/src/cli/hab/sup.rs
+++ b/components/hab/src/cli/hab/sup.rs
@@ -9,9 +9,9 @@ use super::{svc::{ConfigOptSharedLoad,
                    ConfigOptRemoteSup,
                    DurationProxy,
                    RemoteSup,
+                   SocketAddrProxy,
                    SubjectAlternativeName}};
-use crate::{error::Error,
-            VERSION};
+use crate::VERSION;
 use configopt::{self,
                 configopt_fields,
                 ConfigOpt};
@@ -37,8 +37,7 @@ use rants::{error::Error as RantsError,
 use serde::{Deserialize,
             Serialize};
 use std::{fmt,
-          net::{IpAddr,
-                SocketAddr},
+          net::IpAddr,
           path::PathBuf,
           str::FromStr};
 use structopt::{clap::AppSettings,
@@ -129,10 +128,6 @@ impl From<EventStreamAddress> for NatsAddress {
     fn from(address: EventStreamAddress) -> Self { address.0 }
 }
 
-fn parse_peer(s: &str) -> Result<SocketAddr, Error> {
-    Ok(habitat_common::util::resolve_socket_addr_with_default_port(s, GossipListenAddr::DEFAULT_PORT)?.1)
-}
-
 /// Run the Habitat Supervisor
 #[configopt_fields]
 #[derive(ConfigOpt, StructOpt, Deserialize)]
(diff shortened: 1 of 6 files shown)


components/hab/src/cli/hab/sup.rs
```diff
@@ -190,8 +185,8 @@ pub struct SupRun {
     // TODO (DM): Currently there is not a good way to use `parse_peer` when deserializing. Due to
     // https://github.com/serde-rs/serde/issues/723. There are workarounds but they are all ugly.
     // This means that you have to specify the port when setting this with a config file.
```
mwrock, line 187: we should remove this comment

components/hab/src/cli/hab/util.rs
```diff
@@ -232,6 +235,48 @@ impl fmt::Display for DurationProxy {
...
+#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "String")]
+pub struct SocketAddrProxy (SocketAddr);
+
+impl TryFrom<String> for SocketAddrProxy {
+    type Error = Error;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let (domain, addr) =
+            habitat_common::util::resolve_socket_addr_with_default_port(value, GossipListenAddr::DEFAULT_PORT)?;
+        if !domain.is_empty() {
```
mwrock, line 248: I was looking at `resolve_socket_addr_with_default_port` and couldn't figure out under what condition `domain` would be empty.

components/sup/src/main.rs
```diff
@@ -1249,6 +1246,105 @@ sys_ip_address = "7.8.9.0"
...
+                                                                    key_path,
+                                                                    ca_cert_path:
+                                                                        Some(ca_cert_path) }),
+                                       feature_flags: FeatureFlag::empty(),
+                                       event_stream_config: None,
+                                       keep_latest_packages: Some(5),
+                                       sys_ip: "7.8.9.0".parse().unwrap() },
+                       config);
+        }
+
```
mwrock, line 1347: I think you are really only interested in testing the peer values and not all of the other values. We already have a test that tests the other stuff and this test can then be MUCH smaller because you won't need all the cert key stuff or alot of other ceremony.

mwrock (changes requested):

components/hab/src/cli/hab/sup.rs
dikshagupta1: Addressed in commit 8930f52d98b666253fb494c288a5c2cdb3bf686c.

components/hab/src/cli/hab/util.rs
dikshagupta1: Agree. Have removed the unnecessary if check in commit 8930f52d98b666253fb494c288a5c2cdb3bf686c.

components/sup/src/main.rs
dikshagupta1: Have modified the test case to specifically target at peer values.
Please refer  commit 8930f52d98b666253fb494c288a5c2cdb3bf686c

mwrock (changes requested):

components/sup/src/main.rs
```diff
@@ -289,7 +285,7 @@ async fn split_apart_sup_run(sup_run: SupRun,
                         organization: sup_run.organization,
                         gossip_permanent: sup_run.permanent_peer,
                         ring_key,
-                        gossip_peers: sup_run.peer,
+                        gossip_peers: sup_run.peer.iter().map(|x| x.into()).collect(),
```
atrniv, line 288: You could get rid of the clippy lint by using this form:

```rust
gossip_peers: sup_run.peer.iter().map(Into::into).collect()
```

It basically uses the function `into` of the `Into<T>` trait as the mapping function itself. In this case the T would be automatically inferred by the compiler due to the assignment to gossip_peers. If it isn't able to infer it you would have to use a slightly modified form:



```rust
gossip_peers: sup_run.peer.iter().map(Into::<SocketAddr>::into).collect()
```

atrniv (changes requested):

.expeditor/scripts/verify/shellcheck.sh
```diff
@@ -24,6 +24,7 @@ shellcheck --version
 # https://github.com/koalaman/shellcheck/wiki/SC1117
 # https://github.com/koalaman/shellcheck/wiki/SC2148
 # https://github.com/koalaman/shellcheck/wiki/SC2034
+echo "start"
```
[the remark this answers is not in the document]
dikshagupta1: Yes, they were for debugging. Thankyou. I have removed them.

components/sup/src/main.rs
dikshagupta1: It was able to infer by just using (Into::into). I have re-confirmed the same by necessary debugging logs.

mwrock (approved):

atrniv (approved):

PR #8564, opened 2022-07-18
Fixes several Rolling Update bugs

### LINKED ISSUE #8441: Channel tracking rolling updates hang if a package is demoted from the tracked chennel before the rolling update completes
This was observed by a customer where a service was set to update and track a specific channel. In this case there were 4 supervisors involved in the rolling update:

* sup1 - leader - No problems
* sup2 - first follower - No problems
* sup3 - next follower in line. The package to be updated was demoted while this sup was waiting for sup2 to finish. When it bacame this sup's turn to update, the package was no longer in the channel and downloading resulted in a 404. It continued to retry this indefinitely. The 404 was coming from [here](https://github.com/habitat-sh/habitat/blob/5840db1492e64972535cbbb144bd1f5af61ba326/components/sup/src/manager/service_updater/package_update_worker.rs#L119).
* sup4 - while waiting for sup3 (which would be forever) as soon as the package to be updated was demoted and the leader started a new update, sup4 thought the leader [had died](https://github.com/habitat-sh/habitat/blob/5840db1492e64972535cbbb144bd1f5af61ba326/components/sup/src/manager/service_updater/rolling_update_worker.rs#L252).

mwrock:
Fixes #8441 

This fixes the following bugs with rolling updates:

* There is a race condition that I can only replicate on Windows where a supervisor is updated, it gossips the new service version, it starts a new rolling update worker, and then updates it census state. It is possible (and is in fact happening reliably on windows) that after starting the rolling update thread, that thread will read and act upon stale census data and indefinitely wait for a new update. This fix registers the new updater workers after the census is repopulated.
* There is a delay that should be waited for before checking for updates. In the case of rolling updates, the leader checks to see if there is an update and then the followers all update to that version. The leader should honor the delay before each update check, but the followers should update asap after one another. There was some refactoring here a couple years ago that causes each follower to incur the delay that potentially causes a rolling update to take much longer than it should. This also opens the possibility for other undesirable edge cases making it easier for a roll back to happen in the middlw of an update or a update leaser to be removed during an update. If the followers do not wait, there is a much smaller liklihood that these events would occur in the middle of an update. This PR still fixes those cases, but by removing the delay for followers, those cases would heve been much harder to reproduce.
* Rolling update followers that track the head of a channel could get into a state where they are indefinitely receiving 404s from builder if an update has been rolled back before a rolling update has completed. Although the followers have been told the exact version to update to, they are still checking the channel to see what is at the head. If their version has been demoted, that version will not be there. In the case where rolling followers are updating to a fully qualified version, we should just update them to that version without checking the channel. That is the leader's responsibility.
* In the case where an update has been rolled back before an update completes, we ideally want to stop followers from updating that have not yet updated. Once we see that they are already on the leader's version, we should just cancel their update.
* In the case where a leader dies mid update, a new follower will be elected and promoted to leader. That leader may initially have a stale version and we don't want to initiate a rollback for followers that have already updated.

### DIFF
--- a/components/sup/src/manager/service_updater/rolling_update_worker.rs
+++ b/components/sup/src/manager/service_updater/rolling_update_worker.rs
@@ -1,11 +1,11 @@
-use super::package_update_worker::PackageUpdateWorker;
+use super::{package_update_worker::PackageUpdateWorker,
+            IncarnatedPackageIdent};
 use crate::{census::{CensusGroup,
                      CensusRing},
             manager::service::{Service,
                                Topology}};
 use habitat_common::owning_refs::RwLockReadGuardRef;
-use habitat_core::{package::PackageIdent,
-                   service::ServiceGroup};
+use habitat_core::service::ServiceGroup;
 use log::{debug,
           error,
           trace,
@@ -33,7 +33,7 @@ enum FollowerUpdateStartEvent {
     PromotedToLeader,
     /// An update started and we have a specific package to update to.
     // TODO (DM): This should use FullyQualifiedPackageIdent.
-    UpdateTo(PackageIdent),
+    UpdateTo(IncarnatedPackageIdent),
 }
 
 /// Possible events when a follower is waiting for its turn to update.
@@ -44,10 +44,10 @@ enum FollowerUpdateTurnEvent {
     /// leader. When this happens the new leader needs to update to the exact version the
     /// rolling update was started with.
     // TODO (DM): This should use FullyQualifiedPackageIdent.
-    PromotedToLeaderMidUpdate(PackageIdent),
+    PromotedToLeaderMidUpdate(IncarnatedPackageIdent),
     /// An update started and we have a specific package to update to.
     // TODO (DM): This should use FullyQualifiedPackageIdent.
-    UpdateTo(PackageIdent),
+    UpdateTo(IncarnatedPackageIdent),
 }
 
 /// The worker for handling rolling updates.
@@ -75,7 +75,7 @@ impl RollingUpdateWorker {
                butterfly }
     }
 
-    pub async fn run(self) -> PackageIdent {
+    pub async fn run(self) -> IncarnatedPackageIdent {
         // Determine this services suitablity and start the update leader election
         let suitability = self.update_election_suitability(self.topology).await;
         self.butterfly
@@ -85,11 +85,15 @@ impl RollingUpdateWorker {
             Role::Leader => {
                 // Wait for an update which will trigger follower updates through the census
                 // protocol
-                self.package_update_worker.update().await
+                let mut pkg = self.package_update_worker.update().await;
+                // bump the incarnation of the update that the leader performed
+                // this will eventually get gossiped after the service restarts
+                pkg.incarnation = Some(self.census_group().await.pkg_incarnation + 1);
+                pkg
             }
             Role::Follower => {
-                // Wait till it is our turn to update. It is possible that while we are waiting the
-                // leader dies and we are promoted to update leader.
+                // Wait till it is our turn to update. It is possible that while we are waiting
+                // the leader dies and we are promoted to update leader.
                 match self.follower_wait_for_update_turn().await {
                     FollowerUpdateTurnEvent::PromotedToLeader => {
                         // Wait for an update which will trigger follower updates through the
@@ -97,8 +101,8 @@ impl RollingUpdateWorker {
                         self.package_update_worker.update().await
                     }
                     FollowerUpdateTurnEvent::PromotedToLeaderMidUpdate(new_ident) => {
-                        // Update to the same package as the old leader allowing all followers to
-                        // finish updating
+                        // Update to the same package as the old leader allowing all followers
+                        // to finish updating
                         self.package_update_worker.update_to(new_ident).await
                     }
                     FollowerUpdateTurnEvent::UpdateTo(new_ident) => {
@@ -207,15 +211,35 @@ impl RollingUpdateWorker {
                                    self.service_group);
                             break FollowerUpdateStartEvent::PromotedToLeader;
                         }
-                        if leader.pkg != me.pkg {
+
+                        if leader.pkg_incarnation != census_group.pkg_incarnation {
+                            trace!("leader incarnation {} is not caught up with cencus group \
+                                    incarnation {}",
+                                   leader.pkg_incarnation,
+                                   census_group.pkg_incarnation);
+                        }
+
+                        if leader.pkg_incarnation > me.pkg_incarnation {
                             // The leader has a new package starting a rolling update
-                            debug!("'{}' started a rolling update: leader='{}' follower='{}'",
-                                   self.service_group, leader.pkg, me.pkg);
-                            break FollowerUpdateStartEvent::UpdateTo(leader.pkg.clone());
+                            debug!("'{}' started a rolling update: leader='{}/{}' \
+                                    follower='{}/{}'",
+                                   self.service_group,
+                                   leader.pkg_incarnation,
+                                   leader.pkg,
+                                   me.pkg_incarnation,
+                                   me.pkg);
+                            break FollowerUpdateStartEvent::UpdateTo(IncarnatedPackageIdent::new(leader.pkg.clone(),
+                                                                     Some(leader.pkg_incarnation)));
                         } else {
                             // The leader still has the same package as this follower so an update
                             // has not started
-                            trace!("'{}' is not in a rolling update", self.service_group);
+                            trace!("'{}' is not in a rolling update: leader='{}/{}' \
+                                    follower='{}/{}'",
+                                   self.service_group,
+                                   leader.pkg_incarnation,
+                                   leader.pkg,
+                                   me.pkg_incarnation,
+                                   me.pkg);
                         }
                     }
                     _ => {
@@ -235,7 +259,7 @@ impl RollingUpdateWorker {
             FollowerUpdateStartEvent::PromotedToLeader => {
                 return FollowerUpdateTurnEvent::PromotedToLeader
             }
-            FollowerUpdateStartEvent::UpdateTo(ident) => ident,
+            FollowerUpdateStartEvent::UpdateTo(ident) => (ident),
         };
         loop {
             {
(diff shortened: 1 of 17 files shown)


mwrock:
Goona keep this a draft while I add tests

mwrock:
For testing this fix, provision a windows VM with docker if you do not already have one. You can use the following `docker-compose.yml` file:

```
version: '3'
services:
  first:
    image: mwrock/nginx
    environment:
      # - RUST_LOG=warn,habitat_sup::manager::service_updater=trace
      - HAB_LICENSE=accept-no-persist
    command: --strategy rolling --channel rolling --update-condition track-channel
  next:
    image: mwrock/nginx
    environment:
      # - RUST_LOG=warn,habitat_sup::manager::service_updater=trace
      - HAB_LICENSE=accept-no-persist
    links:
      - first
    depends_on:
      - first
    command: --peer first --strategy rolling --channel rolling --update-condition track-channel
```

Note that you can uncomment the RUST_LOG setting if you want to see much more verbose logging of the updater, but you should not need to to validate this.

Replicating buggy behavior:

- Export mwrock/nginx to a container using the latest stable hab: `hab pkg export container mwrock/nginx --channel rolling`
- Bring up a 4 node ring: `docker-compose up --scale next=3`
- Promote to `mwrock/nginx/1.21.6/20220705160727`: `hab pkg promote mwrock/nginx/1.21.6/20220705160727 rolling`
- This should eventually work however notice how each node takes up to a minute to update themselves after the previous node
- Promote `mwrock/nginx/1.21.6/20220705160747`: `hab pkg promote mwrock/nginx/1.21.6/20220705160747 rolling`
- You should see the leader update itself but most and likely all of the followers will be stuck and will not update

Reviewing the fix:
- Make sure to demote the two versions of mwrock/nginx that you promoted above
- Export using built versions of this PR: `hab pkg export container mwrock/nginx --channel rolling --sup-pkg core/hab-sup/1.6.539/20220802124011 --hab-pkg core/hab/1.6.539/20220802120851`
- Bring up a 6 node ring: `docker-compose up --scale next=5`
- Promote to `mwrock/nginx/1.21.6/20220705160727`: `hab pkg promote mwrock/nginx/1.21.6/20220705160727 rolling`
- This should work and after the leader updates, the followers should update right after each other without delay
- Promote `mwrock/nginx/1.21.6/20220705160747`: `hab pkg promote mwrock/nginx/1.21.6/20220705160747 rolling`
- This should update all nodes just like the first promote
- Demote both of the promoted packages again
- Promote to `mwrock/nginx/1.21.6/20220705160727`: `hab pkg promote mwrock/nginx/1.21.6/20220705160727 rolling`
- Demote `mwrock/nginx/1.21.6/20220705160727` after a couple nodes update: `hab pkg demote mwrock/nginx/1.21.6/20220705160727 rolling`
- The rest of the nodes will likely continue to update and then you should see all the nodes roll back.

components/sup/src/manager/service_updater/rolling_update_worker.rs
```diff
@@ -207,15 +256,20 @@ impl RollingUpdateWorker {
...
                             break FollowerUpdateStartEvent::UpdateTo(leader.pkg.clone());
                         } else {
                             // The leader still has the same package as this follower so an update
                             // has not started
-                            trace!("'{}' is not in a rolling update", self.service_group);
+                            trace!("'{}' is not in a rolling update: leader='{}' follower='{}'",
+                                   self.service_group,
+                                   leader.pkg,
+                                   me.pkg);
                         }
```
atrniv, line 273: @mwrock , here's a scenario you may want to consider:

- Given 3 nodes A, B, C. A is the leader. B & C are followers.
- A, B & C currently have package version v1
- A updates the package to version v2
- The following happens concurrently:
  - A dies
  - package version v2 gets demoted 
- A new package version v3 is promoted  
- B becomes the new leader and updates to package version v3.

Now in C:
- The older leader is A, which had updated to package version v2
- The new leader is B, which has a package v3 which is not equal to v2

Thus C will always think that B is not ok to lead, since it's package version v3 does not match the version (v2) of the old leader (A). Thus C might always be stuck in this loop


## Using Lamport Clocks to track version change ordering

A possible solution to this problem, and also perhaps a general solution to the problem of rolling back updates across followers could be achieved using [Lamport clocks](https://martinfowler.com/articles/patterns-of-distributed-systems/lamport-clock.html) as follows:

- Given 3 nodes A, B, C. A is the leader. B & C are followers.
- A, B & C currently have package version 0/v1 . Here `0` serves as a lamport clock to track the order of version changes.
- A updated the package to 1/v2
- The following happens concurrently:
  - A dies
  - Package version v2 gets demoted
- A new package version v3 is promoted
- B becomes the new leader and updates the package version to 2/v3. Note that B has incremented the lamport clock from 1 to 2. 

Now in C:
- The older leader is A, which had updated to package version v2
- The new leader is B, which has package 2/v3 which is after 1/v2. Hence C will now update to the latest version which happens to be 2/v3.

The `ok_to_lead` function could be changed to check that the new leader's package version's lamport clock is newer than the old leader's version's lamport clock.

## Scenario of Package version Rollback

- Given 3 nodes A, B, C. A is the leader. B & C are followers.
- A, B & C currently have package version 0/v1.
- A updates to 1/v2.
- B updates to 1/v2 because it's previous peer (A) has updated to the leader's version (1/v2).
- Package version v2 gets demoted.
- A detects that  v1 is now the latest version and updates it's version to 2/v1.
- B & C receive the latest version (2/v1) of the package from the leader.
- C does not update, because it's previous peer (B) has version 1/v2, which is not equal to the leader's version 2/v1.
- B updates to 2/v1 because it's previous peer (A) has updated to the leader's version (2/v1). This is effectively a rollback. 
- C updates to 2/v1 because it's previous peer (B) has updated to the leader's (A) version (2/v1). C effectively does nothing in this case, as it transitions from 0/v1 -> 2/v2.

The use of a Lamport clock to track package update version changes removes all assumptions about how the package's version should progress. It could go forward or backward by any number of steps and still behave correctly.

components/sup/src/manager/service_updater/rolling_update_worker.rs
```diff
@@ -207,15 +256,20 @@ impl RollingUpdateWorker {
                                    self.service_group);
                             break FollowerUpdateStartEvent::PromotedToLeader;
                         }
-                        if leader.pkg != me.pkg {
+
+                        rolling_leader_state.update(leader);
+                        if rolling_leader_state.ok_to_lead() && leader.pkg != me.pkg {
```
atrniv, line 261: @mwrock , here's a scenario you may want to consider:

- Given 3 nodes A, B, C. A is the leader. B & C are followers.
- A, B & C currently have package version v1
- A updates the package to version v2
- The following happens concurrently:
  - A dies
  - package version v2 gets demoted 
- A new package version v3 is promoted  
- B becomes the new leader and updates to package version v3.

Now in C:
- The older leader is A, which had updated to package version v2
- The new leader is B, which has a package v3 which is not equal to v2

Thus C will always think that B is not ok to lead, since it's package version v3 does not match the version (v2) of the old leader (A). Thus C might always be stuck in this loop


## Using Lamport Clocks to track version change ordering

A possible solution to this problem, and also perhaps a general solution to the problem of rolling back updates across followers could be achieved using [Lamport clocks](https://martinfowler.com/articles/patterns-of-distributed-systems/lamport-clock.html) as follows:

- Given 3 nodes A, B, C. A is the leader. B & C are followers.
- A, B & C currently have package version 0/v1 . Here `0` serves as a lamport clock to track the order of version changes.
- A updated the package to 1/v2
- The following happens concurrently:
  - A dies
  - Package version v2 gets demoted
- A new package version v3 is promoted
- B becomes the new leader and updates the package version to 2/v3. Note that B has incremented the lamport clock from 1 to 2. 

Now in C:
- The older leader is A, which had updated to package version v2
- The new leader is B, which has package 2/v3 which is after 1/v2. Hence C will now update to the latest version which happens to be 2/v3.

The `ok_to_lead` function could be changed to check that the new leader's package version's lamport clock is newer than the old leader's version's lamport clock.

## Scenario of Package version Rollback

- Given 3 nodes A, B, C. A is the leader. B & C are followers.
- A, B & C currently have package version 0/v1.
- A updates to 1/v2.
- B updates to 1/v2 because it's previous peer (A) has updated to the leader's version (1/v2).
- Package version v2 gets demoted.
- A detects that  v1 is now the latest version and updates it's version to 2/v1.
- B & C receive the latest version (2/v1) of the package from the leader.
- C does not update, because it's previous peer (B) has version 1/v2, which is not equal to the leader's version 2/v1.
- B updates to 2/v1 because it's previous peer (A) has updated to the leader's version (2/v1). This is effectively a rollback. 
- C updates to 2/v1 because it's previous peer (B) has updated to the leader's (A) version (2/v1). C effectively does nothing in this case, as it transitions from 0/v1 -> 2/v2.

The use of a Lamport clock to track package update version changes removes all assumptions about how the package's version should progress. It could go forward or backward by any number of steps and still behave correctly.

mwrock:
Actually the above hang scenario should not happen. If an update (or rollback) is in progress and a leader dies, the new leader once chosen will do one of 2 things:

1. If the new leader was a follower who had not yet updated, it will immediately update itself to the version of the former leader. The new leader will not query builder for the "latest" version until the next check period (60 seconds by default). So in this case B would update to v2 even though v3 is the new version. Then it would wait 60 seconds or whatever is configured before checking again. In that time C should deem B `ok_to_lead`.
2. If the new leader was a follower who had already updated, it will immediately pause for the check period before querying builder. So in the case where A dies after B updates but before C...well...in that case C would already be in the middle of an update. However, lets add D all the way to Z just to add some time for this scenario. If A dies after B is updated but before M thru Z, then B will wait 60 (or whatever is configured) seconds before seeing v3 as the latest and in the mean time will have a version that was equal to the previous leader and thus be `ok_to_lead`.

Now I suppose there could possibly be some network latency that might delay the rumors and someone has configured the update period to a smaller period. This is really unlikely since most who configure the delay period configure to a larger value. What we could do here id add a timeout to the `RollingLeaderState` of (DELAY * 2) and render the new leader `ok_to_lead` if still not ok after that timeout. In that case we remove the chance if a follower ever getting stuck.

I totally agree that the Lamport clock is a sound approach. I'm just wondering if the census refactoring is worth it given the very small chance for this to occur where you lose a follower (fairly rare), in the middle of an update (more rare) that gets updated again before the update competes (more rare still) and the network latency is high enough that rumors take longer that `DELAY` to reach the nodes (super rare).

atrniv:
@mwrock, in my opinion these are some reasons we should favor a correct implementation vs one that usually works:

1. Following the nuances of the timing and the reasons for all the various delays for a future maintainer will be far more complicated compared to a sound Lamport clock implementation which will guarantee correctness (conditional on the correctness of the gossip protocol). Not to mention actually reproducing the cause of failure in a production cluster will require a lot of effort because it is implicitly depends on the timing of the test cases, the clock speeds on each host and the reliability of the network. We actually face this problem right now while working on this issue. The complex success and failure scenarios we have considered above are testament to this and Lamport clocks were invented to solve exactly these kinds of complex distributed system scenarios.

2. Sometimes certain customer configurations or usage patterns will also guarantee that this failure occurs consistently. In my own experience given enough scale anything that can break, will break in production no matter how low the probability of it occurring. The only solution to this is to have an implementation that assumes the bare minimum necessary in order to work.

3. Any user of Habitat would assume that we have taken care of all the fiddly edge cases and hard problems. Even one failure of this behavior in some critical production cluster may cause a large enough business issue that they loose trust in the implementation of this feature.

Regarding the cost of refactoring the census data. The rumor structure is defined and stored as protobuf messages. So adding a new field say `pkg_incarnation` with a default value of 0 should be both backward and forward compatible with no issues even if someone chose to rollback after updating the supervisor. The older versions of the code would just ignore newer protobuf fields while the newer versions of the code would start with a default pkg_incarnation of 0 for existing services which should be fine for our purposes.

atrniv (approved):
Looks good to me 👍 !

PR #8576, opened 2022-07-21
Bump crates and pin the crate versions for rants

sajjaphani:
Use the [v0.6.0](https://github.com/habitat-sh/rants/tree/v0.6.0) branch of [rants](https://github.com/habitat-sh/rants), which pinned the dependency crates. It is good to pin the dependency crate versions for libraries. The dependabot is messing up with the dependent crate versions for rants while bumping the versions.

Also Closes
https://github.com/habitat-sh/habitat/pull/8575
https://github.com/habitat-sh/habitat/pull/8574
https://github.com/habitat-sh/habitat/pull/8573
https://github.com/habitat-sh/habitat/pull/8572

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ dependencies = [
  "memchr",
  "pin-project-lite",
  "tokio",
- "tokio-util 0.7.3",
+ "tokio-util",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


mwrock:
I think we should push the latest changes to the upstream main to our fork so we get the TLS fixes and then edit the cargo.toml file on top of that.

mwrock (approved):

PR #8607, opened 2022-08-19
fix hung rolling update if leader dies while waiting for an update

mwrock:

Should fix end to end rolling tests.


sajjaphani (approved):
Looking good to me!

PR #8611, opened 2022-08-29
cleanly close ctl_gateway server stream after message is sent

mwrock:

With the recent updates to rustls, calls to the supervisor's ctl-gateway are resulting in an `UnexpectedEof`.

On our (supervisor client) side, this is coming from [`poll_read`](https://github.com/habitat-sh/habitat/blob/586e7f218a9a6f5b3c29d496522c4051450705f3/components/core/src/tls/rustls_wrapper/tcp_or_tls_stream.rs#L92) in `TcpOrTlsStream`.

Further down the stack in rustls, this error is coming from the call to [`read`](https://github.com/rustls/rustls/blob/162bea2f41d08613ee5e6cdafb0b257d4ea0633e/rustls/src/conn.rs#L237) in the underlying rustls connection's `reader` struct. Here, it will return the `UnexpectedEof` if the peer did not cleanly close.

In order to cleanly close, the ctl-gateway server in the supervisor needs to send a `close_notify` once it has sent its message. This is done by calling the stream's `poll_shutdown` function. So the change in this PR is to do exactly that. I have validated that this eliminates the `UnexpectedEof` in a manualy build supervisor and new client and have also run this in the ctl-gateway's e2e tests where they passed.

Note that in rustls v0.19 which is what we were using before, this need to call `close_notify` was not present. That version would call the [`read`](https://github.com/rustls/rustls/blob/3c390ef7c459cc1ef2504bd9d1fefdcb7eea1c20/rustls/src/session.rs#L771) function of the underlying `session` struct. This would simply return a result<Ok(0)> to `tokio-rustls` and ultimately result in oue own `poll_read` mentioned above receiving a `Ready(Ok(()))`.

### DIFF
--- a/components/sup/src/ctl_gateway/server.rs
+++ b/components/sup/src/ctl_gateway/server.rs
@@ -56,7 +56,8 @@ use std::{error,
           sync::{Arc,
                  Mutex},
           time::Duration};
-use tokio::{net::TcpListener,
+use tokio::{io::AsyncWrite,
+            net::TcpListener,
             task,
             time};
 use tokio_util::codec::Decoder;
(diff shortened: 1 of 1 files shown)


jasonheath (approved):
Looks reasonable.  New async handling of an error resulting from API changes.

atrniv (approved):
Looks good. Tried out the changes on an x86_64 linux system, seems to be working fine.

PR #8634, opened 2022-09-20
address easy warnings in service.rs of component/sup

jasonheath:
Signed-off-by: Jason Heath <jason.heath@progress.com>

### DIFF
--- a/components/sup/src/manager/service.rs
+++ b/components/sup/src/manager/service.rs
@@ -1653,6 +1653,8 @@ impl<'a> Serialize for ServiceProxy<'a> {
 }
 
 #[cfg(test)]
+#[cfg(all(any(target_os = "linux", target_os = "windows"),
+          target_arch = "x86_64"))]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


mwrock:
What is the warning this is addressing? Looks like it is likely not valid based on the pipeline failures.

jasonheath:
> What is the warning this is addressing? Looks like it is likely not valid based on the pipeline failures.

Capturing offline discussion here.  This file was triggering warn(dead_code) on a use stmt and a function.  Both are used in an async block so my suspicion is that the async was throwing the linter off.  Change is ultimately to #[allow(dead_code)] with a comment as to why.  Build failed the first time around because the "use crate::test_helpers::*" couldn't be removed as it was being used async.

mwrock (approved):

jasonheath:
Addresses 2 warnings (1 for unused_imports and 1 for dead_code) on macOS where a test wasn't appropriate for the platform but not completely conditionally compiled out.

PR #8642, opened 2022-10-04
remove use of try when statting in studio

mwrock:

### DIFF
--- a/components/studio/bin/hab-studio.sh
+++ b/components/studio/bin/hab-studio.sh
@@ -1018,7 +1018,7 @@ chown_artifacts() {
   if [ -z "${NO_MOUNT}" ] \
   && [ -z "${NO_ARTIFACT_PATH}" ] \
   && [ -d "$ARTIFACT_PATH" ]; then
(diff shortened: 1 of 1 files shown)


mwrock:
I had some ideas for some further research. Here is what I found:

* the buildkite image just changed from ubuntu 18.04 to 20.04
* Locally 20.04 "seems" to work but it actually exhibits the same behavior on the `try` function and leaves behind the mounts.
* Locally 20.04 will succeed after leaving behind the mounts. There is logic in the studio script to only create a new mount if there is no existing mount.
* The reason it is failing to create the mount in the buildkite image is because when it looks to see if the mount already exists, it does not see the mount. For some reason `busybox mount` is not returning the existing mounts. Normal `mount` does. I am guessing this is a docker related issue.
* The version of bash seems irrelevant. While ubuntu 18.04 (4.4.20) and 20.04 (5.0.17) have different versions of bash, the studio script uses /bin/sh which uses `dash` and not bash. I am not at all an expert here but it seems dash is yied to ubuntu version.

I still think we should move forward with the change in this PR because that will prevent the orphaned mounts which seems like something we want even if the build was succeeding.

Issue #8501, opened 2022-05-10
Proposal: Native Packages, enabling Supervisor service support for platforms without packages.


atrniv:
## Summary

This is a proposal for introducing the idea of habitat package types. There will primarily be two types of packages:
- **Standard Packages** : These are currently the only type of packages we have. A standard package is built in a cleanroom environment and it is guaranteed to run in any environment once it has been built. This is possible because every dependency of a standard package is also another standard package.
- **Native Packages**: This is a new type of package. Native packages are built in the host environment that the habitat packaging tool gets run in. These packages transfer the responsibility of managing runtime dependencies to the user. A native package is essentially a thin wrapper around an existing application that makes use of hooks to define how to manage the application lifecycle.

## Motivation

The primary advantages we gain by introducing the concept of a *Native Package* is the ability to run the Habitat Supervisor on platforms where we currently do not have a package ecosystem bootstrapped (eg: ARM, Solaris, M1 Macs). This will enable us to open a broader set of use cases for Habitat usage.

## Design

To enable support for native packages, the `HAB_NATIVE_PACKAGE_SUPPORT` environment variable must be set to `true`.

To create a Native Package you must:
- Put `# package_type: "native"` as the topmost line of your `plan.sh` / `plan.ps1` file.
- You cannot have the following plan parameters:
  - pkg_deps
  - pkg_build_deps
  - pkg_interpreters 
- You must have a run hook defined for your package. 
- You can use any interpreter for your hooks. You must use a shebang line at the start of all your hook files with an appropriate path to the interpreter you wish to use.
- There will be a new Metafile in a package called `PKG_TYPE`. It will contain `native` or `standard` to indicate the type of package.

## Behavior Changes
- Invoking `hab pkg build` will automatically run the hab plan build script on the host system when it detects that the plan builds a native package.
- Invoking `build` within a studio environment for a native package plan will result in an error message suggesting the user to run `hab pkg build` outside the studio.
- Adding any unsupported plan parameter such as `pkg_deps` will result in a build time error message.
- Native Package hooks will have to ensure that the interpreter they use and any commands that are used in them are available in the runtime environment.

## Design Issues Considered

- Supporting dependencies between native packages is likely to be error prone as the user has to ensure that they always build and run the package in an environment that is compatible with all dependencies. Hence we will also not be supporting native packages that are libraries at this time.
[...]

PR #8650, opened 2022-10-20
Native Package enhancements to enable package bootstrapping

atrniv:
This PR adds additional features to the native packages capability that are required to bootstrap a set of packages on a new architecture.

- It adds the ability to specify `pkg_deps` and `pkg_build_deps` for native packages.
- It adds a new type of chroot studio `bootstrap`. The `bootstrap` studio type makes use of various `build-tools` native packages from the [bootstrap plans repo](https://github.com/habitat-sh/bootstrap-plans) to create a minimal studio cleanroom environment that can be used to make isolated builds.
- Fixes a bug that caused failed native package builds to make the hab cli command die with a success exit code (0). Any failure in a native package build will now return a non-zero exit code.


mwrock:
Do we care if a native package has a dep on a standard package or if a standard package has a dep on a native package? I would guess we would care more about the later. If a standard has a dep on a native package then the standard package may not be as portable as one might expect.

mwrock:
I think we should add some end to end tests for native packages.

atrniv:
> Do we care if a native package has a dep on a standard package or if a standard package has a dep on a native package? I would guess we would care more about the later. If a standard has a dep on a native package then the standard package may not be as portable as one might expect.

@mwrock , there are valid use cases for all combinations of dependencies, I've added a comment to the original proposal elaborating on this: https://github.com/habitat-sh/habitat/issues/8501#issuecomment-1286772725

atrniv:
> I think we should add some end to end tests for native packages.

Yes, perhaps a E2E tests that a native package build succeeds in an environment with the necessary tools and fails otherwise might make sens.

mwrock (approved):

PR #8682, opened 2022-12-02
add SYSTEM ACL to /hab/svc in docker studio

mwrock:

The LocalSystem account that the Habitat windows service runs under is failing when started with an Access Denied error. After digging further this is due to a missing ACL in /hab/svc and its child files and folders for `NT AUTHORITY\SYSTEM`. Adding this ACL recursively fixes the Access Denied error.

It is still unclear why this recently started. I'm assuming it is only related to Windows 11 because my laptop upgraded around the time this started happening. Also perplexing is why if I build the image locally, the ACL gets added automatically but not on the buildkite server. My suspicion is this is related to building on a client SKU (win 11) vs a server SKU (server 2022) which has slightly different rules for default folder ACLs.


mwrock:
Did a bunch more research and experimentation. I could not reproduce the problem on win 2022 aws VMs. I created a win 11 vm on azure with nested virtualization and setup docker and hyperv. Even there, which closely matched my environment, I was not able to reproduce. I upgraded my docker desktop to 4.15.0 which is the same version on the VM but the container still had the missing ACL and the habitat service failed to start.

Finally I removed the studio image locally and then reran `hab studio enter -D` which reinstalled it. That seemed to fix the issue and the ACL was there. It's hard to say how this ever happened. Could have been a docker upgrade or windows update but I am going to close this for now.

PR #8714, opened 2023-01-18
Fix windows 7 support

atrniv:
* The version of the "ipc-channel" crate is frozen at 0.15.0 to avoid the usage of the "GetOverlappedResultEx" system call, which was introduced in version 0.16.0.
* An error type wrapper named "IPCError" is created around "ipc_channel::ipc::IpcError" because it does not implement the "std::error::Error" trait, which is required to use it with the "#[from]" attribute of the "thiserror" crate's "Error" derive macro.

It is important to note that these changes should ideally be reverted once support for Windows 7 is discontinued around January 2024. This will allow for the usage of the "GetOverlappedResultEx" system call and the "ipc_channel::ipc::IpcError" trait. More information about "GetOverlappedResultEx" can be found at the following link: https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-getoverlappedresultex And the information about the commit that introduced the change can be found at: https://github.com/servo/ipc-channel/commit/eb08381a30bc71a534a0a73ab98c05bca7a12f82

Signed-off-by: Johny Jose <johny.jose@progress.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1767,9 +1767,9 @@ dependencies = [
 
 [[package]]
 name = "ipc-channel"
-version = "0.16.0"
+version = "0.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7cb1d9211085f0ea6f1379d944b93c4d07e8207aa3bcf49f37eda12b85081887"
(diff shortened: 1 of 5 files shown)


mwrock (approved):
Looks good!

PR #8720, opened 2023-01-25
fixes rolling update leader resetting pkg_incarnation

atrniv:
During rolling updates, the update worker relies on the value of the service group's "pkg_incarnation" to determine if an update has occurred. If a leader in the group has a higher pkg_incarnation than the followers, the followers will update to the leader's package.

This commit addresses a bug that caused a leader to reset it's pkg_incarnation to 0 in certain scenarios, resulting in followers remaining in a perpetual state of waiting for the leader to catch up to the latest pkg_incarnation.

Signed-off-by: Johny Jose <johny.jose@progress.com>

### DIFF
--- a/components/sup/src/manager.rs
+++ b/components/sup/src/manager.rs
@@ -657,8 +657,8 @@ pub struct Manager {
     // something else (maybe a HashMap?) in order to cleanly manage
     // the different operations.
     busy_services: Arc<Mutex<HashSet<PackageIdent>>>,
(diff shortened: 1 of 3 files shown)


mwrock:
In the specific customer scenario, this was happening when the running service could not be gracefully terminated and had to be forcefully killed. So I was able to reproduce the problem by adding the following to the run hook:

```
_term() {
  echo "run hook is terminating"
  sleep 30
}

trap _term TERM
```

This PR resolves the issue and performs the rollback without the followers getting hung.

mwrock (approved):

PR #8737, opened 2023-02-14
Revved rust from 1.62.1 to 1.67.1

jasonheath:
Signed-off-by: Jason Heath <jason.heath@progress.com>


components/core/src/os/process/exec/unix.rs
```diff
@@ -37,6 +40,14 @@ pub fn hook_command<X, I, K, V>(executable: X, env: I, ids: Option<(Uid, Gid)>)
         with_user_and_group_information(&mut cmd, uid, gid);
     }
 
+    unsafe {
```
mwrock, line 43: I think I would add a comment here about why we had to do this with a link to the rust github pr.

mwrock (approved):
Just a note on adding the comment. Feel free to merge after adding that.

components/core/src/os/process/exec/unix.rs
jasonheath: Added comment per feedback and will now merge.

PR #8741, opened 2023-02-16
get rid of traitobject and typemap deps

mwrock:
This should get rid of the "critical" security alerts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -651,6 +651,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "destructure_traitobject"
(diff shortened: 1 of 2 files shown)


jasonheath:
log4rs has a dependency but I'm not sure if or where we are using the old typemap or the new typemap_or crates.  I really focused on trying to find use statements so I may have missed something but is it possible that we could drop the direct dependency from habitat_core?

mwrock:
Good catch. I removed typemap-ors from habitat_core and it compiles fine.

jasonheath (approved):

PR #8773, opened 2023-03-03
cargo audit github action

jasonheath:
There's may be some noise on this branch as I rev different things working to test out this new functionality.  This commit is the first cut of the files and after this I intend to upgrade some things in order to see it working.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,44 @@
+# cargo audit config file
+# https://github.com/rustsec/rustsec/blob/main/cargo-audit/audit.toml.example
+#
+# All of the options which can be passed via CLI arguments can also be
+# permanently specified in this file.
+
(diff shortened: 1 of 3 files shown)


jasonheath:
One point of discussion.  As currently committed `cargo audit` will run for every push.  We can instead run it only when the Cargo.lock, a Cargo.toml and the .cargo/audit.toml files change during a push and I think that's a useful change.  We could also have it run daily via a cron job.  I kind of like that as well.  Both are easy to add.

mwrock (approved):
Looks good!

jasonheath:
So as is?  No changes as to pushing for those three files or putting it on a schedule?

mwrock:
I think it's fine as is. It's not expensive and will finish way before the verify check and we are more likely to see it in a PR than on a schedule.

PR #8776, opened 2023-03-06
bumped buildkit timeout from 30 to 60 

jasonheath:
I asked Sean Simmons about the ARM builds that were consistently failing all day.  He suggested that I bump the buildkit timeout from 30 to 60.  In so doing several whitespace formatting changes came along.  I believe that's due to my having installed Red Hat's YAML plugin which did the formatting on file save.  I can scrap those but thought I would initially submit this with them in place.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -1,7 +1,7 @@
 expeditor:
   defaults:
     buildkite:
-      timeout_in_minutes: 30
+      timeout_in_minutes: 60
 
 steps:
 #######################################################################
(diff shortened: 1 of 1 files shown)


mwrock (changes requested):
I totally agree with the timeout increase but there are a lot of whitespace changes here that are unrelated.

jasonheath:
I've reverted all of the whitespace changes except one where there were somehow two blank space on an otherwise empty line.  I spent a bit fighting with the formatting settings in vscode and then realized that all other blank lines are simply line breaks without any other bit of whitespace on it so I stopped fighting it.

mwrock (approved):

PR #8777, opened 2023-03-06
Restore notify crate for all OSes

jasonheath:
When I did the work on the notify crate I tried to limit dependencies we weren't directly relying on by specifying it as a dependency for linux and widows without default features.  This broke the build on macOS and this commit puts it back as a unbounded dependency for all OSes with its defaults.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -914,6 +914,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "fuchsia-zircon"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jasonheath:
I also commented out some tests that best I could tell weren't appropriate for macOS anymore.  Definitely want the review of others on that decision.

components/hab/src/cli.rs
```diff
@@ -143,12 +143,15 @@ mod tests {
 
     #[test]
     fn legacy_appliction_and_environment_args() {
-        let r = get(no_feature_flags()).get_matches_from_safe(vec!["hab",
-                                                                   "sup",
-                                                                   "run",
-                                                                   "--application",
-                                                                   "--environment=env"]);
-        assert!(r.is_ok());
+        #[cfg(not(target_os = "macos"))]
```
mwrock, line 146: I don't think we want to add this constraint. None of our pipelines run these tests on macos. If we do add macos, then we would likely want them to run these tests.

components/hab/src/cli.rs
```diff
@@ -178,6 +181,7 @@ mod tests {
         }
 
         #[test]
+        #[cfg(not(target_os = "macos"))]
```
mwrock, line 184: same as above.

mwrock (changes requested):

components/hab/src/cli.rs
jasonheath: Removed

components/hab/src/cli.rs
jasonheath: Removed

mwrock (approved):

PR #8800, opened 2023-03-14
Jah/rust 1.68.0

jasonheath:


components/common/src/ui.rs
```diff
@@ -53,21 +53,14 @@ impl From<Color> for termcolor::Color {
     }
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Default)]
 pub enum GlyphStyle {
+    #[default]
```
mwrock, line 58: seems like the default woul be full for both platforms now whereas it was limited for windows.

mwrock (changes requested):

mwrock (approved):

PR #8819, opened 2023-03-24
remove special aarch64-linux handling in the installer

mwrock:
On a generic ubuntu aarch64-linux instance, I cannot run the supervisor due to the inability to find libzmq. I feel like there was a reason we have aarch64 pull the sup and launcher from the archive rather than `hab pkg install`ing them which would include the libzmq deps and link correctly. This PR does not pull the sup or launcher from the archive which would force `hab` to download them. With this installer, I can install hab and then run `hab sup run` successfully assuming it is pointing to a builder that has those packages for arm. I am just putting this up as a draft for now to allow @atrniv to weigh in on this.

### DIFF
--- a/components/hab/install.sh
+++ b/components/hab/install.sh
@@ -254,86 +254,24 @@ install_hab() {
       install -v "${archive_dir}"/hab /usr/local/bin/hab
       ;;
     linux)
-      case "${arch}" in
-        x86_64)
-          local _ident="core/hab"
+      local _ident="core/hab"
 
-          if [ -n "${version-}" ] && [ "${version}" != "latest" ]; then
-            _ident+="/$version";
-          fi
+      if [ -n "${version-}" ] && [ "${version}" != "latest" ]; then
+        _ident+="/$version";
+      fi
 
-          info "Installing Habitat package using temporarily downloaded hab"
-          # NOTE: For people (rightly) wondering why we download hab only to use it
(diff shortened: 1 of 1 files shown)


atrniv:
@mwrock, we installed the archive for arm differently than the x86_64 platform. This was necessary because there were no native or standard packages available on the public SAAS builder for aarch64-linux. Therefore, we had to install everything from the release archive manually, without involving any packages.

The aarch64-linux habitat build was also a direct cargo build without using a habitat plan. Its primary purpose was to support use cases for native packages with a supervisor on aarch64-linux. In addition to running the install script, you needed to make sure that zmq was available in the runtime environment for aarch64-linux. I think 

If we merge these changes in right now, it will break the installation of hab-sup and hab-launcher on aarch64-linux for the existing use case of running native packages. This would also block the next habitat release on the publishing of the aarch64-linux packages to the SAAS builder. 

We will need to re-work and extend a lot of the build and packaging CI and test pipeline once we officially make the aarch64-linux packages available on the SAAS builder. We can also make the aarch64-linux package installation identical to the x86_64-linux installation once that's done.

mwrock:
After speaking about this offline, we agreed that this installer should provide a "standard" package (vs. native) experience across all platforms. Until we officially support ARM, this will require setting `HAB_BLDR_URL` to `https://bldr.acceptance.habitat.sh` (or a on-prem builder that has the ARM bundle) where we are currently keeping the ARM packages.

Issue #8470, opened 2022-03-25
Refactor Supervisor Gateway HTTP Server Code


atrniv:
The Supervisor HTTP Gateway serializes core internal structures into a JSON string to break any direct dependencies on the internal types. It does this through a proxy class which holds a reference to the internal structure and implements serde serialize to read the internal structure and serialize it to a string. This string is then de-serialized into a generic JSON value and re-serialized into HTTP JSON responses.

This is counter productive for the following reasons:
- The JSON string representation of an object almost always with consume more memory than it's native in-memory representation.
- The serialized string is de-serialized, read and re-serialized on every HTTP request.
- There is no type-safe way to access the serialized data from the HTTP gateway, making the code more prone to runtime bugs because of a difference in data structure.

A better approach would be the following:
- Create a new API type for each internal struct with fully owned fields that conforms to the HTTP API.
- When the internal structures change, re-construct the API type.
- Serialize the API type as required on HTTP calls.

This would give us the following benefits:
- Reduced memory usage, due to representing the data as in-memory structs vs a JSON string
- Reduced cpu usage, as we don't need to repeatedly de-serialize and re-serialize. We only serialize the API structs as required.
- Simpler serialization code, we just clone the internal struct's fields onto the API struct and derive `Serialize` on the API struct.
- The HTTP gateway server will no longer have to deal with raw strings, it can use fully typed structs, making it more robust to change.

PR #8845, opened 2023-06-06
Refactor Supervisor HTTP Gateway to eliminate service data inconsistencies

atrniv:
This PR refactors the HTTP Gateway code to use in memory structs as the source of truth instead of serialized JSON data. This also allows us to directly update these structs on service changes and prevent any data inconsistencies.
Namely the health check data between the `/services` and `/services/{svc}/{group}/{org}/health` endpoints should be identical.

This also addresses #8470.


mwrock (approved):
This looks good! Before merge, can you fix the windows clippy error and rebase on main which will fix the cargo audit failure?

PR #8987, opened 2023-08-24
use rustls-webpki instead of webpki

jasonheath:
This PR changes direct dependencies on webpki to rustls-webpki due to RUSTSEC-2023-0052. It does not completely address RUSTSEC-2023-0052 as of yet.  I believe for that to happen that actix-web will have to address their dependency and that their work to do so is starting with [this pull request](https://github.com/actix/actix-web/pull/3112).  See 
[webpki-reverse-dependencies.txt](https://github.com/habitat-sh/habitat/files/12429497/webpki-reverse-dependencies.txt) to review remaining indirect depending on webpki.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1274,6 +1274,7 @@ dependencies = [
  "reqwest",
  "retry",
  "rustls",
+ "rustls-webpki",
  "same-file",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


jasonheath:
Myself, @mwrock, and @atrniv discussed this during today's daily. We agreed to go ahead and merge this PR so that we address our direct dependencies so that actix-web lands the rustsec advisory should clear.  In the meantime the continued presence of the rustsec advisory in PRs will keep that in our minds.  I'll also keep link dependabot PRs back to this if it feels appropriate to push them on through.

mwrock (approved):

PR #9010, opened 2023-09-05
update to rustls 0.21.7

jasonheath:
Combining the work of a couple of dependentbot branches into one to untangle a dependency knot.

The actual work behind the changes can be seen in [rustls-0_21-update.sh](https://github.com/habitat-sh/habitat/files/12528006/rustls-0_21-update.sh.v2.txt) and the actual execution can be seen in [rustls-0_21-update.log](https://github.com/habitat-sh/habitat/files/12528010/rustls-0_21-update.log.v2.txt).

This will close the dependabot PRs for [rustls](https://github.com/habitat-sh/habitat/pull/8997) and [actix-web](https://github.com/habitat-sh/habitat/pull/9003) and resolve [RUSTSEC-2023-0052](https://rustsec.org/advisories/RUSTSEC-2023-0052.html)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,18 +21,18 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.3.0"
+version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0070905b2c4a98d184c4e81025253cb192aa8a73827553f38e9410801ceb35bb"
+checksum = "a92ef85799cba03f76e4f7c10f533e66d87c9a7e7055f3391f09000ad8351bc9"
 dependencies = [
  "actix-codec",
  "actix-rt",
  "actix-service",
  "actix-tls",
  "actix-utils",
- "ahash",
+ "ahash 0.8.3",
  "base64 0.21.3",
- "bitflags 1.3.2",
+ "bitflags 2.4.0",
  "bytes",
  "bytestring",
  "derive_more",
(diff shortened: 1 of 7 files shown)


jasonheath:
I'll squash this down to a single commit before merging but I left some extra commits in there that are a bit more thematic.

mwrock:
It looks like this is a global reformat instead of just formatting the files effected by this change. Was that intended? I thought rustfmt only cared about changed files.

mwrock:
Looks like this is downgrading `rand` and `regex` which does not seem right. Seems unrelated to the main cargo updates unless I am missing something.

jasonheath:
> Looks like this is downgrading `rand` and `regex` which does not seem right. Seems unrelated to the main cargo updates unless I am missing something.

Yeah, I missed that.  Will address that.  I think that slipped in when I was address other similar regressions.  Tact should be the same as for lazy_static and base64 so I won't update those files illustrating what I was doing unless it ends up unexpectedly different.

jasonheath:
> It looks like this is a global reformat instead of just formatting the files effected by this change. Was that intended? I thought rustfmt only cared about changed files.

I was surprised by that too and part of the reason I left the formatting as its own commit.  I tried using the command in the Makefile but it was crashing saying it didn't have the right toolchain (looking for linux when I'm on macOS). I tried to fix that for a moment and then I just ran `cargo +$(RUSTFMT_TOOLCHAIN) fmt` which is what the make fmt ultimately should end up doing.  I think the reason we may be seeing all of these is because this might be the first commit since the release revved nightly.  Looking back through the history though that merge seemed to bring in formatting changes so I don't know why it didn't bring all of these at that time.  Trying some other experiments trying to reproduce whatever I ran I'm not getting the same results.  But yeah, I specifically left those as their own commit to make it easy to toss.

mwrock (approved):
Just got one formatting issue to clear up.

jasonheath:
Addressed the fmt issue and squashed commits.

PR #9027, opened 2023-09-13
Removed RUSTSEC-2022-0048 from .cargo/audit.toml

jasonheath:
RUSTSEC-2022-0048 was rescinded as the xml-rs crate was forked and is maintained once more.


jasonheath:
[This dependabot PR](https://github.com/habitat-sh/habitat/pull/9015) may also our remove our warnings about [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071) as chrono is removing any dependency on the time crate.  I've moved this back to draft and will take appropriate action later.

jasonheath:
Will create new branch and PR after all of the other changes.

PR #9077, opened 2023-10-27
Fix the verify pipeline to get macOS building again

jasonheath:
### Problem

Our buildkit pipelines are failing because the macOS install script step is timing out.

### Root Cause

- [p11-kit releases 0.25.1 stable](https://github.com/p11-glue/p11-kit/releases)
- [Homebrew rolls out a 0.25.1 forumlae requiring meson and updates the 0.25.0 bottle dropping macOS 11 (Big Sur)](https://github.com/Homebrew/homebrew-core/commits/d75d5dde165275005185aee61c10b6dc1c597c6d/Formula/p/p11-kit.rb)
- Our gnupgp dependency is now using the updated p11-kit which brings in meson, ninja, and python as new dependencies and the install step is timing out.

### Solution

- Change the timeout for the step to 60m matching the timeout for other macOS steps

### While we're here...

- Since macOS 11 (Big Sur) has apparently gone EOL as of Sept 26, 2023, change macOS version to 12 (Monterey) as that's [the last approved version in expeditor](https://github.com/chef/expeditor/blob/5e8f0b19519b9abe40dea19c433f4443ddaa259e/expeditor-ruby/lib/expeditor/cli/buildkite/executor/macos.rb#L54).

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -335,9 +335,9 @@ steps:
     expeditor:
       executor:
         macos:
-          os-version: "11"
+          os-version: "12"
           inherit-environment-vars: true
-    timeout_in_minutes: 10
+    timeout_in_minutes: 60
     retry:
       automatic:
         limit: 10 # Addressing current Anka system timeouts due to oversubscription
@@ -998,7 +998,7 @@ steps:
     expeditor:
       executor:
         macos:
-          os-version: "10.15"
+          os-version: "12"
           inherit-environment-vars: true
     timeout_in_minutes: 60
     retry:
@@ -1018,7 +1018,7 @@ steps:
     expeditor:
       executor:
         macos:
-          os-version: "11"
+          os-version: "12"
           inherit-environment-vars: true
     timeout_in_minutes: 60
     retry:


jasonheath:
The cargo audit failure is due to [the yanking of many version of the ahash crate](https://github.com/tkaitchuck/aHash/issues/163).  We don't have a direct dependency on ahash so we're waiting for updates from the projects leveraging ahash.  This is commit is one of those projects and the relationship to ahash changes a bit but actix needs changes and I think more upstream changes may be needed in hashbrown before tokio clears.

.expeditor/verify.pipeline.yml
```diff
@@ -335,9 +335,9 @@ steps:
     expeditor:
       executor:
         macos:
-          os-version: "11"
+          os-version: "12"
           inherit-environment-vars: true
-    timeout_in_minutes: 10
+    timeout_in_minutes: 60
```
mwrock, line 340: I don't think we need this bump.

mwrock:
The audit failure should be cleared after https://github.com/habitat-sh/habitat/pull/9080 is merged.

mwrock:
I went ahead and merged #9080 to clear up all the audit failures which also bumped macos so I am closing this but thanks for submitting this because it helped me realize what #9080 needed to pass a few weeks ag.

PR #9081, opened 2023-11-06
toToml helper fix

sajjaphani:
The issue has been observed during the local setup of the builder.

The issue is failing with the following scenarios:
[default,toml] 
my_values = ["a", "b", "c"]
[config,toml]
my_values = {{toToml cfg.my_values}}

If we configure it as:
build_targets = {{cfg.my_values}}
It writes as:
my_values = [a, b, c,]

**Here are the simple instructions to reproduce it using builder source:**

1. Enter the Habitat Studio: hab studio enter
2. Start the builder: start-builder

The error is particularly occurring on the line `build_targets = {{toToml cfg.build_targets}}` in the `config.toml` file of the builder-jobsrv component.

The reason seems to be the updating toml crate (https://github.com/habitat-sh/habitat/commit/4c8b5fbef186ac5db04236226c15742b56238393), particularly replacing `to_vec` with `to_string`.

According to the change log (https://github.com/toml-rs/toml/blob/main/crates/toml/CHANGELOG.md#compatibility-2), `to_string` should be used for TOML documents, not values.

I've also added a few test cases to cover these scenarios. However, I'm not sure if this is the only case where we need this.

### DIFF
--- a/components/common/src/templating/helpers/to_toml.rs
+++ b/components/common/src/templating/helpers/to_toml.rs
@@ -17,13 +17,27 @@ impl HelperDef for ToTomlHelper {
         // `serde_json` has been compiled with the `preserve_order`
         // feature, since order *is* important for TOML (values must
         // be emitted before tables).
-        let toml = toml::ser::to_string(&param).map_err(|e| {
-                                                   RenderError::new(format!("Can't serialize \
-                                                                             parameter to TOML: \
-                                                                             {}",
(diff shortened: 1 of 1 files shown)


mwrock (approved):

sajjaphani:
The Hab version 1.6.652 does not have this problem, and the issue has been identified with version 1.6.826.

PR #9082, opened 2023-11-15
CHEF-3310: Ensure hab pkg install verifies all deps are installed

atrniv:
This PR fixes a bug that causes Habitat to ignore missing tdeps during package installation if the parent package is already installed


mwrock (approved):
This looks good and works as advertised. I think we should add an e2e test for this.

mwrock (approved):

mwrock:
@atrniv if you rebase this, it should get pas the mac and audit failures.

PR #9273, opened 2024-06-06
CHEF-12227: vulnerable owning_ref crate

jasonheath:
- change owning_ref crate to safer_owning_ref crate
- use habitat.sh/rants 6.1 which uses safer_owning_ref crate
- added .cargo/advisory-db..lock to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.sig.key
 *.sym.key
 .cargo/advisory-db/
+.cargo/advisory-db..lock
 .DS_Store
 .envrc
 .secrets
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 *.sig.key
 *.sym.key
 .cargo/advisory-db/
+.cargo/advisory-db..lock
```
mwrock, line 6: was this intentional?

mwrock (approved):

.gitignore
jasonheath: It was.  That file was always empty when it was created and while I'm generally in favor of committing lock files I wasn't finding anything useful regarding that particular file. It also just felt weird with the .. in the filename so I decided to add it to the .gitignore now and reverse that decision later if appropriate.

jasonheath:
This PR also depended upon https://github.com/davidMcneil/rants/pull/29

PR #9302, opened 2024-07-09
CHEF-12417: Remove hab bldr job commands

jasonheath:
Replaces hab bld job cmds with a message to contact account team


jasonheath:
This first version of this code takes a rather conservative route.  It augments the doc comments with the message and any invocation of the subcommands also prints the message.  The functionality replaced my the messaging printing it commented out and simply dead at this goal bing to keep the changes as small as possible as other things are in flight.  I can quickly remove the dead code if desired.

components/hab/src/cli/hab.rs
```diff
@@ -163,6 +166,9 @@ pub enum Channel {
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(no_version, aliases = &["j", "jo"], settings = &[AppSettings::ArgRequiredElseHelp, AppSettings::SubcommandRequiredElseHelp])]
 /// Commands relating to Habitat Builder jobs
```
agadgil-progress, line 168: I am just thinking whether we should add `(Deprecated and UNSUPPORTED)` or something similar at the end - so this is readily available in the help when someone runs the command `hab bldr help`

components/hab/src/cli/hab/bldr.rs
```diff
@@ -52,6 +52,9 @@ pub struct BldrJobStatusSourceGroup {
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(no_version)]
 /// Commands relating to Habitat Builder jobs
```
agadgil-progress, line 54: Same as above

components/hab/src/cli/hab/bldr.rs
```diff
@@ -61,6 +64,9 @@ pub enum Job {
 }
 
 /// Get the status of one or more job groups
```
agadgil-progress, line 66: Same as above so that user knows when (s)he checks help.

components/hab/src/cli/hab/bldr.rs
```diff
@@ -61,6 +64,9 @@ pub enum Job {
...
+///
+/// NOTICE: Public Builder Build Functions are no longer supported. Please reach
+/// out to your account team if you were using these features.
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(name = "status", no_version)]
 pub struct JobStatus {
```
agadgil-progress, line 72: Not sure whether these arguments make sense any more for example if one runs `hab bldr job status`, a user is shown CLI parsing error (which is not same as the NOTICE above). When this is anyways not supported, don't think we need to deal with these options anymore.

components/hab/src/cli/hab/bldr.rs
```diff
@@ -77,6 +83,9 @@ pub struct JobStatus {
 }
 
 /// Cancel a build job group and any in-progress builds
```
agadgil-progress, line 85: Same as above - Add (Deprecated and UNSUPPORTED)?

components/hab/src/cli/hab/bldr.rs
```diff
@@ -77,6 +83,9 @@ pub struct JobStatus {
...
+///
+/// NOTICE: Public Builder Build Functions are no longer supported. Please reach
+/// out to your account team if you were using these features.
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(name = "cancel", no_version)]
 pub struct JobCancel {
```
agadgil-progress, line 91: We might want to get rid of the options to this as well. same as `JobStatus` above.

components/hab/src/cli/hab/bldr.rs
```diff
@@ -93,6 +102,9 @@ pub struct JobCancel {
 }
 
 /// Demote packages from a completed build job from a specified channel
```
agadgil-progress, line 104: Same as above.

components/hab/src/cli/hab/bldr.rs
```diff
@@ -93,6 +102,9 @@ pub struct JobCancel {
...
+///
+/// NOTICE: Public Builder Build Functions are no longer supported. Please reach
+/// out to your account team if you were using these features.
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(name = "demote", no_version)]
 pub struct JobDemote {
```
agadgil-progress, line 110: Same as above.

components/hab/src/cli/hab/bldr.rs
```diff
@@ -118,6 +130,9 @@ pub struct JobDemote {
 }
 
 /// Promote packages from a completed build job to a specified channel
```
agadgil-progress, line 132: Same as above

components/hab/src/cli/hab/bldr.rs
```diff
@@ -118,6 +130,9 @@ pub struct JobDemote {
...
+///
+/// NOTICE: Public Builder Build Functions are no longer supported. Please reach out to your account
+/// team if you were using these features.
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(name = "promote", no_version)]
 pub struct JobPromote {
```
agadgil-progress, line 138: Same as above.

components/hab/src/cli/hab/bldr.rs
```diff
@@ -143,6 +158,9 @@ pub struct JobPromote {
 }
 
 /// Schedule a build job or group of jobs
```
agadgil-progress, line 160: Same as above

components/hab/src/cli/hab/bldr.rs
```diff
@@ -143,6 +158,9 @@ pub struct JobPromote {
...
+///
+/// NOTICE: Public Builder Build Functions are no longer supported. Please reach
+/// out to your account team if you were using these features.
 #[derive(ConfigOpt, StructOpt)]
 #[structopt(name = "start", no_version)]
 pub struct JobStart {
```
agadgil-progress, line 166: Same as above

components/hab/src/error.rs
```diff
@@ -83,6 +84,11 @@ impl fmt::Display for Error {
         let msg = match *self {
             Error::APIClient(ref e) => e.to_string(),
             Error::ArgumentError(ref e) => e.to_string(),
+            Error::BuilderBuildFunctionsRemoved => {
```
agadgil-progress, line 87: This is perfect, a single place to deal with errors!

components/hab/src/cli/hab/bldr.rs
```diff
@@ -77,6 +83,9 @@ pub struct JobStatus {
 }
 
 /// Cancel a build job group and any in-progress builds
+///
+/// NOTICE: Public Builder Build Functions are no longer supported. Please reach
+/// out to your account team if you were using these features.
```
agadgil-progress, line 88: Should it be if you **are** using these features? instead of you **were** ?

agadgil-progress (changes requested):
The code changes per se look okay. A couple of main comments (repeated a few times for each sub command) - 

1. Being a bit more explicit about **unsupported** features in the CLI help. 
2. We don't have to handle the options for individual `job <subcommand>`

components/hab/src/cli/hab/bldr.rs
agadgil-progress: nitpick: Also, this text is not exactly identical to one mentioned in the Jira issue. It should be using ***this feature*** and not ***these features***.

components/hab/src/cli/hab.rs
jasonheath: Went with prefixing 'REMOVED: '

components/hab/src/cli/hab/bldr.rs
jasonheath: As discussed in daily and now noted in commit leaving these so that any scripting done leveraging these commands fail with the failure message instead of "missing option".

components/hab/src/cli/hab/bldr.rs
jasonheath: mesg now matches Chaithra's.

components/hab/src/cli/hab/bldr.rs
agadgil-progress: I just tried locally something - and it works fine -  If we change 

```rust 
#[structopt(name = "status", no_version)]
``` 

to 

```rust
#[structopt(name = "demote", no_version, settings = &[AppSettings::ArgRequiredElseHelp])]
``` 

This avoids one round trip for user to get the `NOTICE: ...` 

I think it's a good idea to include that.

agadgil-progress (approved):

jasonheath:
So it's the CPP action that's failing on this.  Pushing over that since there's no CPP in our project.  Something we ought to fix at some point.

PR #9313, opened 2024-07-17
Upgrade Rust to 1.79.0

sougata-progress:
issue : #CHEF-13974

Updated Habitat to use 1.79.0

### DIFF
--- a/components/butterfly/src/rumor/dat_file.rs
+++ b/components/butterfly/src/rumor/dat_file.rs
@@ -96,7 +96,7 @@ impl DatFileReader {
         let size = OpenOptions::new().create(true)
                                      .read(true)
                                      .write(true)
-                                     .truncate(true)
+                                     .truncate(false)
                                      .open(&data_path)
                                      .map_err(|err| Error::DatFileIO(data_path.clone(), err))?
                                      .metadata()
--- a/components/butterfly/src/rumor/dat_file.rs
+++ b/components/butterfly/src/rumor/dat_file.rs
@@ -96,6 +96,7 @@ impl DatFileReader {
         let size = OpenOptions::new().create(true)
                                      .read(true)
                                      .write(true)
+                                     .truncate(true)
                                      .open(&data_path)
                                      .map_err(|err| Error::DatFileIO(data_path.clone(), err))?
                                      .metadata()
--- a/components/core/src/os/ffi/windows.rs
+++ b/components/core/src/os/ffi/windows.rs
@@ -7,17 +7,6 @@ pub trait OsStrExt3 {
     fn as_bytes(&self) -> &[u8];
 }
 
-pub trait OsStrExt2 {
-    fn starts_with(&self, s: &[u8]) -> bool;
-    fn split_at_byte(&self, b: u8) -> (&OsStr, &OsStr);
-    fn split_at(&self, i: usize) -> (&OsStr, &OsStr);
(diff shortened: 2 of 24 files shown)


mwrock (changes requested):
Your changes look good but there are some build failures tha tneed investigation (windows and rustfmt). Also please rename this PR to something like "Upgrade rust to 1.79.0" and you will need to make sure your commits are signed.

If you rebase on main, that should get rid of the cargo audit failure.

components/core/src/os/ffi/windows.rs
```diff
@@ -8,13 +8,27 @@ pub trait OsStrExt3 {
 }
 
 pub trait OsStrExt2 {
+    #[allow(dead_code)]
```
mwrock, line 11: I think you can actually remove `OsStrExt2`. I really does look like dead code that is truly dead.

components/core/src/package/install.rs
```diff
@@ -1280,17 +1280,28 @@ mod test {
...
-            fs::fs_rooted_path(
-                &pkg_prefix_for(&pkg_install).join("/should/really/not/be/ignored"),
-                &fs_root_path,
-            ),
-        ]).unwrap()
-                        .to_string_lossy()
-                        .into_owned(),
+            "PSModulePath".to_string(),
+            env::join_paths(vec![
+                fs::fs_rooted_path(
```
mwrock, line 1285: I was looking at this change and also clippy's complaint with the old code and the test failure you are seeing. I think clippy's suggestion about `Path::new` is the way to go:

```
        expected.insert(
                        "PSModulePath".to_string(),
                        env::join_paths(vec![
            fs::fs_rooted_path(&PathBuf::from("/should/not/be/ignored"), &fs_root_path),
            PathBuf::from("c:/my/dir"),
            fs::fs_rooted_path(
                &PathBuf::from("/should/really/not/be/ignored"),
                &fs_root_path,
            ),
        ]).unwrap()
                        .to_string_lossy()
                        .into_owned(),
        );
```

sougata-progress: okay sure ill make the changes

sougata-progress:
@mwrock can you please check why few of the tests are failing? im am not able to figure it out. Thank you

components/butterfly/src/rumor/dat_file.rs
```diff
@@ -96,6 +96,7 @@ impl DatFileReader {
         let size = OpenOptions::new().create(true)
                                      .read(true)
                                      .write(true)
+                                     .truncate(true)
```
mwrock, line 99: This should be `false` which will fix the rumor unit test. It looks like function checks the size of the file using this `create` call. Not entirely sure why this approach was used, but truncating it will obviously wipe it for subsequent reading.

components/sup/src/lock_file.rs
```diff
@@ -348,6 +348,7 @@ mod tests {
               C: AsRef<str>
     {
         let mut file = std::fs::OpenOptions::new().write(true)
+                                                  .truncate(true)
```
sougata-progress, line 351: @mwrock should the same apply here as well?

mwrock: that one looks ok. This is a helper function for tests and all of its callers expect an empty file.

sougata-progress: okay

sajjaphani (approved):

mwrock (approved):

PR #9322, opened 2024-07-22
Removed dependency on crate `env`

agadgil-progress:
`components/builder-api-client` had a dependency on crate `env` and this crate is possibly yanked from `crates.io`, this resulted in builds starting to fail.

Removed this dependency as this is an old dependency lying around and not used anywhere in the package.

Did a very basic `cargo build` and `cargo test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -904,12 +904,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env"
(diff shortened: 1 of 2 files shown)


sajjaphani (approved):
Looking good!

mwrock (approved):

PR #9325, opened 2024-07-23
Agadgil/e2e test failure container export

agadgil-progress:
Fixes issues in end to end test [failure](https://buildkite.com/chef/habitat-sh-habitat-main-end-to-end/builds/1607#0190dd06-e879-4083-938b-d2aa2550258e) 

1. Fix for `test_cointainer_export` e2e test
2. Fix for `test_external_binaries` e2e test

### DIFF
--- a/components/pkg-export-container/src/cli.rs
+++ b/components/pkg-export-container/src/cli.rs
@@ -28,11 +28,12 @@ pub fn cli() -> Command {
                                 .arg(Arg::new("IMAGE_NAME").long("image-name")
                                                            .short('i')
                                                            .value_name("IMAGE_NAME")
-                                                           .help("Image name template: supports: \
-                                                                  {{pkg_origin}}/{{pkg_name}} \
-                                                                  (default), {{pkg_origin}}, \
-                                                                  {{pkg_name}}, {{pkg_version}}, \
-                                                                  {{pkg_release}}, {{channel}})"));
+                                                           .help("Image name template: supports \
+                                                                  {{pkg_origin}}, {{pkg_name}}, \
+                                                                  {{pkg_version}}, \
+                                                                  {{pkg_release}}, {{channel}} \
+                                                                  variables. [default template: \
+                                                                  {{pkg_origin}}/{{pkg_name}}]"));
 
     let cmd = add_base_packages_args(cmd);
     let cmd = add_builder_args(cmd);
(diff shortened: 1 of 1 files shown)


components/pkg-export-container/src/cli.rs
```diff
@@ -31,10 +38,10 @@ pub fn cli() -> Command {
...
                                                      .help("Image name template: supports: \
                                                             {{pkg_origin}}/{{pkg_name}} \
-                                                            {{pkg_origin}}, {{pkg_name}}, \
-                                                            {{pkg_version}}, {{pkg_release}}, \
-                                                            {{channel}})")
-                                                     .default_value("{{pkg_origin/pkg_name}}"));
+                                                            (default), {{pkg_origin}}, \
+                                                            {{pkg_name}}, {{pkg_version}}, \
+                                                            {{pkg_release}}, {{channel}})")
+                                                     .value_parser(SUPPORTED_IMAGE_NAME_TEMPLATES));
```
mwrock, line 44: This validation looks incorrect. As I understand it, this enforces that the value of `--image-name` must be one of the values of `SUPPORTED_IMAGE_NAME_TEMPLATES`, which is not what we want. The template could contain any text and promises to treat each of the tokens in `SUPPORTED_IMAGE_NAME_TEMPLATES` as substitutable for their effective values. For example `{{pkg_name}}-MY_ORG` should be a valid template while this does not allow it.

agadgil-progress: Oh okay, I was referring to `help` of that option and thought only those are supported. Removed that validation now.

components/pkg-export-container/src/cli.rs
```diff
@@ -19,22 +19,20 @@ pub fn cli() -> Command {
...
+                                .version(VERSION)
+                                .author("\nAuthors: The Habitat Maintainers <humans@habitat.sh>")
+                                .help_template("{name} {version} {author-section} {about-section} \
+                                                \n{usage-heading} {usage}\n\n{all-args}")
+                                .arg(Arg::new("IMAGE_NAME").long("image-name")
+                                                           .short('i')
+                                                           .value_name("IMAGE_NAME")
+                                                           .help("Image name template: supports: \
+                                                                  {{pkg_origin}}/{{pkg_name}} \
+                                                                  (default), {{pkg_origin}}, \
```
mwrock, line 33: I think you can remove `{{pkg_origin}}/{{pkg_name}}` and `(default)`. It's just {{pkg_origin}}, {{pkg_name}}, {{pkg_version}}, {{pkg_release}}, and {{channel}} that we can substitute.

agadgil-progress: Updated the help text

mwrock (approved):

PR #9327, opened 2024-07-23
Updated channel from stable to LTS-2024

sougata-progress:
Issue CHEF-6569


sougata-progress:
Pausing dev on this pr as some of the hab packages are not available on the LTS channel for windows. will continue dev when the packages are available.

PR #9380, opened 2024-09-02
clap-v4 support for RST Reader

agadgil-progress:
Moved `rst-reader` binary to Clap V4. (CHEF-14728)

### DIFF
--- a/components/rst-reader/src/main.rs
+++ b/components/rst-reader/src/main.rs
@@ -1,6 +1,6 @@
 use crate::error::Result;
-use clap::{App,
-           Arg};
+use clap::Parser;
+
 use habitat_butterfly::rumor::{dat_file,
                                Departure,
                                Election,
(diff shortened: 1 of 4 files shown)


components/rst-reader/src/main.rs
```diff
@@ -12,33 +12,36 @@ use log::error;
...
+    stats: bool,
+}
+
+fn main() -> error::Result<()> {
     env_logger::init();
-    let matches =
-        App::new("Habitat RST Reader").about("Introspection for the butterfly RST file")
-                                      .arg(Arg::with_name("FILE").required(true)
-                                                                 .index(1)
-                                                                 .help("Path to the RST file"))
```
mwrock, line 23: looks like we lose this arg specific help and for STATS. Is there a way to retain that?

agadgil-progress: Oh!! Sorry had missed that! Fixed now.

mwrock (approved):

PR #9449, opened 2024-10-30
Stops setting permissions in ensure_path_permissions when we reach pkgs directory

jasonheath:
Stops setting permissions in ensure_path_permissions when we reach pkgs directory

An interim version of the code had some print stmts that show the path and the ancestors it walked up.  This show one of those runs where it stoped walking up the path hierarchy when it hit pkgs.

```
PATH: "/tmp/.tmpJ2NRUr/rootfs/hab/pkgs/core/linux-headers/4.20.17"
 ANC:"/tmp/.tmpJ2NRUr/rootfs/hab/pkgs/core/linux-headers/4.20.17"
 ANC:"/tmp/.tmpJ2NRUr/rootfs/hab/pkgs/core/linux-headers"
 ANC:"/tmp/.tmpJ2NRUr/rootfs/hab/pkgs/core"
 ANC:"/tmp/.tmpJ2NRUr/rootfs/hab/pkgs"
```

### DIFF
--- a/components/core/src/util/posix_perm.rs
+++ b/components/core/src/util/posix_perm.rs
@@ -69,7 +69,7 @@ pub(crate) fn ensure_path_permissions(path: &Path, permissions: u32) -> Result<(
     let euid = users::get_effective_uid();
     let egid = users::get_effective_gid();
     for ancestor in path.ancestors() {
-        if ancestor.ends_with("hab/pkgs") {
+        if ancestor.ends_with(crate::fs::PKG_PATH) {
             break;
         }
         if euid_egid_matches(&euid, &egid, ancestor) {
--- a/components/core/src/util/posix_perm.rs
+++ b/components/core/src/util/posix_perm.rs
@@ -69,6 +69,9 @@ pub(crate) fn ensure_path_permissions(path: &Path, permissions: u32) -> Result<(
     let euid = users::get_effective_uid();
     let egid = users::get_effective_gid();
     for ancestor in path.ancestors() {
+        if ancestor.ends_with("hab/pkgs") {
+            break;
+        }
         if euid_egid_matches(&euid, &egid, ancestor) {
             set_permissions(ancestor, permissions)?
         }


components/core/src/util/posix_perm.rs
```diff
@@ -69,6 +69,9 @@ pub(crate) fn ensure_path_permissions(path: &Path, permissions: u32) -> Result<(
     let euid = users::get_effective_uid();
     let egid = users::get_effective_gid();
     for ancestor in path.ancestors() {
+        if ancestor.ends_with("pkgs") {
```
mwrock, line 72: could we make this `ends_with("/pkgs")` in the event you have a package name or origin that just so happens to end with pkgs?

mwrock (approved):
Preemptively approving but please have a look at the comment and consider this small change.

components/core/src/util/posix_perm.rs
jasonheath: I didn't think '/pkgs' vs 'pkgs' would work and based on my testing it didn't.  However, 'hab/pkgs' does work and feels line with your suggestion.  That'll be be force pushed up here in a moment.

jasonheath: I don’t think the method works that way, see the examples/tests associated with [ends_with](https://doc.rust-lang.org/std/path/struct.Path.html#method.ends_with).  I think what you’re asking would require the 4th assertion `assert!(!path.ends_with("/resolv.conf”));` to drop the ! operator.

As for your concerns about changes to FS_ROOT I’ve gone through and traced the code and I don’t think that will be a problem.  To show that I’ve meaningfully gone through it I present the following. Laying it out here will hopefully save you time in reviewing and if I’m wrong hopefully there’s enough here that you can concretely point to where I’ve muffed things because this all tracks for me.

- Everything around [FS_ROOT](https://github.com/habitat-sh/habitat/blob/a56dd608ee879b09061fc85b4de7ec049d149ef7/components/core/src/fs.rs#L142) is really about where the system is or where we want to root everything when we’re doing a test.  So ‘/‘ for unix, “SYSTEMDRIVE” for windows, or from what I’ve seen some arbitrary location in the system’s tmp.
- From there the code is always going to partition itself under [ROOT_PATH](https://github.com/habitat-sh/habitat/blob/a56dd608ee879b09061fc85b4de7ec049d149ef7/components/core/src/fs.rs#L34) which is a compile time constant of ‘hab’.
- Similarly, there a number of other constants of the forms “hab/xxx” and one of those is [PKG_PATH](https://github.com/habitat-sh/habitat/blob/a56dd608ee879b09061fc85b4de7ec049d149ef7/components/core/src/fs.rs#L55) with the unix value being ‘hab/pkgs’ and the windows value being ‘hab\pkgs’.  
    - Even though windows isn’t a concern for us in this code, this is really what I should have used instead of the magic str I used so I’ve made that change.
- I think that these “hab-centric” constants show that we don’t have to worry about hab changing arbitrarily based on FS_ROOT or any other environment variable.  

So, given the documentation around the ends_with method and what I present above about FS_ROOT/PKG_PATH I feel like this is a trustworthy change.

jasonheath:
Capturing some debug prints in case I wish I had them later.

```shell
☛ Verifying core/hab/1.6.1041/20240411141804
path: "/tmp/.tmphCoP04/rootfs/hab/pkgs/core/hab/1.6.1041/20240411141804"
base: "/tmp/.tmphCoP04/rootfs/hab/pkgs/core/hab/1.6.1041"
FS_ROOT_PATH: "/"
PKG_PATH: "hab/pkgs"
PATH: "/tmp/.tmphCoP04/rootfs/hab/pkgs/core/hab/1.6.1041"
 ANC: "/tmp/.tmphCoP04/rootfs/hab/pkgs/core/hab/1.6.1041"
 ANC: "/tmp/.tmphCoP04/rootfs/hab/pkgs/core/hab"
 ANC: "/tmp/.tmphCoP04/rootfs/hab/pkgs/core"
 ANC: "/tmp/.tmphCoP04/rootfs/hab/pkgs"
```

mwrock (approved):

PR #9481, opened 2024-11-21
fixes for release pipeline (aarch64)

sajjaphani:
Currently, the release pipeline is failing because it is falling back to the LTS-2024 channel, where none of the `hab*` packages exist. To unblock the release pipeline, the fallback channel has been updated to `stable`.

Note: The hab, hab-studio, hab-backline, and hab-plan-build packages with version ~~1.6.1178~~ **1.6.1218** have been promoted to the stable channel.

### DIFF
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -101,6 +101,9 @@ finish_setup() {
   _hab pkg binlink core/bash bash
   _hab pkg binlink core/bash sh
 
+  # binlink sleep command
+  _hab pkg binlink "$coreutils_path/bin/sleep" sleep
+
   # Create a wrapper to `build` so that any calls to it have a super-stripped
   # `$PATH` and not whatever augmented version is currently in use. This should
   # mean that running `build` from inside a `studio enter` and running `studio
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -101,9 +101,6 @@ finish_setup() {
   _hab pkg binlink core/bash bash
   _hab pkg binlink core/bash sh
 
-  # binlink sleep command
-  _hab pkg binlink core/coreutils sleep
-
   # Create a wrapper to `build` so that any calls to it have a super-stripped
   # `$PATH` and not whatever augmented version is currently in use. This should
   # mean that running `build` from inside a `studio enter` and running `studio
--- a/components/studio/libexec/hab-studio-type-default.sh
+++ b/components/studio/libexec/hab-studio-type-default.sh
@@ -102,7 +102,7 @@ finish_setup() {
   _hab pkg binlink core/bash sh
 
   # binlink sleep command
-  _hab pkg binlink "$coreutils_path/bin/sleep" sleep
+  _hab pkg binlink core/coreutils sleep
 
   # Create a wrapper to `build` so that any calls to it have a super-stripped
   # `$PATH` and not whatever augmented version is currently in use. This should
--- a/.expeditor/release_habitat.pipeline.yml
+++ b/.expeditor/release_habitat.pipeline.yml
@@ -118,7 +118,7 @@ steps:
     agents:
       queue: default-privileged-aarch64
     env:
-      HAB_STUDIO_SECRET_HAB_FALLBACK_CHANNEL: "stable"
+      HAB_STUDIO_SECRET_HAB_FALLBACK_CHANNEL: "LTS-2024"
       BUILD_PKG_TARGET: "aarch64-linux"
     command:
       - sudo -E .expeditor/scripts/release_habitat/build_component.sh plan-build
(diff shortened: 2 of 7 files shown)


mwrock:
You actually want the release channel to fall back t oLTS-2024. The "primary" HAB_BLDR_CHANNEL should be a dynamically created channel based on the `BUILDKITE_BUILD_ID`. As components are built in the pipeline, the they are uploaded to this channel. So core/hab is built and uploaded to that channel and then core/hab-plan-build should get hab from that channel.

I'm looking at https://buildkite.com/chef/habitat-sh-habitat-main-release-habitat/builds/2002 where the release pipeline is failing. For some reason hab-plan-build is not honoring `HAB_BLDR_CHANNEL`, but I have not looked deeply into this. Anyways, I think that is the problem you want to investigate. If you fall back to stable, then all of the deps will come from stable instead of LTS-2024.

sajjaphani:
For aarch64, we have core packages in the LTS-2024 channel, and the hab* packages built against them are placed in the stable channel.
We introduced the `HAB_REFRESH_CHANNEL` to pull core dependencies, and the aarch64 packages in the stable channel are built with these changes.

https://github.com/habitat-sh/habitat/blob/9ef672b40e323cc45874f3ca93d39586cd849a84/components/plan-build/bin/hab-plan-build.sh#L691-L698

It appears that with the introduction of HAB_REFRESH_CHANNEL, the HAB_BLDR_CHANNEL is no longer honored for core packages.

For `aarch64`, the `HAB_REFRESH_CHANNEL` is set to `LTS-2024` in both the verify and release pipelines to pull dependencies, while the `HAB_FALLBACK_CHANNEL` is configured to pull the **hab**, **hab-studio**, and similar packages from the stable channel. No other packages, except for hab, hab-studio, and related ones, are present in the stable channel.

mwrock:
I ran a test release build from this branch and the built components used in the front of the pipeline are not getting consumed by their dependent components. This can be seen in https://buildkite.com/chef/habitat-sh-habitat-main-release-habitat/builds/2009#019397ab-5b97-486b-8433-7dc6b7bfb6c1/429-693. Here the core/hab built at the beginning of the pipeline is `core/hab/1.6.1218/20241205163418` and was promoted to `habitat-release-019397aa-b2e3-45a5-8ecc-893aa8c7c0ba`. However, the hab-plan-build component is pulling `core/hab/1.6.1218/20241127073909` from `stable`.

I think what you need to do to fix this is set `HAB_STUDIO_SECRET_HAB_REFRESH_CHANNEL` to the release channel in build_component.sh. For the non-hab components, it should pull from the fallback of LTS-2024.

sajjaphani:
If we set HAB_STUDIO_SECRET_HAB_REFRESH_CHANNEL to the release channel with a fallback to LTS-2024, it will only fetch the latest hab binary from the release channel. This raises the question of where hab-studio is being sourced from, as it is essential for building any component.

mwrock:
You want the core/hab that builds with the down stream components to be pulled from the release channel. Otherwise, plan-build and the exporters won't include the correct core/hab. The core/hab and studio used to actually do the build will also come from the release channel if they have both been build, but I'm pretty sure that in the earlier stages of the pipeline, it just uses the latest stable hab.

At any rate, in the end, we have to make sure that the release pipeline produces artifacts that all reflect  the components that were built off of main. Currently it doesn't look like that is happening here.

sajjaphani:
Updated to point the hab refresh channel to the release channel and set the fallback to LTS-2024 for aarch64.
https://buildkite.com/chef/habitat-sh-habitat-main-release-habitat/builds/2010#0193ab42-8820-4621-baa7-5dd69c954712

components/studio/libexec/hab-studio-type-default.sh
```diff
@@ -101,6 +101,9 @@ finish_setup() {
   _hab pkg binlink core/bash bash
   _hab pkg binlink core/bash sh
 
+  # binlink sleep command
+  _hab pkg binlink "$coreutils_path/bin/sleep" sleep
```
mwrock, line 105: I think this should be `_hab pkg binlink core/coreutils sleep`

mwrock:
@sajjaphani still seeing issues with `sleep` in https://buildkite.com/chef/habitat-sh-habitat-main-verify/builds/13599#0193b0e4-1d0a-417c-aff3-f822ec666930/194-787

sajjaphani:
> @sajjaphani still seeing issues with `sleep` in https://buildkite.com/chef/habitat-sh-habitat-main-verify/builds/13599#0193b0e4-1d0a-417c-aff3-f822ec666930/194-787

This is happening because the older studio does not have sleep symlinked. I’ve rebuilt and uploaded the package, and it is now passing the pipeline.

mwrock:
In the release pipeline, I see it built the studio fine and then when it builds the launcher, it uses that studio and I see it binlinking `sleep` (https://buildkite.com/chef/habitat-sh-habitat-main-release-habitat/builds/2015#0193b66e-cab1-467f-87b2-18dfadde2b04/297-417) but then it fails at the end with the same old `sleep: No such file or directory` error (https://buildkite.com/chef/habitat-sh-habitat-main-release-habitat/builds/2015#0193b66e-cab1-467f-87b2-18dfadde2b04/297-1640).

mwrock (approved):

PR #9611, opened 2025-03-04
downgrade native-tls

mwrock:
The upgrade to native-tls 0.2.13, breaks arm when reqwest tries to build a request. I'm can't tell exactly what broke from the native-tls changelog and the error returned on arm from a `hab pkg install` which will produce the error is completely useless. It just returns "builder error". It's inner error array is empty.

There is a native-tls 0.2.14 we could try but it requires at least rust v 1.80. We need to get that available, upgrade rust and give 0.2.14 a try but for now downgrading fixes arm.

### DIFF
--- a/.expeditor/scripts/shared.sh
+++ b/.expeditor/scripts/shared.sh
@@ -48,6 +48,17 @@ install_rustup() {
   fi
 }
 
+install_rust_toolchain() {
+  local toolchain="${1?toolchain argument required}"
+
+  if rustup component list --toolchain "$toolchain" &>/dev/null; then
+    echo "--- :rust: Rust $toolchain is already installed."
+  else
+    echo "--- :rust: Installing rust $toolchain."
+    rustup toolchain install "$toolchain"
+  fi
+}
+
 get_toolchain() {
     dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
     tail -n 1 "$dir/../../rust-toolchain" | cut -d'"' -f 2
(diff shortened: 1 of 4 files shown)


.expeditor/scripts/shared.sh
```diff
@@ -48,6 +48,17 @@ install_rustup() {
   fi
 }
 
+install_rust_toolchain() {
```
agadgil-progress, line 51: Is this explicitly required? I believe if we have `rustup`, then the toolchain argument in the `rust-toolchain` file is automatically honoured? 

Not in the scope of current issue, but I believe it's probably a good idea to specify `minimal` profile and explicitly add components `clippy` and `rustfmt` like here - https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

agadgil-progress: Oh I see - you just moved this from `verify/shared` to `shared`

.expeditor/scripts/verify/build_mac_package.sh
```diff
@@ -31,11 +31,13 @@ macos_sync_cache_signing_keys
...
 install_rustup
 
+rust_toolchain="$(tail -n 1 rust-toolchain  | cut -d'"' -f 2)"
 if [ "$BUILD_PKG_TARGET" == "aarch64-darwin" ]; then
     rustup target add aarch64-apple-darwin
+    rust_toolchain="${rust_toolchain}-aarch64-apple-darwin"
 fi
 
-rust_toolchain="$(tail -n 1 rust-toolchain  | cut -d'"' -f 2)"
+install_rust_toolchain "$rust_toolchain"
```
agadgil-progress, line 40: Is this explicitly required?

mwrock: It never has been before but in in the last few days, the mac builds are breaking on this and it looks like it corresponds to a change in rustup so I'm just trying to experiment and get it to work.

PR #9644, opened 2025-03-25
Fixes failure when `NO_INSTALL_DEPS` is set

agadgil-progress:


mwrock (approved):
nice catch. I imagine we will need to make the same fix for windows too. Please address the lint errors before merge.

PR #9645, opened 2025-03-25
CHEF-19259: perm issues with services

jasonheath:
This PR implements change to ensure that the following directories have permissions of 755 on systems when they are hardened such that a umask 077 is in effect.
- /hab/svc 
- /hab/svc/hooks
- /hab/svc/logs

In addition,
- test/end-to-end/test_supervisor_load_service.ps1 is now used to test both Linux and windows
- test/end-to-end/hab-svc-load.exp has been removed
- .expeditor/scripts/verify/run_psscriptanalyzer.ps1 was updated to output version information.

Justifications 
- test_supervisor_load_service.ps1 executes equivalents tests as in hab-svc-load.exp but also goes a bit further.
- While expect is venerable its a small rounding error of our testing. Switching adheres to our team decision to focus on PowerShell based testing for both platforms and helps to reduce cognitive load in terms of the things we all need to know.
- I was debugging PSScriptAnalyzer issues and wanted to know the version so I added it.  The answer surprised me so I kept it as it might be helpful in the future.


agadgil-progress (approved):

agadgil-progress:
I only had one comment - if at all (not super important) - the check 

```rust
    ... 
     if ancestor.ends_with(crate::fs::PKG_PATH) {
            break;
    }
    ...
```
kind of becomes redundant with new usage.

jasonheath:
> I only had one comment - if at all (not super important) - the check
> 
> ```rust
>     ... 
>      if ancestor.ends_with(crate::fs::PKG_PATH) {
>             break;
>     }
>     ...
> ```
> 
> kind of becomes redundant with new usage.

Removed.

mwrock (approved):

PR #9658, opened 2025-04-04
Corrects 2s timeout to 20s in test_supervisor_load_service.ps1

jasonheath:
This PR fixes an issues with test_supervisor_load_service where the wait for nginx was 2s instead of the 20s used elsewhere.

After submitting the PR I hit [a RUSTSEC advisory that came out today](https://rustsec.org/advisories/RUSTSEC-2025-0022) so this also fixes that.


jasonheath:
Awesome.  I hit the RUSTSEC lottery.  https://rustsec.org/advisories/RUSTSEC-2025-0022

mwrock (approved):

PR #9675, opened 2025-04-14
Calls release_habitat/package_and_upload_binary.sh when branch == main

jasonheath:
This PR modifies the release_habitat pipeline such that package_and_upload_binary.sh will run whenever the build is being executed against the main branch.  This is an added OR'ed condition to the logic that previously existed.

The motivating issue is that the release_habitat pipeline is a bit fragile and often fails.  One may argue that the real fix would be "fix the pipeline" which is accurate but that's likely to be quite an undertaking, this is fast.  Also, fixing could mean higher costs. For example, bigger and more instances with more processing power would likely help a lot of this fragility because the failure are often related to builds timing out.

However, beyond that there is motivation in that buildkite UI has two flaws IMO.  

First, when the if conditional in the yaml is not met there is a very subtle indicator that steps aren't being run which is the addition of an eye on the right hand side.  In addition, the steps not being run are hidden  It's just so easy to overlook that indicator also to not realize that steps won't be run when you're busy and keeping an eye on the pipeline executing off to the side.

Beyond that, clicking the rebuild button which feels like it should do what you want really doesn't in this case.  The executing user now matches who clicks the button and you don't have the ability to specify the BUILD_AND_PROMOTE environment variable.  To do that you have to click new build, set the envvar, set a build name which you should probably copy and paste it forward.

In short, there's just friction and sharp edges and ruunning package_and_upload_binary.sh when the branch is main should honor the goals that brought about the addition of the if stmt.

### DIFF
--- a/.expeditor/release_habitat.pipeline.yml
+++ b/.expeditor/release_habitat.pipeline.yml
@@ -326,7 +326,7 @@ steps:
   - label: "[:linux: upload hab binary]"
     command:
       - .expeditor/scripts/release_habitat/package_and_upload_binary.sh
-    if: build.creator.name == 'Chef Expeditor' || build.env("UPLOAD_AND_PROMOTE") == 'true'
+    if: build.creator.name == 'Chef Expeditor' || build.env("UPLOAD_AND_PROMOTE") == 'true' || build.branch == 'main'
     expeditor:
       executor:
         docker:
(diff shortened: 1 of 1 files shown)


jasonheath:
> I was just thinking more about this and wondering would it be possible for `build.branch` to me `main` but the commit not to be at the HEAD? In other words, if I were to go to some old build and hit the "rebuild" button, would it be building from the `main` branch but an old commit? I wouldn't want that to get uploaded.

Kind of a edge case but an awesome call out.  At the moment, don't know.  AFAIK the list of toys we have to play with is [here](https://buildkite.com/docs/pipelines/configure/conditionals#variable-and-syntax-reference-variables).  There is a `build.commit` but that talks about a build number so I don't know that HEAD would work but if it doesn't then maybe something from the different kinds of [revision syntax](https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emem) would?

mwrock:
My guess is that be build.commit would mirror the BUILDKITE_COMMIT variable which would be a specific commit in the branch.

jasonheath:
I still think this or a variant is a good idea but I'm closing it for now. May reopen it or create a new PR in the future.

PR #9712, opened 2025-05-29
add --pretty to cargo tests

mwrock:


mwrock:
this was just to get buildkite to print out the tests. don't need permanently.

PR #9717, opened 2025-06-05
Using `base` channel as default channel for packages in `core` origin

agadgil-progress:
This PR adds support for `base` channel as default channel for the packages in the `core` origin. `stable` remains default channel for packages in other origins - 

1. Major changes in the `download` and `install` packages sub command
2. `HAB_REFRESH_CHANNEL` default is set to `base`
3. Updates in supervisor to use appropriate default channel based on the packet origin

As the `hab` and related packages are moving to `chef` origin, some other parts are not updated (notably `pkg-export-*`)


components/hab/src/command/pkg/promote.rs
```diff
@@ -37,11 +37,19 @@ pub async fn start(ui: &mut UI,
...
     ui.begin(format!("Promoting {} ({}) to channel '{}'", ident, target, channel))?;
 
-    if channel != &ChannelIdent::stable() && channel != &ChannelIdent::unstable() {
+    let default_channel_for_ident = if ident.origin() == "core" {
+        channel == &ChannelIdent::base()
+    } else {
+        channel == &ChannelIdent::stable()
+    };
+
+    if !default_channel_for_ident && channel != &ChannelIdent::unstable() {
```
agadgil-progress, line 52: Wondering whether we should just have something like `reserved_channels` and let people not create those channels than the logic here.

components/hab/src/command/pkg/upload.rs
```diff
@@ -197,9 +197,16 @@ async fn promote_to_channel(ui: &mut UI,
...
     ui.begin(format!("Promoting {} to channel '{}'", ident, channel))?;
 
-    if channel != ChannelIdent::stable() && channel != ChannelIdent::unstable() {
+    let default_channel_for_ident = if ident.origin() == "core" {
+        channel == ChannelIdent::base()
+    } else {
+        channel == ChannelIdent::stable()
+    };
+
+    if !default_channel_for_ident && channel != ChannelIdent::unstable() {
```
agadgil-progress, line 209: Same as above.

sajjaphani:
> This PR adds support for `base` channel as default channel for the packages in the `core` origin. `stable` remains default channel for packages in other origins -
> 
> 1. Major changes in the `download` and `install` packages sub command
> 2. `HAB_REFRESH_CHANNEL` default is set to `base`
> 3. Updates in supervisor to use appropriate default channel based on the packet origin
> 
> As the `hab` and related packages are moving to `chef` origin, some other parts are not updated (notably `pkg-export-*`)
Moving to chef origin is now merged, Can you consider updating pkg-export-* from using the same base channel?

components/hab/src/cli_v4/pkg/download.rs
```diff
@@ -75,6 +76,8 @@ pub(crate) struct PkgDownloadOptions {
 
 impl PkgDownloadOptions {
     pub(super) async fn do_download(&self, ui: &mut UI) -> HabResult<()> {
+        use habitat_core::package::Identifiable;
```
jasonheath: This is the one item I noticed while reviewing the code and it's the nit pickiest of nits that you should feel free to ignore.  You embed a use stmt for this in a number of functions and it's usually the sole instance of the use stmt.  In this file you have it twice and in another you have 3 use stmts.  My personal inclination is to add all use stmts at the top and I think that's generally what we do in the code.  However, I get the arguments for putting inline so my reason for mentioning it is that I noticed in main.rs you added the include at the top.I guess my thought would be to maybe eliminate the duplication here and in main_v2.rs and leave the rest alone but then it's kind of inconsistent with the practice every else.  I'd personally DRY them all up top but if you've got something you're reaching for, keep 'em.

components/hab/src/command/pkg/promote.rs
jasonheath: Since we've talked this morning and I've looked at the code I think I get what you're after here but after thinking about I think we should create a work item and do this seperately.  Mostly I'm thinking it would be good to get @mwrock's and @sajjaphani's input on the list of channel names.

jasonheath:
Address @sajjaphani's comment and I think this can be approved.

jasonheath:
Address @sajjaphani's [comment](https://github.com/habitat-sh/habitat/pull/9717#issuecomment-3073466680) and I think this can be approved.

components/hab/src/cli_v4/pkg/download.rs
agadgil-progress: Now I know why I did that - `Identifiable` is a `trait` that needs to be in the scope. This was one way of saying only these functions use that `trait`.

agadgil-progress:
> > This PR adds support for `base` channel as default channel for the packages in the `core` origin. `stable` remains default channel for packages in other origins -
> > 
> > 1. Major changes in the `download` and `install` packages sub command
> > 2. `HAB_REFRESH_CHANNEL` default is set to `base`
> > 3. Updates in supervisor to use appropriate default channel based on the packet origin
> > 
> > As the `hab` and related packages are moving to `chef` origin, some other parts are not updated (notably `pkg-export-*`)
> > Moving to chef origin is now merged, Can you consider updating pkg-export-* from using the same base channel?

I checked the plan files for both `pkg-export-container` and `pkg-export-tar` for this and they seem to contain the `chef` origin. Also did a quick `grep` on the codebase for `core`. Couldn't find it trivially. Are there places where I should look at in more details that you remember off-hand?

sajjaphani:
@agadgil-progress  The base packages (hab, sup, and launcher) are added from the `stable` channel in `export*`. You can find references to BASE_PKGS_CHANNEL for this configuration.

agadgil-progress:
> @agadgil-progress The base packages (hab, sup, and launcher) are added from the `stable` channel in `export*`. You can find references to BASE_PKGS_CHANNEL for this configuration.

`BASE_PKGS_CHANNEL` default is `stable` which it will continue to remain as the `hab` and related packages are moved to `chef` origin and thus have `stable` as default channel?

PR #9718, opened 2025-06-05
Move habitat components to chef origin

sajjaphani:
This PR moves the components from the core origin to the chef origin for:

- Linux (x86_64 and aarch64)
- Windows (x86_64)

A few tests are being skipped, as we need to revisit them during the release process.

**Note:** I think we should continue to default to the `core` origin in `install.sh` until we're ready for a release. Otherwise, consumers of `install.sh` may experience failures.


.expeditor/scripts/shared.ps1
```diff
@@ -1,3 +1,6 @@
+# Set the Habitat channel to install chef/hab from.
+$HabChannel = "dev"
```
mwrock: Instead of this global variable here, you change `function Install-Habitat` below to `function InstallHabitat($HabChannel = "dev")` This is basically the equivilent to what you have in the bash version where you have an optional argumant that defaults to `dev`.

mwrock: On a separate note, I'm not sure I would default to `dev`. `acceptance` might be better wince that would have passed e2e tests.

.expeditor/scripts/verify/build_package-aarch64.sh
```diff
@@ -19,5 +19,11 @@ HAB_CACHE_KEY_PATH="$JOB_TEMP_ROOT/keys"
 
 echo "--- :key: Generating temporary origin key"
 hab origin key generate "$HAB_ORIGIN"
+
+# Install hab-studio from the chef origin via the dev channel.
+# By default, it installs from the stable channel only,
+# so this may need updating to support other channels.
+hab pkg install chef/hab-studio -c dev
```
mwrock: You can `export HAB_INTERNAL_BLDR_CHANNEL=dev` ind `hab pkg build` will pull the studio from dev.

.expeditor/scripts/verify/build_package.ps1
```diff
@@ -19,6 +19,12 @@ $env:HAB_CACHE_KEY_PATH="$job_temp_root/keys"
 
 Write-Host "--- :key: Generating fake origin key"
 hab origin key generate
+
+# Install hab-studio from the chef origin via the dev channel.
+# By default, it installs from the stable channel only,
+# so this may need updating to support other channels.
+hab pkg install chef/hab-studio -c dev
+
```
mwrock, line 27: See comment above but use `$env:HAB_INTERNAL_BLDR_CHANNEL="staging"`

.expeditor/scripts/verify/build_package.sh
```diff
@@ -21,5 +21,11 @@ HAB_CACHE_KEY_PATH="$JOB_TEMP_ROOT/keys"
 
 echo "--- :key: Generating fake origin key"
 ${hab_binary} origin key generate
+
+# Install hab-studio from the chef origin via the dev channel.
+# By default, it installs from the stable channel only,
+# so this may need updating to support other channels.
+hab pkg install chef/hab-studio -c dev
```
mwrock: see above comment

.expeditor/verify.pipeline.yml
```diff
@@ -10,6 +10,10 @@ expeditor:
         HAB_BLDR_CHANNEL: "base-2025"
         HAB_STUDIO_SECRET_HAB_REFRESH_CHANNEL: "base-2025"
         HAB_REFRESH_CHANNEL: "base-2025"
+        HAB_FALLBACK_CHANNEL: "dev"
+        HAB_STUDIO_SECRET_HAB_FALLBACK_CHANNEL: "dev"
+        BUILD_PKG_ORIGIN: "chef"
+        BUILD_PKG_CHANNEL: "dev"
```
mwrock: The variable names `BUILD_PKG_ORIGIN` and `BUILD_PKG_CHANNEL` feel confusing to me. Could we make this something like `HAB_CLI_ORIGIN` and `HAB_CLI_CHANNEL`? Or just use the values explicitly when calling the install script instead of using these variables.

mwrock (commented):
I'd consider using `acceptance` instead of `dev` because it would be too easy for a bad binary to land in dev and then break future builds.

Also, I did a search for `core/hab*` in this branch and found several references that need changing. Some examples include the release.md, the e2e multi-supervisor Dockerfile, the rootless studio dockerfile, the aarch64 part of the install.sh, etc.

.expeditor/scripts/shared.ps1
sajjaphani: Update channel from 'dev' to 'acceptance' in all places

.expeditor/scripts/verify/build_package.sh
sajjaphani: After updating with export HAB_INTERNAL_BLDR_CHANNEL=acceptance, the build is failing:https://buildkite.com/chef/habitat-sh-habitat-main-verify/builds/14464#0197cb12-51d5-4e28-b7af-c62edbb44811

.expeditor/verify.pipeline.yml
sajjaphani: Removed both environment variables from here.

.expeditor/scripts/verify/build_package.sh
mwrock: The problem was you needed to remove the `hab pkg install chef/hab-studio` because what `HAB_INTERNAL_BLDR_CHANNEL` does is make it so that `hab pkg build` will use that channel when it sees that it needs to download the studio first.

While that solved the error you were seeing, it created a new problem. Apparently when a hab cli command like hab studio or hab sup or hab export determines that it needs to download another hab binary, it does not honor the `HAB_AUTH_TOKEN` and thus some of the core deps in base-2025 were not being downloaded.

I will revert this to the code you originally had but this looks like a bug we will need to fix in hab 2.0.

mwrock (approved):

PR #9744, opened 2025-07-03
Updated hab config to use clap_v4

sougata-progress:
[CHEF-23350](https://progresssoftware.atlassian.net/browse/CHEF-23350)

[CHEF-23350]: https://progresssoftware.atlassian.net/browse/CHEF-23350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/hab/src/cli_v4/config/apply.rs
+++ b/components/hab/src/cli_v4/config/apply.rs
@@ -6,7 +6,6 @@ use clap::Parser;
 use clap_v4 as clap;
 use habitat_common::{cli::clap_validators::FileExistsOrStdinValueParser,
                      ui::UI};
-use std::path::PathBuf;
 
 #[derive(Debug, Clone, Parser)]
 #[command(arg_required_else_help = true,
@@ -38,8 +37,8 @@ pub(crate) struct ConfigApplyOptions {
     config_version: u64,
 
     /// Path to local file on disk (ex: /tmp/config.toml, "-" for stdin)
-    #[arg(value_parser = FileExistsOrStdinValueParser, value_name = "FILE")]
-    file: PathBuf,
+    #[arg(value_parser = FileExistsOrStdinValueParser, value_name = "FILE", default_value = "-")]
+    file: String,
 }
 
 impl ConfigApplyOptions {
--- a/components/hab/src/command/config.rs
+++ b/components/hab/src/command/config.rs
@@ -17,13 +17,12 @@ use std::{convert::TryFrom,
           fs::File,
           io::{self,
                Read},
-          path::{Path,
-                 PathBuf},
+          path::PathBuf,
           process};
 
 pub(crate) async fn sub_svc_set<U>(ui: &mut U,
                                    grp: ServiceGroup,
-                                   cfg_path: &Path,
+                                   cfg_path: &String,
                                    version: u64,
                                    user_opt: Option<String>,
                                    remote_sup: Option<ResolvedListenCtlAddr>,
@@ -32,7 +31,7 @@ pub(crate) async fn sub_svc_set<U>(ui: &mut U,
     where U: UIWriter
 {
     let mut buf = Vec::with_capacity(MAX_SVC_CFG_SIZE);
-    let len = if cfg_path.as_os_str() == "-" {
+    let len = if cfg_path.as_str() == "-" {
         io::stdin().read_to_end(&mut buf)?
     } else {
         let mut f = File::open(cfg_path)?;


components/hab/src/cli_v4.rs
```diff
@@ -9,7 +9,10 @@ use crate::{error::Result as HabResult,
             AFTER_HELP,
             VERSION};
 
+pub mod config;
```
agadgil-progress: Is `pub` mod required? The best is to start with a private module and then if required, go for `pub(crate)` very rarely `pub` access is required.

components/hab/src/cli_v4.rs
```diff
@@ -36,6 +39,7 @@ enum Hab {
     /// Commands relating to Habitat runtime config
     Cli(CliCommand),
 
+    #[clap(subcommand)]
```
agadgil-progress, line 43: Add a doctoring similar to `Cli(CliCommand)` above or `Pkg(PkgCommand)`

components/hab/src/cli_v4/config/mod.rs
```diff
@@ -0,0 +1,29 @@
+use clap_v4 as clap;
```
agadgil-progress: Though this is not wrong per se, we have been keeping a practice of `cli_v4/config.rs` for the `config` module and `cli_v4/config/apply.rs` etc. for the submodule. Better to stick with that practice. so you may want to move `config/mod.rs` to `config.rs`.

components/hab/src/cli_v4/config/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+use clap::Subcommand;
+use habitat_common::{ui::UI};
+use crate::error::Result as HabResult;
+
+mod apply;
+mod show;
+
+pub use apply::ConfigApplyOptions;
```
agadgil-progress: no `pub` qualifier`.

components/hab/src/cli_v4/config/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+pub use show::ConfigShowOptions;
```
agadgil-progress: no `pub` qualifier.

components/hab/src/cli_v4/config/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[derive(Debug, Clone, Subcommand)]
```
agadgil-progress: Apply `docstrings` here to individual enum variants. They become available as help strings.

components/hab/src/cli_v4/config/show.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[arg(long)]
+    pub remote_sup: Option<String>,
+
+    #[arg(long)]
+    pub ident: String,
+}
+
+impl ConfigShowOptions {
+    pub async fn do_show(&self) -> HabResult<()> {
+        let raw_args: Vec<_> = std::env::args_os().skip(2).collect();
```
agadgil-progress: You do not have to use `get_matches_from` again - all the hard work of parsing is done by `cli::parse` the values should be available as `self.remote_sup` and `self.ident` options.

components/hab/src/cli_v4/config/show.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Debug, Clone, Parser)]
+#[command(
+    arg_required_else_help = true,
+    rename_all = "kebab-case",
+    about = "Show the current config of a running service"
+)]
+pub struct ConfigShowOptions {
+    #[arg(long)]
+    pub remote_sup: Option<String>,
```
agadgil-progress: Also here - rather than using `String`, you can take the semantic `RemoteSup` from `cli_v4/utils.rs` see how it is done in `sup/status.rs` Something like 

```rust 
    /// Remote supervisor connection option
    #[command(flatten)]
    remote_sup: RemoteSup,
```

components/hab/src/cli_v4/config/show.rs
agadgil-progress: with `clap_v4`'s `cli::parse` command line properly parsed, you almost never have to use `matches` API, this reduces a lot of code that needs to be written.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::{error::Result as HabResult, command::config::sub_svc_set};
+
+
+#[derive(Debug, Clone, Parser)]
+#[command(
+    arg_required_else_help = true,
+    rename_all = "kebab-case",
+    about = "Apply a configuration to a running service"
+)]
+pub struct ConfigApplyOptions {
```
agadgil-progress: Add proper `docstring`s here.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,60 @@
...
+use habitat_common::{ui::UI, types::ResolvedListenCtlAddr};
+use habitat_core::service::ServiceGroup;
+use crate::{error::Result as HabResult, command::config::sub_svc_set};
+
+
+#[derive(Debug, Clone, Parser)]
+#[command(
+    arg_required_else_help = true,
+    rename_all = "kebab-case",
+    about = "Apply a configuration to a running service"
```
agadgil-progress: Also use the `help_template` attribute (This comment may apply in general to *all* structures). See for example `pkg/install.rs` or any `pkg`/`sup` subcommands.

components/hab/src/cli_v4/config.rs
```diff
@@ -0,0 +1,31 @@
...
+use habitat_common::{ui::UI};
+use crate::error::Result as HabResult;
+
+mod apply;
+mod show;
+
+use apply::ConfigApplyOptions;
+use show::ConfigShowOptions;
+
+#[derive(Debug, Clone, Subcommand)]
```
agadgil-progress, line 12: you will need to add the clap header similar to `cli_v4/pkg.rs` like below - 

```rust
#[command(author = "\nThe Habitat Maintainers <humans@habitat.sh>",
          arg_required_else_help = true,
          help_template = "{name} {version} {author-section} {about-section} \n{usage-heading} \
                           {usage}\n\n{all-args}\n")]
```

So the usage shows properly.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Service group identifier, e.g. `core/redis.default`
+    #[arg(long)]
+    pub group: String,
+
+    /// Path to the config file ("-" for stdin)
+    #[arg(long)]
+    pub file: PathBuf,
+
+    /// Configuration version number to set
+    #[arg(long, default_value_t = 0)]
```
agadgil-progress: you may need to rename the attribute to `config_version` (~and use `long = "version"` in the field attribute~). When I run the following command I get a panic 

```rust
$ ./target/debug/hab config apply --help

thread 'main' panicked at /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.29/src/builder/debug_asserts.rs:86:13:
Command apply: Argument names must be unique, but 'version' is in use by more than one argument or group (call `cmd.disable_version_flag(true)` to remove the auto-generated `--version`)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This might be related to the `version` field being present on both top level cli structure and here.

components/hab/src/command.rs
```diff
@@ -10,3 +10,4 @@ pub mod studio;
 pub mod sup;
 pub mod supportbundle;
 pub mod user;
+pub mod config;
```
agadgil-progress: `pub` really required here? `pub(crate)` should work no?

components/hab/src/command/config.rs
```diff
@@ -1 +1,165 @@
...
+    service::ServiceGroup,
+    package::PackageIdent,
+    crypto::keys::{KeyCache, Key},
+};
+use habitat_sup_client::{SrvClient, SrvClientError};
+use habitat_sup_protocol::{self as sup_proto, butterfly::MAX_SVC_CFG_SIZE, net::ErrCode};
+use futures::StreamExt;
+use std::{convert::TryFrom, fs::File, io::{self, Read}, path::Path, process};
 
+pub async fn sub_svc_set<U>(
```
agadgil-progress: `pub(crate)` is not working?

components/hab/src/command/config.rs
```diff
@@ -1 +1,165 @@
...
+            }
+        }
+    }
+
+    ui.end("Applied configuration")?;
+
+    Ok(())
+}
+
+pub async fn sub_svc_config(
```
agadgil-progress: `pub(crate) is not working?

components/hab/src/cli_v4/config.rs
```diff
@@ -0,0 +1,31 @@
...
+pub enum ConfigCommand {
```
agadgil-progress: `pub(crate)` visibility.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Debug, Clone, Parser)]
+#[command(
+    arg_required_else_help = true,
+    rename_all = "kebab-case",
+    help_template = "{name} {version} {author-section} {about-section}\n\
+                     {usage-heading} {usage}\n\n{all-args}\n",
+    about = "Apply a configuration to a running service"
+)]
+pub struct ConfigApplyOptions {
```
agadgil-progress: pub(crate) visibility.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Supervisor control address (overrides HAB_SUP_CTL_ADDR)
+    #[command(flatten)]
+    pub remote_sup: RemoteSup,
```
agadgil-progress: These all should be private?

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[arg(long, default_value_t = 0)]
+    pub version: u64,
+
+    /// Encrypt the payload for this username
+    #[arg(long)]
+    pub user: Option<String>,
+}
+
+impl ConfigApplyOptions {
+    pub async fn do_apply(&self, ui: &mut UI) -> HabResult<()> {
```
agadgil-progress: pub(crate)

components/hab/src/cli_v4/config/show.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Debug, Clone, Parser)]
+#[command(
+    arg_required_else_help = true,
+    rename_all = "kebab-case",
+    help_template = "{name} {version} {author-section} {about-section}\n\
+                     {usage-heading} {usage}\n\n{all-args}\n",
+    about = "Show the current config of a running service"
+)]
+pub struct ConfigShowOptions {
```
agadgil-progress: pub(crate)

components/hab/src/cli_v4/config/show.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Remote Supervisor control address (overrides HAB_SUP_CTL_ADDR)
+    #[command(flatten)]
+    pub remote_sup: RemoteSup,
```
agadgil-progress: These all should be private visibility.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,53 @@
...
+    about = "Apply a configuration to a running service"
+)]
+pub struct ConfigApplyOptions {
+    /// Supervisor control address (overrides HAB_SUP_CTL_ADDR)
+    #[command(flatten)]
+    pub remote_sup: RemoteSup,
+
+    /// Service group identifier, e.g. `core/redis.default`
+    #[arg(long)]
+    pub group: String,
```
agadgil-progress: This is an `argument` and not an option, hence you will need to remove the `long` there and add the name for the argument `SERVICE_GROUP` (should use the same name for attribute ie. `service_group`).

agadgil-progress: Check how `v2` CLI makes use of this and some other arguments.

components/hab/src/cli_v4/config/apply.rs
agadgil-progress, line 30: This is also an `argument` and not `option` , you may have to rename it as above. (Also you will have to use a value parser similar to `pkg/upload.rs` ).  The `FileExistsOrStdinValueParser` is already implemented, you'd need to use that.

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Path to the config file ("-" for stdin)
+    #[arg(long)]
```
agadgil-progress: This is also an argument and not option , you may have to rename it as above. (Also you will have to use a value parser similar to `pkg/upload.rs` ). The `FileExistsOrStdinValueParser` is already implemented, you'd need to use that.

components/hab/src/cli_v4/config/apply.rs
agadgil-progress: This is an argument and not an option. Also you may not want to use the default value for this. Please refer to `v2` usage. Also, this argument should be defined *before* the `file` argument above, so that the order is preserved.

components/hab/src/command/config.rs
sougata-progress: making this `pub(crate)` is resulting in clippy errors

agadgil-progress: This happens because in the `v2` feature, these functions are not used. One way to fix this issue is - 

```rust 


#[cfg(feature = "v4")]
pub mod config;

```

components/hab/src/cli_v4/config/apply.rs
```diff
@@ -0,0 +1,51 @@
...
+                     ui::UI};
+use std::path::PathBuf;
+
+#[derive(Debug, Clone, Parser)]
+#[command(arg_required_else_help = true,
+          rename_all = "kebab-case",
+          help_template = "{name} {version} {author-section} {about-section}\n{usage-heading} \
+                           {usage}\n\n{all-args}\n",
+          about = "Apply a configuration to a running service")]
+pub(crate) struct ConfigApplyOptions {
```
agadgil-progress, line 16: The help for the attributes has different text between v2 and v4. For example -

1. v4 output

```rust
Arguments:
  <SERVICE_GROUP>   Service group identifier, e.g. `core/redis.default`
  <VERSION_NUMBER>  Configuration version number to set
  <FILE>            Path to the config file ("-" for stdin)

Options:
  -r, --remote-sup <REMOTE_SUP>  Address to a remote Supervisor's Control Gateway [default: 127.0.0.1:9632]
      --user <USER>              Encrypt the payload for this username
  -h, --help                     Print help
  -V, --version                  Print version

```

2. v2 output - 
```rust
OPTIONS:
        --cache-key-path <CACHE_KEY_PATH>    Cache for creating and searching for encryption keys [env:
                                             HAB_CACHE_KEY_PATH=]  [default: /home/ubuntu/.hab/cache/keys]
    -r, --remote-sup <REMOTE_SUP>            Address to a remote Supervisor's Control Gateway [default: 127.0.0.1:9632]
    -u, --user <USER>                        Name of a user key to use for encryption

ARGS:
    <SERVICE_GROUP>     Target service group service.group[@organization] (ex: redis.default or foo.default@bazcorp)
    <VERSION_NUMBER>    A version number (positive integer) for this configuration (ex: 42)
    <FILE>              Path to local file on disk (ex: /tmp/config.toml, default: <stdin>)
```

In `v4` `CACHEY_KEY_PATH` is missing and help text for other format options appear to be different. Also in the `v2` usage, `FILE` argument is temporary and in our case it is permanent. But if we explicitly ask for `-` for `<stdin>` that might be okay.

PR #9749, opened 2025-07-09
documentation updates to reflect move to chef origin

mwrock:


jasonheath (approved):
@IanMadd should ought do a final review but these changes all make sense to me if @mwrock wants to go ahead and merge.

IanMadd (approved):

PR #9751, opened 2025-07-09
update tests to run on aarch64

sajjaphani:


jasonheath (approved):
Looks good. Runs to the same effect as its amd64 counterpart.  Part of me would like to see the amd64 version use Docker too given the work done here but that can be done in the future if its found to be desireable.

PR #9752, opened 2025-07-09
Updated `hab file` to use clap v4

sougata-progress:
[CHEF-23448](https://progresssoftware.atlassian.net/browse/CHEF-23448)

[CHEF-23448]: https://progresssoftware.atlassian.net/browse/CHEF-23448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/hab/src/cli_v4/file.rs
+++ b/components/hab/src/cli_v4/file.rs
@@ -1,0 +1,25 @@
+use crate::error::Result as HabResult;
+use clap::Subcommand;
+use clap_v4 as clap;
+use habitat_common::ui::UI;
+
+mod upload;
+use upload::FileUploadOptions;
+
+#[derive(Debug, Clone, Subcommand)]
+#[command(about = "Commands relating to Habitat files",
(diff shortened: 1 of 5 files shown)


components/hab/src/cli_v4/file.rs
```diff
@@ -0,0 +1,26 @@
...
+use clap::Subcommand;
+use clap_v4 as clap;
+use habitat_common::ui::UI;
+
+mod upload;
+use upload::FileUploadOptions;
+
+#[derive(Debug, Clone, Subcommand)]
+#[command(about = "Commands relating to Habitat files",
+          override_usage = "hab file <SUBCOMMAND>",
```
agadgil-progress: Just curious - why `override_usage` ? Not wrong per se, but just trying to understand and then if it makes sense whether we should consider it for other subcommands.

PR #9764, opened 2025-07-15
Updated `hab cli` to use clap v4

sougata-progress:
[CHEF-23450](https://progresssoftware.atlassian.net/browse/CHEF-23450)

[CHEF-23450]: https://progresssoftware.atlassian.net/browse/CHEF-23450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/hab/src/cli_v4/config/apply.rs
+++ b/components/hab/src/cli_v4/config/apply.rs
@@ -14,8 +14,6 @@ use habitat_common::{cli::clap_validators::FileExistsOrStdinValueParser,
                            {usage}\n\n{all-args}\n",
           about = "Sets a configuration to be shared by members of a Service Group")]
 pub(crate) struct ConfigApplyOptions {
-    /// Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=]  [default:
-    /// /hab/cache/keys]
     #[command(flatten)]
     cache_key_path: CacheKeyPath,
 
--- a/components/hab/src/cli_v4/config/apply.rs
+++ b/components/hab/src/cli_v4/config/apply.rs
@@ -15,7 +15,7 @@ use habitat_common::{cli::clap_validators::FileExistsOrStdinValueParser,
           about = "Sets a configuration to be shared by members of a Service Group")]
 pub(crate) struct ConfigApplyOptions {
     /// Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=]  [default:
-    /// /home/ubuntu/.hab/cache/keys]
+    /// /hab/cache/keys]
     #[command(flatten)]
     cache_key_path: CacheKeyPath,
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -685,6 +685,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.5.55"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


components/hab/src/cli_v4/config/apply.rs
```diff
@@ -15,7 +15,7 @@ use habitat_common::{cli::clap_validators::FileExistsOrStdinValueParser,
           about = "Sets a configuration to be shared by members of a Service Group")]
 pub(crate) struct ConfigApplyOptions {
     /// Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=]  [default:
-    /// /home/ubuntu/.hab/cache/keys]
```
agadgil-progress, line 18: Wondering why did this get changed?

sougata-progress: In the V2 commands the default path is /hab/cache/keys. That's why

agadgil-progress: Actually these values are taken directly from the `env` field attribute and `default` field attribute in the derive. So no need to explicitly define it. 

I believe you saw `/hab/cache/keys` because you were probably running that command as root. In fact you can remove it from the help completely and will still be taken properly using the `clap` `derive` values

agadgil-progress: For Example - 

```rust
$ sudo ./target/debug/hab pkg build 
hab-pkg-build 2.0.160  
Builds a plan using Habitat Studio
 
Usage: hab pkg build [OPTIONS] <PLAN_CONTEXT>

Arguments:
  <PLAN_CONTEXT>  A directory containing a plan file or a `habitat/` directory which contains the plan file

Options:
  -k, --keys <HAB_ORIGIN_KEYS>             Installs secret origin keys (ex: "unicorn", "acme,other,acme-ops")
  -r, --root <HAB_STUDIO_ROOT>             Sets the Studio root (default: /hab/studios/<DIR_NAME>)
  -s, --src <SRC_PATH>                     Sets the source path [default: $PWD]
      --cache-key-path <CACHE_KEY_PATH>    Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=] [default: /hab/cache/keys]
  -N, --native-package                     Build a native package on the host system without a studio
  -R, --reuse                              Reuses a previous Studio for the build (default: clean up before building)
  -D, --docker                             Uses a Dockerized Studio for the build
  -f, --refresh-channel <REFRESH_CHANNEL>  Channel used to retrieve plan dependencies for Chef supported origins [env: HAB_REFRESH_CHANNEL=] [default: stable]
  -h, --help                               Print help
  -V, --version                            Print version
```

agadgil-progress: Also, please try to fix the same in `cli` and `file` - for example following - 

```shell

components/hab/src/cli_v4/cli/setup.rs:    /// Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=]  [default:
components/hab/src/cli_v4/config/apply.rs:    /// Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=]  [default:
components/hab/src/cli_v4/file/upload.rs:    /// Cache for creating and searching for encryption keys [env: HAB_CACHE_KEY_PATH=]  [default:
```

agadgil-progress: Alright to do in the same PR, but in a commit of it's own probably.

PR #9769, opened 2025-07-17
fixes the release pipeline with chef origin

mwrock:


components/plan-build-ps1/bin/hab-plan-build.ps1
```diff
@@ -340,8 +340,10 @@ function Install-Dependency($dependency, $install_args = $null) {
         $origin = $dependency.Split("/")[0]
         $channel = $env:HAB_BLDR_CHANNEL
         $ignoreLocal = ""
-        if ($origin -eq "core") {
-            $channel="$env:HAB_REFRESH_CHANNEL"
+        if ($origin -eq "core" -or $origin -eq "chef") {
+            if ($origin -eq "core") {
+                $channel = $env:HAB_REFRESH_CHANNEL
```
jasonheath, line 345: This logic here and in the linux shell version is the one question I have about this PR. In thinking about it I'm sure you're correct but I'm still going to ask "why don't we want to set the HAB_REFRESH_CHANNEL for the chef origin?"

mwrock: good question. It's because the chef origin packages are not part of a "refresh." That word, "refresh," may not be the greatest choice of words but it is what we have historically called the set of core deps we support that are included in the core origin.

In this change, what I needed to do was to make sure we downloaded previous hab packages built earlier in the release pipeline (now in the chef origin) even if a package existed locally. For example, plan-build was not pulling in the newly built hab package but was using the one already installed with the current studio.

PR #9779, opened 2025-07-22
Fixed `HAB_AUTH_TOKEN` bug

sougata-progress:
[CHEF-23525](https://progresssoftware.atlassian.net/browse/CHEF-23525)


agadgil-progress:
@sougata-progress : I believe this problem exists on some of the `hab` subcommands ported to `cli_v4` do you mind checking those and adding those as a part of this? In fact any structure that takes `-z` needs to handle this `maybe_` we are handling the `env` or cli switch, but we may need to handle stuff from `cli.toml` as well.

PR #9785, opened 2025-07-25
fixes e2e tests broken by chef origin and builder restrictions

mwrock:


.expeditor/scripts/end_to_end/setup_environment.sh
```diff
@@ -40,7 +40,7 @@ echo "--- Installing latest core/netcat from ${HAB_BLDR_URL}, stable channel"
 sudo -E hab pkg install core/netcat \
     --binlink \
     --force \
-    --channel="stable" \
+    --channel="base-2025" \
```
agadgil-progress, line 43: Not sure, Why is this required from the `base-2025` channel? Also a *nit* the comment above should be changed.

agadgil-progress (approved):
One really small nitpick and a question. Based on my understanding this looks alright.

.expeditor/scripts/end_to_end/setup_environment.sh
mwrock: It's not required but I wanted to start using base-2025 over stable where we can. Will update the comment.

PR #9806, opened 2025-08-06
cli tests for `hab-sup` binary

agadgil-progress:


components/hab/src/cli_v4/sup/sup_run.rs
```diff
@@ -292,3 +323,185 @@ impl SupRunOptions {
...
+                return Err(HabError::ArgumentError(format!("'{}' is not a valid \
+                                                            toml config file.",
+                                                           config_file.display())));
+            }
+        }
+
+        Ok(other)
+    }
+
+    #[allow(clippy::cognitive_complexity)]
```
agadgil-progress, line 372: The idea here is to merge with another one - if the current one is using the `default`. Note: the other could just be using default and override the `self` (which is a bit wasteful but not wrong). 

I don't think this is the most ideal way to do it - if there's a better way to do it, would surely be keen to do that.

dikshagupta1 (approved):

dikshagupta1:
@agadgil-progress I have approved the PR. I think you need to make a commit with `-s` flag for the DCO check to pass.

PR #9816, opened 2025-08-12
Replaced rusoto with aws-sdk and rust update to 1.89

sougata-progress:
Updated rust to 1.89 as it is required to run the sdk

### DIFF
--- a/components/pkg-export-container/src/lib.rs
+++ b/components/pkg-export-container/src/lib.rs
@@ -131,7 +131,7 @@ impl Credentials {
                 let token = client.get_authorization_token()
                                   .send()
                                   .await
-                                  .map_err(Error::TokenFetchFailed)
+                                  .map_err(|e| Error::TokenFetchFailed(Box::new(e)))
                                   .and_then(|resp| {
                                       resp.authorization_data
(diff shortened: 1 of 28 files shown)


components/pkg-export-container/src/lib.rs
```diff
@@ -116,11 +114,20 @@ impl Credentials {
...
+                                                                       password.to_string(),
+                                                                       None,
+                                                                       None,
+                                                                       "static"));
                 // TODO TED: Make the region configurable
-                let client = EcrClient::new_with(HttpClient::new()?, provider, Region::UsWest2);
-                let auth_token_req = GetAuthorizationTokenRequest {};
-                let token = client.get_authorization_token(auth_token_req)
+                let loader = aws_config::defaults(BehaviorVersion::latest())
+                    .region(ecr::config::Region::new("us-west-2".to_string()))
```
agadgil-progress: Pretty much the only comment - if `us-west-2` is the default we want to have - better to have it as some constant, else we will not remember later if something gets changed where in the code `us-west-2` is coming from?

mwrock (approved):

PR #9858, opened 2025-08-23
Add GitHub Security Actions Workflow

sean-sype-simmons:
Pr is needed for enabling the required security scans workflow. https://github.com/chef/common-github-actions/blob/main/.github/workflows/ci-main-pull-request.yml


mwrock:
Looks like this is missing the DCO

PR #9872, opened 2025-08-28
remove v1.6 from the release_branch so that it will be added back again

muthuja:


jasonheath (approved):

jasonheath:
Muthuja Bavaji Vempalli requested this to be done.  Given that it was acting on expeditor config and related to work he was actively engaged in I bypassed and merged.

PR #9892, opened 2025-09-02
Creates HAB-2-README.md and documents handlebars upgrade

jasonheath:
This PR creates a new file named HAB-2-README.md and documents the effects of the handlebars upgrade detailing the breaking changes and steps users will need to take to mitigate them.

### DIFF
--- a/HAB-2-README.md
+++ b/HAB-2-README.md
@@ -1,0 +1,27 @@
+# Habitat 2.0.x README
+
+## Handlebars Crate Upgraded
+
+The [handlebars crate](https://crates.io/crates/handlebars) was upgraded from an early version that habitat had pinned due to breaking changes after habitat was released for general use.  The code base was upgraded to the most recent version available at the time the work was done.  Also it has been, and will continue to be, updated as new releases of the crate become available.
+
+The impact of this is that you may have to update your templates as described in the following sections.
+
+### Object Access Syntax Removed
+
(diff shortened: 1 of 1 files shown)


mwrock:
Do we plan on adding to this doc? I wonder if we should call it something like breaking_changes or just handlebars_update if this is all this is going to include.

jasonheath:
> Do we plan on adding to this doc? I wonder if we should call it something like breaking_changes or just handlebars_update if this is all this is going to include.

I created this file thinking that we could document all of our hab 2 related bits in it so what's here would be the first of many pieces that people would add.  We could still do that but since then I've been wondering if all such documentation should only be added to the new docs repo or if we want to add it to a document such as this and the new repo.  Personally, I think one place and that it should be the new doc repo. With that in mind I would move this doc to the new doc repo and close this PR unmerged. What do you think @mwrock?

mwrock:
> > Do we plan on adding to this doc? I wonder if we should call it something like breaking_changes or just handlebars_update if this is all this is going to include.
> 
> I created this file thinking that we could document all of our hab 2 related bits in it so what's here would be the first of many pieces that people would add. We could still do that but since then I've been wondering if all such documentation should only be added to the new docs repo or if we want to add it to a document such as this and the new repo. Personally, I think one place and that it should be the new doc repo. With that in mind I would move this doc to the new doc repo and close this PR unmerged. What do you think @mwrock?

yeah that sounds good.

jasonheath:
Superceded by https://github.com/habitat-sh/chef-habitat-docs/pull/15

PR #9941, opened 2025-09-24
hab plan CLI fixes

jasonheath:
- Fixes `hab plan render` by fixing the default.toml parsing
- Fixes parsing of the argument passed to --scaffolding restoring scaffold plan generation
- Restores scaffolding functionality on windows
- Updates template_plan.sh and template_plan.ps1 to fix `hab plan init`
  - Removes almost all use of `{{~` and `~}}` to stop lines from running together after handlebars upgrade
  - introduces newlines to separate settings that were previously running together
  - Upcases 'Required.' and 'Optional.' and drops the period
  - Replaces no longer available doc URLs with current viable doc URLs
  - Normalizes whitespace in the handlebars templates as it was inconsistent
- requires --scaffolding for a non-placeholder scaffoding setting
  - for some reason the code was always defaulting to "core/scaffolding-ruby" 
- add an end-to-end test for `hab plan init` and `hab plan render`

### DIFF
--- a/components/hab/src/cli_v4/plan/init.rs
+++ b/components/hab/src/cli_v4/plan/init.rs
@@ -1,13 +1,9 @@
-// Implementation of `hab plan init` subcommand
-use std::str::FromStr;
-
 use clap_v4 as clap;
 
 use clap::Parser;
 
 use habitat_common::ui::UI;
-use habitat_core::{origin::Origin,
-                   package::PackageIdent};
+use habitat_core::origin::Origin;
 
 use crate::error::Result as HabResult;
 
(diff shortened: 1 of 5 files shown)


jasonheath:
I plan to add a e2e test for the hab plan cmd but I think the code is ready for review.

agadgil-progress:
So may be we should get rid of `--scaffolding` for now for windows and then we can plan and add that support later.

mwrock:
I think this brings us closer to 1.6 parity. Not having support for `--scaffolding` is technically a breaking change. There are windows scaffolding packages in the effortless repo. The only difference I see here with 1.6 is that it supports non-ident scaffolding values (ex. `go` and not just `core/scaffolding-go`). This really isn't necessary but it is not breaking and I think it's fine.

PR #9958, opened 2025-09-30
Fixes `hab pkg ...` testing

agadgil-progress:
Fixed Usage for various subcommands

1. Added details to `-z` about the env variable and config file.
2. Added details to `-u` about the env variable and config file.
3. Removed unsupported options from `bulkupload` and `ulpoad`
4. Fixed typos

### DIFF
--- a/components/hab/src/cli_v4/pkg/bulk_upload.rs
+++ b/components/hab/src/cli_v4/pkg/bulk_upload.rs
@@ -13,8 +13,6 @@ use habitat_common::{cli::clap_validators::DirExistsValueParser,
 use habitat_core::{crypto::keys::KeyCache,
                    ChannelIdent};
 
-use habitat_api_client::BuildOnUpload;
-
 use crate::{cli_v4::utils::{AuthToken,
                             BldrUrl},
             command::pkg::bulkupload,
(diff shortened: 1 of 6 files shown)


components/hab/src/cli_v4/pkg/bulk_upload.rs
```diff
@@ -43,7 +43,7 @@ pub(crate) struct PkgBulkUploadOptions {
 
     // TODO: This option is to be removed?
     /// Enable auto-build for all packages in this upload. Only applicable to SaaS Builder
-    #[arg(name = "AUTO_BUILD", long = "auto-build", action = ArgAction::SetTrue)]
+    #[arg(name = "AUTO_BUILD", long = "auto-build", hide = true, action = ArgAction::SetTrue)]
```
mwrock: can we completely remove this instead of hiding it?

components/hab/src/cli_v4/pkg/upload.rs
```diff
@@ -43,7 +43,7 @@ pub(crate) struct PkgUploadOptions {
     force: bool,
 
     /// Disable auto-build for all packages in this upload
-    #[arg(name = "NO_BUILD", long = "no-build", action = ArgAction::SetTrue)]
+    #[arg(name = "NO_BUILD", long = "no-build", hide = true, action = ArgAction::SetTrue)]
```
mwrock: same comment as above. We should remove instead of hiding it.

mwrock (changes requested):

mwrock (approved):

PR #9967, opened 2025-10-02
Adds test_hab_plan_cli to e2e pipeline

jasonheath:
Checked the e2e execution and realized that the new test still needed to be added to the e2e pipeline.


jasonheath:
One step is failing on the verify pipeline but these changes won't have any effect until the end-to-end pipeline runs so they should be unrelated.

PR #9977, opened 2025-10-06
Removes 'succeeds with autobuild' from test_pkg_bulkupload.ps1

jasonheath:
Removes 'succeeds with autobuild' from test_pkg_bulkupload.ps1 as --auto-build was disabled in the context of hab pkg bulkupload


jasonheath:
This is the isolated removal of a test step from an e2e test so I'm pushing it a bit quicker and not waiting for the verify pipeline to complete as i'm doing this for other things I'm about to push.

PR #9978, opened 2025-10-06
Corrects casing for write-host for psscriptanalyzer

jasonheath:


jasonheath:
This is the isolated correction of case to make PSScriptAnalyzer happy again (and it's passed) so I'm pushing it a bit quicker and not waiting for the full verify pipeline to complete as I'm doing this for other things I'm about to push.

PR #10357, opened 2026-04-22
thread token passed from --auth through to check for interpreter

mwrock:


agadgil-progress (approved):
Looks good - couldn't test the end-to-end test, for that we've to wait till everything goes through the pipelines.

PR #10385, opened 2026-05-11
Fixes install.sh for Darwin aarch64 

agadgil-progress:
Also added fixes to e2e test failures.


mwrock (commented):
wouldn't this break anyone trying to download a stable aarch64 mac cli today? That won't support `hab pkg install` right?

agadgil-progress:
> wouldn't this break anyone trying to download a stable aarch64 mac cli today? That won't support `hab pkg install` right?

Oh yes!! that's true. Will fix that

mwrock (approved):

PR #10396, opened 2026-05-15
Fully automate SBOM generation and import to blackduck

mwrock:
The standard blackduck rust plugin uses a simple `cargo tree` to determine crate dependencies. This misses 2 key classes of crates:
1. Platform specific crates that are specific to a platform where `cargo tree` is not being run.
2. stdlib crates compiled into the rust toolchain
This change catches all of these crates and produces a SBOM that it uploads to blackduck.

This also adds all of the habitat package components. For now these are imported from a static list that we have t omanually maintain but at least we don't have to add them in the UI which is very tedious and they do not change often.

Lastly this bumps the minor version to 2.1.


.github/workflows/ci-main-pull-request-stub.yml
agadgil-progress: Wondering whether we can make use of this and the next setting to make sure our version is always latest? What I couldn't figure out is - how to use this to set the `BD_VERSION_NAME` env below.

agadgil-progress: Like `detect-version-source-type: 'file'` and `detect-version-source-parameter: 'VERSION'`

agadgil-progress (commented):
More like a question as I don't understand much about BlackDuck.

.github/workflows/ci-main-pull-request-stub.yml
```diff
@@ -107,4 +107,241 @@ jobs:
...
+      # Note: cross-compilation is NOT required here — cargo-cyclonedx uses
+      # `cargo metadata --filter-platform` which only needs the target registered,
+      # not a full cross toolchain.
+      - name: Install target triples
+        run: |
+          rustup target add \
+            x86_64-unknown-linux-gnu \
+            x86_64-pc-windows-msvc \
+            aarch64-apple-darwin \
+            x86_64-apple-darwin
```
agadgil-progress, line 172: Is there a reason why we are not considering `aarch64-linux`?

.github/workflows/ci-main-pull-request-stub.yml
```diff
@@ -107,4 +107,241 @@ jobs:
...
+
+      - name: Set BD_VERSION_NAME from VERSION file
+        run: echo "BD_VERSION_NAME=$(cat VERSION)" >> "$GITHUB_ENV"
+
+      # Use the same toolchain pinned in rust-toolchain
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        id: toolchain
+        with:
+          toolchain: '1.91.1'
```
agadgil-progress: A nitpick here: Can we get the version from the `rust-toolchain` file?

agadgil-progress (commented):
Again more like questions - 

1. Is there a reason why we are not generating for `aarch64-linux` ? 
2. Also, can we get Rust version from the `rust-toolchain` file (similar to how `BD_VERSION_NAME` is done?)

agadgil-progress (approved):

PR #10416, opened 2026-06-01
Removed core/docker from pkg export tar plan.ps1

sougata-progress:


components/pkg-export-tar/habitat/plan.ps1
```diff
@@ -15,7 +14,7 @@ $pkg_build_deps = @(
 )
 
 function pkg_version {
-    Get-Content "$SRC_PATH/../../VERSION"
+    Get-Content "$SRC_PATH/../../../VERSION"
```
mwrock: This does not look right. It deviates from all the other windows plans.

mwrock (approved):
preemptively approving assuming we fix the version path.

components/pkg-export-tar/habitat/plan.ps1
sougata-progress: fixed

PR #10417, opened 2026-06-03
Using core/libiconv as a dependency on macOS

agadgil-progress:
This pull request updates the `plan.sh` for the `aarch64-darwin` Habitat package to ensure that the build process uses the Habitat-managed `libiconv` package instead of relying on the system version. The most important changes are:

Dependency management improvements:
* Added `core/libiconv` to the `pkg_deps` array to ensure the Habitat package uses its own managed version of `libiconv` rather than the system-provided one.

Build process updates:
* Set the `RUSTFLAGS` environment variable in the `do_prepare()` function to include the library path for `libiconv`, ensuring that Rust links against the correct library during the build.

### DIFF
--- a/components/hab/habitat/aarch64-darwin/plan.sh
+++ b/components/hab/habitat/aarch64-darwin/plan.sh
@@ -8,7 +8,7 @@ pkg_license=('Apache-2.0')
 # This is getting resolved from the system, while we have a habitat package for that
 # so we shoud use the one from 'our' habitat package.
 pkg_deps=(
-	core/libiconv
+    core/libiconv
 )
 
 pkg_build_deps=(
@@ -101,7 +101,8 @@ do_prepare() {
     build_type="--release"
     build_line "Building artifacts with \`${build_type#--}' mode"
 
-    export RUSTFLAGS="-L $(pkg_path_for libiconv)/lib"
+    RUSTFLAGS="${RUSTFLAGS:+$RUSTFLAGS }-L $(pkg_path_for libiconv)/lib"
+    export RUSTFLAGS
     build_line "Setting RUSTFLAGS=$RUSTFLAGS"
 }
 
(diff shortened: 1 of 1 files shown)


mwrock (commented):
How will this work since hab is (or should be) a static binary?

agadgil-progress:
> How will this work since hab is (or should be) a static binary?

It is not a static binary on macOS. Basically it is a massive pain to build statically linked binaries.

I just checked even our 1.6 hab binaries on macOS (aarch64) are dynamically linked.

mwrock:
Then we will need to somehow account for this in the installer right? Because it currently assumes that we initially download a single hab binary Will that hab binary be able to do the `hab pkg install` of the hab package without the core/libiconv package?

agadgil-progress:
> Then we will need to somehow account for this in the installer right? Because it currently assumes that we initially download a single hab binary Will that hab binary be able to do the `hab pkg install` of the hab package without the core/libiconv package?

We need to test it - but that should work - here is how it is - for versions *after 2.0.504* we indeed do a `hab pkg install` of the latest hab binary that should pull the dependent `core/libiconv` . Or am I missing something? For older versions, it will continue to use the libiconv on the system.

agadgil-progress:
> > Then we will need to somehow account for this in the installer right? Because it currently assumes that we initially download a single hab binary Will that hab binary be able to do the `hab pkg install` of the hab package without the core/libiconv package?
> 
> We need to test it - but that should work - here is how it is - for versions _after 2.0.504_ we indeed do a `hab pkg install` of the latest hab binary that should pull the dependent `core/libiconv` . Or am I missing something? For older versions, it will continue to use the libiconv on the system.

Or you think it's not worth going down that path and let it continue to use the system libiconv as it is doing today, because we know that path works? May be that's a good idea too.

mwrock:
I'm not sure. Basically when this is released, when a user installs hab 2.1, they will first just pull down the hab binary as a single file. Then they run `hab pkg install` to populate their /opt/hab folder. So when they get the initial hab 2.1 file, unless we change te install behavior, it will be missing the core/libiconv binaries and I am wondering if that will render the hab binary useless if it is looking for the absolute /opt/hab/pkgs path to link to it.

agadgil-progress:
As discussed - closing this PR

PR #10534, opened 2026-08-18
use previous buildah package to fix e2e pipeline

mwrock:


dikshagupta1 (approved):
Looks good. But is the pipeline suppose to fail?

